PR #4, opened 2017-09-14
Expirement with default url scheme

deadtrickster:
Hi!

Simple helper to add "http://" if no scheme found in the url.
Probably a good idea to make it configurable and move to parameter.
Added [quri](https://github.com/fukamachi/quri) to dependencies. Used for parsing url. 
Let me know if it's the right direction.

### DIFF
--- a/next/lisp/document-mode.lisp
+++ b/next/lisp/document-mode.lisp
@@ -51,7 +51,14 @@
 	(|setUrl| (buffer-view buffer) url)))
 
 (defun set-url (input-url)
-  (set-url-buffer input-url *active-buffer*))
+  (let ((url (maybe-add-default-url-scheme input-url)))
+    (set-url-buffer url *active-buffer*)))
+
+(defun maybe-add-default-url-scheme (input-url)
+  (let ((url (quri:uri input-url)))
+    (if (quri:uri-scheme url)
+        input-url
+        (concatenate 'string "http://" input-url ))))
 
 (defun document-mode ()
   "Base mode for interacting with documents"
--- a/next/dependencies.lisp
+++ b/next/dependencies.lisp
@@ -1,2 +1,3 @@
 (ql:quickload :cl-strings)
 (ql:quickload :cl-string-match)
+(ql:quickload :quri)
--- a/next/next.asd
+++ b/next/next.asd
@@ -2,7 +2,7 @@
 ;;;; next.asd
 (defsystem :next
   :serial t
-  :depends-on (:cl-strings :cl-string-match)
+  :depends-on (:cl-strings :cl-string-match :quri)
   :components ((:file "lisp/package")
 	       (:file "lisp/macro")
 	       (:file "lisp/qt")


next/lisp/document-mode.lisp
```diff
@@ -51,7 +51,14 @@
 	(|setUrl| (buffer-view buffer) url)))
 
 (defun set-url (input-url)
-  (set-url-buffer input-url *active-buffer*))
+  (let ((url (maybe-add-default-url-scheme input-url)))
+    (set-url-buffer url *active-buffer*)))
+
+(defun maybe-add-default-url-scheme (input-url)
```
jmercouris: the name could maybe something like format-url

jmercouris:
Thank you very much for your contribution!!

jmercouris:
Also to answer your question, I think it is the right direction, I've since changed the name a little bit on master, in time we'll want to make a more intelligent mini buffer completion based on the URL, which is why in version 0.03 we'll be adding history in the form of a sqlite database.

Version 0.02 is almost done, at which point we'll kind of finalize the buffer architecture, and we can hopefully start on the 0.03 features.

I'd be very interested to hear about your ideas for version 0.03 history and bookmark persistence, as it is something I am still thinking a lot about.

Please keep in mind that for the history and bookmarks, ultimately it'd be great to support intelligent mini-buffer completion for anything about the bookmark

jmercouris:
Also, quri, unfortunately breaks the build system because it depends on ASDF being available in the binary form, we might have to substitute quri for something of our own, or use a diff library, please see the below conversation

`> hey everyone, I'm getting a strange error when I run my program:
    https://pastebin.com/AYRtnk5D
> any idea what that means?
<pjb> check the backtrace.
<pjb> Somebody's calling ASDF/SYSTEM::SYSTEM-RELATIVE-PATHNAME but
    it's not defined.
<pjb> It may be a problem of incompatible asdf version. Or not having
    loaded asdf early enough?
> right, but that's the thing, I'm never calling that function
> can it be possible that one of my dependencies is calling it?
<pjb> indeed.
<pjb> Hence the use of the backtrace.
> what is the backtrace?
> what do you mean?
> aha, I found it now, it's in one of my dependencies, how can I fix
    it?
<pjb> the list of stacked function calls.
> I thought that was called a stacktrace
<pjb> same thing.
<edgar-rft> the debugger should spit out a list of the functions that
    were called before the error happened or at least have an option
    to produce such a list
> unfortunately there were no functions called beforehand
> it seems to be at the base level of the file
> I'm looking at it here
> https://github.com/fukamachi/quri/blob/master/src/etld.lisp
<edgar-rft> (defvar *default-etld-names*
    (asdf:system-relative-pathname :quri
    #P"data/effective_tld_names.dat")) probably calls that fuction
    internally somewhere
<edgar-rft> have you tried if the same error happens if you evaluate
    (asdf:system-relative-pathname :quri
    #P"data/effective_tld_names.dat")
> edgar-rft: right yeah
<pjb> jmercouris: possibly, the problem here is that defvar doesn't
    evaluate the initial-value at compilation-time, but at
    (:load-toplevel :execute) time.
<pjb> jmercouris: (similarly to the call to the load-etld-data line
    40.
<pjb> )
> I know the function will work normally on my machine, I'm able to
    load and use the library just fine
<pjb> jmercouris: what implementation are you using?
> I'm using ecl
<pjb> I guessed it.  With ecl, the load-toplevel occurs when you run
    the binary.
> but what is happening is that I'm basically making a monolithic lib
    using asdf, and I am embedding it into an executable, so asdf is
    not available in that executable, causing this issue
<pjb> At that time, asdf is not loaded.
> pjb: so, ideas?
<pjb> And it's a tad difficult to include asdf in the binary, because
    it may be loaded/reloaded by the rc-file at run-time.
> could I just redefine the symbol to do nothing?
<pjb> jmercouris: the problem here is that you cannot make an asdf
    system depend on asdf.
> pjb: Right yeah, I was actually thinking down that path at first
<pjb> So when ecl uses asdf to compile and load systems, it doesn't
    compile and load asdf.
`

deadtrickster:
Hi, unfortunately paste expired. But I'm on #lisp too. Not at 4am though :-)
Probably 90% times it is ok to just prepend "http://", however (as always) there is a catch - spec is somewhat more complex and to properly detect schema we need a parser. 
Also we need to catch unsupported scheme too.
I'll try to build binary myself, but please if you still have that paste around, upload it again.

SQLite seems to be ok. I'd like to match url and title, so it's kind of implicit history search.

jmercouris:
Here's the error originally in the pastebin:

screen shot 2017-09-15 at 22 02 42

jmercouris:
In terms of the error, it's in that file I linked to in the conversation about QURI, it's trying to load a .dat file that contains a bunch of entries of domain name suffixes. It is using ASDF to do this kind of loading for some reason, apparently to get a path relative to the library during runtime.

The problem is that we need the .dat file for the library to function, so we either need to examine exactly what part of that code does, or maybe try to find another library, not sure.

Issue #8, opened 2017-10-13
Non-existent file ~/.next.d/bookmark.db


eadmund:
When I attempt to run nEXT for the first time, I get an error:

```
…
[package sqlite]..
; Loading "next"
[package cl-strings]..............................
[package puri]....................................
[package next].
Condition of type: FILE-ERROR
Filesystem error with pathname "~/.next.d/bookmark.db".
Either
 1) the file does not exist, or
 2) we are not allowed to access the file, or
 3) the pathname points to a broken symbolic link.

Available restarts:

1. (TRY-RECOMPILING) Recompile bookmark and try loading it again
2. (RETRY) Retry loading FASL for #<cl-source-file "next" "bookmark">.
3. (ACCEPT) Continue, treating loading FASL for #<cl-source-file "next" "bookmark"> as having been successful.
4. (ABORT) Give up on "next"

Top level in: #<process TOP-LEVEL>.
>
```

It's fixable by running `mkdir ~/.next.d`

jmercouris:
Hi, sorry for the problem! You need to update your home directory with a ~/.next.d folder. This will allow the system to create the bookmarks database, it’s a really new feature and I haven’t yet updated the manual. Please let me know if that works!

eadmund:
It should be possible to ensure that `~/.next.d` exists with `(uiop:ensure-directories-exist #"~/.next.d/")`. I'll work on a PR for this.

jmercouris:
Ok sounds good, I look forward to it :)

PR #9, opened 2017-10-13
No side effects

eadmund:
I started to hack on nEXT and noticed that the error referenced in #8 happens at when loading nEXT: `bookmark.lisp` runs INITIALIZE-BOOKMARK-DB at toplevel. I think that it'd be preferable to be able to quickload nEXT and _then_ run it, rather than having it run as a side-effect.

This might help with making things work with SLIME (which runs in a separate thread), although I could be wrong.


jmercouris:
Hi, firstly, thank you for your extensive PR!! I don't have time to go through it today, but I will try to take a look tomorrow. In terms of SLIME, that is not the issue with loading it, it's a little more complex than that and has to do with somethings going on in QT, you can see more information here: 
https://gitlab.com/eql/EQL5/tree/master/slime

next/lisp/keymap.lisp
```diff
@@ -133,5 +133,6 @@
   	    (push key-chord key-sequence)))
     key-sequence))
 
-;;; set standard keybindings
-(define-key global-map (kbd "C-x C-c") #'qquit)
+(defun initialize-keymap ()
```
jmercouris: I think with the amount of definitions that exist, this could also be pulled out of this file (as the other define-keys were), and loaded into a single lisp file called something like define-key-bindings.lisp or something like that

next/run.lisp
```diff
@@ -3,3 +3,12 @@
 (push "./" asdf:*central-registry*)
 (ql:quickload "next")
 (asdf:load-system "next")
+
+(in-package #:next)
+
+;; start nEXT
+(start)
+
+;; load the user configuration if it exists
```
jmercouris: I think it's always preferable to load a base configuration, and THEN load the user configuration to override or reset any of those changes. That way, the start of the system is more predictable

jmercouris:
Hey Bob, I just wanted to thank you very much for your support/contribution. It looks great, and hopefully it may make it easier to actually launch the program from slime!

The REPL integration with slime I was imagining originally involved launching nEXT, and then using a hotkey to start a swank server that slime would connect to, but the other way around should also be possible I suppose :)

PR #31, opened 2017-11-29
assets: fix incorrect spelling of "next" -> "nEXT"

seanfarley:
I found that the binary release didn't work due to the Info.plist not having
the correct case of the binary on the file system. Found by using
case-sensitive APFS.


jmercouris:
Hey, thank you for figuring this out!!!

PR #36, opened 2017-12-12
nEXT crashes on malformed minibuffer inputs

jnjcc:
* `puri:parse-uri` fails on malformed url strings
  `handler-case` might not be the best solution, though.
  `drakma:url-encode` might be better

* `ccl --no-init --load make.lisp` will fail by default
  because quicklisp was installed under ${HOME}/quicklisp/ by default.
  also, I replaced tabs with spaces in `make.lisp`, if that's OK with you, :-)

### DIFF
--- a/next/make.lisp
+++ b/next/make.lisp
@@ -1,12 +1,12 @@
 ;;;; make.lisp --- create binary files for nEXT
 ;;;;
-;;;; See Next/next/README.org for more information on installing the
+;;;; See nEXT/next/README.org for more information on installing the
 ;;;; dependencies necessary to build nEXT from source
 ;;;;
 ;;;; Please note that this script must be run from the directory
-;;;; Next/next.
+;;;; nEXT/next.
 
-(let ((quicklisp-init (merge-pathnames ".quicklisp/setup.lisp" (user-homedir-pathname))))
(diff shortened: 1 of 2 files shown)


jmercouris:
Firstly, thank you very much for your contribution! It looks great! I think you're right about the puri:parse-uri. I too thought about using a url-encoding library, but I couldn't settle on one, so I decided to hold off on that for now. I think your solution is definitely better than what is there now, but it is something we could work on in the future.

In terms of compilation via the `make.lisp` script, I would like it to check a fallback directory instead of just selecting the default one. It is true that normally it is in `~/quicklisp`, but a lot of people have it in `~/.quicklisp` (myself included).

Lastly, about the tabs vs spaces, I don't really mind it either way. In Python I use exclusively spaces, but in Lisp, I let M-q decide. If there is a way to make M-q in Emacs use spaces only, I can setup my configuration that way (so the Lisp code will also show up properly in other editors).

Once again, thank you so much for your interest and contribution! If you are willing to update your code for the `~/quicklisp` fallback I am ready to merge your PR, if you'd rather I make that change, that's fine too. Please let me know! Thanks!

jnjcc:
That makes sense. quicklisp path reverted.

Cool project, :-)

jmercouris:
Looks great, thank you! We'll still have to do the fall back `~/.quicklisp` and `~/quicklisp`, but for now at least random URLs won't crash the program :D

PR #41, opened 2017-12-18
switch buffer if deleting active buffer

jnjcc:
switch buffer when `delete-buffer` tries to delete the active buffer.

otherwise, the active buffer just won't disappear. I'm not quite sure
if this has anything to do with cocoa's `(#/release view)`

### DIFF
--- a/next/source/buffer.lisp
+++ b/next/source/buffer.lisp
@@ -61,22 +61,27 @@
 (defun switch-buffer-next ()
   (let ((active-buffer-index (position *active-buffer* *buffers* :test #'equalp)))
     (if (< (+ active-buffer-index 1) (length *buffers*))
-	(set-visible-active-buffer (nth (+ active-buffer-index 1) *buffers*))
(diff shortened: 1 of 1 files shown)


jmercouris:
Hi, thank you very much for your pull request! This is a long standing bug I was meaning to address!

A great way to crash the (current) compiled version is by killing the current buffer via the kill-buffer then attempting to switch to the next buffer :)

Of course this is because the *active-buffer* is no longer part of *buffers*, so it is trying to get a reference to something that does not exist, and because the compiled version of nEXT has no restarts, we get a crash.

Anyways, this PR fixes that, so thank you! One thing I might suggest though is this: instead of setting the page to *start-page-url* after killing the last buffer, I think it would make more sense to generate a new buffer and set that as the *active-buffer* (then it would have no history and be a truly fresh buffer). What do you think of this idea?

Once again, thank you very much for your time and contribution!

jnjcc:
That does seem like a better idea

PR #43, opened 2017-12-20
add page zooming functionality

jnjcc:
1. horizontal scrolling and page zooming functionality
Some webpages (like github) don't fit nEXT horizontally (for me at least), add
horizontal scrolling and page zooming functionality to make life easier

2. a workaround solution to make `kbd` recognize `C-x C--`

### DIFF
--- a/next/source/global.lisp
+++ b/next/source/global.lisp
@@ -15,6 +15,11 @@
 (defvar *hl-scroll-distance* 50
   "Horizontal scroll distance. The distance scroll-left or scroll-right
   will scroll.")
+(defvar *current-zoom-ratio* 1.0)
+(defvar *zoom-ratio-step* 0.2)
+(defvar *zoom-ratio-min* 0.2)
+(defvar *zoom-ratio-max* 5.0)
+(defvar *zoom-ratio-default* 1.0)
 (defvar *swank-port* 4006
   "The port that swank will open a new server on (default Emacs slime port
   is 4005, default set to 4006 in nEXT to avoid collisions).")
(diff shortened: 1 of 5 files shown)


next/source/global.lisp
```diff
@@ -12,6 +12,14 @@
   "A list of all existing buffers.")
 (defvar *scroll-distance* 50
   "The distance scroll-down or scroll-up will scroll.")
+(defvar *hl-scroll-distance* 50
```
jmercouris: I renamed this to `*horizontal-scroll-distance*`

jmercouris:
Hello! Once again, thank you for your contribution! This is great!

If you look at master, you'll see that I fixed the kludge by making an update I had wanted to make for a long time actually, special chars should be converted by the input system and replaced with strings like "RETURN" and "HYPHEN", if you look at the diff it will make sense

Finally, zooming actually breaks link-hints, so when you are zoomed, the position of a true object is not reveled. This is honestly not a big deal as I was thinking about how link-hints needs to be redone anyway to account for this (among many other issues with the current link-hints implementation), so it is definitely something we can get to later

Once again, thank you for your help! I hope you are enjoying nEXT!

jnjcc:
> Finally, zooming actually breaks link-hints, so when you are zoomed, the position of a true object is not reveled.

This did come to my mind before implementing zooming, just got lost at the end, :-(

Looking forward to your new implementation. I might try taking a look at it, too.

jmercouris:
It will be some time as I am now focused on doing the GTK port, so if you do want to look at it / work on it, absolutely feel free to!

just to have a quick idea of what I had in mind for version 2

1. more reliably detect position of anchor elements, also execute faster
 
version 3 will involve a few more changes that will allow selection of buttons/text-input boxes, but I want to follow the same approach of the json return to nEXT which then handles the next action

PR #49, opened 2018-01-03
Make calls into main-loop synchronous

jasom:
This should reduce the impedence mismatch between gtk and cocoa a bit.

It also allows one to get meaningful values out of expressions evaluated
in the main loop


jmercouris:
This is great, thank you so much!

jmercouris:
1 step closer to a complete gtk port :D!!!

Issue #58, opened 2018-01-11
Distinguishing capital and lower case letters in key bindings


lehmacdj:
I took a stab at implementing some vi keybindings, however encountered a problem while doing so. Unfortunately capital and lower case letters are recognized as the same keystroke. Consider the following simple `init.lisp` as an example:

```
(in-package :next)
(define-key *global-map* (kbd "j") #'scroll-down)
(define-key *global-map* (kbd "J") #'scroll-to-top)
```

Regardless of whether "J" or "j" is pressed the window always scrolls to the top instead of differentiating its behavior as necessary. I think this is because the way key strokes are reported, "J" and "j" produce the same key code just with the shift modifier either set or unset. It would be ideal if we could allow using the different letters for defining keybindings, and differentiate them however necessary.

I can try to implement the ability to differentiate these in the next day or two.

jmercouris:
Hi, thank you for the report! I did not know about this.

I think I know actually what may be the cause of this. If you take a look at `keymap.lisp` and look at `push-key-chord`, you'll see the function you need to examine. There is where you can test to see whether the system is actually pushing lowercase and uppercase letters. If it is pushing upper and lower case letters correctly, then the problem must lie within the consumption of key-chords or within the key-chord struct itself.

For this, I would take a look at `consume-key-sequence` also within `keymap.lisp`. Possibly the equality function being used for the keymap is too liberal when comparing the key-chord structs. You can see the clhs entry here: http://www.lispworks.com/documentation/HyperSpec/Body/f_mk_has.htm

make-hash accepts a test parameter, in this case I have set it to `'equalp` for most of the key mappings within next. You can see an example of this in `global.lisp`, on the line that starts with `defvar *global-map*`. Therefore perhaps making a keymap with a different equality comparator might fix the issue.

There are a couple of equality comparators that may be more suitable. A list of the types of equalities can be seen here: http://www.gigamonkeys.com/book/syntax-and-semantics.html in the section titled `Truth, Falsehood, and Equality`

Once again, thank you for your report, please let me know how the investigation goes!

lehmacdj:
The problem is that structs cannot be compared with `'equal`, but `'equalp' uses a case insensitive comparison. The keys are being reported with the correct case.

I think the correct solution is to use a custom comparison function, for which we would also need a custom hash function, or as a simpler hack, instead of using a struct we could use a list of a character followed by 3 booleans to represent the key struct and then use `'equal`, but that seems bad from a readability perspective.

[...]

PR #62, opened 2018-01-13
store char-code in struct instead of key

lehmacdj:
Because `'equalp` compares strings and characters case insensitively we
need to store the key code in the key struct used for key recognition.

This implements the changes suggested in #58.

### DIFF
--- a/next/source/cocoa/cocoa.lisp
+++ b/next/source/cocoa/cocoa.lisp
@@ -4,7 +4,6 @@
 
 (defparameter *window* nil)
 (defparameter *next-view* nil)
-(defparameter *character-conversion-table* (make-hash-table :test 'equalp))
 
 (defclass minibuffer-view (ns:ns-view)
   ((input-buffer :accessor input-buffer)
@@ -164,17 +163,17 @@
 (defun process-event (event)
   (let* ((flags (#/modifierFlags event))
 	 (character (ns-to-lisp-string (#/charactersIgnoringModifiers event)))
-         (mapped-character (gethash character *character-conversion-table* character)))
+         (key-code (char-code (char character 0))))
     (next:push-key-chord
      (> (logand flags #$NSControlKeyMask) 0)
      (> (logand flags #$NSAlternateKeyMask) 0)
      (> (logand flags #$NSCommandKeyMask) 0)
-     mapped-character)))
+     key-code)))
 
 (defun make-window ()
   (gui::assume-cocoa-thread)
   (ns:with-ns-rect (r 100.0 100.0 1024.0 768.0)
-    (ccl::with-autorelease-pool 
+    (ccl::with-autorelease-pool
       (let* ((.window. (make-instance
 		      'next-window
 		      :with-content-rect r
@@ -192,9 +191,7 @@
 	(setf *next-view* .next-view.)
 	*window*))))
 
-(defun initialize ()
-  (setf (gethash "" *character-conversion-table*) "RETURN")
-  (setf (gethash "-" *character-conversion-table*) "HYPHEN"))
+(defun initialize ())
 
 (defun start ()
   (on-main-thread
--- a/next/source/gtk/gtk.lisp
+++ b/next/source/gtk/gtk.lisp
@@ -5,7 +5,6 @@
 (defparameter *cookie-type* :webkit-cookie-persistent-storage-text)
 (defparameter *cookie-accept-policy* :webkit-cookie-policy-accept-always)
 (defparameter *next-interface* nil)
-(defparameter *character-conversion-table* (make-hash-table :test 'equalp))
 
 (defclass next-interface ()
   ((window :accessor window :initarg :window)
@@ -71,9 +70,7 @@
 	       (funcall ,thunk)))
 	   (lparallel:force ,promise))))))
 
-(defun initialize ()
-  (setf (gethash #\Return *character-conversion-table*) "RETURN")
-  (setf (gethash #\- *character-conversion-table*) "HYPHEN"))
+(defun initialize ())
 
 (defun start ()
   (synchronous-within-main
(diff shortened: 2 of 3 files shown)


lehmacdj:
Just realize that we need a little bit more work on this. Unfortunately keys like HYPHEN are reported as a string of more than one character so just storing the char-code is not sufficient for the map. It is almost like we need to be able to translate that string into the char-code for `-`.

jmercouris:
Thank you for your awesome work! Looks really nice and clean!

I would like to keep "HYPHEN" and "RETURN" for usage in config files, because "C--" is hard to read as a keybinding

I think I know how we can fix the hyphen issue, all we need to do is have a `process-token` function, or something that will just map "HYPHEN" to "-", which will then get fed back into the existing process. We can use easily integrate this into (kbd)

Then, we will have to fix the backends to send the appropriate char for `-` and `M^`.

jmercouris:
If you like I can perform the changes required to the two backends for this, please let me know what works best for you

lehmacdj:
I can perform the necessary necessary.

While I am at it I think it would make sense to move the logic for translating characters (e.g. the (*character-conversion-table*`) into a textual representation to a separate file, so we can modify it only in one place instead of separately in cocoa, gtk, and the next package  when adding new keys, e.g. tab, escape. I think it would be the most sense to put such functions in the next package, but it seems like it might also make sense to include them in a separate "core" package, or something like that, to avoid having the interface call too heavily into the next package and to at least somewhat protect it from modification in user configuration files. Which do you think would be the most reasonable?

jmercouris:
I think it is a high priority that we isolate as much of platform specific code to the interface package for each platform. It is because we get great advantage by decoupling the interface as much as possible from the core (e.g. the :next package). Eventually, the dream is that the core will run in its own entirely separate process from the GUI. 

Advantages:

1. can't run random lisp code from the browser via some sort of attack as the GUI and the Core interface via a strict API
2. stability of the GUI should not effect the stability of the core. if there is a failure in the GUI, the core should be able to restart the GUI and restore its state
3. portability, whatever GUI system or technology for rendering web views arises in the future, it should be possible to easily port next to it by simply implementing the few functions necessary in the interface

this flexibility so far has made it easy to make the GTK port. Hopefully it will pay off in the future when new engines, technologies, and platforms arise. This is a popular paradigm in Common Lisp GUI development, and it has proven to be very powerful.

that is why I think we should leave the character-conversion-table within it's current setup/contexts. Please let me know what you think is best too, and if there are any holes in my logic :D


thanks,

-John

lehmacdj:
The only thing about the conversion table is that it may end up being the same if not almost identical for all platforms. I see that you are right though, the different platforms have somewhat different conversion tables right now anyways.

lehmacdj:
@jmercouris I think this is ready to be merged now.

I ended up sending the char-code directly to the key-map from the interface, since I thought that was better than converting first to a textual representation and then back.

jmercouris:
Hey, looks good, I am very much agreeing with your choices here! I believe you're right in just using the char-code directly in the interface, adds a layer of simplicity instead of passing around strings. It also avoids having those individual maps, as long as we can rely on the consistency of the external implementations (e.g. obj-c and gtk) being consistent in the codes they produce AND using ASCII the way we expect. I think most likely this will not be a problem for most ports, and in the worst case scenario, we can write a wrapper  if necessary (e.g. on some windows port or something).

I'm not as sure that breaking apart `split-chord` yields any real benefits, but it is a little bit cleaner. So that is good at least!

I would also like to say, whenever possible, I think we should avoid using shortened or abbreviated forms of words. Maybe it is my objective-c background speaking, but I prefer `chord-sequence` to `chord-seq` as there is absolutely no potential for ambiguity.

Lastly, I would like to involve you in some discussions about some deep changes I'll be making to the architecture that will change how the system blocks, sets bindings, and makes closures around functions that interface with windows and buffers.

This should result in some nice benefits for writing interactive functions/getting minibuffer input. You can already see the beginnings here:

https://github.com/next-browser/next/blob/master/next/source/command.lisp
and here: https://github.com/next-browser/next/blob/master/next/source/macro.lisp#L41

All part of the objectives listed here of course: https://github.com/next-browser/next/tree/master/next#008

what's the best way to reach you therefore? e.g. email or something, would you like that we setup a mailing list? I had one but discontinued it due to lack of activity (at least there are people on IRC :P)

Finally, thank you for your contributions and interest thus far. You've been the biggest contributor I think by a long margin, and have shown an interest in the project which has inspired me and made very happy. So, thanks!

-John

lehmacdj:
> I would also like to say, whenever possible, I think we should avoid using shortened or abbreviated forms of words. Maybe it is my objective-c background speaking, but I prefer chord-sequence to chord-seq as there is absolutely no potential for ambiguity.

In my opinion, such shortened names are okay (I might have even named the variable seq) when working with a tiny scope, as in this case, but perhaps my Haskell background betrays me here?

> what's the best way to reach you therefore? e.g. email or something, would you like that we setup a mailing list? I had one but discontinued it due to lack of activity (at least there are people on IRC :P)

Email would be fine. I would also be happy to use Slack or a similar service, but that has the problem of messages disappearing eventually on free accounts and also has a somewhat higher barrier of entry to email for future users.

jmercouris:
Once again, thank you for your contribution! Looks really good apart from some minor details:

I added #\Escape in the conversion map instead of hard-coding ASCII code 27

I had to revert split-chord as it was not generating prefix keys properly and C-x b was not functional.

GTK would simply not compile :P, I also fixed that

Now we can distinguish C-j, C-J, etc. It also now properly detects Escape :)!

lehmacdj:
Thanks for fixing my little mistakes!

I think when reversing split-chord you also reverted change I made to fix a subtle bug I found while working on this. I noticed that it is impossible to bind C or S or M because these characters are used to determine if the control, meta or super modifiers are pressed. I had a weird dance of reversing the list of characters split on `-` and then processing the first one separately and always interpreting it as a literal character. I must have accidentally introduced a bug while doing so.

jmercouris:
Hey, no problem! I know how to fix this bug, I’ll try to make a patch ASAP.

In the meantime I’ve been working on the blog/some marketing efforts, as well as conversion of all interactive functions to use the new defcommand macro + continuation passing style (macro with-result) to avoid writing blocking code or strange def keybind forms…

Anyways, all that aside, I’ll keep you updated when I push a fix for this!

jmercouris:
@lehmacdj This issue has been fixed in a recent by PR :)

-John

Issue #29, opened 2017-11-29
honor the XDG directory specification


jmtd:
https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

so instead of hardcoding `~/.next.d` you'd check for `$XDG_CONFIG_HOME` and if set use `$XDG_CONFIG_HOME/next` (or `next.d`), otherwise default to `~/.config/next` (or `next.d`)

This helps to prevent clutter in ~ for people and also gives a measure of control as to where config, data are kept.

Thanks

jmercouris:
Hi! Thank you for your feature suggestion! That is a great idea! I will look into it! I too am bothered when an application just decides to dump something in a directory of its choosing, I wasn't aware of XDG_CONFIG_HOME, but that's a great feature!

jmercouris:
@lehmacdj Has implemented this! @jmtd sorry about the long wait, thank you for your patience!

PR #63, opened 2018-01-13
Use the xdg home directory specification for configuration and data files

lehmacdj:
As requested in #29. In order to maintain backwards compatibility with older versions of the browser I check for the existence of the old files before changing the location they exist in the file system. This means that users will have to remove ~/.next.d for the changes to take effect.

### DIFF
--- a/next/source/base.lisp
+++ b/next/source/base.lisp
@@ -3,7 +3,7 @@
 (in-package :next)
 
 (defun start ()
-  (ensure-directories-exist (uiop:physicalize-pathname "~/.next.d/"))
+  (ensure-directories-exist (xdg-data-home))
   (initialize-default-key-bindings)
   ;; load the user configuration if it exists
   (load *init-file-path* :if-does-not-exist nil)
--- a/next/source/global.lisp
+++ b/next/source/global.lisp
@@ -45,25 +45,16 @@
   "The package global variables available, populated by helper
   function load package-globals")
 (defvar *init-file-path*
-  (if (uiop:file-exists-p "~/.next.d/init.lisp")
-    "~/.next.d/init.lisp"
-    (xdg-config-home "init.lisp"))
+  (xdg-config-home "init.lisp")
   "The path where the system will look to load an init file from.")
 (defvar *history-db-path*
-  (if (uiop:file-exists-p "~/.next.d/history.db")
-    "~/.next.d/history.db"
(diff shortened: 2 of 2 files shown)


jmercouris:
Hi, thank you for implementing this feature! I like it! :)

I have a suggestion about the default behavior, because it is tricky, we can take it a number of ways,

what if we have a function called `get-home` which will use the xdg home if it is set, otherwise, will just use the home dir, then we can `merge-pathnames` with `xdg` if exists, and `~` if not. This way, we don't have to do a check for every place a path is used. There is a loss of granularity in this method, but I think we will benefit from the consistency when we have a ton of paths

you can see an example of this in a predecessor browser:  https://github.com/AeroNotix/lispkit/blob/master/settings.lisp#L3

here is the function that they use for `get-xdg-config-dir`:
https://github.com/AeroNotix/lispkit/blob/master/user.lisp#L4

please let me know what you think! thank you so much for your contributions! I was quite excited to see two new prs today!!! :D

lehmacdj:
I definitely agree with using `merge-pathnames` that is much better than `concatenate` for file paths.

Unfortunately I don't think defaulting to `~` if the XDG environment variables are unset correctly follows the XDG directory specification. If the variables are unset, the directories should be put at a default location (e.g. ~/.local/share for cookies/history/bookmarks or ~/.config for init.lisp). The problem I am trying to solve by making separate checks is simply maintaining backward compatibility with files that are already at the old location in `~/.next.d`. For new files added to next, we would simply use `xdg-config-home` or `xdg-data-home` to get a path to a file and that would be it, there would be no need to check that the file is at the old location because it wouldn't be created yet.

jmercouris:
I just looked up the specification, you are 100% correct. My apologies for assuming details about the spec without fully understanding it. I wouldn't worry too much about breaking compatibility as we are so early in this, and not even at 1.0.

If a user upgrades, it is sort of an at-risk thing, of course release notes will detail breaking changes, but until we hit 1.0 we should maintain some flexibility.

I can't say I like it though, I like how emacs has its own self contained directory with all files relating to it. It makes a nice little area where you can just delete files. This does seem to be the way of the future though, so I will have to let go of my biases :D

lehmacdj:
As far as having a single directory is concerned, in practice it will still work out that way for configuration because everything will be located under ~/.config/next.

PR #65, opened 2018-01-17
Simplify switch-mode method 

selesdepselesnul:
Use `when` instead of `if`


jmercouris:
Ah yes, good call! Thank you! I still think in other languages some times :D

PR #72, opened 2018-02-02
Continuation passing style

jmercouris:
convert things to continuation passing style

makes bindings for functions easier

additionally converts all interactive functions to use define-command

much cleaner and more obvious code (rather than the indirection with the `:input` and `:input-complete` macros)


jmercouris:
looks great, did you test it?

jmercouris:
well, of course I did!

PR #83, opened 2018-02-22
Fix #'kbd

shrdlu68:
Fix #'kbd to distinguish modifiers from upcase chars.


jmercouris:
looks great, thank you!

PR #101, opened 2018-05-22
parse command line arguments and open urls

### LINKED ISSUE #99: Commandline argument support?
I might be missing something, but it seems that `next` does not support commandline arguments at the moment.  That is,

```
next <some-url>
```
fires up Next and opens the start page, not `<some-url>`.

I can provide a patch if required.

vindarel:
I suggest this. We can put many urls as arguments and interleave them with `-h` or `-v`.

fixes #99

### DIFF
--- a/next/next.asd
+++ b/next/next.asd
@@ -4,7 +4,7 @@
 (asdf:defsystem :next
   :serial t
   :depends-on (:alexandria :cl-strings :cl-string-match :puri
-               :queues.simple-queue :sqlite :parenscript :cl-json :swank)
+               :queues.simple-queue :sqlite :parenscript :cl-json :swank :unix-opts)
   :pathname "source/"
   :components ((:file "package")
 	       (:file "macro")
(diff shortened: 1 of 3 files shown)


jmercouris:
Hey! Thank you for the pull request! I'll get to testing this ASAP!!!

jmercouris:
There's some strange behavior happening here:

screen shot 2018-05-26 at 11 08 16

Not sure why it is, perhaps somehow unix-opts isn't being saved into the image?

vindarel:
unix-opts is a wrapper library, it should work for CCL too. Strange indeed. 

```

jmercouris:
I can't get it to work on CCL, but it doesn't much matter as CCL is intended for OSX where it is meant to really be launched as an APP bundle.

I would just like to confirm that you've tested it with SBCL and it works, if so, I'll go ahead and merge this, thanks!

vindarel:
I can confirm on SBCL :)

selection_237

jmercouris:
Thank you :)!

PR #105, opened 2018-06-19
makefile: --non-interactive

vindarel:
Hello,

That's a little but handy thing that I don't know from long… let's not be trapped into the debugger, it isn't very friendly on the command line. Never had to C-z and `kill sbcl` ??


jmercouris:
Looks good, thank you!!!

PR #117, opened 2018-08-21
Update README.org

chuchana:
Added Arch package name for WebKitGTK+
Clarified path for next-gtk binary

### DIFF
--- a/next/README.org
+++ b/next/README.org
@@ -90,6 +90,7 @@ Find a package for your distribution which provides
 
 For Ubuntu this package is ~libwebkit2gtk-4.0-dev~
 For Debian this package is ~libwebkit2gtk~
+For Arch this package is [~webkit2gtk~](https://www.archlinux.org/packages/extra/x86_64/webkit2gtk/)
 
 #+begin_quote
 A note of caution about installing Webkit on Linux via your package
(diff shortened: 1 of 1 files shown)


next/README.org
```diff
@@ -126,7 +127,7 @@ In a new Terminal execute the following:
 : cd next/next
 : make build-gtk
 
-the compiled binary will be present in source/next-gtk.
+the compiled binary will be present in the directory ~source~ as ~next-gtk~.
```
Ambrevar: Was `source/next-gtk` confusing?  Don't know, I might be biased.

next/README.org
```diff
@@ -90,6 +90,7 @@ Find a package for your distribution which provides
 
 For Ubuntu this package is ~libwebkit2gtk-4.0-dev~
 For Debian this package is ~libwebkit2gtk~
+For Arch this package is [~webkit2gtk~](https://www.archlinux.org/packages/extra/x86_64/webkit2gtk/)
```
Ambrevar: The syntax for links in Org is `[[URL][TITLE]]`.

next/README.org
chuchana: I misread it as a path to a directory because of the word `in`, so I had a bit of trouble finding it at first. It might just be me, but I thought getting rid of that potential ambiguity couldn't hurt.

next/README.org
chuchana: Thanks

PR #118, opened 2018-08-23
add: function to bookmark current-url

watkinsr:
Would be nice to bookmark the current URL I think.

### DIFF
--- a/next/source/bookmark.lisp
+++ b/next/source/bookmark.lisp
@@ -45,6 +45,11 @@
   (with-result (url (read-from-minibuffer (mode *minibuffer*)))
     (%bookmark-url url)))
 
+(define-command bookmark-url-from-buffer ()
+  "Allow the user to bookmark a URL from the current active buffer."
+  (let ((url (interface:web-view-get-url (view *active-buffer*))))
+    (%bookmark-url url)))
+
 (define-command bookmark-delete ()
   "Delete a bookmark from the bookmark database."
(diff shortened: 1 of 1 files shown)


jmercouris:
Hi, thank you very much for your interest!

Your function is nice and clean, but I believe that it has the same functionality as:

‘’’bookmark-current-page’’’

so perhaps instead clean up bookmark-current-page and implement it in the same way, or make an alias or something, I’m not sure.

watkinsr:
@jmercouris as far as I could tell and manually test. "BOOKMARK-CURRENT-PAGE" will only bookmark the buffer name. It does not get the latest URL so if you happened to navigate away from the initially entered URL, it would bookmark the first entry. Hence, the need for this function.

jmercouris:
Ah, I understand now, my apologies! Thank you for the clarification!!!

jmercouris:
Congratulations! to the first of hopefully many more :)


if you need tasks, I have a huge list :D

watkinsr:
@jmercouris Could you send the list through to ryanwatkins54@gmail.com ? 👍

Ambrevar:
Better: could you publish it in the repo?
An Org file would be great! :)

jmercouris:
I usually have a task list for each version in the developer readme at the bottom, if you open it in org mode you can see it,

the only problem with that is that it tracks predictable and major releases, I tend not to put really ambitious and speculative projects, such as the current branch that I am working on, I can probably link to that branch though,

-John

PR #126, opened 2018-10-11
Fix set-url-new-buffer duplicate

Ambrevar:
SET-URL-NEW-BUFFER is declared twice which is probably a bug and also leads to
compilation warning/error on some Common Lisp implementations.

### DIFF
--- a/next/source/base.lisp
+++ b/next/source/base.lisp
@@ -53,7 +53,7 @@
     ;; We can have many urls as positional arguments.
     (if free-args
         (loop for url in free-args do
-             (set-url-new-buffer url t))
+             (%set-url-new-buffer url t))
         (set-url *start-page-url*))))
 
 (defun initialize-default-key-bindings ()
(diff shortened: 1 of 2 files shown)


jmercouris:
This error is because there is a command called `set-url-new-buffer`. You can see the `define-command` macro within `command.lisp`.

Instead of `(define-command set-url-new-buffer () ...)` producing a function `(defun set-url-new-buffer () ...)` it could produce `(defun command-set-url-new-buffer () ...)` to avoid these collisions, or we could use the `%` as you have done.

I'm not sure what is better in the long run, probably the `%` to indicate that it is internal, not meant to be public facing. Thoughts?

jmercouris:
That does make sense, there is no reason why we could not employ both strategies!

Ambrevar:
Actually some functions are already "command-" prefixed, e.g. `command-documentation-style-warning`.  I see two options here:

- Either we rename those functions to something else.
- Or we prefix command with something else.  In this case, what about using a different special character, such as `@`.  The benefit is that it makes for shorter command names.

Ambrevar:
Additional strategy: in the `define-command` macro, why not erroring out if the function is already bound with `fboundp` and if the instance is a command?

jmercouris:
Yes, it should probably emit a signal of some sort when redefining a function, that will help catch a lot of problems,

-John

PR #128, opened 2018-10-15
fix gtk-entry-text not external to gtk package

vindarel:
I am on Quicklisp "2018-04-30", if that makes a difference.

a `make build-gtk` was failing due to this.

### DIFF
--- a/next/source/gtk/gtk.lisp
+++ b/next/source/gtk/gtk.lisp
@@ -33,7 +33,7 @@
         (write-to-string element))))
 
 (defmethod get-input ((self minibuffer-view))
-  (gtk:gtk-entry-text (input-entry self)))
+  (gtk::gtk-entry-text (input-entry self)))
 
 (defmethod process-set-completions ((self minibuffer-view))
   "Process and set completions for the minibuffer"


jmercouris:
Thanks for fixing this, make cl-cffi-gtk was updated and that symbol was no longer being exported?

Not sure, at any rate, this is greatly appreciated, thank you!

vindarel:
I had started a CI on Gitlab: https://gitlab.com/vindarel/next-browser/blob/linux-binary/.gitlab-ci.yml (linux-binary branch) https://gitlab.com/vindarel/next-browser/pipelines

Time to re-think about some CI maybe !

jmercouris:
Hmm, it does seem like you are getting farther at least!

Just out of curiosity, why do you think it has so many issues in compiled form? When I was running it in a REPL, most everything was working. Granted, I was only testing SBCL with GTK on an Ubuntu VM, so my environment may be quite different.

vindarel:
Good question. I suppose it has to do with library versions. We checked cl-webkit2 (it has to be your fork). What version of cl-ffi-gtk are you using ? (I see no info in the [dev doc](https://github.com/atlas-engineer/next/tree/master/next#developer-installation)). I might try with the one not in Quicklisp.

jmercouris:
Unfortunately I've updated quicklisp, so I can't tell you which version
of cl-cffi-gtk I was using during development :\

Ambrevar:
Link to the two cl-cffi-gtk:
- Old: https://github.com/crategus/cl-cffi-gtk
- New: https://github.com/Ferada/cl-cffi-gtk

vindarel:
seems better, I can browse a bit, but now

```
"cl-cffi-gtk main thread" running
                                    {10090BB463}>:
  nil is not a Lisp string or pointer.

Backtrace for: #<SB-THREAD:THREAD "cl-cffi-gtk main thread" RUNNING {10090BB463}>
0: ((LAMBDA NIL :IN SB-DEBUG::FUNCALL-WITH-DEBUG-IO-SYNTAX))
1: (SB-IMPL::CALL-WITH-SANE-IO-SYNTAX #<CLOSURE (LAMBDA NIL :IN SB-DEBUG::FUNCALL-WITH-DEBUG-IO-SYNTAX) {100A46378B}>)
2: (SB-IMPL::%WITH-STANDARD-IO-SYNTAX #<CLOSURE (LAMBDA NIL :IN SB-DEBUG::FUNCALL-WITH-DEBUG-IO-SYNTAX) {100A46375B}>)
3: (PRINT-BACKTRACE :STREAM #<SYNONYM-STREAM :SYMBOL SB-SYS:*STDERR* {1000314B73}> :START 0 :FROM :INTERRUPTED-FRAME :COUNT NIL :PRINT-THREAD T :PRINT-FRAME-SOURCE NIL :METHOD-FRAME-STYLE NIL)
4: (SB-DEBUG::DEBUGGER-DISABLED-HOOK #<SIMPLE-ERROR "~A is not a Lisp string or pointer." {100A461153}> #<unavailable argument>)
5: (SB-DEBUG::RUN-HOOK *INVOKE-DEBUGGER-HOOK* #<SIMPLE-ERROR "~A is not a Lisp string or pointer." {100A461153}>)
6: (INVOKE-DEBUGGER #<SIMPLE-ERROR "~A is not a Lisp string or pointer." {100A461153}>)
7: (ERROR "~A is not a Lisp string or pointer." NIL)
8: ((:METHOD CFFI:TRANSLATE-TO-FOREIGN (T CFFI::FOREIGN-STRING-TYPE)) NIL #<unavailable argument>) [fast-method]
9: (GOBJECT:G-VALUE-SET-STRING #.(SB-SYS:INT-SAP #X7FFFE5AF7FE0) NIL)
10: (GOBJECT:G-OBJECT-SET-PROPERTY #.(SB-SYS:INT-SAP #X7FFFE0186250) "text" NIL #<GTYPE :name "gchararray" :id 64>)
11: ((:METHOD (SETF SB-MOP:SLOT-VALUE-USING-CLASS) (T GOBJECT:GOBJECT-CLASS T GOBJECT::GOBJECT-PROPERTY-EFFECTIVE-SLOT-DEFINITION)) NIL #<unavailable argument> #<GTK:GTK-ENTRY {1009165043}> #<GOBJECT::GOBJECT-PROPERTY-EFFECTIVE-SLOT-DEFINITION GTK::TEXT>) [fast-method]
12: ((:METHOD NEXT::READ-FROM-MINIBUFFER (T NEXT::MINIBUFFER-MODE)) #<FUNCTION (LAMBDA (NEXT::URL) :IN NEXT::SET-URL-CURRENT-BUFFER) {100496022B}> #<NEXT::MINIBUFFER-MODE {100983DE53}> :COMPLETION NEXT::HISTORY-TYPED-COMPLETE :SETUP NEXT::SETUP-URL :CLEANUP NIL :EMPTY-COMPLETE T) [fast-method]
13: (NEXT::CONSUME-KEY-SEQUENCE)
```

may be a Next error.

jmercouris:
Hi @vindarel thank you for your continued investigation!

Probably the error is in read-from-minibuffer trying to get the value of the mini buffer upon each keypress. 

Can please you describe exactly which sequence of actions leads to this crash?

vindarel:
It happens with a `C-l` (L). `C-g` works fine, point and click too, `C-x b` shows the prompt.

vindarel:
The newer cl-ffi-gtk has been included into Quicklisp (october).

vindarel:
The newer cl-ffi-gtk has been included into Quicklisp (october).

vindarel:
The newer cl-ffi-gtk has been included into Quicklisp (october).

PR #130, opened 2018-10-15
fix beware nil url

vindarel:
The landing page is a 404, without this fix I couldn't browse a bit, it lead to a `cl-dbi` error. Now a bit better.

This may not be the root cause.


jmercouris:
that makes a lot of sense, a URL should exist if we are going to append it to the history db, thank you!

Issue #133, opened 2018-10-19
Title and executable name


Ambrevar:
The readme's first section is "Next Web Browser".  It's the only part where "web" is used.

Should we use "Next", "Next Browser" or "Next Web Browser"?
Should it be capital "B" in "Browser?

What about the executable name?  I think `next-browser` would be less ambiguous and reduce the change of running into a name conflict in the PATH.

jmercouris:
I think "Next Browser", and I agree that the executable should be `next-browser`

-John

Ambrevar:
Makes sense, thanks for clarifying!

PR #134, opened 2018-10-19
assets/next.desktop: Init

Ambrevar:
Before merging this, we need to settle on a few points:

- Name: See issue #133 .
- Icon names: Should we rename them to `next_*`?
- Exec: See issue #133 as well.
- Actions: What do we support now?

### DIFF
--- a/assets/next.desktop
+++ b/assets/next.desktop
@@ -1,0 +1,15 @@
+[Desktop Entry]
+Version=0.8
+Name=Next Browser
+Comment=Be Productive
+GenericName=Web Browser
+Keywords=Internet;WWW;Browser;Web;Explorer
+Exec=next-browser %u
+Terminal=false
+X-MultipleArgs=false
+Type=Application
+Icon=next
+Categories=GTK;Network;WebBrowser;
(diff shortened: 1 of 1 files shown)


jmercouris:
Hi Pierre, I've replied to the questions in the other issues, interested in your thoughts as well,

-John

Ambrevar:
> What do we support now? Currently we only support MacOS, our subsequent release we will > support Linux & MacOS

Sorry for the confusion, that's not what I meant.  From the .desktop file:

```
Actions=NewWindow;NewPrivateWindow;
```

jmercouris:
With regards to the desktop file, we have not yet implemented a private window mode, so I would just leave it as "NewWindow"

Ambrevar:
What about the icons?  On GNU/Linux, they should have a unique name or else it will confuse the XDG specifications.

jmercouris:
I guess then, make them whatever the XDG spec says, "next-browser" in this case?

Ambrevar:
I thought you did not want to use "next-browser" but "next" instead?

Ambrevar:
OK, I've corrected the file and merged.

Issue #109, opened 2018-08-06
`M-l` leads to crash


emacsomancer:
I can use `C-l` fine to open a new url in the current tab. But `M-l` leads to Next completely freezing and needing to be xkill'ed (on my Arch machine) or crashing (on my GuixSD machine) with the following error displayed in terminal:

```
(next-gtk:6410): Gtk-CRITICAL **: 09:27:05.739: gtk_tree_path_to_string: assertion 'path != NULL' failed
Unhandled SB-INT:SIMPLE-PROGRAM-ERROR in thread #<SB-THREAD:THREAD "cl-cffi-gtk main thread" RUNNING
                                                   {1004655193}>:
  invalid number of arguments: 0

Backtrace for: #<SB-THREAD:THREAD "cl-cffi-gtk main thread" RUNNING {1004655193}>
0: (NEXT::SET-URL-NEW-BUFFER) [external]
1: (NEXT::CONSUME-KEY-SEQUENCE)
2: (GOBJECT::CALL-WITH-RESTARTS #<FUNCTION (LAMBDA (INTERFACE::WINDOW INTERFACE::EVENT) :IN INTERFACE:START) {22286A3B}> (#<GTK:GTK-WINDOW {10046B5673}> #S(GDK:GDK-EVENT-KEY :TYPE :KEY-PRESS :WINDOW #<GDK:GDK-WINDOW {1005713D03}> :SEND-EVENT NIL :TIME 36894816 :STATE (:MOD1-MASK) :KEYVAL 108 :LENGTH 1 :STRING "l" :HARDWARE-KEYCODE 46 :GROUP 0 :IS-MODIFIER 0)))
3: ((LAMBDA (GOBJECT::CLOSURE GOBJECT::RETURN-VALUE GOBJECT::COUNT-OF-ARGS GOBJECT::ARGS GOBJECT::INVOCATION-HINT GOBJECT::MARSHAL-DATA) :IN "/home/slade/common-lisp/quicklisp/dists/quicklisp/software/cl-cffi-gtk-20160208-git/gobject/gobject.signals.lisp") #.(SB-SYS:INT-SAP #X7FFFDC145AE0) #.(SB-SYS:INT-SAP #X7FFFE16A6100) 2 #.(SB-SYS:INT-SAP #X7FFFE16A61B0) #<unused argument> #<unused argument>)
[...]

PR #137, opened 2018-10-25
Fix nil path.

Ferada:
- Fixes one case of a null value being passed to a tree path function, that's only visible on stdout.
- Cleaner main loop handling; this might a bit of explanation: `within-main-loop` will always add one more level of nesting for the Lisp side handling of the main loop (via `ensure-gtk-main`) and thus needs to match the number of invocations for `leave-gtk-main` (eventually, the last one would terminate the main loop and the program). Therefore, this change makes the macro use `within-gtk-thread`, which we've added exactly for this reason, to just ensure a function is executed in the main loop thread, but without further nesting. For the built image there's no change, for interactive development you might want to ensure you always have one level of nesting left, so before the main window is closed again, use `ensure-gtk-main` and observe that second return value should be 2 (but not much more than that). When the window is then closed, the GTK main thread is still kept alive, with nesting level 1 and `(start)` could be called again (increasing the level back to two). I've also seen a crash (in the WebKit part probably) when the main thread was completely killed and then restarted via `start`, that's why I'm suggesting leaving the main thread alive.
- Stored the main window in a variable, basically just to be able to remove it later, that makes for a bit nicer experience with interactive development as the window doesn't stay around when it was supposed to be closed.
- One of the `destroy` handlers was duplicated, so would've called `leave-gtk-main` more often than necessary.

### DIFF
--- a/source/gtk/gtk.lisp
+++ b/source/gtk/gtk.lisp
@@ -65,14 +65,17 @@
 	 (funcall ,thunk)
 	 (let ((*within-main-nesting* t)
 	       (,promise (lparallel:promise)))
-	   (gtk:within-main-loop
+	   (gtk:within-gtk-thread
 	     (lparallel:fulfill ,promise
 	       (funcall ,thunk)))
 	   (lparallel:force ,promise))))))
 
 (defun initialize ())
 
+(defvar *window* NIL)
+
 (defun start ()
+  (gtk:ensure-gtk-main)
   (synchronous-within-main
    (let* ((window
            (make-instance 'gtk:gtk-window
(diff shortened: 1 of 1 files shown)


Ambrevar:
Excellent!  Thank you so much for working on this!
I'll test as soon as possible.

@jmercouris: Insights?

Ambrevar:
It fixes both #109 and the unclean shutdown of Next!

jmercouris:
The changes look very good and reasonable to me! I’m no expert on GTK so I can’t offer anything more than that, if you’re okay to merge Pierre, then I would say let’s do it!

Ambrevar:
@Ferada : Thanks you so much!

PR #157, opened 2018-12-05
warn about error loading the init file and don't fail

vindarel:
Next failed to start because of an error in my `init.lisp`, which was working with the previous version (the command used `(interface:web-view-get-url …`).

This addition catches all errors so than Next can start. Or should it fail, or warn the user in the interface ?

disclaimer: I couldn't test it !! (I still don't have a working installation)

### DIFF
--- a/source/base.lisp
+++ b/source/base.lisp
@@ -81,7 +81,10 @@ Set to '-' to read standard input instead."))
         (loop for object = (read *standard-input* nil :eof)
               until (eq object :eof)
               do (eval object)))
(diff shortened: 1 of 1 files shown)


vindarel:
It prints the condition (which comes with some stacktrace if I recall correctly).

jmercouris:
@vindarel you are correct it prints the condition, line number, and a little stacktrace, thanks for this addition!

now within the compiled versions we'd probably (instead of just printing to stdout) want to print to a place that the user can see it, maybe in a messages buffer or something, but we first need to implement messages :D

PR #175, opened 2018-12-23
Document dev installation via Guix

swedebugia:


Ambrevar:
Thanks for the suggestion!

PR #176, opened 2018-12-23
Add clarification that cloning is needed

swedebugia:
...for guix.scm to work.


Ambrevar:
Sure, but I wonder how there could possibly be a confusion.
Could you give an example?

PR #196, opened 2019-04-22
Add visits column to typed history, and sort by it

FrozenPigs:
This will add a visits column to new databases, as well as ones that already exists. It's incremented on each visit to the url, and it is used to sort the history so the things you visit often come first.

### DIFF
--- a/next.asd
+++ b/next.asd
@@ -7,7 +7,8 @@
   :license "BSD 3-Clause"
   :serial t
   :defsystem-depends-on ("trivial-features")
-  :depends-on (:alexandria
+  :depends-on (:anaphora
+               :alexandria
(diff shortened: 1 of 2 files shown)


jmercouris:
This is a very useful patch. It makes a lot of sense, so you get a fuzzy match that also considers frequency! Thank you for your submission! We'll try to get this tested and merged ASAP (we are making a lot of changes on branches and there might be some conflicts).

FrozenPigs:
This branch has been rebased te master, so it should be good for merging again.

Ambrevar:
Thanks, I'll look into it as soon as possible.

Ambrevar:
Merged, thanks!

PR #197, opened 2019-04-22
Add space to cursor-forwards-word and cursor-backwards-word

FrozenPigs:
This seems like something that should have already been there, but if not feel free to say why space was not included in the `cursor-backwards-word` and `cursor-forwards-word` functions. Also a little question, do you happen to have an IRC you guys hang out on?

### DIFF
--- a/next.asd
+++ b/next.asd
@@ -7,7 +7,8 @@
   :license "BSD 3-Clause"
   :serial t
   :defsystem-depends-on ("trivial-features")
-  :depends-on (:alexandria
+  :depends-on (:anaphora
+               :alexandria
                :cl-strings
                :cl-string-match
                :puri
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -18,26 +18,31 @@
      db "insert into typed (url) values (?)" "about:blank")
     (sqlite:disconnect db)))
 
+(defun ensure-history-db ()
+  "Returns the pathname of the history database"
+  (ensure-file-exists 
+   (anaphora:aif (window-active *interface*)
+                 (history-db-path anaphora:it)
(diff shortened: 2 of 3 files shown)


jmercouris:
No reason really, at the time I was thinking about URLs, and URLs don't have spaces in them, however you can interact with more than just URLs in the minibuffer, so that does make sense

on IRC Freenode you can find us on #next-browser

jmercouris:
Ok, so I ran this patch a little bit, and I realized that `GENERATE-INPUT-HTML` has to be modified to accommodate this change as consecutive spaces in Webkit are breaking. I tried to simply do a string substitution on the `input-buffer` for spaces with `&nbsp`, however that was simply emitted over and over again into the mini buffer when typing. So, the change is good, but some other changes are necessary to properly accommodate it.

FrozenPigs:
What exactly is the issue with consecutive spaces? If there is any extra chanages you need from this pull request feel free to ask, though my common lisp knowledge is fairly limited. I am also around in the IRC if you need to talk.

Ambrevar:
@jmercouris I'm not sure I understand the problem either.
@FrozenPigs : there has been some recent changes on those very functions, can you rebase your pull request on master?  Thanks!

FrozenPigs:
@Ambrevar Was that what I did to that branch right? I'm not the best at all this git stuff just yet heh. And that problem he was talking about is when you enter space in the minibuffer multiple times, only one appears. I belive he said that it was already an issue before this commit.

Ambrevar:
Thanks, I'll look into this as soon as possible.

Ambrevar:
Merged, thanks for your contribution!
I'll work on @jmercouris point with the inability of inserting multiple spaces.

PR #198, opened 2019-04-23
Add *search-engines* alist for defining search engines

FrozenPigs:
This adds as association list that the user can fill with short names plus a search url to use for searching in the minibuffer. This also brings back the searching with s, but allows the user to set it in the alist by changing the 'default' key.

I wasn't sure if this is the kind of feature you wanted after I saw that searching with s was removed, but I've had this feature in other web browsers in the past and find it quite useful.

### DIFF
--- a/source/global.lisp
+++ b/source/global.lisp
@@ -35,10 +35,11 @@
 (defvar *package-globals* nil
   "The package global variables available, populated by helper
   function load package-globals")
-(defvar *search-engines* '(("default" . "https://duckduckgo.com/?q=")
-                           ("wiki" . "https://en.wikipedia.org/w/index.php?search="))
+(defvar *search-engines* '(("default" . "https://duckduckgo.com/?q=~a")
+                           ("wiki" . "https://en.wikipedia.org/w/index.php?search=~a"))
   "An association list of all the search engines you can use in the minibuffer.
-The 'default' engine is use for searching with 's' and as a backup.")
+The 'default' engine is use when the query is not a valid URL, or the first
+keyword is not recognised.")
 
 (defparameter +version+
   (let ((version (asdf/component:component-version (asdf:find-system :next)))
--- a/source/global.lisp
+++ b/source/global.lisp
@@ -35,6 +35,10 @@
 (defvar *package-globals* nil
   "The package global variables available, populated by helper
   function load package-globals")
+(defvar *search-engines* '(("default" . "https://duckduckgo.com/?q=")
+                           ("wiki" . "https://en.wikipedia.org/w/index.php?search="))
+  "An association list of all the search engines you can use in the minibuffer.
+The 'default' engine is use for searching with 's' and as a backup.")
 
 (defparameter +version+
   (let ((version (asdf/component:component-version (asdf:find-system :next)))
--- a/source/global.lisp
+++ b/source/global.lisp
@@ -35,11 +35,6 @@
 (defvar *package-globals* nil
   "The package global variables available, populated by helper
   function load package-globals")
-(defvar *search-engines* '(("default" . "https://duckduckgo.com/?q=~a")
-                           ("wiki" . "https://en.wikipedia.org/w/index.php?search=~a"))
-  "An association list of all the search engines you can use in the minibuffer.
-The 'default' engine is use when the query is not a valid URL, or the first
-keyword is not recognised.")
 
 (defparameter +version+
   (let ((version (asdf/component:component-version (asdf:find-system :next)))
--- a/source/utility.lisp
+++ b/source/utility.lisp
@@ -27,10 +27,11 @@ slime. Default port is 4006."
   (swank:create-server :port *swank-port* :dont-close t))
 
 (defun parse-url (input-url)
-  (let ((engine (assoc (first (cl-strings:split input-url))
-                       *search-engines* :test #'string=))
-        (default (assoc "default"
-                        *search-engines* :test #'string=)))
+  (let* ((window (window-active *interface*))
+         (engine (assoc (first (cl-strings:split input-url))
+                       (search-engines window) :test #'string=))
+         (default (assoc "default"
+                         (search-engines window) :test #'string=)))
     (if engine
         (generate-search-query (subseq input-url
                                        (length (first (cl-strings:split input-url))))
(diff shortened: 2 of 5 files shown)


source/global.lisp
```diff
@@ -89,6 +89,10 @@
      (.selected :background-color "gray"
                 :color "white")))
   "The CSS applied to a minibuffer when it is set-up.")
+(defvar *search-engines* '(("default" . "https://duckduckgo.com/?q=")
+                           ("wiki" . "https://en.wikipedia.org/w/index.php?search="))
+  "An association list of all the search engines you can use in the minibuffer.
+The 'default' engine is use for searching with 's' and as a backup.")
```
Ambrevar: Please remove the 's' support and mentions.  Explain that 'default' is used when query is not a valid URL and first keyword is not recognized.

source/global.lisp
```diff
@@ -98,6 +102,7 @@
 (deferredvar *cookie-path-dir* (xdg-data-home)
   "The path for cookies in the GTK Version of Next")
 
+
```
Ambrevar: No need to add a newline here.

source/utility.lisp
```diff
@@ -30,23 +30,33 @@ slime. Default port is 4006."
...
-      ;; - https://www.google.com/search?q=hello world
-      (let ((url (puri:parse-uri input-url)))
-        (cond
-          ((puri:uri-scheme url) input-url)
-          ((probe-file input-url)
-           (concatenate 'string "file://" input-url))
-          (t (generate-search-query input-url))))
-    (puri:uri-parse-error ()
-      input-url)))
+  (let ((engine (assoc (nth 0 (cl-strings:split input-url))
```
Ambrevar: You can replace all the `nth 0` with `first`.

source/utility.lisp
```diff
@@ -30,23 +30,33 @@ slime. Default port is 4006."
...
+                       *search-engines* :test #'string=))
+        (default (assoc "default"
+                        *search-engines* :test #'string=)))
+    (if engine
+        (generate-search-query (subseq input-url
+                                       (length (nth 0 (cl-strings:split input-url))))
+                               (cdr engine))
+      (if (string= "s" (nth 0 (cl-strings:split input-url)))
+          (generate-search-query (subseq input-url 2) (cdr default))
```
Ambrevar: You can remove this "s" case.

source/utility.lisp
```diff
@@ -30,23 +30,33 @@ slime. Default port is 4006."
...
+               (t (generate-search-query input-url (cdr default)))))
+          (puri:uri-parse-error ()
+                                input-url))))))
 
-(defun generate-search-query (search-string)
+(defun generate-search-query (search-string search-url)
   (let* ((encoded-search-string
            (cl-string-match:replace-re "  *" "+" search-string :all t))
-         (url (concatenate 'string "https://duckduckgo.com/?q=" encoded-search-string)))
+         (url (concatenate 'string search-url encoded-search-string)))
```
Ambrevar: Better: replace concatenate with `(format ` and have the search engines use `"...~a..."` so that the encoded-search-string can be inserted anywhere in the URL.

Ambrevar:
Thanks for your contribution!

I originally intended to replace the whole search-engine system with generalized bookmarks, but that is going to take a little while.  So in the meantime, I'd be happy to merge your patch! :)

FrozenPigs:
The latest coommit should have fixed all the changes requested by @Ambrevar, but there is one extra change I would request, which is chinging the `input-url` on line 52 to a `generate-search-query`. The reason for that is as it is now, if I open next, then do 'C-l hi there RET' I end up on the url 'https://next.atlas.engineer/hi%20there'.

I assume the reason for the above is what is explained in the comment, so I think that this seems like a sensible change. Looking forward to more feedback.

source/utility.lisp
```diff
@@ -30,23 +30,30 @@ slime. Default port is 4006."
...
-      input-url)))
+  (let ((engine (assoc (first (cl-strings:split input-url))
+                       *search-engines* :test #'string=))
+        (default (assoc "default"
+                        *search-engines* :test #'string=)))
+    (if engine
+        (generate-search-query (subseq input-url
+                                       (length (first (cl-strings:split input-url))))
+                               (cdr engine))
+      (handler-case
```
Ambrevar: Make sure the indentation is correct.  Emacs should get this right.

FrozenPigs:
The latest commits should address all the points mentioned by @Ambrevar, though after moving *SEARCH-ENGINES* into WINDOW I occationally get this start up error:
```
    #<STANDARD-GENERIC-FUNCTION NEXT::SEARCH-ENGINES (1)>
  when called with arguments
    (NIL).
See also:
```

Ambrevar:
Hi,
sorry for the delays, there has been some more important architectural changes.  I've now merged your patch, it seems to be working well.  I've fixed a few issues.  Let me know if you experience the startup error.

Regarding `input-url` on line 52, it seems to be working fine by me.  Can you describe your problem more explicitly?
Also try pressing `C-RET` instead to validate a raw input without using the suggestion.

Ambrevar:
To change the defaults from your init.lisp:

```
(set-default 'window 'search-engines
             '(("default" . "https://duckduckgo.com/?q=~a")
               ("yt" . "https://www.youtube.com/results?search_query=~a")
               ("wiki" . "https://en.wikipedia.org/w/index.php?search=~a")))
```

Note that when your init.lisp is parsed, *interface* does not exist yet, so you can't change the object.  You can only change the defaults of the class.
Alternatively, you could change the object in a hook after *interface* initialization.  Hooks are a bit lacking at the moment though.  Suggestions are welcome!

Ambrevar:
Update: I've just replaced `set-default` with a setf-able `get-default`.
The right way to do it now is

```lisp
(setf (get-default 'window 'search-engines)
      '(("default" . "https://duckduckgo.com/?q=~a")
        ("yt" . "https://www.youtube.com/results?search_query=~a")
        ("wiki" . "https://en.wikipedia.org/w/index.php?search=~a")))
```
To append to existing engines, you can use
``` lisp
(nconc (eval (get-default 'window 'search-engines))
       '(("yt" . "https://www.youtube.com/results?search_query=~a")))
```

PR #216, opened 2019-05-20
documents/MANUAL.org: how to set search engines

vindarel:


jmercouris:
Thank you @vindarel!

Ambrevar:
Yup, that was very much needed ;)
Although I'd rather not advertise Youtube.  Something like Peertubes / Dtube instead?
Or simply Quickdocs?

PR #222, opened 2019-05-28
Remove dependency on Anaphora

vindarel:


Ambrevar:
Mark had introduced it, I guess it's a matter of style.  I'm personally not a big fan of anaphoric macros, I prefer `if-let`, etc.  So I will probably not use it.
@jmercouris?

jmercouris:
I am not a fan of them either

PR #224, opened 2019-05-29
parse-url: if input contains a point, suppose it's an url

vindarel:
so we go to the site.

the previous behavior was to search with the default search
engine. Typing "wikipedia.com" searched for wikipedia on
duckduckgo. Now we go to it. 

Yes ?

### DIFF
--- a/next.asd
+++ b/next.asd
@@ -9,6 +9,7 @@
   :defsystem-depends-on ("trivial-features")
   :depends-on (:alexandria
                :cl-strings
+               :str
                :cl-string-match
                :puri
                :sqlite
--- a/source/utility.lisp
+++ b/source/utility.lisp
@@ -20,6 +20,13 @@ SLIME."
   (swank:create-server :port swank-port :dont-close t))
 
 (defun parse-url (input-url)
+  "From user input, return the full url to visit.
+
+If the first word references a search engine, generate a search query.
+If the input starts with an uri scheme, open it as is.
+If the input is actually a file path, open it.
+Suppose the user omitted the scheme: if the input prefixed by 'https://' gives a valid uri, go to it.
+Otherwise, build a search query with the default search engine."
   (let* ((window (%%window-active *interface*))
          (engine (assoc (first (cl-strings:split input-url))
                         (search-engines window) :test #'string=))
@@ -39,6 +46,9 @@ SLIME."
             ((probe-file input-url)
              (format nil "file://~a"
                      (uiop:ensure-absolute-pathname input-url *default-pathname-defaults*)))
+            ((puri:uri-p (ignore-errors
+                           (puri:parse-uri (str:concat "https://" input-url))))
+             (str:concat "https://" input-url))
             (t (generate-search-query input-url (cdr default))))))))
 
 (defun generate-search-query (search-string search-url)


alexanderbarbosa:
hoho, I was about to do just that! That's neat! :D

source/utility.lisp
```diff
@@ -39,6 +46,8 @@ SLIME."
             ((probe-file input-url)
              (format nil "file://~a"
                      (uiop:ensure-absolute-pathname input-url *default-pathname-defaults*)))
+            ((str:contains? "." input-url)
+             (str:concat "http://" input-url))
```
jmercouris: I think it should always concatenate https instead of http

Ambrevar: Agreed.

Ambrevar:
I'd choose a different strategy since it won't work with "foo.bar baz":
if input is not a URL, prepend with "https://" and test with puri if the result is a valid URL.  If so, load it.  If not, go on with the search.

vindarel:
oh indeed. 

@alexanderbarbosa me first :)

vindarel:
done and rebased the commits.

source/utility.lisp
```diff
@@ -39,6 +46,14 @@ SLIME."
...
              (format nil "file://~a"
                      (uiop:ensure-absolute-pathname input-url *default-pathname-defaults*)))
+            ((str:contains? "." input-url)
+             (let* ((new-url (str:concat "https://" input-url))
+                    (is-uri (ignore-errors
+                              (puri:uri-p (puri:parse-uri new-url)))))
+               (if is-uri
+                   new-url
+                   (generate-search-query input-url (cdr default)))
+               ))
```
Ambrevar: Hanging parens ;)

source/utility.lisp
```diff
@@ -39,6 +46,14 @@ SLIME."
             ((probe-file input-url)
              (format nil "file://~a"
                      (uiop:ensure-absolute-pathname input-url *default-pathname-defaults*)))
+            ((str:contains? "." input-url)
```
Ambrevar: We can remove this check, nah?  Testing if `is-uri` should be enough.

next.asd
```diff
@@ -9,6 +9,7 @@
   :defsystem-depends-on ("trivial-features")
   :depends-on (:alexandria
                :cl-strings
+               :str
```
Ambrevar: Can we get rid of `cl-strings` then?

jmercouris: I believe cl-strings has some things that str does not have yet.

vindarel: Indeed, I had a look, and the only blocking function is cl-strings:insert. cl-strings also has functions to change case, which we don't use. I'll improve `str` :) Also, we'll have to compose with Quicklisp's delay (annoying mostly for users who install from source).

source/utility.lisp
vindarel: \+1. I ended with a cleaner logic but a repetition:

```
            ((puri:uri-p (ignore-errors
                           (puri:parse-uri (str:concat "https://" input-url))))
             (str:concat "https://" input-url))
```

you'd prefer it at the first `let*` ?

Ambrevar: I'm fine with both, it's a trivial repetition.  Feel free to merge :D

PR #226, opened 2019-05-31
command.lisp: define-command - typo in docstring

alexanderbarbosa:
Fix a minor typo on define-command docstring.

PS: I've seen a "]" in some of the **/source** files...I reckon someone fixed it already.


Ambrevar:
Good catch, thanks!
What do you mean with the "]"?

Issue #28, opened 2017-11-29
Augment the key binding facility (advanced VI features)


averrin:
I am emacs user, but big fun of vim-like bindings. So i use cVim and already cannot use other keys=(

I think something like spacemacs will be very popular

jmercouris:
Hi, thank you for your interest!

I am sorry to disappoint, but Evil mode is not planned, though something akin to hydra is planned, if you'd like to see evil mode incorporated, I am happy to accept any PR or suggestion on how to maybe incorporate Evil mode, we could put it in the timeline, and if it gets enough traction maybe we can even complete it :P

LaloHao:
A good starting point would be defining what evil-mode is in the browser.

So far i'd think of `jk` instead of `C-p` and `C-n`
Remapping `C-g` to `/` maybe?
`ESC` to exit minibuffer

Perhaps we could work out something together

jgkamat:
I don't think mapping j/k is feasable unless there's a mode system, as you won't be able to type j/k into web pages unless you switch modes. You could try `C-j` or `C-k` though. This could work out fine though if you just use an external editor to edit text fields.

averrin:
I need not vim in text fields, but i need j/k navigation and other one-letter shortcuts. Evil is about modes not text editing.

jmercouris:
one needs modes to distinguish “j” from typing in a “j” into a text field. to answer the question, nEXT does support modes, therefore one could program an evil emulation thing

HiPhish:
If I may chime in, I think one should distinguish between vi-keys for the browser interface and vi-keys for typing in text fields. vi-keys for the browser could be implement as simple mappings for the most part, but text fields are a different beast.

Rather than emulating a complex system I think it would be easier to integrate nEXT with Neovim. A headless Neovim process could be started (maybe lazily only after the user as actually entered a text input) and then the text field would serve as a a GUI for the headless Neovim. This would effectively give users their whole Neovim setup with all their customizations for entering text rather than just an approximation. I for instance rely a lot on key bindings because the default vi bindings don't make sense on a non-English keyboard.

I don't expect this sort of thing to be part of nEXT though, it should be a plugin instead.

There is something else I wonder: what if a web page introduces its own key bindings? Like, if it does something when the user presses `j`, does the page get the key before nEXT gets it? After? Does the key press get consumed or would both the web page react and the page get scrolled?

braham-snyder:
[...]

PR #235, opened 2019-06-07
Add VI bindings

Ambrevar:


alexanderbarbosa:
ITS HAPPENING :D

Ambrevar:
The following features are missing and should be implemented as soon as possible for an ideal experience.

- Allow setting vi-normal-mode by default.
- Declare known keymap scheme names (`:emacs`, etc.) to avoid typos.
- Define `keymap` structure (so that we have KEYMAP-P).
- Disable VI mode properly.
- Auto-switch to insert mode when clicking on HTML text fields
- Translate "shift + key" to "KEY", e.g. "s-:" should translate to ":" if unbound.
- Add support for override maps.

## Implementation details

If you look at the `vi.lisp` files, you'll see it's some 30 trivial lines of code for basic VI support!

I thought quite a lot about it and decided to first make Next as general as possible regarding the key facilities so that any key binding scheme could be implemented with the least amount of effort as possible.

In fact, the beauty of Next is that we don't depend on either the history of Emacs nor the limitations of WebExtensions or Firefox.  We have full control here, so let's design something neat from the ground up.

- Buffers

`buffer` is a Common Lisp class and defines a whole bunch of slots.
One of those slots, `modes`, holds the list of active modes for this buffer.
The first mode has highest priority.

Another slot is the `current-key-scheme` (see below).

- Modes

A mode is a class that runs some initialization code and some destruction code.
A mode has only one mandatory slot: the buffer it is associated to.
A common of use modes is to hold _key schemes_.

- Key schemes

A key scheme is roughly a plist of keymaps, e.g.

```
(:emacs my-emacs-map
 :vi-normal my-vi-map)
```

- Logic

When we press a key, Next will go through the active modes of the current buffer and fetch the keymap corresponding the `current-key-scheme` until it hits a binding.

This makes it possible to change binding scheme at run time, for individual buffers (or globally).

From there, VI implementation is trivial: the "insert mode" is just one binding scheme where `ESCAPE` is bound to `vi-normal-mode` and vice-versa, with `i` being bound to `vi-insert-mode` in vi-insert-mode.

This does not cover the advanced features mentioned in #28, such as
- Automatic resolution of conflicting key bindings with a timeout
- Command repeating
- Numeric argument
- Macros

All this will have to wait for some time :)

I'm planning to merge on Monday.
Thoughts? @jmercouris @vindarel @wasamasa @lehmacdj @jgkamat @noctuid

Ambrevar:
Update:
- VI can now be set to be the default with
```
(add-to-default-list 'vi-normal-mode 'buffer 'default-modes)
```

- VI mode can now be properly de-activated.

- Shifted keys are properly translated, e.g. "s-o" is translated to "O" and "s-:" is translated to ":".

jmercouris:
Everything looks OK to me so far. Will the missing features be implemented in a separate branch?

Ambrevar:
Incoming just now!

source/keymap.lisp
```diff
@@ -1,27 +1,59 @@
...
+                         for fun = (gethash key keymap)
+                         unless fun
+                           do (log:debug "Key-chord ~a normalized to ~a" key key-normal)
+                              (setf fun (gethash key-normal keymap))
+                         when fun
+                           return (cons fun mode))))
+    (when fun+mode
+      (if (eq (first fun+mode) #'prefix)
+          (first fun+mode)
+          (lambda () (funcall (first fun+mode) (cdr fun+mode)))))))
```
vindarel: nitpicking: mixed "first" and "cdr".

source/keymap.lisp
```diff
@@ -42,103 +74,60 @@ For example, it may add C-M-s or C-x to a stack which will be consumed by
...
-         (key-maps (list (root-mode-default-keymap) local-map))
-         (serialized-key-stack (mapcar #'serialize-key-chord (key-chord-stack *interface*))))
-    (dolist (map key-maps)
-      (let ((bound-function (gethash serialized-key-stack map)))
-        (cond ((equalp "prefix" bound-function)
-               (return-from consume-key-sequence t))
-              (bound-function
+          ((minibuffer-active active-window)
+           (if (member "R" (key-chord-modifiers (first (key-chord-stack *interface*)))
+                       :test #'string=)
```
vindarel: Maybe this pattern can be extracted, I saw it twice or more (`"member … (key-chord-modifiers (first …) :test #'string=`).

source/mode.lisp
```diff
@@ -7,34 +7,72 @@
...
-
-(defun root-mode-default-keymap ()
-  "Return the default keymap of root mode."
-  (eval (closer-mop:slot-definition-initform
-         (find-slot (find-class 'root-mode) 'keymap))))
+     (buffer :accessor buffer :initarg :buffer)
+     (activate :accessor activate :initarg :activate) ; TODO: This can be used in the future to temporarily turn off modes without destroying the object.
+     (destructor :accessor destructor  ; TODO: Better name?
+                 :documentation
+                 "A lambda function which takes the mode as argument.")
```
vindarel: better docstring for destructor and docstrings for other slots ?

vindarel:
\+1. Don't forget the user and developer documentation though. You already explained quite a bit in your above comment.

source/keymap.lisp
Ambrevar: Do you know the equivalent of `first` for `cdr`?
Or do you mean I should use `car` and `cdr` here?

source/keymap.lisp
Ambrevar: Yes, this `(member ... :test #'string=) is a bit verbose.  Isn't there some higher-level function in the standard already?  Maybe in the library?

source/keymap.lisp
vindarel: yes `cdr` == `rest`.

(https://lispcookbook.github.io/cl-cookbook/data-structures.html#carcdr-or-firstrest-and-second-to-tenth ;) )

Ambrevar: Oops!  Must have been tired on this one... :)

source/minibuffer.lisp
```diff
@@ -386,26 +386,42 @@
...
+                          (body :border-top "4px solid dimgray"
+                                :margin "0"
+                                :padding "0 6px")
+                          (p :margin "0")))))
+            (set-input minibuffer
+                       (cl-markup:markup
+                        (:head (:style style))
+                        (:body
+                         (:p (apply #'format nil args)))))))
+        (log:warn "Can't echo ~a without minibuffer or interface" args))))
```
jmercouris: Possibly also echo what it was trying to echo to stdout here?

Ambrevar: Hmm?  Isn't it exactly what `log:warn` does?

noctuid:
I think the idea of key schemes is much better than trying to imitate what evil does. Can only one key scheme be active at a time? It seems like it would make sense to be able to have inheritance in some cases like evil does.

Is there currently some hook for when a new url is opened that could be used to set up local modes (e.g. I create a mode for reddit and enable it if the url matches reddit.com)?

Ambrevar:
Only one key scheme is active at a time for a buffer.
By inheritence, you mean inheritance across schemes?

For instance if C-x C-c is defined in the `:emacs` scheme but not in VI, you'd like that it still works when the VI scheme is selected?

I'm not too sure about that, since it could be confusing to the user.  If we enable inheritance, can the user disable it?  How?  Can it be disable on a per-mode / per-buffer basis?

Hooks: while it's possible to program something, a convenient interface for hooking is missing at the moment.  I'll work on it soon, probably when I get down to adblocking.

vindarel:
> a convenient interface for hooking is missing at the moment.

We can use CLOS's before/after methods for our classes, but it's like we had one single hook.

(https://lispcookbook.github.io/cl-cookbook/clos.html#method-qualifiers-before-after-around)

jmercouris:
My vote is to restore the old hook logic, I think it was pretty simple/straightforward.

noctuid:
> By inheritence, you mean inheritance across schemes?

Yeah, across schemes.

> For instance if C-x C-c is defined in the :emacs scheme but not in VI, you'd like that it still works when the VI scheme is selected?

That could be something someone would want, so I think it would make sense to allow that, but I was thinking more like evil's inheritance between states. Some keybindings will potentially make sense to have in multiple key schemes (e.g. normal and with a visual selection; probably a lot less keybindings will make sense in multiple key schemes than do for evil). You could duplicate them or have a variable to hold common keybindings, but I think it would make more sense to allow the user to change keybindings for both in the same place using `define-key`. There would be a key scheme whose only purpose was to hold common keybindings. It could be simpler to just bind in multiple keyschemes at once, but this seems a little less clean to me and could also potentially cause confusion. It's something worth considering I think.

> If we enable inheritance, can the user disable it?  How? Can it be disable on a per-mode / per-buffer basis?

Make the current active schemes optionally a list. It being global would be fine for the use case I had in mind, but having a buffer-local `current-key-scheme` might be useful in some cases. By default, I wouldn't do anything like inherit `:emacs` in a vi scheme.

Ambrevar:
@jmercouris: The old hook system had a few issues, among which that it only ran at a predefined point in a command.  So it would not run on class initialization for instance, and it would not run before, after or around like is customary for auxiliary methods (or Emacs `advice` system).

I suggest we keep thinking about it for a couple of days until we find something that flexible enough: in the end it's a feature that is going to be exposed to the user and that will be key to customization.  For everything.

Ambrevar:
@noctuid: Lots of good points here.  Merging schemes would be easy, but then you lose the ability to set a key in one place.  Hmm... Need to think a bit more about it.  Anyways, I'll probably merge this and keep inheritance for later.

jmercouris:
Merge it! I’m excited to see it on master!

PR #238, opened 2019-06-12
Titlecase

Ambrevar:
I've renamed the `s` modifier to `Shift`, and while I was at it, I thought it would be nice to enforce consistency.

This brings up some questions:
- Do you people like titlecase for special keys?  Lowercase would be more Lispy.
- We lack type-checking when it comes to the special key names.  For instance binding "shit-Return" won't raise an error and the user might stay befuddled until they figure they forgot the "f". Can we do something about it?

Regarding the last question, well, we've used strings to encode key-chord, but should we?  We did it this way so far taking inspiration from Emacs / StumpWM, but we don't have to.  In fact, this is Lisp, so why not leveraging the language?  We could use some nifty macros to encode keycodes.  Some examples:

- "a" -> (a)
- "a b" -> (a b)
- "Return" -> (return)
- "Dummy" -> (dummy) ; can error out at macro-expansion time
- "C-x s" -> `((c x) s)`
- "C-M-s" -> `((c (m s)))`

Basically a key binding is a list of symbols.
Symbols are lookep up in a map.  If it does not exist and is a single character, insert this character.
In a sublist, the first character is a modifier.  Sublists can be nested, which enables us to stack modifiers.

We could even encode timeouts, for instance to bind a quick "j k" to vi-normal-mode:
`((timeout j 0.2) k)`, where 0.2 is the timeout in seconds.

And that's the beauty of it: we would be able to extend this mini-language while remaining backward-compatible.  Something that's way harder to do with strings.

And this brings us back to the topic of this PR: if we adopt a Lispy syntax to encode bindings, then should we still use titlecase?

@vindarel @jmercouris 

Edit: s/compile/macro-expansion/

### DIFF
--- a/ports/gtk-webkit/buffer.h
+++ b/ports/gtk-webkit/buffer.h
@@ -16,7 +16,7 @@ typedef struct {
 
 // See the documentation of "enum GdkModifierType".
 static Modifier modifier_names[] = {
-	{.mod = GDK_SHIFT_MASK, .name = "s"},
+	{.mod = GDK_SHIFT_MASK, .name = "Shift"},
 	{.mod = GDK_LOCK_MASK, .name = "Lock"},
 	{.mod = GDK_CONTROL_MASK, .name = "C"},
 	{.mod = GDK_MOD1_MASK, .name = "M"}, // Usually "Alt".
--- a/ports/gtk-webkit/buffer.h
+++ b/ports/gtk-webkit/buffer.h
@@ -208,7 +208,7 @@ gboolean buffer_web_view_decide_policy(WebKitWebView *_web_view,
 	gchar *mouse_button = "";
 	if (action) {
 		guint button = webkit_navigation_action_get_mouse_button(action);
-		mouse_button = g_strdup_printf("button%d", button);
+		mouse_button = g_strdup_printf("Button%d", button);
 
 		guint modifiers = webkit_navigation_action_get_modifiers(action);
 		for (int i = 0; i < (sizeof modifier_names)/(sizeof modifier_names[0]); i++) {
--- a/ports/gtk-webkit/window.h
+++ b/ports/gtk-webkit/window.h
@@ -19,13 +19,13 @@ typedef struct {
 } KeyTranslation;
 
 static KeyTranslation key_translations[] = {
-	{.old = "", .new = "BACKSPACE"},
-	{.old = " ", .new = "SPACE"},
-	{.old = "", .new = "DELETE"},
-	{.old = "-", .new = "HYPHEN"},
-	{.old = "", .new = "ESCAPE"},
-	{.old = "\r", .new = "RETURN"},
-	{.old = "\t", .new = "TAB"},
+	{.old = "", .new = "Backspace"},
+	{.old = " ", .new = "Space"},
+	{.old = "", .new = "Delete"},
+	{.old = "-", .new = "Hyphen"},
+	{.old = "", .new = "Escape"},
+	{.old = "\r", .new = "Return"},
+	{.old = "\t", .new = "Tab"},
 };
 
 guint window_string_to_modifier(gchar *s) {
@@ -421,7 +421,7 @@ gboolean window_button_event(GtkWidget *_widget, GdkEventButton *event, gpointer
 		}
 	}
 
-	gchar *event_string = g_strdup_printf("button%d", event->button);
+	gchar *event_string = g_strdup_printf("Button%d", event->button);
 	return window_send_event(window,
 		       event_string, event->state,
 		       0, event->button,
@@ -487,7 +487,7 @@ gboolean window_scroll_event(GtkWidget *_widget, GdkEventScroll *event, gpointer
 	}
 	}
 
-	gchar *event_string = g_strdup_printf("button%d", button);
+	gchar *event_string = g_strdup_printf("Button%d", button);
 	return window_send_event(window,
 		       event_string, event->state,
 		       event->direction, button,
--- a/source/document-mode.lisp
+++ b/source/document-mode.lisp
@@ -68,7 +68,7 @@
           "C-f" 'scroll-page-down
           "C-b" 'scroll-page-up
           "SPACE" 'scroll-page-down
-          "s-SPACE" 'scroll-page-up)
+          "Shift-SPACE" 'scroll-page-up)
         (list :emacs emacs-map
               :vi-normal vi-map))))
   ;; Init.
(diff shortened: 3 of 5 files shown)


jmercouris:
I think UPCASE would be more LISPY. I like the idea of using symbols and lists. I need to think about it some more before giving a more refined answer.

Ambrevar:
I think you mean that the printed representation is upcase, but here I
meant the written representation, that is, how we write the bindings.
No one writes Lisp code in uppercase :D

vindarel:
If you rename `s` to `shift`, why not use `control` and `meta`, specially in the potential lispy notation ? `(c (m x))` vs `((control (meta x))`. In that case I'm all for lowercase. Otherwise, I think that an upper letter in `C` or `Return` denote that the key is somewhat special. It isn't a letter.

Also, let's rename `meta` to `alt` ! (or allow both)

Good ideas again. Could we not keep a string notation (simpler for the user), but we parse it, so we are able to fail on malformed keybindings too? `"C-M-x"` is parsed to an internal structure, which is checked for validity. But sure, if we write a lispy keybinding we can script the keybinding… (`(timeout j *default-timeout*…)`. infinite extensibility O_o (do we really want this?)

Yesterday I found `((C (M x))` really strange, it's better now :]

Ambrevar:
I had the exact same thoughts yesterday!

I think (c (m x)) is shorter which is good for something that's used so often, but true so it's more confusing.  For
instance `((c (m c) m))` starts being confusing.

Solution: We could use uppercase there too: `((C (M c) m))`.

That said, Alt is Option on mac, and Super is the infamous flag on many
keyboards, while it's Command on mac.

The benefit of using `M` and `S` is that they are rather neutral.

Or at least `meta` and `super` do not stand for one or the other, so
they are more universal.  All in all, I'm rather opposed to naming it `alt`.

String parsing: yes, we can totally do it too, you are right.

Lisp flexbility: also right, it's cool that we can script the
keybindings, but really I can't think of a single use case :D

Ambrevar:
(Beside your timeout example, but this could be implemented differently too.)

jmercouris:
Thoughts:
- I like Vincent's more explicit writing out of "shift" and "super" in the the list. This I think it is better than the sometimes capitalized letters. For anyone who doesn't have case sensitive on, everything will be up-cased and thus ambiguous. This may or may not be a problem depending on how you parse these s-exp, it will however add overhead if you want to show useful error messages to the user, you'll have to write a function to make a reasonable/understandable printed representation.
- I agree with Pierre's assessment of using the more neutral terms. On keyboard the symbols manifest in many different ways, sometimes super is a Windows key, sometimes it is a Command Key, etc.

The one example that I am not understanding is this:
- "C-M-s" -> `((c (m s)))`

Why would this not produce instead `(c m s)` (or in the new syntax `(control meta s)`)?

Ambrevar:
I don't see why the printed case would be a problem when reporting to the user: it's like the rest of the code, the error message can be printed the way we want.

About `(c m s)`: How do you tell the difference between "c m s" and "C-M-s" in the new notation?
My proposal was to use nested list.  Anything symbol as the top level is a key, while a sublist is a modifier.

jmercouris:
Aha, I see now.

Well you could tell by the fact that the keywords are "control" "meta" and not "c" and "m". I think it would be harder by trying to stick to single letter symbols.

Ambrevar:
Good point.  So with full words, then we can have unambiguously `(control meta x control super s n)`.  But that's harder to parse in code.  Might also be harder to read for human beings than `((control meta x) (control super s) n)`.

jmercouris:
I was actually trying to suggest your second suggestion (I think we are on the same page):

`(control s (control x))`

that would represent `C-x C-s`. Which leads to the question do we actually need to nest s-exp?Because it could also have been written like this:

`((control x)(control s))`

thoughts?

Ambrevar:
`(control s (control x))` was not my suggestion, this is a hybrid of my 2 suggestions :)
I find it less logicial, harder to make sense of from the user's perspective and harder to parse.
I prefer `((control s) (control x))`.

jmercouris:
Hi Pierre, just to be sure

`((control s) (control x))` would be `C-s C-x` in emacs notation?

vindarel:
just don't go too deep on that notation until we choose to do it ;)

>  String parsing: yes, we can totally do it too, you are right.

Ambrevar:
@jmercouris: Yes it would!
@vindarel: Agreed!

jmercouris:
Ok, I think we have reached consensus then. I am OK with the most recent proposed notation of

`((control s) (control x))`

Ambrevar:
We haven't answered the most important question: What about sticking to the current strings and do type-checking on them?

If we do stick to strings, should we use "C", "Control" or "control"?  "Return" or "return"?

jmercouris:
Oh I see. I think we don't have to stick to strings, but we can build support for them later, that is what I had understood from @vindarel 

So it would be (in order of implementation):

1. symbol notation
2. string -> symbol notation macro

Ambrevar:
I think there is a misunderstanding: my point was to get rid of strings, not to keep both.  In fact one of your points previously was that we didn't want multiple notations for keys (and I agree with that).

So the user would do

``` lisp
(define-key ((C x) (C c)) 'kill)
```

jmercouris:
Ok, now I finally understand. `(define-key ((C x) (C c)) 'kill)` works OK for me as a complete replacement for the current string based system *and* without a string converter macro.

I would like to point out though that in your example, `DEFINE-KEY` will have to be a macro or the list otherwise quoted as such: `(define-key '((C x) (C c)) 'kill)`.

vindarel:
> my point was to get rid of strings, not to keep both.

disagree:

> We lack type-checking when it comes to the special key names

your point was the lack of type checking in keybindings notations. So, could we keep the string notation, parse it and warn users as well as with the list notation ? I would prefer this by much, unless we can think of obvious benefits of the list notation, but it seems we don't think of any:

> Lisp flexbility: also right, it's cool that we can script the keybindings, but really I can't think of a single use case :D

Ambrevar:
@jmercouris: You are correct, define-key must be turned to a macro.

@vindarel: Well, no it was not my _original_ point, I was just
explaining why I thought we should only keep one.

So I vote for the simplest road then: keep the strings and add type
checking.

Then what about the casing and the full-word modifiers?

vindarel:
> Well, no it was not my _original_ point,

sorry, got it.

> So I vote for the simplest road then: keep the strings and add type checking. 

\+1 

> Then what about the casing and the full-word modifiers?

no strong opinion. I'm fine with  capital `C` or `control`. Might prefer `return` over `Return`.

jmercouris:
May we close this PR and revisit it in a new branch after the FFI changes?

Ambrevar:
I suggest we open an issue linking to this broken PR and summing up the conclusion.

Ambrevar:
Closing and leaving the discussion to #564 .

PR #241, opened 2019-06-15
feat: add Roswell support

sheepduke:
I read the documentation and the `Makefile` source code and found that actually `(asdf:make)` is called to build the Lisp core.

Many Lipers, including me, are using Roswell to maintain the REPLs and systems. 

In my humble opinion it would be really nice to be able to build the Lisp core using Roswell, so that local systems will be reused.


sheepduke:
By the way is it encouraged to submit Pull Requests like this, i.e. out of the TODO plan?

alexanderbarbosa:
Next as quicklisp package would be great too! hehe

documents/README.org
```diff
@@ -106,6 +106,14 @@ these fixes.
 
 ** Compile
 
+The Next program consists of 2 dedicated executable files: the Lisp core and
+the WebKit port. There are 2 ways to build them according to whether you want
```
Ambrevar: Replace "WebKit" with "platform", a PyQt platform port is coming soon.

documents/README.org
```diff
@@ -106,6 +106,14 @@ these fixes.
...
+to use [[https://github.com/roswell/roswell][Roswell]].
+
+*** Not Using Roswell
+
+The following command will build both Lisp core and the WebKit port.
```
Ambrevar: Same.

documents/README.org
```diff
@@ -121,6 +129,26 @@ global installation of Quicklisp, or to automatically download and
 install a separate Quicklisp installation for Next. Run ~make~ to display some
 documentation. Please see the Makefile for more details.
 
+*** Using Roswell
+
+Roswell can be used to build the Lisp core. A Roswell script is provided under
+=roswell/= directory. In order to compile it, simple run:
```
Ambrevar: Can we move the `next.ros` file at the root?

documents/README.org
```diff
@@ -121,6 +129,26 @@ global installation of Quicklisp, or to automatically download and
...
+
+#+BEGIN_SRC sh
+ros build next.ros
+#+END_SRC
+
+An executable file will be generated under =roswell/= directory. Please note
+that this method uses the systems maintained by the Roswell.
```
Ambrevar: "The" Roswell?

documents/README.org
```diff
@@ -121,6 +129,26 @@ global installation of Quicklisp, or to automatically download and
...
+
+Afterwards, you need to build the WebKit port:
```
Ambrevar: Platform port... Also mention that this is just an example, since soon there will be another command to build the PyQt port.
Run `make` for the doc.

roswell/next.ros
```diff
@@ -0,0 +1,17 @@
...
+#|-*- mode:lisp -*-|#
+#|
+exec ros -Q -- $0 "$@"
+|#
+(progn ;;init forms
+  (ros:ensure-asdf)
+  #+quicklisp(ql:quickload '(next) :silent t)
+  )
+
+(defpackage :ros.script.next.3769603967
```
Ambrevar: What's this magic number?

documents/README.org
jmercouris: Or, we make a build directory and put the guix.scm file and the next.ros file in there?

jmercouris:
A welcome patch! Thank you for spending some time on this! However, one thing I do not like is the the compilation instructions are "With Roswell" and "Not with Roswell". I believe it would be better if it read "Compile Image with LISP" and "Compile with Roswell".

jmercouris:
> Next as quicklisp package would be great too! hehe

A good idea! It is not so simple though, QL is for libraries, not applications. One could argue that Next is a library... At any rate, we would like to maintain a little bit more control over updating and distributing Next and its dependencies.

alexanderbarbosa:
@jmercouris 
Interesting, I may be get confused as StumpWM is available at quicklisp.

Anyway, would be great CL packages have its own package manager.. :D

documents/README.org
sheepduke: In my understanding the Qt port is also a wrapper of WebKit right?

I mean, it is like gtk-webkit vs qt-webkit.

Am I wrong?

documents/README.org
sheepduke: For Roswell the location is not important as long as the system can be found as "local project".

I will add this requirement to the doc.

documents/README.org
sheepduke: Erhh...Wrong use of "the"?

documents/README.org
sheepduke: > Run make for the doc

What does it mean? Could you please elaborate?

roswell/next.ros
sheepduke: This is a random number generated by Roswell to avoid name mangling.

It can be removed since it is only a trivial symbol.

It looks nicer without it in fact.

documents/README.org
sheepduke: > build dire

In my humble opinion `build` might not be a good name...

sheepduke:
> Interesting, thanks for the patch! I am not very familiar with Roswell but I plan to give it a shot soon. Can you share some pros and cons compared to Quicklisp or Guix?

I am not familiar with Guix, so I cannot help here.

Roswell plays a different role with Quicklisp. Quicklisp is a system manager, while Roswell is a implementation manager (like PyEnv) plus a script launcher:

* It can install and manage different Lisp implementations and makes it easy to switch between them.
* It can generate `*.ros` files that can be used as a script. The start-up time would be a little bit long (around 1 second accordingly) compared to bare-metal binaries, but it is OK.

documents/README.org
Ambrevar: The PyQt platform port will use QtWebengine, which is based on Blink (Chrome's renderer).

documents/README.org
Ambrevar: Agreed, it sounds like a folder where build files will end (CMake does that).
What about `build-scripts`, `recipes`, `scripts`?
Or else simply at the root :)

documents/README.org
Ambrevar: I don't know.  If it's customary to say "the Roswell", then no problem.  Otherwise if Roswell is a program / system like any other, I would remove "the" there.

documents/README.org
Ambrevar: Try it: go in the Next folder and run `make`: it will print out a help message.
I suggest you mention it as well along those lines.

roswell/next.ros
Ambrevar: Thanks!

Ambrevar:
Guix also features an implementation manager (`guix environment`) which has the great benefit of not only allowing the user to select a Common Lisp implementation, but also any (possibly non-Lisp) package!  Which is great for libraries / programs that depend on, say, SQLite, etc.

Regarding the scripting possibilities, I've never tried it.  I guess it's a portable way to run scripts, while `sbcl --script` only works for sbcl.

Regarding the executable: my understanding is that with standard ASDF you need a `program-op` in your `.asd` and then run `(asdf:make)`.  It's cool that Roswell can do it for any ros script, but I'm not sure we are gaining much here for Next in this regard.

documents/README.org
sheepduke: Understood.

documents/README.org
sheepduke: I think for now the root directory is OK. This might be changed later if needed.

documents/README.org
sheepduke: Reasonable. Removing "the" makes it read smoother.

sheepduke:
> Regarding the executable: my understanding is that with standard ASDF you need a program-op in your .asd and then run (asdf:make). It's cool that Roswell can do it for any ros script, but I'm not sure we are gaining much here for Next in this regard.

I think the point of providing Roswell here is to make the building process more convenient for Roswell users. For a typical Roswell user, there are following inconveniences:
* Since `ros run` is used to invoke Lisp REPL, there is no `sbcl` command in the system. So I had to make it in the `$PATH` manually.
* Quicklisp is installed to `~/.roswell/lisp/quicklisp/` so it is required to provide several build flags in order to use my existing systems. Compared with simply running `ros build`..what more can I say?

PR Updated.

documents/README.org
jmercouris: Final change: Can you please put it into a subfolder called `build-scripts`?

Ambrevar:
Note that regarding those last two points, our Makefile supports them both.  Run `make` and you'll see that two variables are documented, LISP and NEXT_INTERNAL_QUICKLISP.  For instance you can call

```
make LISP=/path/to/your/sbcl NEXT_INTERNAL_QUICKLISP=false
```

You can even specify a path to the quicklisp directory with the QUICKLISP variable.

That said, I understand that there is a Roswell user base out there that would like to see Next behave as expected, and the ros script is completely non-intrusive, so I think it's a good idea to include it.

documents/README.org
Ambrevar: @sheepduke: Would you like to change the folder to `build-scripts`?  I'll merge then.

sheepduke: @jmercouris Sorry, I did not notice your reply. My fault.

@Ambrevar Done.

Ambrevar:
Thanks!

jmercouris:
Indeed nhooks is not in Quicklisp. You'll need to clone it to some place CL can find it. It is at atlas-engineer/nhooks.

glyh:
Thank you, I got the system installed via the following commands:
```
ros install atlas-engineer/nhooks omar-polo/phos atlas-engineer/nyxt
```
But this does not seem to create an executable in ~/.roswell/bin, and roswell complains it did not find any roswell scripts.

Ambrevar:
Indeed, Nyxt does not support Roswell (for now, at least), hence this
issue ;)

PR #242, opened 2019-06-19
Download manager

Ambrevar:
A first draft for the pure-lisp download manager, and it already works!

Next steps:
- Refactor the library into a `libraries/download-manager` subdirectory.
- Update progress live.
- Add authenticated download support.
- Add FTP support.
- Add magnet support (for now we can just leverage Transmission's web interface).
- Allow stopping downloads.
- Optional: Allow pausing and resuming downloads.


source/download-mode.lisp
```diff
@@ -0,0 +1,30 @@
...
+(define-command download-list (root-mode &optional (interface *interface*))
+  "Display a buffer listing all downloads."
+  (let* ((download-buffer (or (find-buffer 'download-mode)
+                              (make-buffer
+                               :name "*Downloads*"
+                               :default-modes (cons 'download-mode
+                                                    (get-default 'buffer 'default-modes)))))
+         (contents (cl-markup:markup
+                    (:h1 "Downloads")
+                    (:span              ; TODO: Do we need this span?
```
jmercouris: You do not need a span necessarily it could just as well be `<p>`, however to collect and output something in cl-markup you do need to have a tag of some sort.

source/download.lisp
```diff
@@ -0,0 +1,112 @@
+;;; Download manager backend.
+(in-package :next/engine)
+
+(defmethod cache ((type (eql :uri)) uri)
+  (locally-cache uri))
+
+(defun downloads-directory ()
+  (ensure-directories-exist (truename #p"~/Downloads/")))
```
jmercouris: Not sure I agree with this, on the other hand I can't think of a different download directory on any system...

source/download-mode.lisp
Ambrevar: So we need a span, right? :p

source/download.lisp
Ambrevar: From the XDG standard, it should be `XDG_DOWNLOAD_DIR="$HOME/Downloads"`.  So I suggest we first check the environment variable, then default to ~/Downloads.

Ambrevar:
I've fixed the first point: refactor the library to a separate directory.

source/download.lisp
Ambrevar: Actually we should first query the `xdg-user-dir` system tool.  Is there a way to do this in pure Common Lisp?

source/download-mode.lisp
jmercouris: effectively yes :D

alexanderbarbosa:
Wow

source/download.lisp
alexanderbarbosa: nope, could not find any xdg implementation...

source/remote.lisp
```diff
@@ -135,7 +135,29 @@ commands.")
...
+(defun download-watch ()
+  ;; TODO: Add a (sleep ...)?  If we have many downloads, this loop could result
+  ;; in too high a frequency of refreshes.
+  (loop while (lparallel:receive-result download-manager:notifications)
+        do (let ((buffer (find-buffer 'download-mode)))
+             ;; Only update if buffer exists.  We update even when out of focus
+             ;; because if we switch to the buffer after all downloads are
+             ;; completed, we won't receive notifications so the content needs
+             ;; to be updated already.
+             ;; TODO: Disable when out of focus?  Maybe need hook for that.
```
jmercouris: We don't have a `buffer-changed` hook. I also don't think we have a `buffer-focused` method. We do have `buffer-active`, but it may still be on a non-visible screen, and therefore not actually in focus.

source/proxy-mode.lisp
```diff
@@ -6,7 +6,7 @@
     "Enable forwarding of all network requests to a specific host.
 This can apply to specific buffer."
     ((server-address :accessor server-address :initarg :server-address
-                     :initform "socks://127.0.0.1:9050"
+                     :initform "socks5://127.0.0.1:9050"
```
jmercouris: what is `socks` and `socks5`?

Ambrevar: That's because there are multiple SOCKS protocols (socks4, socks5, etc.).
WebKitGTK understands "socks" as socks5, and thus it works for Tor (which needs socks5).
Dexador does not get it however, so we need to specify socks5 explicitly.

source/remote.lisp
Ambrevar: Yes, we need a buffer changed hook, or at least something that can hook on `set-url`.
Shall we implement hooks before or after 1.3?

Ambrevar:
Update: Download progress is now displayed! :D 

I've investigated authenticated downloads, and it seems that I need to add an end point to forward a serialized cookie-jar.  Dexador accepts them.  It should not be too hard.  Stay tuned.

source/remote.lisp
jmercouris: after, it isn't mission critical

jmercouris:
Nice progress! A serialized cookie-jar? If the location of the cookies is known, why do you need WebKit to serialize them? Are they in some sort of proprietary format?

Ambrevar:
No they are not, but I don't understand the format completely.  (It's Firefox' format.)
I wan't to serialize them with WebKit to know how they look like structurally so that I can re-generate them in pure list from `cookies.txt`.  I'll experiment with it and see what I can do.

libraries/download-manager/native.lisp
```diff
@@ -0,0 +1,111 @@
...
+                         :resolved-uri resolved-uri
+                         :header response-headers
+                         :file file
+                         :status status
+                         :downstream stream)))
+    (error (c)
+      (error c))))
+
+;; TODO: Rename DOWNLOAD method to FETCH?
+(defmethod download ((download download)
```
jmercouris: I'm not sure if fetch makes it more clear since it is a download manager. If I had to rename it I would pick `READ` because I see no semantics difference between reading on a local or a remote device. I am still reading from some I/O interface.

Ambrevar: You are right, `READ` is probably better.

Ambrevar:
I take that back: I think we need to serialize the cookies between the platform port (here webkit) and Lisp because of "session cookies" which are not persisted to disk.

Ambrevar:
And... Done!  Authenticated downloads work :D 
There might be some edge cases that we need to tune, but I guess it's a good start.

I'll keep FTP support and the other tasks for later, probably post 1.3.

So I think this is ready for merding.  @jmercouris Let me know what you think and let's merge!

libraries/download-manager/native.lisp
```diff
@@ -0,0 +1,94 @@
...
+                           :requested-uri requested-uri
+                           :resolved-uri resolved-uri
+                           :header response-headers
+                           :file file
+                           :status status
+                           :downstream stream)))
+      (error (c)
+        (error c)))))
+
+(defmethod fetch ((download download)
```
jmercouris: I don't agree with the name fetch, but it is OK

libraries/download-manager/native.lisp
```diff
@@ -0,0 +1,94 @@
...
+  "Return the number of bytes fetched."
+  (let* ((buffer (make-array buffer-size :element-type '(unsigned-byte 8))))
+    (with-open-file (output (file download)
+                            :direction :output
+                            :if-exists :supersede
+                            :element-type '(unsigned-byte 8))
+      (log:info "Downloading~%  ~a~%to~%  '~a'.~%"
+                (resolved-uri download) (file download))
+      (loop :for byte-position = (read-sequence buffer (downstream download))
+
```
jmercouris: Why the extra newlines here?

jmercouris:
I of course can't test, especially the GTK stuff, but the Lisp code looks logical to me. I like the fact that this is broken out into a library. I think it is ready to merge! Well done! And as always, I'm sure we will iterate on this many times as we learn more when we try to use it, so even if I missed something no worries. I would not say we are afraid of rewrites :D

libraries/download-manager/native.lisp
Ambrevar: As I explained in the commit message, the problem with `READ` is that is conflicts with CL's `READ`.  `FETCH` is just easier.

libraries/download-manager/native.lisp
Ambrevar: Hmmm, well just to highlight that the 3 paragraphs underneath are separate actions.  Not really useful.

Ambrevar:
Thanks!  Alright, I'll merge later this weekend then!

libraries/download-manager/native.lisp
jmercouris: Oh I see, READ is a function?

jmercouris:
Great!

libraries/download-manager/native.lisp
Ambrevar: Yes, it's the Lisp reader :)

jmercouris: Ah, for some reason I thought it was a method or something... whoops :D

alexanderbarbosa:
Wonderful. I wont need chromium for anything, now! :D

PR #243, opened 2019-06-22
Download Manager: remove test (next.asd)

alexanderbarbosa:
Hi, 

next.asd has a defsystem **download-manager/tests** that requires prove-asdf package, issuing an error as Next is building. 

Removing it and Next builds correctly.

PS2: at last!

### DIFF
--- a/next.asd
+++ b/next.asd
@@ -69,22 +69,13 @@
   :entry-point "next:start-with-port")
 
 (asdf:defsystem download-manager
-  :depends-on (lparallel
-               log4cl
-               dexador
-               quri
-               cl-ppcre
-               str)
-  :components ((:module source :pathname "libraries/download-manager/"
-                :components ((:file "package")
-                             (:file "engine")
-                             (:file "native")))))
-
-(asdf:defsystem download-manager/tests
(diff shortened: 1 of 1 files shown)


jmercouris:
Hi, while it is nice that it does build now for you, it should always build. Instead of removing the test, can you find a way to make the test work for you? Did you quick load `download-manager/tests`? That should allow you to load the system without having to delete the test.

alexanderbarbosa:
fair enough!

Ambrevar:
Thanks for reporting.  I've updated the Makefile to include prove-asdf
in the dependencies.

That said, I wonder how ql:quickload is supposed to work with
`:defsystem-depends-on`.

Is there a way to automate it?  Mark had inserted the following bits:

;;; TODO fix upstream ASDF to successfully locate and load
;;;      DEFSYSTEM-DEPENDS-ON clauses
(eval-when (:compile-toplevel :load-toplevel :execute)
  (unless (ignore-errors (asdf:find-system :prove-asdf))
    (ql:quickload :prove-asdf)))

which is mostly a workaround, but still a nice one.  Thoughts?

PR #261, opened 2019-07-05
Bookmark: rename function, binds semantic, add table to manual.org

alexanderbarbosa:
Hi,

This a small commit, tested in emacs and vi bindings:

- Rename `set-url-from-bookmark` function to `bookmark-open-bookmark`
Cosmetic change, keeping all bookmarks functions starting with `bookmark-...`

- Move all bookmarks bindings to document-mode
There are bindings everywhere, maybe a dedicated `bindings.lisp` file would be interesting. Anyway, I just bundled all bookmarks bindings together.

- Bookmarks bindings following its name: bookmark-anchor (c-m a)
At first I tried to keep all bindings following its function name, but I noticed that some would be easier to type with nearest letters  instead. So I just changed two bindings while keeping delete and current intact.

- Update bookmark se change than importantction on manual.org
Vim bindings are not yet on Manual.org, so I add it and a table.

PS: If you all agree I might improve manual.org by adding tables, improving text and adding lacking bindings and documentation...

:D

### DIFF
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -219,16 +219,20 @@ forward command will simply visit the first child of the current node
 in the tree.
 
 ** Bookmarks
-Bookmarks are stored in a database located in
-=~/.local/share/next/bookmark.db= by default. This (SQLite) database contains
-one table with two columns: id, url. In order to navigate and manage your
-bookmarks, a few functions are provided:
+   Next bookmarks are stored in a database file located in =~/.local/share/next/bookmark.db=
+   by default.
 
-1. ~C-m k~: Delete Bookmark
-2. ~C-m o~: Open Bookmark
-3. ~C-m s~: Bookmark Current Page
-4. ~C-m u~: Bookmark URL (input URL via minibuffer)
-5. ~C-m g~: Bookmark Anchor (input URL via link hints)
+   This, SQLite, database contains one table with two columns: id, url.
+
+   In order to navigate and manage your bookmarks, a few commands are provided:
+
+   | Command                                    | Emacs | Vi  |
+   |--------------------------------------------+-------+-----|
+   | Open Bookmark                              | ~C-m o~ | ~m o~ |
+   | Bookmark Current Page                      | ~C-m m~ | ~m m~ |
+   | Delete Bookmark                            | ~C-m k~ | ~m k~ |
+   | Bookmark URL (input URL via minibuffer)    | ~C-m u~ | ~m u~ |
+   | Bookmark Anchor (input URL via link hints) | ~C-m g~ | ~m a~ |
 
 ** Clearing the Echo Area
 In the area at the bottom of the screen where the minibuffer resides,
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -74,8 +74,8 @@
       (when selected-link
         (%bookmark-url selected-link)))))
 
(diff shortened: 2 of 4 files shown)


documents/MANUAL.org
```diff
@@ -219,16 +219,20 @@ forward command will simply visit the first child of the current node
...
-
-1. ~C-m k~: Delete Bookmark
-2. ~C-m o~: Open Bookmark
-3. ~C-m s~: Bookmark Current Page
-4. ~C-m u~: Bookmark URL (input URL via minibuffer)
-5. ~C-m g~: Bookmark Anchor (input URL via link hints)
+   Next bookmarks are stored in a database file located in =~/.local/share/next/bookmark.db=
+   by default.
+
+   This, SQLite, database contains one table with two columns: id, url.
```
Ambrevar: You can remove the commas around SQLite.

source/bookmark.lisp
```diff
@@ -74,8 +74,8 @@
       (when selected-link
         (%bookmark-url selected-link)))))
 
-(define-command set-url-from-bookmark ()
-  "Set the URL for the current buffer from a bookmark."
+(define-command bookmark-open-bookmark ()
```
Ambrevar: What about `bookmark-open` or  `bookmark-load`?

source/bookmark.lisp
```diff
@@ -74,8 +74,8 @@
...
+  "Open current highlighted bookmark in the current buffer."
```
Ambrevar: "current highlighted" is superfluous, "Open a bookmark in the current buffer." is enough I think.

source/document-mode.lisp
```diff
@@ -3,87 +3,91 @@
...
-          "?" 'remove-search-hints
-          "G" 'scroll-to-bottom
-          "g g" 'scroll-to-top
-          "C-f" 'scroll-page-down
-          "C-b" 'scroll-page-up
-          "SPACE" 'scroll-page-down
-          "s-SPACE" 'scroll-page-up)
-        (list :emacs emacs-map
-              :vi-normal vi-map))))
+  "Base mode for interacting with documents."
```
Ambrevar: The 4-space indentation is because `define-mode` is a macro and only its body argument is indented with 2 spaces.  Now it's true that for modes this is a little unfortunate, because then most mode declaration will have awfully wide indentations.
Can we fix this at the macro level?  @jmercouris @vindarel?

source/bookmark.lisp
alexanderbarbosa: Indeed, I forgot to improve that docstring!

source/bookmark.lisp
alexanderbarbosa: I thought so at first, but I had to go the verbose way! :D 

That's better!

alexanderbarbosa:
Oh, I thought so, Manual seems to be yet evolving.

As for bindings, yep, don't make sense.

C-h b is great indeed! 

Instead, I will try to implement those issues are filled in. :D

source/document-mode.lisp
jmercouris: It’s a consequence of slime, try making a macro called deffish and then make one called define-fish, you’ll see the difference. As per how to get the other parts to dedent to 2 spaces I haven’t figured that out yet, I believe Shinmera does have some library for modifying indentation rules, but I cannot remember the name.

PR #262, opened 2019-07-06
README.org: point Guix link to next package directly

azzamsa:
I think this is better, cause user can find directly the next package.


Ambrevar:
Problem is, this would have to be updated on every version update.
Besides, I think the package information is not so useful if the users already have access to the README.  What do you think?

Note: "guix.info" needs to be updated to "guix.gnu.org".

azzamsa:
Oh Sorry, I think the link will be static as like debian packages.

PR #269, opened 2019-07-18
port.lisp: lookup for two possible executables

vindarel:
look for either "next-gtk-webkit" or "next-pyqt-webengine" in the
current directory, ports/, or return "name".

we can now run ./next inside our pyqt dist/ directory.

It works©, but a review is welcome and this might render obsolete some code, such as `ports/gtk-webkit.lisp`.

### DIFF
--- a/source/port.lisp
+++ b/source/port.lisp
@@ -39,24 +39,31 @@ as a string.")
   (port-accessor port 'log-file (name port)))
 
 (defun derive-path-from-name (name)
-  "Find platform port binary with base path NAME.
-First NAME is looked up relative to current directory.
+  "Find a platform port binary.
+Find one of next-gtk-webkit and next-pyqt-webengine in the current directory.
 If not found, then it is looked in the ports/ subfolder corresponding the NAME (e.g. \"gtk-webkit\").
 If still not found, NAME is used as is and will be looked for in PATH.
 
 This is an acceptable value for the PATH slot of the PORT class."
-  (or
-   (when (probe-file name) (truename name))
-   (let* ((port-name (file-namestring name))
-          (port-prefix "next-")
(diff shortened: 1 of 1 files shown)


alexanderbarbosa:
at last.

That means, users wont need to set executyable port path?! if so, great!

alexanderbarbosa:
It correctly find ports executable, if Next is installed as global package.

It fails on Local builds.

I reckon, that is exactly how it should be!

jmercouris:
You're correct about it rendering `ports/gtk-webkit.lisp` obsolete, that was one of the ideas we'd discussed. The idea being that, nothing should be port specific within the Lisp code. All of those details should be handled by the platform port.

The second thing that we talked about was actually just renaming the executable to output as next-platform on both GTK and PyQtWebEngine so that we don't have any confusion. Few people will run both binaries simultaneously, and if they want to, they can always customize the Lisp core to allow for this.

vindarel:
> It correctly find ports executable, if Next is installed as global package.

thanks for testing.

> It fails on Local builds.

@alexanderbarbosa What do you actually mean by this ? For me it works when I run `./next` at the project root or on `ports/gtk-webkit` (if there is a `next-gtk-webkit` binary).

> The second thing that we talked about was actually just renaming the executable to output as next-platform on both GTK and PyQtWebEngine so that we don't have any confusion. Few people will run both binaries simultaneously,

That is actually a solution too. I'd go by keeping how it is now for the few of us that run both.

alexanderbarbosa:
@vindarel 

Cant say why, but symlink next binary to a $PATH folder it dont behave as if executing in project root folder.

jmercouris:
This branch has been covered by the merging of pyqt-webengine.

PR #270, opened 2019-07-25
PyQt Webengine

jmercouris:
There are a lot of Lisp specific changes I strongly disagree with. Particularly the adding of a test client and excessive logging, which makes logs quite difficult to follow. I've outlined my comments below.


source/base.lisp
```diff
@@ -124,8 +124,11 @@ PATH or set in you ~/.config/next/init.lisp, for instance:
                 (set-url-buffer url buffer))))
           (progn
             (log:error "Could not connect to platform port: ~a" (path (port interface)))
-            (kill-program (port interface))
-            (kill-interface interface)
+            (handler-case
```
jmercouris: We want the debugger to be invoked on failure, otherwise how could one debug?

source/remote.lisp
```diff
@@ -615,15 +624,19 @@ Deal with URL with the following rules:
...
-    (funcall (resource-query-function buffer)
-             buffer
-             :url url
-             :cookies cookies
-             :event-type event-type
-             :is-new-window is-new-window
-             :is-known-type is-known-type
-             :mouse-button mouse-button
-             :modifiers modifiers)))
+    (if buffer
```
jmercouris: Buffer will always exist. The platform port cannot make its own buffers without being instructed to do so by the Lisp core.

source/utility.lisp
```diff
@@ -227,3 +227,52 @@ Optional second argument FLAVOR controls the units and the display format:
...
+  "List the windows of the given interface. Dev purposes."
+  (maphash (lambda (key val)
+             (format t "~a ~a~&" key val))
+          (windows interface)))
+
+(defun list-buffers (interface)
+  (maphash (lambda (key val)
+             (format t "~a ~a~&" key val))
+           (buffers interface)))
+
```
jmercouris: All of the above new functions are not part of Next, no user will want to use them. That is why I suggest they do not belong in `utility.lisp` and they should not be compiled into the image. I think they could belong in a system called `next/debug`.

vindarel: Ok, I removed them. We'll see.

source/base.lisp
vindarel: This should happen only when we run the executables on the terminal, and it still shows the stacktrace.

jmercouris: I'm not sure what you mean by that, are you saying we don't want the debugger when the GUI fails?

vindarel: I mean: *as a user running Next*, I don't want the clunky SBCL debugger in the terminal when an error occurs. I want to see a stacktrace. A developer wants to have the debugger when developing with Slime.

jmercouris: Ah!!! I see now, OK, that makes a lot of sense!

jmercouris: I think in the future though, when the user is modifying their config file and the like, we'd probably want to show the debugger in the GUI window, like Emacs does, but that is a down the road thing.

source/remote.lisp
vindarel: Right. It happened to me during development, but it probably© won't happen otherwise.

jmercouris: I'm personally a huge fan of: It works, it compiled on my machine ™ :D

source/port.lisp
```diff
@@ -84,8 +91,10 @@ This is an acceptable value for the PATH slot of the PORT class."
...
-         (write-to-string
-          (uiop/launch-program:process-info-pid
-           (running-process port))))))
+  (if (uiop:process-alive-p (running-process port))
+      (uiop:run-program
+       (list "kill" "-15"
+             (write-to-string
+              (uiop/launch-program:process-info-pid
+               (running-process port)))))
+      (format *error-output* "process ~a is not running.~&" (running-process port))))
```
Ambrevar: Use `log:error` instead.

source/proxy-mode.lisp
```diff
@@ -5,6 +5,8 @@
 (define-mode proxy-mode ()
     "Enable forwarding of all network requests to a specific host.
 This can apply to specific buffer."
+    ;TODO: for the PyQt side, we now want the protocol, the IP and the
+    ;port on different slots.
```
Ambrevar: Use `;; ` command style.  Also I prefer when sentences start with a capital ("For" here) :)

source/minibuffer.lisp
```diff
@@ -431,7 +431,7 @@ interpreted by `format'. "
                         (:head (:style style))
                         (:body
                          (:p (apply #'format nil args)))))))
-        (log:warn "Can't echo ~a without minibuffer or interface" args))))
+        (log:warn "Can't echo '~a' without minibuffer or interface" (apply #'format nil args)))))
```
Ambrevar: What's up here?

jmercouris: Had to apply the arguments otherwise it would simply output the sexp, which isn't exactly useful. It is also done above. I thought about putting it into the parent LET, but decided to keep the change as small as possible.

Ambrevar:
But the s-exp displays everything, right?  This change is purely
cosmetic, isn't it?
Considering it was part of a `log:warn` which mostly targets developers,
I am actually more in favour of a less ambiguous s-exp :)

source/minibuffer.lisp
jmercouris: No, it shows just a variable or whatever about what it was trying to echo to the minibuffer rather than what would have literally been echoed in the minibuffer. I can post an example later, but it is a lot more informative this way, that’s why it is done that way in the LET above

source/port.lisp
vindarel: two comments: done, thanks.

PR #272, opened 2019-07-29
Vindarel/improve fuzzy match

vindarel:
Improved M-x fuzzy match:

- Major improvement: the order of input words is not important: `M-x "swit buf"` gives the same results as `"buf swit"`
- Candidates are sorted by "best matching" with the levenshtein distance
- Candidates that start with the first written word come first.

(edit) However: there's `mk-strings-metrics` to bundle for Guix. Maybe can we find a light replacement.

It looks ok, it doesn't break the previous behavior (`sbp` gives `switch-buffer-previous`).

See examples in the test file. Ok I put a little gif :)  (low quality sorry)

vokoscreen-2019-07-29_23-33-23

Possible improvements: better function names. Put this in a package of its own for further re-usability.

### DIFF
--- a/source/utility.lisp
+++ b/source/utility.lisp
@@ -2,7 +2,6 @@
 
 (in-package :next)
 
-(export '(fuzzy-match))
 
 (defmethod object-string ((object t))
   (princ-to-string object))
@@ -70,6 +69,18 @@ Otherwise, build a search query with the default search engine."
          (url (format nil search-url encoded-search-string)))
     url))
 
+(defun interleave-chars-with-regexp (input)
+  "Insert '.*' between all characters of `input'."
+  (check-type input string)
+  (with-output-to-string (stream)
+    (loop for char across input do
+         (princ #\. stream)
+         (princ #\* stream)
+         (princ char stream))
+    ;; match any chars after final char in cleaned-input
+    (princ #\. stream)
+    (princ #\* stream)))
+
 (defun all-permutations (list)
   "From a list of words, return a list of lists containing all the permutations.
 (foo bar) => ((foo bar) (bar foo))"
@@ -80,19 +91,23 @@ Otherwise, build a search query with the default search engine."
               append (mapcar (lambda (l) (cons element l))
                              (all-permutations (remove element list)))))))
 
-(defun join-permutations-to-re (permutations)
-  "`permutations': list of lists.
-((foo bar) (bar foo)) => (foo.*bar)|(bar.*foo)"
+(defun join-permutations-to-regexp (permutations)
+  "Join the `permutations' (list of lists of strings) into a 'OR' regexp.
+  For example: ((foo bar) (bar foo)) => (foo.*bar)|(bar.*foo)"
   (check-type permutations list)
   (str:join "|"
-            (loop for couple in permutations collect
-                 (format nil "(~a)" (str:join ".*" couple)))))
+            (loop for innerlist in permutations
+               for innerlist-with-regexp = (mapcar (lambda (word)
+                                                    (interleave-chars-with-regexp word))
+                                                  innerlist)
+               collect
+                 (str:join "" innerlist-with-regexp))))
 
-(defun build-input-re-permutations (input)
+(defun build-input-regexp-permutations (input)
   "Build a regexp of all words permutations from `input'.
 'foo bar' => '(foo.*bar)|(bar.*foo)'"
   (let* ((input-permutations (all-permutations (str:words input)))
-         (input-re (join-permutations-to-re input-permutations)))
+         (input-re (join-permutations-to-regexp input-permutations)))
     input-re))
 
 (defun candidate-match-p (input-re candidate)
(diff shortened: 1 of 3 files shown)


source/utility.lisp
```diff
@@ -2,6 +2,8 @@
 
 (in-package :next)
 
+(export '(fuzzy-match))
```
Ambrevar: I prefer to put the `export` next to the definition, this way we reduce the opportunity to run out of sync.  Maybe use cl-annot?

source/utility.lisp
```diff
@@ -68,42 +70,91 @@ Otherwise, build a search query with the default search engine."
...
+    ;; match any chars after final char in cleaned-input
+    (princ #\. stream)
+    (princ #\* stream)))
+
+(defun all-permutations (list)
+  "From a list of words, return a list of lists containing all the permutations.
+(foo bar) => ((foo bar) (bar foo))"
+  ;; thanks https://stackoverflow.com/questions/2087693/how-can-i-get-all-possible-permutations-of-a-list-with-common-lisp
+  (cond ((null list) nil)
+        ((null (cdr list)) (list list))
```
Ambrevar: `(list list)` can be a bit confusing.  Maybe rename the argument?

source/utility.lisp
```diff
@@ -68,42 +70,91 @@ Otherwise, build a search query with the default search engine."
...
+        (t (loop for element in list
+              append (mapcar (lambda (l) (cons element l))
+                             (all-permutations (remove element list)))))))
```
Ambrevar: You could use `match` to improve readability a bit.  No deal-breaker though :p

source/utility.lisp
```diff
@@ -68,42 +70,91 @@ Otherwise, build a search query with the default search engine."
...
+
+(defun join-permutations-to-re (permutations)
+  "`permutations': list of lists.
```
Ambrevar: I didn't understand the docstring at first :p  Please make it a sentence.

source/utility.lisp
```diff
@@ -68,42 +70,91 @@ Otherwise, build a search query with the default search engine."
...
+  (when (and input candidates)
+    (let (results)
+      (mapcar (lambda (candidate)
+                (when (candidate-match-p (build-input-re-permutations input) candidate)
+                  (push candidate results)))
+              candidates)
+      results)))
+
+(defun search-or-loose (word x)
+  "Search for substring `word' in x, or return a big number (and not nil). Used for sorting."
```
Ambrevar: I don't understand the docstring:
- I think it should be "`x'".
- "or return" -> shouldn't it be "and return"?
- Is it "loose" or "lose"?
- What is a big number?  Why are we doing this?
- Sorting what?

source/utility.lisp
```diff
@@ -68,42 +70,91 @@ Otherwise, build a search query with the default search engine."
...
+  "`permutations': list of lists.
+  ((foo bar) (bar foo)) => (foo.*bar)|(bar.*foo)"
+  (check-type permutations list)
+  (str:join "|"
+            (loop for innerlist in permutations
+               for innerlist-with-regex = (mapcar (lambda (word)
+                                                    (interleave-chars-with-regexp word))
+                                                  innerlist)
+               collect
+                 (str:join "" innerlist-with-regex))))
```
Ambrevar: Inconsistency between `re`, `regex` and `regexp`.  I'd favour the longest: `regexp`.

source/utility.lisp
```diff
@@ -68,42 +70,91 @@ Otherwise, build a search query with the default search engine."
...
+
+(defun match-permutation-candidates (input candidates)
+  "Return the list of candidates that contain all the words of the input."
+  (when (and input candidates)
+    (let (results)
+      (mapcar (lambda (candidate)
+                (when (candidate-match-p (build-input-re-permutations input) candidate)
+                  (push candidate results)))
+              candidates)
+      results)))
```
Ambrevar: Nit: Could make this a bit shorter with a `loop ... collect`.

vindarel: indeed.

source/utility.lisp
vindarel: done.

source/utility.lisp
vindarel: I reworked that.

source/utility.lisp
vindarel: ok

source/utility.lisp
vindarel: ok. I am currently not in favor of cl-annot, it imposes a user configuration for Slime integration. It doesn't work out of the box.

source/utility.lisp
jmercouris: It seems clear to me that it is applying the function LIST on list. I feel as if changing the name of the list argument would be confusing, because the name would become non descriptive or overly-descriptive (as in the case of naming it LIST-ARGUMENT). Just my two cents!

source/utility.lisp
jmercouris: Export next to the definition is good, I've looked into cl-annot, I'm not a huge fan of it honestly. Less magic, the better.

jmercouris:
I would also like to say a +1 on this work, it offers a nice incremental improvement to fuzzy search!

source/utility.lisp
vindarel: > My only concern is about the permutations, I'm not sure that regexps are the best way to deal with it, it could have some hidden costs (regexps usually do) that will shoot us in the foot at some point, e.g. when matching long strings like URLs and titles.

So in this PR we introduce permutations and we insert the `.*` "match-all" regexp in-between characters of the input string (to comply to the previous behavior, your first implementation). I think permutations are not costly because they deal with user input, that will usually be two to three words long ("swich buf cur"). But inserting the `.*` regexp can be costly indeed. If we don't do it, the user must insert spaces to have the "match all" effect. That's ok to me, that's how I'm used to input completion (helm-swoop and all). To match titles and URLs, we certainly can do without inserting this `.*` because we prefer to rely on a clever matching algorithm like the Levenhstein distance (which we use) to accept typos (match "wiki levensh" for "wiki levenhs").

We might be able to know when we do completion of titles and URLs so we can give a parameter to the fuzzy function to not do the `.*`.

Shall we remove `.*` in-between chars now or wait ?

jmercouris: I would vote to not remove it, the cost has not yet shown to be significant. Premature optimization and all that :D

vindarel:
@Ambrevar what about

> there's mk-strings-metrics to bundle for Guix. Maybe can we find a light replacement.

is it ok or do we search a lightweight replacement ?

Ambrevar:
`mk-string-metrics` has no dependency, it will take 5 minutes to package it for Guix.

jmercouris:
@Ambrevar I assume that closing this PR was unintentional?

Ambrevar:
Indeed, I don't remember closing it at all, really :/

PR #273, opened 2019-07-30
Fuzzy-match on display values and return real values.

Ambrevar:
@jmercouris @vindarel: This should fix the recent issue introduced with the new fuzzy-matching function.

This makes the distinction between real-value and display-value for candidates:
display-value is what's used for display _and_ for fuzzy matching, while the real-value is the data that's handled (commands, buffers, etc.).  This works exactly like Helm.

Note that this issue is not new: it's always been around, but the reason it used to work was that John had tailored the former fuzzy-match function to handle this gracefully, but then it could not be extended with multiple sorting functions the way we are doing it now.
Anyways, it should now be fixed.

On a related note, I've added a TODO in a previous commit regarding the display-value vs. object-string that is used in `generate-completion-html` to display the candidates.
At the moment, `object-string` and `accessor-function` can be different, which could result in weird behaviour (the fuzzy-matching may not match the display).
I suggest we remove the `accessor-function` and only implement the `object-string` method for the candidates.

Thoughts?

### DIFF
--- a/source/utility.lisp
+++ b/source/utility.lisp
@@ -113,12 +113,12 @@ Otherwise, build a search query with the default search engine."
 (defun candidate-match-p (input-re candidate)
   (ppcre:scan input-re candidate))
 
-(defun match-permutation-candidates (input candidates)
+(defun match-permutation-candidates (input candidate-pairs)
   "Return the list of candidates that contain all the words of the input."
-  (when (and input candidates)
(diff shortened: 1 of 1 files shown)


vindarel:
Thanks !

please add a docstring or comment about these two values.

Ambrevar:
Comment is already there.
I'll add more docstrings.

PR #280, opened 2019-07-31
new command: download-open-file

vindarel:
in addition, we should be able to click to open a file, and we should have an
option to automatically open it after download.

### DIFF
--- a/source/download-mode.lisp
+++ b/source/download-mode.lisp
@@ -46,7 +46,8 @@
                              ") "
                              (:u (quri:render-uri (download-manager:resolved-uri d)))
                              " as "
-                             (:b (file-namestring (download-manager:file d)))))))))
+                             (:b (file-namestring (download-manager:file d)))))))
+                    (:p (:em "Open a file with M-x download-open-file."))))
          (insert-content (ps:ps (setf (ps:@ document Body |innerHTML|)
                                       (ps:lisp contents)))))
     (rpc-buffer-evaluate-javascript interface download-buffer insert-content)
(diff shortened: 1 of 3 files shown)


source/download-mode.lisp
```diff
@@ -46,7 +46,8 @@
                              ") "
                              (:u (quri:render-uri (download-manager:resolved-uri d)))
                              " as "
-                             (:b (file-namestring (download-manager:file d)))))))))
+                             (:b (file-namestring (download-manager:file d)))))))
+                    (:p (:em "Open a file with M-x download-open-file."))))
```
Ambrevar: I'd put it as `(:code "M-x download-open-file") "."`.

source/download-mode.lisp
```diff
@@ -71,3 +72,22 @@
...
+
+(define-command download-open-file (root-mode &optional (interface *interface*))
+  "Open a downloaded file.
+Open the file with the system's default (xdg-open)."
+  ;; xxx: have a customizable association list of extensions -> program.
+  (with-result (filename (read-from-minibuffer
+                          (minibuffer *interface*)
+                          :input-prompt "Open file:"
+                          :completion-function (downloaded-files-completion-fn interface)))
+    (uiop:launch-program (list "xdg-open" filename))))
```
Ambrevar: Should catch a condition here in case xdg-open is missing.
And better: make this a customizable function that take the filename as argument, so that the user can customize what to do with what file.
For instance, if `filename` has the `.pdf` extension, fire up `zathura`, etc.

Ambrevar:
What the?  Why did it just merge by itself??!?

jmercouris:
Using GitHub as a remote has problems...

source/download-mode.lisp
vindarel: done

source/download-mode.lisp
vindarel: \+1. A function is a good start.

PR #285, opened 2019-08-01
switch-buffer: the current buffer shouldn't be the first of the list

### LINKED ISSUE #179: [C-x b] Switch to next buffer if there only two buffers open
Hey, I dont think that even Emacs has it:

If there are only two buffers open, C-x b would just switch to  the next buffer imeddiately...

That is it! :D

vindarel:
it is a little bit annoying.

fixes #179

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -21,7 +21,12 @@ See the `make-buffer' function for Lisp code."
   (make-buffer :name name :default-modes modes))
 
 (defun buffer-completion-fn ()
-  (let ((buffers (alexandria:hash-table-values (buffers *interface*))))
+  (let ((buffers (alexandria:hash-table-values (buffers *interface*)))
+        (active-buffer (active-buffer *interface*)))
+    ;; For commodity, the current buffer shouldn't be the first one on the list.
+    (when (equal (first buffers)
+                 active-buffer)
+      (setf buffers (alexandria:rotate buffers -1)))
     (lambda (input)
       (fuzzy-match input buffers))))
 


source/utility.lisp
```diff
@@ -287,3 +287,9 @@ Optional second argument FLAVOR controls the units and the display format:
                "B")
               ((eq flavor 'iec) "iB")
               (t "")))))
+
+(defun rotate-left (list)
+  "The first element of `list' becomes the last one."
+  (when list
+    (append (cdr list)
+            (cons (car list) '()))))
```
Ambrevar: Alexandria can already do this with `(alexandria:rotate SEQ -1)`.

source/buffer.lisp
```diff
@@ -21,7 +21,12 @@ See the `make-buffer' function for Lisp code."
   (make-buffer :name name :default-modes modes))
 
 (defun buffer-completion-fn ()
-  (let ((buffers (alexandria:hash-table-values (buffers *interface*))))
+  (let ((buffers (alexandria:hash-table-values (buffers *interface*)))
+        (active-buffer (active-buffer *interface*)))
+    ;; For commodity, the current buffer shouldn't be the first one on the list.
+    (when (equal (first buffers)
+                 active-buffer)
+      (setf buffers (rotate-left buffers)))
```
Ambrevar: I guess this works.  Another possibly more general way to do this would be to add an `access-time` slot and to sort by it in `fuzzy-match`.

source/utility.lisp
vindarel: \+1

source/buffer.lisp
vindarel: Yeah. Ok if I merge and I create an issue for this ?

Ambrevar: Yes

PR #286, opened 2019-08-01
[WIP] fix shell example

vindarel:
but still, on M-x shell:

```
   Method handler (NEXT::CORE-OBJECT
                   NEXT::PUSH-INPUT-EVENT) signaled an error: The value
                                                                NEXT::SHELL-MODE

                                                              is not of type
                                                                SEQUENCE.
; No value
```

@Ambrevar, @jmercouris ?


Ambrevar:
Look at the website git repo, it's already fixed there.  Completely forgot that we had an example here too, thanks for the catch!

Ambrevar:
Sorry, I forgot about this and merged it as 12ca0ad33c5c89b8e28b3b89f299c2d10b2c9370 instead.
Well, should be good now.

Issue #155, opened 2018-12-05
Sort and indent headings


phw:
The jump to heading feature is one of the coolest feature of this browser. However, the heading listing seems to be sorted rather randomly and has no indication of the structure of headings.

It would be great if headings could be sorted in the order they appear (or at least document order) and indicate their level e.g. by indenting sub headings

jmercouris:
That is a very good suggestion! I assume H1 at the top level, and then H2 indented further etc etc. I really like your idea! Thank you!

With regards to sort order, I have no idea honestly what it is based on, somehow JS decides to traverse the document, I am really not sure how.

I would also like to make it a little bit more intelligent and be able to type in keywords or something, so type "Jump to heading" and start typing phrases, like "fish, food, salmon" and it suggests to me headings that have paragraphs below it containing those words. A little bit of a smarter jump, do you know what I mean?

Lastly, I would like to implement a "jump across all buffers" which I think would be really cool, allowing you to jump anywhere within all of your open buffers, but that is quite a bit more complex!

Ambrevar:
That'd be awesome indeed!

vindarel:
About the sort order, where do you experience this ? It is ok for me here:

- https://next.atlas.engineer/quickstart
- https://lispcookbook.github.io/cl-cookbook/editor-support.html
- http://wikemacs.org/
- https://en.wikipedia.org/wiki/Lisp_(programming_language)

aartaka:
4df71688e7a49809353527e4404ef40edb5e2eb3 should finally get us sorted headings, given that our `document-model` has clear ordered idntifiers matching the structure of the DOM.

PR #287, opened 2019-08-02
jump-heading: clean headings, specially for Wikipedia

vindarel:
Wikipedia headings end with "[edit]", in french it's an even longer
"[modifier | modifier le code]". That's ugly.
We strip what comes after the first bracket.

for #155

We need a list of real candidates and a list of user-friendly names, hence the elegant use of a closure. I found no built-in mechanism for this in `read-from-minibuffer`…

### DIFF
--- a/source/jump-heading.lisp
+++ b/source/jump-heading.lisp
@@ -2,6 +2,25 @@
 
 (in-package :next)
 
+(defclass heading ()
+  ((inner-text :accessor inner-text
+               :initarg :inner-text
+               :type :string))
+  (:documentation "A heading. The inner-text must not be modified, so
+  than we can jump to the anchor of the same name."))
+
+(defmethod object-string ((heading heading))
+  "Cleaned-up text of this heading, to show the user in the minibuffer.
+For example, Wikipedia ones end with '[edit]'. We strip what comes after the first bracket."
+  (with-slots (inner-text) heading
+    (subseq inner-text 0 (position #\[ inner-text))))
+
+(defun make-headings (list/str)
+  "Make a list `heading's from a list of strings."
+  (mapcar (lambda (text)
+            (make-instance 'heading :inner-text text))
+          list/str))
+
 (define-parenscript get-headings ()
   (defun qsa (context selector)
     "Alias of document.querySelectorAll"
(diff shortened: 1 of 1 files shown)


source/jump-heading.lisp
```diff
@@ -20,15 +20,36 @@
...
-                           :input-prompt "Jump to heading:"
-                           :completion-function (lambda (input)
-                                                  (fuzzy-match
-                                                   input
-                                                   (cl-json:decode-json-from-string
-                                                    headings))))))
-    (paren-jump-to-heading :heading-inner-text heading)))
+(defun clean-heading (heading)
+  "Cleanup a heading.
+For example, wikipedia ones end with '[edit]'. We strip what comes after the first bracket."
```
Ambrevar: Wikipedia (capital).

source/jump-heading.lisp
```diff
@@ -20,15 +20,36 @@
...
+      (cleaned-headings nil))
+  ;; We need to preserve the list of real headings, the ones that are the names of links,
+  ;; but we want to present cleaner names to the user in the completion buffer.
+  (defun clean-headings (headings/json)
+    "Transform the json string to a list of headings, and clean them up."
+    (check-type headings/json string)
+    (let ((headings (cl-json:decode-json-from-string headings/json)))
+      (setf current-headings headings)
+      (setf cleaned-headings (mapcar #'clean-heading headings))
+      cleaned-headings))
```
Ambrevar: The current way we use to general display value from real values is with the `object-string` method.  So maybe it's easier to just create a `heading` class or type and implement the method.
Thoughts?

source/jump-heading.lisp
```diff
@@ -20,15 +20,36 @@
...
+
+  (defun heading-from-selection (selection)
+    "From the user selection, which is a heading cleaned up, get back the original heading string, so than we can jump to it."
```
Ambrevar: Line is too long.

source/jump-heading.lisp
```diff
@@ -20,15 +20,36 @@
...
+    (elt current-headings
+         (position selection cleaned-headings :test #'string=)))
+
+  (define-command jump-to-heading ()
+    "Jump to a particular heading, of type h1, h2, h3, h4, h5, or h6"
```
Ambrevar: Period missing.

vindarel:
It's much better indeed with object-string.

PR #288, opened 2019-08-02
Use single history database table so that clicked-on links also appea…

Ambrevar:
…r in minibuffer.

### DIFF
--- a/source/document-mode.lisp
+++ b/source/document-mode.lisp
@@ -152,7 +152,7 @@
         (return-from add-or-traverse-history t)))
     ;; if we made it this far, we must create a new node
     (when url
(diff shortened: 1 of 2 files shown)


vindarel:
\+1 LGTM

jmercouris:
There is not the data here differentiating from inputted URLs and typed URLs, do we not want to do that via a column?

Ambrevar:
In practice I find the current behaviour to be very usable.  If we are going to drop SQlite and use a "how it was added" slot, then I'd rather do that straight away.  (After 1.3.)

jmercouris:
OK, then!

PR #289, opened 2019-08-02
Print bindings next to commands name in minibuffer.

Ambrevar:
@vindarel: This is how I use `object-string` to change the display value in `fuzzy-match`.

### DIFF
--- a/source/keymap.lisp
+++ b/source/keymap.lisp
@@ -22,7 +22,7 @@ mode map, which are
   \"C-x\" \"C-s\"         - set to \"prefix\"
   \"C-x\"                 - set to \"prefix\"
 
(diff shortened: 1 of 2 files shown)


source/keymap.lisp
```diff
@@ -57,6 +57,21 @@ sequence of keys longer than one key-chord can be recorded."
   (mapcar (lambda (k) (serialize-key-chord k :normalize normalize))
           key-chord-stack))
 
+;; TODO: Ideally, we wouldn't need the serialized intermediary representation
+;; and we could remove thie function altogether.
```
jmercouris: typo "thie"

vindarel:
\+1 it works beautifuly !

PR #291, opened 2019-08-02
M-x open-file: open any file from the minibuffer with fuzzy completion

vindarel:
That's pretty basic but useful. 

TODOs:
- [X] open a file from the downloads directory
- [X] go one directory up
- [X] enter one directory
- [x] create a mode (don't clobber minibuffer keymap)
  - ~~move stuff to class slost~~: a PR is coming to put everything under a package namespace.
- [x] ensure we have good configuration options (only overriding a function for now, manual ok).

Much, much can and will be added. We'll have a great resource with Lem's directory mode (https://github.com/cxxxr/lem/blob/188a5881b9ee054ab26d057371ce8ce133b33136/lib/core/directory-mode.lisp).

### DIFF
--- a/documents/CHANGELOG.org
+++ b/documents/CHANGELOG.org
@@ -70,7 +70,7 @@ Press =Enter= to  visit a file, =M-Left= or =C-l=  to go one directory
 up, =M-Right= or =C-j= to browse the directory at point.
 
 By default, it uses the =xdg-open=  command. The user can override the
-=next:*open-file-fn*= variable  with a new function,  which takes the
+=next:*open-file-function*= variable  with a new function,  which takes the
 filename  (or directory  name) as  parameter.  See the  manual for  an
 example.
 
--- a/documents/CHANGELOG.org
+++ b/documents/CHANGELOG.org
@@ -57,6 +57,27 @@ MAJOR.MINOR.PATCH, we increment the:
   imported, say, from Firefox, or exported to a SQLite database or a custom text
   format.
 
+** TASK 1.3.2 (Upcoming)
+*** New open-file command (C-x C-f)
+
+This new command allows to open a file from the filesystem.
+
+The user is prompted with the minibuffer, files are browsable with the fuzzy completion.
+
+The default directory is the one from the download manager.
+
+Press =Enter= to  visit a file, =M-Left= or =C-l=  to go one directory
+up, =M-Right= or =C-j= to browse the directory at point.
+
+By default, it uses the =xdg-open=  command. The user can override the
+=next:*open-file-fn*= variable  with a new function,  which takes the
+filename  (or directory  name) as  parameter.  See the  manual for  an
+example.
+
+The default keybinding is =C-x C-f=.
+
+__Note__: this feature is alpha and is meant to grow in Next 1.4 and beyond!
+
 ** DONE 1.3.1 (Upcoming)
 *** Print page title in buffer list
 And the title is matched when fuzzy-searching a buffer!
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -472,10 +472,10 @@ The default keybindings for ~vi-normal-mode~ are:
 ** Decide how to open files
 
 The commands  =open-file= and  =download-open-file= call  the function
-=open-file-fn <filename>=. Its  default behaviour is to  open the file
+=open-file-function <filename>=. Its  default behaviour is to  open the file
 with the  system's default, using  =xdg-open=.  You can  override this
 behaviour   by    binding   another    function   to    the   variable
-=next:*open-file-fn*=.
+=next:*open-file-function*=.
 
 For example:
 
@@ -487,10 +487,10 @@ For example:
                     ((member extension '("flv" "mkv" "mp4") :test #'string=)
                      (uiop:run-program (list "mpv" filename)))
                     (t
-                     (next::open-file-fn-default filename))))
-    (error (c) (log:error "Error opening pdf ~a: ~a" filename c))))
+                     (next::open-file-function-default filename))))
+    (error (c) (log:error "Error opening ~a: ~a" filename c))))
 
-(setf next:*open-file-fn* #'my-open-videos)
+(setf next:*open-file-function* #'my-open-videos)
 #+end_src
 
 ** Loading Files
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -16,6 +16,7 @@ end-user allowing for infinite customization.
   - [[#history][History]]
   - [[#bookmarks][Bookmarks]]
   - [[#download-manager][Download manager]]
+  - [[#opening-files][Opening files]]
   - [[#clearing-the-echo-area][Clearing the Echo Area]]
   - [[#exiting][Exiting]]
   - [[#proxy-and-tor-support][Proxy and Tor support]]
@@ -266,6 +267,19 @@ as usual, and also with =M-x download-list=.
 To open  a file, use  =M-x download-open-file=. See  the customization
 section to control how files are open.
 
+** Opening files
+
+With =M-x open-file= (bound to =C-x  C-f=), you are prompted a list of
+files, and you can select one with the usual fuzzy completion. You can
+go one directory  up with =M-Left= or =C-l=, and  enter the directory
+at point with =M-Right= or =C-j=.
+
+Next  will open  the  file  with the  system's  default program  using
+=xdg-open=.  See  the  command  help  for  further  details,  and  the
+customization section to override the default behavior.
+
+__Note__: this feature is alpha and is meant to grow in Next 1.4 and onwards.
+
 ** Clearing the Echo Area
 In the area at the bottom of the screen where the minibuffer resides,
 Next will occasionally display messages. These can be dismissed by
@@ -457,22 +471,26 @@ The default keybindings for ~vi-normal-mode~ are:
 
 ** Decide how to open files
 
-The command =M-x download-open-file=, which is  used to open a file we
-downloaded  in  the  current  session,  calls  in  turn  the  function
-=open-file <filename>=. Its default behaviour is to open the file with
-the  system's  default,  using   =xdg-open=.  You  can  override  this
-behaviour by re-defining this function.
+The commands  =open-file= and  =download-open-file= call  the function
+=open-file-fn <filename>=. Its  default behaviour is to  open the file
+with the  system's default, using  =xdg-open=.  You can  override this
+behaviour   by    binding   another    function   to    the   variable
+=next:*open-file-fn*=.
 
 For example:
 
 #+begin_src lisp
-(defun open-file (filename)
-  "Open PDFs with Emacs."
+(defun my-open-videos (filename)
+  "Open videos with `mpv'."
   (handler-case (let ((extension (pathname-type filename)))
-                  (str:string-case extension
-                                   ("pdf" (uiop:run-program (list "emacs" filename)))
-                                   (otherwise (next::open-file-default filename))))
-    (error (c) (format *error-output* "Error opening pdf ~a: ~a" filename c))))
+                  (cond
+                    ((member extension '("flv" "mkv" "mp4") :test #'string=)
+                     (uiop:run-program (list "mpv" filename)))
+                    (t
+                     (next::open-file-fn-default filename))))
+    (error (c) (log:error "Error opening pdf ~a: ~a" filename c))))
+
+(setf next:*open-file-fn* #'my-open-videos)
 #+end_src
 
 ** Loading Files
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -481,14 +481,14 @@ For example:
 
 #+begin_src lisp
 (defun my-open-videos (filename)
-  "Open videos with `mpv'."
+  "Open videos with mpv."
   (handler-case (let ((extension (pathname-type filename)))
-                  (cond
-                    ((member extension '("flv" "mkv" "mp4") :test #'string=)
-                     (uiop:run-program (list "mpv" filename)))
-                    (t
+                  (match extension
+                    ((or "flv" "mkv" "mp4")
+                     (uiop:launch-program (list "mpv" filename)))
+                    (_
                      (next::open-file-function-default filename))))
-    (error (c) (log:error "Error opening ~a: ~a" filename c))))
+    (error (c) (log:error "Error opening pdf ~a: ~a" filename c))))
 
 (setf next:*open-file-function* #'my-open-videos)
 #+end_src
--- a/next.asd
+++ b/next.asd
@@ -63,6 +63,7 @@
                  (:file "proxy-mode")
                  (:file "noscript-mode")
                  (:file "download-mode")
+                 (:file "file-manager-mode")
                  ;; About
                  (:file "about")
                  ;; Port Compatibility Layers
--- a/source/file-manager-mode.lisp
+++ b/source/file-manager-mode.lisp
@@ -41,17 +41,18 @@
 ;; "open-file-function" into a download-mode slot, which is then called from
 ;; `download-open-file' with `(funcall (open-file-function download-mode)
 ;; filename).
-(defun open-file-fn (filename)
-  "Open `filename'.
-`filename' is the full path of the file (or directory), as a string.
-By default, try to open it with the system's default external program, using `xdg-open'.
-The user can override this function to decide what to do with the file."
-  (open-file-fn-default filename))
+;; (defun open-file-fn (filename)
+;;   "Open `filename'.
+;; `filename' is the full path of the file (or directory), as a string.
+;; By default, try to open it with the system's default external program, using `xdg-open'.
+;; The user can override this function to decide what to do with the file."
+;;   (open-file-fn-default filename))
 
 ;; note: put under the function definition.
-(export '*open-file-fn*)   ;; the user is encouraged to override this in her init file.
-(defparameter *open-file-fn* #'open-file-fn
-  "Function triggered to open files.")
+;; (export '*open-file-fn*)
+;; the user is encouraged to override this in her init file.
+;; (defparameter *open-file-fn* #'open-file-fn
+;;   "Function triggered to open files.")
 
 (defun open-file-from-directory-completion-fn (input &optional (directory *current-directory*))
   "Fuzzy-match files and directories from `*current-directory*'."
@@ -59,7 +60,19 @@ The user can override this function to decide what to do with the file."
         (dirnames (uiop:subdirectories directory)))
     (fuzzy-match input (append filenames dirnames))))
 
-(define-command display-parent-directory (minibuffer-mode &optional (minibuffer (minibuffer *interface*)))
+;TODO: define a new mode and specialize keybinings on it !
+
+(define-mode open-file-mode (minibuffer-mode)
+    "doc"
+    ())
+
+(defclass open-file-instance ()
+  ((default-modes :initform '(open-file-mode))
+   (open-file-function :accessor open-file-function
+                       :initform #'open-file-fn-default))
+  (:documentation "Open a file interactively."))
+
+(define-command display-parent-directory (open-file-mode &optional (minibuffer (minibuffer *interface*)))
   "Get the parent directory and update the minibuffer.
 
 Default keybindings: `M-Left' and `C-l'."
@@ -67,7 +80,7 @@ Default keybindings: `M-Left' and `C-l'."
   (erase-input minibuffer)
   (update-display minibuffer))
 
-(define-command enter-directory (minibuffer-mode &optional (minibuffer (minibuffer *interface*)))
+(define-command enter-directory (open-file-mode &optional (minibuffer (minibuffer *interface*)))
   "If the candidate at point is a directory, refresh the minibuffer candidates with its list of files.
 
 Default keybindings: `M-Right' and `C-j'. "
@@ -78,7 +91,7 @@ Default keybindings: `M-Right' and `C-j'. "
       (erase-input minibuffer)
       (update-display minibuffer))))
 
-(define-command open-file (root-mode &optional (interface *interface*))
+(define-command open-file ()
   "Open a file from the filesystem.
 
 The user is prompted with the minibuffer, files are browsable with the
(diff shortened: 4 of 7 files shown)


source/file-manager-mode.lisp
```diff
@@ -0,0 +1,41 @@
...
+;;     ())
+
+(defun open-file-fn-default (filename)
+  "Open this file with `xdg-open'."
+  (handler-case (uiop:run-program (list "xdg-open" (namestring filename)))
+    ;; We can probably signal something and display a notification.
+    (error (c) (format *error-output* "Error opening ~a: ~a~&" filename c))))
+
+;; TODO: Remove `open-file-fn` (it's just a one-liner) and instead store the
+;; "open-file-function" into a download-mode slot, which is then called from
```
Ambrevar: Then it would be in the file-manager-mode slot when implemented! :)

Ambrevar:
Ready for merging?

vindarel:
No, not yet. We need to access parent directories. I have local changes, but I couldn't update the minibuffer with the new listing yet. I'll show you.

vindarel:
I commited, see comments. We can do `M-x open-file`, we can use `C-l` to go one directory up, but we have to re-run the command to see the new files list.

vindarel:
What should be the good keybinding ? `C-o` is load-file, `C-f` is forward-history. `C-x C-f` is very emacsy, we might want a more usual one?

Ambrevar:
C-x C-f seems good, "opening a file" is not a very browserish task so I
don't think many people will complain :)

vindarel:
Ok. It's ready for tests.

documents/CHANGELOG.org
```diff
@@ -59,6 +59,29 @@ MAJOR.MINOR.PATCH, we increment the:
 
 ** TASK 1.3.1 (Upcoming)
 
+*** New M-x open-file command (C-x C-f)
```
Ambrevar: M-x and command are a bit reduncant.  I would write "New `open-file` command (`C-x C-f`)".

documents/CHANGELOG.org
```diff
@@ -59,6 +59,29 @@ MAJOR.MINOR.PATCH, we increment the:
...
+The user is prompted with the minibuffer, files are browsable with the fuzzy completion.
+
+The default directory is the one from the download manager.
+
+Press =Enter= to visit a file, =C-l=  to go one directory up, =C-j= to
+browse the directory at point.
+
+By default, it uses the =xdg-open=  command. The user can override the
+=next:open-file-fn= function,  which takes the filename  (or directory
+name) as parameter. See the manual for an example.
```
Ambrevar: I would still not recommend function overrides, but setting a variable instead.

documents/MANUAL.org
```diff
@@ -457,21 +457,22 @@ The default keybindings for ~vi-normal-mode~ are:
...
+  "Open videos with `mpv'."
   (handler-case (let ((extension (pathname-type filename)))
-                  (str:string-case extension
-                                   ("pdf" (uiop:run-program (list "emacs" filename)))
-                                   (otherwise (next::open-file-default filename))))
+                  (cond
+                    ((member extension '("flv" "mkv" "mp4") :test #'string=)
+                     (uiop:run-program (list "mpv" filename)))
+                    (t
+                     (next::open-file-fn-default filename))))
```
Ambrevar: A good opportunity to use `match` here :)

next.asd
```diff
@@ -10,6 +10,7 @@
   :depends-on (:alexandria
                :bordeaux-threads
                :cl-css
+               :cl-fad
```
Ambrevar: I'm not too familiar with cl-fad.  What are the benefits over uiop and standard CL?

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,97 @@
...
+;;;
+;;; "file manager" is a big excessive for now. Currently, we can:
+;;; - browse files, with fuzzy-completion
+;;; - go one directory up (C-l)
+;;; - enter a directory (C-j)
+;;; - open files. By default, with xdg-open. See `open-file-fn'.
+;;;
+;;; ***********************************************************************
+;;; *disclaimer*: this feature is meant to grow with Next 1.4 and onwards !
+;;; ***********************************************************************
```
Ambrevar: Punctuation: No space before `!` and capital at the beginning.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,97 @@
...
+(in-package :next)
+
+(defvar *current-directory* download-manager::*default-download-directory*
+  "Default directory to open files from. Defaults to the downloads directory.")
+
+(defun open-file-fn-default (filename)
+  "Open this file with `xdg-open'."
+  (handler-case (uiop:run-program (list "xdg-open" (namestring filename)))
+    ;; We can probably signal something and display a notification.
+    (error (c) (format *error-output* "Error opening ~a: ~a~&" filename c))))
```
Ambrevar: Use `log:error` instead?

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,97 @@
...
+;;; - open files in Next
+;;; - a UI to list files
+;;; - lazy loading for large directories
+;;; - many things...
+;;;
+
+(in-package :next)
+
+(defvar *current-directory* download-manager::*default-download-directory*
+  "Default directory to open files from. Defaults to the downloads directory.")
```
Ambrevar: I'm not sure a global for the current directory is the right approach:
- We already have (uiop:getcwd).
- A global means we can't nest calls to functions that use it.

What about calling `(setf (completion-function minibuffer) (lambda (input) (open-file-from-directory-completion-fn input NEW-DIR-HERE)))`

Otherwise, uiop also has `with-current-directory`, which you can use as follows:
```
(uiop:with-current-directory ("/home/ambrevar/dotfiles/") 
					 (uiop:directory-files "."))
```
Not sure how it can be useful here, but it's always good to keep it in mind :)

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,97 @@
...
+    ;; We can probably signal something and display a notification.
+    (error (c) (format *error-output* "Error opening ~a: ~a~&" filename c))))
+
+;; TODO: Remove `open-file-fn` (it's just a one-liner) and instead store the
+;; "open-file-function" into a download-mode slot, which is then called from
+;; `download-open-file' with `(funcall (open-file-function download-mode)
+;; filename).
+(export 'open-file)  ;; the user is encouraged to override this in her init file.
+(defun open-file-fn (filename)
+  "Open this file. `filename' is the full path of the file (or directory), as a string.
```
Ambrevar: Open FILENAME.
FILENAME is the full path...

Ambrevar: By the way, instead of specifying the type in the docstring, I suggest we start using more `(declaim (ftype...))` :)

Moreoever, instead of support just strings for pathname, we could support `(or string pathname)`, and wrap `filename` in `(namestring filename)`.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,97 @@
...
+
+The default directory is the one from `next/download-manager::*default-download-directory*' (which is `~/Downloads' by default).
+
+Press `Enter' to visit a file, `C-l' to go one directory up, `C-j' to browse the directory at point.
+
+By default, it uses the `xdg-open' command. The user can override the `next:open-file-fn' function, which takes the filename (or directory name) as parameter.
+
+The default keybinding is `C-x C-f'.
+
+Note: this feature is alpha, get in touch for more !"
```
Ambrevar: Some `M-q` to fold the lines?

Also I would remove some of those empty lines, I think it's harder to read with so much spacing.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,97 @@
...
+               (cl-fad:directory-exists-p filename))
+      (setf *current-directory* filename)
+      (update-display minibuffer))))
+
+(define-command open-file (root-mode &optional (interface *interface*))
+  "Open a file from the filesystem.
+
+The user is prompted with the minibuffer, files are browsable with the fuzzy completion.
+
+The default directory is the one from `next/download-manager::*default-download-directory*' (which is `~/Downloads' by default).
```
Ambrevar: Actually no, the default download directory is dynamically computer with `default-download-directory`, and only falls back to `*default-download-directory*`.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,97 @@
...
+                            :input-prompt (file-namestring directory)
+                            :completion-function #'open-file-from-directory-completion-fn))
+
+      (open-file-fn (namestring filename)))))
+
+
+(define-key  "C-x C-f" #'open-file)
+
+(define-key :mode 'minibuffer-mode  "C-l" #'display-parent-directory)
+(define-key :mode 'minibuffer-mode  "C-j" #'enter-directory)
```
Ambrevar: I know this is the default for Helm, but are those good bindings?  I find them rather counter-intuitive.
We could at least at arrows, or `C-right` / `C-left`, or maybe `C-ret` / `C-backspage`.  Thoughts?

documents/MANUAL.org
```diff
@@ -460,20 +474,23 @@ The default keybindings for ~vi-normal-mode~ are:
 The commands  =open-file= and  =download-open-file= call  the function
 =open-file-fn <filename>=. Its  default behaviour is to  open the file
 with the  system's default, using  =xdg-open=.  You can  override this
-behaviour by re-defining this function.
+behaviour   by    binding   another    function   to    the   variable
+=next:*open-file-fn*=.
```
Ambrevar: Are you using some kind of justification?  The spacing is strange, and I see that the text is aligned to a column.
I think we should all use the same justification for consistency (Emacs' default here I think).

vindarel:
Note: we can use new binaries to try new features. For this one: https://gitlab.com/atlas-engineer/next/-/jobs/284443816 (download the artifact).

vindarel:
@ambrevar: I pushed a last change, it seems ok to me. I left the open-file function as a parameter, not as a class slot, because a PR that puts everything under packages is coming.

One review not adressed: the global var for the current directory, tips on uiop:with-cwd etc.

Ambrevar:
Regarding the global, actually maybe a simpler solution would be to store the current directory as a class slot!

documents/CHANGELOG.org
```diff
@@ -57,6 +57,27 @@ MAJOR.MINOR.PATCH, we increment the:
   imported, say, from Firefox, or exported to a SQLite database or a custom text
   format.
 
+** TASK 1.3.2 (Upcoming)
+*** New M-x open-file command (C-x C-f)
```
Ambrevar: Still redundant "M-x" :)

documents/CHANGELOG.org
```diff
@@ -57,6 +57,27 @@ MAJOR.MINOR.PATCH, we increment the:
...
+
+By default, it uses the =xdg-open=  command. The user can override the
+=next:*open-file-fn*= variable  with a new function,  which takes the
+filename  (or directory  name) as  parameter.  See the  manual for  an
+example.
+
+The default keybinding is =C-x C-f=.
+
+__Note__: this feature is alpha and is meant to grow in Next 1.4 and beyond!
+
```
Ambrevar: Isn't this a bit too long for a changelog entry?  What about shrinking it to just a few lines and linking to the manual instead?

documents/CHANGELOG.org
```diff
@@ -65,6 +86,7 @@ And the title is matched when fuzzy-searching a buffer!
 This can be configured by setting =(open-external-link-in-new-window-p
 *interface*)= to the desired boolean.
 *** Return to last created buffer when deleting buffer
+*** Buffer management improvements
```
Ambrevar: ?

documents/MANUAL.org
```diff
@@ -266,6 +267,19 @@ as usual, and also with =M-x download-list=.
...
+** Opening files
+
+With =M-x open-file= (bound to =C-x  C-f=), you are prompted a list of
+files, and you can select one with the usual fuzzy completion. You can
+go one directory  up with =M-Left= or =C-l=, and  enter the directory
+at point with =M-Right= or =C-j=.
+
+Next  will open  the  file  with the  system's  default program  using
+=xdg-open=.  See  the  command  help  for  further  details,  and  the
+customization section to override the default behavior.
```
Ambrevar: You are using some sort of text justification here.  What about using Emacs' default for consistency?

documents/MANUAL.org
Ambrevar: If you don't want to use match now, could you add a "TODO"?

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
+;;; file-manager-mode.lisp -- Manage files.
+;;;
+;;; Open any file from within Next, with the usual fuzzy completion.
+;;;
+;;; `M-x open-file (C-x C-f)'
+;;;
+;;; "file manager" is a big excessive for now. Currently, we can:
```
Ambrevar: Typo: "bit"

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+;;; - configuration options to choose what to open with
+;;; - more configuration in general
+;;; - sort by last access, etc
+;;; - multi-selection
+;;; - bookmarks
+;;; - open files in Next
+;;; - a UI to list files
+;;; - lazy loading for large directories
+;;; - many things...
+;;;
```
Ambrevar: Spurious line.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+
+(in-package :next)
+
+;; XXX: a global isn't satisfactory.
+;; - store it as a class slot ?
+;; see also: uiop:with-current-directory, uiop:getcwd
```
Ambrevar: We use "TODO" in the code base, I suggest we stick to it for consistency.

Ambrevar: You didn't explain why it's not satisfactory.  You can use my argument below if you like.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+  (handler-case (uiop:run-program (list "xdg-open" (namestring filename)))
+    ;; We can probably signal something and display a notification.
+    (error (c) (log:error "Error opening ~a: ~a~&" filename c))))
+
+(defun open-file-fn (filename)
+  "Open `filename'.
+`filename' is the full path of the file (or directory), as a string.
+By default, try to open it with the system's default external program, using `xdg-open'.
+The user can override this function to decide what to do with the file."
+  (open-file-fn-default filename))
```
Ambrevar: Previously I had noticed that this function is just a one-liner.  Do we need it?
What about removing it and simply call `open-file-fn-default` at the caller sites?

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+
+;; note: put under the function definition.
+(export '*open-file-fn*)
+;; the user is encouraged to override this in her init file.
+(defparameter *open-file-fn* #'open-file-fn
+  "Function triggered to open files.")
```
Ambrevar: Shouldn't this be #'open-file-fn-default?

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+  "Fuzzy-match files and directories from `*current-directory*'."
+  (let ((filenames (uiop:directory-files directory))
+        (dirnames (uiop:subdirectories directory)))
+    (fuzzy-match input (append filenames dirnames))))
+
+;; (defclass open-file-instance ()
+;;   ((default-modes :initform '(open-file-mode))
+;;    (open-file-function :accessor open-file-function
+;;                        :initform #'open-file-fn-default))
+;;   (:documentation "Open a file interactively."))
```
Ambrevar: What is this?

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+(define-command display-parent-directory (open-file-mode &optional (minibuffer (minibuffer *interface*)))
+  "Get the parent directory and update the minibuffer.
+
+Default keybindings: `M-Left' and `C-l'."
+  (setf *current-directory* (uiop:pathname-parent-directory-pathname *current-directory*))
+  (erase-input minibuffer)
+  (update-display minibuffer))
+
+(define-command enter-directory (open-file-mode &optional (minibuffer (minibuffer *interface*)))
+  "If the candidate at point is a directory, refresh the minibuffer candidates with its list of files.
```
Ambrevar: Justification.  Also ideally the first sentence should tell what the function does  and should fit on one line.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+Default keybindings: `M-Right' and `C-j'. "
+  (let ((filename (get-candidate minibuffer)))
+    (when (and (uiop:directory-pathname-p filename)
+               (uiop:directory-exists-p filename))
+      (setf *current-directory* filename)
+      (erase-input minibuffer)
+      (update-display minibuffer))))
+
+(defun clean-up-open-file-mode ()
+  "Remove open-file-mode from the minibuffer modes, so that we don't clutter the usual minibuffers with our keybindings."
```
Ambrevar: Use quotes around `open-file-mode'.
Split in 2 sentences and move the 2nd sentence to a new line.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+  (setf (modes (minibuffer *interface*))
+        (last (modes (minibuffer *interface*)))))
```
Ambrevar: This only works if the mode is in first position.
You can be more reliable with, for instance, `(setf (modes (minibuffer *interface*)) (delete-if (lambda (m) ...) (modes (minibuffer *interface*)))`.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+
+(define-command open-file ()
+  "Open a file from the filesystem.
+
+The user is prompted with the minibuffer, files are browsable with the
+fuzzy completion.
```
Ambrevar: "with fuzzy completion"

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+
+The default directory is the one computed by
+`download-manager:default-download-directory' (certainly `~/Downloads').
```
Ambrevar: "certainly" -> "usually"

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+
+The default keybinding is `C-x C-f'.
+
+Note: this feature is alpha, get in touch for more !"
+  (let ((directory *current-directory*)
+        mode)
+    ;; "Activate" open-file-mode
+    (setf mode (make-instance  'open-file-mode))
+    ;; populate needed slots...
+    (setf (buffer mode) (minibuffer *interface*))
```
Ambrevar: You can use the `:buffer` key parameter in `make-instance`.

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+    ;; Allow the current minibuffer to recognize our keybindings.
+    (push mode (modes (minibuffer *interface*)))
+    (with-result (filename (read-from-minibuffer
+                            (minibuffer *interface*)
+                            :input-prompt (file-namestring directory)
+                            :completion-function #'open-file-from-directory-completion-fn
+                            :cleanup-function #'clean-up-open-file-mode))
+
```
Ambrevar: Spurious new line?

source/file-manager-mode.lisp
```diff
@@ -0,0 +1,153 @@
...
+      (funcall *open-file-fn* (namestring filename)))))
+
+
+(define-key  "C-x C-f" #'open-file)
+
+;; This currently dosen't work, we must use the keymap at the mode definition:
+;; (define-key :mode 'open-file-mode  "M-Left" #'display-parent-directory)
```
Ambrevar: Spurious comment?

Ambrevar:
One last thing: In commit message f6c92dbd6867c179c01e01ff935517e4533d0906, you say that

```
default open function as parameter is ok  …

no need to put it under a class slot, there's a PR coming to put
everything under the package namespace.
```

I think there is a misunderstanding.  The package namespacing allows us to define variables, functions and commands within a mode namespace.  But when it comes to configuration options, those should ideally be stored in a mode slot because we (usually) don't want them to be globals.

This is by design: we don't want to reproduce the Emacs "everything is global" nightmare.
So every time something is configurable, we should think "what if I have two separate instances of those".  We don't want one instance to override the configuration of the other.

Does that make sense?  What do you think?

vindarel:
> "what if I have two separate instances of those"

yes, thanks for the clarification. An easy example is when we call open-file in two different windows. The current behaviour is acceptable (the first window changes the cwd, the second uses it), but should be configurable.

> match

not that I don't want, I welcome an example.

Ambrevar:
Match example:

``` lisp
(match extension
						 ((or "flv" "mkv" "mp4")
							(uiop:run-program (list "mpv" filename)))
						 (t
							(next::open-file-fn-default filename)))
```

vindarel:
mmh I have a compilation warning

```
; in: DEFUN MY-OPEN-VIDEOS
;     (NEXT::OPEN-FILE-FUNCTION-DEFAULT NEXT::FILENAME)
; ==>
;   NEXT::FILENAME
; 
; note: deleting unreachable code
; 
; compilation unit finished
```

Ambrevar:
Sorry, the fallthrough is not `t`, it's `_`.

source/file-manager-mode.lisp
Ambrevar: You didn't fix FILENAME.
The convention is to uppercase the argument names, and to use quotes (`') for other references.

Ambrevar:
> An easy example is when we call open-file in two different windows. The current behaviour is acceptable (the first window changes the cwd, the second uses it), but should be configurable.

@vindarel: Please add a TODO so that we remember to fix this.

source/file-manager-mode.lisp
Ambrevar: @vindarel: No thoughts on this?  I'd rather not have more functions than necessary.

PR #297, opened 2019-08-06
Add INVISIBLE-INPUT-P to READ-FROM-MINIBUFFER.

noogie13:
Want to be able to enter passwords from the minibuffer.

Add INVISIBLE-INPUT-P to minibuffer class, and read-from-minibuffer method. As well add generate-input-html-invisible for update-display to generate the input with *'s. 

INVISIBLE-INPUT-P is reset to nil every time the cleanup function is called (if is non-nil).

### DIFF
--- a/source/minibuffer.lisp
+++ b/source/minibuffer.lisp
@@ -66,6 +66,7 @@ This should not rely on the minibuffer's content.")
    (input-prompt :accessor input-prompt :initform "Input:")
    (input-buffer :accessor input-buffer :initform "")
    (input-buffer-cursor :accessor input-buffer-cursor :initform 0)
+   (invisible-input-p :accessor invisible-input-p :initform nil)
    (completions :accessor completions)
    (completion-cursor :accessor completion-cursor :initform 0)
    (minibuffer-style :accessor minibuffer-style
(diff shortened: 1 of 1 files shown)


source/minibuffer.lisp
```diff
@@ -409,7 +424,9 @@ This should not rely on the minibuffer's content.")
       ;; Don't add input-buffer to completions that don't accept arbitrary
       ;; inputs (i.e. empty-complete-immediate is nil).
       (push input-buffer completions))
-    (let ((input-text (generate-input-html input-buffer input-buffer-cursor))
+    (let ((input-text (if (invisible-input-p minibuffer)
+                           (generate-input-html-invisible input-buffer input-buffer-cursor)
+                           (generate-input-html input-buffer input-buffer-cursor)))
```
Ambrevar: The `if` block is not indented properly.  I've fixed it :)

Ambrevar:
Great addition, consider it merged!
And sorry for the delay :p

noogie13:

> Great addition, consider it merged!
> And sorry for the delay :p

Thanks! Good stuff. I'm trying to figure out how to manage the ring now.
The easy part of modifying the paste function to paste from the ring,
the hard part is making it work with the global copy paste. One idea was
to only partitially implement it, it will paste from the ring (if it
exists) and after 30 seconds, it will set to nil again.

This solution doesn't feel very complete, but it does prevent you from
pasting the password outside of Next.

What do you think?

Regards,
Sol

P.S. After this I'm pretty sure I'm ready to PR the entire password
management system!

PR #307, opened 2019-08-11
macos: fix makefile and Info.plist

seanfarley:
I'm not sure the `cp` command with `*` is correct; hopefully someone can correct
that.

----

#


Ambrevar:
@jmercouris: What do you think?

jmercouris:
It is good! Thank you! It is not the ideal makefile, but we'll get there :D

PR #321, opened 2019-08-17
Use OBJECT-STRING for current minibuffer candidate

goose121:
Using the call to FORMAT, the completion returned would be some sort
of unprintable-object representation (e.g. #<BUFFER {223f7}>) which
would not be accepted by the calling code. This change makes it return
the string associated with the object instead, as the caller expects.


Ambrevar:
Good catch, thanks!

PR #323, opened 2019-08-17
Overhaul use of command class.

Ambrevar:
The command class has additional slots:
- `bindings' that replaces the `%%command-key-bindings' hash table (easier and
  faster).
- `mode' for convenience.
- `access-time' so we can sort by commands by recency of the call in the minibuffer.

The `object-string' method now displays all bindings of a command instead of
just the latest.

`execute-command' used to be broken in how it found the mode associated to a
command.  This is now fixed.

Many redundant and obsolete functions have been removed (command-symbol,
all-available-commands, etc.)

Know issue:
When calling `define-key` over a keymap, the `:scheme` argument is used to
register the binding in the command object.  This is not very dynamic: if we
manually add a keymap instance to a mode, the command `bindings' slot won't be updated.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -2,6 +2,7 @@
 
 (in-package :next)
 
+;; TODO: Use standard `print-object' instead?
 (defmethod object-string ((buffer buffer))
   (name buffer))
 
--- a/source/command.lisp
+++ b/source/command.lisp
@@ -68,10 +68,10 @@ deprecated and by what in the docstring."
     symbols))
 
 (defun package-variables ()
-  (remove-if-not #'boundp (package-defined-symbols)))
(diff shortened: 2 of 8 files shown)


Ambrevar:
Fixes issue mentioned in #318 where only one binding would be listed in the minibuffer.

Ambrevar:
Update: I've fixed the above "known issue" by removing the `bindings` slot from the `command` class.

Now the main question that remains is: Do we need a `mode` slot?  I'd argue that we could have a `mode` method instead that simply fetches the first specializer of the command with closer-mop.

Ambrevar:
Answer: yes we need either a mode slot of a slot that holds a reference to the actual method, so that we can disambiguate commands with the same name for different modes.

Ambrevar:
Merged.

PR #330, opened 2019-08-21
Remove unneeded reset of INVISIBLE-INPUT-P when callback function cal…

noogie13:
…led.

Resetting INVISIBLE-INPUT-P is irrelevant because it defaults to NIL when READ-FROM-MINIBUFFER is called. Also, resetting it would break the functionality when trying to read an invisible and visible input in a `WITH-RESULT*` as follows:

```
(with-result* ((foo (read-from-minibuffer (minibuffer *interface*)))
               (invisible-foo (read-from-minibuffer (minibuffer *interface*) :invisible-input-p t)))
  (echo (format nil "FOO:~a INVISIBLE-FOO:~a" foo invisible-foo)))
```

### DIFF
--- a/source/minibuffer.lisp
+++ b/source/minibuffer.lisp
@@ -157,8 +157,7 @@ This should not rely on the minibuffer's content.")
         ;; if there's no completion function
         (return-immediate (first (modes minibuffer)) minibuffer))
     (when cleanup-function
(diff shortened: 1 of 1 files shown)


jmercouris:
Looks good to me!

PR #336, opened 2019-08-22
Print title in buffer list.

Ambrevar:
A rather trivial patch, but it shows one current annoyange with the asynchronous callbacks: it makes simple tasks like fetching the title non-trivial.
In particular, the return value of anything in `with-result` is a callback number and not the result of the body, which can be jarring to newcomers.

I propose to implement synchronous versions of `rpc-buffer-evaluate-javascript` and `rpc-minibuffer-evaluate-javascript`.
One approach would be to have the caller wait on a lock indexed by the callback-id and then have `buffer-javascript-call-back` unlock that lock and return the right value.

Thoughts?

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -4,7 +4,7 @@
 
 ;; TODO: Use standard `print-object' instead?
 (defmethod object-string ((buffer buffer))
-  (name buffer))
+  (format nil "~a  ~a" (name buffer) (title buffer)))
 
 (defun make-buffer (&key (name "default")
                          default-modes)
--- a/source/remote.lisp
+++ b/source/remote.lisp
@@ -33,6 +33,7 @@ keyword is not recognized.")))
 (defclass buffer ()
   ((id :accessor id :initarg :id)
    (name :accessor name :initarg :name)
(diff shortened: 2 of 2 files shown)


jmercouris:
Looks good to me! Synchronous calls may lock the Lisp program if the JS does not complete.

Ambrevar:
> Looks good to me! Synchronous calls may lock the Lisp program if the JS does not complete.

Yes, but we can always add a timer to the lock so that it does not hang
the Lisp.

jmercouris:
We would have to add in threading then, no?

Ambrevar:
We already have it.
We use lparallel for downloads, we use bordeaux-threads on startup, and
finally threads are a prerequisite for async communication with dbus (as
with all RPC systems in this context).

jmercouris:
Yes, I am aware, I just meant for these specific operations :D

Ambrevar:
It should be no problem at all, Common Lisp's gracious handling of
multiprocessing is one of the reasons we are using the language after
all :)

Ambrevar:
Merged in 873c754e116c41b5fbb7bca166d95fb8dbe71a75

PR #343, opened 2019-08-25
Add copy & paste ring.

noogie13:
Add a ring for the clipboard.

Keep all copied and pasted entries into the ring, for access by PASTE-FROM-RING.


alexanderbarbosa:
what about `c-y` for copying?

noogie13:
Well, it should replace all copy and pasting duties, so I figure most people would use C-c/C-v, but we could bind both C-c/C-v and C-w/C-y.

Ambrevar:
So for now we have Emacs and VI schemes.  We could (should?) add a CUA scheme.  In this sense, `C-w`/`C-y` would only be in the Emacs scheme and `C-c`/`C-v` in the CUA scheme.

Ambrevar:
Also please rebase, I think this the total changes are readable enough to fit in 1-2 commits.  Thanks!

source/ring.lisp
```diff
@@ -0,0 +1,59 @@
...
+(defmethod ring-index ((ring ring) index)
+  "Convert to internal ring index, where items are ordered from newest to oldest."
+  (mod (1- (+ (head-index ring) (- (item-count ring) index))) (ring-size ring)))
+
+(defmethod ring-insert ((ring ring) new-item)
+  "Insert item into RING.
+If RING is full, delete oldest item."
+  (setf (aref (items ring)
+              (mod (+ (head-index ring) (item-count ring))
+                   (ring-size ring))) new-item)
```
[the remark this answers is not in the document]
noogie13: Isn't the point of a ring to have a fixed size and overwrite the old entries when we get to the end of the length?

Example: the 1001st item that we copy should replace the oldest one?

If not, then why not just use a list & expand it when we get towards the end?

source/ring.lisp
```diff
@@ -0,0 +1,59 @@
...
+
+(defmethod ring-ref ((ring ring) index)
+  "Return from RING using conversion to internal index."
+  (aref (items ring) (ring-index ring index)))
+
+(defmethod ring-recent-list ((ring ring))
+  "Return list of items ordered by most recent."
+  (loop for index from 0 to (1- (ring-size ring))
+        for item = (ring-ref ring index)
+        if (not (null item)) collect item))
```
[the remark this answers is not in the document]
jmercouris: why not `(when item collect item)`?

source/ring.lisp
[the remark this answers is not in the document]
Ambrevar: A list has linear access time.  What's great with a ring is that it has constant access time.

In your current implementation, `head-index` is never updated, and `item-count` is constantly growing, which is not possible for a ring.
So if we keep head-index as the index of the oldest element and we add the 1001st item:
- it should replace array index 0
- item-count should stick to 1000
- head-index should be 1
Now the most recent element is at array index 0, the oldest at array index 1.
Makes sense?

source/ring.lisp
```diff
@@ -34,7 +52,7 @@ If RING is full, replace the oldest item."
   "Return list of items ordered by most recent."
   (loop for index from 0 below (ring-size ring)
         for item = (ring-ref ring index)
-        if (not (null item)) collect item))
+        unless (null item) collect item))
```
Ambrevar: As @jmercouris suggested, `when item collect item` is even simpler!

source/ring.lisp
```diff
@@ -6,25 +6,43 @@
...
+               :documentation "Index of oldest item.")))
 
 (defmethod ring-size ((ring ring))
   "Return the maximum number of elements it can contain."
   (length (items ring)))
 
+(defmethod +1-ring-index ((ring ring) index)
+  "Add one to index, with wraparound."
+  (let ((new-index (1+ index)))
+    (if (eq new-index (length (items ring)))
```
Ambrevar: I prefer `=` for numbers.

source/ring.lisp
```diff
@@ -6,25 +6,43 @@
...
+        0 new-index)))
+
+(defmethod -1-ring-index ((ring ring) index)
+  "Minus one from index, with wraparound."
+  (let ((new-index (1- index)))
+    (if (eq new-index -1)
+        (1- (length (items ring))) new-index)))
```
Ambrevar: I believe both functions can be replaced with `(mod (1+ index) (ring-size ring))` (respectively same with `1-`).  Or did you mean something else?
Also what happens if `index` is -1 or `(ring-size ring)`?

source/ring.lisp
```diff
@@ -6,25 +6,43 @@
...
   "Insert item into RING.
 If RING is full, replace the oldest item."
-  (setf (aref (items ring)
-              (mod (item-count ring)
-                   (ring-size ring)))
-        new-item)
-  (incf (item-count ring)))
+  (prog1 (setf (aref (items ring)
+               (mod (+ (head-index ring) (item-count ring))
+                    (ring-size ring)))
```
Ambrevar: Indentation missing on `(mod...)`

Ambrevar:
I'm wondering if we need a ring deletion function.
I believe that the right implementation when removing an element is too shift all newer elements one index down.

Ambrevar:
@noogie13: Any blocker?

Ambrevar:
Closing in favor of #349.

PR #346, opened 2019-08-29
Wip mode package hooks and more

Ambrevar:
Get ready for the biggy! :)

This needs a bit of testing, but it's mostly working at the moment.

Question:
- Should we move minibuffer-mode and help-mode to separate packages?  I'd think so.

Next steps:
- Update tests and doc.
- Add hooks!

### DIFF
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -599,3 +599,40 @@ default modes:
 #+begin_src lisp
 (add-to-default-list 'proxy-mode 'buffer 'default-modes)
 #+end_src
+
+** Hooks
+
+A hook is a variable that holds a list of functions.
+We say a hook is executed when all its functions are run one after the other,
+over its arguments (which are decided at the call site).
+
+Many hooks are executed at different points in Next.
+
+Hooks are exposed to the users so that they can customize the behaviour of
+specific actions in arbitrary ways.
+
+Hooks are added at different sites, among others:
+
+- Global hooks, such as ~load-hook~ hook and ~after-init-hook~.
+- Window / buffer related hooks.
+- Commands "before" and "after" hooks.
+- Modes "enable" and "disable" hooks.
+
+For instance, if you want to force =old.reddit.com= over =www.reddit.com=, you
+can set a hook like the following in you =~/.config/next/init.lisp=:
+
+#+begin_src lisp
+(defun old-reddit-hook (url)
+  (let* ((uri (quri:uri url)))
+    (if (search "www.reddit" (quri:uri-host uri))
+        (progn
+          (setf (quri:uri-host uri) "old.reddit.com")
+          (let ((new-url (quri:render-uri uri)))
+            (log:info "Switching to old Reddit: ~a" new-url)
+            new-url))
+        url)))
+(add-to-default-list #'old-reddit-hook 'buffer 'load-hook)
+#+end_src
+
+Some hooks like the above example expect a return value, so it's important to
+make sure we return ~url~ here.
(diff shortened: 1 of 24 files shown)


Ambrevar:
I've added hook support!  In terms of code, it's really not hard (I've leveraged cl-hooks).
In terms of design, it raises tons of questions:

- Where do we add hooks?  Before or after of both?

- I feel like tons of hooks are missing, in particular around networking, but I'm still fumbling here.  Time will tell I suppose.

- I'd like to type the hooks so that only functions with the right signature are accepted.

- I realized that it would be beneficial to control _how_ handlers are executed, that is to say, should they be `#'mapcar` ed over  the arguments, or `#'compose` d, etc.?  For instance, the `set-url` hook should certainly chain the handlers over the successive results of the change made to the URLs, otherwise it makes no sense.  For now I've implemented a hacky `run-composed-hook`, but I'd like to give more control to the user so they can configure the hook on how to execute the handlers.
I've reported this issue upstrea: https://github.com/scymtym/architecture.hooks/issues/2.

Ambrevar:
If you want to test, you can try this cute hook which force-switches the URL to old.reddit.com:

```lisp
(defun old-reddit-hook (url)
  (let* ((uri (quri:uri url)))
    (if (search "www.reddit" (quri:uri-host uri))
        (progn
          (setf (quri:uri-host uri) "old.reddit.com")
          (let ((new-url (quri:render-uri uri)))
            (log:info "Switching to old Reddit: ~a" new-url)
            new-url))
        url)))
(add-to-default-list #'old-reddit-hook 'buffer 'load-hook)
```

You can add it to ~/.config/next/init.lisp.

Ambrevar:
Other hook question: wouldn't it be nice to be able to list _all the hooks_?
But I'm not sure how to proceed.  With `cl-hooks` hooks don't really have a type, they are just list.  But cl-hooks stores some stuff internally, so maybe we can look there.

Ambrevar:
About the hook handler composition: it seems that it's a novel idea, Emacs does not support it and I don't think that Lem or Nomad do either.  Maybe an idea worth sharing!

vindarel:
LGTM, but needs tests and a minimal doc. I can write a little tutorial afterwards.

Also, let's merge my PR first :D

vindarel:
oh, and it's awesome :) (hooks could have been made in another PR, no?)

Ambrevar:
For doc and tests, I left them out on purpose until someone (you :p) reviewed it.

For future "big" efforts, I believe this is a more efficiant workflow: should the patch set be rejected (or modified significantly), the effort put in documentating and testing would go to waste.

I'll add doc and tests as soon as possible.

For the PR: I needed hooks to be based off the same branch, but I could have created another PR indeed, my bad! :p

Ambrevar:
What do you think of the following TODO:
```
   ;; TODO: Rename `load-hook' to `set-url-hook'?
```

Ambrevar:
There could be 2 hooks: one that is run whenever the buffer displays something, e.g. a REPL is refreshed; the other one could be set whenever we "set the URL", that is to say, we query the load of a network resource.

Ambrevar:
I've updated the docstrings and the manual.

Ambrevar:
Merged.

Ambrevar:
Commit f6ffa60a67d8221fe93e20a2bd137435d846af38.

Ambrevar:
@vindarel: You mentioned tests but I'm not sure what to test here.
The separate package mostly have to do with compilation, so the test here is "does it compile or not"?
For hooks, I believe the cl-hooks library upstream should to the tests.

Thoughts?

PR #356, opened 2019-09-07
Instantiate a new minibuffer on every call and add a status bar.

Ambrevar:
The status bar is (for now) still a minibuffer, albeit one that's only
instantiated once per window.

To allow nesting of minibuffers, we add a `active-minibuffers' stack to the
window.
When a minibuffer is popped, the previous minibuffer is restores thanks to the
HTML content which was saved in the `content' slot.


Ambrevar:
To do:

- Fix open-file mode.
- set-url is very slow.
- Instantiate read-from-minibuffer with a minibuffer object.

Ambrevar:
To clarify, for now the status-bar display in the same web view as the minibuffer.  We will need to update the platform port to support an additional web view.

vindarel:
What is broken with open-file-mode ? (anything specific to it?)

Ambrevar:
My changes break it, which is desired because I got rid of the global
minibuffer.  We can now instantiate a new minibuffer just for
file-manager-mode.
I'm on it.

Ambrevar:
Fixed!

Performance is also much much better!  Previously Next used to slow down significantly on set-url with a big history.  This is now fixed too.

Last thing I'd like to add to this pull request to make sure I'm doing things right: the minibuffer history.

@vindarel: Let me know what you think for now.  This pull request  is the foundation for many future changes, including the separate status bar, HiDPI support, s-exp in candidate display, minibuffer multi selection, and more file manager goodies!

Ambrevar:
We also need this before implementing completion timeout and auto-selection on unique entry.

Ambrevar:
This is ready for merging!

I've implemented minibuffer history, bound to `M-p` by default.
For now, I've added 3 histories:
- generic (the default)
- set-url
- search

Set-url has the specificity to always add the current URL to the input history, so that if you want to edit the current URL you can simply press `C-l M-p RET`.

This could be a good opportunity to tune the minibuffer of `open-file`, @vindarel ;)

source/minibuffer.lisp
```diff
@@ -443,18 +449,35 @@ The new webview HTML document it set as the MINIBUFFER's `content'."
...
                                (:span (subseq input-buffer-password (+ 1  cursor-index))))))))
 
-(defun generate-completion-html (completions cursor-index)
-  (cl-markup:markup (:ul (loop for i from 0 for completion in completions
-                               collect
-                               (cl-markup:markup
-                                (:li :class (when (equal i cursor-index) "selected")
-                                     :id (when (equal i cursor-index) "selected")
-                                     (object-string completion)))))))
+(defun generate-completion-html (completions cursor-index minibuffer)
```
jmercouris: Am I understanding this correctly in that it will simply redraw the completions as necessary? Does this mean that the user can no longer scroll within the mini buffer (using the mouse wheel)?

Ambrevar:
They can, we just need to map the scrolls buttons to scroll-next /
scroll-previous!

Good catch, I had forgotten about this!

Ambrevar:
@jmercouris : Fixed!

I'd like to merge this tomorrow morning if possible.  Let me know if that's alright :)

jmercouris:
no worries from my side, looks good

Ambrevar:
Merged in 84156bd5122d7aca5ccc5c60ffe7f01ab0aa1f88.

PR #363, opened 2019-09-09
Add COPY-TO-CLIPBOARD function, which copies to clipboard and ring.

noogie13:
Replace bare trivial-clipboard calls in COPY-URL and COPY-TITLE with COPY-TO-CLIPBOARD.


Ambrevar:
Thanks for this!

PR #365, opened 2019-09-10
Reopen tab

### LINKED ISSUE #253: Reopen closed tab
Hi,

Yet another feature would that of reopening just recently closed tab.

Thanks!

4t0m:
Please let me know if anything needs fixing.  I wasn't totally sure about where to define things or what names to use.

For some reason, when I change the variables on my machine, something is failing to update somewhere and I am getting errors in slime because undefined variables are being invoked.  I figure we can just change the name here when the code reaches an approvable state.

Fixes #253.


4t0m:
Hmm.  I'm currently working on bringing this up to date with master.  Probably best to ignore this for now.

Ambrevar:
Thanks so much for working on this, and welcome to the still very small circle of Next contributors! :)

I had a quick look and it's looking good!  I'll go for a proper review after you've rebased.

I just had an idea: you could add a command which prompts the user with the minibuffer which URL to restore.  Awesome, nah? :)

Also instead of storing just the URL, I recommend you store both the URL and the title so that it displays in the minibuffer just like for other buffers.

Ambrevar:
Also as mentioned in #253, "buffer history" could be confused with the history of visited URLs for a given buffer.  I suggest to use the term "recent buffers" instead, or something like that.

4t0m:
Thanks for the notes!  I'll change the variable shortly, and saving the title as well sounds like a good idea.  

I was also thinking of doing something like what you describe with the minibuffer.  In that case, should the buffers be removed from the ring (as they normally are)?

I don't think there's a neat way to remove things from the middle of the ring, but it's not as if there will be any performance issues as a result.

4t0m:
> Also instead of storing just the URL, I recommend you store both the URL and the title so that it displays in the minibuffer just like for other buffers.

Do you mean like how urls and titles are shown for the current buffer?  I don't see where the minibuffer displays both title and url during selection.

I'm looking into what I'll have to do to display/return something other than a url string.

Ambrevar:
Are you using 1.3.1 or master?  Then in `M-x switch-buffer` you should
see URL and titles.

To customize the display of candidates, you need to create a class, say
`dead-buffer`, and customize the object-string method.

For the ring deletion, I suggest the following:

- Create a ring-delete ((ring ring) index) procedure that removes
  element at index INDEX.
  If the INDEX is closer to HEAD-INDEX then the tail, shift all recent
  elements to the "right".  Watch for the wrap around.  If the INDEX is
  closer to the tail, shift all older elements to the "left".  This
  should be the fastest.

- The ring-pop procedure can then just call to ring-delete.

4t0m:
Do you mean shifting left/right by setting the value of the array at index for each element?  Would it be faster to create a new ring with the starting array of the concatenation of all elements before the deleted value and all elements after the deleted value?  Or set `:items` to such an array?

Another thing: How does one close the minibuffer?  Right now it's remaining open after reopening a buffer, and I can't see what I'm doing differently from `delete-buffer` (which seems to close the minibuffer after running).

And finally: should adding the recent-buffer to the ring be part of `rpc-delete` as it currently is, or is it better as a hook?

4t0m:
Okay it doesn't seem like concatenating is any better.  I'll do as you described!

4t0m:
Hmm, it seems kind of tedious to delete a buffer from the ring when there are duplicates, since I don't know how to determine the index.  Is there a way for me to get the minibuffer to return the index of the selected item (the index in the original list passed to `fuzzy-match`)?  Or maybe there's a better way.

Also, is this worth it? :P

libraries/ring/ring.lisp
```diff
@@ -55,3 +55,10 @@ Return NEW-ITEM."
 
 (defun make (&key (size 1000))
   (make-instance 'ring :items (make-array size :initial-element nil)))
+
+(defmethod pop-most-recent((ring ring))
+  "Return the most-recently-added item in RING, and remove it from the RING.
+TODO What if the ring is empty?"
+  (let ((most-recent-item (ref ring 0)))
+    (decf (item-count ring))
```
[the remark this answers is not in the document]
4t0m: I'm confused about this.  Can you explain?

Example:
head-index (oldest): 1, item-count: 3
[c a b]
ring-pop -> c

head-index: 1 , item-count: 2
[c a b]
ring-pop -> b

head-index: 1 , item-count: 1
[c a b]
ring-pop -> a

[c a b]
head-index: 1 , item-count: 0
ring-pop -> nil
^ either check (plusp item-count), or set elements to nil when they are popped from the ring.

Really sorry if I'm missing something obvious.

(Also, my sincere apologies for all of the sloppy errors you had to catch for me.  I'll be sure to check those things locally before putting code up here in the future.  I'm also sorry & embarrassed by how slow I've been with this.  I'll have more time starting in a couple of days and I'd love to tackle another issue once this is finished up.)

source/remote.lisp
```diff
@@ -288,6 +292,20 @@ The handlers take the URL as argument.")
                         :documentation "Hook run after a download has completed.
 The handlers take the `download-manager:download' class instance as argument.")))
 
+(defmethod initialize-instance :after
+  "Creates a recent-buffers ring with the length set in recent-buffers-length."
+  ((interface remote-interface) &rest args)
+  (setf (recent-buffers interface)
+        (ring:make :size (recent-buffers-length interface))))
+        ;; (make-instance 'ring :items (make-array (recent-buffers-length interface)
+                                                ;; :initial-element nil))))
```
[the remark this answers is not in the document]
Ambrevar: Also see https://github.com/atlas-engineer/next/issues/381

libraries/ring/ring.lisp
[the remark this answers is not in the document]
Ambrevar: My bad, when I reviewed this I had forgotten that head was the oldest item, I thought it was the newest.
Don't spend to much time on this, I have plans to reuse cl-containers instead: https://github.com/gwkkwg/cl-containers/issues/10

And no need to apologize for anything, quite the opposite, your contribution is very much appreciated!  There is no time constraint or quality requirements, every little bit helps!

source/minibuffer.lisp
```diff
@@ -603,6 +603,7 @@ Return most recent entry in RING."
       (ring:insert ring clipboard-content)))
   (string (ring:ref ring 0)))
 
+
```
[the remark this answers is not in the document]
4t0m: Should be fixed.

source/recent-buffer.lisp
```diff
@@ -0,0 +1,37 @@
+;;; recent-buffer.lisp --- lisp subroutines for creating and managing recent buffers
```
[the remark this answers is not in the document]
4t0m: Fixed.

source/remote.lisp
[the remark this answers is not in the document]
4t0m: Fixed.

source/remote.lisp
```diff
@@ -243,6 +243,10 @@ by Next when the user session dbus instance is not available.")
    (minibuffer-generic-history :accessor minibuffer-generic-history :initform (ring:make))
    (minibuffer-search-history :accessor minibuffer-search-history :initform (ring:make))
    (minibuffer-set-url-history :accessor minibuffer-set-url-history :initform (ring:make))
+   (recent-buffers-length :accessor recent-buffers-length :initform 50
+                          :documentation "The maximum length of the recent-buffers.")
+   (recent-buffers :accessor recent-buffers
```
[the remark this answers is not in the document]
4t0m: I think fixed.

source/remote.lisp
```diff
@@ -288,6 +292,20 @@ The handlers take the URL as argument.")
...
 
+(defmethod initialize-instance :after
+  "Creates a recent-buffers ring with the length set in recent-buffers-length."
+  ((interface remote-interface) &rest args)
+  (setf (recent-buffers interface)
+        (ring:make :size (recent-buffers-length interface))))
+        ;; (make-instance 'ring :items (make-array (recent-buffers-length interface)
+                                                ;; :initial-element nil))))
+
+(defmethod add-buffer-to-recent-buffers-ring ((interface remote-interface) (buffer buffer))
```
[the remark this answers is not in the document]
4t0m: Fixed.

source/remote.lisp
```diff
@@ -288,6 +292,20 @@ The handlers take the URL as argument.")
...
+  "Add the url of a given buffer to the buffers ring."
```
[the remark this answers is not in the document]
4t0m: Fixed.

source/recent-buffer.lisp
```diff
@@ -0,0 +1,37 @@
...
+(in-package :next)
+(annot:enable-annot-syntax)
+
+@export
+@export-accessors
+(defclass recent-buffer()
+  ((name :accessor name :initarg :name
+        :initform nil)
+   (title :accessor title :initarg :title
+          :initform nil)))
```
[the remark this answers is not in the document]
4t0m: I think fixed.

4t0m:
Okay, I think it's working and ~complete.  

One thing I wasn't sure how to handle was echo-ing to the minibuffer.  The following doesn't seem to be working for me.

    (define-command undo-buffer-deletion ()
      "Open a new buffer with the URL of the most recently deleted buffer."

      (if (plusp (ring:item-count (recent-buffers *interface*)))
          (make-buffers (list (name (ring:pop-most-recent (recent-buffers *interface*)))))
          (echo "There are no recently-deleted buffers.")))

Also, should we add some keybindings? 
Maybe 'u' and 'U' for VI-MODE? (And something else for normal mode?)

next.asd
```diff
@@ -50,6 +50,7 @@
                  (:file "window")
                  (:file "minibuffer")
                  (:file "keymap")
+                 (:file "recent-buffer")
```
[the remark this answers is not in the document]
4t0m: Fixed.

libraries/ring/ring.lisp
```diff
@@ -17,6 +17,7 @@
 
 (defmethod index ((ring ring) index)
   "Return index converted to internal ring index, where items are ordered from newest to oldest."
+  ;; TODO: shouldn't that read "oldest to newest", since index 0 is the oldest?
```
4t0m: To clarify, index 0 in the internal array is the oldest item, so this line is confusing to me.

Ambrevar: The `index` method returns a "virtual" index.  The internal index 0 is _not_ the oldest item is there has been a wraparound.

For instance, in a ring of size 3,
- in [a b]: `a` at index 0 is indeed the oldest element.
- in [a b c], you add `d`, then you get [d b c]: there `d` at index 0 is the newest element.

libraries/ring/ring.lisp
```diff
@@ -29,7 +30,7 @@ Return NEW-ITEM."
                      (mod (+ (head-index ring) (item-count ring))
                           (size ring)))
                new-item)
-    (if (= (item-count ring) (length (items ring)))
+    (if (= (item-count ring) (size ring))
```
Ambrevar: Good catch! ;)

libraries/ring/ring.lisp
```diff
@@ -55,3 +56,40 @@ Return NEW-ITEM."
...
+  "Return the most-recently-added item in RING, and remove it from the RING."
+    (delete-index ring (index ring 0)))
+
+(defmethod shift-left ((ring ring) start end)
+  (loop for index from start below end
+        do (setf (aref (items ring)
+                       (mod index (size ring)))
+                 (aref (items ring)
+                       (mod (1+ index) (size ring)))))
+  (setf (aref (items ring) (mod end (size ring))) nil))
```
Ambrevar: Could use something like `(replace (items ring) (items ring) :start2 1)`.

libraries/ring/ring.lisp
```diff
@@ -55,3 +56,40 @@ Return NEW-ITEM."
...
+
+(defmethod shift-right ((ring ring) start end)
+  (loop for index from start above end
+        do (setf (aref (items ring)
+                       (mod index (size ring)))
+                 (aref (items ring)
+                       (mod (1- index) (size ring)))))
+  (setf (aref (items ring) (mod end (size ring))) nil))
```
Ambrevar: Same as above, with `:start1`.

libraries/ring/ring.lisp
```diff
@@ -55,3 +56,40 @@ Return NEW-ITEM."
...
+
+(defmethod delete-index ((ring ring) index)
+  "Delete the item at (internal oldest-first) index and move items to fill the space."
```
Ambrevar: Hmmm, I think it's confusing that this is the only function where `index` is the internal index.  Why not using the ring virtual index instead?

Ambrevar: No big deal though, we might merge the ring library into cl-containers later.

Ambrevar:
Good job!  It's mostly ready for merging, the only nit is if you could try using the `replace` function instead, that could make the code significantly shorter.

Regarding the issue with `echo`: Don't worry about it, this is a known issue, it will be solved when we implement a status buffer and a *Messages* buffer.  Stay tuned ;)

Ambrevar:
Let me know when you grow tired of working on this issue, I can complete the finishing touches if need be, no problem! :)

libraries/ring/ring.lisp
4t0m: Ah got it.

libraries/ring/ring.lisp
4t0m: How about this current version?


    (defmethod external-index ((ring ring) internal-index)
      (1- (+ (- (head-index ring)
                 internal-index)
             (item-count ring))))

    (defmethod delete-index ((ring ring) index)
      "Delete the item at external index and move items to fill the space."
      (let* ((internal-index (index ring index))
             (deleted-item (aref (items ring) (mod internal-index (size ring))))
             (head (head-index ring))
             (midpoint (mod (floor (+ head (item-count ring)) 2)
                            (size ring))))
        (if (< midpoint internal-index)
            (shift-left ring internal-index (+ head (item-count ring)))
            (progn
              (shift-right ring internal-index head)
              (setf (head-index ring) (mod (1+ (head-index ring)) (size ring)))))
        (decf (item-count ring))
        deleted-item))

    (defmethod delete-match ((ring ring) match-test)
      (let ((match-index (position-if match-test (items ring))))
        (when match-index (delete-index ring (external-index ring match-index)))))

libraries/ring/ring.lisp
4t0m: I'm having sort of a hard time visualizing how this handles wrap arounds.  Would you need to replace once from start to end, once from 0 to start, and then separately set element that wraps around?

4t0m:
Please feel free to make any edits/finishing touches.  This is my first time using common lisp so I've been enjoying learning how it all works, but I am also eager to look at other issues.

libraries/ring/ring.lisp
Ambrevar: Hmm... I'm afraid `external-index` might add extra confusion to the library.  I think we should try to keep it to just one "index" method and keep the "internal index" within that method only.
No problem, I'll merge and tweak the code until I find a nice and simple way to put this together.

libraries/ring/ring.lisp
Ambrevar: Good point, this might not be worth the complexity of having to deal with the internal index.
I'll think about it.

libraries/ring/ring.lisp
4t0m: Another idea: Maybe we should run `position-if' on the output of ring:recent-list, and then delete the index that returns?

Ambrevar:
That's an option :)

libraries/ring/ring.lisp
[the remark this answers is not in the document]
vindarel: > how slow I've been with this

same feeling everytime in a new project (or a new line of code).

But: you're doing it! kuddos.

Ambrevar:
I've rebased and merged in befea72c0f6053ec4f7227306deff3bc84822986.
Thanks for this great and very useful contribution!

A small tip to ease the burden or merging: when catching up with master, dont "merge", but "rebase against master" instead.  This way all your commits will be packed together instead of being spread around merge branches.

Also I noticed that you'd used different authors.  I've picked one which does not seem to be recognized by GitHub, sorry :p 

Last but not least?  Would you like to be included in the list of contributors in `M-x about`?

4t0m:
Awesome :).  I'll rebase in the future and hopefully figure out what's going on with the authorship thing.  

I'd love to be included on the list of contributors.

Ambrevar:
Done in commit 
1df811d0927af3e9c7906e20f952592215fc561b.

PR #367, opened 2019-09-10
Moved ring to libraries, fixed testing for download-manager and ring …

noogie13:
…systems.

You can run tests like this 
```
;;; Load SB-COVER
(require :sb-cover)

;;; Turn on generation of code coverage instrumentation in the compiler
(declaim (optimize sb-cover:store-coverage-data))

;;; Load some code, ensuring that it's recompiled with the new optimization
;;; policy.
(asdf:oos 'asdf:load-op :next/download-manager :force t)

;;; Run the test suite.
(prove:run :next/download-manager/tests)

(sb-cover:report "coverage/")

(declaim (optimize (sb-cover:store-coverage-data 0)))
```

Basically fixing systems.


noogie13:
Sorry, that's how you can generate coverage reports, you can still run them with `(prove:run :next/ring/tests)` or with `(asdf:test-system :next/ring/tests)` also.

Ambrevar:
Looking good!  Now that we have a separate package for rings, what about removing the "ring-" prefix in the defuns/defvars?

This way, from Next we would call `ring:insert`, `ring:make`, etc.

Ambrevar:
I've just merged the minibuffer overhaul branch, so there might be a few more conflicts when you rebase here, sorry about that :p  Not in particular that I've added a `copy` function to the ring library.

libraries/ring/ring.lisp
```diff
@@ -16,62 +11,52 @@
...
   "Return a copy of RING."
   (let ((copy (make-instance 'ring :items nil)))
     (setf
      (head-index copy) (head-index ring)
      (item-count copy) (item-count ring)
      (items copy) (copy-seq (items ring)))
     copy))
+
+(defun make-ring ()
+  (make-instance 'ring))
```
Ambrevar: I would add a `size` key parameter to set the array size.

libraries/ring/tests/tests.lisp
```diff
@@ -0,0 +1,36 @@
+(in-package :cl-user)
+
+(prove:plan nil)
+
+(defvar test-ring1 (ring:make-ring))
+(setf test-ring1 (ring:make-ring))
```
Ambrevar: I think you forgot to remove the "ring-" prefix in this file.

libraries/ring/ring.lisp
Ambrevar: Also shall we name it just `make`?

noogie13:
Ok. Added &key size, changed name to `(make)`. Else I think ready.

Also, I've noticed the open-file command doesn't seem to work, is there something I changed that's interfering or is it slightly broken?

Ambrevar:
Thanks for the notification, I'll fix open-file.
Good work, merging now!

Issue #79, opened 2018-02-17
Search engine dynamic suggestions (alias "webjumps")


Ambrevar:
I think it's important to discuss the bookmark format at an early stage.

1. A binary format is probably not a good idea.  It's not versionable and you are going to lose a huge user base who likes to keep their bookmarks in a git repo.
A few years back, another power-browser transited from text format to SQLite, it went defunct in the few following months... (I think it was Luakit.)  Also see https://github.com/qutebrowser/qutebrowser/issues/882#issuecomment-136807011.

2. What do you think about using Emacs `eww`'s format?  That would allow the two to share the same bookmarks transparently, a big bonus for Emacs users I think.

3. Qutebrowser is currently discussing the merge of bookmarks, quickmarks and search engines.  I think it's a neat idea.  Basically every bookmark can be recalled with a simple keybind (if specified) or used as a search engine (again, if specified).
See:
https://github.com/qutebrowser/qutebrowser/issues/882.

jmercouris:
The area of bookmarks is a huge area of contention for me. Currently the bookmarks are already in a sqlite database as is the history stored in `~/.local/share/next/bookmark.db` and `~/.local/share/next/history.db` respectively. The schema is very simple, just a single column for URL.

The reason I would like to stick with sqlite is because I am going to add some "smart" word matching, text analysis, and document tagging capabilities in the future (as that is one of my specialties as a computer scientist). 

In terms of importing from other browsers and formats, something to do that would be greatly appreciated. If you are interested on working on a parser for EWW bookmarks in Common Lisp, or some mechanism to add them to, that would be great!

In regards to bullet 3: I have no idea what quick marks are, I looked through the thread and I wasn't able to really figure it out. If you could provide a brief explanation, that would be very useful.

Finally, it is currently really easy to combine multiple sources to do something like search history, bookmarks, current tabs etc. So implementing these kinds of "go-to-anything" functions should be relatively trivial :)

Ambrevar:
I understand the need for a database, and Qutebrowser does that too.
The thing is, you don't need to _store_ the bookmarks as a SQL database, you can simply dump them as plain text on save, and create the database on startup.

This is a must, and I insist: _bookmarks_ significantly lose their value if their are not human readable / versionable.
(That would personally deter me from using Next Browser altogether.)

Qutebrowser has opted for storing the _history_ as a SQL binary though, because nobody was really interested in versioning it.
[...]

PR #371, opened 2019-09-12
WIP: Persistence (session, history, bookmarks)

Ambrevar:
Draft for now.

I can restore URLs on startup.  Hooray! :)

I've extracted the startup precedure to make it configurable by the user.
The session "store" and "restore" functions can be customized.

Before proceding, I need to clean up parts of the code:

- buffer-set-url is used in places, like following link hints, which bypasses the `set-url` function.  I believe there should be only one entry point for URL change.  I suggest renaming buffer-set-url to `%set-url` and using it only from `set-url`.
- We could add an option to `set-url` to not call `parse-url`, e.g. `:raw-url-p`.
- I want to delete new-window and new-buffer.
- I want to merge `make-buffers` into `make-buffer`.

Question:

- Should the session persistence functions be per window or per interface?  Considering a session persistence might save information about windows, maybe it would make more sense to store it in the remote-interface class.  But what about "private windows" then?

### DIFF
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -34,6 +34,7 @@ end-user allowing for infinite customization.
   - [[#styles][Styles]]
   - [[#proxy][Proxy]]
   - [[#hooks][Hooks]]
+  - [[#startup-behaviour][Startup behaviour]]
 
 * Basics
 ** Navigation
@@ -655,3 +656,18 @@ can set a hook like the following in you =~/.config/next/init.lisp=:
 
 Some hooks like the above example expect a return value, so it's important to
 make sure we return ~url~ here.
+
+** Startup behaviour
+
+Once the platform port has been started, the default action of Next is to run
+
+#+begin_src lisp
+(funcall (startup-function *interface*) (or urls *free-args*))
+#+end_src
+
+~startup-function~ defaults to ~default-startup~ and takes URLs that are passed
+to Next as command line arguments.
+
+You can assign you own function to ~startup-function~ to change the behaviour of
+Next on startup, such as which URL it should display, if it should restore the
+previous session or not, etc.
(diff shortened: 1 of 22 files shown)


Ambrevar:
I've just pushed some changes to serialize the complete `buffer` data structure.
cl-prevalence does an amazing job here, short of 2 edge cases:
- pathnames
- functions

I've fixed the pathname serialization.

For functions, well it's not easy to serialize a function :p  For now, I'm serializing named functions to `#'my-function-name`, and unnamed functions to `#'ignoring`, where `ignoring` is a function that takes any argument and does nothing.  (Is there already such a function in the standard?)
It's not very pretty, but it allows us to serialize many class slots which have a function value (e.g. `resource-query-function`). 

I'd like to report those ideas upstream, see if we can get them merged.

The great news is that I'm now able to restore buffers with their history and all.  There are still some rough edges.
For instance, shall we re-run the mode constructors when restoring the buffer?  The constructor could set some mode slots, so we'd rather not.  But on the other side, the constructor could also make some RPC calls (e.g. enable the proxy) which are required for the mode to be on at all.

My take is that constructors should _not_ change the state of Next beyond the RPC calls.
I'm not sure how we can enforce that though.

Thoughts?

Ambrevar:
I gave this a little more thoughts, and I'm realizing one thing: session persistence is a hard problem and I don't know if anyone ever got it right.  Emacs `desktop-mode` does not work so well in my experience (so many edge cases) and I prefer to rely on `recentf` to quickly re-access previously opened buffers.

As for restoring the right modes for the right URL, I just opened the issue #376.  So this could simply implemented orthogonally.

To sum up, the only thing we would have to serialize and restore are the buffer _histories_.  Which is a simple tree of URLs, so nothing hard here.

Thoughts?

Ambrevar:
TODO: Update doc (search engines have been moved).

Ambrevar:
Ready to merge, after 1.3.2.

Ambrevar:
After the merge, we need to tackle #376, #377 and #378.
There is also #79 that could be dealt with while we are at it.

Ambrevar:
And #380.

Ambrevar:
Merged with 1eae4c8fe0fa12e462cb35ae4498c1d5e377a30e.

PR #386, opened 2019-09-19
GitHub cloner

vindarel:
M-x git-clone

### DIFF
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -20,6 +20,7 @@ end-user allowing for infinite customization.
   - [[#clearing-the-echo-area][Clearing the Echo Area]]
   - [[#exiting][Exiting]]
   - [[#proxy-and-tor-support][Proxy and Tor support]]
+  - [[#cloning-git-repositories][Cloning Git repositories]]
 - [[#advanced-topics][Advanced Topics]]
   - [[#execute-extended-command][Execute Extended Command]]
   - [[#help][Help]]
@@ -314,6 +315,22 @@ and other information within your cache.
  At the time of writing, there  are differences between the Gtk and the
  Qt port: the Gtk one sets  proxies per-buffer, whereas it is currently
  global for the Qt one.
+
+** Cloning Git repositories
+
+Use the  =git-clone= command to  clone a  Git repository to  disk. The
+clone command is run asynchronously.
+
+By default, the git-clone command looks into the following directories
+for existing projects:
+
+: "~/projects" "~/src" "~/work" "~/common-lisp" "~/quicklisp/local-projects"
+
+You can change the list like this:
+
+#+begin_src lisp
+(setf next/vcs:*vcs-projects-roots* '("~/my/directory"))
+#+end_src
 
 
 * Advanced Topics
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -318,11 +318,12 @@ and other information within your cache.
 
 ** Cloning Git repositories
 
-Use the  =git-clone= command to  clone a  Git repository to  disk. The
-clone command is run asynchronously.
+Use  the  =vcs-clone=  (alias  =git-clone=) command  to  clone  a  Git
+repository to  disk.  It asks  you for  the destination and  then runs
+asynchronously.
 
-By default, the git-clone command looks into the following directories
-for existing projects:
+By  default, the  command  looks into  the  following directories  for
+existing projects:
 
 : "~/projects" "~/src" "~/work" "~/common-lisp" "~/quicklisp/local-projects"
 
@@ -331,6 +332,8 @@ You can change the list like this:
 #+begin_src lisp
 (setf next/vcs:*vcs-projects-roots* '("~/my/directory"))
 #+end_src
+
+When there is one single choice, it doesn't ask for confirmation.
 
 
 * Advanced Topics
--- a/source/vcs-mode.lisp
+++ b/source/vcs-mode.lisp
@@ -65,16 +65,13 @@ If EXIT-RECURSIVE-SCAN is non-nil, avoid recursive scan of local projects. By de
       (setf result (find-project-directory name :exit-recursive-scan t)))
     result))
 
-;XXX: add ftype declaration.
+(declaim (ftype (function (string)) ensure-directory))
 (defun ensure-directory (base)
   "Create this directory if it doesn't exist.
-Needed to call `truename' to exand a tilde after it.
-BASE: directory name (string)."
+Needed to call `truename' to exand a tilde after it."
   ;; ensure a trailing slash.
-  (setf base
-        (str:concat (string-right-trim (list #\/) base)
-                    "/"))
-  (ensure-directories-exist base))
+  (ensure-directories-exist (str:concat (string-right-trim (list #\/) base)
+                                        "/")))
 
 (defun concat-filenames (base dir)
   "Concat filenames. Expand a tilde in BASE.
--- a/source/vcs-mode.lisp
+++ b/source/vcs-mode.lisp
@@ -42,19 +42,27 @@ changes, browse files in a text editor, use hooks...
   "Scan `*vcs-projects-roots*'."
   (mapcan #'search-git-directories *vcs-projects-roots*))
 
-(defun find-project-directory (name &key exit)
-  "Return the directory pathname of the project named NAME.
-If EXIT is non-nil and the project was not found, don't parse the project roots again."
+(defun find-project-directory (name &key exit-recursive-scan)
+  "Return the directory pathname of the project named NAME by searching the local projects.
+If EXIT-RECURSIVE-SCAN is non-nil, avoid recursive scan of local projects. By default, it will re-scan the local projects to avoid false negatives and false positives."
   (unless *git-projects*
     (setf *git-projects* (parse-projects)))
   (let ((result (find name *git-projects*
                       :key (lambda (dir)
                               (alexandria:last-elt (str:split "/" (namestring dir) :omit-nulls t)))
                       :test #'string=)))
+    ;; If null, re-parse the local projects.
+    ;; It could have been cloned manually.
     (unless (or result
-                exit)
+                exit-recursive-scan)
       (setf *git-projects* (parse-projects))
-      (setf result (find-project-directory name :exit t)))
+      (setf result (find-project-directory name :exit-recursive-scan t)))
+    ;; Avoid false positives: check the directory exists.
+    ;; It could have been deleted.
+    (unless (or exit-recursive-scan
+                (and result (probe-file result)))
+      (setf *git-projects* (parse-projects))
+      (setf result (find-project-directory name :exit-recursive-scan t)))
     result))
 
 ;XXX: add ftype declaration.
--- a/source/vcs-mode.lisp
+++ b/source/vcs-mode.lisp
@@ -113,20 +113,19 @@ Ask for which directory to clone to, expect if there is one single choice."
            (clone-uri (quri:copy-uri uri))
            (existing-repo (next/vcs::find-project-directory project-name))
            target-dir)
-      (if project-name
-          (if existing-repo
-              (echo "This repository exists in ~a" existing-repo)
-              (progn
-                (if (= 1 (length next/vcs:*vcs-projects-roots*))
-                    (progn
-                      (setf target-dir (first next/vcs:*vcs-projects-roots*))
-                      (next/vcs::clone project-name root-name target-dir clone-uri))
-                    (with-result (target-dir (read-from-minibuffer
-                                              (make-instance 'minibuffer
-                                                             :input-prompt "Target directory:"
-                                                             :completion-function #'next/vcs::projects-roots-completion-function)))
-                      (next/vcs::clone project-name root-name target-dir clone-uri)))))
-          (echo "Could not find the project name.")))))
+      (cond
+        ((not project-name)
+         (echo "Could not find the project name."))
+        (existing-repo
+         (echo "This repository exists in ~a" existing-repo))
+        ((= 1 (length next/vcs:*vcs-projects-roots*))
+         (setf target-dir (first next/vcs:*vcs-projects-roots*))
+         (next/vcs::clone project-name root-name target-dir clone-uri))
+        (t (with-result (target-dir (read-from-minibuffer
+                                     (make-instance 'minibuffer
+                                                    :input-prompt "Target directory:"
+                                                    :completion-function #'next/vcs::projects-roots-completion-function)))
+             (next/vcs::clone project-name root-name target-dir clone-uri)))))))
 
 (define-command git-clone ()
   "Alias of `vcs-clone'."
--- a/source/vcs-mode.lisp
+++ b/source/vcs-mode.lisp
@@ -3,9 +3,12 @@
   (:export :*vcs-projects-roots*)
   (:documentation "Interact with Git repositories.
 
-New command: vcs-clone (alias git-clone), to clone a VCS repository on disk (Git only at the moment).
+New command: vcs-clone (alias git-clone), to clone a VCS repository on
+disk (Git only at the moment). rst rst rts rst rst rst rst rt rst rt
+rtr tr strs t rst rs t
 
-Change the `*vcs-projects-roots*' list to define where to look for existing repositories on disk.
+Change the `*vcs-projects-roots*' list to define where to look for
+existing repositories on disk.
 
 The clone command is run asynchronously.
 
@@ -13,13 +16,14 @@ The clone command is run asynchronously.
 *Disclaimer*: this feature is meant to grow with Next 1.4 and onwards!
 ***********************************************************************
 
-We could clone on Github/Gitlab, be notified if we have unpushed changes, browse files in a text editor, use hooks...
+We could clone on Github/Gitlab, be notified if we have unpushed
+changes, browse files in a text editor, use hooks...
 "))
 
 (in-package :next/vcs)
 
 (defparameter *vcs-projects-roots* '("~/projects" "~/src" "~/work" "~/common-lisp" "~/quicklisp/local-projects")
-  "A list of directories to look for VCS repositories to.")
+  "A list of directories to look for VCS repositories into.")
 ;; Possible improvement: specify the depth to look for projects alongside the directory.
 ;; See magit-list-repositories.
 
@@ -40,7 +44,7 @@ We could clone on Github/Gitlab, be notified if we have unpushed changes, browse
 
 (defun find-project-directory (name &key exit)
   "Return the directory pathname of the project named NAME.
-If EXIT is true and the project was not found, don't parse the projects roots again."
+If EXIT is non-nil and the project was not found, don't parse the project roots again."
   (unless *git-projects*
     (setf *git-projects* (parse-projects)))
   (let ((result (find name *git-projects*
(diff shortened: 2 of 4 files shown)


documents/MANUAL.org
```diff
@@ -314,6 +315,19 @@ and other information within your cache.
...
  global for the Qt one.
 
+** Cloning git repositories
+
+Use the  =git-clone= command to  clone a  git repository to  disk. The
+clone command is run asynchronously.
+
+Change  the =next/git:*git-projects-roots*=  list to  define where  to
+look for existing git repositories. By default, the git-clone commands
+looks into =~/projects= and =~/work=.
```
Ambrevar: Can you disable text justification?

documents/MANUAL.org
```diff
@@ -314,6 +315,19 @@ and other information within your cache.
  Qt port: the Gtk one sets  proxies per-buffer, whereas it is currently
  global for the Qt one.
 
+** Cloning git repositories
+
+Use the  =git-clone= command to  clone a  git repository to  disk. The
```
Ambrevar: git => Git (everywhere when it's not a command)

source/git-mode.lisp
```diff
@@ -3,11 +3,14 @@
...
-  - on Github
-  - on Gitlab
-- be notified if you have unpushed changes
+New command: git-clone, to clone a git repository on disk.
+
+Change the `*git-projects-roots*' list to define where to look for
+existing git repositories on disk.
+The clone command is run asynchronously.
+
+Much can be added! We could clone on Github/Gitlab, be notified if we
```
Ambrevar: Prefix with "TODO: "?

source/git-mode.lisp
```diff
@@ -3,11 +3,14 @@
...
+have unpushed changes, browse files in a text editor,...
```
Ambrevar: No comma before an ellipsis.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+"))
+
+(in-package :next/git)
+(annot:enable-annot-syntax)
```
Ambrevar: No need for annot syntax if you don't use it.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+
+
+(defparameter *git-projects-roots* '("~/projects" "~/work")
```
Ambrevar: Maybe add ~/common-lisp and ~/quicklisp/local-projects since they are widespread defaults in the Common Lisp community.
Also ~/src is quite common in the Unix world.

Ambrevar: Shouldn't make this use the syntax of `magit-repository-directories`?  If not now, maybe add a TODO for later?

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+(defparameter *git-projects* '()
+  "Currently registered git projects.")
+
+(defun search-git-directories (dir)
+  "Search all directories that contain a .git/ subdirectory, one level deep inside DIR."
+  (when (uiop:directory-pathname-p (uiop:ensure-directory-pathname dir))
+    (loop for dir in (uiop:subdirectories dir)
+       for git-dir = (merge-pathnames dir ".git/")
+       when (uiop:directory-exists-p git-dir)
+       collect dir)))
```
Ambrevar: Maybe have it search N levels deep, where N is a keyword argument and is unlimited by default.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+
+(defun parse-projects ()
+  "Scan `*git-projects-roots*' and register git repositories."
+  (setf *git-projects*
+        (mapcan #'search-git-directories *git-projects-roots*)))
```
Ambrevar: Maybe no need to make this function non-pure.
Set `*git-projects*` at the caller site.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+
+(defun find-project-directory (name &key exit)
+  "Return the directory pathname of the project named NAME."
```
Ambrevar: Undocumented `exit` key arg.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+  (unless *git-projects*
+    (parse-projects))
+  (let ((result (find name *git-projects*
+                      :key (lambda (dir)
+                             (car (last (str:split "/" (namestring dir) :omit-nulls t))))
```
Ambrevar: car => first

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+                      :test #'string=)))
+    (unless (or result
+                exit)
+      (parse-projects)
```
Ambrevar: Why parsing projects again?  It was done at the beginning, it think once should be enough.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+  (ensure-directories-exist base)
+  ;; truename expands the tilde, but fails if the directory doesn't actually exist.
+  (let* ((base-truename-path (truename base))
+         (base-truename-string (namestring base-truename-path))
+         (dir-string (string-trim (list #\/) (namestring dir)))
+         (joined (str:concat base-truename-string dir-string)))
+    joined))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(in-package :next)
```
Ambrevar: Should we place git-clone in the "next" or in "git" package?
So far we've only used packages for modes, but you don't define a mode here.

I can see 2 options:
- Move everything to the "next" package.
- Add a "git-mode" and leave everything in the "git" package.  Then users can enable git-mode if they wish.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+                (setf (quri:uri-path clone-uri)
+                      (str:concat "/" root-name "/" project-name))
+                (handler-case (progn
+                                (uiop:launch-program
+                                 (list "git" "clone"
+                                       (quri:render-uri clone-uri)
+                                       (next/git::concat-filenames (first next/git::*git-projects-roots*) project-name))))
+                  (error (c)
+                    (log:warn "Error cloning ~a: ~a" project-name c)
+                    (echo "There was an error cloning ~a." project-name)))))
```
Ambrevar: When we have a `*Messages*` buffer, we can have our own `warn` function which combines both `echo` and `log:warn`.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+                      :test #'string=)))
+    (unless (or result
+                exit)
+      (parse-projects)
+      (setf result (find-project-directory name :exit t)))
+    result))
+
+(defun concat-filenames (base dir)
+  "Concat filenames. Handle a tilde in BASE.
+  Create BASE if it doesn't exist."
```
Ambrevar: Indentation to column 0.

source/git-mode.lisp
```diff
@@ -0,0 +1,96 @@
...
+  (setf base
+        (str:concat (string-right-trim (list #\/) base)
+                    "/"))
+  (ensure-directories-exist base)
+  ;; truename expands the tilde, but fails if the directory doesn't actually exist.
+  (let* ((base-truename-path (truename base))
+         (base-truename-string (namestring base-truename-path))
+         (dir-string (string-trim (list #\/) (namestring dir)))
+         (joined (str:concat base-truename-string dir-string)))
+    joined))
```
Ambrevar: Do we really need this function?
UIOP has `truenamize`, `merge-pathnames*`, `ensure-all-directories-exist`, etc.
Or maybe split this function in more canonical and reusable functions?

Ambrevar:
Should we really name this "git"?
Even if we only support Git for now, we could end up adding support for, say, mercurial.  Or the VCS of the future.

So what about naming this "vcs" instead?

source/git-mode.lisp
vindarel: mmh I would not, this is in the documentation, it's a message to the reader, it is also not a TODO but a "nice to have", "will be done if interest".

jmercouris:
> So what about naming this "vcs" instead?

Agreed

source/git-mode.lisp
vindarel: ✓ added more locations.

I added a comment (but not as TODO).

source/git-mode.lisp
vindarel: let's use alexandria:last-elt.

source/git-mode.lisp
vindarel: the list can be outdated. We might have cloned a repo to one of these directories since last time it was parsed.

source/git-mode.lisp
vindarel: We do! We need to expand the tilde from a string (not pathname), so `truename` alone is not enough. `uiop:truenamize` also isn't.

    (uiop:truenamize "~/foo")  => #P"~/foo" 

wrong, the following `ensure-directories-exist` will create the "~/foo" directory at the current working directory, instead of a proper `/home/user/foo`.

vindarel:
It makes sense but now I have to write a lot of "Git only ATM" warnings :S

source/git-mode.lisp
Ambrevar: How can it be outdated if "last time" is within the same (synchronous) function call?

source/git-mode.lisp
Ambrevar: Can you split this into smaller blocks, like a function that just expands the `~`?  (I wonder if there really isn't something that can to this in CL.)

vindarel: `truename` does expand the tilde *if the directory exists*.

Ambrevar:
Yes, so we can leverage this to make a function that expends the tilde
for any path.

vindarel:
> Yes, so we can leverage this to make a function that expends the tilde for any path.

no we can't, the directory must exist.

(or: I'd like such a function, but I got around it).

vindarel:
I splitted the function somehow.

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+*Disclaimer*: this feature is meant to grow with Next 1.4 and onwards!
+***********************************************************************
+
+We could clone on Github/Gitlab, be notified if we have unpushed changes, browse files in a text editor, use hooks...
+"))
+
+(in-package :next/vcs)
+
+(defparameter *vcs-projects-roots* '("~/projects" "~/src" "~/work" "~/common-lisp" "~/quicklisp/local-projects")
+  "A list of directories to look for VCS repositories to.")
```
Ambrevar: "to" what?

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+;; Possible improvement: specify the depth to look for projects alongside the directory.
+;; See magit-list-repositories.
```
Ambrevar: Please put a "TODO:" so that we can easily grep them in the code.

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+       when (uiop:directory-exists-p git-dir)
+       collect dir)))
+
+(defun parse-projects ()
+  "Scan `*vcs-projects-roots*'."
+  (mapcan #'search-git-directories *vcs-projects-roots*))
+
+(defun find-project-directory (name &key exit)
+  "Return the directory pathname of the project named NAME.
+If EXIT is true and the project was not found, don't parse the projects roots again."
```
Ambrevar: projects: no "s".

Ambrevar: "EXIT is non-nil"

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+  (unless *git-projects*
+    (setf *git-projects* (parse-projects)))
+  (let ((result (find name *git-projects*
+                      :key (lambda (dir)
+                              (alexandria:last-elt (str:split "/" (namestring dir) :omit-nulls t)))
+                      :test #'string=)))
+    (unless (or result
+                exit)
+      (setf *git-projects* (parse-projects))
```
Ambrevar: I still don't understand why *git-projects* is being re-computed here.  Does (parse-projects) produce the same output twice, here and at the beginning of the function?

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+      (setf result (find-project-directory name :exit t)))
+    result))
+
+(defun ensure-directory (base)
+  "Create this directory if it doesn't exist.
+Needed to call `truename' to exand a tilde after it.
+BASE: directory name (string)."
+  (check-type base string)
```
Ambrevar: With `ftype` you don't need to document the type of BASE.

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+                (if (= 1 (length next/vcs:*vcs-projects-roots*))
+                    (progn
+                      (setf target-dir (first next/vcs:*vcs-projects-roots*))
+                      (next/vcs::clone project-name root-name target-dir clone-uri))
+                    (with-result (target-dir (read-from-minibuffer
+                                              (make-instance 'minibuffer
+                                                             :input-prompt "Target directory:"
+                                                             :completion-function #'next/vcs::projects-roots-completion-function)))
+                      (next/vcs::clone project-name root-name target-dir clone-uri)))))
+          (echo "Could not find the project name.")))))
```
Ambrevar: This calls for a `cond` / `match` ;)

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+                      :test #'string=)))
+    (unless (or result
+                exit)
+      (setf *git-projects* (parse-projects))
+      (setf result (find-project-directory name :exit t)))
+    result))
+
+(defun ensure-directory (base)
+  "Create this directory if it doesn't exist.
+Needed to call `truename' to exand a tilde after it.
```
Ambrevar: When do you call `truename'?  Here or the caller?

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+                    "/"))
+  (ensure-directories-exist base))
+
+(defun concat-filenames (base dir)
+  "Concat filenames. Expand a tilde in BASE.
+Create BASE if it doesn't exist."
+  (ensure-directory base)
+  ;; truename expands the tilde, but fails if the directory doesn't actually exist.
+  ;; The tilde must be expanded for the following git clone command, that otherwise creates
+  ;; ./~/my/foo instead of /home/user/my/foo.
```
Ambrevar: I don't understand where you fix the ~ expansion.

Ambrevar:
There might have been a misunderstanding about the "tilde expansion": what I mean is that you can write a function like Emacs' `expand-file-name` that simply expands the tilde and that's it.

Then this would simplify the rest of the function.

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+
+New command: vcs-clone (alias git-clone), to clone a VCS repository on disk (Git only at the moment).
+
+Change the `*vcs-projects-roots*' list to define where to look for existing repositories on disk.
+
+The clone command is run asynchronously.
+
+***********************************************************************
+*Disclaimer*: this feature is meant to grow with Next 1.4 and onwards!
+***********************************************************************
```
Ambrevar: I'm not a big fan of the star-boxes, they take 2 lines of precious screen real estate for little gain.  "*Disclaimer*" is already emphasized.

source/vcs-mode.lisp
```diff
@@ -0,0 +1,138 @@
...
+
+We could clone on Github/Gitlab, be notified if we have unpushed changes, browse files in a text editor, use hooks...
```
Ambrevar: You might have misunderstood my comment about justification of the text: I meant the "justification to the right".  The conventional style is to wrap paragraphs around column 80 or so, but not to justify to the right (which is never done in the rest of the code base, so we just want to be consistent here).

vindarel: Look, I'll pay attention, but I'm using emacs' `M-q fill-paragraph` default.

source/vcs-mode.lisp
vindarel: good point to add this, I added a XXX, but I still think telling the type in the docstring helps the developer.

source/vcs-mode.lisp
vindarel: I thought about this workflow:

- user calls git clone on a project
  - find-project-directory: no result (empty projects list)
  - parse is called for the first time, it computes the list of projects
- more internet surfing, time flies
- user clones git repo B on her machine, not using Next
- user calls git clone on the same project, or on a project with the same name, etc
  - find-project-directory: finds nothing in the existing `*git-projects*`:
  - parse is called, 
  - find-project-directory called a second time: it finds the existing project, as expected.

If we don't do this, git clone would fail (without us noticing).

source/git-mode.lisp
vindarel: I like it this way, the package separates concerns. I might give the mode a shot, but I fear it isn't useful right now.

source/vcs-mode.lisp
Ambrevar: This is no big deal, I did not mean to be nagging here, sorry if I came across this way, this is just a friendly conversation.

After a quick search, I found out that if you use the prefix argument on `M-q`, it will right-justify.  You might be pressing C-u by mistake or something.

I've tested the behaviour on `emacs -Q`.

vindarel:
> There might have been a misunderstanding about the "tilde expansion": what I mean is that you can write a function like Emacs' expand-file-name that simply expands the tilde and that's it.

> Then this would simplify the rest of the function.

There's simpler ! Before, I absolutely wanted to merge pathnames with the base that didn't exist: it was impossible. But since I now ensure the base exists (and only the one we're interested in), I can simply

     (merge-pathnames (truename base) dir))

pfiu.

source/vcs-mode.lisp
vindarel: On my left and on my right: Pierre. Go ! 

I like how the logic reads, forgive me… please enlighten me :D

vindarel:
If there are no blockers, I'll merge after the 1.3.2 release.

source/vcs-mode.lisp
Ambrevar: OK, I get it now.

Could you add some of this to the docstring?

Also I suggest renaming `exit` to something like `do-not-rescan`.

source/vcs-mode.lisp
Ambrevar: The type information is displayed when you prompt the documentation with SLIME.  Thus if your docstring contains it, it's only redundant.

source/vcs-mode.lisp
Ambrevar: I think the following is equivalent (untested):

```lisp
(cond
 ((not project-name)
  (echo "Could not find the project name."))
 (existing-repo
  (echo "This repository exists in ~a" existing-repo))
 ((= 1 (length next/vcs:*vcs-projects-roots*))
  (setf target-dir (first next/vcs:*vcs-projects-roots*))
  (next/vcs::clone project-name root-name target-dir clone-uri))
 (t (with-result (target-dir (read-from-minibuffer
                              (make-instance 'minibuffer
                                             :input-prompt "Target directory:"
                                             :completion-function #'next/vcs::projects-roots-completion-function)))
                 (next/vcs::clone project-name root-name target-dir clone-uri))))

```

source/vcs-mode.lisp
vindarel: nice one!

source/vcs-mode.lisp
vindarel: Much simpler:

- user clones project A
- user clones A again
  - find-project doesn't find it (old project list), re-scans: ok

source/vcs-mode.lisp
vindarel: thanks:)

source/vcs-mode.lisp
vindarel: done.

PR #390, opened 2019-09-23
Fix PASTE-FROM-RING (needs to use 'ring::minibuffer class).

noogie13:


jmercouris:
Instead of using a double colon, is there a possibility to export the symbol?

Ambrevar:
Yes, we could add a @export @export-accessors on top of `(defclass minibuffer...)`.
@noogie13 Mind updating the PR?
While we are at it, we can remove all the `next::` prefix in `(make-instance 'minibuffer...)` in `web-mode.lisp`.

PR #393, opened 2019-09-25
Add option to focus on reopened buffers

4t0m:
It doesn't seem like the current implementation focuses on reopened tabs, which I expect many would want it to do.  I added an option to remote-interface and refactored a bit of the recent-buffers code.  Let me know what you think.


jmercouris:
Thank you very much for this pull request! It was very kind of you to spot this problem and find a good solution! One of us will test it and merge it in ASAP!

source/remote.lisp
```diff
@@ -237,6 +237,7 @@ Most recent messages are first.")
    (minibuffer-set-url-history :accessor minibuffer-set-url-history :initform (ring:make))
    (recent-buffers :accessor recent-buffers :initform (ring:make :size 50)
                    :documentation "A ring that keeps track of deleted buffers.")
+   (focus-on-reopened-buffer-p :accessor focus-on-reopened-buffer-p :initform nil)
```
jmercouris: If you expect most people will want this as a default, perhaps setting it to `t` here would make sense?

4t0m: Indeed :)

vindarel:
yay, I was annoyed by this and would like it to be the default too, thanks!

Ambrevar:
I would suggest the following:

- Remove `recreate-buffer`.
- Add a `:focus` key argument to `reload-buffer` and focus from there.  You'll want to make `buffer` a key argument as well.

This way we don't need a new function and it's more consistent with how `open-urls` works.
Does that make sense?

4t0m:
Like so?

Ambrevar:
Thanks!

Issue #384, opened 2019-09-17
Unwanted behavior in VI-INSERT-MODE -> VI-NORMAL-MODE on click


4t0m:
Clicking on the currently-active input box while in VI-INSERT-MODE causes a switch to VI-NORMAL-MODE, which is annoying when using the mouse to move the cursor.  

I would personally want to maintain VI-INSERT-MODE when selecting other input elements as well.

Ambrevar:
I've noticed this, but it only happens on specific input fields it
seems.  I can reproduce consistently on GitHub comments.

I've been annoyed by this for too long, this is high on my priority list
(but not before a week at least).

Patches welcome!  :)
The function to look at is `vi-button1` in vi-mode.lisp.

Ambrevar:
This should be fixed now, feel free to re-open if the issue persists.

PR #394, opened 2019-09-25
add check for textarea tag when determining vi-insert-mode

4t0m:
I think this addresses #384.  It was only checking for input tags and not textarea tags.

### DIFF
--- a/source/vi-mode.lisp
+++ b/source/vi-mode.lisp
@@ -52,7 +52,8 @@
 (define-parenscript %clicked-in-input? ()
   (ps:chain document active-element tag-name))
 
-(defmethod is-input-tag ((tag string))
+(declaim (ftype (function (string)) input-tag-p))
+(defun input-tag-p (tag)
   (or (string= tag "INPUT")
       (string= tag "TEXTAREA")))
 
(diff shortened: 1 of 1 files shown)


Ambrevar:
Thanks for digging this out, it seems to work indeed! :)
(Had been annoying me for way too long :p)

Just a nit: can you rename the function to `input-tag-p` and can you make the defmethod a function, and use type declaration to enforce type checking:
```
(declaim (ftype (function (string)) input-tag-p))
(defun input-tag-p (tag)
  (or (string= tag "INPUT")
      (string= tag "TEXTAREA")))
```

Methods should only be used if we are actually going to dispatch methods (e.g. other objects would have an "input-tag-p" method).  Methods should not be used for type checking.

Ambrevar:
To clarify: in Common Lisp, booleans and boolean-returning function are conventionally suffixed with `-p`, which stands for "predicate".
Then you don't need "is-".

4t0m:
Should be fixed!  Thanks for the pointer about conventions :).  It's all stuff I want to learn/understand.

Ambrevar:
You're welcome, and thanks for this very useful fix!

PR #396, opened 2019-09-26
Vindarel/debian package

### LINKED ISSUE #76: provide a deb package
It's an interesting project, but the compilation steps are too many for somebody who wants to just chekc it out. It would be much easier to achieve a wider adoption if you could provide a deb package for debian/ubuntu.

vindarel:
fixes #76 

https://next.atlas.engineer/static/release/next_1.3.2_amd64.deb


Ambrevar:
Great!
Did you manage to fix the dpkg warnings?

We should mention that it's Debian 10+ / Ubuntu / Mint / etc.  In particular, it won't work for Debian 9.  Wanna update the README?

vindarel:
README updated, and nope not yet (it's debuild warnings, not dpkg, except if I missed something).

jmercouris:
Can you please move this under build-scripts? I don't think this is important as a top-level folder

vindarel:
I think it is for all the tools to find the Makefile, but I can give it a shot.

jmercouris:
OK, thank you. I just want to keep the top-level as clean and as small as possible to not be intimidating and a sprawling mess of files. Also reduces the "distance to fold" so that users can see the README

vindarel:
tried: it is required (unless non-standard rules file mixing).

PR #397, opened 2019-09-26
Wip functional config

Ambrevar:
Let me know if you like https://github.com/atlas-engineer/next/issues/381#issuecomment-535449592.

If so, I'll update the documentation and merge.

### DIFF
--- a/documents/EXTENSION.org
+++ b/documents/EXTENSION.org
@@ -15,7 +15,10 @@ Setting the start page can easily be achieved in the following manner:
 
 #+NAME: start-page-set
 #+BEGIN_SRC lisp
-(setf (get-default 'remote-interface 'start-page-url) "about:blank")
+(defun my-interface-defaults ()
+  (setf (start-page-url *interface*) "about:blank"))
+
+(hooks:add-to-hook '*after-init-hook* #'my-interface-defaults)
 #+END_SRC
 
 * Creating your own Parenscript
@@ -87,7 +90,7 @@ the command output.
   "Open a shell buffer."
   (set-active-buffer *interface* (make-buffer :name "*shell*" :default-modes '(shell-mode))))
 
-(define-key "C-x s" #'shell)
+(define-key :keymap *my-keymap* "C-x s" #'shell)
 #+END_SRC
 
 * Reloading the page
@@ -103,7 +106,7 @@ The following function provides a way to reload the current page.
   "Reload page."
   (%reload-page))
 
-(define-key "C-r" #'reload-page)
+(define-key :keymap *my-keymap* "C-r" #'reload-page)
 #+END_SRC
 
 * Scroll the page up and down by a screen
@@ -127,6 +130,6 @@ one screen.
   (%scroll-page-up))
 
 ;; Bind the new functions.
-(define-key "C-v" #'scroll-page-down)
-(define-key "M-v" #'scroll-page-up)
+(define-key :keymap *my-keymap* "C-v" #'scroll-page-down)
+(define-key :keymap *my-keymap* "M-v" #'scroll-page-up)
 #+END_SRC
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -33,7 +33,6 @@ end-user allowing for infinite customization.
   - [[#getting-input-from-the-user][Getting input from the user]]
   - [[#network-resource-dispatch-including-ad-blocking][Network resource dispatch (including ad-blocking)]]
   - [[#styles][Styles]]
-  - [[#proxy][Proxy]]
   - [[#hooks][Hooks]]
   - [[#startup-behaviour][Startup behaviour]]
 
@@ -113,18 +112,26 @@ The =~a= in the search engine URI is a place holder for the search pattern.
 To set the list of search engines, do:
 
 #+begin_src lisp
-(setf (get-default 'remote-interface 'search-engines)
-      '(("default" . "https://duckduckgo.com/?q=~a")
-        ("yt" . "https://www.youtube.com/results?search_query=~a")
-        ("wiki" . "https://en.wikipedia.org/w/index.php?search=~a")))
+(defun my-interface-defaults ()
+  (setf (search-engines *interface*)
+        '(("default" . "https://duckduckgo.com/?q=~a")
+          ("yt" . "https://www.youtube.com/results?search_query=~a")
+          ("wiki" . "https://en.wikipedia.org/w/index.php?search=~a"))))
+
+(hooks:add-to-hook '*after-init-hook* #'my-interface-defaults)
 #+end_src
 
 and to append a search engine do the list, you can do
 
 #+begin_src lisp
-(nconc (get-default 'remote-interface 'search-engines))
-       '(("q" . "http://quickdocs.org/search?q=~a")
-         ("yt" . "https://www.youtube.com/results?search_query=~a")))
+(defun my-interface-defaults ()
+  (setf (search-engines *interface*)
+        (append
+         '(("yt" . "https://www.youtube.com/results?search_query=~a")
+           ("wiki" . "https://en.wikipedia.org/w/index.php?search=~a"))
+         (search-engines *interface*))))
+
+(hooks:add-to-hook '*after-init-hook* #'my-interface-defaults)
 #+end_src
 
 ** Jumping to Headings
@@ -194,6 +201,21 @@ same instance of a mode (by default at least).
 Modes are typically defined in their separate Common Lisp package.  This allows
 for defining mode-specific functions and variables in a separate namespace.
 
+
+To enable a mode for all buffers by default, add the mode to the list of
+default modes:
+
+#+begin_src lisp
+(defun my-buffer-defaults (buffer)
+  (pushnew 'vi-normal-mode (default-modes buffer)))
+
+(defun my-interface-defaults ()
+  (hooks:add-to-hook (hooks:object-hook *interface* 'buffer-make-hook)
+                     #'my-buffer-defaults))
+
+(hooks:add-to-hook '*after-init-hook* #'my-interface-defaults)
+#+end_src
+
 *** Windows vs. buffers
 
 When opening a link from an external program, or when clicking on a link while
@@ -207,7 +229,10 @@ You can change the default behaviour by adding the following to your
 configuration file:
 
 #+begin_src lisp
-(setf (get-default 'remote-interface 'open-external-link-in-new-window-p) t)
+(defun my-interface-defaults ()
+  (setf (open-external-link-in-new-window-p *interface*) t))
+
+(hooks:add-to-hook '*after-init-hook* #'my-interface-defaults)
 #+end_src
 
 ** Searching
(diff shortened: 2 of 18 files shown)


vindarel:
Please wait a bit, this requires more thinking and tests (at least on my side), thanks.

Ambrevar:
This is my new `init.lisp`:

``` lisp
(in-package :next)

;; Use development platform port.
(setf +platform-port-command+
      "~/common-lisp/next/ports/gtk-webkit/next-gtk-webkit")

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(defun eval-in-emacs (&rest s-exps)
  "Evaluate S-exps with `emacsclient'."
  (let ((s-exps-string (str:replace-all
                        ;; Discard the package prefix.
                        "next::" ""
                        (write-to-string
                         `(progn ,@s-exps) :case :downcase))))
    (log:debug "Sending to Emacs: ~a" s-exps-string)
    (ignore-errors (uiop:run-program
                    (list "emacsclient" "--eval" s-exps-string)))))

(defvar *my-keymap* (make-keymap)
  "Keymap for `my-mode'.")

(define-command org-capture (&optional (buffer (current-buffer)))
  "Org-capture current page."
  (eval-in-emacs
   `(org-link-set-parameters
     "next"
     :store (lambda ()
              (org-store-link-props
               :type "next"
               :link ,(url buffer)
               :description ,(title buffer))))
   `(org-capture)))
(define-key :keymap *my-keymap* "C-M-o" #'org-capture)

(define-command youtube-dl-current-page (&optional (buffer (current-buffer)))
  "Download a video in the currently open buffer."
  (eval-in-emacs
   (if (search "youtu" (url buffer))
       `(progn (youtube-dl ,(url buffer)) (youtube-dl-list))
       `(ambrevar/youtube-dl-url ,(url buffer)))))
(define-key :keymap *my-keymap* "C-M-c d" 'youtube-dl-current-page)

(define-command play-video-in-current-page (&optional (buffer (current-buffer)))
  "Play video in the currently open buffer."
  (uiop:run-program (list "mpv" (url buffer))))
(define-key :keymap *my-keymap* "C-M-c v" #'play-video-in-current-page)

(define-mode my-mode ()
  "Dummy mode for the custom key bindings in `*my-keymap*'."
  ((keymap-schemes :initform (list :emacs-map *my-keymap*
                                   :vi-normal *my-keymap*))))

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(defvar +youtube-dl-command+ "youtube-dl"
  "Path to the 'youtube-dl' program.")

(defun auto-yt-dl-handler (url)
  "Download a Youtube URL asynchronously to /tmp/videos/.
Videos are downloaded with `+youtube-dl-command+'."
  (let ((uri (quri:uri url)))
    (when (and uri
               (member-string (quri:uri-domain uri) '("youtube.com" "youtu.be"))
               (string= (quri:uri-path uri) "/watch"))
      (log:info "Youtube: downloading ~a" url)
      (uiop:launch-program (list +youtube-dl-command+ url "-o" "/tmp/videos/%(title)s.%(ext)s"))))
  url)

(defun old-reddit-handler (url)
  (let* ((uri (quri:uri url)))
    (if (search "www.reddit" (quri:uri-host uri))
        (progn
          (setf (quri:uri-host uri) "old.reddit.com")
          (let ((new-url (quri:render-uri uri)))
            (log:info "Switching to old Reddit: ~a" new-url)
            new-url))
        url)))

(defvar *my-unproxied-domains*
  '("jit.si"
    "wikipedia.org"))

(defun auto-proxy-handler (url)
  (let* ((uri (quri:uri url))
         (domain (and uri (quri:uri-domain uri))))
    (when domain
      (next/proxy-mode:proxy-mode
       :activate
       (not (member-string domain *my-unproxied-domains*)))))
  url)

(defvar *my-blocked-hosts*
  '("platform.twitter.com"
    "syndication.twitter.com"
    "m.media-amazon.com"))

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

(defun my-buffer-defaults (buffer)
  (dolist (mode '(vi-normal-mode proxy-mode my-mode))
    (pushnew mode (default-modes buffer)))
  ;; TODO: Pros and cons of the following approach to mode init?
  ;; The problem with DEFAULT-MODES is that it does not allow for configuring
  ;; the modes.
  (next/blocker-mode:blocker-mode :hostlists *my-blocked-hosts*
                                  :buffer buffer)
  (dolist (handler (list #'old-reddit-handler #'auto-proxy-handler))
    (hooks:add-to-hook (hooks:object-hook buffer 'load-hook)
                       handler)))

(defun my-window-defaults (window)
  (setf (minibuffer-open-height window) 400))

(defun my-minibuffer-defaults (minibuffer)
  (setf
   (minibuffer-line-height minibuffer) "1.1em"
   (minibuffer-font-size minibuffer) "1.1em"))

(defun my-interface-defaults ()
  (setf (download-directory *interface*) "~/temp")
  (hooks:add-to-hook (hooks:object-hook *interface* 'buffer-make-hook)
                     #'my-buffer-defaults)
  (hooks:add-to-hook (hooks:object-hook *interface* 'minibuffer-make-hook)
                     #'my-minibuffer-defaults)
  (hooks:add-to-hook (hooks:object-hook *interface* 'window-make-hook)
                     #'my-window-defaults))

(hooks:add-to-hook '*after-init-hook* #'my-interface-defaults)
```

Ambrevar:
OK, I've tested it quite extensively and this seems to be in pretty good shape.

My only concern is that I'm not sure how to rewire the `open-file` bindings: now that all root-mode bindings are set (in stone) at the definition site of root mode, I dont think we can define "open-file" on top.

I can think of the following solution: define an `open-file-mode` which only contains the C-x C-f binding and have this mode added to the default-modes list.  Thoughts? @vindarel

Ambrevar:
I'm thinking of leaving the `setf get-default` definition as well as `add-to-default-list` for backward compatibility until next verison.  We add a deprecation warning so that users have time to adapt their config.

Ambrevar:
Regarding the file-mode-manager bindings: actually it's no problem to bind them in the root-mode definition directly, so that's even simpler.

Ambrevar:
Let me know if you think this is good to merge, besides the above 2 fixes.

jmercouris:
I'm OK, not too opinionated about this, that said I don't think most people will like this way of doing things.

jmercouris:
Mostly because it feels a rather convoluted way of setting settings.. Rather than what one would expect with a simple setf or other such macro.

vindarel:
same feeling.

vindarel:
So we have such a slot:

```
   (minibuffer-line-height :initarg :minibuffer-line-height
                           :accessor minibuffer-line-height
                           :type string
                           :initform ""
                           :documentation "CSS line height for the minibuffer.
Value is a string, e.g. '1em'.
You might want to configure the value on HiDPI screen.")
```

Your proposal is to configure it with

```
(defun my-minibuffer-defaults (minibuffer)
  (setf
   (minibuffer-line-height minibuffer) "1.1em")
```

then

```
(defun my-interface-defaults ()
  (hooks:add-to-hook (hooks:object-hook *interface* 'minibuffer-make-hook)
                     #'my-minibuffer-defaults))
```

then 

```
(hooks:add-to-hook '*after-init-hook* #'my-interface-defaults)
```

As a CL developer, I expect something like

```
(setf next/minibuffer:*minibuffer-line-height* "1.1em")
```

Did we already talk about that?

Ambrevar:
I understand, but what's really at play here is the _paradigm_.

Most of us are used to a "all global, non-functional" approach to
configuring programs (assuming they can be configured).

Take Emacs: Historically it had dynamic scoping by default, for the very
reason that it made it easier for the user.  This proved
to be a rather big mistake over time, because it makes programming
errors much more pervasive.

Still in Emacs: so many globals all over the place, it's really hard to
modify a value without impacting everything else.
The consequence: every couple of days, my Emacs breaks down in
mysterious ways and I have to restart it.



What I'm proposing here is a system in which the _whole state_ of the
program is a first-class object.  This gives us a lot of flexibility.



With Common Lisp, we are gifted with the package system.  So we can much
more easily isolate stuff.
While I'm insisting on this functional approach for the _core
of Next_, I don't think the external, "leaf" modules have to be this
way.  For instance, the VCS module has a couple of global variables and
that's probably alright if no other module is going to depend on it.



All in all, while this approach of configuring the _core of Next_ might
seem a bit convoluted, it has many long term benefits in my opinion:

- Full control of the whole state.

- If a slot is set from 2 different place, there is trace of this in the
  hooks.  In a way, this gives us a history of all values that are set
  to a slot.

- Resilient _core_: with no globals, it's much harder to break the state
  of Next.  Any broken _part_ can be re-instantiated.

- Locality: every value can be set per-buffer, or per-window, etc.  It's
  like Emacs' buffer-local variables, but for more objects.

- Less programming errors.



To conclude, the main issue I see with the above approach is that we
need to educate our audience to this "new" paradigm.

But hey, it's Lisp after all, the parentheses alone make a lot of people
feel uneasy already :p

Ambrevar:
See my answer above: I think that package globals are OK when used in
"leaf" packages (packages that are not going to be re-used by anyone).

For minibuffers, my fear is that it's too central to risk exposing
globals.

Example:

- Mode/Package A is activated and changes the default value of line-height.

- Same for Mode/Package B.

- When Mode/Package B is turned off / unloaded, the value of A is gone.

If we don't have a global but use the hook approach instead, unload B
would restore A.  In my opinion this is awesome.

vindarel:
> unload B would restore A.

When a mode is activated, can we not use a closure ? The previous value of the global would be preserved.

Ambrevar:
It's an interesting idea.
The closure could be stored in a "deactivation function".

I can see an integrity problem though:

- Enable A.
- Enable B.
- Disable A.

Now we are back with the default value, while B is still active.

What about those things that don't have a deactivation function?

How do you implement this in practice?  It seems to me that you'd need a
dedicated function, so if the user relies on `setf` it won't work.

Ambrevar:
Nothing is carved in stone even if we merge this.  After all if we keep `(setf (get-default ...))` until next version, we can always step back.  Or iterate and come up with something better.  Or use more globals / packages if we deem it fit.  What do you think?

I'd like to merge tomorrow, let me know if I should hold off for a little longer.

jmercouris:
Like I said before, it is OK to. merge :-)

I still have my reservations. I understand statelessness and all that... but there must be a more simple way, if not for the sake of the end-user. Otherwise it would be nice if make the core FAR more minimal and expose more things in packages (such that we can use globals that are easily SETF'able).

Ambrevar:
In my understanding, we don't chose what the "minimal core is": it's a matter of semantic.
The minibuffer is part of the core, because feature-wise it's a the center of everything.
Should we move it to a separate package, it would not change how central it is, and as such we should avoid using globals for it in my opinion.

For now, in my understanding the core is made of the following classes:
- remote-interface
- window
- buffer
- minibuffer
- command
- root-mode
- web-mode

But maybe some parts could be moved away from the slots.
For example, could move bookmarks path / importer / export to separate global variables?  Maybe, but that would prevent third-party packages from cleanly customizing the importer(s)/exporter(s).

I also thought of other approaches, such as keeping a git-like history of the state of everything.  This would be awesome, but that's much more complex :p

Ambrevar:
Maybe a better solution for users: considering the Lisp might already be too hard to write to the uninitiated, we probably need a graphical configuration interface anyways.  So we could have a web page were the user can set the default values for `window`, `buffer`, etc. and upon saving, it would generate a configuration file akin to what I wrote above.

Best of both worlds! :)

jmercouris:
I see you are already thinking of a customization buffer :-)

Well, this is still a ways away I think, but surely we will not pollute the users config as it is in Emacs' default, we should make a `customizations.lisp` as the default :D

Anyways, please merge this in whenever you are ready!

Ambrevar:
Will do!

Ambrevar:
Merged in c8a210ba515df9407b2defb8a1703c214d6499df.

PR #398, opened 2019-09-26
Fix hint box overlapping

alexherbo2:


Ambrevar:
Overlapping with what?  :p

jmercouris:
Probably each other?

alexherbo2:
The site elements.  It happens the site elements get a higher priority in the overlapping stack.  [`z-index`] ensures hint boxes are always on-top.

[`z-index`]: https://developer.mozilla.org/en-US/docs/Web/CSS/z-index

Ambrevar:
OK, sounds good.  Do you have an example website so we can test it?

According to https://developer.mozilla.org/en-US/docs/Web/CSS/integer#Syntax, integers have no standard range.  What would be the most reasonable value then?

Either way, we should probably comment the magic number :)

alexherbo2:
The maximum: 2147483647 (2³¹ − 1)

You can see here with the GitHub Star button.

Ambrevar:
Indeed, this fixes the GitHub buttons, thanks!

I understand the maximum you chose is over 32 signed integers, but the Mozilla documents I've linked explicitly states that this is not standard.

Should we stick to this value, I prefer `(1- (expt 2 31))`, it helps clarifying the meaning of the value.

Ambrevar:
I'll merge and add a link to the Mozilla pages, explaining that the range might have to be tweaked depending on the web renderer.

Ambrevar:
Thanks!

PR #404, opened 2019-09-26
Wip enable disable mode

Ambrevar:
(Note: This is only 3 commits, but I'm based on top of #397 which is not yet merged.)

At last, the list of modes and currently active modes!


Ambrevar:
Merged with 9be32d8e884ef4dd2114c8b5bce310bc3485868d.

PR #411, opened 2019-09-26
Fix "The value 31 is not of type SYMBOL" error at startup.

jl2:
Pulled master and built this afternoon and crashed with "The value 31 is not of type SYMBOL"


vindarel:
Indeed, thanks!

PR #427, opened 2019-09-30
Add password-manager package.

noogie13:
Has KeepassXC and Password-Store interfaces.

KeepassXC is broken because you can't enter the password with :invisible-input because then nothing get's passed to the anonymous function (need fixing).

Two main commands - `save-new-password`: saves a new password.

 `copy-password`: copies password into clipboard for 5 seconds (should we add the clipboard into the copy/paste ring?

### DIFF
--- a/libraries/password-manager/password-keepassxc.lisp
+++ b/libraries/password-manager/password-keepassxc.lisp
@@ -1,4 +1,6 @@
-(in-package :next)
+(in-package :password)
+
+(annot:enable-annot-syntax)
 
 (defmethod list-passwords ((password-interface keepassxc-interface))
   (let* ((st (make-string-input-stream (master-password password-interface)))
--- a/libraries/password-manager/password-keepassxc.lisp
+++ b/libraries/password-manager/password-keepassxc.lisp
@@ -1,17 +1,17 @@
 (in-package :password)
 
-(annot:enable-annot-syntax)
+(defvar *keepassxc-cli-program* (executable-find "keepassxc-cli"))
 
 (defmethod list-passwords ((password-interface keepassxc-interface))
   (let* ((st (make-string-input-stream (master-password password-interface)))
-         (output (uiop:run-program `("keepassxc-cli"
+         (output (uiop:run-program `(*keepassxc-cli-program*
                                      "ls" ,(password-file password-interface))
                                    :input st :output '(:string :stripped t))))
     (remove "Recycle Bin/" (rest (cl-ppcre:split "\\n" output)) :test #'equal)))
 
 (defmethod clip-password ((password-interface keepassxc-interface) password-name)
   (let* ((st (make-string-input-stream (master-password password-interface)))
-         (output (uiop:run-program `("keepassxc-cli"
+         (output (uiop:run-program `(*keepassxc-cli-program*
                                      "show" ,(password-file password-interface)
                                      ,password-name)
                                    :input st :output '(:string :stripped t))))
@@ -23,7 +23,7 @@
                                (format nil "~a~C~a"
                                        (master-password password-interface)
                                        #\newline password)))
-    (uiop:run-program `("keepassxc-cli"
+    (uiop:run-program `(*keepassxc-cli-program*
                         "add" "-p" ,(password-file password-interface) ,password-name)
                       :input st)))
 
@@ -31,7 +31,7 @@
   (when (master-password password-interface)
     (handler-case
         (let* ((st (make-string-input-stream (master-password password-interface)))
-               (output (uiop:run-program `("keepassxc-cli"
+               (output (uiop:run-program `(*keepassxc-cli-program*
                                            "ls" ,(password-file password-interface))
                                          :input st :output '(:string :stripped t))))
           (remove "Recycle Bin/" (rest (cl-ppcre:split "\\n" output)) :test #'equal))
--- a/libraries/password-manager/password-pass.lisp
+++ b/libraries/password-manager/password-pass.lisp
@@ -1,6 +1,6 @@
 (in-package :password)
 
-(annot:enable-annot-syntax)
+(defvar *password-store-program* (executable-find "pass"))
 
 (defmethod list-passwords ((password-interface password-store-interface))
   (let ((raw-list (directory (format nil "~a/**/*.gpg"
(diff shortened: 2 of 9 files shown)


libraries/password-manager/password-pass.lisp
```diff
@@ -0,0 +1,23 @@
...
+
+(annot:enable-annot-syntax)
+
+(defmethod list-passwords ((password-interface password-store-interface))
+  (let ((raw-list (directory (format nil "~a/**/*.gpg"
+                                     (password-directory password-interface))))
+        (dir-length (length (namestring
+                             (truename (password-directory password-interface))))))
+    (mapcar #'(lambda (x)
+                (subseq (namestring x) dir-length (- (length (namestring x)) 4)))
```
Ambrevar: If "4" stands for the `(length ".gpg")`, I suggest you use the latter to make your intention more explicit.

libraries/password-manager/password-keepassxc.lisp
```diff
@@ -0,0 +1,39 @@
...
+                      :input st)))
+
+(defmethod password-correct-p ((password-interface keepassxc-interface))
+  (when (master-password password-interface)
+    (handler-case
+        (let* ((st (make-string-input-stream (master-password password-interface)))
+               (output (uiop:run-program `("keepassxc-cli"
+                                           "ls" ,(password-file password-interface))
+                                         :input st :output '(:string :stripped t))))
+          (remove "Recycle Bin/" (rest (cl-ppcre:split "\\n" output)) :test #'equal))
```
Ambrevar: Same with "Recycle Bin/".

libraries/password-manager/password-pass.lisp
```diff
@@ -0,0 +1,23 @@
...
+            raw-list)))
+
+(defmethod clip-password ((password-interface password-store-interface) password-name)
+  (clip-password-string (uiop:run-program `("pass" "show" ,password-name)
```
Ambrevar: Move all "pass" to a global variable as for keepassxc.

Ambrevar:
Looks very good, excellent work! :)
In my opinion it's ready for merging (let me know if you have time to fix the few nits).

I'll fix the `:invisible-input` issue as soon as possible, sorry about that.

libraries/password-manager/password-pass.lisp
tviti: I'm not a core contributor, so hopefully I'm not out of place commenting on a PR like this, but would it be better to simply call `pass -c <password-name>` instead of using `clip-password-string` to copy to clipboard?

libraries/password-manager/password-keepassxc.lisp
```diff
@@ -0,0 +1,39 @@
+(in-package :password)
+
+(annot:enable-annot-syntax)
+
+(defmethod list-passwords ((password-interface keepassxc-interface))
+  (let* ((st (make-string-input-stream (master-password password-interface)))
+         (output (uiop:run-program `("keepassxc-cli"
+                                     "ls" ,(password-file password-interface))
+                                   :input st :output '(:string :stripped t))))
+    (remove "Recycle Bin/" (rest (cl-ppcre:split "\\n" output)) :test #'equal)))
```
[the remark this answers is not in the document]
noogie13: I think there is a `Recycle Bin/` for keepassxc db's. Will dig a little.

source/password.lisp
```diff
@@ -0,0 +1,32 @@
...
+                                 (make-instance 'minibuffer
+                                                :input-prompt "Name for new password:")))
+                 (master-password (read-from-minibuffer
+                                   (make-instance 'minibuffer
+                                                  :invisible-input-p t
+                                                  :input-prompt "New password:"))))
+    (password:save-password (password-interface *interface*) password-name master-password)))
+
+(defmacro with-password (password-interface &body body)
+  `(if (password:password-correct-p ,password-interface)
```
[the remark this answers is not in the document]
noogie13: Yes. I accidentally reverted back. I wanted to figure out repeating the call to enter password, but I get errors if I try to define recursive macro. Maybe I could just use a loop. I'll play around again.

source/password.lisp
```diff
@@ -0,0 +1,32 @@
...
+  (let ((password-list (password:list-passwords password-instance)))
+    (lambda (input)
+      (fuzzy-match input password-list))))
+
+(define-command save-new-password ()
+  "Save password to password interface."
+  (with-result* ((password-name (read-from-minibuffer
+                                 (make-instance 'minibuffer
+                                                :input-prompt "Name for new password:")))
+                 (master-password (read-from-minibuffer
```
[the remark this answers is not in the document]
noogie13: Yep

libraries/password-manager/password-pass.lisp
Ambrevar: Of course everyone is more than welcome to participate in any of the discussions here, that's why the development is public after all! :)

I believe the incentive here is twofold:
- `pass -c` depends on `xclip`.  Which is OK, since it's also a dependency of Next, but since it's a library that may end up as a separate package, we can avoid an extra dependency easily.
- `pass -c` modifies the system clipboard, which is a side effect the user may not necessarily want.  With `pass show` Next has more control over what we do with the password.

@noogie13 ?

Ambrevar:
I've just tested the input invisibility: it works on master! :)
Maybe you forgot to recompile everything, or maybe you've got some stale .fasls?
Don't forget to run `,delete-system-fasls RET next RET` from the SLIME REPL.

Ambrevar:
Great!  Green light me and I'll merge.

noogie13:
You can merge, but the :invisible-input problem remains. It will enter the top entry in the suggested strings, not the typed string. I suspect this to be an easy fix. Anyway I can fix any other problems with the code later on, if need be.

libraries/password-manager/password-keepassxc.lisp
noogie13: What is a better way to remove Recycle Bin/ or .gpg for example?

Ambrevar: I meant move the hard-coded string to a global variable.

Ambrevar:
Thanks a lot, this is a fantastic contribution!

vindarel:
Well done and thanks, @noogie13 !

vindarel:
nitpick:

```
M-x copy-password

Method handler (NEXT::CORE-OBJECT
                   NEXT::PUSH-INPUT-EVENT) signaled an error: There is no applicable method for the generic function
                                                                #<STANDARD-GENERIC-FUNCTION PASSWORD:PASSWORD-CORRECT-P (2)>
                                                              when called with arguments
                                                                (NIL).
```

I have neither keepassx or pass installed.

Ambrevar:
Fixed.

Ambrevar:
@noogie13 I've fixed some issues in cee2ce92f3296591fac54c92f2910ad4393bd674.
Pro-tip: in SLIME, call `,reload-system RET next RET` to get a full-blown compilation warning report.  The "re" is important :)

I've fixed the issue with the invisible input.

I've added password generation support when password is empty.

noogie13:
@Ambrevar so useful! Thanks!

PR #446, opened 2019-10-04
Fix delete buffer shorcut in manual

hsanjuan:
C-x C-k allows to select which buffer to kill. C-x k kills the current buffer.


jmercouris:
Thank you for the fix!

Ambrevar:
Actually it should be the other way around, so I've reverted this commit
and I've swapped the bindings instead.

Thanks for the catch!

jmercouris:
I did do a double take myself and checked the source :D I was wondering about that!

PR #448, opened 2019-10-05
My pyqt webengine

tviti:
These are modifications I made to the `next-pyqt` port, in order to get it working for me on `macOS`.

The largest change is probably the switch to using `Qt`'s own `QDBus` for `dbus` comms, since I couldn't get the `glib` mainloop working in `dbus-python`. I'm not sure if this was just broken for me, or if it's an issue in general on `macOS`, but I figure either way, the change also gets rid of a dependency.

I have also made some changes to the way that key-events are handled, because the original implementation did not properly distinguish between upper and lower case key events (the shift key in general didn't actually do anything for text input boxes).  The function `QtKeyEvent.key()` does not encode case at all, but from what I could tell, the core engine needs to have the case encoded in the `keycode` in order to properly handle it.

The `is_printable?` functionality that was added to the upstream codebase in a recent commit has also been implemented, since the port just won't work at all without it. The implementation is however a transliteration of what was added to the `gtkwebkit` port, and so may be missing some nuances (the only real testing done on this change amounts to `well, now it works`).

The way that javascript calls are handled by the port has also been changed. The problem with the original implementation (in terms of symptoms), was that introspection functions (i.e. using `C-h c` or `C-h v`) did not display anything in the newly created buffer. My knowledge of how javascript _works_ is very limited, but the conclusion I came to was that the calls to `runJavaScript` were being made before the DOM was ready. To fix this, the `buffers` class now queues JS, and the entire queue is evaluated when the page is _ready_ to run them. At the moment, `self.js_ready` is set when Qt signal `loadFinished` is emitted, and unset on `urlChanged`.

This signal choice has the undesirable side effect that the user is unable to scroll the page (or do any JS interaction) until the page has finished loading (and if they try to do things during load, the actions queue, and then happen all at once). I think that this could allow interaction during page loading by changing the signal to `loadProgress`.


jmercouris:
Hi Taylor, thank you for this extensive PR. I wanted to let you know that I haven't yet gotten a chance to try this yet, but I am aware of it! I will get to this as soon as I am able to. Once again, thank you very much!!

Ambrevar:
Why did you close it?

tviti:
I was looking through the commit history and realized it's pretty *dirty*
(e.g. with BS messages like "woops that didn't work", and also a file that wound
up never getting used). I'm still pretty new to git/VCS, so I was/still am
learning habits/protocol.

I figured it might be good to rebase to try to sort that stuff out, and thought
I should close before trying.

Ambrevar:
No problem!

PR #468, opened 2019-10-19
Add new dependency "trivial-types" to guix.scm

khinsen:
A recent commit added "trivial-types" as a new dependency. This must be reflected in `guix.scm`.


Ambrevar:
Good catch, thanks!

PR #475, opened 2019-10-29
extract and type search-engines, add github

vindarel:
now we get a type error at compile time. The slot type itself doesn't
trigger one.

Example to discuss.

### DIFF
--- a/source/minibuffer.lisp
+++ b/source/minibuffer.lisp
@@ -66,6 +66,27 @@
             :vi-normal map
             :vi-insert map)))))
 
+(declaim (type (integer) *max-lines*))
+@export
+(defparameter *max-lines* 8
+  "Max number of candidate lines to show in the minibuffer.
+You will want edit this to match the changes done to `minibuffer-font-size',
+`minibuffer-line-height' and `minibuffer-open-height'.")
+
+(declaim (type (string) *minibuffer-font-size*))
+@export
+(defparameter *minibuffer-font-size* "1em"
+  "CSS font size for the minibuffer.
+Value is a string, e.g. '1em'.
+You might want to configure the value on HiDPI screen.")
+
+(declaim (type (string) *minibuffer-line-height*))
+@export
+(defparameter *minibuffer-line-height* ""
+  "CSS line height for the minibuffer.
+Value is a string, e.g. '1em'.
+You might want to configure the value on HiDPI screen.")
+
 @export
 @export-accessors
 (defclass minibuffer (buffer)
@@ -125,25 +146,13 @@ candidates.")
             :documentation "The HTML content of the minibuffer.")
    (max-lines :initarg :max-lines
               :accessor max-lines
-              :type integer
-              :initform 8
-              :documentation "Max number of candidate lines to show.
-You will want edit this to match the changes done to `minibuffer-font-size',
-`minibuffer-line-height' and `minibuffer-open-height'.")
+              :initform *max-lines*)
    (minibuffer-font-size :initarg :minibuffer-font-size
                          :accessor minibuffer-font-size
-                         :type string
-                         :initform "1em"
-                         :documentation "CSS font size for the minibuffer.
-Value is a string, e.g. '1em'.
-You might want to configure the value on HiDPI screen.")
+                         :initform *minibuffer-font-size*)
    (minibuffer-line-height :initarg :minibuffer-line-height
                            :accessor minibuffer-line-height
-                           :type string
-                           :initform ""
-                           :documentation "CSS line height for the minibuffer.
-Value is a string, e.g. '1em'.
-You might want to configure the value on HiDPI screen.")
+                           :initform *minibuffer-line-height*)
    (minibuffer-style :accessor minibuffer-style
                      :initform (cl-css:css
                                 '((* :font-family "monospace,monospace")
--- a/source/remote.lisp
+++ b/source/remote.lisp
@@ -5,6 +5,16 @@
 (in-package :next)
 (annot:enable-annot-syntax)
 
+(declaim (type (integer) *status-buffer-height*))
+@export
+(defparameter *status-buffer-height* 36
+  "The height of the status buffer in pixels.")
+
+(declaim (type (integer) *minibuffer-open-height*))
+@export
+(defparameter *minibuffer-open-height* 200
+  "The height of the minibuffer when open.")
+
 @export
 @export-accessors
 (defclass window ()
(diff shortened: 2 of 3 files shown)


impaktor:
Related (I hope), I would expect that when openting a new buffer and starting typing "wiki" or "github" and then hitting TAB that it would auto-complete it? (I know, that's not what the PR is doing, but would be nice functionality)

vindarel:
To discuss in parallel to https://github.com/atlas-engineer/next/issues/419

Here we keep slots to configure things, but we extract their default value to global parameters.

The direct consequence is:

- we get type checking, at compile time, in the REPL, when the user launches Next or reloads the init file. We've seen it is useful.
  Slots types are not checked, despite the `:type` information.
  Type checking might be achieved with a custom macro from 419. Let's see to compare.

We can keep using `get-default` and all.

And actually, this should answer a concern of @Ambrevar: the global parameter and the slot value are two different things. So if the user uses our macros our the slot accessors, he doesn't touch the default parameter, and can use them to re-set the slot to its default value!

- default: `*search-engines*` == `(search-engines *interface*)`
- configuration: `(setf (search-engines *interface* <custom list>))`
- state 2: `*search-engines*` (untouched) != `(search-engines *interface*)`
- oops, I want to reset them: `(setf (search-engines *interface*) *search-engines*))`.
- back to default.

Ambrevar:
Indeed, with *search-engines*, there would be 2 defaults:

- The user defaults.
- The Next defaults.

When the users change *search-engines*, and later change (search-engines
*interface*), then they can revert to *search-engines* (the user
defaults), but not to the Next defaults.

vindarel:
We can explain to change `(search-engines *interface*)` preferably. If he changes `*search-engines*`, no big deal, he can restart Next. For a sign, we can also not export the parameters, nor document it.

Ambrevar:
> For a sign, we can also not export the parameters, nor document it.

I don't understand this, what do you mean?

Ambrevar:
I suggest naming `*search-engines*`I suggest renaming `*search-engines*` to `*search-engines-default*` or `*default-search-engines*`, to make it clear to the user that it won't change the values of a running instance.

I'm wondering how much this approach could confuse the user.

Some slots would have a default defparameter, others wouldn't.

Many users start hacking their config by first copying snippets from the Internet and guessing the rest from there.  This is why this approach add more confusion: users would have to guess which variable has a default defparameter and which one does not.  (Not all users have a Common Lisp-aware editor with completion.)

Regarding the compile-time type checking:

- We would have it only for the default defparameters.

- Note that the default are often set at the top-level of the init.lisp, so in effect it's the same as run-time type checking when init.lisp is loaded.

Ambrevar:
I suggest we close this since it's superseded by #484 which was just merged.

vindarel:
my bad for not answering earlier.

Does 484 catch type errors when we load the init file?

On renaming: sounds good, but it may not be necessary.

If we say in the doc: `setf *param*` will take effect when you reload next, and `(set-param …)` (our macro) changes the value of the current session so you can call `load-init-file`, it's enough. 
And, if we want the user to be able to reset a parameter in the current session, why not provide a command? `M-x reset-parameter RET <choose param> RET` will call the `set-param` macro.

Well, I think all slots that could be customized will have a default parameter. 

> Many users start hacking their config by first copying snippets from the Internet and guessing the rest from there. This is why this approach add more confusion:

this brings more safety. If they don't have the config right, they are warned. 
The double naming might bring confusion but it's not necessary.

>  users would have to guess which variable has a default defparameter and which one does not.

they don't guess, they copy snippets or the documentation. 

> Note that the defaults are often set at the top-level of the init.lisp, so in effect it's the same as run-time type checking when init.lisp is loaded.

I don't get what point you're trying to make here.

Ambrevar:
> Does 484 catch type errors when we load the init file?

Just like this PR: it catches them if they are set at the top-level or
if deftype does not use `satisfies`.

> On renaming: sounds good, but it may not be necessary.
>
> If we say in the doc: `setf *param*` will take effect when you reload next, and `(set-param …)` (our macro) changes the value of the current session so you can call `load-init-file`, it's enough. 

What is `set-param`?

> And, if we want the user to be able to reset a parameter in the current
> session, why not provide a command? `M-x reset-parameter RET <choose param>
> RET` will call the `set-param` macro.

I don't know, I'm a bit uneasy with providing a framework to customize only
_some remote-interface slots_.

> Well, I think all slots that could be customized will have a default parameter. 

You mean using the approach of this PR?  As we discussed before, this would
generate a maintenance burden.  Which, in turn, is prone to introduce bugs.

> this brings more safety. If they don't have the config right, they are warned. 
> The double naming might bring confusion but it's not necessary.

This type-checking is only a tiny guarantee, the config can be wrong in many
more ways.

> they don't guess, they copy snippets or the documentation. 

The documentation cannot cover all cases (it's combinatorial), so users
must understand the configuration system to some extent to use it.

If the variable names make sense, it helps the user.
If they can't configure much without copying snippets, this is seriously
limiting the users.

jmercouris:
My two cents:

More advanced users will understand the system and use Lisp to extend it. Others shall use a customization buffer. If the customization in Emacs were better, fewer people would be copy/pasta'ing bits of Elisp into their configs.

jmercouris:
In short, I agree with Pierre

vindarel:
> This type-checking is only a tiny guarantee, the config can be wrong in many
more ways.

They fix issues we are having right now with user configs!

> What is set-param?

I meant (setf (get-defaut)) or whatever the actual macro is.

> Just like this PR: it catches them if they are set at the top-level or
if deftype does not use satisfies.

So I tried the changes, see https://github.com/atlas-engineer/next/pull/484#issuecomment-555971279

the type uses `satisfies. The type error was caught with the changes I proposed, but it is not with slots as presently. We should do better.

Ambrevar:
> They fix issues we are having right now with user configs!

But the fix has some drawbacks, i.e. increased user-facing complexity
and lack of generality.

I believe that while it fixes some issues, it create others, which is
why I'm would not rush to apply it, I'd like to think more.

In the end, improving type-checking in SBCL seems to be the long term,
more viable solution.

>> What is set-param?
>
> I meant (setf (get-defaut)) or whatever the actual macro is.

This is deprecated since it's extremely buggy, so we should not use it anymore.

>> Just like this PR: it catches them if they are set at the top-level or
> if deftype does not use satisfies.
>
> So I tried the changes, see https://github.com/atlas-engineer/next/pull/484#issuecomment-555971279
>
> the type uses `satisfies. The type error was caught with the changes I proposed, but it is not with slots as presently. We should do better.

Your approach only catches type errors _at the top level_, so it also
has some limitations.

vindarel:
> Your approach only catches type errors _at the top level_, so it also
has some limitations.

sure. It's better than nothing. It'd catch a majority of user config issues IMO.

> But the fix has some drawbacks, i.e. increased user-facing complexity
and lack of generality.

I'm surprised by this, because the present configuration system is more involved. I like the new class-based approach, but it increases user-facing complexity IMO.

---

Anyways, the benefits of type checking with global parameters didn't do consensus between us. We could use CLOS machinery, if the benefits outweights the cost (but with CLOS machinery, I doubt it). Good news is that compile time type checks for class slots is landing into SBCL.

Ambrevar:
> I'm surprised by this, because the present configuration system is more involved. I like the new class-based approach, but it increases user-facing complexity IMO.

What I meant here is that it exposes _more_ user facing variables /
config knobs.  So in the sense of total number of user-facing variables,
it increases complexity.

Issue #419, opened 2019-09-28
Functional config & hooks


Ambrevar:
A discussion on the general idea of _configurability_: This is an iteration on
the question of hooks (https://github.com/atlas-engineer/next/issues/383) and
the controversial topic of functional configuration
(https://github.com/atlas-engineer/next/pull/397).

--------------------------------------------------------------------------------

I just had an epiphany when working with the set-url hook:
the handlers take the URL as argument, but not the buffer.  So how do the
handler know which buffer they apply to.  Calling `(current-buffer)` is wrong:
for instance, when we call `reload-buffer` over multiple buffers,
`(current-buffer)` always returns the same buffer, not the buffer being
reloaded.

Possible fixes:

- Add a `buffer` parameter.  But it fits badly with `run-composed-hook` which
  calls the next handler over the return value of the previous.  Even if we pass
  the buffer along, we've got no strong guarantee that no handler is returning a
  different buffer, thus breaking the integrity of the hook.
  Beside, it really feels like we are passing along an information that should
  already be there, because the hook _belongs_ to the buffer.

- Replace the named handler by a closure over the buffer.  This seems right, but
  there is a problem: a closure is anonymous, and hooks don't like lambdas as
  the Emacs experience tells us:

  - Lambdas are not comparable and thus duplicates can't be removed and will
    stack up.

  - They are black boxes to the user, we can't know what they do.

  - It's impractical to remove a lambda interactively since it has no name.

So what if we created a `handler` class with the following slots:

- Name / description, e.g. `handler that sets the search engine to '("foo" "bar")`.
- The function, possibly a lambda.
- `place`: (Optional) If the function is a setter, describe the place being set,
  e.g. `'(remote-interface buffer-make-hook load-hook)`.
- `value`: (Optional) If the function is a setter, describe the value,
  e.g. `'(1 2 3)`.

We would modify `add-to-hook` to either accept functions or `handler`s but not
anonymous functions.
Then `run-hooks` should be able to run both regular functions and
the above `handler`s in a hook.

Handlers should be comparable so that the hooks can remove duplicates.
I can think of various criteria for comparison:

- If function is the same (non-lambda only).
- Or, if place and value are set, if they are `EQUAL`.  This is only useful for
  setters.
- I'm not sure it makes sense to compare the name/description.

--------------------------------------------------------------------------------

In https://github.com/atlas-engineer/next/pull/397, we discussed new
"functional" (stateless) ways to configure Next.
I was suggesting

```lisp
[...]

Issue #383, opened 2019-09-17
Hooks enhancements


vindarel:
So hooks open up for many many cool stuff. Here's a meta issue on what can be improved to work with them.

- as a user/developer trying to extend Next in the REPL, I define hooks and suddenly I ask myself: "what did I just add"? We should have a way to list all existing hooks. 
- some hooks expect a return value, my hook will break if my function accepts say a buffer instead of an url: compilation-time type warnings would help.
- when I break Next with a bad hook, I'd like a way to disable them altogether. For these two, see https://github.com/scymtym/architecture.hooks/issues/3
- currently, redefining a function that was added to a hook doesn't actually update the function of the hook. See https://github.com/scymtym/architecture.hooks/issues/4
- add-to-default-list or (setf) get-default doesn't add the hook to existing objects, pushing to say `(load-hook (active-buffer *interface*))` doesn't add it to the default list: an interface might help (with the double benefit that the user doesn't directly touch the internal data structures: its ini file is more resilient to future changes).

Ambrevar:
If I'm not mistaken, the all the above is now merged on master except for the ability to list all hooks.

jmercouris:
You are correct, ability to list all hooks is missing. However you can easily find the hooks in slime with completion by typing "next:" TAB and then searching for "hook" in the completions.

aartaka:
Closing this. I'm pretty sure our `describe-any`/`describe-slot` can handle that, especially given that all the hooks have a `-hook` suffix in their names.

To implement hook listing as a command/function, one'd need to go through all the slots and variables and find those that are `nhooks:hook`-typed, which not all of the hooks are (let alone the strictness of typing on different implementations). Listing instances of `nhooks:hook` and subclasses may be better, but the best approach to it would be non-portable at best (ﾉ´ｰ`)ﾉ

Ambrevar:
There is another way: in the nhooks library we can tweak the initializer to references all the hooks, which would then be reported by a `nhooks:all-hooks` functions or the like.

But this does not report the bound symbols, if any.  
@aartaka's approach does it however.

Do we want both approaches?

This is a fairly general problem in CL, others must have dealt with it before, nah? 

Issue opened at https://github.com/atlas-engineer/nhooks/issues/14.

aartaka:
Huh, registering hook instances can be a good way there, because then we can check the symbols we find for being bound to these instance!

PR #483, opened 2019-11-05
General-purpose hooks and functional config

Ambrevar:
Work in progress, not done yet.

So I've addressed many of the points mentioned in #419.
Those hooks are based off Serapeum's hooks.  They fix most (all?) issues in #383. 

The biggest issue I'm facing at the moment is type-checking: I'd like to check that the handlers match the type specified in the the `hook` object, but it seems impossible to do in Common Lisp:

https://old.reddit.com/r/Common_Lisp/comments/ds4tpy/programmable_and_compiletime_function_typechecking/?

This is a general issue: how do we validate the type of a variable / slot that's supposed to host a function?

I've implemented some form of global, ahead-of-time hooks that can be bound to any symbol / object.  It's a very naive implementation, I don't know if that would suit everyone's needs.

This implementation makes some parts of Serapeum's hooks irrelevant, such as `run-hook-with-args` and derivatives.

@ruricolist @scymtym What do you think?

### DIFF
--- a/libraries/hooks/hooks.lisp
+++ b/libraries/hooks/hooks.lisp
@@ -208,10 +208,10 @@ To add anonymous functions to HOOK, use
                                   (when (equals f fn)
                                     (setf found-handler f)
                                     t))
-                                handlers)))))
-      (find-handler (handlers hook))
-      (unless found-handler
-        (find-handler (disabled-handlers hook)))
+                                handlers))))
+        (find-handler (handlers hook))
+        (unless found-handler
+          (find-handler (disabled-handlers hook))))
       found-handler)))
 
 (defmethod remove-hook ((hook hook) handler-name)
@@ -223,10 +223,10 @@ To add anonymous functions to HOOK, use
                                   (when (eq handler-name (name f))
                                     (setf found-handler f)
                                     t))
-                                handlers)))))
-      (find-handler (handlers hook))
-      (unless found-handler
-        (find-handler (disabled-handlers hook)))
+                                handlers))))
+        (find-handler (handlers hook))
+        (unless found-handler
+          (find-handler (disabled-handlers hook))))
       found-handler)))
 
 (defmethod run-hook ((hook hook))
--- a/libraries/hooks/hooks.lisp
+++ b/libraries/hooks/hooks.lisp
@@ -241,8 +241,8 @@ This is an acceptable `combination' for `hook'."
 
 ;; TODO: Because our hooks are typed, having both `run-hook` and
 ;; `run-hook-with-args' makes little sense.  Remove `run-hook-with-args'?
-;; (defmethod run-hook-with-args ((hook hook) &rest args)
-;;   (apply (combination hook) hook args))
+(defmethod run-hook-with-args ((hook hook) &rest args)
+  (apply (combination hook) hook args))
 
 ;; TODO: Implement the following methods? Isn't the `combination' slot more general?
 ;; (defmethod run-hook-with-args-until-failure ((hook hook) &rest args)
--- a/libraries/hooks/hooks.lisp
+++ b/libraries/hooks/hooks.lisp
@@ -180,9 +180,13 @@ This is an acceptable `combination' for `hook'."
             (with-hook-restart (apply (fn handler) args)))
           (handlers hook)))
 
-(defmethod combine-hook-until-failure ((hook hook) &rest args) ; TODO: Result difference between "all fail" and "no handlers"?
+(defmethod combine-hook-until-failure ((hook hook) &rest args)
   "Return the list of values until the first nil result.
-Handlers after the successful one are not run.
+Handlers after the failing one are not run.
+
+You need to check if the hook has handlers to know if a NIL return value is due
+to the first handler failing or an empty hook.
+
 This is an acceptable `combination' for `hook'."
   (let ((result nil))
     (loop for handler in (handlers hook)
@@ -191,9 +195,13 @@ This is an acceptable `combination' for `hook'."
           always res)
     (nreverse result)))
 
-(defmethod combine-hook-until-success ((hook hook) &rest args) ; TODO: Result difference between "no success" and "no handlers"?
+(defmethod combine-hook-until-success ((hook hook) &rest args)
   "Return the value of the first non-nil result.
 Handlers after the successful one are not run.
+
+You need to check if the hook has handlers to know if a NIL return value is due
+to all handlers failing or an empty hook.
+
 This is an acceptable `combination' for `hook'."
   (loop for handler in (handlers hook)
      thereis (apply (fn handler) args)))
--- a/libraries/hooks/hooks.lisp
+++ b/libraries/hooks/hooks.lisp
@@ -1,3 +1,5 @@
+(in-package :next-hooks)
+
 (defvar *hook* nil
   "The hook currently being run.")
 
@@ -5,23 +7,23 @@
   (:documentation "Add FN to the value of HOOK.")
   (:method ((hook symbol) fn &key append)
     (declare (type (or function symbol) fn))
-    (synchronized (hook)
+    (serapeum:synchronized (hook)
       (if (not append)
           (pushnew fn (symbol-value hook))
           (unless (member fn (symbol-value hook))
-            (appendf (symbol-value hook) (list fn)))))))
+            (alexandria:appendf (symbol-value hook) (list fn)))))))
 
 (defgeneric remove-hook (hook fn)
   (:documentation "Remove FN from the symbol value of HOOK.")
   (:method ((hook symbol) fn)
-    (synchronized (hook)
-      (removef (symbol-value hook) fn))))
+    (serapeum:synchronized (hook)
+      (alexandria:removef (symbol-value hook) fn))))
 
 (defmacro with-hook-restart (&body body)
   `(with-simple-restart (continue "Call next function in hook ~s" *hook*)
      ,@body))
 
-(defun run-hooks (&rest hooks)
+(defun run-hooks (&rest hooks)          ; TODO: What about running hooks over arguments?
   "Run all the hooks in HOOKS.
 The variable `*hook*' is bound to the name of each hook as it is being
 run."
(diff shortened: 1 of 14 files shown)


Ambrevar:
Left to do:

- doc
- tests
- helper macros, like the setters detailed in #419.
- integrate with Next.

Ambrevar:
@ruricolist In https://github.com/ruricolist/serapeum/issues/41#issuecomment-538606901 you mentioned a trick to check the arity of a function.  I'm not sure I understand how that would work in practice, can you explain?

ruricolist:
The idea I had was that when you add the function to the hook, you get the arity of the hook and then (at runtime) compile a lambda that calls the function with that arity.

Here's what it looks like with SBCL:

``` lisp
(defun x (y z) (list y z))

(compile nil '(lambda (a b c) (x a b c)))
; caught WARNING:
;   The function X is called with three arguments, but wants exactly two.

(let ((fn (lambda (y z) (list y z))))
  (compile nil `(lambda (a b c) (funcall ,fn a b c))))
; caught WARNING:
;   The function (LAMBDA (Y Z)) is called with three arguments, but wants exactly two.
```

Clozure signals a warning for the global definition, but unfortunately not for the local definition.

Ambrevar:
What about just using `declare` then?  SBCL seems to do the right thing,
+ it allows for catching type errors for the return value as well:

```lisp

> (let ((fn (lambda (y z) (list y z))))
				(declare (type (function (integer integer) string) fn))
				fn)
; in: LET ((FN (LAMBDA (Y Z) (LIST Y Z))))
;     (LET ((NEXT::FN (LAMBDA (NEXT::Y NEXT::Z) (LIST NEXT::Y NEXT::Z))))
;       (DECLARE (TYPE (FUNCTION (INTEGER INTEGER) STRING) NEXT::FN))
;       NEXT::FN)
; 
; caught WARNING:
;   Derived type of (LAMBDA (Y Z)) is
;     (FUNCTION (T T) (VALUES CONS &OPTIONAL)),
;   conflicting with its asserted type
;     (FUNCTION (INTEGER INTEGER) (VALUES STRING &REST T)).
;   See also:
;     The SBCL Manual, Node Handling of Types
; 
; compilation unit finished
;   caught 1 WARNING condition
#<FUNCTION (LAMBDA (Y Z)) {539C099B}>

```

Ambrevar:
Actually I have a problem with `declare`: how do we get the type
programmatically from the hook?

It seems that we may only be able to do this if `add-hook` were a macro.
Thoughts?

jmercouris:
It occurred to me that we could use a macro called set-default or something which expands into a hook that runs after initialization, I’m much more of the opinion that compile time checking and slots is not important, the slot set function can do the checking. This is much more towards proper encapsulation of OO data structure anyways. I know this is not strictly related to only the contents of this pull request, but I can’t think of anywhere else it belongs.

Ambrevar:
I think it belongs to #419 where I've offered a non-hook solution ;)

Ambrevar:
Great news!  I've figured a way to use macros to generate typed hooks and typed handlers, and now `add-hook` will report a warning if called over the wrong combination of hook / handler.

The only drawback: The types must be declared before use.

In practice, it looks like this:

```lisp
(define-hook-type string->string (function (string) string)) ; Need to be done only once for this type.

;; Then
(defvar *my-hook* (make-instance 'hook-string->string))

(defun my-settings1 ()
  (add-hook *foo* (make-handler-string->string (lambda (s) (str:concat "foo-" s)) :name 'foo)

(defun my-settings2 ()
  (add-hook *foo* (make-handler-string->string (lambda (s) 17) :name 'foo)))

; caught WARNING:
;   Derived type of (LAMBDA (S) :IN MY-SETTINGS2) is
;     (FUNCTION (T) (VALUES (INTEGER 17 17) &OPTIONAL)),
;   conflicting with its asserted type
;     (FUNCTION (STRING) (VALUES STRING &REST T)).
```

What do you think?  @ruricolist ?

Ambrevar:
In particular, the above approach may supersede @ruricolist arity-checking technique since it non only checks the arity but also the argument and return value types.

jmercouris:
Is there anything yet blocking us from merging this in?

Ambrevar:
Yes, everything mentioned at the top:
- Replace cl-hooks with this.
- Document.
- Add tests.

I'm on it.

Ambrevar:
Update:
- I've integrated this library with Next.  Let me know if you like the new API.  I've remove cl-hooks.
- I wrote tests.

Left to do:
- Documentation.

Questions left:

- Should we add a facility to list all known hooks?  This could be useful considering they are user-facing.  But then we need the add a function to _delete_ a hook or else temporary hook won't be garbage collected.  I can't think of a use case for temporary hooks though.

- For now we can create hooks with `(make-instance 'hook-TYPE :combination ...)`.  Do you think it's useful to add a constructor like `(make-hook 'hook-TYPE ...)`?

- The disable-hook function can be used to move all handlers to the list of disabled-handlers in the hook.  Those won't be executed by run-hook, but they can be reactivated later.  Do you think it would be useful to add a `disable-handler` function to selectively disable handlers from a hook?  Maybe better: add a list argument to `disable-hook` that lists which handlers to disable.  This way we don't need to add a new function.

- For now I'm exposing `make-handler` which allows for creating untyped handlers.  But untyped handlers are risky and I don't think it's useful at all considering we have `make-handler-any` which accepts functions of any type.

- The serapeum hook library comes with both `run-hook` and `run-hook-with-args`.  For typed hooks there is no difference because the presence of argument is decided by the hook itself.  So I'd like to use `run-hook-with-args` only, but the name is a bit unfortunate then.  Ideally, it would simply be `run-hook`.

- Similarly, `run-hook-with-args-until-failure` and `run-hooks-with-args-until-success` are not needed for the hook class since it has a `combination` slot which determines how the handlers are run.

- What value to we return when we run `run-hook` on a hook with no handler?  `NIL` can conflict with the return value of the handlers.

@ruricolist  @vindarel @jmercouris ?

libraries/hooks/hooks.lisp
```diff
@@ -342,6 +346,7 @@ HANDLER must be of type ~a."
          (add-hook-internal hook handler :append append)))))
 
 ;; TODO: Add make-hook-* function?
+;; TODO: Allow listing all the hooks?
```
jmercouris: Yes, I think this would be valuable

Ambrevar:
I've briefly documented the hooks.

I have a partial answer for my question about the return values of hooks.
Actually, the return value is determined by the combination function.

combine-composed-hook should return its arguments untouched when there is no
handler.
To see why consider the following call:

```lisp
(setf url (next-hooks:run-hook-with-args (load-hook buffer) url))
```

The above does not need to check if load-hook has handlers and we can safely
assumed the above will be a no-op if load-hook has no handlers.

For default-combine-hook, it's simple: we return the empty list when there is no
handler.  This cannot be confused with a handler returning NIL, because then the
return value of `run-hook` is `(NIL)`.

I'm less clear about 
- `combine-hook-until-failure`: How do we distinguish between "all failed" and
"there was no handlers"?
- `combine-hook-until-success`: How do we distinguish between "no success" and
"there was no handlers"?

jmercouris:
Hello Pierre!

Lots and lots of very good questions here. I'm not sure that I can answer all of them. I would like to go over this together and make sure I understand. Below, based on my cursory reading are some of my thoughts:

> * Should we add a facility to list all known hooks?  This could be useful considering they are user-facing.  But then we need the add a function to _delete_ a hook or else temporary hook won't be garbage collected.  I can't think of a use case for temporary hooks though.

I think that a facility to list known hooks would be very valuable to ensure that there aren't duplicate hooks (as a user). It would also serve as feedback (as a user), that a hook has been successfully added (or deleted).

> * For now we can create hooks with `(make-instance 'hook-TYPE :combination ...)`.  Do you think it's useful to add a constructor like `(make-hook 'hook-TYPE ...)`?

I think make-instance suffices more than enough. In my opinion, there is no checking we have to do upon creation that would justify creating a function.

> * The disable-hook function can be used to move all handlers to the list of disabled-handlers in the hook.  Those won't be executed by run-hook, but they can be reactivated later.  Do you think it would be useful to add a `disable-handler` function to selectively disable handlers from a hook?  Maybe better: add a list argument to `disable-hook` that lists which handlers to disable.  This way we don't need to add a new function.

Perhaps I don't understand, but couldn't one disable a handler by deleting it? Would the point of disabling and re-enabling handlers be to keep some state in the handler perhaps without getting rid of it? If so, then I think the user could just as easily keep a reference to it if they really need it. 

> * For now I'm exposing `make-handler` which allows for creating untyped handlers.  But untyped handlers are risky and I don't think it's useful at all considering we have `make-handler-any` which accepts functions of any type.

If you don't do it, some user will do it eventually :-)
We can make the API as strict as we want, but people will make a library or copy paste a snippet for sure that adds `make-handler`. Better that we add it, than it get's copy pasted into thousands of init files :-D

> * The serapeum hook library comes with both `run-hook` and `run-hook-with-args`.  For typed hooks there is no difference because the presence of argument is decided by the hook itself.  So I'd like to use `run-hook-with-args` only, but the name is a bit unfortunate then.  Ideally, it would simply be `run-hook`.

I see no problem with just making it `run-hook`. We don't say `funcall-with-args`, we simply say `funcall`. Since there is no distinction here in our codebase, we do not need it. Thoughts?

> * What value to we return when we run `run-hook` on a hook with no handler?  `NIL` can conflict with the return value of the handlers.

I'm not sure I understand this question. I may be missing too much context. Can you please explain it?

libraries/hooks/hooks.lisp
```diff
@@ -0,0 +1,356 @@
...
+
+;; TODO: Maybe don't export make-handler, so that user is forced to use the
+;; typed versions.
+(defun make-handler (fn &key (class-name 'handler) name handler-type place value)
+  "NAME is a symbol."
+  (setf name (or name (let ((fname (nth-value 2 (function-lambda-expression fn))))
+                        (when (typep fname 'symbol)
+                          fname))))
+  (unless name
+    (error "Can't make a handler without a name"))
```
vindarel: (There's also the `:initform (error "this slot is required")` trick that can be used along with make-instance, in case make-handler is not used).

vindarel:
``` 
    combine-hook-until-failure: How do we distinguish between "all failed" and
    "there was no handlers"?
    combine-hook-until-success: How do we distinguish between "no success" and
    "there was no handlers"?
```

Either catch the absence of handlers early, or throw a condition?

yes `disable-hook` is valuable.

vindarel:
> For now we can create hooks with (make-instance 'hook-TYPE :combination ...). Do you think it's useful to add a constructor like (make-hook 'hook-TYPE ...)?

I do, generally I like this style for its shortness and the encapsulation. And is it not needed? In make-hook you check that a name is given. The argument for a constructor?

(And what about even more precise constructors, like `(make-string->string-hook …)` for the most used ones?)

Ambrevar:
> I think that a facility to list known hooks would be very valuable to ensure
> that there aren't duplicate hooks (as a user). It would also serve as feedback
> (as a user), that a hook has been successfully added (or deleted).

I think you mixed up the terms here.  The `hook` is the value that holds a list
of handlers.  So it's the handler that is added to the hook.

Note that handlers cannot be duplicated with my typed-hook implementation.

My question was about listing hooks, not handlers.

> I think make-instance suffices more than enough. In my opinion, there is no
> checking we have to do upon creation that would justify creating a function.

Actually there is checking that could be done:

- `:combination` must be a function.
- `handlers` must be handlers.

> Perhaps I don't understand, but couldn't one disable a handler by deleting it?
> Would the point of disabling and re-enabling handlers be to keep some state in
> the handler perhaps without getting rid of it? If so, then I think the user
> could just as easily keep a reference to it if they really need it.

Keeping track of the handler has at least 2 use cases:

- It makes it easy to toggle handlers on/off for testing, say, when a handler
misbehaves.
- A handler may be a closure and thus hold internal state.  Thus if the closure
is deleted from the hook, there is no way to recover it later.

The user could very well keep track of the handlers themselves, but then all
users interested in this feature
would have to re-implement something that this library could have provided.

> If you don't do it, some user will do it eventually :-) We can make the API as
> strict as we want, but people will make a library or copy paste a snippet for
> sure that adds `make-handler`. Better that we add it, than it get's copy pasted
> into thousands of init files :-D

It's Common Lisp, un-exported symbols can always be accessed with `::`, so there
won't be any copy-paste anyways.

The question was whether we tell the user "this is OK to use" or not.

> I see no problem with just making it `run-hook`. We don't say
> `funcall-with-args`, we simply say `funcall`. Since there is no distinction here
> in our codebase, we do not need it. Thoughts?

`run-hook` and `run-hook-with-args` are generic functions, so I cannot change
their arguments.  In particular, I cannot pass extra arguments to `run-hook`
which is what I would like to do here for typed hooks.

Some options:

- Change Serapeum's default.
- Stick to `run-hook-with-args`.

> I'm not sure I understand this question. I may be missing too much context. Can
> you please explain it? Edit: just to be clear, I've also read your subsequent
> follow-up, and I still don't quite understand.

When run, a hook may return a value.
But the call site may want to make the distinction between the return value of
the handlers and what is returned when the hook has no handlers.

Besides, for some combination we want to adopt a specific behaviour.
See the example with the load-hook: if it has no handler, it should return the
URL unchanged.

Does that make more sense?

jmercouris:
> I think you mixed up the terms here. The `hook` is the value that holds a list
> of handlers. So it's the handler that is added to the hook.

You are correct, I had mixed up the terms in my head for that paragraph for some reason. I still think a function to list hooks would be useful as part of the discoverability and interactivity of the platform. I imagine we could bind a command like `list-hooks` or something so that people can search for them. 

> Note that handlers cannot be duplicated with my typed-hook implementation.

How so? Could I not add two lambda handlers that do the same thing?
 
> > I think make-instance suffices more than enough. In my opinion, there is no
> > checking we have to do upon creation that would justify creating a function.
> 
> Actually there is checking that could be done:
> 
> * `:combination` must be a function.
> * `handlers` must be handlers.

I think we should allow people to shoot themselves in the foot. I have consistently been against adding in more checking of user valid input. The reason is:

1. It adds overhead to making changes
2. It "freezes" the code at a specific point

It is too soon to do either of those things. Checking types, at this point, for most users, simply does not add as much value as a new feature. In other words, for the time being, I believe our efforts are better spent elsewhere.

> The user could very well keep track of the handlers themselves, but then all
> users interested in this feature
> would have to re-implement something that this library could have provided.

I guess I don't see the purpose here exactly, but if you think it would be useful instead of having every user keep a reference to whatever they are testing, why not!

> The question was whether we tell the user "this is OK to use" or not.

Ah, yes, export the symbol.

> Some options:
> 
> * Change Serapeum's default.
> * Stick to `run-hook-with-args`.

I should think that Serapeum should change, @ruricolist ?

> When run, a hook may return a value.
> But the call site may want to make the distinction between the return value of
> the handlers and what is returned when the hook has no handlers.
> 
> Besides, for some combination we want to adopt a specific behaviour.
> See the example with the load-hook: if it has no handler, it should return the
> URL unchanged.
> 
> Does that make more sense?

That makes sense now, I believe. I however, do not have any ideas... I will think about it :-)

Ambrevar:
Yes listing the hooks would be useful, but the question is how do we list them?
It's a general problem: how do we list variables in a programming language?

- What's the name of the hook?  If we don't have a name, then the listing is
  mostly meaningless.
- How do we store the list?  How do we make sure we are not missing hooks?
- Can we remove hooks?

> How so? Could I not add two lambda handlers that do the same thing?

See the docstrings, the tests and this thread.  Lambdas are only accepted with
the extra name argument, so they can be compared (see the `equals` function).

> I think we should allow people to shoot themselves in the foot.

I think you are talking about a different thing, which is a different problem.
Users still can call `make-instance` anytime.  The question here is whether we
add helper functions (constructors) to provide some extra checks and shorter code.

> Ah, yes, export the symbol.

I'm leaning towards not exporting it.  The main intent of this library is to expose
typed hooks after all, which are much more powerful than untyped hooks.

jmercouris:
> See the docstrings, the tests and this thread. Lambdas are only accepted with
> the extra name argument, so they can be compared (see the `equals` function).

Yes, I saw, but I mean two lambdas, two names, do the same thing. I don't believe this can be prevented :-D

> > I think we should allow people to shoot themselves in the foot.
> 
> I think you are talking about a different thing, which is a different problem.
> Users still can call `make-instance` anytime. The question here is whether we
> add helper functions (constructors) to provide some extra checks and shorter code.

I'm not talking about a different thing. I am referencing a different thing, but is the same idea here. I don't believe constructors with extra checks are necessary.

> > Ah, yes, export the symbol.
> 
> I'm leaning towards not exporting it. The main intent of this library is to expose
> typed hooks after all, which are much more powerful than untyped hooks.

Well...as you wish :-D I don't have a very strong opinion on this.

ruricolist:
I've changed the signature of serapeum:run-hooks to allow arguments.

Is there anything else you need from me for this?

Ambrevar:
@ruricolist: I think you meant `run-hook` :)
Your change looks good, but haven't you broken backward compatibility in the process?

libraries/hooks/hooks.lisp
Ambrevar: Nice tip, didn't know about it, thanks!
Sadly it won't work here because the name can be guessed from the named function (that is, when we don't pass a lambda) and thus it should not fail for named functions.

Ambrevar:
@vindarel 

> Either catch the absence of handlers early, or throw a condition?

An empty hook is a valid hook.  `run-hook` should not fail for empty hooks.  See Emacs for instance.

> yes disable-hook is valuable.

It is, but the question is about handlers: do we need to add a `disable-handler` function, or add a parameter to `disable-hook`?

Ambrevar:
@vindarel 

> I do, generally I like this style for its shortness and the encapsulation. And is it not needed? In make-hook you check that a name is given. The argument for a constructor?

Hooks don't have names.  Did you mean something else?

> (And what about even more precise constructors, like (make-string->string-hook …) for the most used ones?)

Indeed, this is what I had in mind: generate constructors in the `define-hook-type` macro.

vindarel:
> Hooks don't have names. Did you mean something else?

I mixed up with handlers and make-handler.

Ambrevar:
I believe this is ready to merge.  I'll merge tomorrow if there is no objection.

jmercouris:
No objection, please feel free to merge!

jmercouris:
May I merge this in on your behalf?

Ambrevar:
I'll do it as soon as possible, no worries!

jmercouris:
OK, thank you! I was just asking because I want to do quite a few file operations (renaming base.lisp to start.lisp etc) in preparation for our new work and I want the merges to be nice and easy!

Ambrevar:
Merged.

PR #484, opened 2019-11-07
CLOS-style configuration

### LINKED ISSUE #419: Functional config & hooks

Ambrevar:
This should mostly fix #419.

Example init.lisp:

```lisp
(defclass my-interface (remote-interface)
  ((download-directory :initform "~/temp")))

(setf *remote-interface-class* 'my-interface)
```

That's it!
Note that we can combine class settings by using multiple inheritance.

### DIFF
--- a/source/base.lisp
+++ b/source/base.lisp
@@ -189,8 +189,8 @@ useful when Next starts up)."
 If INTERACTIVE is t, allow the debugger on errors. If :running, show an error but don't quit the lisp process.
 "
   (with-result (file-name-input (read-from-minibuffer
-                                 (make-instance 'minibuffer
-                                                :input-prompt "Load file:")))
+                                 (make-minibuffer
+                                  :input-prompt "Load file:")))
     (load-lisp-file file-name-input :interactive :running)))
 
 (define-command load-init-file (&key (init-file (init-file-path))
--- a/source/base.lisp
+++ b/source/base.lisp
@@ -234,10 +234,10 @@ Finally, run the `*after-init-hook*'."
     (when *interface*
       (kill-interface *interface*)
       ;; It's important to set it to nil or else if we re-run this function,
-      ;; (make-instance 'remote-interface) will be run while an existing
+      ;; (make-instance *remote-interface-class*) will be run while an existing
       ;; *interface* is still floating around.
       (setf *interface* nil))
-    (setf *interface* (make-instance 'remote-interface
+    (setf *interface* (make-instance *remote-interface-class*
                                      :non-interactive non-interactive
                                      :startup-timestamp startup-timestamp))
     ;; Start the port after the interface so that we don't overwrite the log when
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -135,11 +135,11 @@ This can be useful to let the user select no tag when returning directly."
   "Bookmark the URL of BUFFER."
   (if (url buffer)
       (with-result (tags (read-from-minibuffer
-                          (make-instance 'minibuffer
-                                         :input-prompt "Space-separated tag(s):"
-                                         :multi-selection-p t
-                                         :completion-function (tag-completion-filter :with-empty-tag t)
-                                         :empty-complete-immediate t)))
+                          (make-minibuffer
+                           :input-prompt "Space-separated tag(s):"
+                           :multi-selection-p t
+                           :completion-function (tag-completion-filter :with-empty-tag t)
+                           :empty-complete-immediate t)))
         (when tags
           ;; Turn tags with spaces into multiple tags.
           (setf tags (alexandria:flatten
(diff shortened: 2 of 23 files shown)


Ambrevar:
As opposed to what I said in #419, we don't have proper type-checking over `*window-class*`, etc.
We could enable it with the following:

```lisp
(defun window-class-symbol-p (class-symbol)
  (closer-mop:subclassp (find-class class-symbol)
                        (find-class 'window)))
(deftype window-type ()
  `(satisfies window-class-symbol-p))
(declaim (type (window-type) *window-class*))
```

But note that SBCL fails to perform compile-time type-checking on non-top-level symbols because of the `satisfies`.

I guess it's not a big deal.  Question is: should I even bother with the deftype then?

Ambrevar:
As mentioned in #489, mode configuration still is a bit cumbersome.  What about generating a `*foo-mode-class*` symbol that evalutes to `'foo-mode` by default?

This way configuring the class defaults would look like this:

```
(defclass my-web-mode (web-mode)
  (keymap-schemes ...))

(setf *web-mode-class* 'my-web-mode)
```

Thoughts?

jmercouris:
I like the idea, but I feel like we could take it one step further using reader macros to possibly automate that whole process for the user... I need to think a little bit about it. As an aside, I am very much liking the CLOS style configuration, I think it is a fantastic idea!

Ambrevar:
Macros or reader-macros?  What do you have in mind?

jmercouris:
I have reader macros in mind, I was thinking of a special syntax for people who are new, something to make things simple for them. On the other hand, maybe that is more suited to a `*customize*` buffer of sorts, and we only need provide an SEXP "API".

Ambrevar:
I have implemented the same configuration approach for modes.

Now mode defaults can be configured by define a subclass and then setting the
default class, e.g.

``` common-lisp
(setf *web-mode-class* 'my-web-mode)
```

jmercouris:
Nice!

Ambrevar:
Merged.
@vindarel: Let me know if there is anything you'd like to revert, no problem.

I'll update the documentation just now.

vindarel:
> should I even bother with the deftype then?

yes if it catches errors when loading the init file.

vindarel:
I tried the example from the doc:

```
(defclass my-remote-interface ()
  ((search-engines :initform
    '(("default"
       "https://duckduckgo.com/?q=~a"
       "https://duckduckgo.com/")
      ("yt"
       "https://www.youtube.com/results?search_query=~a"
       "https://www.youtube.com/")
      ("wiki"
       "https://en.wikipedia.org/w/index.php?search=~a"
       "https://en.wikipedia.org/")))))

(setf *remote-interface-class* 'my-remote-interface)
```

and I get:

```
Error: we could not load the init file because of a type error.:
#<TYPE-ERROR expected-type: (SATISFIES REMOTE-INTERFACE-TYPE-P)
             datum: MY-REMOTE-INTERFACE>
```

on the setf.

btw, if I'm not mistaken the type checks don't check slots types?

```
(defmacro define-class-type (class-sym)
  "Define a type named CLASS-SYM-type.
An object of this type is a subclass of CLASS-SYM."
  (let ((type-pred (intern (str:concat (string class-sym) "-TYPE-P")))
        (type-fun (intern (str:concat (string class-sym) "-TYPE"))))
    `(progn
       (defun ,type-pred (class-symbol)
         (closer-mop:subclassp (find-class class-symbol)
                               (find-class ',class-sym)))
       (deftype ,type-fun ()
         '(satisfies ,type-pred)))))
```

vindarel:
and I confirm for the missing type checks:

```
(defclass my-remote-interface (remote-interface)
  ((search-engines :initform
    '(("default"
       "https://duckduckgo.com/?q=~a"
       "https://duckduckgo.com/")
     ;; wrong:
      (42)))))
```

where search-engines is typed:

```
   (search-engines :accessor search-engines
                   :initform '(("default" "https://duckduckgo.com/?q=~a" "https://duckduckgo.com/")
                               ("wiki" "https://en.wikipedia.org/w/index.php?search=~a" "https://en.wikipedia.org/"))
                   :type alist-of-3tuples-strings
```

```
(setf *remote-interface-class* 'my-remote-interface)
MY-REMOTE-INTERFACE
```

should fail.

PR #491, opened 2019-11-10
next-pyqt fixes

tviti:
This is a resubmission of https://github.com/atlas-engineer/next/pull/448 with the history "cleaned up" a little bit (only my own commits are touched).

Recap of changes:

- Switch to from dbus-python to QtDBus for RPC
- Fixes so that the port can quit gracefully
- Changes to key event handling (should fix shift-modifier and spacebar handling, and I think will help with https://github.com/atlas-engineer/next/issues/191#issuecomment-550558727)
- Port now queues JS calls from core, and holds them till page load complete
- QtWebengine processes are now killed when a buffer is killed (previously they lived on and caused memory usage to balloon over time)


jmercouris:
Hi Taylor, thank you so much for keeping this pull request up, it was very nice to have a visible option for users who wanted to use Qt web engine.

Since we have now merged in the Qt work using cl-webengine (https://github.com/atlas-engineer/cl-webengine), and dbus is no longer supported, I will close this pull request (to avoid confusion). In any case, I would welcome your help and participation on the cl-webengine work, it still has a lot to go, and any help would be appreciated! Please let me know if you are interested, and we could collaborate!

thank you!

tviti:
Hi John,

Sorry for the delayed response on this. Glad to see that dbus has been
removed as a dep, since that will surely help to increase size of the
macOS user-base.

Would love to help out, assuming I even can (my CL experience is just I
have picked up along the way while playing with Next). I guess I should
first get around to updating my config to work with the latest updates,
since I think I'm still running one of the first v1.5.0 commits.

Taylor


John Mercouris writes:

Ambrevar:
You'll find an incomplete list of changes in the CHANGELOG, this should
help with updating the configuration.

Let me know if you have any question!

jmercouris:
No problem, I can help out too! Since I am a macOS user too, I can provide some more specific things. There is also some non-CL things to do that would be very useful. For example, taking the MacPorts WebKitGTK+ port and converting it to also work for Homebrew users.

PR #497, opened 2019-11-26
Clip only first line when getting password from password-store

runejuhl:
Currently all lines from passwords stored in password-store are put into the clipboard buffer. For most purposes only the first line is necessary as this is where the password is stored; any other lines usually contain information related to the secret.

This PR changes `clip-password` for `password-store-interface` to only grab the first line.

I can see that there's currently no handling of errors from `uiop:run-program`, so this would explode on errors from `pass`. `cl-ppcre:split` is fine with a `nil` input, so this PR changes nothing in that regard. I'd be happy to add some error handling around `uiop:run-program`, but I'm not quite sure how to go about it. Do you have any preferences?

### DIFF
--- a/libraries/password-manager/password-pass.lisp
+++ b/libraries/password-manager/password-pass.lisp
@@ -3,8 +3,20 @@
 (defvar *password-store-program* (executable-find "pass"))
 
 (defmethod list-passwords ((password-interface password-store-interface))
-  (let ((raw-list (directory (format nil "~a/**/*.gpg"
-                                     (password-directory password-interface))))
+  (let ((raw-list (directory
+                   (format nil "~a/**/*.gpg"
+                           (password-directory password-interface))
+                   ;; We truncate the root directory so that the password list
+                   ;; resembles the output from `pass list`. To do so, we
+                   ;; truncate `~/.password-store/` in the pathname strings of
+                   ;; the passwords.
+                   ;;
+                   ;; Special care must be taken for symlinks. Say
+                   ;; `~/.password-store/work` points to `~/work/pass`, would we
+                   ;; follow symlinks, we would not be able to truncate
+                   ;; `~/.password-store/` in `~/work/pass/some/password.gpg`.
+                   ;; Because of this, we don't follow symlinks.
+                   :resolve-symlinks nil))
         (dir-length (length (namestring
                              (truename (password-directory password-interface))))))
     (mapcar #'(lambda (x)
--- a/libraries/password-manager/password-pass.lisp
+++ b/libraries/password-manager/password-pass.lisp
@@ -12,9 +12,17 @@
             raw-list)))
 
 (defmethod clip-password ((password-interface password-store-interface) password-name)
-  (clip-password-string (uiop:run-program (list *password-store-program* "show"
-                                                password-name)
-                                          :output '(:string :stripped t))))
+  (clip-password-string
+   ;; The common way to store secret in password-store is to use the first line
+   ;; for the secret; there's no standard for how to encode anything else.
+   ;; Because there's no support in `password` for having additional fields
+   ;; (e.g. username or email address for autofilling input fields), we'll keep
+   ;; it simple and just return the first line of the password file.
+   (first
+    (cl-ppcre:split #\newline
+                    (uiop:run-program (list *password-store-program* "show"
+                                            password-name)
+                                      :output '(:string :stripped t))))))
 
 (defmethod save-password ((password-interface password-store-interface)
                           password-name password)


libraries/password-manager/password-pass.lisp
```diff
@@ -12,9 +12,12 @@
             raw-list)))
 
 (defmethod clip-password ((password-interface password-store-interface) password-name)
-  (clip-password-string (uiop:run-program (list *password-store-program* "show"
-                                                password-name)
-                                          :output '(:string :stripped t))))
+  (clip-password-string
+   (car
```
Ambrevar: `car` -> `first`

libraries/password-manager/password-pass.lisp
```diff
@@ -12,9 +12,12 @@
...
+    (cl-ppcre:split "\\n"
```
Ambrevar: You can use `#\newline` instead of "\\n", it's more readable and portable.

Ambrevar:
Indeed, thanks for catching this!
Can you just fix the few nits I've pointed out?  Thanks!

runejuhl:
> Indeed, thanks for catching this!
> Can you just fix the few nits I've pointed out? Thanks!

You're welcome -- and sure, pushed the changes.

I also added a commit that changes the behavior to not resolving symlinks in `list-passwords`; I've added a detailed explanation in the commit message: https://github.com/atlas-engineer/next/pull/497/commits/cc57810bdcc8c55a10798f7de060637554a62e01

Ambrevar:
This seems nice, but I'm not sure the use case is completely clear to
me.

Could you give a concrete example?

runejuhl:
> This seems nice, but I'm not sure the use case is completely clear to me. Could you give a concrete example?

Sure; I've inserted the text from the commit message below. It's pretty much how I structure my own passwords, which is why I noticed the bug; the passwords in my work pass folder had the first few characters truncated because the path of the resolved symlink was shorter than the `~/.password-store/` prefix.

Resolving symlinks means that we get the full path for the actual file instead
of the path of the symlink. Since `list-passwords` truncates the paths by
droppping the length of the password-store path from password files, this means
that we'll drop arbitrary parts of the path name for symlinked entries.

Consider this tree

~~~
~/.password-store/www/somesite.com.gpg
~/.password-store/www/anothersite.dk.gpg
~/.password-store/work -> ~/work/passwords                            # symlink
~/work/pass/www/mywork.example.org.gpg
~~~

If we truncate all entries with the length of the primary password store
prefix (`~/.password-store/`), we'll get these password entries:

~~~
www/somesite.com
www/anothersite.dk
work.example.org
~~~

...which is not quite what we want. If we don't resolve symlinks we'll start
with this:

~~~
~/.password-store/www/somesite.com.gpg
~/.password-store/www/anothersite.dk.gpg
~/.password-store/work/www/mywork.example.org.gpg
~~~

and end up with this:

~~~
www/somesite.com.gpg
www/anothersite.dk.gpg
work/www/mywork.example.org.gpg
~~~

Ambrevar:
Sorry, I missed the commit message, this makes sense.  However the comment seems
to be misleading: the real issue is with how we truncate the root dir to get the
password entries, isn't it?

If I got it right, could you update the comment to explain the bug?
Sorry for picking on the nits! :)

runejuhl:
> Sorry, I missed the commit message, this makes sense. However the comment seems
> to be misleading: the real issue is with how we truncate the root dir to get the
> password entries, isn't it?
> 
> If I got it right, could you update the comment to explain the bug?
You're talking about [this comment](https://github.com/atlas-engineer/next/pull/497/files#diff-a10b0d99fe850d317f6add5c0866e847R8-R13), right?

Yes, you got it right. As it is right now, I can see no other practical way to truncate the root dir for symlinked folders, but I don't think that would make sense anyway. `pass` shows files under symlinks in "root" tree, which is the behavior with this PR.

Maybe something like

```lisp
;; If we follow symlinks we might end up with a list of secrets that have
;; different root dirs. We'd like to truncate the root dir so that the list
;; resembles the output from `pass list`, but if we resolve symlinks a password
;; under the symlinked folder `~/.password-store/work` pointing to `~/work/pass`
;; would have the resolved path of `~/work/pass/some/password.gpg`, instead of
;; `~/.password-store/work/some/password.gpg`. This means that in order to
;; truncate the root dir we'd need to handle all symlinks under
;; `password-directory` separately.
;;
;; By not resolving symlinks we can avoid the need for any special cases; since
;; all entries will have the same root dir prefix we can simply drop the length
;; of the of the prefix from all entries.
```

...but that might be a bit too verbose and hard to understand too?

> Sorry for picking on the nits! :)

It's all good! They need picking -- and I tend to do the same :wink:

Ambrevar:
Yes, this is better I believe.  I propose the following simplification:

``` common-lisp
;; We truncate the root directory so that the password list
;; resembles the output from `pass list`.
;; To do so, we truncate `~/.password-store/` in the pathname strings of the passwords.
;;
;; Special care must be taken for symlinks.
;; Say `~/.password-store/work` points to `~/work/pass`,
;; would we follow symlinks, we would not be able to truncate
;; `~/.password-store/` in `~/work/pass/some/password.gpg`.
;; Because of this, we don't follow symlinks.
```

Ambrevar:
Perfect, thanks a lot!

jmercouris:
+1 thank you for this fix!

PR #500, opened 2019-11-30
macOS: Fix icon location in appbundle.

tviti:
The icon file was previously being copied to the wrong location in the app
bundle, resulting in the .app having no icon. This PR corrects the location.

I'm pretty sure that somebody pointed this out and posted a patch for in
an issue ticket a while back, which I thought culminated in a PR, but apparently
not.


jmercouris:
Ah, yes, thank you!

PR #501, opened 2019-12-06
Update README.org

equirosa:
As of 2019-07-06, Debian 10 was released as codename "Buster". (It's a bit of a small change but might prevent any confusion pertaining to the Sid name.)


Ambrevar:
Thanks!

PR #505, opened 2019-12-18
Change prompt for set-url command to show if new or current buffer

runejuhl:
Vary the input prompt message to have the text "Open URL in new buffer:" if
using a new buffer, else "Open URL in current buffer:".


Ambrevar:
Good catch, thanks!

PR #510, opened 2020-01-03
Fix slot types in command class

christophejunke:
They used to be keywords. The type error was detected by compiling the code with enough debug/safety levels. In `~/.sbclrc`, I added `(sb-ext:restrict-compiler-policy 'debug 3)` and `(sb-ext:restrict-compiler-policy 'safety 3)`.


Ambrevar:
Good catch, thanks!

PR #512, opened 2020-01-03
Refactor set-url-new-buffer into set-url

### LINKED ISSUE #511: Factor set-url-to-buffer back into set-url*
See https://github.com/atlas-engineer/next/commit/53ad4fad1e1d716782d37b7e9e045b0eb7c83258.
I think we have too many set-url* functions.  I suggest we factor in set-url-to-buffer.

@vindarel ?

vindarel:
fixes #511

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -95,31 +95,27 @@ buffer to the start page."
   (ps:chain document title))
 
 @export
-(defun set-url (input-url &key (buffer (current-buffer))
-                            raw-url-p)
+(defun set-url (input-url &key buffer raw-url-p new-buffer-p)
   "Load INPUT-URL in BUFFER.
 URL is first transformed by `parse-url', then by BUFFER's `load-hook'."
   (let* ((url (if raw-url-p
                   input-url
-                  (parse-url input-url))))
+                  (parse-url input-url)))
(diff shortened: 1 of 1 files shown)


jmercouris:
Looks good! Thanks for the quick turnaround!

One thing I did notice is this: you can check to see if a keyword argument is supplied as an argument to a function or if it is using the default value:

```
(defun foo (&key (a 0) (b 0 b-supplied-p)))
```

in the above case if `b-supplied-p` is non-nil, then you know it has been supplied. This may allow you to simplify the body implementation checking of current-buffer.

Ambrevar:
Looks good but a few nits:

- Document the keyword arguments in the docstring.

- You forgot to update set-url-to-buffer in file-manager-mode.

- Actually, do we really need an extra argument?  Currently there is an
  ambiguity if we just look at the function signature: what happens if
  both BUFFER and NEW-BUFFER-P are set?
  
  What about this instead: Add a `switch-p` keyword argument to
  `make-buffer`, then we can call:

  (set-url "http://foo" :buffer (make-buffer :switch-p t))

It would be more general, simplify the code and less ambiguous.

Ambrevar:
Fixed with #555.

Issue #514, opened 2020-01-13
Adblock example in manual does not work


chrisboeg:
**Describe the bug**

The adblock example [in the manual](https://next.atlas.engineer/documentation#resource-blocking-ad-blocking) does not work. I get an error when starting Next after adding the code.

To make it work, I have to add a (documentation) string as the first argument to the `define-mode` block so it looks something like this:

```
(define-mode my-blocker-mode (next/blocker-mode:blocker-mode)
  "my adblocker mode"
  ((hostlists :initform (list *my-blocked-hosts* next/blocker-mode:*default-host-list*))))
```

**Precise recipe to reproduce the issue**

- Install next from AUR
- Copy [adblock example code](https://next.atlas.engineer/documentation#resource-blocking-ad-blocking) from manual and add it to `~/.config/next/init.lisp`.
- Start next
- next won't start, gives error:

```
Next version 1.4.0
Loading configuration from #P"/home/cb/.config/next/init.lisp"...
 <WARN> [11:32:40] next utility.lisp (setf get-default) -
  The (setf (get-default ...)) and (add-to-default-list ...) forms are deprecated.
While evaluating the form starting at line 11, column 0
  of #P"/home/cb/.config/next/init.lisp":
Error: we could not load the init file
Error while parsing arguments to DEFMACRO NEXT:DEFINE-MODE:
  too few elements in
    (NEXT::MY-BLOCKER-MODE (NEXT/BLOCKER-MODE:BLOCKER-MODE)
     ((NEXT::HOSTLISTS :INITFORM
       (LIST NEXT/BLOCKER-MODE:*DEFAULT-HOST-LIST*))))
  to satisfy lambda list
    (NEXT::NAME NEXT::DIRECT-SUPERCLASSES NEXT::DOCSTRING NEXT::DIRECT-SLOTS):
  exactly 4 expected, but got 3
```
- Modify the adblock code by adding a (documentation) string as the first argument to the `define-mode` block in the example code
- now next will start

**Information**
- OS name+version: Arch Linux (up-to-date)
- Next version (from =M-x next-version= or =next --version=): 1.4.0
- How you installed Next (Guix pack, package manager, build from source): AUR

jmercouris:
Thank you for your report! The documentation string should be indeed optional, it seems we've got a small bug with our macro, we'll get that fixed!

PR #516, opened 2020-01-13
make docstring optional

chrisboeg:
As per issue #514, the example in the manual with `define-mode` does not work since the `docstring` is not actually optional. As @jmercouris mentioned in #514, the `docstring` should be optional, so I have here created a quick PR that makes the `docstring` optional. However it also changes the position of the `docstring`, which I don't know whether is acceptable or not. However since `&optional` arguments have to come at the end of the argument list, I only see two options, this being the first one. The second option is to merge the `docstring` and `direct-slots` arguments into a `&rest` argument and then in the body of the macro definition act on how many arguments are in the `&rest`-list.

I'm inclined to prefer the first option since it's simpler, but I also don't know the goals of the current implementation and whether the position of the `docstring` is important or not. It will probably break some existing code at least to change it like I've done in this PR, but I thought it better to bring this issue up and send the change along, and then a decision can be made in either direction.

I should add that I haven't had any luck with actually testing this code since I'm getting some C++ errors when I'm trying to build the code locally, so this code change is untested, but I'm pretty sure it will introduce bugs since the `docstring`s position is expected is some parts of the codebase as far as I can see.

### DIFF
--- a/source/remote.lisp
+++ b/source/remote.lisp
@@ -164,16 +164,9 @@ platform ports might support this.")
                            :border "1px #C38A22 solid"
                            :font-weight "bold"
                            :padding "1px 3px 0px 3px"
-                           :text-transform "lowercase"
-                           :padding "1px 3px 0px 3px"
                            :text-align "center"
                            :text-shadow "0 3px 7px 0px rgba(0,0,0,0.3)"
                            :border-radius "3px"
-                           ;; Ensure the hint is above all the page elements.
-                           ;; https://developer.mozilla.org/en-US/docs/Web/CSS/z-index
-                           ;; TODO: The highest integer value is non-standard,
-                           ;; so the following might depend on the web renderer.
-                           ;; https://developer.mozilla.org/en-US/docs/Web/CSS/integer#Syntax
                            :z-index #.(1- (expt 2 31))))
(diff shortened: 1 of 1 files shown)


jmercouris:
Hello, thank you for your pull request!

Unfortunately, this is not the ideal way to do things. The position of the docstring should not change in the lambda list, as it would break common expectations.

To see another way to do this, please look in command.lisp to see how it is currently done elsewhere in the codebase:

```
(defmacro define-command (name (&rest arglist) &body body)
  "Define new command NAME.
ARGLIST must be a list of optional arguments.
This macro also define two hooks, NAME-before-hook and NAME-after-hook.
Regardless of the hook, the command returns the last expression of BODY."
  (let ((documentation (if (stringp (first body))
                           (first body)
                           (warn (make-condition
                                  'command-documentation-style-warning
                                  :name name))))
        (body (if (stringp (first body))
                  (rest body)
                  body))
```

as you can see, there is a check to see if there is a docstring or not, and a warning emitted when it is missing.

chrisboeg:
Yes, indeed, that what was I was thinking.

I'll see if I can get something like that working, but I first have to get my local build of next working :)

PR #519, opened 2020-01-15
fix so `define-mode`s docstring is optional

chrisboeg:
`define-mode`s docstring should be optional but it isn't. This PR fixes that.

Reference issue: #514

### DIFF
--- a/source/remote.lisp
+++ b/source/remote.lisp
@@ -164,16 +164,9 @@ platform ports might support this.")
                            :border "1px #C38A22 solid"
                            :font-weight "bold"
                            :padding "1px 3px 0px 3px"
-                           :text-transform "lowercase"
-                           :padding "1px 3px 0px 3px"
                            :text-align "center"
                            :text-shadow "0 3px 7px 0px rgba(0,0,0,0.3)"
                            :border-radius "3px"
-                           ;; Ensure the hint is above all the page elements.
-                           ;; https://developer.mozilla.org/en-US/docs/Web/CSS/z-index
-                           ;; TODO: The highest integer value is non-standard,
-                           ;; so the following might depend on the web renderer.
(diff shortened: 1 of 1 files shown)


chrisboeg:
Come to think of it, there should probably be something other than an "empty" string if there is no documentation string provided...

jmercouris:
Hi, thank you for this fix. To answer your question, no an empty string is not what we want to do. We would really like to emit a warning, and store NIL, or nothing as the documentation.

jmercouris:
This is because NIL means something much different than an empty string. You can use NIL as a type of flag too see that something is unbound (if you are consistent about it).

chrisboeg:
@jmercouris Makes sense. I'll take a look at how warnings are done in the code base, and then see about simply not setting a documentation string when no one is provided and then update the PR.

jmercouris:
Thank you!

chrisboeg:
I initially wanted to only have a single `defclass` as is the case in the old version, and then do something inside it like this:

```
,(when docstring (:documentation ,docstring))`
```

but since `when` returns a `nil` on the "off-case", and I couldn't figure out a way of making that not happen, or any other way of achieving something like this, I ended up just creating the *big* `if`-condition from line 38 that either writes the class definition with the docstring if there is one, or without one if there isn't, which isn't a solution I'm thrilled about, but it's the best I've got at the moment.

jmercouris:
I understand, maybe you could have a list of things that you unpack into the CLASS-OPTION where you add the :documentation and the string IF it exists. I would like Pierre's thoughts on this as well. I am however OK with this (@Ambrevar).

chrisboeg:
I'm gonna try your idea and get back later about it.

chrisboeg:
Was it something like what the newest commit does that you were imagining, @jmercouris?

PR #521, opened 2020-01-17
Define mode optional docstring

chrisboeg:
This is the same change we have discussed at #519, but because I needed to move this change out into another branch so I can work on another feature at the same time, #519 was automatically closed when I rolled my master branch brack to upstream/master, so I'm just reopening the PR here.

### DIFF
--- a/source/mode.lisp
+++ b/source/mode.lisp
@@ -5,10 +5,14 @@
 (annot:enable-annot-syntax)
 
 @export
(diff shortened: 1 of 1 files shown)


Ambrevar:
I haven't tested but this looks good to me.

jmercouris:
I just tested, looks good to me! Thank you very much for this fix! The commits tell a nice little story about how the code evolved and how we came to a good solution.

PR #522, opened 2020-01-17
add command to delete all other buffers

chrisboeg:
I have tried implementing the "delete-other-buffers" command you mentioned on IRC, @jmercouris.

I know that you wished to have it called "kill-other-buffers", but since all the "kill-" commands are about lines, and all the "buffer deletion commands" use the term "delete", I decided to use "delete-other-buffers". If you still wish to call it "kill-other-buffers", it can be arranged.


jmercouris:
Works perfect, and the name is good! Thank you, merging in!

PR #525, opened 2020-01-18
Multi buffer search

chrisboeg:
So here's a working (but not finished) implementation of search over multiple buffers, as we talked about, @jmercouris.

It uses the same interface as `delete-buffer` for choosing which buffers to do the search on, and then it shows the minibuffer as per the screenshot below where all search results - across all buffers - are presented. You can choose a match with the return key, or "go back and forth" through the matches with "C-s" like with single-buffer search, but with the buffer getting changed as needed. The "N:" prefix to the left of a match's identifier is the index of the buffer.

multi-buffer-search

So other than the code changes, I also think it's worth discussing how the interface of this feature should actually look. This is just what I came up with, but I can imagine you also have some ideas and/or wishes, @jmercouris, so this is an open invitation to that.

There's also a bug that I'm not yet sure how to deal with. There's a [comment in the code about it here](https://github.com/chrisboeg/next/blob/multi-buffer-search/source/search-buffer.lisp#L86-L88). The deal is that when next is about to focus a search-match, if the match is in another buffer than the currently visible one, it needs to switch to it before it can focus it. The way the code is currently, these two actions are not synced. The focus "attempt" happens and fails before the buffer-switch has happened. I'm not sure what approach to take here, if there is some primitive built in to next that can be used or something like that.

### DIFF
--- a/source/search-buffer.lisp
+++ b/source/search-buffer.lisp
@@ -81,11 +81,6 @@ character-preview-count)))."
     (ps:chain -j-s-o-n (stringify *matches*))))
 
 (define-parenscript focus-match (match)
-  (ps:lisp (when (not (equal (buffer match) (current-buffer)))
-             (set-current-buffer (buffer match))))
-  ;; TODO: scroll doesn't work properly because it seems it tries to scroll
-  ;; before switching buffer in cases where the buffer should be switched
-  ;; so have to find a way to sync that up
   (let* ((rel-identifier
            (ps:lisp
             (let ((id (identifier match)))
@@ -94,6 +89,11 @@ character-preview-count)))."
                   id))))
          (element (ps:chain document (get-element-by-id rel-identifier))))
     (ps:chain element (scroll-into-view t))))
+
+(defun handle-match (match)
+  (when (not (equal (buffer match) (current-buffer)))
+    (set-current-buffer (buffer match)))
+  (focus-match :match match))
 
 (defclass match ()
   ((identifier :accessor identifier :initarg :identifier)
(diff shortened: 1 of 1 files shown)


jmercouris:
Hi Chris,

thank you for this pull request! This is a super cool feature and I am very excited that we now have it!

With regards to your first question about interface, I think it looks perfect for a first version! If you wish, the only change I would make is this:

"buffer number:result number ...match text... buffer title"

In practice:

"1:3 ...goldfish are golden... Wikipedia: All about Goldfish"

In order to achieve this you edit the `object-string` method that specializes on `match` objects, in other words: `(defmethod object-string ((match match))`. The product of this method is what is displayed in the Minibuffer.

The reason we do not use simply `(print object)` and `(object-string method)` is for a reluctance in changing the behavior of the Lisp printer and Reader. This will help clean up adding concatenated data into the `identification` slot of the `match` object. In other words you will not have to do:
```
  (let* ((rel-identifier
           (ps:lisp
            (let ((id (identifier match)))
              (if (stringp id)
                  (cadr (str:split ":" id))
                  id))))
```

To fix the bug of the buffer switching and not focusing in the right spot, I did the following:
```
modified   source/search-buffer.lisp
@@ -81,11 +81,6 @@ character-preview-count)))."
     (ps:chain -j-s-o-n (stringify *matches*))))
 
 (define-parenscript focus-match (match)
-  (ps:lisp (when (not (equal (buffer match) (current-buffer)))
-             (set-current-buffer (buffer match))))
-  ;; TODO: scroll doesn't work properly because it seems it tries to scroll
-  ;; before switching buffer in cases where the buffer should be switched
-  ;; so have to find a way to sync that up
   (let* ((rel-identifier
            (ps:lisp
             (let ((id (identifier match)))
@@ -187,8 +182,11 @@ provided buffers."
     (define-key :keymap keymap "C-s"
                 #'(lambda ()
                     (when (completions minibuffer)
-                      (focus-match :match (nth (completion-cursor minibuffer)
-                                               (completions minibuffer))))))
+                      (let ((match (nth (completion-cursor minibuffer)
+                                        (completions minibuffer))))
+                        (when (not (equal (buffer match) (current-buffer)))
+                          (set-current-buffer (buffer match)))
+                        (focus-match :match match)))))
     (with-result (input (read-from-minibuffer minibuffer))
       (focus-match :match input))))
```

What was happening was the `PS:LISP` form was not getting evaluated in time (out of order) making the focus match function be called on the wrong buffer. I simply switched it to put it in the Lambda body. The same will have to be done before the line: `(focus-match :match input))))`. Probably you will want to reorganize things a little bit :-)

Please let me know if that all makes sense, and once again thank you!

chrisboeg:
Thank you for the comment, tips, and code, @jmercouris. I've incorporated your fix, added the change to how the matches are shown in the minibuffer, and cleaned up the code a bit afterwards as well as merged the newest stuff from upstream/master in.

jmercouris:
Looks good to me! I will merge this in! The only thing a user *may* want is a way to clear search hints from all searched buffers. However, that is what I see as an additional feature.

The reason I see it as an additional feature is that we don't want to automatically clear the search hints when the user presses RET as this might make them lose sight of what they are searching for.

Thank you very much for this wonderful feature!

Issue #524, opened 2020-01-18
Choosing multiple buffers when deleting: can't choose all


chrisboeg:
**Version**: current state of the master branch as per right now

I've notified that if you use the "choose multiple buffers" feature that's available if you for example run the `delete-buffer` functions and then use the `C-Space` shortcut, then if I fire that shortcut until I get down to the last buffer in the list, it will show the number of selected buffers as n-1, and only if I press my up-arrow key will it show the last buffer as selected.

Am I using the feature wrong, or is there a bug?

chrisboeg:
Closing because the issue has been fixed in #526.

PR #526, opened 2020-01-18
Always update display after toggling mark

chrisboeg:
This fixes the bug I filed at #524 about not being able to choose all buffers with C-Space.

Turned out the problem is that `minibuffer-toggle-mark` relies on `select-next` to `update-display`, but if there is no items left to select, `select-next` doesn't update the display. This patch makes `minibuffer-toggle-mark` itself request an update of the display and thus fixes the bug.

### DIFF
--- a/source/minibuffer.lisp
+++ b/source/minibuffer.lisp
@@ -895,6 +895,7 @@ Only available if minibuffer `multi-selection-p' is non-nil."
         (match (member candidate marked-completions)
           ((guard n n) (setf marked-completions (delete candidate marked-completions)))
           (_ (push candidate marked-completions)))))
+    (update-display minibuffer)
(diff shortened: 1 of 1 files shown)


jmercouris:
Looks good, thank you! This is a good fix!

To any future reader wondering why not just pull out the UPDATE-DISPLAY from the WHEN in SELECT-NEXT: consider that they are two discrete operations that update the display and will not necessarily always be chained together- the definition of MINIBUFFER-TOGGLE-MARK may be changed to remove SELECT-NEXT.

A way to potentially update the display multiple times without redrawing may be to have some sort of buffer of changes and then update only when necessary (as is frequently done in games). However, this kind of optimization at this point would be premature and unnecessary.

PR #528, opened 2020-01-23
Add link/button hint and matches follow-focus

chrisboeg:
This makes link/button-hints and search results be able to be
followed with M-n and M-p which are mapped to the two new commands
select-next-follow and select-previous-follow.

Because M-p was previously mapped to minibuffer-history,
minibuffer-history has been remapped to M-h.

When you select a result in the minibuffer
with either of the new commands, the selected result will be
highlighted, the buffer will scroll it into view, and if it is on
another buffer than the current one, which can be the case with
multi-buffer searches, a switch will happen to the correct buffer.

Here's how the current highlight looks. Feel free to come up with a better styling of it:

hint-highlight

This also changes the the C-s shortcut that was available for search
results which will focus the currently-selected result. Now it uses
the same feature as above, so it will also get the highlight.

~~Right now the scrolling that's done is kinda annoying, but I will follow up with another pr fixing that, but since this is not the only place where things are getting scrolled, it's should be a separate patch that fixes it everywhere.~~


chrisboeg:
I talked to @impaktor about this, and he suggests that the highlighting should always be done, but the "follow" (scroll, switch buffer) should only be on M-n and M-p. I think this is a good suggestion.

What I suggest is keeping this patch as is (featurewise), but then when the new pr, as we talked about, @jmercouris, that adds a *selection-changed* callback to the minibuffer, when that is implemented, then I think we should use this to just always call the function that highlights the hints and matches. But we should then change the function so it doesn't "follow" by default, only highlights, and take an optional "follow" parameter that select-next-follow and select-previous-follow will use.

Ambrevar:
Another approach would be that of Emacs Helm: add a shortcut to toggle
"follow mode".  When follow mode is on, going up and down follows the
hints on screen.

Helm's default binding is `C-c C-f`.

chrisboeg:
@Ambrevar Yes, I like both approaches.

I think I would personally prefer the toggle. However I think it would be pretty easy to have both. M-n/M-p just always follows, but it would be easy to add a toggle that changes the behavior of C-n/C-p/Up/Down. This will also the need the new patch I mentioned above. I've made a pr for it here: #529

chrisboeg:
I noticed that the behavior for scrolling in jump-to-heading is actually on point, so the only place where scrolling was annoying where here in the follow mode, so I've fixed it so it's quite better now, I think.

Ambrevar:
Or the other way around: implement M-n / M-p around follow-mode, e.g.

```
(define-command scroll-down-with-follow ()
  (with-follow-mode
    (scroll-down ...)))
```

chrisboeg:
I think that's a good solution indeed.

jmercouris:
Hi Chris, thank you again for this very nice concept! I think we are close. I like the idea of name spacing next hints and next classes, it is a very smart idea to prevent conflicts.

With regards to implementation I have an idea of what I think could be a simpler and faster implementation. Instead of creating a new element for highlighting something and overlaying it on top, what one can do is create a style assigned to a class.

Imagine that we have the following:

```
<style>
.higlighted-box {
  color: green;
}
</style>
```
then, we have whatever elements we have on screen:

```
<span class="some-class">Some element</span>
```
Instead of doing all of the manipulations and creating a new element and overlaying it on top of it, we just add a class to the span.
```
<span class="some-class highlighted-box">Some element</span>
```
Et voila! The latest CSS shall override the previous classes and allow us to highlight the element without completely removing and adding nodes from the DOM. In other words, `highlighted-box` will shadow `some-class` with regards to presentation.

In this case the function to focus match would like like this:

1. find the highlighted element via its ID from the match object
2. adding the `highlighted-box` class to it
3. remove the `highlighted-box` class from all other elements

What do you think? Thanks!

chrisboeg:
@jmercouris I've amended the patch to use the method you suggested.

So I think I've ended in this position before where I want a piece of code in two different parenscripts, but I cannot figure out how to get it working. In this patch I have `add-stylesheet` both in `element-hint.lisp` and in `search-buffer.lisp`. Do you know how?

jmercouris:
Looks good! Merging in, thank you!

PR #529, opened 2020-01-23
minibuffer changed callback

chrisboeg:
This adds a callback to the minibuffer that gets called every time
the minibuffer has been updated. This is useful for when you e.g.
want to react to changes in the minibuffer that should be reflected
in a buffer, like for example when highlighting the currently selected
search result.


chrisboeg:
I'm not sure if its good to call the changed-callback in update-display as is the case in the current patch. It might be called too often? Is there a perhaps a better place?

jmercouris:
Looks good! Thank you for the changes!

PR #534, opened 2020-01-26
Always highlight sometimes follow

chrisboeg:
This adds the the functionality of always highlighting hints using the new `changed-callback` on the minibuffer, and only following (switching buffer for say a multiple buffer search) with M-n/p.

Initially, it was planned that it shouldn't scroll either with C-n/p, and only scroll and switch buffer with M-n/p, but after testing the feature as such, I thought it was really weird to C-n/p to hints that was out of view, so in this implementation, C-n/p both highlights and scroll, and M-n/p also changes buffer as needed.

What do you think about this, @jmercouris & @impaktor?

There's a bug in this that I'm gonna fix, so it's not ready to get merged yet, but I wanted to hear what you thought about the above. I will request a review when I think it's ready to get merged.

### DIFF
--- a/source/element-hint.lisp
+++ b/source/element-hint.lisp
@@ -228,28 +228,32 @@ identifier for every hinted element."
 (defun update-selection-highlight-hint (&key completions scroll follow
                                           (minibuffer (current-minibuffer))
                                           (buffer (current-buffer)))
-  (defun hintp (hint-candidate)
-    (if (typep hint-candidate '(or link-hint button-hint match))
-        hint-candidate
-        nil))
-  
-  (let ((hint (if completions
-                  (hintp (first completions))
-                  (when minibuffer
-                    (let ((hint-candidate (nth (completion-cursor minibuffer)
-                                               (completions minibuffer))))
-                      (hintp hint-candidate))))))
-    (when hint
-      (when (and follow
-                 (slot-exists-p hint 'buffer)
-                 (not (equal (buffer hint) buffer)))
-        (set-current-buffer (buffer hint))
-        (setf buffer (buffer hint)))
-      (if (or (not (slot-exists-p hint 'buffer)) ;; type link-hint or button-hint
-              (and (slot-exists-p hint 'buffer) ;; type match, can be multi-buffer
-                   (equal (buffer hint) buffer)))
-          (highlight-selected-hint :buffer buffer :link-hint hint :scroll scroll)
-          (remove-focus)))))
+  (flet ((hintp (hint-candidate)
+           (if (typep hint-candidate '(or link-hint button-hint match))
+               hint-candidate
+               nil)))
+    (let ((hint (if completions
+                    (hintp (first completions))
+                    (when minibuffer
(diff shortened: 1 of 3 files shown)


source/element-hint.lisp
```diff
@@ -214,20 +219,31 @@ identifier for every hinted element."
 (defmethod %copy-hint-url ((button-hint button-hint))
   (echo "Can't copy URL from button."))
 
-(defun update-selection-highlight-hint (&optional completions)
+(defun update-selection-highlight-hint (&key (completions nil) (scroll nil) (follow nil))
```
Ambrevar: Style: `nil` is the default, you can simplify this to `&key completions scroll follow`.

source/element-hint.lisp
```diff
@@ -214,20 +219,31 @@ identifier for every hinted element."
...
+  (defun hintp (hint-candidate)
```
Ambrevar: `defun` in Common Lisp always defines at the top level (unlike Scheme).  Use `flet` or `labels` instead.

source/element-hint.lisp
chrisboeg: That's neat, thanks!

source/element-hint.lisp
chrisboeg: Of course, yes, thank you very much!

chrisboeg:
I think it works pretty well now.

I've ironed out different bugs, improved the code quality thanks to @Ambrevar, and adjusted the behaviour to what I think makes sense.

The state of things is now that when you initiate `search-buffer`, `search-buffers`, and `follow-hint` the minibuffer appear with results, and the top result - the one where the cursor is at in the minibuffer - is highlighted. However the buffer does not get scrolled to it. So say you are a long way down a page and invokes :follow-hint, you won't get scrolled to the top which could be very annoying. But if you start navigating the minibuffer completions with C-n/C-p/arrow up/arrow down, the buffer get scrolled so item where the cursor is at is visible if it is not already visible. However, it does not follow the item to other buffers, say if it's a multi-buffer search item.

If you want to follow the items around to other buffers too, that's where C-s and M-n/M-p comes in. And of course, if you have navigated to a multi-buffer search match in the minibuffer and you have not navigated to the buffer where the match appears, say, with C-s or M-n/M-p, if you hit return, a buffer switch will happen and the buffer will get scrolled if needed so the match is visible.

chrisboeg:
This is not ready - I just ran into a bug. Will investigate and see if it's caused by this patch and return.

jmercouris:
> ~This is not ready - I just ran into a bug. Will investigate and see if it's caused by this patch and return.~
> 
> Bug is fixed, see pending review below.

If I understand correctly, you did have to supply a dummy completion function? I assume this is because some slots are being set in `make-minibuffer` that wouldn't otherwise be set if you just `SETF`'d the completion function?

chrisboeg:
@jmercouris Indeed. When I took a cursery look, it couldn't see a way that that wouldn't be the solution in this context anyways, so I didn't dig deeper.

jmercouris:
Fantastic, thank you!

PR #539, opened 2020-01-27
no need to provide a default nil value, thank you pascal

chrisboeg:


Ambrevar:
Actually no need for the surrounding parentheses either ;)

chrisboeg:
Woah!!!

Issue #537, opened 2020-01-27
make-keymap is undefined


matzebond:
Hello,

I am trying to customize some keybindings in my config file but when I start `next` I get this error

> Next version 1.4.0
> Loading configuration from #P"/home/maschm/.config/next/init.lisp"...
> 
> ; file: /home/maschm/.config/next/init.lisp
> ; in: DEFVAR *MY-KEYMAP*
> ;     (MAKE-KEYMAP)
> ; 
> ; caught STYLE-WARNING:
> ;   undefined function: COMMON-LISP-USER::MAKE-KEYMAP
> ; 
> ; compilation unit finished
> ;   Undefined function:
> ;     MAKE-KEYMAP
> ;   caught 1 STYLE-WARNING condition
> While evaluating the form starting at line 1, column 0
>   of #P"/home/maschm/.config/next/init.lisp":
> Error: we could not load the init file
> The function COMMON-LISP-USER::MAKE-KEYMAP is undefined.

This is my config file. It is mostly copied from the manual.

```
(defvar *my-keymap* (make-keymap)
  "My keymap.")

(define-key :keymap *my-keymap*
  "g g" #'scroll-to-top
  "G" #'scroll-to-bottom
  "C-d" #'scroll-page-down
  "C-u" #'scroll-page-up
  "SPACE b" #'switch-buffer)

(define-mode my-mode ()
  "Dummy mode for the custom key bindings in `*my-keymap*'."
  ((keymap-schemes :initform (list :emacs *my-keymap*
                                   :vi-normal *my-keymap*))))
(defclass my-buffer (buffer)
  ((default-modes :initform
    (get-default 'buffer 'default-modes))))
     ;; (cons 'vi-normal-mode (get-default 'buffer 'default-modes)))))

(setf *buffer-class* 'my-buffer)
```

I am on Arch Linux using the `next-browser` package (version 1.4.0) from the AUR.
Without the init file the browser starts and works.

Thanks for any help and keep up the work on this promising browser :)

Ambrevar:
If you are going to refer next symbols without namespace prefix, you
must place your init.lisp in the "next-user" package at the beginning:

```lisp
(in-package :next-user)
```

Makes sense?

matzebond:
Thank you. I am only starting with lisp. It makes perfect sense and now `next` starts and does not reject my config file but my keybindings are not applied. Did I do something wrong.

Starting from the example in the manual I only changed this to not use vi-mode by default.
```
[...]

PR #540, opened 2020-01-27
documents/MANUAL: Add in-package to keymap example

matzebond:
Since other config snippets also have the `in-package` header we should also add it to the keymap example.
#537 for reference


Ambrevar:
Thanks!

PR #547, opened 2020-02-07
Show log4cl logs in *Messages* buffer

ebrasca:


Ambrevar:
Nice, looks good to me!

Ambrevar:
Actually, wait, we already print the `echo-*` messages in the Messages
buffer.

Maybe your proposal would make more sense if we printed the messages to
a dedicated log buffer.  What do you think?

jmercouris:
I thought we should print them to the same buffer, but have an option or key that you can press to show all logs, or echoed, or just logs.

jmercouris:
@Ambrevar - are you OK with this PR as an incremental step?

Ambrevar:
Sure!

jmercouris:
Alright! Done!

Issue #445, opened 2019-10-04
Tabbing on the "Open URL in browser" options should only fill in the URL (w/o title)


hsanjuan:
When `C-l` the user is presented with a list of options in the form "url - Title". Hitting the tab on one of those options will replace the input line with the option.

This is super useful since usually I see a url similar to the one I want to visit and I just want to edit part of that url. However, hitting the tab also inserts the page title which I then have to delete to be left with the URL only.

So my request is to make Tab only fill in the URL, rather than the url+title as it appears on the list of options.

Ambrevar:
Yup, that's annoying :p

Current TAB is very dumb and inserts the content of the candidate _as
is_.

To fix this, maybe we could specialize TAB on the candidate type.

jmercouris:
The candidate object could provide a function for when TAB is pressed much how the display is isolated from the data, is that what you are suggesting?

Ambrevar:
If I understand you right, yes, just like we have `object-string` we could have a `real-string`  method that TAB would use to know what to insert in the input prompt.

jmercouris:
Correct! That's exactly what I was thinking. I think that would be really nice!

PR #551, opened 2020-02-10
Add object-display method beside object-string for finer control.

Ambrevar:
In particular, this lets the user copy or insert the candidate using a different
string than the display, e.g. only the URL instead of "URL  TITLE" as it used to be.


Ambrevar:
Fixes #445.

jmercouris:
Perfect! I like!

PR #552, opened 2020-02-10
Fix tab-insertion of first candidate.

Ambrevar:
Search engine completion broke the insertion of the first candidate for all
minibuffers.
This change fixes it.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -130,6 +130,34 @@ URL is first transformed by `parse-url', then by BUFFER's `load-hook'."
         (set-current-buffer buffer))
       (set-url input-url)))
 
+(define-command insert-candidate-or-search-engine (&optional (minibuffer (current-minibuffer)))
+  "Paste clipboard text or to input.
+If minibuffer input is not empty and the selection is on first position,
+complete against a search engine."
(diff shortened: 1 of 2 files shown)


Ambrevar:
Fixes #445 .

Ambrevar:
Oops, wrong PR, this is not related to #445.

source/buffer.lisp
```diff
@@ -130,6 +130,34 @@ URL is first transformed by `parse-url', then by BUFFER's `load-hook'."
...
+       (let ((name (search-engine-starting-with candidate)))
+         (when name
+           (kill-whole-line minibuffer)
+           (insert (str:concat name " ")))))
+      (t
+       (when candidate
+         (kill-whole-line minibuffer)
+         (insert candidate minibuffer))))))
+
+(define-mode %set-url-mode (minibuffer-mode)
```
jmercouris: why % prefix?

Ambrevar:
I figured this was mostly an internal mode and was not meant to be
reused by the user in their init.lisp.  No strong opinion though.

jmercouris:
I would prefer not to prefix with % or so unless it is importantly private API/forbidden

Ambrevar:
Done!

PR #553, opened 2020-02-10
Sort buffer by last access time.

Ambrevar:
This fixes the erratic buffer order when deleting the current buffer or when
switching to the next / previous one.


jmercouris:
Looks perfect! Might we not just delete switch-buffer-next?

Ambrevar:
Why deleting it?  I use it quite a lot! :)

PR #556, opened 2020-02-11
Add enable-hook-handler and disable-hook-handler.

Ambrevar:
This allows the user to easily toggle hook handler while Next is running,
without having to enter code.

### DIFF
--- a/source/command.lisp
+++ b/source/command.lisp
@@ -200,3 +200,73 @@ This function can be `funcall'ed."
                           :show-completion-count nil)))
     (setf (access-time command) (get-internal-real-time))
     (run command)))
+
+(defclass hook-description ()
+  ((name :accessor name :initarg :name
+         :initform ""
+         :type string
+         :documentation "The hook name.")
+   (value :accessor value :initarg :value
+          :documentation "The hook value.")))
+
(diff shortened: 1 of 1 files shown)


jmercouris:
Nice!

Does this also mean (since there is a completion handler) that the user can effectively list all hooks? In my opinion it is not necessary to make a function called "list-hooks" in that case. All of this said, I also didn't see the need for a "reload page" function since the user could type in C-l RET and it would reload the page. What would a more typical user expect?

Ambrevar:
What I currently do to "find all hooks" is lookup the slots ending with
"-hook" in the buffer, window and remote-interface classes.  It's enough
for our needs, but true so a more general approach would still be
welcome.

I'm not sure I understood what you meant with "reload page".  Is there
anything you'd like me to fix in this PR?

jmercouris:
There is nothing to fix in this PR, just a question of whether it would be advantageous for us to also add a "list-hooks" function. The point I was trying to make is, in my opinion it is not necessary (due to the completion handler), but others may think otherwise. What is your opinion?

Ambrevar:
No strong opinion at the moment :)

jmercouris:
OK, then we leave it as is, and we see if it proves to be necessary.

Issue #518, opened 2020-01-14
resource-query-function slot is not used if buffer have a blocker-mode enabled


chrisboeg:
If I add the following to my `init.lisp`:

```
(defclass my-buffer (buffer)
  ((default-modes :initform (list get-default 'buffer 'default-modes))
   (resource-query-function
     :initform (lambda (&rest args)
                 (progn
                   (log:info "args: ~a" args)
                   nil)))))

(setf *buffer-class* 'my-buffer)
```

and then run next in a shell, I can see the log-output from my `resource-query-function` in the `stdout` from next and indeed no urls are allowed to load. However if I add a `blocker-mode` like so:

```
(define-mode my-blocker-mode (next/blocker-mode:blocker-mode)
  "A docstring."
  ((hostlists :initform (list next/blocker-mode:*default-host-list*))))

(defclass my-buffer (buffer)
  ((default-modes :initform (append
                              '(my-blocker-mode)
                              (get-default 'buffer 'default-modes)))
   (resource-query-function
     :initform (lambda (&rest args)
                 (progn
                   (log:info "args: ~a" args)
                   nil)))))

(setf *buffer-class* 'my-buffer)
```

now my `resource-query-function` is completely ignored. It seems like the `blocker-mode` overwrites the `resource-query-function` slot behind the scene, which I think seems like a bug.

**Information**
- Next version (from =M-x next-version= or =next --version=): 1.4.0
- WebKitGTK+ / PyQt-Webengine version: WebKitGTK+

Ambrevar:
Hi Chris,

yes, blocker-mode sets resource-query-function.

If you want to use both blocker-mode and your own query function, the
solution is simple: define your custom my-blocker-mode which inherits
from blocker-mode and assign it your own resource-query-function.  This
function will probably fall back on blocker-mode's query function.

As you realized first hand, this is not the greatest design decision as
it can trip users easily.

A while back I though we should make resource-query-function a list of
functions and run them all in order.
At this point, it looks very much like a hook, and now that we've merged
the overpowered hook system, maybe we could leverage that.

chrisboeg:
Thank you for the tip on how to work around it.

I at least think it would be a good idea to allow multiple functions in some way, because as it is now is confusing until you read the code.

Ambrevar:
Maybe reopen this issue until we've implemented a better approach, for
instance with hooks.

chrisboeg:
Sorry.

Ambrevar:
[...]

PR #558, opened 2020-02-11
FFI Rework

jmercouris:
This is the FFI branch that I've been working on.

I've changed quite a few concepts. It works on macOS too, which is great.

Tasks Left (in-exhaustive list):
+ Use *buffer*, *interface* class etc
+ Finish IPC methods, RPC -> IPC
+ Work on C -> C++ Glue code @chrisboeg 

I would be interested in some preliminary feedback as I am working on this first draft. Ideally I would like to get something working, merge it onto master so that things do not diverge too far, and then we incrementally improve upon it. Interested in your early thoughts and questions @Ambrevar.

The first draft will only feature a GTK port that is cross-platform. It is tested on SBCL, with some functionality tested on CCL.


Ambrevar:
Very quick feedback:

- It mostly works for now!  Hooray!
- REPL is busy.
- Key presses are doubled.

I'll look more into it Very Soon©.

Ambrevar:
Also let me know when you want to take a break and I can review the
whole thing ;)

jmercouris:
OK! I let you know when first alpha is done, thanks!

Ambrevar:
Some early comments:

- What about renaming `interface` to `instance`?  I think it would make more sense.
- The Makefile is broken, you need to remove ` source/ports/*.lisp` in

```
lisp_files := next.asd source/*.lisp source/ports/*.lisp
```

- The `port` related slots in `interface` can be removed.  As well as `active-connection`.

jmercouris:
> Some early comments:
> 
> * What about renaming `interface` to `instance`?  I think it would make more sense.
> * The Makefile is broken, you need to remove ` source/ports/*.lisp` in
> 
> ```
> lisp_files := next.asd source/*.lisp source/ports/*.lisp
> ```
> 
> * The `port` related slots in `interface` can be removed.  As well as `active-connection`.

Done!

Except for the renaming. I'll need to think about that. `instance` does seem a good name, though the name could just as easily be `next`, which might be more consistent with other programs.

jmercouris:
Fixed keypress doubling, keymap.lisp now properly reports keypress consumption so that the keypress event can propagate or not propagate through the widget/responder tree.

Ambrevar:
`next` is fine by me, but it's not very resistant to refactoring
(e.g. porting pieces of the code to another project or as a library).

`interface` is not correct in my opinion, since it no longer is an interface.

Ambrevar:
keypress: Great, I'll test and start using it then!

jmercouris:
I will be pausing development over the weekend, so please give it a full review at your own convenience. Thank you!

jmercouris:
JavaScript problem has been "solved". Not exactly how I would like. Any more elegant solutions are greatly appreciated!

The problem is in the line in cl-webkit:
```
(let ((js-result (webkit-web-view-run-javascript-finish (callback-web-view callback) result))...
```
which calls:
```
(defun webkit-web-view-run-javascript-finish (web-view result)
  (glib:with-g-error (err)
    (%webkit-web-view-run-javascript-finish web-view result err)))
(export 'webkit-web-view-run-javascript-finish)
```
This function accepts a gerror pointer as can be seen here: https://webkitgtk.org/reference/webkit2gtk/2.5.1/WebKitWebView.html#webkit-web-view-run-javascript-finish, that is why it is wrapped in glib error macro in cl-webkit2:

You can see the fixes here: 
https://github.com/atlas-engineer/cl-webkit/commit/9e725eef6d73d6e6400ff99b76f98676cdabd7ab
https://github.com/atlas-engineer/cl-webkit/commit/073b8eb98324641c798234afb244814bdaefba7b

I would do something better, but we have to use the CFFI defcallback. Which, as you can see we are not calling directly, and is being invoked by C code in `webkit-web-view-evaluate-javascript`, otherwise I would handle the condition elsewhere.

It is very possible that there is a way to do just that, and I don't know it, it would certainly be better to handle the condition in the callback that we pass into `webkit-web-view-evaluate-javascript` rather than calling the callback with nil as a return value to signify failure of javascript evaluation.

I hope that makes sense, it is a bit difficult to explain :-)

Ambrevar:
Yes, you are making sense.
I've fetched your fix and this fixes the crash on restart, so that's a
great start! :)

Sadly I've experienced a similar error in a different context.  Needs
more testing.

I've pushed a couple of changes on the `ambrevar/ffi` branch.

Important fixes:

- CLOS-style config works again.
- REPL does not close on exit, but we need to kill the app when run as a
  stand-alone.  To do later.
- Killing all windows terminates Next.

I tried to persist the cookies but it does not work yet.  Still need to
figure out why.

Cheers!

Ambrevar:
Regarding the CFFI rearchitecture: Great work!
I think it's looking very good overall!

Some functions might need splitting, we'll see.

What I suggest we do, before 2.0 release: encapsulate all the GTK code,
probably rename interprocess-communication to ipc-gtk.lisp.
As a proof of concept, I think including a 2nd implementation would
validate the current one.  I think the most trivial would be to have a
cl-gobject-introspection implementation, say as "ipc-gtk-gi.lisp".

Ambrevar:
I think I've figured out the issue with cookies.

Ambrevar:
Note: Previously we used D-Bus to support running Next in
single-instance.  It's a common use of D-Bus.

I don't know how to do it without D-Bus.  We could roll our own lock
file or something like that, but it's brittle and it reinvents the
wheel.

So what about reintroducing D-Bus for single-instance checking?

jmercouris:
Hello Pierre, thank you for the review so far! I've read through your branch as well as your comments. I will probably cherry pick some things on Monday or so. I agree with most changes, but I would like to discuss a few (for example: I believe we should specialize `ipc-window-make` because other interfaces may have different requirements, things that need to be done during the creation of a window. If we are not to specialize it on `gtk-interface`, I think it could at least be specialized on `interface` rather than using a `defun`).

> Note: Previously we used D-Bus to support running Next in single-instance. It's a common use of D-Bus. I don't know how to do it without D-Bus. We could roll our own lock file or something like that, but it's brittle and it reinvents the wheel. So what about reintroducing D-Bus for single-instance checking?

As per this, I have actually had an opposite request from a user, that we *allow* multiple instances of Next. I personally think we should allow multiple instances. If you are wondering how we can support a single instance without the use of d-bus, I believe it can more reliably be done via a unix domain socket. D-bus is difficult to obtain on macOS is unlikely to exist, and requires set-up. Of course, not even on Linux have we seen it universally available. On the other hand, unix domain sockets are available everywhere, and we only need to send a single message type to handle single instance.

jmercouris:
> I think I've figured out the issue with cookies.

I believe it has to be set-up during the creation of the buffer or so. I can see an example of it here: https://github.com/AeroNotix/lispkit/blob/master/browser.lisp#L95

jmercouris:
> Regarding the CFFI rearchitecture: Great work! I think it's looking very good overall! Some functions might need splitting, we'll see. What I suggest we do, before 2.0 release: encapsulate all the GTK code, probably rename interprocess-communication to ipc-gtk.lisp. As a proof of concept, I think including a 2nd implementation would validate the current one. I think the most trivial would be to have a cl-gobject-introspection implementation, say as "ipc-gtk-gi.lisp".

Thank for the kind words! interprocess-communication could definitely be renamed, I would prefer `interprocess-communication-gtk` if you are OK with that.

As per a 2nd implementation before 2.0, I'm not sure I have the stamina for that :-D. I instead suggest we release 2.0 and in 2.1 we release a 2nd implementation. As per what it should be, I agree cl-gobject-introspection would be relatively trivial to translate- you are right- unfortunately I am not a huge fan of it, and would rather proceed with working on the Qt implementation for that as that is the ultimate goal. Having two implementations for GTK doesn't get us anything extra :-D

thoughts?

jmercouris:
> Yes, you are making sense. I've fetched your fix and this fixes the crash on restart, so that's a great start! :) Sadly I've experienced a similar error in a different context. Needs more testing. I've pushed a couple of changes on the `ambrevar/ffi` branch. Important fixes: - CLOS-style config works again. - REPL does not close on exit, but we need to kill the app when run as a stand-alone. To do later. - Killing all windows terminates Next. I tried to persist the cookies but it does not work yet. Still need to figure out why. Cheers!

What do you mean by a similar error in a different context? A different type of g-error or?

- Thank you for the CLOS style config fix, that was on my task list :-)
- REPL does not close on exit, but need to kill the app: instead of detecting if we are on the REPL or not, might we instead have two different commands?
- Killing all windows terminates Next, I didn't even think about that one :-D good catch!

jmercouris:
What about `core` as a replacement name for `interface`?

Ambrevar:
> What about `core` as a replacement name for `interface`?

Or `browser` like Lispkit did :)

Ambrevar:
> Thank for the kind words! interprocess-communication could definitely be renamed, I would prefer `interprocess-communication-gtk` if you are OK with that.

It's a bit longish, but I guess that does not really matter.
Otherwise gui-gtk, or backend-gtk.

> As per a 2nd implementation before 2.0, I'm not sure I have the
> stamina for that :-D. I instead suggest we release 2.0 and in 2.1 we
> release a 2nd implementation. As per what it should be, I agree
> cl-gobject-introspection would be relatively trivial to translate- you
> are right- unfortunately I am not a huge fan of it, and would rather
> proceed with working on the Qt implementation for that as that is the
> ultimate goal. Having two implementations for GTK doesn't get us
> anything extra :-D

I strongly suggest we do this before 2.0 because it will validate the
API.
Doing it after will probably reveal some API-breaking requirements.
It does not cost much to implement (it can probably be done in less than
a week) and I'd be happy to do it.

What cl-object-introspection buys us is nearly-instant Qt compatibility
thanks to GIQt.

Also consider this:

https://github.com/joachifm/cl-webkit/issues/33

cl-gobject-introspection might be more maintained / maintainable in the future.

Ambrevar:
> As per this, I have actually had an opposite request from a user, that
> we *allow* multiple instances of Next. I personally think we should
> allow multiple instances. If you are wondering how we can support a
> single instance without the use of d-bus, I believe it can more
> reliably be done via a unix domain socket. D-bus is difficult to
> obtain on macOS is unlikely to exist, and requires set-up. Of course,
> not even on Linux have we seen it universally available. On the other
> hand, unix domain sockets are available everywhere, and we only need
> to send a single message type to handle single instance.

Totally agree, we need to support both single-instance and
multi-instance Next.

On second thought, you are right, a Unix socket is the trivial way to
implement this.

D-Bus gives us the optional benefit of controlling Next from a foreign
application; something useful to keep in mind but I guess we can make
this an opt-in build-time option.
Emacs supports this for instance.

Conclusion on my end: let's implement the Unix-socket way :)

Ambrevar:
I suggest we use the `webkit` package nickname instead of `cl-webkit2`.

Ambrevar:
For Unix sockets: do you know anything that supports it in Common Lisp
beside IOLib?

Ambrevar:
Never mind the javascript error, I had forgotten to merge cl-webkit
properly.
I've pushed a fix to your fix by the way ;)

Ambrevar:
To create a local socket with IOLib:

(iolib:make-socket :address-family :local :local-filename "/path/to/dummy.socket")

Ambrevar:
I've been using Next for a couple of hours.
I can report one main issue so far: it seems to randomly hang after a
few dozen minutes.

Ambrevar:
I've implement initial proxy support.
Code is both on cl-webkit and in Next.

I don't know how to pass a "const gchar * const *" with CFFI, so I don't
know how to set the ignore-hosts.  Any idea?

Ambrevar:
Another (big) bug: when running multiple instance, the former instances
tend to hang.

Probably related, but I've noticed that from the REPL, if I quit Next
and restart it from the same Lisp instance, the WebKit process is
reused.  I can see that from the cookies session being still there (when
I disable cookie persistence).

jmercouris:
> What about `core` as a replacement name for `interface`?
> Or `browser` like Lispkit did :)

browser is OK! I shall do it after we merge this branch or so.

jmercouris:
> I strongly suggest we do this before 2.0 because it will validate the API. Doing it after will probably reveal some API-breaking requirements. It does not cost much to implement (it can probably be done in less than a week) and I'd be happy to do it. What cl-object-introspection buys us is nearly-instant Qt compatibility thanks to GIQt. Also consider this: [joachifm/cl-webkit#33](https://github.com/joachifm/cl-webkit/issues/33) cl-gobject-introspection might be more maintained / maintainable in the future.

OK, no problem. That said, let us please merge this into master sooner than later :-)

jmercouris:
> I suggest we use the `webkit` package nickname instead of `cl-webkit2`.

OK!

jmercouris:
As I am cherry picking I am making some comments on commits with my modifications, this an invitation for discussion, and I'm willing to change most things :-)

Ambrevar:

> OK, no problem. That said, let us please merge this into master sooner than later :-)

To clarify, no problem with merging on master once the main pain points
are solved.
From there we can implement the cl-gobject-introspection backend.
Once both work, we release 2.0.

Sounds good?

jmercouris:
> To clarify, no problem with merging on master once the main pain points are solved. From there we can implement the cl-gobject-introspection backend. Once both work, we release 2.0. Sounds good?

Exactly what I was thinking! No problem!

jmercouris:
> I've implement initial proxy support. Code is both on cl-webkit and in Next. I don't know how to pass a "const gchar * const *" with CFFI, so I don't know how to set the ignore-hosts. Any idea?

No idea, but I will look into it. I will be trying to get ipc-buffer-get-proxy working firstly though.

jmercouris:
I believe all that is a *must* have before merging is decide-policy and proxies?

Ambrevar:
Also now is a good time to refactor some of the code and move the
`window` and `buffer` class to their respective files.

The `interface.lisp` file is a big mess at the moment :)

Ambrevar:
That and
- single-instance
- stability: it hangs for me much too often.  It would help if we would
log all of stdout / stderr to a file.

jmercouris:
OK, then the final list:

- decide policy (in progress (I'm working on it))
- proxies (you are working on?)
- move window and buffer classes to correct files
- single instance support
- improve stability

Ambrevar:
> - proxies (you are working on?)

Just need to pass the ignore-hosts via CFFI.  Will give it a try.

> - move window and buffer classes to correct files

This can be done after the merge on master, but before 2.0.

> - single instance support
> - improve stability

+ logging, which should help soving the last point.

jmercouris:
Acknowledged!

Ambrevar:
Once merged, we can edit the GitHub issue template since we won't need
that many details.

jmercouris:
Something interesting I've noticed, it almost never crashes from me when I launch Next from the REPL, but if I launch the app bundle, it will crash on certain URLs. I guess logging stdout/err will help with this.

Ambrevar:
Another things before 2.0 (maybe unrelated):

use `alex:` and `sera:` local nicknames instead of the lengthy
`alexandria:` and `serapeum:`.

jmercouris:
Only thing left is proxy support passing gchar const, I'll look at that now. Then we can continue to work on this on master!

As soon as this is on master I will: rename interface -> browser, move methods to appropriate spots for window and buffer

After that I will begin working on the Qt implementation (I assume you will during that time try to implement the g-object introspection version)

Ambrevar:
Great!

Have you managed to log the output and fix the hangs?

jmercouris:
I have managed to log the output, you can take a look at the appender I've created. I cannot reproduce the hangs as it does not ever hang on my machine. I've ran it many hours at a time via the terminal with absolutely no problem.

jmercouris:
Sorry, I haven't made an appender I made a method called write-output-to-log which can be used in a binary version. I would have to setup the appender so that log4cl can append to the appropriate streams and also get written.

jmercouris:
I've added ignore hosts support in, can you please test it?

jmercouris:
If it fails right away, you can also try recompiling without the cffi:foreign-free, it is possible that it needs the pointer to ignore-hosts instead of copying the data.

Ambrevar:
I can't test today, but will do as soon as possible.

jmercouris:
OK, well, then I will assume it works and merge onto master :-P

just kidding, let me know when you test it :-)

jmercouris:
I would like to merge this today, do you have time to test? Otherwise I think it is OK to merge onto master.

Ambrevar:
Sorry, didn't find time to check it out last weekend.  Will do as soon
as possible, could you please hold the merge for a little while?  Thanks!

jmercouris:
I can wait one more day, I want to get started on the Qt port and finish the first milestone.

Ambrevar:
I started reviewing the latest changes, sadly I could not start next because of
the first item below.

It's in good shape overall, but have you managed to redirect the log?

- Fails to start: Many dangling references to RESOURCE-QUERY-FUNCTION and RESOURCE-QUERY-DEFAULT
  need to be fixed.

- We need to restore certificate-whitelist-mode.  At the
  very least, add an obvious TODO about it.

- Rename interface.lisp -> browser.lisp.

- Commit 0cda227bf49ec144407a8ff487f73a489e711171 contains snippets that belong
  to the previous commit.

- To temporarily disable expressions, use `#+nil` instead of `(when (and ... nil)`.

- ring:
```diff
-(declaim (ftype (function (ring) ring) ring-copy))
+(declaim (ftype (function (ring) ring) copy))
```

- Our build will fail for all users who haven't checkout our cl-webkit.
  1. Send pull requests upstream and see if they get merged.
  2. Edit the Makefile to clone our cl-webkit.

- process-decide-policy: Use match, it will unclutter the code:

```lisp
(defmethod process-decide-policy ((buffer gtk-buffer) response-policy-decision policy-decision-type-response)
  (let ((is-new-window nil) (is-known-type t) (event-type nil)
        (navigation-action nil) (navigation-type nil)
        (mouse-button nil) (modifiers ())
        (url nil) (request nil))
    (match policy-decision-type-response
      (:webkit-policy-decision-type-navigation-action
       (setf navigation-type (webkit:webkit-navigation-policy-decision-navigation-type response-policy-decision)))
      (:webkit-policy-decision-type-new-window-action
       (setf navigation-type (webkit:webkit-navigation-policy-decision-navigation-type response-policy-decision))
       (setf is-new-window t))
      (:webkit-policy-decision-type-response
       (setf request (webkit:webkit-response-policy-decision-request response-policy-decision))
       (setf is-known-type
             (webkit:webkit-response-policy-decision-is-mime-type-supported
              response-policy-decision))))
    ;; Set Event-Type
    (setf event-type
          (match navigation-type
            (:webkit-navigation-type-link-clicked :link-click)
            (:webkit-navigation-type-form-submitted :form-submission)
            (:webkit-navigation-type-back-forward :backward-or-forward)
            (:webkit-navigation-type-reload :reload)
            (:webkit-navigation-type-form-resubmitted :form-resubmission)
            (:webkit-navigation-type-other :other)))
    ;; Get Navigation Parameters from WebKitNavigationAction object
    (when navigation-type
      (setf navigation-action (webkit:webkit-navigation-policy-decision-get-navigation-action response-policy-decision))
      (setf request (webkit:webkit-navigation-action-get-request navigation-action))
      (setf mouse-button (format nil "button~d" (webkit:webkit-navigation-action-get-mouse-button navigation-action)))
      (setf modifiers (translate-modifier-list *browser* (webkit:webkit-navigation-action-get-modifiers navigation-action))))
    (setf url (webkit:webkit-uri-request-uri request))
    (request-resource buffer
                      :url url
                      :mouse-button mouse-button
                      :event-type event-type
                      :is-new-window is-new-window
                      :is-known-type is-known-type
                      :modifiers modifiers)))
```

Ambrevar:
Regarding the request-resource functions, why did you remove the slot from the
BROWSER class?

Ambrevar:
Other issues:

- Sometimes in VI mode, pressing "g" brings me to the
  top of the page, while it should be "gg".
  
- shift-space does not work.

Ambrevar:
Next is still hanging for me.  Recipe:

- Start Next twice.
- Navigate to a page in both instances.
- Wait for 5 minutes.
- Go back to both instances: one of them will be hung.

I think this happens because of cl-webkit.
Something is wrong with how it deals with WebKit processes.  A way to see this
is by setting the cookie persistance file to nothing, start Next from the REPL, login to a
website, quit, then
start again: you'll still be logged in.  So it seems that the WebKit process is
reused somehow.



`quit commant`: We should not call `uiop:quit`, or it will exit the REPL.
On Linux, not calling it does the right thing (it exists SBCL if run from a binary.)

If the process still persists on macOS, maybe do the following: 

- Set a global parameter `*keep-alive*` to `t`.
- In ENTRY-POINT, set `*keep-alice*` to `nil`.
- In `quit`, call `uiop:quit` if `*keep-alive*` is nil.

What do you think?



Other issue:

- COmmit 970ff4de0213e0c23f0bf7c4f195ba211e8dafc2:
Commit message first line does not tell what it does.
"do not use match, makes code harder to understand": I invite you to
reconsider.  `match` can save a lot of error-prone repetitions:

```lisp
(defun character->string (character &optional key-value)
  (match character
    (#\Return "RETURN")
    (#\Backspace "BACKSPACE")
    (#\Esc "ESCAPE")
    (#\- "HYPHEN")
    (#\Space "SPACE")
    (#\Tab "TAB")
    (_ (match key-value
         (65361 "Left")
         (65362 "Up")
         (65364 "Down")
         (65363 "Right")
         (_ (unless (eq character #\Nul)
              (string character)))))))
```

Ambrevar:
About than hangs: there is no SBCL error / output, so logging won't help us here.

jmercouris:
> It's in good shape overall, but have you managed to redirect the log?

Yes, but it does not help :-)

> * Fails to start: Many dangling references to RESOURCE-QUERY-FUNCTION and RESOURCE-QUERY-DEFAULT
>   need to be fixed.

OK will RM

> * We need to restore certificate-whitelist-mode.  At the
>   very least, add an obvious TODO about it.

Added Note

> * Rename interface.lisp -> browser.lisp.

Completed

> * Commit [0cda227](https://github.com/atlas-engineer/next/commit/0cda227bf49ec144407a8ff487f73a489e711171) contains snippets that belong
>   to the previous commit.
> * To temporarily disable expressions, use `#+nil` instead of `(when (and ... nil)`.

A good point!

> * ring:
> 
> ```diff
> -(declaim (ftype (function (ring) ring) ring-copy))
> +(declaim (ftype (function (ring) ring) copy))
> ```
> 
> * Our build will fail for all users who haven't checkout our cl-webkit.
>   
>   1. Send pull requests upstream and see if they get merged.
>   2. Edit the Makefile to clone our cl-webkit.

OK, I made a note in the developer README

> * process-decide-policy: Use match, it will unclutter the code:
>
> ```lisp
> (defmethod process-decide-policy ((buffer gtk-buffer) response-policy-decision policy-decision-type-response)
>   (let ((is-new-window nil) (is-known-type t) (event-type nil)
>         (navigation-action nil) (navigation-type nil)
>         (mouse-button nil) (modifiers ())
>         (url nil) (request nil))
>     (match policy-decision-type-response
>       (:webkit-policy-decision-type-navigation-action
>        (setf navigation-type (webkit:webkit-navigation-policy-decision-navigation-type response-policy-decision)))
>       (:webkit-policy-decision-type-new-window-action
>        (setf navigation-type (webkit:webkit-navigation-policy-decision-navigation-type response-policy-decision))
>        (setf is-new-window t))
>       (:webkit-policy-decision-type-response
>        (setf request (webkit:webkit-response-policy-decision-request response-policy-decision))
>        (setf is-known-type
>              (webkit:webkit-response-policy-decision-is-mime-type-supported
>               response-policy-decision))))
>     ;; Set Event-Type
>     (setf event-type
>           (match navigation-type
>             (:webkit-navigation-type-link-clicked :link-click)
>             (:webkit-navigation-type-form-submitted :form-submission)
>             (:webkit-navigation-type-back-forward :backward-or-forward)
>             (:webkit-navigation-type-reload :reload)
>             (:webkit-navigation-type-form-resubmitted :form-resubmission)
>             (:webkit-navigation-type-other :other)))
>     ;; Get Navigation Parameters from WebKitNavigationAction object
>     (when navigation-type
>       (setf navigation-action (webkit:webkit-navigation-policy-decision-get-navigation-action response-policy-decision))
>       (setf request (webkit:webkit-navigation-action-get-request navigation-action))
>       (setf mouse-button (format nil "button~d" (webkit:webkit-navigation-action-get-mouse-button navigation-action)))
>       (setf modifiers (translate-modifier-list *browser* (webkit:webkit-navigation-action-get-modifiers navigation-action))))
>     (setf url (webkit:webkit-uri-request-uri request))
>     (request-resource buffer
>                       :url url
>                       :mouse-button mouse-button
>                       :event-type event-type
>                       :is-new-window is-new-window
>                       :is-known-type is-known-type
>                       :modifiers modifiers)))
> ```

If you want to change it to use MATCH, that is OK with me, however I am not a fan of using it myself. I read through the documentation, and I still prefer traditional control flow structures. Maybe I'll grow a liking to it over time.

jmercouris:
> Regarding the request-resource functions, why did you remove the slot from the
> BROWSER class?

I believe it is easier and more straightforward for the user to redefine the method in their init file rather than setting the value of a slot in the browser class. So the idea is you don't need to have two steps, just one.

jmercouris:
> Other issues:
> 
> * Sometimes in VI mode, pressing "g" brings me to the
>   top of the page, while it should be "gg".
> * shift-space does not work.

Seems that it is still double pressing even with consumption of keys. I have this issue occasionally where I can not type multiple combinations like `C-m s` because it will have double clicked `C-m` which is not bound to anything. Other keys like `M-x` will continue to work though. Needs more investigation.

jmercouris:
Submitted pull request to `cl-webkit` https://github.com/joachifm/cl-webkit/pull/36.

jmercouris:
Also, is the failure to start due to your init file setting up VI mode? with an empty init file do you still have failure to start?

jmercouris:
As per uiop:quit and quit command, no need to have a separate behavior on macOS and Linux. Should you wish to just kill the interface invoke kill-interface, should you wish to exit Next, invoke the quit command. Does that work OK?

Ambrevar:
> As per uiop:quit and quit command, no need to have a separate behavior on macOS and Linux. Should you wish to just kill the interface invoke kill-interface, should you wish to exit Next, invoke the quit command. Does that work OK?

I don't think it's a good idea.

1. By default (for now) the REPL is blocked, so it's not easy to call
(kill-interface).
2. It's too easy (muscle memory) to press C-x C-c (or ZZ) to quit the
browser and it will accidentally quit the REPL.

Ambrevar:
> Also, is the failure to start due to your init file setting up VI mode? with an empty init file do you still have failure to start?

The failure to start is due to the request-query function, which I've
fixed on my ambrevar/ffi-2 branch.

Ambrevar:

> I believe it is easier and more straightforward for the user to redefine the method in their init file rather than setting the value of a slot in the browser class. So the idea is you don't need to have two steps, just one.

Why not, but consider issue #518.
Either way, the implementation we have in 1.5 is not ideal.

We need a way that clearly tells the user what's going on.
The user will want to know "What are the currently active resource query functions?"

I'm not sure that methods are a good way to do it because it's hard to
list them.

A probably better way, now that we have merged them: hooks!  I believe
it's idea because we can define how the handlers are composed + users
can easily list them, customize them, disable them, etc.

Ambrevar:
More progress on the hangs:  I can reproduce as follows:

- Start Next from a binary.
- Let it run for almost precisely 10 minutes.
- At minute 10, it will hang and the process will start eating a CPU.

No need for multiple instances actually.

Can you reproduce?

I'll try to reproduce with a minimal cl-webkit example see if it really comes
from the library, or if we are using it badly.

jmercouris:
> As per uiop:quit and quit command, no need to have a separate behavior on macOS and Linux. Should you wish to just kill the interface invoke kill-interface, should you wish to exit Next, invoke the quit command. Does that work OK?
> I don't think it's a good idea. 1. By default (for now) the REPL is blocked, so it's not easy to call (kill-interface). 2. It's too easy (muscle memory) to press C-x C-c (or ZZ) to quit the browser and it will accidentally quit the REPL.

What if we had a prompt with a y-n question that one could toggle via their init file ?

jmercouris:
> Also, is the failure to start due to your init file setting up VI mode? with an empty init file do you still have failure to start?
> The failure to start is due to the request-query function, which I've fixed on my ambrevar/ffi-2 branch.

OK thank you! I will cherry pick your fixes!

jmercouris:
> More progress on the hangs: I can reproduce as follows:
> 
> * Start Next from a binary.
> * Let it run for almost precisely 10 minutes.
> * At minute 10, it will hang and the process will start eating a CPU.
> 
> No need for multiple instances actually.
> 
> Can you reproduce?
> 
> I'll try to reproduce with a minimal cl-webkit example see if it really comes
> from the library, or if we are using it badly.

I can not reproduce, I can leave Next running for several days now without crashing. In the REPL do you see any X errors or GLIB warnings? Same behavior from
compiled vs running in REPL?

jmercouris:
Any messages in the inferior lisp? Is Lisp still responsive?

jmercouris:
If your problem only occurs with the binary I have some theories to fix it.

Ambrevar:
I can confirm that it only occurs in the binary.

Ambrevar:
> What if we had a prompt with a y-n question that one could toggle via their init file ?

Sounds like an implementation detail we would ideally hide from the
user.

ENTRY-POINT is only every run by the binary, so in my opinion my
proposal fits both use cases without exposing anything to the user.

Ambrevar:
No message, no warning.

What's your webkitgtk version?

Ambrevar:
Breaking news: compiling this demo
(https://github.com/joachifm/cl-webkit/blob/master/demo/simple-browser.lisp)
into a binary does not hang!

Something is wrong in our code or in the more advanced use of cl-webkit
/ cffi-gtk.

Maybe the threads or the signals.

jmercouris:
> No message, no warning. What's your webkitgtk version?

jmercouris@Fruit lib % ls libwebkit2gtk-4.0.*
libwebkit2gtk-4.0.37.42.0.dylib	libwebkit2gtk-4.0.dylib
libwebkit2gtk-4.0.37.dylib

jmercouris:
> Breaking news: compiling this demo (https://github.com/joachifm/cl-webkit/blob/master/demo/simple-browser.lisp) into a binary does not hang! Something is wrong in our code or in the more advanced use of cl-webkit / cffi-gtk. Maybe the threads or the signals.


Interesting find!

Some things that stand out to me: it doesn't use Javascript and it uses the recommended `within-gtk-loop` macro (which does not work on macOS), we can make it a conditional for Linux if it results in your binary not crashing.

jmercouris:
> What if we had a prompt with a y-n question that one could toggle via their init file ?
> Sounds like an implementation detail we would ideally hide from the user. ENTRY-POINT is only every run by the binary, so in my opinion my proposal fits both use cases without exposing anything to the user.

I'm not sure if that will work, that's all (and it may be a lot of work to get it to work, when I want to move on and complete the Qt port as soon as possible).

jmercouris:
Shall I now merge while we continue to work on this?

Ambrevar:
> Shall I now merge while we continue to work on this?

Considering it breaks on Linux, I'd say no, let's wait until it's fixed.
Your hunch about the within-gtk-loop is good, I'll have a look today.
Fingers crossed and maybe we will merge on master today / this week :)

Ambrevar:
>> What if we had a prompt with a y-n question that one could toggle via their init file ?
>> Sounds like an implementation detail we would ideally hide from the user. ENTRY-POINT is only every run by the binary, so in my opinion my proposal fits both use cases without exposing anything to the user.
>
> I'm not sure if that will work, that's all (and it may be a lot of work to get it to work, when I want to move on and complete the Qt port as soon as possible).

I think you misunderstood, what I'm suggesting is super simple.  I'll
send a patch later.

jmercouris:
> > What if we had a prompt with a y-n question that one could toggle via their init file ? > Sounds like an implementation detail we would ideally hide from the user. ENTRY-POINT is only every run by the binary, so in my opinion my proposal fits both use cases without exposing anything to the user. I'm not sure if that will work, that's all (and it may be a lot of work to get it to work, when I want to move on and complete the Qt port as soon as possible).
> I think you misunderstood, what I'm suggesting is super simple. I'll send a patch later.

OK :-)

jmercouris:
> Shall I now merge while we continue to work on this?
> Considering it breaks on Linux, I'd say no, let's wait until it's fixed. Your hunch about the within-gtk-loop is good, I'll have a look today. Fingers crossed and maybe we will merge on master today / this week :)

OK, I can also take a look on my VM and see if I can reproduce some of the behavior.

jmercouris:
I can reproduce the crashing in my VM. I will work on the C++ API for now, and get back to this if you have not already handled it.

Ambrevar:
Which GTK versions are you using in the VM and on macOS?
Mine is 3.24.13.

jmercouris:
I’m using 3.24.12 on my machine, I don’t know how to check the VM

Ambrevar:
Great news!  It works now for me on Linux with within-main-loop!

I'll send a patch in a little while.

Ambrevar:
Depends on the distro's package manager.
Which distribution is it?

jmercouris:
Great!!! That’s fantastic!

It is Ubuntu, anyways if within-main-loop solves it, then I don’t need to look at it, please make it conditional so that it is only used on Linux, as that line actively makes it malfunction on macOS

Ambrevar:
I've just sent a patch on ambrevar/ffi-2.

I haven't put a conditional yet because I'd like you to give it a try,
see if you really get a problem in this case.
If so, could you report the issue you are getting to cffi-gtk?

If move the code around a little bit so that it should be trivial to put
a conditional.

Let me know! :)

jmercouris:
> I've just sent a patch on ambrevar/ffi-2. I haven't put a conditional yet because I'd like you to give it a try, see if you really get a problem in this case. If so, could you report the issue you are getting to cffi-gtk? If move the code around a little bit so that it should be trivial to put a conditional. Let me know! :)

I just tried it, and it did not work. This problem took me like two weeks to solve already :-D. It involves not being able to draw when not on the main thread of macOS, which is why I cannot join the main GTK loop. It has to do with an implementation detail of WebKitGTK+.

Whenever you get a chance, please put the conditional in place and I will merge onto this branch, and then onto master!

jmercouris:
I added a conditional, I will test on Linux, and if it works, merge onto master.

jmercouris:
Works for me! Just waiting for your confirmation.

jmercouris:
As per report to cl-cffi-gtk, I already made a bug report some time ago: https://github.com/crategus/cl-cffi-gtk/issues/72

jmercouris:
Well, since you already fixed it, and the conditional works on my machine, merge time! :-D

by all means, this is not finished, we'll keep working on it on master, I just don't want to keep this branch open for so long

jmercouris:
Any objections? :-D

Ambrevar:
Works for me!

I guess you can merge now if you like.
I'll keep working on all the nits we mentioned in this thread plus:

- Single instance support.
- Don't quit the REPL when started from it.
- Fix the key doubling.

I guess this won't interfere with the WebEngine IPC so you can safely
work on it in the mean time.

When I'm done with the above I'll start working on a GI IPC.

Cheers!

Ambrevar:
As I said, go ahead! :)

jmercouris:
Yay!!! :-)

PR #573, opened 2020-02-23
Ensure BYTES-TOTAL returns an integer rather than string.

hdasch:
This is intended to fix:
,----
|  <INFO> [14:33:16] download-manager native.lisp (fetch download) - Downloading
|   http://www.informatimago.com/articles/cl-types/cl-types-in-sbcl.ps
| to
|   '/home/hugh/Downloads/cl-types-in-sbcl.ps'.
|
| WARNING:
|    Method handler (NEXT::CORE-OBJECT
|                    NEXT::REQUEST-RESOURCE) signaled an error: The value
|                                                                 "108439"
|                                                               is not of type
|                                                                 NUMBER
|                                                               when binding SB-KERNEL::X.
`----
which occurred when selecting the link "cl-types-in-sbcl.ps" on page
http://www.informatimago.com/articles/cl-types/cl-types-in-sbcl.ps.

The expression "(= 0 (download-manager:bytes-total d))" in
DOWNLOAD-REFRESH suggests that the expected return type is integer
rather than string.

### DIFF
--- a/libraries/download-manager/engine.lisp
+++ b/libraries/download-manager/engine.lisp
@@ -96,8 +96,9 @@ downloads."
    (format nil "~a.part" (namestring (file download)))))
 
 (defmethod bytes-total ((download download))
-  (gethash "content-length"
-           (header download) 0))
+  (let ((bytes (gethash "content-length"
+                        (header download) 0)))
+    (if (stringp bytes) (parse-integer bytes) bytes)))
 
 (defmethod progress ((download download))
   "Return progress ratio.


Ambrevar:
Hmmm, this is strange, I can't reproduce.

`bytes-total` is set from the `content-length` header as per `dex:get`.
Maybe there is a bug in the `dex` library.

How did you install Next?  Is your local installation of `dex` up to date?

hdasch:
Strange indeed.  I'm a neophyte, so this may well be operator error.

I've built from github master at 64fa5fd.  SBCL version 2.0.2.

I found several versions of dexador in my quicklisp tree.  So I:
- Removed old versions
- Ran "make clean" which blew away the quicklisp directory
- Ran "rm -rf ~/.cache/common-lisp/sbcl-2.0.0-linux-x64/" to purge the fasl cache. 
- make next
I still see (gethash "content-length" (header download) 0) return a string.

On the rebuild, quicklisp fetches http://beta.quicklisp.org/archive/dexador/2020-02-18/dexador-20200218-git.tgz, so I guess I'm picking up a version that's less than a week old.

If I run (dex:get "http://www.informatimago.com/articles/cl-types/cl-types-in-sbcl.ps") from the repl then inspect the returned header hash-table, the value of the "content-length" entry is a string rather than an integer.

I notice that when dexador/src/backend/usocket.lisp references content-length, it uses etypecase to ensure the type is coerced to an integer.  That looks more sane than my conditional, but suggests the returned value is a string sometimes, an integer or nil other times.

What am I missing?

Thanks for your help.

Ambrevar:
Nothing, you are absolutely right, I updated dex and it now returns a string for me as well :(
Thanks for the thorough investigation!

PR #588, opened 2020-03-06
README.org: add dependencies install steps for Fedora

Asalle:
I think it will make life of fedora folks easier if we put the whole command that installs dependecies in the readme.


jmercouris:
Thank you!

PR #591, opened 2020-03-07
Gtk abstraction

Ambrevar:
I've extracted the high-level code from the GTK renderer so that it can be factorer with other renderers (e.g. running hooks).

This saves some 50 lines of code in GTK, which is important considering we are saving them from all future renderers.

### DIFF
--- a/source/window.lisp
+++ b/source/window.lisp
@@ -13,6 +13,15 @@
     (lambda (input)
       (fuzzy-match input windows))))
 
+(declaim (ftype (function (window)) window-delete))
+(defun window-delete (window)
+  "This function must be called by the renderer when a window is deleted."
+  (ipc-window-delete window)
+  (next-hooks:run-hook (window-delete-hook window) window)
+  (remhash (id window) (windows *browser*))
+  (when (zerop (hash-table-count (windows *browser*)))
+    (quit)))
+
 (define-command delete-window ()
   "Delete the queried window(s)."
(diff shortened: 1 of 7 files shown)


source/window.lisp
```diff
@@ -13,6 +13,25 @@
     (lambda (input)
       (fuzzy-match input windows))))
 
+(declaim (ftype (function (browser)) window-make))
+(defun window-make (browser)
+  "This function must be called by the renderer when a window is deleted."
```
jmercouris: I assume this is a bit of a typo :-)

jmercouris:
It looks great! Thank you! As I said in my email I would like to use defmethods because there are in fact scenarios with dispatching that may very well be relevant, and since the performance cost is so minuscule, I would prefer it very much.

source/window.lisp
Ambrevar: Oops :p  Good catch!

Ambrevar:
Personally I prefer to use defmethod when I know which are the different implementations.  Otherwise it feels a bit like shooting in the dark.

Anyways, I propose the following: let's merge this and then you can change what you feel like could be a good defmethod (because it's not too clear to me).  What do you think?

jmercouris:
OK, that sounds like a plan to me! Feel free to merge this whenever you like.

Ambrevar:
Merged.

PR #10, opened 2017-10-13
Ensure that ~/.next.d exists

eadmund:
Note that on non-macOS Unix, this should actually be in `$XDG_CONFIG_HOME`, not `~`.  But that's a different story.

Fixes #8.

PR #599, opened 2020-03-10
WIP: Initialize keymap library.

Ambrevar:
Do not merge.

Important changes:

- Moved to a separate library.
- Multiple inheritance.
- `define-key` validates the keyspec at compile time (to be tested... done).
- Prefix keymaps.
- Shorter, simpler code!

Left to do:

- [x] Tests!
- [x] Implement keycode keyspec syntax (e.g. "#10").
- [x] Implement default key translation function.
- [x] Detect cycles in lookup-key.
- [x] List keys?
- [x] List all bindings associated to a symbol.
- [x] Exports.
- [x] Fallback value when not found.
- [x] Integrate.
- [x] Fix dead keys if they still don't work in the minibuffer.
- [x] Test with SBCL 2.0.2 (has stronger warnings)
- [ ] Move `scheme` support to `keymap` library.
- [x] Use new string protocol (all lower case)
- [ ] Implement CUA bindings, make schemes inherit from each other.
- [x] Test key code support (try switching system keyboard layout)
- [x] Document how modifiers can be remapped.


Ambrevar:
@jmercouris This is ready for review!

In a nutshell:
- 200 lines of code remove from source!
- The input system is super simple now (compared to the convoluted stuff we inherited from the RPC architecture).
- SBCL 2.0.2 compatible.
- Lots of new features, including keymap inheritance, prefix keymap, composition, etc.
- Library should be well tested, documented, typed, etc.  Robust and hopefully future-proof.

jmercouris:
Yay!!! I will look at this as soon as I can tomorrow!

Ambrevar:
Thanks!

A few TODO to consider:

- Can a keycode be 0?  I've used 0 as the unset value.

- How shall we call the strings like "C-c C-d"?  I've called them
  `keyspec`s, but what about `keydesc` instead?

- The function `lookup-key` is named after its Emacs counterpart, but
  really it should be `lookup-keys` for consistency with the rest of the
  code.  Thoughts?

- symbol-keys returns an alist of (KEYSPEC KEYMAP), which might be more
  usable than a hash-table since it chains well with #'mapcar, etc.
  Do we really need the KEYMAPs though?  In Next we don't.
  Shall we return them as second VALUE?

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
+(in-package :keymap)
+
+(defstruct modifier
+  (string "" :type string)
+  (shortcut "" :type string))
+
+(declaim (ftype (function ((or string modifier) (or string modifier)) boolean) modifier=))
+(defun modifier= (string-or-modifier1 string-or-modifier2)
```
jmercouris: Is there a possibility to just always convert to a string for comparison?

jmercouris: I am just thinking it may simplify the logic, but there are still many things I could be missing.

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+      ((stringp string-or-modifier2)
+       (match-modifier string-or-modifier1 string-or-modifier2))
+      (t
+       (or (string= (modifier-string string-or-modifier1)
+                    (modifier-string string-or-modifier2))
+           (string= (modifier-shortcut string-or-modifier1)
+                    (modifier-shortcut string-or-modifier2)))))))
+
+(defmethod fset:compare ((x modifier) (y modifier))
+  "Needed to user the KEY structure as keys in Fset maps."
```
jmercouris: Probable typo on "user"

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+  (fset:compare-lexicographically (modifier-string x) (modifier-string y)))
+
+(defvar +control+ (make-modifier :string "control" :shortcut "C"))
+(defvar +meta+ (make-modifier :string "meta" :shortcut "M"))
+(defvar +shift+ (make-modifier :string "shift" :shortcut "s"))
+(defvar +super+ (make-modifier :string "super" :shortcut "S"))
+(defvar +hyper+ (make-modifier :string "hyper" :shortcut "H"))
+
+(defparameter *modifier-list*
```
jmercouris: Probably also want to make this +modifier-list+ since it is not meant to be mutated?

jmercouris: Or, is the idea that people could append any modifiers that they want to this list?

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+  (list +control+ +meta+ +shift+ +super+ +hyper+)
+  "List of known modifiers.
+`make-key' and `define-key' raise an error when setting a modifier that is not
+in this list.")
+
+(deftype key-status-type ()
+  `(or (eql :pressed) (eql :released)))
+
+;; Must be a structure so that it can served as a key in a hash-table with
```
jmercouris: Is the problem of the case insensitivity of equalp solved elsewhere?

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+;; #'equalp tests.
+(defstruct (key (:constructor %make-key (code value modifiers status))
+                (:copier %copy-key))
+  (code 0 :type integer) ; TODO: Can a keycode be 0?  I think not, so 0 might be a good non-value.
+  (value "" :type string)
+  (modifiers (fset:set) :type fset:wb-set)
+  (status :pressed :type key-status-type))
+
+(defun key= (key1 key2)
+  "Two keys are equal if the have the same modifiers, status and key code.
```
jmercouris: Is the value the printed representation of the character? I assume based on the struct, but I am not fully sure :-D

libraries/keymap/keymap.lisp
jmercouris: Or would it be possible to use `modspec->modifier`?

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+                                (:modifiers list) (:status keyword))
+                          key)
+                make-key))
+(defun make-key (&key (code 0 explicit-code) (value "" explicit-value)
+                      modifiers
+                      (status :pressed))
+  "Return new `key'.
+Modifiers can be either a `modifier' type or a string that will be looked up in
+`*modifier-list*'."
+  ;; TODO: Display warning on duplicate modifiers?
```
jmercouris: it would be nice I guess, but how far do we go? There are an infinite amount of error permutations

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+in this list.")
+
+(deftype key-status-type ()
+  `(or (eql :pressed) (eql :released)))
+
+;; Must be a structure so that it can served as a key in a hash-table with
+;; #'equalp tests.
+(defstruct (key (:constructor %make-key (code value modifiers status))
+                (:copier %copy-key))
+  (code 0 :type integer) ; TODO: Can a keycode be 0?  I think not, so 0 might be a good non-value.
```
jmercouris: I was not able to find anything with a value of 0 other than Nul, which I don't think one can enter

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+   code
+   value
+   (modspecs->modifiers modifiers)
+   status))
+
+(declaim (ftype (function (key &key (:code integer) (:value string)
+                               (:modifiers fset:wb-set) (:status keyword))
+                          key)
+                copy-key))
+(defun copy-key (key &key (code (key-code key)) (value (key-value key))
```
jmercouris: I assume this exists because %copy-key does not recurse (as far as I understand based on the Hyperspec)?

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+             (string= "#" (subseq code-or-value 0 1)))
+        (setf code (or (parse-integer code-or-value :start 1 :junk-allowed t)
+                       code))
+        (setf value code-or-value))
+    (make-key :code code :value value :modifiers modifiers)))
+
+(declaim (ftype (function (string) list-of-keys) keyspecs->keys))
+(defun keyspecs->keys (spec)
+  "Parse SPEC and return corresponding list of keys."
+  ;; TODO: Return nil if SPEC is invalid?
```
jmercouris: A condition will be raised upon each invalid key, should the user pick a restart that allows them to return a valid list of keys, then it should be OK, no?

jmercouris: s/invalid key/invalid keyspec->key invokation

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+
+(defun translate-remove-shift (keys)
+  "With shift, keys without shift: 'shift-a' -> 'a'."
+  (let ((shift? (find +shift+ keys :key #'key-modifiers :test #'fset:find)))
+    (when shift?
+      (mapcar (lambda (key)
+                (copy-key key :modifiers (fset:less (key-modifiers key) +shift+)))
+              keys))))
+
+(defun translate-remove-but-first-control (keys)
```
jmercouris: What if we want to bind C-x C-c and C-x c to different things? I guess it doesn't matter since it doesn't appear in the default translator, maybe some people will want it :-)

source/renderer-gtk.lisp
```diff
@@ -86,12 +86,8 @@
     (gobject:g-signal-connect
      gtk-object "key_press_event"
      (lambda (widget event) (declare (ignore widget))
-       (push-modifier *browser* event)
        (process-key-press-event window event)))
-    (gobject:g-signal-connect
-     gtk-object "key_release_event"
-     (lambda (widget event) (declare (ignore widget))
-       (pop-modifier *browser* event)))
+    ;; REVIEW: No need to handle release events?
```
jmercouris: Correct, no need to handle release events. Insofar as I can tell most UI elements (exception being buttons), the click on matters, not the click off.

source/renderer-gtk.lisp
```diff
@@ -148,63 +153,39 @@
...
+      (dispatch-input-event event (active-buffer sender) sender))))
 
 (defmethod process-button-press-event ((sender gtk-buffer) event)
   (let* ((button (gdk:gdk-event-button-button event))
-         (x (gdk:gdk-event-button-x event))
-         (y (gdk:gdk-event-button-y event))
-         (key-string (format nil "button~s" button))
-         (window (find sender (window-list)
-                       :key #'active-buffer)))
+         ;; REVIEW: No need to store X and Y?
```
jmercouris: I do think we should remove it since we are not using it and have not yet found cause to use it, yes

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+                  (list #'translate-remove-shift-toggle-case
+                        #'translate-remove-shift
+                        #'translate-remove-but-first-control
+                        #'translate-remove-shift-but-first-control
+                        #'translate-remove-shift-but-first-control-toggle-case))))
+
+(defvar *default-translator* #'translate-shift-control-combinations
+  "Default key translator to use in `keymap' objects.")
+
+;; TODO: Enable override of default and translator in lookup?
```
jmercouris: Yes, I can think of when it would be useful in many cases! :-)

jmercouris: That said, someone could just use `let`... but maybe if it is sanctioned with an &key people will feel more comfortable doing it

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+                  :translator (or translator *default-translator*)
+                  ;; We cannot use the standard (make-hash-table :test #'equalp)
+                  ;; because then (set "a") and (set "A") would be the same thing.
+                  :entries (fset:empty-map default))
+   'keymap))
+
+(defun keymap-p (object)
+  (typep object 'keymap))
+
+(deftype keyspecs-type ()               ; TODO: Rename to KEYDESC?
```
jmercouris: In my opinion it is a specification not a description. I could describe the keys in a number of ways, but I can only provide an official representation via a specification. The connotation being that a specification is a logically sound description, a subset of a description.

libraries/keymap/keymap.lisp
```diff
@@ -0,0 +1,556 @@
...
+  `(satisfies keyspecs->keys))
+
+;; We need a macro to check that bindings are valid at compile time.
+;; This is because most Common Lisp implementations or not capable of checking
```
jmercouris: s/or not/are not

jmercouris:
Well, well! Pierre! I must say a most fantastic job! The functions are clear, simple and concise! Easy to follow logic! Even though I did not have you walk me through the codebase, I still understand a large part of it. There are some things I must learn like `fset` before I can leave a full review.

Though I'm not yet fully understanding this part of the codebase, I left some preliminary comments to start a discussion.

Lastly (for now), there is a form that does not compile: 
```
../../source/vi-mode.lisp:37:7:
  error: 
    during macroexpansion of (DEFINE-KEY :KEYMAP MAP ...). Use *BREAK-ON-SIGNALS*
    to intercept.
    
     The value
       MAP
     is not of type
       STRING
     when binding KEYMAP::SPEC
```

jmercouris:
Some quick thoughts:
>- The function `lookup-key` is named after its Emacs counterpart, but really it should be `lookup-keys` for consistency with the rest of the code. Thoughts?

Let's go with `lookup-keys`, we do not have to be 1:1 Emacs, we already differ in so many concepts

> - symbol-keys returns an alist of (KEYSPEC KEYMAP), which might be more usable than a hash-table since it chains well with #'mapcar, etc. Do we really need the KEYMAPs though? In Next we don't. Shall we return them as second VALUE?

A second VALUE should be fine in this case. Again though, don't forget about `maphash` :-)

libraries/keymap/keymap.lisp
Ambrevar: Can't use modspec->modifier because it relies on `modifier=`.

Converting to string would result in the same logic I think.  If you can think of a simplification, please share it :)

libraries/keymap/keymap.lisp
Ambrevar: Fixed.

libraries/keymap/keymap.lisp
Ambrevar: That latter indeed!

libraries/keymap/keymap.lisp
Ambrevar: See `fset:compare key key`.

libraries/keymap/keymap.lisp
Ambrevar: Yes, couldn't think of a better name... :(

libraries/keymap/keymap.lisp
Ambrevar: I think you misunderstood what I mean: we just need to report extraneous modifiers, e.g. in ("C" "C").
Anyways, fixed in e195556070102165a935479b6c93346d93147082.

libraries/keymap/keymap.lisp
Ambrevar: Thanks.

libraries/keymap/keymap.lisp
Ambrevar: Not exactly: the copier is automatically defined by `defstruct`, so if I want to override it, I need to give vanilla copier another name and implement my own `copy-key` that uses it.

libraries/keymap/keymap.lisp
Ambrevar: I'm not sure I understood what you meant: that we added a restart?
I'm not sure a restart is too useful here though.  This function is internal to `keymap` and it's only called by `define-key`, which is a macro and thus detects errors at compile time.  Restarts are not very useful during compilation.

libraries/keymap/keymap.lisp
Ambrevar: It's supported: translation only happens if the exact keyspec does not match.

libraries/keymap/keymap.lisp
Ambrevar: `let` would not work here since `default` and `translator` are `keymap` slots.
I'll use a keyword argument.

libraries/keymap/keymap.lisp
Ambrevar: Fair enough, thanks for making a point here! :)

Ambrevar:
vi-mode: Oops!  Fixed.  Weird I didn't see it though :p 

lookup-keys: OK.

maphash: The big drawback of maphash is that it does not return a value.  In general it's nice to return values that can be passed to higher-order functions.
I think I'll go with a (VALUES LIST ALIST) return value.

Ambrevar:
Actually about lookup-key: we have define-key without `s`, so lookup-key is consistent with that at least.

libraries/keymap/keymap.lisp
jmercouris: Right, but my question is, why did you want to override it?

libraries/keymap/keymap.lisp
jmercouris: We don't have to add a restart, the user could make some changes and use one of the default restarts, I could for example edit the define-key form and attempt a recompile

libraries/keymap/keymap.lisp
jmercouris: Ah, whoops :-)

libraries/keymap/keymap.lisp
Ambrevar: Then the current code behaves well, doesn't it?

libraries/keymap/keymap.lisp
Ambrevar: To support modspecs argument.

Ambrevar:
Fset tutorial: https://common-lisp.net/project/fset/Site/FSet-Tutorial.html

jmercouris:
By the way, feel free to merge this whenever you feel ready, in so far as I am concerned, the design is fully sound! The only things I could make comments would be trivial stylistic choices.

libraries/keymap/keymap.lisp
jmercouris: Yes, that is what I was saying, I think it just fine!

Ambrevar:
I've made some structural changes, so I'd like you to let me know what you think
before I merge:

- Made keymap-local translators  global: indeed, keymap-specific translators
  would not do the right thing (e.g. "g g" and "G" are bound to different
  keymaps, "g g" is in the first keymap, "G" in the second, if the user presses
  "s-G" it used to hit "g" instead of "G").
  
- Removed DEFAULT slot since we are not using it, plus since LOOKUP-KEY supports
  multiple keymaps as argument, there is no intuitive way to use DEFAULT.

- Added a mandatory NAME slot for keymaps.  This is useful to describe the
  keymap.
  Try "command-inspect" on a command that has bindings!

Also great news, dead keys are finally supported in the minibuffer.

documents/CHANGELOG.org
```diff
@@ -39,6 +39,15 @@ MAJOR.MINOR.PATCH, we increment the:
   of a particular text field
 *** Add online suggestions from search engines
 
+** 2.0.0
+- Change key string specifications: all lowercase without underscore.
+  For instance "Page_Up" becomes "pageup".
```
jmercouris: I think it should be case insensitive actually if the user has C-SPACE or C-space it should not matter. Thoughts?

Ambrevar:
The problem is that key "A" is different from "a".  So if we want to be
case insensitive for only some keys, we need a way to distinguish them.

I guess the criteria here is "printability".  Although it does not work
for "space", "hyphen", "tab" and maybe a few others which are printable.

The drawbacks may outweigh the benefits.

source/start.lisp
```diff
@@ -201,11 +201,8 @@ EXPR must contain one single Lisp form. Use `progn' if needed."
                                       :omit-nulls t)
                       ((guard urls urls)
                        (log:info "External process requested URL(s): ~{~a~^, ~}" urls)
-                       ;; `open-urls` must be done on the GTK thread.
-                       (gdk:gdk-threads-add-idle (lambda ()
-                                                   (open-urls urls)
-                                                   ;; Return nil so that the callback is called only once.
-                                                   nil)))
+                       (gtk:within-gtk-thread
```
jmercouris: I am pretty sure this will cause freezing on macOS

jmercouris:
> The problem is that key "A" is different from "a". So if we want to be case insensitive for only some keys, we need a way to distinguish them. I guess the criteria here is "printability". Although it does not work for "space", "hyphen", "tab" and maybe a few others which are printable. The drawbacks may outweigh the benefits.

I know you are pro lower case, but I am pro upper case, it would seem to me that the best compromise would be using the lisp printer settings :-D

jmercouris:
> I've made some structural changes, so I'd like you to let me know what you think
> before I merge:
> 
> * Made keymap-local translators  global: indeed, keymap-specific translators
>   would not do the right thing (e.g. "g g" and "G" are bound to different
>   keymaps, "g g" is in the first keymap, "G" in the second, if the user presses
>   "s-G" it used to hit "g" instead of "G").

Ah, I did not even think of this potential conflict! In any case it will now allow for let binding it no?

> * Removed DEFAULT slot since we are not using it, plus since LOOKUP-KEY supports
>   multiple keymaps as argument, there is no intuitive way to use DEFAULT.

Agreed

> * Added a mandatory NAME slot for keymaps.  This is useful to describe the
>   keymap.

This is a good change!

> Also great news, dead keys are finally supported in the minibuffer.

Yay!

source/start.lisp
jmercouris: Actually I can't even find the symbol within-gtk-thread in cl-cffi-gtk

jmercouris: I changed it to within-main-loop, doesn't appear to be causing problems

jmercouris:
Just tested on macOS, I cannot type M-x, M->, etc, so I cannot test describe-* :-\

source/start.lisp
Ambrevar: It's in GTK init.
It will work on macOS, because the macro does exactly what the previous code did:
```
(defmacro within-gtk-thread (&body body)
  "Executes the body within the GTK thread. Unlike GTK:WITHIN-MAIN-LOOP, this
macro does not initialize GTK, and can be used several times in a program. It
is mainly useful to execute some GTK code from a thread different from the main
GTK thread."
  `(gdk:gdk-threads-add-idle ; Could probably be replaced by glib:g-idle-add, but this is safer (see GDK Threads doc)
    #'(lambda ()
        ,@body
        nil)))
```

Ambrevar:
Can you report precisely what does not work?  If you enable debug log, you should see exactly what's happening.

Also check `translate-modifiers`, I've updated the function a bit, maybe I broke something with the magic macOS modifier number.

Ambrevar:
> I know you are pro lower case, but I am pro upper case, it would seem to me that the best compromise would be using the lisp printer settings :-D

Unfortunately these are strings, not symbols, so the Lisp printer does
not intervene here.

source/start.lisp
Ambrevar: I meant `gtk-init.lisp` from the CFFI-GTK library.

Ambrevar:
> Ah, I did not even think of this potential conflict! In any case it will now allow for let binding it no?

Correct!

source/start.lisp
Ambrevar: You _cannot_ use within-main-loop here, look at the documentation of `within-gtk-thread`:
```
  "Executes the body within the GTK thread. Unlike GTK:WITHIN-MAIN-LOOP, this
macro does not initialize GTK, and can be used several times in a program. It
is mainly useful to execute some GTK code from a thread different from the main
GTK thread."
```

Ambrevar:
> Just tested on macOS, I cannot type M-x, M->, etc, so I cannot test describe-* :-\

If you have a init.lisp, make sure it's compatible with the changes :p

Ambrevar:
I've pushed some last tweaks.  This is ready to merge as far as I'm concerned.
Let me know if it works for you and when it does, let's merge!

source/start.lisp
jmercouris: Does not exist in my version of cl-cffi-gtk

jmercouris:
nothing is in my init file.

jmercouris:
Here's what it does when I type in M-x:

Screen Shot 2020-03-20 at 18 09 15

Also right clicking crashes the program:

```
debugger invoked on a TYPE-ERROR in thread
#<THREAD "main thread" RUNNING {1000560083}>:
  The value
    0
  is not of type
    LIST
  when binding NEXT::MODIFIER-STATE

  0: [RETURN-FROM-G-CLOSURE] Return value from closure
  1: [ABORT                ] Exit debugger, returning to top level.

(NEXT::TRANSLATE-MODIFIERS 0 #S(GDK:GDK-EVENT-BUTTON :TYPE :BUTTON-PRESS :WINDOW #<GDK:GDK-WINDOW {1005E3E383}> :SEND-EVENT NIL :TIME 4176280826 :X 273.75d0 :Y 459.62109375d0 :AXES #(273.75d0 459.62109375d0) :STATE 0 :BUTTON 3 :DEVICE #<GDK:GDK-X11-DEVICE-XI2 {1005F3D883}> :X-ROOT 273.75d0 :Y-ROOT 481.62109375d0)) [external]
   source: (DEFUN TRANSLATE-MODIFIERS (MODIFIER-STATE &OPTIONAL EVENT)
             "Return list of modifiers fit for `keymap:make-key'.
See `gtk-browser's `modifier-translator' slot."
             (LET ((MODIFIERS NIL)
                   (MACINTOSH-ALT-KEY-VALUE 65406)
                   (KEY-VALUE
                    (MATCH (TYPE-OF EVENT) ('GDK:GDK-EVENT-KEY #) (_ 0))))
               (WHEN (MEMBER :CONTROL-MASK MODIFIER-STATE)
                 (PUSH "control" MODIFIERS))
               (WHEN
                   (OR (MEMBER :MOD1-MASK MODIFIER-STATE)
                       (= KEY-VALUE MACINTOSH-ALT-KEY-VALUE))
                 (PUSH "meta" MODIFIERS))
               (WHEN (MEMBER :SHIFT-MASK MODIFIER-STATE)
                 (PUSH "shift" MODIFIERS))
               (WHEN
                   (OR (MEMBER :SUPER-MASK MODIFIER-STATE)
                       (AND (MEMBER :MOD2-MASK MODIFIER-STATE)
                            (MEMBER :META-MASK MODIFIER-STATE)))
                 (PUSH "super" MODIFIERS))
               (WHEN (MEMBER :HYPER-MASK MODIFIER-STATE)
                 (PUSH "hyper" MODIFIERS))
               MODIFIERS))
0] 
```

source/start.lisp
jmercouris: I had to update to the one on Quicklisp, I was using the one from crategus cloned locally

jmercouris:
Here's the debug output:
```
NEXT> (log:config :debug)
; No value<DEBUG> [18:14:54] next renderer-gtk.lisp (process-key-press-event gtk-window) -
  NEXT::KEY-STRING: "arabicswitch" NEXT::KEYCODE: 66 CHARACTER: #\Nul
  NEXT::KEYVAL-NAME: "Arabic_switch" NEXT:MODIFIERS: ("meta") 
<DEBUG> [18:14:54] next input.lisp (dispatch-input-event) -
  Forward key "M-arabicswitch (M-#66)"
<DEBUG> [18:14:54] next renderer-gtk.lisp (process-key-press-event gtk-window) -
  NEXT::KEY-STRING: "x" NEXT::KEYCODE: 15 CHARACTER: #\x NEXT::KEYVAL-NAME: "x" 
<DEBUG> [18:14:54] next input.lisp (dispatch-input-event) -
  Forward key "x (#15)"
<DEBUG> [18:14:55] next renderer-gtk.lisp (process-key-press-event gtk-window) -
  NEXT::KEY-STRING: "arabicswitch" NEXT::KEYCODE: 66 CHARACTER: #\Nul
  NEXT::KEYVAL-NAME: "Arabic_switch" NEXT:MODIFIERS: ("meta") 
<DEBUG> [18:14:55] next input.lisp (dispatch-input-event) -
  Forward key "M-arabicswitch (M-#66)"
```

Ambrevar:
Clean the .fasls (CFFI and Next), fetch the latest commit from this
branch and rebuild.

> NEXT> (log:config :debug)
> ; No value<DEBUG> [18:14:54] next renderer-gtk.lisp (process-key-press-event gtk-window) -
>   NEXT::KEY-STRING: "arabicswitch" NEXT::KEYCODE: 66 CHARACTER: #\Nul
>   NEXT::KEYVAL-NAME: "Arabic_switch" NEXT:MODIFIERS: ("meta") 

Which key did you press?

jmercouris:
I pressed the "alt" key, as in "alt-x" or "M-x"

Ambrevar:
> Here's what it does when I type in M-x:

Are you saying that typing "M-x" when the browser starts brings up the
"command-inspect" view of execute-command?

jmercouris:
> Here's what it does when I type in M-x:
> Are you saying that typing "M-x" when the browser starts brings up the "command-inspect" view of execute-command?

No, I typed in `C-h k`, then I typed in `M-x` so that you could see how it reports Meta/Alt on my machine.

jmercouris:
I cleaned all fasls, recompiled, issue persists unfortunately :-(

Ambrevar:
> No, I typed in `C-h k`, then I typed in `M-x` so that you could see how it reports Meta/Alt on my machine.
>

Then key-inspect works for you! :)

Ambrevar:
> I pressed the "alt" key, as in "alt-x" or "M-x"

I've just added Arabic_switch in the last commit, try again.

For crash with click and left/right, please report for each input key:

- the key,
- the debug output or the trace.

Ambrevar:
Which issue?

source/utility.lisp
```diff
@@ -96,60 +78,11 @@ The second value is the initfunction."
...
-(declaim (ftype (function (fixnum &optional fixnum)) kill-program))
-(defun kill-program (pid &optional (signal 15))
-  (handler-case (uiop:run-program
-                 (list "kill" (format nil "-~a" signal)
-                       (write-to-string pid)))
-    (error ()
-      (log:error "Process with PID ~a is not running" pid))))
-
-(declaim (ftype (function (string &rest string)) run-program-to-string))
-(defun run-program-to-string (program &rest args)
```
jmercouris: Is this not used anywhere?

jmercouris:
Still no luck:

output:

```
<DEBUG> [19:34:38] next renderer-gtk.lisp (process-key-press-event gtk-window) -
  NEXT::KEY-STRING: NIL NEXT::KEYCODE: 66 CHARACTER: #\Nul
  NEXT::KEYVAL-NAME: "Arabic_switch" NEXT:MODIFIERS: ("meta") 
<DEBUG> [19:34:38] next renderer-gtk.lisp (process-key-press-event gtk-window) -
  NEXT::KEY-STRING: "x" NEXT::KEYCODE: 15 CHARACTER: #\x NEXT::KEYVAL-NAME: "x" 
<DEBUG> [19:34:38] next input.lisp (dispatch-input-event) -
  Forward key "x (#15)"
<DEBUG> [19:34:38] next renderer-gtk.lisp (process-key-press-event gtk-window) -
  NEXT::KEY-STRING: "x" NEXT::KEYCODE: 15 CHARACTER: #\x NEXT::KEYVAL-NAME: "x" 
<DEBUG> [19:34:38] next input.lisp (dispatch-input-event) -
  Forward key "x (#15)"
```

now it just acts as if I am not pressing anything

jmercouris:
In other words, I do `C-h k`, and then I type in `M-x`, but it just prints `x` with no modifiers in the echo area

source/utility.lisp
Ambrevar: Nope :p

Ambrevar:
Looks like Meta (Alt) is not a modifier on macOS.
Is this why you implemented the push-modifiers / pop-modifiers functions before?
Can you check if 87b57377aedde7e161c30601c31bd6549fee0391
is the first commit where meta stops working for you?

Regarding your click issue, I think I've fixed it in the last commit (please pull).

jmercouris:
> Looks like Meta (Alt) is not a modifier on macOS.
> Is this why you implemented the push-modifiers / pop-modifiers functions before?
> Can you check if [87b5737](https://github.com/atlas-engineer/next/commit/87b57377aedde7e161c30601c31bd6549fee0391)
> is the first commit where meta stops working for you?
> 
> Regarding your click issue, I think I've fixed it in the last commit (please pull).

that's correct, thats why I implemented a function to push/pop modifiers :-)

it is a bug in GTK. You cannot query which modifiers exist (reliably), I thought I had documented that, sorry if I forgot!

Ambrevar:
Can you tell me which modifiers needs this form of stack?
Control seems to be fine.   How is Super?

jmercouris:
Right click and left click are now working! :-)

jmercouris:
I believe it is Meta, and Super. Only Control appears to behave reliably

jmercouris:
> Looks like Meta (Alt) is not a modifier on macOS.
> Is this why you implemented the push-modifiers / pop-modifiers functions before?
> Can you check if [87b5737](https://github.com/atlas-engineer/next/commit/87b57377aedde7e161c30601c31bd6549fee0391)
> is the first commit where meta stops working for you?

I can confirm that this is the first commit in which meta does not work.

jmercouris:
> I know you are pro lower case, but I am pro upper case, it would seem to me that the best compromise would be using the lisp printer settings :-D
> Unfortunately these are strings, not symbols, so the Lisp printer does not intervene here.

I'm know, I'm saying we let the lisp printer settings decide how we evaluate the strings

Ambrevar:
> I'm know, I'm saying we let the lisp printer settings decide how we evaluate the strings

In any case this is an easy fix, we can always do this later.

Ambrevar:
I've just pushed a fix for the modifiers on macOS.  Let me know if it works for you.

Ambrevar:
https://github.com/atlas-engineer/next/pull/599/commits/1eb5a1c9342465c20aa912cb4eb7ca974c3982a9

jmercouris:
Yay! It has fixed it!!!

I should also point out that bindings for:

```
"C-[" #'switch-buffer-previous
"C-]" #'switch-buffer-next
```

are no longer functioning because it now reports that as C-bracketleft and C-bracketright, not sure if that is intended or not. If it is intended, then we just need to change those bindings, and it should be ready for merging!

Ambrevar:
Actually all "C-" + punctuation bindings were broken.  Should be fixed now.

I've also cleaned up the macOS workaround, can you test again?   It should allow you to remap the modifiers very easily (example in the doc).

jmercouris:
Everything working! Yay :-)

Ambrevar:
Thanks!

jmercouris:
More bugs :-D, C-space does not work, does not actually do anything

Ambrevar:
C-space fixed.

Question: the keymap library supports "-" as a key value.
E.g. "-" and "C--" are valid.

Shall we still use "hyphen" or shall we use "-"?  The latter would
remove one exception and be more consistent with Emacs.

jmercouris:
C-- looks so much like a typo that I think C-hyphen is better, but either is OK with me

Ambrevar:
OK, C-hyphen it is then!
Big merge inbound! :)

jmercouris:
Let's go!!! :-)

jmercouris:
Can we close this pull request since it appears merged on master?

Ambrevar:
Yup!

PR #600, opened 2020-03-11
Instantiate windows from *window-class* parameter.

jellelicht:


Ambrevar:
Good catch!  And sorry if this bug wasted your time :p

PR #601, opened 2020-03-12
Add dependency on dconf under Guix

### LINKED ISSUE #592: TLS/SSL support not available; install glib-networking
**Describe the bug**
When I try to open a HTTPS URL, I get the error message "TLS/SSL support not available; install glib-networking".

- Start `next`
- Open any HTTPS URL

**Information**
- OS name+version: Guix commit 1c9e3e9f36621bb91d153bc155c6cbce704e9220
- Kernel name+version:
- Graphics card and driver:
- Desktop environment / Window manager name+version: None. I ssh into Guix from macOS running XQuartz.
- Next version (from =M-x next-version= or =next --version=): commit e3e3d2bfa003f0b1d4e428b1d6e9a2eff9129b99
- How you installed Next (Guix pack, package manager, build from source): I run Next from source without installation
- SBCL version (if built from source): 2.0.2
- WebKitGTK+ / PyQt-Webengine version:

khinsen:
Fixes #592

### DIFF
--- a/build-scripts/guix.scm
+++ b/build-scripts/guix.scm
@@ -226,6 +226,7 @@ key-bindings and is fully configurable and extensible in Common Lisp.")
          ("unix-opts" ,sbcl-unix-opts)
          ;; WebKitGTK deps
          ("glib-networking" ,glib-networking)
+         ("dconf" ,dconf)
          ("gsettings-desktop-schemas" ,gsettings-desktop-schemas)
          ;; Local deps
          ("next-download-manager" ,sbcl-next-download-manager)


Ambrevar:
Will test as soon as possible, thanks for investigating!

Ambrevar:
Finally had time to review this! :)

glib is the package that sets the GIO_EXTRA_MODULES search path.
glib is a propagated dependency of dconf.  So I presume that the right
input to add here is glib, not dconf.

Furthermore, I don't understand how adding glib/dconf as an _input_
could change anything for you: since the package is not in the profile,
the search path won't be updated.  I just tried it on my machine.

Instead, adding glib as a propagated-input does the trick for me: I now
have

export GIO_EXTRA_MODULES=\
...

in my ~/.guix-profile/etc/profile.

That said, none of the browsers in web-browsers.scm sets propagated-inputs.
Is this a mistake on Guix' side?

What shall we do?  Any more clue?
Otherwise I'll ask upstream.

khinsen:
Just checked, I had `dconf` as a propagated input. But `glib` does the job as well.

To me this looks like a Guix issue. `GIO_EXTRA_MODULES` becomes necessary only because Guix doesn't use the standard directory layout. But I don't know the details of how Guix handles such environment variables.

Ambrevar:
Epiphany propagates the dconf input.
But vimb does not and it seems to work nonetheless.  I wonder how.

Ambrevar:
Got it: it's because vimb uses the glib-or-gtk-build-system, which
automatically creates the "vimb" wrapper:

#!/gnu/store/29jhbbg1hf557x8j53f9sxd9imlmf02a-bash-minimal-5.0.7/bin/bash
export XDG_DATA_DIRS="/gnu/store/gb2v82aflbxjbq1rkgb56xd07f3s01x3-shared-mime-info-1.10/share:/gnu/store/b8pr2k0i2zd07zmb7kpffmcimqi337if-glib-2.60.6/share:/gnu/store/8m64m3f1s3rdib1aqj2c2zq6gciq3kr5-gtk+-3.24.14/share:/gnu/store/igp43vz0zkmbp7qcrsgswjim2x3nxiry-gsettings-desktop-schemas-3.32.0/share${XDG_DATA_DIRS:+:}$XDG_DATA_DIRS"
export GTK_PATH="/gnu/store/8m64m3f1s3rdib1aqj2c2zq6gciq3kr5-gtk+-3.24.14/lib/gtk-3.0${GTK_PATH:+:}$GTK_PATH"
export GIO_EXTRA_MODULES="/gnu/store/yb95bx7wzifkhydy9v4z5655x1l0cqg5-glib-networking-2.60.3/lib/gio/modules${GIO_EXTRA_MODULES:+:}$GIO_EXTRA_MODULES"
exec -a "$0" "/gnu/store/g7gc04h2k9069da3wgk2f9j943c8kjyf-vimb-3.6.0/bin/.vimb-real" "$@"

I'll do something similar for next to avoid the propagated input.

Ambrevar:
Fixed in ebd569858479af709f58f71a349449846de647eb.

PR #602, opened 2020-03-14
session.lisp: prevent restoring nil histories.

jellelicht:
observe the `history-tree:data` being passed `nil` when running e.g. `next "https://google.nl"

### DIFF
--- a/source/session.lisp
+++ b/source/session.lisp
@@ -61,6 +61,7 @@ Currently we store the list of current URLs of all buffers."
           ((list version buffer-histories)
            (unless (string= version +version+)
              (log:warn "Session version ~s differs from current version ~s" version +version+))
+           (setf buffer-histories (delete-if-not #'htree:current buffer-histories))
            (when buffer-histories
              (log:info "Restoring ~a"
                        (mapcar #'object-string (mapcar (alexandria:compose #'htree:data #'htree:current)


Ambrevar:
Good catch, but a few comments:

- Why #'htree-root?  Shouldn't it be #'htree-data?  Sorry I don't have the code base here and I haven't touched this in a while.
- You probably want to alter buffer-histories before so that the log:info does not print the NILs either.
- Use delete-if-not to modifier the list in place.

jellelicht:

> Good catch, but a few comments:
>
> - Why #'htree-root?  Shouldn't it be #'htree-data?  Sorry I don't have the code base here and I haven't touched this in a while.
#'history-tree:root and #'history-tree:current both return nil in these
cases. The code subsequently calls #'history-tree:data on the nil from
#'history-tree:current, which is the cause of the warning. So the
problem is not that there is no data, the problem is that we have a
history-tree that is unrooted (root = nil) and has no current node (current
= nil).

> - You probably want to alter buffer-histories before so that the log:info does not print the NILs either.
> - Use delete-if-not to modifier the list in place.
Done!

Ambrevar:
This is good to merge because it fixes the issue we could have when ~/.local/share/next/session.lisp does not return a list of htree objects.

That said, this error should not occur in the first place unless the file is corrupted.
@jellelicht Do you mind sharing the content of the file assuming it triggers the issue without the patch?

Ambrevar:
Hmm, if you can't reproduce then I must guess there must be something else.  I don't see how this would be related to the other window fix.
Oh well... :/

Ambrevar:
Thanks for the fix!

PR #603, opened 2020-03-14
start.lisp: Only open one window.

jellelicht:
If I pass a list of urls to next, I do not want a semi-functional window that has the help buffer visible.

### DIFF
--- a/source/start.lisp
+++ b/source/start.lisp
@@ -178,11 +178,11 @@ EXPR must contain one single Lisp form. Use `progn' if needed."
 (defun default-startup (&optional urls)
   "Make a window and load URLS in new buffers. This function is
    suitable as a `browser' `startup-function'."
-  (when urls
-    (open-urls urls))
   (let ((window (window-make *browser*))
         (buffer (help)))
-    (window-set-active-buffer window buffer))
+    (if urls
+        (open-urls urls)
+        (window-set-active-buffer window buffer)))
   (match (session-restore-function *browser*)
     ((guard f f)
      (when *use-session*


Ambrevar:
Thanks for this, it had been annoying me for the past few days!
Actually I see that you are reverting what @jmercouris did in afa549600f1e922ce4fdb61f5cbd0542e3844e3b.  @jmercouris Is there a good reason for the `when` or was it a mistake?

jmercouris:
It was not a mistake, I guess I had misunderstood what you had wanted. I assumed that the user would *always* want the help until such a time as they were to customize the default-startup-function. I'm not sure that this change would be a good one.

Ambrevar:
You may have misunderstood the previous function:
The condition on URLS is for when the browser is started with `next
foo-url`.

Previously, the function used to do the right thing because `open-urls`
opens new windows on demand (see `open-external-link-in-new-window-p`).

Now it opens two windows on startup, including an empty one which is
not useful I think and may be confusing.

What do you think?  Does that make sense?

Regarding the offending commit, ideally it's nice to have
separate, atomic commits with a message that explains the intent.  Thanks!

jmercouris:
The problem is actually in my opinion then the behavior of `open-urls` which should by default not make a new window. That is probably why I expected the behavior to be different (resulting in only one window).

As per the commit, it could be better. With this kind of large scale refactoring, I do not make commits per change, but per functioning revision. In this case, I had to edit this (I can't remember why) as part of getting windows to show up.

Anyways, in the meantime, I am OK with this change, but in the long-term my vote is for changing the behavior of `open-urls`.

jmercouris:
Actually, maybe the behavior should be by default to open a new window. I don't know, but in this case, I don't think it should open new windows.

jmercouris:
To summarize, one should be able to pass a key to open-urls called something like :open-new-window-p to handle this behavior. I do already see the open-external-link-in-new-window-p, but I think it should be possible to pass it in instead of having to let that variable or something. That is, should we use a `let` to bind that temporarily, we would be misusing the variable (as it is currently named), in my opinion. So I would probably do something like this:

```
(defun open-urls (urls &key no-focus (open-new-window-p nil open-new-window-supplied-p))
...)
```

jmercouris:
Thank you @jellelicht !!

Ambrevar:
Hmm... I think there is still a misunderstanding.  Or maybe we want different things.

The `open-urls' function is the one that gets called when Next is told to open
URLS _on the command line_.  For instance in a shell you can write `next
foo-url`, or when clicking on a URL from an external program.

If Next is not running, then it will start up with the given URL.  If Next is
already running, the running instance is told to load the given URL.

Either way, it's impossible to contextually tell the browser whether the URL
should be open in a new window or not.  This can only be configured at the
global level.

Does that make sense?

Ambrevar:
To clarify (just in case): The control of whether or not we should open new windows is
unrelated to this issue in which the startup function accidentally
opened 2 windows.

PR #607, opened 2020-03-16
query-buffer: avoid overwritting a global variable, fixes #567

### LINKED ISSUE #567: search-over-buffers corrupt matched elements
**Describe the bug**
When I use **(C-s s)** to do incremental search, already matched elements end getting their contents set to undefined unless I quit the search and then manually clear the hints with **(C-s k)**

I experimented a bit, and so far I know that:

- Already matched elements gets altered as soon as I modify the query
- Exiting the search, then making another without clearing the hints first still modify the matched elements
- Does not seems to affect the marked elements left with hints from bug #566 in e621
- On [e621](https://e621.net) next makes a complain not unlike the one from using follow-hint* there, but unlike that function, search-buffer(s) doesn't seem to be affected

Some pictures:

search-error
search-error-2
search-error-3
search-error-4
search-error-5
search-error-6
search-error-7

1. Do **(C-s s)** 
2. Match some elements first
3. Modify the query with previous matches around

**Information**
- OS name+version: Debian 10 Buster (Stable)
- Kernel name+version: Linux 4.9.0-9 amd64
- Graphics card and driver: Intel Q 33 (Integrated)
- Desktop environment / Window manager name+version: Stumpwm 19.11
- Next version (from =M-x next-version= or =next --version=): 1.5.0
- How you installed Next (Guix pack, package manager, build from source): Guix package
- SBCL version (if built from source):
- WebKitGTK+ / PyQt-Webengine version:

gravitational-dark-light:
This is the cause of #567, and it turned out to be a simple thing, the global variable NODES has global state that must be preserved, and query-buffer was overwriting it regardless

### DIFF
--- a/source/search-buffer.lisp
+++ b/source/search-buffer.lisp
@@ -5,7 +5,8 @@
 (define-parenscript query-buffer (query)
   (defvar *identifier* 0)
   (defvar *matches* (array))
-  (defvar *nodes* (ps:new (-Object)))
+  (when (= (typeof (ps:chain self *nodes*)) "undefined")
+    (defvar *nodes* (ps:new (-Object))))
 
   (defun qs (context selector)
     "Alias of document.querySelector"


jmercouris:
I am closing this since search-buffer.lisp has changed substantially. If you still have this issue, please let us know.

PR #612, opened 2020-03-19
Noscript-mode only blocks javascript markup

### LINKED ISSUE #595: no-script mode breaks Next
It was reported to me that no-script does not work on master.
Needs investigation.

jellelicht:
@Ambrevar I figured out this issue and fixed it, only to then find out at commit-time that you already did so in a revision of next from 9 months ago. I shamelessly stole your commit message. 

*NOTE*: This should only be merged after joachifm/cl-webkit#38 is upstreamed (and subsequently `build-scripts/guix.scm` should be updated to refer to that version of cl-webkit)

Also fixes #595

### DIFF
--- a/source/noscript-mode.lisp
+++ b/source/noscript-mode.lisp
@@ -8,8 +8,8 @@
     ((destructor
       :initform
       (lambda (mode)
-        (ipc-buffer-enable-javascript (buffer mode) t)))
+        (ipc-buffer-enable-javascript-markup (buffer mode) t)))
      (constructor
       :initform
       (lambda (mode)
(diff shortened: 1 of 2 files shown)


Ambrevar:
Excellent, thanks!

Ambrevar:
@jmercouris: Notice the subtle difference in the WebKitGTK API between `enable-javascript` and `enable-javascript-markup` ;)

jmercouris:
I was looking at the two one was for external scripts and one was for scripts loaded through the API, I thought I had used the right one :-D

PR #665, opened 2020-04-17
Echo status

Ambrevar:
This adds 2 components to the UI:
- A status view.
- A message view.

It's a fantastic improvement in my opinion.  No more screenspace contentions, no more race conditions!
`M-x next-version` and `M-x next-init-time` work again!

Aestetically there is more work to be done.  The dimensions are not ideal, the borders are not well set, etc.

We could add a command to toggle the status view and message view.
Shall we make it a mode?

I figured that "message" was more familiar than "echo" (which is more of an Emacs term).
Shall we rename all `echo*` commands to `message*`?

In my opinion this is ready to be merged.  Next I'll clean up minibuffer.lisp a little.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -20,10 +20,10 @@
    (active-buffer :accessor active-buffer :initform nil)
    (active-minibuffers :accessor active-minibuffers :initform nil
                        :documentation "The stack of currently active minibuffers.")
-   (status-buffer-height :accessor status-buffer-height :initform 26
+   (status-buffer-height :accessor status-buffer-height :initform 16
                          :type integer
                          :documentation "The height of the status buffer in pixels.")
-   (message-buffer-height :accessor message-buffer-height :initform 26
+   (message-buffer-height :accessor message-buffer-height :initform 16
                          :type integer
                          :documentation "The height of the message buffer in pixels.")
    (minibuffer-open-height :accessor minibuffer-open-height :initform 256
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -661,7 +661,7 @@ proceeding."
     (setf (last-access buffer) (local-time:now))
     (setf (last-active-buffer *browser*) buffer)
     (set-window-title window buffer)
-    (echo-dismiss)
+    (print-status)
     (setf (active-buffer window) buffer)))
 
 (defun %get-inactive-buffer ()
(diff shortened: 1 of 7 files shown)


jmercouris:
May I commit changes on this branch?

jmercouris:
I would suggest that the status view and message view get flipped in terms of position

Ambrevar:
No strong opinion, go ahead :)

Ambrevar:
Note about the position of the message view and the status view: the status view would probably never be more than 1 line high by default.  The message view is prone to be resized, e.g. when we get a message on multiple lines, a message with a list, etc.

Maybe it would be more natural that the message view does not push the status view up when it gets resized.  I know Emacs does this, but I'm not sure it's a good idea.

We could also set the idle height of the message view to 0 to save screen space.
If we do so (or when the users do so), the status view would move up and down every time we hover links for instance.

Ambrevar:
I think this is ready to be merged.  @jmercouris what do you think?

jmercouris:
I would like to have a cursory glance a priori first, I'll get to this as soon as possible!

jmercouris:
I've set the message and status height to smaller sizes. I do not suggest changing the idle height. It'll be jarring to constantly have things being occluded/unoccluded as the user moves the mouse and messages are posted. It looks really good to me! Well done!

jmercouris:
As per really long lines on the message view, there are two possibilities:

1. it is a long message, scroll it sideways over time
2. it is several messages at once, show a [1/3] or something and then with a timer show the second, and third message

jmercouris:
Once again, thank you for the excellent work!

Ambrevar:
Well done, thanks!

Ambrevar:
And thank you for the dramatically improved styling: I was quite
frustrated with my own result just because of the style :)

jmercouris:
No problem!

impaktor:
So how would one go about showing how many buffers one has in the status line?

jmercouris:
From the docstring :-)

```
Function of a window argument that returns
a string to be printed in the status view.

Example formatter that prints the buffer indices over the total number of buffers:

\(defun my-format-status (window)
  (declare (ignore window))
  (let* ((buffer (current-buffer))
         (buffer-count (1+ (or (position buffer
                                         (sort (alexandria:hash-table-values (buffers *browser*))
                                               #'<
                                               :key #'id))
                               0)))
    (format nil \"[~{~a~^ ~}] (~a/~a) ~a — ~a\"
            (mapcar (lambda (m) (str:replace-all \"-mode\" \"\"
                                                 (str:downcase
                                                  (class-name (class-of m)))))
                    (modes buffer))
            buffer-count
            (hash-table-count (buffers *browser*))
            (url buffer)
            (title buffer))))
```

then you wold need to do something like:

```
(define-configuration browser
  ((status-formatter #'my-format-status)))

```

PR #685, opened 2020-04-22
Contextual keys

Ambrevar:
@hsanjuan Can you confirm this works for you?

I haven't fixed SPACE for now.
There is also an issue with button2 that this pull request fails to fix properly.
button2 (the mouse wheel button) is supposed to open a link in a new buffer.
But if an input element is active, it will paste instead.

Is there a way to detect the HTML element under the mouse cursor?

### DIFF
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -3,8 +3,9 @@
   (:import-from #:keymap #:define-key #:define-scheme)
   (:documentation "Mode for web pages"))
 (in-package :next/web-mode)
-
-(trivial-package-local-nicknames:add-package-local-nickname :sera :serapeum)
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (trivial-package-local-nicknames:add-package-local-nickname :alex :alexandria)
+  (trivial-package-local-nicknames:add-package-local-nickname :sera :serapeum))
 
 ;; TODO: Remove web-mode from special buffers (e.g. help).
 ;; This is required because special buffers cannot be part of a history (and it breaks it).
@@ -34,7 +35,7 @@
        "C-u M-g" #'follow-hint-new-buffer
        "C-x C-w" #'copy-hint-url
        "C-v" #'paste
-       "button2" #'paste
+       "button2" #'maybe-paste
        "C-c" #'copy
        "button9" #'history-forwards
        "button8" #'history-backwards
@@ -53,8 +54,8 @@
        "C-f" #'search-buffer
        "M-f" #'remove-search-hints
        "C-." #'jump-to-heading
-       "end" #'scroll-to-bottom
-       "home" #'scroll-to-top
+       "end" #'maybe-scroll-to-bottom
+       "home" #'maybe-scroll-to-top
        "C-down" #'scroll-to-bottom
        "C-up" #'scroll-to-top
        "M-v" #'scroll-page-up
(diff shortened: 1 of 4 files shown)


source/web-mode.lisp
```diff
@@ -207,6 +208,52 @@
...
+                    (last-event buffer))
+                   (funcall-safely command)))))
+
+(define-command maybe-paste (&optional (buffer (current-buffer)))
+  "Paste text if active element is an input tag, forward event otherwise."
+  ;; TODO: This does not work properly.
+  ;; If an input element is active and we press button2, it pastes instead of
+  ;; opening the link in a new tab.
+  ;; We should check if the mouse cursor is over a link.
+  (call-input-command-or-forward #'paste :buffer buffer))
```
jmercouris, line 247: You can see what the mouse is hovering over via javascript

jmercouris:
I just want to confirm: the only time we'll have to call out to parenscript before execution of a command would be for things that are contextually sensitive?

Ambrevar:
> I just want to confirm: the only time we'll have to call out to parenscript before execution of a command would be for things that are contextually sensitive?

I don't know if it's going to be the only time ever, but for now, yes.

source/web-mode.lisp
Ambrevar: How?

jmercouris:
```
window.onmouseover=function(e) {
        console.log(e.target);
};
```

jmercouris:
I think for now it is OK, but I would like to think through this a bit, because I think in the end there will probably be many context sensitive commands. C-a, C-e... or when a user binds something themselves, we need to know if it is context aware, if it conflicts with a special set of bindings that are meant to interact with the web page.

Ambrevar:
If the Javascript does not cut it or if we need another way to be more
general, maybe

https://webkitgtk.org/reference/webkit2gtk/stable/WebKitHitTestResult.html

would do.

Ambrevar:
I couldn't find how to configure the default WebKitGTK bindings.

source/web-mode.lisp
Ambrevar: This won't work because the onmouseover event will be sent _after_ the mouse has moved to a _new_ element.  I'll see what I can do with WebKit.

jmercouris: In that case, you could iterate through all of the children in the dom, in this case, all of the text areas and inputs- and see if any of them are hovered

Ambrevar:
Merged with 553c6602df4474cf23f9f3a8fde49a79cfb6c714.

@jmercouris: I found a simple fix: we already know the hovered URL because we print it in the echo area.  So I decided to just store it in a buffer slot.  If this is set, it means we can open the link with button2.  If not, then we try pasting.

@hsanjuan Don't hesitate to report if this does not work for you!

jmercouris:
that is a very simple fix indeed! nice work!

Ambrevar:
> In that case, you could iterate through all of the children in the dom, in this case, all of the text areas and inputs- and see if any of them are hovered

Is there a hovered property?

jmercouris:
Unfortunately there isn't, in my opinion this is the best one can do https://stackoverflow.com/a/14798046

hsanjuan:
>  iterate through all of the children in the dom

This sounds like a candidate to cause lag.

@Ambrevar might be related, but I cannot open links in Prontonmail emails: clicking does not work, middle click says:

```
<WARN> [11:24:05] JavaScript error: GError: Domain: "WebKitJavascriptError", Code: 699, Message: https://beta.protonmail.com/inbox/<deleted>:5:55: TypeError: undefined is not an object (evaluating 'activeElement1.value.substring')
```

Right click plus choosing from menu does not work.

C-g plus choosing link works (I guess I should be using that anyways!).

Ambrevar:
>>  iterate through all of the children in the dom
>
> This sounds like a candidate to cause lag.

Agreed :p

> @Ambrevar might be related, but I cannot open links in Prontonmail
> emails: clicking does not work,

The "left click" does not work?  Without modifiers?  That's odd.
"button1" is not bound to anything by default, should work out of the box.

Are you using vi-mode?

Does Next say anything?  If you inspect the element, is it different
from other links?  Did you enable blocker-mode?

> middle click says:
>
> ```
> <WARN> [11:24:05] JavaScript error: GError: Domain: "WebKitJavascriptError", Code: 699, Message: https://beta.protonmail.com/inbox/<deleted>:5:55: TypeError: undefined is not an object (evaluating 'activeElement1.value.substring')
> ```

I guess we miss a check.

> Right click plus choosing from menu does not work.

That's a long standing issue! :p

hsanjuan:
> > iterate through all of the children in the dom This sounds like a candidate to cause lag.
> Agreed :p
> @Ambrevar might be related, but I cannot open links in Prontonmail emails: clicking does not work,
> The "left click" does not work? Without modifiers? That's odd. "button1" is not bound to anything by default, should work out of the box. Are you using vi-mode? Does Next say anything? If you inspect the element, is it different from other links? Did you enable blocker-mode?

I will never ever use vi-mode.

I did nothing special, but if I remember well Protonmail has some JS pop up that should ask me if I really want to open that link, so javascript is probably intercepting the click but I don't see the pop up (in Epiphany I do).

Ambrevar:
Very strange.  Can you open an issue about this and report the exact recipe?

PR #690, opened 2020-04-23
Makefile: rebuild when something changes in /libraries

hsanjuan:


Ambrevar:
Good catch!
What separating the .asd though?  Why not keeping it on the lisp_files line?

hsanjuan:
Hows this now?

Ambrevar:
Perfect, thanks!

Issue #668, opened 2020-04-17
Update buffer URL when clicking on anchors


Ambrevar:
Example:
https://wiki.archlinux.org/index.php/GTK#Configuration

Ambrevar:
See #613, maybe the same fix.

Ambrevar:
Fixed with origin/master 86b24e375bc292eb310db763aa769019e1af8c4b.

Issue #211, opened 2019-05-16
DuckDuckGo: Going backward on history returns a weird char "l/"


lxbarbosa:
Hi,

Searching any term on ddg in a new tab (m-l), visiting any site and going back to that ddg search, now it returns a weird char "l/" instead of "asd"

https://i.ibb.co/jkj9TLq/2019-05-17-010242-1366x768-scrot.png

**Reproduce:**
- m-l (new tab)
- searching for "asd"
- enter any site of that search
- backward on history (c-b)
- returns a ddg page with "l/" 

--------------------------
Next: sbcl-next (guix)
OS: Debian SID 10

Ambrevar:
I've experienced this myself.
I'm not sure where that could come from.

@jmercouris any idea?

alexanderbarbosa:
@jmercouris @Ambrevar 

I just noticed that the URL as user go backward on history do change, it is not the early DDG search URL :

**ddg search:**  https://duckduckgo.com/?q=meugas+asda&ia=web
**backward history:**  https://duckduckgo.com/l/?kh=-1&uddg=https%3A%2F%2Fgroceries.asda.com%2Fshelf%2Fdining-glassware%2Fsingle-mugs%2F_%2F111421

Maybe DDG do something just before we access a site, like providing an analytic data for websites know that we accessed  their site through DDG.

As Next look for the its parent url, maybe we just need to look for the grand parent url, that I suppose the correct DDG early search url.

Ambrevar:
This is what I found as well.
I suspect that we need to check the HTTP return code (probably 200 here)
and decide whether or not to add the URL to the history.

Ambrevar:
Could be fixed by #613.

jmercouris:
This is fixed! I knew a recipe to reproduce it and it no longer happens :-)

Ambrevar:
I can reproduce.  Which recipe did you use?
Mine:

- C-l foo
- Click on the first link.
- M-x history-backwards

Result: it reloads the same page instead of going back to the duckduckgo results.

Ambrevar:
I believe this could be fixed if we used WebKitGTK history, which we should for performance reasons.  See #613.

jmercouris:
Works for me...

jmercouris:
no idea :-D

Ambrevar:
Implemented a workaround in dc238e253c39383850573f90f94801484dd16857.

[...]

PR #694, opened 2020-04-24
History uri fixes

### LINKED ISSUE #613: Use renderer's history cache
When going backward or forward in the buffer's history, we can leverage the renderer's cache to speed up the loading.

This would make Next browser experience much faster overall.

Ambrevar:
I wanted to fix all history-related issues together because I wanted to make sure that the variouas fixes would be consistent.

1. DuckDuckGo /l spurious history entry.  See #211.  I've inspected WebKitGTK's history and noticed that the spurious entry gets added there too, but as soon as the page is done loading it gets replace by the final URL.  I don't know how they do that.  In the mean time, I've implemented a stupid URL-blacklist workaround.  Should be good enough for 2.0 at least.

2. WebKitGTK history cache is now used!  Try going back and forth, it should now be lightning fast!  (If you have a fast connection, enable Tor to notice the difference.)  This is one of my favourite updates in weeks (with the status bar)!  Fixes #613 .

3. Update URL when browsing GitHub files and clicking on anchors.  See #668 .
This problem was re-introduced with the FFI bindings.  @jmercouris: To set the URL in WebKitGTK we can't rely on the URL passed during the :load-changed signals because Javascript URL changes or TOC clicks won't be registered.  Instead we must use the notify:uri signal.  We had the same problem some versions ago with the platform ports ;)

### DIFF
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -19,6 +19,17 @@
             :type htree:history-tree
             :initarg :history
             :initform (htree:make))
+   (history-blacklist :accessor history-blacklist
+                      :type list-of-strings
+                      :initarg :history-blacklist
+                      ;; TODO: Find a more automated way to do it.  WebKitGTK
+                      ;; automatically removes such redirections from its
+                      ;; history.  How?
+                      :initform '("https://duckduckgo.com/l/")
+                      :documentation "URI prefixes to not save in history.
+Example: DuckDuckGo redirections should be ignored or else going backward in
+history after consulting a result reloads the result, not the duckduckgo
+search.")
    (keymap-scheme
     :initform
     (define-scheme "web"
@@ -427,7 +438,7 @@ Otherwise go forward to the only child."
     (copy-to-clipboard input)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;; Warning: To specialize `on-signal-load-committed' we must be in the right package.
+;; Warning: To specialize `on-signal-load-finished' we must be in the right package.
 (in-package :next)
 (defmethod on-signal-load-committed ((mode next/web-mode::web-mode) url)
   nil)
@@ -441,13 +452,17 @@ Otherwise go forward to the only child."
     (unzoom-page :buffer buffer)
     (set-window-title (current-window) buffer)
     (print-status)
-    (htree:add-child (make-instance 'buffer-description
-                                    :url url
-                                    :title (title buffer))
-                     (next/web-mode::history mode)
-                     :test #'equals)
-    (when url
-      (history-add url :title (title buffer)))
+
+    (unless (find-if (alex:rcurry #'str:starts-with? url)
+                     (next/web-mode:history-blacklist mode))
+      (htree:add-child (make-instance 'buffer-description
+                                      :url url
+                                      :title (title buffer))
+                       (next/web-mode::history mode)
+                       :test #'equals)
+      (when url
+        (history-add url :title (title buffer))))
+
     (match (session-store-function *browser*)
       ((guard f f) (funcall-safely f))))
   (echo "Finished loading: ~a." url)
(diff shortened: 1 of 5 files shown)


source/web-mode.lisp
```diff
@@ -19,6 +19,17 @@
             :type htree:history-tree
             :initarg :history
             :initform (htree:make))
+   (history-blacklist :accessor history-blacklist
+                      :type list-of-strings
+                      :initarg :history-blacklist
+                      ;; TODO: Find a more automated way to do it.  WebKitGTK
```
jmercouris, line 25: Is there anything in the HTTP status codes or some header that contains this information?

source/web-mode.lisp
```diff
@@ -19,6 +19,17 @@
             :type htree:history-tree
             :initarg :history
             :initform (htree:make))
+   (history-blacklist :accessor history-blacklist
```
jmercouris, line 22: I think it should be history-ignorelist because it is history elements that we ignore

jmercouris:
Great fixes! Other than my single comment about blacklist-> ignore list

blacklist makes sense, but ignore list I think would be more intuitive

jmercouris:
whilst we investigate the strange URL issue, the solution is acceptable. It could also be useful for users, maybe they don't want all of their search engine searches to appear in their history (of any kind) and this could be used to automatically do that, maybe it could be a regex instead of a string.

Anyways, things for later. Once again, great work! Feel free to merge it whenever you are ready :-)

P.S. this means Next won't show "Loading"in the status bar as I've implemented it while loading, but I think that is acceptable since JS is doing its own loading and every site implements their own "loading" bar (think youtube, github)

Ambrevar:
> Great fixes! Other than my single comment about blacklist-> ignore list
>
> blacklist makes sense, but ignore list I think would be more intuitive 

We use "whitelist" for the certificate-whitelits-mode, and we use the
"blacklist" term for blocker-mode.  Out of consistency, if would make
sense to use blacklist here too I think.

The benefit of "blacklist" is that it has an obvious opposite,
"whitelist".  I can't think of the opposite of "ignore list".

Ambrevar:
> Is there anything in the HTTP status codes or some header that contains this information?

Sadly I don't know how to get the status code with WebKit.

Ambrevar:
> P.S. this means Next won't show "Loading"in the status bar as I've
> implemented it while loading, but I think that is acceptable since JS
> is doing its own loading and every site implements their own "loading"
> bar (think youtube, github)

Hmm, unless I'm missing something, I don't think my patch set changes
anything with regard to Javascript loading: it does not trigger the
:load-changed signal so we didn't display anything in the echo view /
status view before.

jmercouris:
> Great fixes! Other than my single comment about blacklist-> ignore list blacklist makes sense, but ignore list I think would be more intuitive
> We use "whitelist" for the certificate-whitelits-mode, and we use the "blacklist" term for blocker-mode. Out of consistency, if would make sense to use blacklist here too I think. The benefit of "blacklist" is that it has an obvious opposite, "whitelist". I can't think of the opposite of "ignore list".

Ah! I didn't think about it in the greater context of the codebase, that makes sense then!

jmercouris:
> P.S. this means Next won't show "Loading"in the status bar as I've implemented it while loading, but I think that is acceptable since JS is doing its own loading and every site implements their own "loading" bar (think youtube, github)
> Hmm, unless I'm missing something, I don't think my patch set changes anything with regard to Javascript loading: it does not trigger the :load-changed signal so we didn't display anything in the echo view / status view before.

We did not, and we still won't, is basically what I was saying :-D

jmercouris:
WebKit URI response object has a status code, headers, etc

jmercouris:
The puzzle may go far deeper than we could anticipate, so I would suggest, let's leave it alone for now. It is a cool feature to have in any case (even if it is a hack :-P). And we can ultimately figure out at some point in the future what is going on with that strange "L" character

Ambrevar:
To clarify, the explanation for the "L" character is clear: when
clicking on a result, DuckDuckGo actually loads

  https://duckduckgo.com/l/...<the actual url>

which in turn loads the final URL.

Most browsers don't store the intermediate "/l/" URL and we shouldn't either.

jmercouris:
I don't think it is so clear, I have a lot of questions

why this intermediate step before loading the result?
why this character?
why do other browsers not store the imtermediate URL? is there some information in the response that tells us not to store that (which we are not reading?)
...
etc :-D

Ambrevar:
> why this intermediate step before loading the result?

It's a redirect (not an HTTP one though), see https://webapps.stackexchange.com/questions/73312/are-duckduckgo-redirects-a-privacy-issue?noredirect=1.

> why this character?

It's just part of the URL, could be any string.

> why do other browsers not store the imtermediate URL? is there some information in the response that tells us not to store that (which we are not reading?)

It's done at the engine level in WebKitGTK and all other browsers that I
know of use the engine's history functions, so they don't have to deal
with this kind of subtleties.

Ambrevar:
Merging, I'll work on the URI response code later.

Issue #517, opened 2020-01-14
load-hook on buffer is only called when loading url via minibuffer


chrisboeg:
**Describe the bug**

As far as I can see, the `load-hook`s on a buffer is only called when you try to load an URL via minibuffer. If you click on a link on a website, the `load-hook`s doesn't get called.

**Precise recipe to reproduce the issue**

Add this code to your init.lisp file:

```
(defun my-load-hook (url)
  (let* ((uri (quri:uri url)))
    (log:info "want to load url: ~a" url)
    url))

(defclass my-buffer (buffer)
  ((load-hook :initform (next-hooks:make-hook-string->string
                          :handlers (list #'my-load-hook)
                          :combination #'next-hooks:combine-composed-hook))))

(setf *buffer-class* 'my-buffer)
```

Now open next via a shell so you can monitor log output via stdout and try to load urls via e.g. C-l shortcut as well as by clicking links on websites. By monitoring the stdout from next, you should be able to see that the `(log:info ...)` only writes to stdout when visiting urls via the C-l binding, not by clicking on links on webpages.

**Information**
- Next version (from =M-x next-version= or =next --version=): 1.4.0
- WebKitGTK+ / PyQt-Webengine version: WebKitGTK+

jmercouris:
Thank you for the report! We will look into it!

chrisboeg:
@jmercouris I can see if I can figure it out and send a PR. I just want to have it confirmed that the load-hook is indeed supposed to be called in all cases where an url is called.

jmercouris:
It is not supposed to be called, looking at the documentation string:

```
Hook run in `set-url' after `parse-url' was
processed.  The handlers take the URL going to be loaded as argument and must
return a (possibly new) URL.
```

I can't remember what all the hooks are, there may already be a hook for your use case. If not, there should be one.

jmercouris:
I should note that did-commit-navigation and did-finish-navigation are invoked upon the loading of URLs, however they are not designed to be friendly to users. @Ambrevar may know more about a more appropriate hook.

chrisboeg:
Ah, okay. I must be confused. I saw the hook in the "old-reddit-hook" example from the manual where "load-hook" is used to redirect reddit.com to old.reddit.com, and where old-reddit is actually only redirected to in the case of navigating to reddit via for example the minibuffer, not when clicking a reddit-link on a website. So if `load-hook` is actually not supposed to be used in such a case, then the manual should probably also not use it for such case.

[...]

PR #698, opened 2020-04-26
Composed request resource

Ambrevar:
See the commit messages.

After this gets merged:

- Shall we rename the lengthy `request-resource-hook` to `request-hook`?
- `set-url-hook` seems to be completely superseded by `request-hook`.  I suggest removing it for now because it's not very useful and might confuse users as in #517.  Or at least unexport it for now and if it ever gets requested by some users we can reenable it later.  It's easier in this direction because we can't remove / unexport after 2.0 (it would be API-breaking).

### DIFF
--- a/libraries/hooks/hooks.lisp
+++ b/libraries/hooks/hooks.lisp
@@ -228,6 +228,7 @@ This is an acceptable `combination' for `hook'."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 (defun add-hook-internal (hook handler &key append)
   "Add HANDLER to HOOK if not already in it.
+Return HOOK.
 HANDLER is also not added if in the `disabled-handlers'.
 If APPEND is non-nil, HANDLER is added at the end."
   (serapeum:synchronized (hook)
@@ -235,7 +236,8 @@ If APPEND is non-nil, HANDLER is added at the end."
                 (member handler (disabled-handlers hook) :test #'equals))
       (if append
           (alexandria:appendf (symbol-value hook) (list handler))
-          (pushnew handler (handlers hook) :test #'equals)))))
+          (pushnew handler (handlers hook) :test #'equals)))
+    hook))
 
 (declaim (ftype (function ((or handler symbol) list) (or handler boolean)) find-handler))
 (defun find-handler (handler-or-name handlers)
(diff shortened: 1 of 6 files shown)


jmercouris:
Unexport is sufficient I believe. I still think it has a use case :-)
I would not rename request-resource-hook because the name is perfectly descriptive, request-hook is ambiguous

jmercouris:
I'll take a look at this on Monday

Ambrevar:
A "request" is descriptive enough I think.
A user might wonder "what is a resource?"

Ambrevar:
In case the consequences of this change were not clear: now we can hi-jack the URLs of any link that is clicked, downloaded, requested otherwise.  And we can _compose_ those hi-jacks.

For instance, an "old-reddit" handler may change "www.reddit.com" to "old.reddit.com".
And another handler may decide to switch to the mobile version.  Both handlers can work together (and be toggled at will) without having to be _aware_ of each other.

jmercouris:
How is the order of the execution decided?

jmercouris:
In other words, how can hooks that compose chain together in such a way that they all process the data as expected?

Would some orders not be invalid? Could we leverage an expert system to automatically deduce the order in which the hooks should execute?

jmercouris:
OR is it that every hook can work with ANY type of data in ANY order?

Ambrevar:
The hooks library offers a very nice features: hooks have a `:combination` slot that specifies how the handlers are executed.
`hooks` offers these by default:
- default-combine-hook (emacs style)
- combine-hook-until-failure
- combine-hook-until-success
- combine-composed-hook

For our needs in resource-request-hook, I've added `combine-composed-hook-until-non-nil-second-value` which composes all the handlers from head to tail until one returns a non-nil second value.

Does that make sense?

Ambrevar:
Practically, this means that if we enable a mode that adds a handler to the hook, it gets executed before the existing handlers.  In practice, I think this is what we want.

Note that a mode is free to place the the handler wherever they want in the hook.  For instance, if A needs to be after B if present, we can always search for B and add A there; or, if B is missing, add A at the beginning.

As you said, we can derive a system for sorting the handlers if the need ever arises.  I think it's an orthogonal issue that we can work on later.  Thoughts?

jmercouris:
That sounds just fine to me, I didn't realize that logic was already in the hooks library, I guess I did not study it well enough! As per the expert system, you are absolutely correct, that is another issue for another time :-)

Anyways, looks ready to me! Nice work!

Ambrevar:
OK, I'll merge soon.
What about renaming the hook, in the end?

jmercouris:
I don't feel too strongly about it, I am OK with it in any case. My opinion is still that we are requesting a resource, and not for example a remote execution call...I also like that the nomenclature is consistent with the web renderer when using "request resource"

Ambrevar:
Merged in 4a8b0e890a38fdebe1cd5a6f7a6f3032955e6487.

Issue #672, opened 2020-04-18
Echo / message error on load failures


Ambrevar:
Recipe:
- disconnect from the Internet.
- load a page.

We need to tell the user that the page could not be loaded instead of doing nothing, since it's very confusing.

Ideally, the status view should keep the load status displayed since the echoed message is ephemeral.

jmercouris:
Interesting, mine shows an error, says page could not be loaded after some time

jmercouris:
Status view is now updated to show when a page is loading. The renderer will try as long as possible, and eventually report an error message. There are no events that should overwrite the status until it is done.

jmercouris:
Works as expected:
![Screen Shot 2020-04-24 at 13 16 35](https://user-images.githubusercontent.com/1691662/80207131-20c24e80-861d-11ea-9e0d-ecd3e1308a98.png)

jmercouris:
I find it hilarious that my interface is half in English and half in Greek :-D

Ambrevar:
Thanks, that's already an improvement!

We can do a little bit better I think, consider this:
Run `follow-hint-new-buffer` and follow a hint in the background.
Notice that both the status view and the echo area report that they are
loading the buffer.  This is redundant, plus the current buffer does not
display any meaningful status anymore.

I suggest the following approach instead: when the buffer loads, have
print-status display the regular status bar with (loading...) appended
before or after the URL.  This way it's contextual to the buffer that's
actually loading.

Thoughts?

jmercouris:
Ah, I see what is happening, that behavior with the background was not intended... :-D

jmercouris:
I don't understand your proposed solution, can you explain how that would fix the problem? The real issue is that the status message is currently tied to a window, and not a buffer

Ambrevar:
My suggestion:

- Don't use print-status like you did on a buffer load.
- Instead, in the print-status function, get the load state of the
  buffer and print it next to the URL
  (we might need a new buffer slot for that).

By the way, I'm realizing that the current fix is going off topic: the issue is about
load _failures_, e.g. when there is not internet connection.  Right now
we get buffers like "Internel WebKit error" which means little to the
user.

What we must do here is handle the load-failed signal.

jmercouris:
print-status would still need to be informed by someone, that would have to be on buffer load. are you suggesting the following:

print-status invoked in buffer load with no arguments
print status default status formatter adds loading if a buffer is currently loading?

[...]

PR #699, opened 2020-04-26
Loading status

Ambrevar:


Ambrevar:
This expands on the fix for #672.

Ambrevar:
We still have to report network errors properly (i.e. with the :load-failed signal) before closing #672.

jmercouris:
I think (Loading) is too subtle, but that's OK for now

I was goign to add a little CSS loading circle, but decided against it, we can think about it later. Thanks for the fixes!

Ambrevar:
> I was goign to add a little CSS loading circle, but decided against it, we can think about it later. Thanks for the fixes!

Agreed, the loading circle would be nicer, but hey, we worry about
cosmetic after usability :)

jmercouris:
I also decided against the loading circle because I thought it would be distracting, maybe it would be enough to just write "Loading" in bold text or something, anyways we'll see!

PR #702, opened 2020-04-27
fix declaration for set-window-title

gravitational-dark-light:
This issue did break the build for for me at commit b56d0d2d568e8af3f8dea7bf5d289f792461d475


Ambrevar:
Good catch, thanks!

Issue #646, opened 2020-04-11
Start options to set the location from individual files xdg-data-home files + profile support


Ambrevar:
Currently we have only two knobs to move the xdg-data-home files (sessions, cookies, hostlist, history, bookmarks, socket).  XDG_DATA_HOME and --session.

The downside it that it's tricky to use a different file for only one component, say, use a different cookie file but the same history.

We already have configurable knobs for sessions and the init file (which is not in XDG_DATA_HOME, but the mechanism is the same).

1. Let's implement a general way to configure the location of all files individually.

2. It'd be nice if the `start` function would allow setting the same knobs as the command line, so that we have the same flexibility from the REPL.  In particular, it's useful to set the socket file from the REPL to avoid calling the binary single instance that could be running.

jmercouris:
1. Is too abstract for me to understand, can you please elaborate on how you imagine that may look?

2. OK!

Ambrevar:
Actually I've fixed 2. yesterday! :)

About 1, the goals are:

- Allow precise and configuration of where the individual persisted data
  is stored.

- Make it controllable from the command line.  The point is that the
  user should not have to write Lisp code just to change the path of the
  persisted data files.

Suggested strategy to avoid a profusion of command line options:

- Allow configuring the "next" part of "$XDG_DATA_HOME/next".

- Use the same session path expansion function for bookmarks, cookies, etc.

- Remove the --session option.

- Add a --profile command line option.

- Add a PROFILE slot for the BROWSER class and move bookmarks-path,
  cookies-path, etc. in there.
  This slot would be an association between a profile name and paths for
  bookmarks, cookies, etc.

When starting Next, the browser is instantiated with the profile
matching "--profile $PROFILE_NAME", or the "default" profile otherwise.

In their init file, users could define their own profiles, i.e.

```
'(("default"  ...)
  ("private" nil) ; no paths, no data is persisted.
  ("work" (:bookmarks-path "work-bookmarks" :cookie-path "..." ...)))
```

It's important to get this right before 2.0 cause it would be
API-breaking afterwards.

Another cool feature would be to add a command to start Next with a
given profile.  Many browsers can "open a private window", this would
allow us to do this (and more).

Thoughts?

Ambrevar:
Note: we also need to add a --socket / --no-socket option, since it's
the only file with the init file that's required before the browser is
instantiated.

--no-socket would allow to start Next in multi-instance.

Ambrevar:
[...]

PR #704, opened 2020-04-29
Replace manual file persistence with the data-storage library.

Ambrevar:
This is a work in progress: GPG support needs testing, download-mode, file-manager-mode and video-mode need integration.  We need command line customization as well (easy).

What this does:

- Streamline file persistence with two functions: `ensure-path` and `with-data-file`.
  These functions make sure to create the parent folder, to resolve environment variables, to crypt with GPG if the file has the .gpg extension, etc.

- Individual data-path as well as data-profile are both configurable from the init.lisp and soon from command line.  The configuration composes: it's possible to configure a given data-path given a profile and vice-versa.

I'll finish it as soon as possible.  In the meantime, you are welcome to test and let me know what you think about the naming.


Ambrevar:
This closes #646.

Ambrevar:
Let's not forget the --socket / --no-socket option! :)

source/start.lisp
```diff
@@ -2,6 +2,18 @@
 
 (in-package :next)
 
+(export-always 'init-file-path)
+(defvar init-file-path (make-instance 'init-file-data-path :basename "init") ; TODO: Rename with ear-mufs to global.
```
jmercouris, line 6: I think these could belong in global.lisp, or shall we move all of the globals out of global.lisp and delete the file?

source/blocker-mode.lisp
```diff
@@ -45,11 +42,11 @@ See the `hostlist' class documentation."
   "Fetch HOSTLIST and return it.
 If HOSTLIST has a `path', persist it locally."
   (unless (uiop:emptyp (url hostlist))
-    (log:info "Updating hostlist ~a from ~a" (path hostlist) (url hostlist))
+    (log:info "Updating hostlist ~a from ~a" (expand-path (path hostlist)) (url hostlist))
```
jmercouris, line 45: I see this `(expand-path ...)` paradigm appearing in many places, perhaps it would be better as a key argument to `(path `?

source/data-storage.lisp
```diff
@@ -0,0 +1,190 @@
+(in-package :next)
+
+;; TODO: Make this a separate library?
```
jmercouris, line 3: difficult to justify as a library, becomes a bit leftpad++ in my opinion :-D

jmercouris: I could see some parts of it though, maybe a trivial-xdg library or something

source/data-storage.lisp
```diff
@@ -0,0 +1,190 @@
...
+function result as a boolean in conditions."
+  (expand-default-path path :root (uiop:xdg-config-home +data-root+)))
+
+(defmethod expand-data-path ((path session-data-path) (profile data-profile))
+  (expand-default-path path :root (uiop:xdg-data-home +data-root+ "sessions")))
+
+(defun xdg-download-dir ()
+  (let ((dir (ignore-errors (uiop:run-program '("xdg-user-dir" "DOWNLOAD")
+                                              :output '(:string :stripped t)))))
+    (when (or (null dir) (string= dir (uiop:getenv "HOME")))
```
jmercouris, line 83: could just check return code of uiop:run-program, I believe one of the values is the return code

source/data-storage.lisp
```diff
@@ -0,0 +1,190 @@
...
+    (unless dir
+      (setf dir (str:concat (uiop:getenv "HOME") "/Downloads/")))))
+
+(defmethod expand-data-path ((path download-data-path) (profile data-profile))
+  ;; TODO: Move default-download-directory to here?
+  (expand-default-path path :root (xdg-download-dir)))
+
+(defmethod expand-data-path ((path data-path) (profile (eql +private-data-profile+)))
+  "Don't persist anything in private mode."
+  ;; (eql (bookmarks-path *browser*)) won't work because *browser* is not instantiated.
```
jmercouris, line 94: could use moptilities?

source/data-storage.lisp
```diff
@@ -0,0 +1,190 @@
...
+
+(defmacro with-data-file ((stream data-path &rest options) &body body)
+  "Evaluate BODY with STREAM bound to DATA-PATH.
+DATA-PATH can be a GPG-encrypted file if it ends with a .gpg extension.
+If DATA-PATH expands to NIL or the empty string, do nothing.
+OPTIONS are as for `open'.
+Parent directories are created if necessary."
+  `(let ((path (expand-path ,data-path)))
+     (when path
+       (let ((gpg? (str:ends-with? ".gpg" path :ignore-case t)))
```
jmercouris, line 178: small stylistic thing: I don't believe the let is really needed here

source/session.lisp
```diff
@@ -28,36 +28,33 @@ Currently we store the list of current URLs of all buffers."
...
   "Restore the current Next session from the browser `session-path'."
   (handler-case
-      (match (with-open-file (file (session-path *browser*)
-                                   :direction :input
-                                   :if-does-not-exist nil)
-               (when file
+      (match (with-data-file (input (session-path *browser*)
+                                    :direction :input
+                                    :if-does-not-exist nil)
+               (when input
```
jmercouris, line 52: the with-data-file macro could be updated to not execute body when all vars are not properly set? thoughts?

source/start.lisp
Ambrevar: Yes, I'll move them to globals.lisp.

source/blocker-mode.lisp
Ambrevar: here `path` is a slot of `hostlist` of type `data-path`.  What do you mean?

source/data-storage.lisp
Ambrevar: The gpg macro (which is not finished) is not trivial.  This macro alone could justify a library! :)

source/data-storage.lisp
Ambrevar: Yes but why?  We don't need the return code.
I'm ignoring errors mostly in case the `xdg-user-dir` executable is not on the system.

source/data-storage.lisp
Ambrevar: I've got a better way, wait for it ;)

source/data-storage.lisp
Ambrevar: Good catch!

source/blocker-mode.lisp
jmercouris: you can make a method called `path` that specializes on hostlist class that accepts an optional keyword to automatically expand the path. in other words, no need to use default slot accessor

source/session.lisp
Ambrevar: You mean if the file does not exist for instance?  Then it's already the case :)

source/data-storage.lisp
jmercouris: return code will tell you whether the dir exists without having to check `(null dir)`

source/session.lisp
jmercouris: so (when input is just checking to see the file has a body?

source/blocker-mode.lisp
Ambrevar: I'll update the code to save `expand-path` in a let-bound variable.

I'm not sure about setting the accessor to call `expand-path`.  I need to think about it.

source/data-storage.lisp
Ambrevar: It does not cover the case when `dir` is nil because `xdg-user-dir` is missing.

source/session.lisp
Ambrevar: Good point.
The HyperSpec of `with-open-file` says

```
It is possible by the use of :if-exists nil or :if-does-not-exist nil for stream to be bound to
nil. Users of :if-does-not-exist nil should check for a valid stream. 
```

Actually a nil value is useful, it allows the body to deal with the non-existing case.
I think it's saner to follow `with-open-file` behaviour.

Ambrevar:
This should be ready to merge.

The main question that remains is the naming.
For now I've used --profile and --list-profiles.  I know you mentioned that you wanted to reserve the "profile" name for other things, but I can't think of a better name.  "--data-profile" and "--list-data-profiles" is lengthy, but it could do.

Thoughts?

jmercouris:
I still think data-profile is the best, profile will be useful for other things. If we never end up using profile, we can always alias it to data-profile, but we cannot as easily mv profile to data-profile, better to keep our options open that way I think.

Nice work! Next get's better with every commit!
Feel ready to merge whenever you like!

Ambrevar:
Merged with e5cedaf9de1181825e9ec1362af1a3fd42586f44.

I've renamed to `data-profile`.

PR #710, opened 2020-04-30
Print TLS error in same buffer.

Ambrevar:
Before merging, we need to fix https://github.com/atlas-engineer/next/issues/672 with the cl-cffi-gtk error we get because of the gerror type.

### DIFF
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -216,10 +216,7 @@ A command is a special kind of function that can be called with
 (defun tls-help (buffer url)
   "This function is invoked upon TLS certificate errors to give users
 help on how to proceed."
-  (let* ((help-buffer (next/help-mode:help-mode
-                       :activate t
-                       :buffer (make-buffer :title "TLS Error")))
-         (help-contents
+  (let* ((help-contents
            (markup:markup
             (:h1 (format nil "TLS Certificate Error: ~a" url))
             (:p "The address you are trying to visit has an invalid
@@ -227,16 +224,20 @@ certificate. By default Next refuses to establish a secure connection
 to a host with an erroneous certificate (e.g. self-signed ones). This
 could mean that the address you are attempting the access is
 compromised.")
-            (:p "If you trust the address
-nonetheless, you can add an exception for the current hostname with
-=add-domain-to-certificate-whitelist=.  The =certificate-whitelist-mode= must be
-active for the current buffer (which is the default).")
+            (:p "If you trust the address nonetheless, you can add an exception
+for the current hostname with the "
+                (:code "add-domain-to-certificate-whitelist")
+                " command.  The "
+                (:code "certificate-whitelist-mode")
+                " must be active for the current buffer (which is the
+default).")
             (:p "You can persist hostname exceptions in your init
-file, see the =add-domain-to-certificate-whitelist= documentation.")))
+file, see the "
+                (:code "add-domain-to-certificate-whitelist")
+                " documentation.")))
          (insert-help (ps:ps (setf (ps:@ document Body |innerHTML|)
                                    (ps:lisp help-contents)))))
-    (ffi-buffer-evaluate-javascript help-buffer insert-help)
-    (set-current-buffer help-buffer)))
+    (ffi-buffer-evaluate-javascript buffer insert-help)))
 
 (defun describe-key-dispatch-input (event buffer window printable-p)
   "Display bound value documentation.
--- a/source/renderer-gtk.lisp
+++ b/source/renderer-gtk.lisp
@@ -352,14 +352,18 @@ Warning: This behaviour may change in the future."
   "Return nil to propagate further (i.e. raise load-failed signal), T otherwise."
   (let* ((context (webkit:webkit-web-view-web-context (gtk-object buffer)))
          (host (quri:uri-host (quri:uri url))))
-    (when (and (certificate-whitelist buffer)
-               (member-string host (certificate-whitelist buffer)))
-      (webkit:webkit-web-context-allow-tls-certificate-for-host
-       context
-       (gobject:pointer certificate)
-       host)
-      (set-url* url :buffer buffer)
-      t)))
+    (if (and (certificate-whitelist buffer)
+             (member-string host (certificate-whitelist buffer)))
+        (progn
+          (webkit:webkit-web-context-allow-tls-certificate-for-host
+           context
+           (gobject:pointer certificate)
+           host)
+          (set-url* url :buffer buffer)
+          t)
+        (progn
+          (tls-help buffer url)
+          t))))
 
 (defmethod on-signal-decide-policy ((buffer gtk-buffer) response-policy-decision policy-decision-type-response)
   (let ((is-new-window nil) (is-known-type t) (event-type nil)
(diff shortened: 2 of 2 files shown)


jmercouris:
I think we'd be better off just removing the lambda and merging :-D

Ambrevar:
I don't think this would work.  If you look at the source:

```lisp
(defun g-signal-connect (instance detailed-signal handler &key after)
 #+cl-cffi-gtk-documentation
 "@version{2013-7-30}
...
  @see-function{g-signal-connect-after}"
  (g-signal-connect-closure (pointer instance)
                            detailed-signal
                            (create-closure instance handler)
                            after))
```

it would fail either way I think.

jmercouris:
Ah, so this does not work as intended?

Ambrevar:
Nope, as I mentioned in the original post we first need to fix the crash you discovered in #672.

jmercouris:
I don't understand the connection with the crash I discovered

Ambrevar:
To print in the TLS error in the current buffer, we _must_ intercept :load-failed or else it will load the failing URL a second time, without TLS error signal this time.

But intercepting :load-failed  crashes Next as you've seen.

jmercouris:
Ah, I understand now!

jmercouris:
Maybe I didn't read the documentation correctly, but I believe we can prevent the propagation to a load-failed form if we handle the signal correctly (t vs nil), in which case the load-failed is not necessary

Ambrevar:
WebKitGTK makes the difference between a signal that's merely registered
from one that is handled.

In this case, it seems that simply registering load-failed makes a
difference, even if we don't reach there in load-failed-on-tls-error.

The problem is that it can reach there on other errors, in which case it
will crash.

jmercouris:
I tested the crategus fork, but I remembered that we are using gtk:within-gtk-thread which does not exist in the @crategus version. I could patch it, but I would rather use the version on Quicklisp.

jmercouris:
just for the record, the bug does NOT exist in the crategus repository

jmercouris:
more investigation reveals that older version of the cl-cffi-gtk crategus fork also suffered from this issue. the fix was committed between 2-3 months ago, will continue to investigate

jmercouris:
the fix occurs between: ef5a8cd and 52aaa70

the obvious candidate is 9cfa3fc

however, not every commit is functional (perhaps I am having cache issues)

jmercouris:
First functional commit is: b658905

will try to figure out why and patch Ferada's fork

Ambrevar:
Indeed, Crategus version was in standby for a while and it got a new
wave of patches this year.  Some of those patches made it to the
Ferada's version but not all of them.

We are getting closer!

jmercouris:
Tried cherry picking some commits to Ferada branch with limited success... I got it to compile and run, but problem was not RM'd. Feel free to look at it too ;-)

Ambrevar:
What is RM?

Open an issue on Ferada's repo, he will probably be able to figure it
out rapidly!

jmercouris:
Done! https://github.com/Ferada/cl-cffi-gtk/issues/40

Ambrevar:
Perfect, thanks!

jmercouris:
RM(1)                     BSD General Commands Manual                    RM(1)

NNAAMMEE
     rrmm, uunnlliinnkk -- remove directory entries

SSYYNNOOPPSSIISS
     rrmm [--ddffiiPPRRrrvvWW] _f_i_l_e _._._.
     uunnlliinnkk _f_i_l_e

DDEESSCCRRIIPPTTIIOONN
     The rrmm utility attempts to remove the non-directory type files specified
     on the command line.  If the permissions of the file do not permit writ-
     ing, and the standard input device is a terminal, the user is prompted
     (on the standard error output) for confirmation.

     The options are as follows:

     --dd          Attempt to remove directories as well as other types of
                 files.

     --ff          Attempt to remove the files without prompting for confirma-
                 tion, regardless of the file's permissions.  If the file does
                 not exist, do not display a diagnostic message or modify the
                 exit status to reflect an error.  The --ff option overrides any
                 previous --ii options.

     --ii          Request confirmation before attempting to remove each file,
                 regardless of the file's permissions, or whether or not the
                 standard input device is a terminal.  The --ii option overrides
                 any previous --ff options.

     --PP          Overwrite regular files before deleting them.  Files are
                 overwritten three times, first with the byte pattern 0xff,
                 then 0x00, and then 0xff again, before they are deleted.

     --RR          Attempt to remove the file hierarchy rooted in each file
                 argument.  The --RR option implies the --dd option.  If the --ii
                 option is specified, the user is prompted for confirmation
                 before each directory's contents are processed (as well as
                 before the attempt is made to remove the directory).  If the
                 user does not respond affirmatively, the file hierarchy
                 rooted in that directory is skipped.

     --rr          Equivalent to --RR.

     --vv          Be verbose when deleting files, showing them as they are
                 removed.

     --WW          Attempt to undelete the named files.  Currently, this option
                 can only be used to recover files covered by whiteouts.

     The rrmm utility removes symbolic links, not the files referenced by the
     links.

     It is an error to attempt to remove the files ``.'' or ``..''.

     When the utility is called as uunnlliinnkk, only one argument, which must not
     be a directory, may be supplied.  No options may be supplied in this sim-
     ple mode of operation, which performs an unlink(2) operation on the
     passed argument.

     The rrmm utility exits 0 if all of the named files or file hierarchies were
     removed, or if the --ff option was specified and all of the existing files
     or file hierarchies were removed.  If an error occurs, rrmm exits with a
     value >0.

NNOOTTEE
     The rrmm command uses getopt(3) to parse its arguments, which allows it to
     accept the `--' option which will cause it to stop processing flag
     options at that point.  This will allow the removal of file names that
     begin with a dash (`-').  For example:
           rm -- -filename
     The same behavior can be obtained by using an absolute or relative path
     reference.  For example:
           rm /home/user/-filename
           rm ./-filename

SSEEEE AALLSSOO
     rmdir(1), undelete(2), unlink(2), fts(3), getopt(3), symlink(7)

BBUUGGSS
     The --PP option assumes that the underlying file system is a fixed-block
     file system.  In addition, only regular files are overwritten, other
     types of files are not.

CCOOMMPPAATTIIBBIILLIITTYY
     The rrmm utility differs from historical implementations in that the --ff
     option only masks attempts to remove non-existent files instead of mask-
     ing a large variety of errors.  The --vv option is non-standard and its use
     in scripts is not recommended.

     Also, historical BSD implementations prompted on the standard output, not
     the standard error output.

SSTTAANNDDAARRDDSS
     The rrmm command is almost IEEE Std 1003.2 (``POSIX.2'') compatible, except
     that POSIX requires rrmm to act like rmdir(1) when the _f_i_l_e specified is a
     directory.  This implementation requires the --dd option if such behavior
     is desired.  This follows the historical behavior of rrmm with respect to
     directories.

     The simplified uunnlliinnkk command conforms to Version 2 of the Single UNIX
     Specification (``SUSv2'').

HHIISSTTOORRYY
     A rrmm command appeared in Version 1 AT&T UNIX.

BSD                            January 28, 1999                            BSD

jmercouris:
RM as in the command, remove

PR #712, opened 2020-05-02
Add instructions for using Sly/Slynk

### LINKED ISSUE #658: Slynk support
It'd be nice to add an build option to replace Swank by slynk.

khinsen:
Fixes #658

### DIFF
--- a/documents/MANUAL.org
+++ b/documents/MANUAL.org
@@ -475,9 +475,9 @@ it to the clipboard.
 You can execute any command by name by typing =M-x=. This will bring up a list
 of candidates that you can fuzzily complete.
 
-** SLIME with a compiled version of Next
+** SLIME (or SLY) with a compiled version of Next
 =SLIME= provides a way of interacting with Next, and with Lisp code in
-general (e.g. in a [[https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop][REPL]]).
+general (e.g. in a [[https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop][REPL]]). =SLY= is a fork of =SLIME= with additional functionality.
 
 From the SLIME manual:
 #+begin_quote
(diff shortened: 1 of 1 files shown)


jmercouris:
Perfect! Thank you so much!

Ambrevar:
Thanks!

Fix: it's a command, not a function.

Instead of quickloading, what about using `asdf:load-system` instead?  It's more universal.
Then you could remove the comment at the end about Quicklisp since it should work with Guix too.

`asdf:load-system` raises an error when the module is not installed.
We could provide our helper that returns T when the system was found and loaded, NIL otherwise.

```lisp
(defun load-system (system)
  (when (asdf:locate-system system)
    (asdf:load-system system)))
```

Then in `init.lisp`

```lisp
(when (load-system :slynk)
  (define-command ...))
```

Ambrevar:
Oops!  John, you might have merged a little fast! :)

jmercouris:
I think quickloading is better because if the system does not exist on their machine it will pull it. That said, I really do see the merits of load system, I would say a -quicklisp declarative would be best.

so something like

```
#-quicklisp
(ql:quickload :slynk)
```

jmercouris:
thoughts?

jmercouris:
Also what is going on with GitHub :-D, I didn't even see your comment, look at this:

Screen Shot 2020-05-02 at 11 53 37

jmercouris:
How is it possible that a comment you made before me, comes up AFTER me on my display?

Ambrevar:
We posted at the same time, can't blame GitHub here :D

Good point with the :quicklisp feature, although I think you meant
#+quicklisp.

What about this then:

```
(defun load-system (system)
  #+quicklisp
  (ql:quickload system)
  #-quicklisp
 	(when (asdf:locate-system system)
 		(asdf:load-system system)))
```

jmercouris:
that looks great!

Ambrevar:
Also we should ignore errors when Quicklisp cannot find the system and
return nil.

Ambrevar:
@khinsen: Why did you define the command in the :next package and not in the :next-user package?

khinsen:
@Ambrevar Without switching to the `:next` package, the command doesn't show up when I do `M-x`.

Ambrevar:
Commands defined in :next-user should be lookup up when calling M-x.
Are you running a recent-enough version?
Can you produce a minimal init.lisp to reproduce the issue?

khinsen:
Here's a minimal `init.lisp` that does nothing but define the new command. With the most current commit (`adf038ee820064d7cf2231123438c270c35df87e`) built from source, `start-slynk` does not show up in the command list.
```
(ql:quickload "slynk")
(define-command start-slynk (&optional (slynk-port *swank-port*))
  "Start a Slynk server that can be connected to, for instance, in
Emacs via SLY.

Warning: This allows Next to be controlled remotely, that is, to
execute arbitrary code with the privileges of the user running Next.
Make sure you understand the security risks associated with this
before running this command."
  (slynk:create-server :port slynk-port :dont-close t)
  (echo "Slynk server started at port ~a" slynk-port))
```

If I do the definition in `:next`, it works fine.

Ambrevar:
Fixed in 3b2319c90ce15977b43009434458fbf270333a9a, thanks for reporting!

PR #716, opened 2020-05-04
Wip application mode

Ambrevar:
Fixes #117 . 

Previously I mentioned that we could do it by setting an input dispatcher.  I was wrong, because the input dispatcher works at the window level, while here we want something that works at the buffer level.

I think the right place to hi-jack the input at the buffer level is when we collect the keymaps in `current-keymaps`.  Currently, `current-keymaps` takes a window parameter.  I believe it should take a buffer (or minibuffer) parameter.

Thoughts?

### DIFF
--- a/source/application-mode.lisp
+++ b/source/application-mode.lisp
@@ -1,11 +1,31 @@
-;;; application-mode.lisp --- major mode for "web applications"
-;;; has no history or bindings to avoid interference with "web applications"'
+(uiop:define-package :next/application-mode
+    (:use :common-lisp :trivia :next)
+  (:documentation "Forward all keybindings to the web view except those in the `override-map'."))
+(in-package :next/application-mode)
 
-(in-package :next)
-
-;; TODO: Finish it!  We can move the active modes to a list of backed up modes,
-;; and have one "escape key" bound to a function that would restore the backup modes.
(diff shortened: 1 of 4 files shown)


source/application-mode.lisp
```diff
@@ -1,11 +1,31 @@
-;;; application-mode.lisp --- major mode for "web applications"
-;;; has no history or bindings to avoid interference with "web applications"'
+(uiop:define-package :next/application-mode
+    (:use :common-lisp :trivia :next)
```
jmercouris, line 2: trivia needed?

jmercouris:
You are correct, the current-keymaps should accept a buffer parameter. The minibuffer is also a buffer, so that is fine.

jmercouris:
That will be more important as we transition to having multiple buffers per window

source/application-mode.lisp
Ambrevar: Nope, just copy-pasted!

Ambrevar:
OK to merge then?

jmercouris:
Go for it! Thank you!

Ambrevar:
Merged with f54ce0c7d702d3471b0432c53a7d6b8bac19045c.  I'll update current-keymaps just now.

PR #718, opened 2020-05-04
Bind keys to function symbols instead of functions.

### LINKED ISSUE #575: define-command does not redefine + creates doubles in M-x
From the REPL, evaluating a  new definition of a `define-command` does not change the command of the running instance.

Besides, it's possible to create duplicate entries in M-x while it should not be.

Ambrevar:
Now calling a redefined function from a binding works.

Fixes #575.


jmercouris:
Yay!

Ambrevar:
OK to merge?

jmercouris:
Yes, looks ready to me!

do the forms: `(:code (binding-keys #'search-buffer))` need to be changed?

jmercouris:
I would imagine that they do... just checking though :-) I see no other issues

Ambrevar:
Yes they do, thanks for catching this!

Ambrevar:
Merged with 92a2905cfed39b0f9a0d0967f444a234d89e5658.

PR #723, opened 2020-05-08
build-scripts/guix: Use cl-* inputs and rely on the Makefile.

### LINKED ISSUE #680: Next hangs after a while (Guix build)
It seems that Next hangs after a while.

It happened to me after many hours, and it seems when I tried to open a URL externally.
(Could be a red herring.)

Possible causes:
- Unhandled exception at the FFI / javascript level.  If so, it should be easy to fix.
- Corruption at the FFI library level (like the C-v issue).  If so, we need to fix the libraries or find a way to not trigger the issue.

@jellelicht If you have a recipe please share :)

Ambrevar:
Fixes #680.
First we must wait until upstream has merged https://issues.guix.info/issue/41135.

### DIFF
--- a/build-scripts/guix.scm
+++ b/build-scripts/guix.scm
@@ -25,10 +25,10 @@
              (guix packages)
              (guix licenses)
              (guix git-download)
-             (guix build-system asdf)
+             (guix build-system gnu)
              (guix build-system glib-or-gtk)
-             (guix build-system trivial)
              (gnu packages)
+             (gnu packages lisp)
              (gnu packages lisp-xyz)
              (gnu packages gnome)
              (gnu packages gstreamer)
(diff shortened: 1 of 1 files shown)


jellelicht:
I get the following on this branch, with the 41135 applied to guix:

```
Unhandled LOAD-FOREIGN-LIBRARY-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                                  {1000580083}>:
  Unable to load any of the alternatives:
   ("libgobject-2.0.so.0" "libgobject-2.0.so")
```

Ambrevar:
Did you apply patch v2 (in 4 parts)?

jellelicht:
There is no v2 4/4, at least not in my mailbox, nor the webinterface

Ambrevar:
Damn!  Why is git send-email dropping so many of my patches?  I've resent the patch manually as an attachment.

Ambrevar:
Merged with b24e9a3459aefbe1981625405b222e6530ff7dcf.

PR #726, opened 2020-05-10
Tag completion filter

Ambrevar:
@jmercouris This fixes the tag completion we talked about.

It also fixes the search engine completion while we are at it.  Try it, it's super useful! (set-url -> press any char then TAB).

While working on it, I couldn't help struggling with the minibuffer implementation.  Minibuffer rewrite is highest priority post 2.0 :)

### DIFF
--- a/source/minibuffer.lisp
+++ b/source/minibuffer.lisp
@@ -448,13 +448,13 @@ The new webview HTML content it set as the MINIBUFFER's `content'."
         (ffi-window-set-minibuffer-height (current-window) 0))))
 
 (defun insert (characters &optional (minibuffer (current-minibuffer)))
-  ;; Set cursor before buffer to ensure cursor is never higher than buffer length.
-  (let ((old-cursor-position (input-cursor-position minibuffer)))
-    (incf (input-cursor-position minibuffer) (length characters))
-    (setf (input-buffer minibuffer)
-          (str:insert characters
-                      old-cursor-position
-                      (input-buffer minibuffer))))
+  (with-accessors ((buffer input-buffer) (cursor input-cursor-position)) minibuffer
+    ;; Set cursor before buffer to ensure cursor is never higher than buffer length.
+    (let ((old-cursor-position cursor))
+      (incf cursor (length characters))
+      (setf buffer (str:insert characters
+                               old-cursor-position
+                               buffer))))
   (state-changed minibuffer)
   (update-display minibuffer))
 
--- a/source/minibuffer.lisp
+++ b/source/minibuffer.lisp
@@ -576,6 +576,37 @@ The new webview HTML content it set as the MINIBUFFER's `content'."
         (incf position)
         position)))
 
+(defvar +white-spaces+ '(#\space #\no-break_space))
+
+(defun word-start (s position)
+  "Return the index of the beginning word at POSITION in string S."
+  (apply #'max
+         (mapcar (lambda (char)
+                   (let ((pos (position char s
+                                        :end position
+                                        :from-end t)))
+                     (if pos
+                         (1+ pos)
+                         0)))
+                 +white-spaces+)))
+
+(defun word-end (s position)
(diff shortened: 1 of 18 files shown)


jmercouris:
This is amazing! I just tried it out! I had looked at the source earlier, it looks good to me as well! I think I had one or two comments, but as far as I am concerned it is ready.

And yeah, you are right, time for the minibuffer factorization/librification, post 2.0!

source/minibuffer.lisp
```diff
@@ -448,11 +448,13 @@ The new webview HTML content it set as the MINIBUFFER's `content'."
...
-  (setf (input-buffer minibuffer)
-        (str:insert characters
-                    (input-cursor-position minibuffer)
-                    (input-buffer minibuffer)))
-  (incf (input-cursor-position minibuffer) (length characters))
+  (with-accessors ((buffer input-buffer) (cursor input-cursor-position)) minibuffer
+    ;; Set cursor before buffer to ensure cursor is never higher than buffer length.
+    (let ((old-cursor-position cursor))
+      (incf cursor (length characters))
+      (setf buffer (str:insert characters
```
jmercouris, line 455: I'm not sure I understand this. Would it be possible to ?
```
(setf buffer ...)
(incf cursor (length characters))
```

that way you do not need the `LET`

source/minibuffer.lisp
```diff
@@ -485,11 +487,14 @@ The new webview HTML content it set as the MINIBUFFER's `content'."
   "Delete character before cursor."
   (with-accessors ((buffer input-buffer) (cursor input-cursor-position)) minibuffer
     (unless (= cursor 0)
-      (setf buffer
-            (concatenate 'string
-                         (subseq buffer 0 (- cursor 1))
-                         (subseq buffer cursor (length buffer))))
-      (decf cursor)))
+      (let ((old-cursor cursor))
```
jmercouris, line 490: Same comment as above, I am not sure I understand why we need a `LET` to capture the old cursor state when we can just perform operations against the cursor and THEN mutate it

source/minibuffer.lisp
```diff
@@ -622,8 +659,9 @@ The new webview HTML content it set as the MINIBUFFER's `content'."
 (define-command kill-whole-line (&optional (minibuffer (current-minibuffer)))
   "Delete all characters in the input."
   (with-accessors ((buffer input-buffer) (cursor input-cursor-position)) minibuffer
-    (setf buffer ""
-          cursor 0))
+    ;; Set cursor before buffer to ensure cursor is never higher than buffer length.
```
jmercouris, line 662: Why the reordering :-D?

source/minibuffer.lisp
Ambrevar: See the above comment! :)
The minibuffer is highly stateful, hence this iterative execution order issue.

source/minibuffer.lisp
Ambrevar: Nope, because then between the two mutation the cursor might be out of bounds compared to the new state of the buffer.

source/minibuffer.lisp
Ambrevar: Same as above, same comment :)

source/minibuffer.lisp
jmercouris: mutating the cursor position updates the minibuffer display or something ?

Ambrevar:
The buffer has a writer method that updates the list of candidates.

jmercouris:
Ah right, I remember now... had to be modified for input-buffer!

Ambrevar:
OK to merge as is, then?

jmercouris:
Yes! :-)

Ambrevar:
I'm away today, feel free to merge, otherwise I'll do it tomorrow!

jmercouris:
Fertig!

Ambrevar:
Thanks!

PR #730, opened 2020-05-13
Search engines

### LINKED ISSUE #729: Make a class for search engines
Currently we represent search engines as an alist.  This is poor practice, our code is not as clear as it could be.

Turning it into a class is trivial.

Bonus: We can add an `object-string` and an `object-display` method to display search engines in the minibuffer.

Ambrevar:
Fixes #729.

While I was at it, we now display the full URL of search-engines when listed in the minibuffer.
Besides, we now fall back to the first search engines when none of them have the "default" shortcut.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -417,6 +417,35 @@ BUFFER's modes."
     (dolist (mode (modes buffer))
       (on-signal-load-finished mode url))))
 
+(defclass-export search-engine ()
+  ((shortcut :initarg :shortcut
+             :accessor shortcut
+             :type string
+             :initform ""
+             :documentation "The word used to refer to the search engine, for
+instance from the `set-url' commands.")
+   (search-url :initarg :search-url
+               :accessor search-url
+               :type string
+               :initform ""
+               :documentation "The URL containing a '~a' which will be replaced with the search query.")
+   (fallback-url :initarg :fallback-url
+                 :accessor fallback-url
+                 :type string
+                 :initform ""
+                 :documentation "The URL to fall back to when given an empty
+query.  This is optional: if empty, use `search-url' instead with ~a expanded to
+the empty string.")))
+
+(defmethod object-string ((engine search-engine))
+  (shortcut engine))
+
+(defmethod object-display ((engine search-engine))
+  (format nil "~a~a ~a"
+          (shortcut engine)
+          (make-string (max 0 (- 10 (length (shortcut engine)))) :initial-element #\no-break_space)
+          (search-url engine)))
+
 (defclass-export browser ()
   ((remote-execution-p :accessor remote-execution-p
                        :initarg remote-execution-p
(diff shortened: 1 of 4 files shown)


source/browser.lisp
```diff
@@ -417,6 +417,35 @@ BUFFER's modes."
     (dolist (mode (modes buffer))
       (on-signal-load-finished mode url))))
 
+(defclass-export search-engine ()
+  ((shortcut :initarg :shortcut
+             :accessor shortcut
+             :type string
+             :initform ""
```
jmercouris, line 424: initform does not appear needed. if we use it, perhaps it should be an error so that we cannot make a search-engine without passing in a shortcut (same for search-url)

so:
```
:initform (error ...)
```

thoughts?

source/buffer.lisp
```diff
@@ -207,26 +218,24 @@ complete against a search engine."
     ;; Complete a search engine name.
     ((and (not (str:emptyp (input-buffer minibuffer)))
           (zerop (completion-cursor minibuffer)))
-     (let* ((engines (mapcar #'first (search-engines *browser*)))
+     (let* ((engines (search-engines *browser*))
             (matching-engines
               (remove-if (complement (alex:curry #'str:starts-with-p (input-buffer minibuffer)))
```
jmercouris, line 223: could use remove-if-not instead?

jmercouris: if you wish to be explicit and turn the result into a generalized boolean, can do `(null i)`

source/types.lisp
```diff
@@ -59,33 +61,6 @@ Example:
   (assert (not (typep '() 'alist-of-strings)))
   (assert (not (typep nil 'alist-of-strings))))
 
-(defun alist-of-string+2strings-p (alist)
```
jmercouris, line 62: Yay nice rm :-)

jmercouris:
Looks great! I left some comments about some ideas for changing the API a little (like forcing args to make-instance)

source/buffer.lisp
Ambrevar: remove-if-not is deprecated, like all the -not functions.
Are you saying that you'd like (lambda (s) (not (str:starts-with-p ...)) instead?

source/browser.lisp
Ambrevar: Initform is almost needed unless you want to catch uninitialized sltos at runtime.
But you are right, (error ...) is better fore the first 2 slots since they are mandatory.

source/buffer.lisp
jmercouris: apparently it is deprecated! however #lisp assured me that it is OK, either way is OK, I would say just leave it then, if it ain’t broke don’t fix it!

Ambrevar:
Merged with 9e22579bf97320f79776c2d0baba0f599d72a4a7.

PR #762, opened 2020-05-27
Correct slight typo in MANUAL.org

DieracDelta:
Having just done this for the first time, it took me a minute to realize `start-swank` was meant here instead of `start-slime`. Small correction, but should help make things clearer.


jmercouris:
Thank you!

PR #766, opened 2020-05-29
Add an RPM build.

ralt:
Some caveats:

- This needs an SBCL built with --sb-linkable-runtime
- This needs to skip sbcl 2.0.4
- This needs current master of https://github.com/cffi/cffi
- This is missing some bit about not reloading foreign libraries on
  startup, but shouldn't be hard to do once the other questions are
  answered.

The current code assumes that during the release build, the `VERSION`
environment variable exists.

The code currently relies on https://github.com/jordansissel/fpm. This
means creating a .deb is also very easy.

That said, to build the package correctly, you need to run it on the
distribution you intend to distribute the package to. Which means you
need one build per distribution version, e.g. fedora 31, fedora 32,
debian 9, debian 10, debian 11, ubuntu 20.04, etc etc. Which shouldn't
be too hard to do on docker, but that opens the question of: how are
the current release builds done?

Anyway, here is some goodies:

```
[ralt@tag next]$ ls -lh next-1.0.0-1.x86_64.rpm
-rw-rw-r--. 1 ralt ralt 28M May 30 01:21 next-1.0.0-1.x86_64.rpm

[ralt@tag next]$ rpm -ql next
/usr/bin/next
/usr/lib/.build-id
/usr/lib/.build-id/1e
/usr/lib/.build-id/1e/875a286ae31e5065ce73525b20d7db4ecd2bbd

[ralt@tag next]$ rpm -qi next
Name        : next
[... skipping ...]

[ralt@tag next]$ rpm -qR next
cairo
gdk-pixbuf2
glib2
glibc
gtk3
libfixposix-devel
openssl-libs
pango
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
webkit2gtk3-devel
```

### DIFF
--- a/build-scripts/build-rpm.lisp
+++ b/build-scripts/build-rpm.lisp
@@ -1,0 +1,74 @@
+(defpackage :next/build-rpm
+  (:use :cl)
+  (:import-from :asdf #:perform #:system)
+  (:import-from :cffi-toolchain #:static-program-op)
+  (:import-from :cffi
+                #:foreign-library-pathname
+                #:foreign-library-type
+                #:list-foreign-libraries
+                #:close-foreign-library)
+  (:import-from :ppcre #:split)
+  (:import-from :uiop
+                #:run-program
+                #:getenv
+                #:register-image-dump-hook)
+  (:import-from :next #:+version+))
+
+(in-package :next/build-rpm)
+
+(defun find-library-rpm-packages (ldconfig library)
+  (with-input-from-string (s ldconfig)
+    ;; skip the first line
+    (read-line s)
+
+    (loop for line = (read-line s nil 'eof)
+       until (eq line 'eof)
+       when (let ((parts (split " " (subseq line 1))))
+              (string= (first parts) (namestring (foreign-library-pathname library))))
+       collect (let ((parts (split " " (subseq line 1))))
+                 (run-program
+                  `("rpm" "-q" "--whatprovides" ,(first (last parts)) "--qf" "%{NAME}")
+                  :output '(:string))))))
+
+(defun find-rpm-packages (libraries)
+  (let ((ldconfig (run-program "ldconfig -p" :output '(:string))))
+    (delete-duplicates
+     (reduce (lambda (packages library)
+               (append
+                packages
+                (unless (eq (foreign-library-type library) :grovel-wrapper)
+                  (find-library-rpm-packages ldconfig library))))
+             libraries
+             :initial-value nil)
+     :test #'string=)))
+
+;;; Make sure we close statically linked libraries.
+;;; Remove when this or similar is done in cffi: https://github.com/cffi/cffi/pull/163
+(register-image-dump-hook
+ (lambda ()
+   (loop for library in (list-foreign-libraries)
+      when (eq (foreign-library-type library) :grovel-wrapper)
+      do (close-foreign-library library))))
+
+(defclass build-rpm (static-program-op)
+  ())
+
+(defmethod perform ((operation build-rpm) (system system))
+  (call-next-method operation system)
+
+  (let ((deps (delete-duplicates
+               (append
+                (find-rpm-packages (list-foreign-libraries))
+                ;; sbcl images always depends on those
+                #+sbcl '("zlib" "glibc")))))
+    (run-program `("fpm" "-s" "dir"
+                         "-t" "rpm"
+                         "-n" "next"
+                         "-v" ,+version+
+                         ,@(mapcar (lambda (dep) (format nil "--depends=~a" dep)) deps)
+                         "next=/usr/bin/"
+                         "assets/next.desktop=/usr/share/applications/")
+                 :output :interactive
+                 :error-output :interactive)))
+
+(setf (find-class 'asdf::build-rpm) (find-class 'build-rpm))
--- a/next.asd
+++ b/next.asd
@@ -135,6 +135,17 @@
   :build-pathname "next"
   :entry-point "next:entry-point")
 
+(asdf:defsystem :next/build-rpm
+  :depends-on (:cffi-toolchain :cl-ppcre :next)
+  :pathname "build-scripts/"
+  :components ((:file "build-rpm")))
+
+(asdf:defsystem :next/gtk-application-rpm
+  :depends-on (:next/build-rpm :next/gtk)
+  :build-operation "build-rpm"
+  :build-pathname "next"
+  :entry-point "next:entry-point")
+
 (asdf:defsystem :next/qt-application
   :depends-on (:next/qt)
   :build-operation "program-op"


ralt:
PS: I should make a library out of this asdf class.

build-rpm/build-rpm.lisp
```diff
@@ -0,0 +1,43 @@
...
+                  (format nil "rpm -q --whatprovides ~a --qf '%{NAME}'" (first (last parts)))
+                  :output '(:string))))))
+
+(defun find-rpm-packages (libraries)
+  (let ((ldconfig (uiop:run-program
+                   "ldconfig -p"
+                   :output '(:string))))
+    (remove-duplicates
+     (alexandria:flatten
+      (remove-if
```
Ambrevar, line 24: Use `delete nil` instead.

build-rpm/build-rpm.lisp
```diff
@@ -0,0 +1,43 @@
+(defpackage :next/build-rpm
+  (:use :cl :asdf :cffi-toolchain))
```
Ambrevar, line 2: Instead of `:use`, write `(import-from :asdf)` and `(import-from :cffi-toolchain)`.

build-rpm/build-rpm.lisp
```diff
@@ -0,0 +1,43 @@
...
+
+(defun find-rpm-package (ldconfig library)
+  (with-input-from-string (s ldconfig)
+    (read-line s)
+    (loop for line = (read-line s nil 'eof)
+       until (eq line 'eof)
+       when (let ((parts (ppcre:split " " (subseq line 1))))
+              (string= (first parts) (namestring (cffi:foreign-library-pathname library))))
+       collect (let ((parts (ppcre:split " " (subseq line 1))))
+                 (uiop:run-program
```
Ambrevar, line 14: uiop:run-program takes either a string or a list.  If a string, it invokes the shell, which is unreliable here.
Prefer a list, e.g. `'("rpm" "-q" ...)`.

build-rpm/build-rpm.lisp
```diff
@@ -0,0 +1,43 @@
+(defpackage :next/build-rpm
+  (:use :cl :asdf :cffi-toolchain))
+
+(in-package :next/build-rpm)
+
+(defun find-rpm-package (ldconfig library)
+  (with-input-from-string (s ldconfig)
+    (read-line s)
```
Ambrevar, line 8: Why do you read the line here?  Do you need to skip the first line?  If so, please add an explicit comment.

build-rpm/build-rpm.lisp
```diff
@@ -0,0 +1,43 @@
...
+       until (eq line 'eof)
+       when (let ((parts (ppcre:split " " (subseq line 1))))
+              (string= (first parts) (namestring (cffi:foreign-library-pathname library))))
+       collect (let ((parts (ppcre:split " " (subseq line 1))))
+                 (uiop:run-program
+                  (format nil "rpm -q --whatprovides ~a --qf '%{NAME}'" (first (last parts)))
+                  :output '(:string))))))
+
+(defun find-rpm-packages (libraries)
+  (let ((ldconfig (uiop:run-program
```
Ambrevar, line 19: Same.

build-rpm/build-rpm.lisp
```diff
@@ -0,0 +1,43 @@
...
+                   "ldconfig -p"
+                   :output '(:string))))
+    (remove-duplicates
```
Ambrevar, line 22: delete-duplicates

build-rpm/build-rpm.lisp
```diff
@@ -0,0 +1,43 @@
...
+  ())
+
+(defmethod perform ((o build-rpm) (s system))
+  (call-next-method o s)
+  (let ((deps (find-rpm-packages (cffi:list-foreign-libraries))))
+    (uiop:run-program (format
+                       nil
+                       "fpm -s dir -t rpm -n next ~a -v ~a next=/usr/bin/"
+                       (format nil "~{--depends ~a~^ ~}" deps)
+                       (or (uiop:getenv "VERSION") "1.0.0"))
```
Ambrevar, line 39: Is this supposed to be the Next version? If so, you can get it programmatically from the .asd or from next:+version+ value.  Let me know if you need help with this.

build-rpm/build-rpm.lisp
```diff
@@ -0,0 +1,43 @@
...
+      (remove-if
+       #'null
+       (mapcar (lambda (library) (find-rpm-package ldconfig library)) libraries)))
+     :test #'string=)))
+
+(defclass build-rpm (static-program-op)
+  ())
+
+(defmethod perform ((o build-rpm) (s system))
+  (call-next-method o s)
```
Ambrevar, line 33: Nit: Instead of `o` and `s`, you could use full words, e.g. `operation` and `system`.

next.asd
```diff
@@ -135,6 +135,17 @@
   :build-pathname "next"
   :entry-point "next:entry-point")
 
+(asdf:defsystem :next/build-rpm
+  :depends-on (:cffi-toolchain :cl-ppcre)
+  :pathname "build-rpm/"
```
Ambrevar, line 140: Can this go in `build-scripts` like the rest?

build-rpm/build-rpm.lisp
Ambrevar: And alexandria.

next.asd
```diff
@@ -135,6 +135,17 @@
   :build-pathname "next"
   :entry-point "next:entry-point")
 
+(asdf:defsystem :next/build-rpm
+  :depends-on (:cffi-toolchain :cl-ppcre)
```
Ambrevar, line 139: Alexandria is missing.

build-rpm/build-rpm.lisp
ralt: Sure :)

Ambrevar:
Excellent, thanks a lot!

- Could you also include the assets like the icons and the .desktop file?

- --sb-linkable-runtime should be the default, isn't it?  
- Why isn't sbcl 2.0.4 working?
- Which version of CFFI do you need?  Isn't 0.21 enough? Why master?  Do you need a specific fix?

About your question for the different distros: there is nothing for now but I'd like to set up a pipeline to build the Guix pack, the .rpm / .deb, etc.
I wonder if we can build the macOS .app in a pipeline too.

build-rpm/build-rpm.lisp
ralt: Makes sense about the comment.

build-rpm/build-rpm.lisp
ralt: It's not "unreliable", it's sh, which must be a POSIX shell. So it's really "don't use with untrusted input", which isn't the case there, so it doesn't matter much. But I'll do the change, sure.

build-rpm/build-rpm.lisp
ralt: Why is that so? Apparently delete-duplicates can potentially modify the sequence, which is not really what I want?

build-rpm/build-rpm.lisp
ralt: That's where it was! Thanks, will use it.

build-rpm/build-rpm.lisp
ralt: This is actually following asdf (and cffi-toolchain) coding standards there. Still want the long ones?

Ambrevar:
> It's not "unreliable", it's sh, which must be a POSIX shell. So it's really "don't use with untrusted input", which isn't the case there, so it doesn't matter much. But I'll do the change, sure.

It's "unreliable" in the sense that it will perform all the shell stuff,
including globbing, etc.  It might have unintended effects plus it adds a
small overhead that we don't need here.

build-rpm/build-rpm.lisp
Ambrevar: You've created a new list with mapcar, so it's safe to modify it in place.

ralt:
> Could you also include the assets like the icons and the .desktop file?

Absolutely! Just need to know the full list there.

> --sb-linkable-runtime should be the default, isn't it?

I'm not sure.

> Why isn't sbcl 2.0.4 working?

A bug in the sbcl home path preventing static-program-op from working.

> Which version of CFFI do you need? Isn't 0.21 enough? Why master? Do you need a specific fix?

Actually, the fixes I was thinking about are not necessary here, my bad.

build-rpm/build-rpm.lisp
Ambrevar: Standard, or just de facto conventions?  Do you have a reference that explains why we expect `o` and `s` here?  In general, long names are always preferable to one-letter names since they are more readable and meaningful.

next.asd
ralt: Sure.

build-rpm/build-rpm.lisp
ralt: But why prefer the unsafe version when the safe version exists and does the job just as well?

build-rpm/build-rpm.lisp
ralt: De facto convention, which counts as "mostly standard" in my book :P

Again, I don't particularly care.

ralt:
> About your question for the different distros: there is nothing for now but I'd like to set up a pipeline to build the Guix pack, the .rpm / .deb, etc.

I would recommend gitlab CI but... :)

build-rpm/build-rpm.lisp
Ambrevar: Your list is a trash-away list.  All `remove` versions have a `delete*` counterpart exactly meant for these kinds of list.
 
If you want to do purely functional code, you should not use mapcar and list.  (See the fset library for instance, but I don't recommend it here for such simple code.)

Since this is not purely functional, calling `remove*` here is a waste of resources.
Certainly a nit here but it's good memory management practice.

build-rpm/build-rpm.lisp
Ambrevar: If the de facto practice can be argued against with rational arguments, I'd rather go for the rational one ;)

Ambrevar:
> I would recommend gitlab CI but... :)

But... This is GitHub, is this what you meant?
I've used GitLab CI before, I'll see if I can do something similar on
GitHub.  Either way, it's not problem.

vindarel:
Nice!

> how are the current release builds done?

On our machines. I built the .deb so far

> I would recommend gitlab CI but... :)

\+1 

https://about.gitlab.com/solutions/github/

> This will automatically prompt GitLab CI/CD to run whenever code is pushed to GitHub and post CI/CD results back to both GitLab and GitHub when completed.

:D

So the fpm command is

                           "fpm -s dir -t rpm -n next [dependencies] -v [version] next=/usr/bin/"

Does that mean the `next` binary was built previously and fpm creates a binary (no source) package?

ralt:
> Does that mean the `next` binary was built previously and fpm creates a binary (no source) package?

Yup! That's why the asdf system is subclassing `CFFI-TOOLCHAIN:STATIC-PROGRAM-OP`, so that the `CALL-NEXT-METHOD` will create the statically linked binary.

ralt:
@Ambrevar would you mind reviewing again? I have done a bunch of changes following your review, plus some more.

That said, what is remaining is figuring out the full list of assets and where to put them. I'd appreciate some help on that :)

ralt:
If using gitlab CI is what you guys want to do, you can probably draw some inspiration from what I previously did here: https://gitlab.com/ralt/ballish/-/blob/master/.gitlab-ci.yml (YAML references are pretty useful!)

Just to be clear, this is what I use with this PR to generate the .rpm:

```
sbcl --load ~/quicklisp/setup.lisp --eval '(ql:quickload :cffi-grovel)' --eval '(ql:quickload :next/gtk-application-rpm)' --eval '(asdf:make :next/gtk-application-rpm)' --quit
```

And to be even more clear... the idea of this binary is that it statically links the grovel/wrapper files, and for shared libraries, this PR is figuring out which rpm packages are providing them, and adding them as dependencies of the `next` package. This way, when it will be installed, the dependencies providing the shared libraries will be installed alongside.

build-scripts/build-rpm.lisp
```diff
@@ -0,0 +1,74 @@
...
+  (:import-from :cffi
+                #:foreign-library-pathname
+                #:foreign-library-type
+                #:list-foreign-libraries
+                #:close-foreign-library)
+  (:import-from :ppcre #:split)
+  (:import-from :uiop
+                #:run-program
+                #:getenv
+                #:register-image-dump-hook)
```
Ambrevar, line 14: Maybe you misunderstood what I meant: write `(:import-from :cffi)` without actually importing anything.  This tells Common Lisp what you depend on, then you can prefix all the CFFI symbols with `cffi:`.

build-scripts/build-rpm.lisp
```diff
@@ -0,0 +1,74 @@
...
+  (:import-from :next #:+version+))
+
+(in-package :next/build-rpm)
+
+(defun find-library-rpm-packages (ldconfig library)
+  (with-input-from-string (s ldconfig)
+    ;; skip the first line
```
Ambrevar, line 21: But why?  What does it contain?

build-scripts/build-rpm.lisp
Ambrevar: Same with UIOP, etc.

build-scripts/build-rpm.lisp
```diff
@@ -0,0 +1,74 @@
...
+       until (eq line 'eof)
+       when (let ((parts (split " " (subseq line 1))))
+              (string= (first parts) (namestring (foreign-library-pathname library))))
+       collect (let ((parts (split " " (subseq line 1))))
+                 (run-program
+                  `("rpm" "-q" "--whatprovides" ,(first (last parts)) "--qf" "%{NAME}")
+                  :output '(:string))))))
+
+(defun find-rpm-packages (libraries)
+  (let ((ldconfig (run-program "ldconfig -p" :output '(:string))))
```
Ambrevar, line 34: Nit: Pass a list to `run-program`.

build-scripts/build-rpm.lisp
```diff
@@ -0,0 +1,74 @@
...
+    (delete-duplicates
+     (reduce (lambda (packages library)
+               (append
+                packages
+                (unless (eq (foreign-library-type library) :grovel-wrapper)
+                  (find-library-rpm-packages ldconfig library))))
+             libraries
+             :initial-value nil)
```
Ambrevar, line 42: Nit: I think the initial-value of nil is the default, so you can leave this out.

Ambrevar:
Looking great!  Just a few more nits and we are good to go!

About the assets: the official way to list them is from `make install-assets`.
In short, it's the .desktop file and the .png.

Maybe you could run `make DESDIR=wherever/you/want install-assets`, then install everything from there.  I'm not sure what's best for a .rpm.

build-scripts/build-rpm.lisp
vindarel: We can use the prefixes when we declare the dependency into the .asd, even without `(:import-from :cffi)` on the package declaration. So what does this import bring? Is it (only) for package-inferred-systems?

Ambrevar:
vindarel <notifications@github.com> writes:

> We can use the prefixes when we declare the dependency into the .asd, even without `(:import-from :cffi)` on the package declaration. So what does this import bring? Is it (only) for package-inferred-systems?

You are right.  The import-from is optional, but since the initial patch
_use_d the packages I thought we could keep it.

`import-from` is useful to declare which libraries the package depends on.

jmercouris:
@Ambrevar, Florian requested that one of us take over and finish up the PR and get done with all of the nitty gritty :-)

If you like, I can try to do that

Ambrevar:
OK!

I'll have a look at the CI in the meanwhile.

Asalle:
any progress on this one? thanks for the amazing browser guys, it's been a bit tough building it from source on fedora every time :)

jmercouris:
No progress from me yet, I'm wrapping up a couple of things and will be trying to take a shot at this. I'm not a Linux user, so it may take some extra time :-)

Ambrevar:
@jmercouris: I'm working on the pipeline which should generate the RPM.
I haven't had much time to dedicate to this recently, but I'll get
there, no worries.

jmercouris:
Closing this in favor of https://github.com/atlas-engineer/nyxt/pull/825

jmercouris:
https://github.com/atlas-engineer/nyxt/pull/825

PR #789, opened 2020-06-12
Nyxt renaming

Ambrevar:
- Compiles OK
- Binary OK
- Tests OK

@jmercouris Let me know what you think.


Ambrevar:
Guix package builds.

Can you test if the macOS bundle builds properly?

libraries/download-manager/engine.lisp
```diff
@@ -36,7 +36,7 @@ The channel return value is a `download'.")
   "Initialize the download manager.
 This is called automatically from `resolve', but it can also be called manually
 beforehand in order to specify different initialization argument."
-  (setf lparallel:*kernel* (lparallel:make-kernel worker-count :name "next-kernel"))
+  (setf lparallel:*kernel* (lparallel:make-kernel worker-count :name "download-manager-kernel"))
```
jmercouris, line 39: why the kernel suffix?

source/browser.lisp
```diff
@@ -48,7 +48,21 @@ Parent directories are created if necessary."
        (with-maybe-gpg-file (,stream path ,@options)
          ,@body))))
 
-(defclass-export window ()
+(eval-when (:compile-toplevel :load-toplevel :execute)
```
jmercouris, line 51: Why eval when? why not export symbols all the time?

source/browser.lisp
```diff
@@ -723,22 +804,6 @@ cl-enchant (broker-list-dicts).")
                   :documentation "The external editor to use for
 editing files. It should be specified as a complete string path to the
 editor executable.")))
-(unexport
```
jmercouris, line 726: I really like this change :-) exporting explicitly rather than un-exporting explicitly makes more sense to me :-D

source/command.lisp
```diff
@@ -104,12 +104,12 @@ deprecated and by what in the docstring."
            (echo-warning "~a is deprecated." ',name)
            ,@body)))))
 
-(defun next-packages ()                 ; TODO: Export a customizable *next-packages* instead?
-  "Return all package designators that start with 'next' plus Next own libraries."
+(defun nyxt-packages ()                 ; TODO: Export a customizable *nyxt-packages* instead?
```
jmercouris, line 107: TODO makes a good point, worth considering, I'm undecided

source/jump-heading.lisp
```diff
@@ -7,12 +7,16 @@
   (:documentation "A heading. The inner-text must not be modified, so
   than we can jump to the anchor of the same name."))
 
-(defmethod object-string ((heading heading))
+(defmethod nyxt:object-string ((heading heading))
```
jmercouris, line 10: nice catch :-)

source/browser.lisp
jmercouris: Never-mind, I realized it some minutes into reviewing :-D

jmercouris:
I've made a commit to fix the macOS deployment, otherwise works!

Nicely done!

jmercouris:
Looking forward to fixing my branches :-D

libraries/download-manager/engine.lisp
Ambrevar: The kernel must be named.
Our external libraries should be nyxt-agnostic, say, if we publish them independently.

source/browser.lisp
Ambrevar: It's still very cumbersome to write and maintain while prone to errors (typos).
What I'd really like a an `:export` keyword in each slot, with an `:export` class option that specifies the default value.

Maybe something for Serapeum / defclass-std / defstar.

source/jump-heading.lisp
Ambrevar: We had a lot of these issues :(
Is there a way to raise an error when using defmethod without an existing defgeneric?

jmercouris: There is, SBCL supports this behavior if I remember correctly

Issue #778, opened 2020-06-04
Add force-https-mode


Ambrevar:
See the old-reddit-handler, we can implement it this way.

Ambrevar:
Need to fix #672 and #687 first.

Issue #687, opened 2020-04-23
Unnaceptable TLS certificate error


hsanjuan:
Trying to open things like https://post.de throws an `Unnaceptable TLS Certificate` message.

Why is it unacceptable?

Logs say:

```
WARNING:
   Couldn't handle multiple interfaces for #.(SB-SYS:INT-SAP #X7F60982DB370) of type GTlsCertificateGnutls: NIL.
```

which is very cryptic... so I am not sure if it's a matter of the certificate having weak crypto, or some bug in webkitgtk or even Next?

jmercouris:
Thank you for the report! My first inclination is that our bindings are slightly problematic, and that WebKitGTK is presenting us with some case we did not anticipate...

Ambrevar:
Héctor: post.de is indeed not included in the nss-certs.
You can try with Epiphany, same problem.

You can force-accepting the cerfiticates with
`add-domain-to-certificate-whitelist'.

If you want to make this permanent, see the doc of the above command
(press "f1 c") on commit 2b324fd5be2c58d3d1bddfc344ae5155ecf3e458.

Spoiler:

(setf next/certificate-whitelist-mode:*default-certificate-whitelist*
      '("post.de"))

hsanjuan:
@Ambrevar ok, but this is more about telling me "Could not find a trusted root authority for this certificate", and perhaps even printing the certificate information (and ultimately allowing me to ignore the issue).

But in fact the problem is that the CN declared by the certificate does not belong to the domain. Epiphany correctly tells about this.

Ambrevar:
I've opened https://github.com/atlas-engineer/next/issues/689 about
this, is that what you mean?

We can go fancy on certificates after 2.0.

hsanjuan:
No, what I mean is mostly specifying the reason that the certificate is invalid in the error page.

Ambrevar:
Yes, we can do that when fixing the aforementioned issue.

jmercouris:
Beyond the fact that the certificate is not acceptable, we should still probably address the bindings at some point... :-D

Ambrevar:
Which bindings?

jmercouris:
WebKitGTK+ bindings

jmercouris:
“Couldn’t handle multiple interfaces ...” whatever that means :-D

jmercouris:
Master has a fix for this now, please let us know if this works for you :-)

Issue #376, opened 2019-09-14
Automatic proxy/tor/noscript/adblocker black- and whitelisting


Ambrevar:
It would be nice if, when we disable/enable a mode for a URL, Next would remember it next time it loads that URL.

For instance, say you have noscript-mode enabled by default but you disable it for example.org.  In a new buffer or new Next instance, it'd be nice to have Next automatically disable noscript-mode when loading example.org again.

(Note: to make it explicit, we need a separate status buffer / bar to show the list of active modes.)

There are a couple of ways to do this:

1. With a set-url-hook handler, we can set the modes for the current buffer loading a given URL.  The modes are looked up in a `(URL . modes)` hash-table (which can be persisted to disk).  This hash-table can be filled from the enable/disable hook of each mode.  Doing this in hooks gives the user the flexibility to choose which mode status matters for the hash table (e.g. `vi-insert-mode` should probably not end up there). 

2. Add "toggle-noscript" commands that black/whitelist the URL, toggle Javascript on the platform port but leave the mode active in the buffer.  The mode would add a handler to set-url-hook.  When disabled, the mode removes the hook.  This way the user on has to add the mode to the list of default modes and then "teach" Next which website to black/whitelist with "toggle-noscript".  This logic should be reflected on all modes then.

Thoughts?

Ambrevar:
What about unnamed modes, for instance `(proxy-mode :proxy (make-instance ...))`?
This can only be evaluate, it cannot be stored as a symbol in the buffer "default-modes" for instance.

Technically we could serialize the anonymous mode and restore it with cl-prevalence, but is it worth it?

vindarel:
> For instance, say you have noscript-mode enabled by default but you disable it for example.org. In a new buffer or new Next instance, it'd be nice to have Next automatically disable noscript-mode when loading example.org again.

No, this is dangerous for the user. Specially if you replace "noscript" by "tor/proxy".

> Add "toggle-noscript" commands that black/whitelist the URL

toggle-noscript should toggle noscript temporarily and no more, we'd have another `noscript-whitelist` command that explicitely saves the state. Again, specially important with proxy mode.

The ability to call whitelist after toggle will be doable with a user hook, but it is the user's responsibility to add it. I don't expect it to be the default.

aartaka:
[...]

PR #792, opened 2020-06-15
Add force-https-mode

aartaka:
This adds the mode for obligatory switching to `https://` even if the link (be it button, hyperlink, or something user enters in the minibuffer) explicitly uses `http://` scheme.

## Motivation
There is a necessity behind this mode, because some, especially hardcoded, links/buttons/inputs contain non-HTTPS links and lead to less secure versions of websites. Enforcing HTTPS can make one's browsing safer, and it's in line with Nyxt care about users' safety.

## Caveats
Several issues that ideally need to be resolved before merging this are still open, but the `force-https-mode` works well enough even without resolving these issues (namely, #672 and #687). As a precaution measure, `*http-only-urls*` variable is added to disable HTTPS enforcement for individual addresses.

## How Has This Been Tested?
- Several certified websites were requested with `http://` scheme that was changed to `https://` on the requests hook. No surprises there.
- Several links from [Why No HTTPS](https://whynohttps.com/) were opened with this mode enabled. Some did break, but it was due to the websites themselves improperly handling TLS connection, so I can do nothing about that.
- [Website](https://post.de) with the certificate problem from #687 was forced to HTTPS too, with an expectable error message. After whitelisting it with `add-domain-to-certificate-whitelist`, the address loaded flawlessly.
- [HTTP-only-website](http://go.com) that was breaking in this mode (and on any attempt to use HTTPS in general, even in other browsers), have loaded flawlessly (yet insecurely) after whitelisting in `nyxt/force-https-mode:*http-only-urls*`.

I'll be glad to see your feedback on the code! If there are some inconsistencies with the appropriate style, I'll try to correct them as fast as I can. If you consider that this mode needs more testing and I can help with that, let me know too!

Closes: #778

### DIFF
--- a/source/force-https-mode.lisp
+++ b/source/force-https-mode.lisp
@@ -3,18 +3,10 @@
   (:documentation "Mode for enforcing HTTPS on any URL clicked/hinted/set by user."))
 (in-package :nyxt/force-https-mode)
 
-(serapeum:export-always '*http-only-urls*)
-(defparameter *http-only-urls* '()
-  "A list of the URLs to be ignored by force-https-mode.")
-
 (defun force-https-handler (resource)
   "Imposes HTTPS on any link with HTTP scheme."
   (let ((uri (quri:uri (url resource))))
-    (if (and (string= (quri:uri-scheme uri) "http")
-             (not (member-if #'(lambda (url)
-                                 (search (quri:uri-host uri)
-                                         (quri:uri-host (quri:uri url))))
-                             *http-only-urls*)))
+    (if (string= (quri:uri-scheme uri) "http")
         (progn (log:info "HTTPS enforced on ~a" (quri:render-uri uri))
                ;; FIXME: http-only websites are displayed as "https://foo.bar"
                ;; FIXME: some websites (e.g., go.com) simply time-out
@@ -31,10 +23,6 @@ and websites that still don't have HTTPS version (shame on them!).
 To escape \"Unacceptable TLS Certificate\" error:
 \(setf nyxt/certificate-whitelist-mode:*default-certificate-whitelist*
        '(\"your.unacceptable.cert.website\"))
-
-To escape lags and inaccessibility of some stubborn non-HTTPS websites:
-\(setf nyxt/force-https-mode:*http-only-urls*
-       '(\"stubborn.website\"))
 
 Example:
 
--- a/source/force-https-mode.lisp
+++ b/source/force-https-mode.lisp
@@ -1,0 +1,55 @@
+(uiop:define-package :nyxt/force-https-mode
+  (:use :common-lisp :nyxt)
+  (:documentation "Mode for enforcing HTTPS on any URL clicked/hinted/set by user."))
+(in-package :nyxt/force-https-mode)
+
+(serapeum:export-always '*http-only-urls*)
+(defparameter *http-only-urls* '()
+  "A list of the URLs to be ignored by force-https-mode.")
+
+(defun force-https-handler (resource)
+  "Imposes HTTPS on any link with HTTP scheme."
+  (let ((uri (quri:uri (url resource))))
+    (if (and (string= (quri:uri-scheme uri) "http")
+             (not (member-if #'(lambda (url)
+                                 (search (quri:uri-host uri)
+                                         (quri:uri-host (quri:uri url))))
+                             *http-only-urls*)))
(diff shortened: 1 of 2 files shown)


source/force-https-mode.lisp
```diff
@@ -0,0 +1,55 @@
+(uiop:define-package :nyxt/force-https-mode
+  (:use :common-lisp :nyxt)
+  (:documentation "Mode for enforcing HTTPS on any URL clicked/hinted/set by user."))
+(in-package :nyxt/force-https-mode)
+
+(serapeum:export-always '*http-only-urls*)
+(defparameter *http-only-urls* '()
+  "A list of the URLs to be ignored by force-https-mode.")
```
Ambrevar, line 8: I would leave this unexported for now until we figure out a more general way to do it in #376.

source/force-https-mode.lisp
```diff
@@ -0,0 +1,55 @@
...
+
+(defun force-https-handler (resource)
+  "Imposes HTTPS on any link with HTTP scheme."
+  (let ((uri (quri:uri (url resource))))
+    (if (and (string= (quri:uri-scheme uri) "http")
+             (not (member-if #'(lambda (url)
+                                 (search (quri:uri-host uri)
+                                         (quri:uri-host (quri:uri url))))
+                             *http-only-urls*)))
```
Ambrevar, line 17: Maybe shorter with `(find (quri:uri-host uri) *http-only-urls* :test #'search :key (alex:compose #'quri:uri-host #'quri:uri)`? (untested)

source/force-https-mode.lisp
```diff
@@ -0,0 +1,55 @@
...
+               ;;        is empty (old-reddit-handler causes the same effect,
+               ;;        I guess the problem is in the request hook)
+               (setf (quri:uri-scheme uri) "https"
+                     (quri:uri-port uri) (quri.port:scheme-default-port "https")
+                     (url resource) (quri:render-uri uri)))))
+  (values resource :forward))
+
+(define-mode force-https-mode ()
+  "Impose HTTPS on any link being set. Use at your own risk --
+can break websites whose certificates are not included in nss-certs
```
Ambrevar, line 31: "whoe certificates are not known" instead of citing nss-certs.  The use of nss-certs is an implementation detail of WebKitGTK which we don't control.

source/force-https-mode.lisp
```diff
@@ -0,0 +1,55 @@
...
+and websites that still don't have an HTTPS version (shame on them)!
```
Ambrevar, line 32: Why not a bit of humour :)
Typo: "an HTTPS version. (Shame on them!)"

Ambrevar:
Excellent work, thank you so much!

Can you have a look at the few nits I've posted, then I'll merge!

source/force-https-mode.lisp
aartaka: I'd better go with the parenthesised phrase as a part of the sentence it relates to, i.e. "version (shame on them!)." Sounds okay to you?

source/force-https-mode.lisp
aartaka: Sounds reasonable. 

This variable was made to be exported for the final user to tweak. Will leaving it out be better then? Leaving it out will simplify the code significantly and will be more in-line with the mechanism you've described in the mentioned issue.

source/force-https-mode.lisp
aartaka: I've left the article out too, if that's what you meant)

source/force-https-mode.lisp
aartaka: Sounds good, but we won't need it if the `*http-only-hosts*` will not be used)

aartaka:
Thank you for review!

Addressed things that you mentioned. Notably, `*http-only-urls*` variable is no longer there. If there will be a need in it, it's not that hard to bring it back, but the variable itself not that necessary now :)

Ambrevar:
Excellent, thanks a lot for this great addition!

Ambrevar:
Oops, I've merged too fast!  I forgot to ask you to squash the commits.  Well, for next time!

Ambrevar:
I've fixed some typos / conventions in d50f65d4e28e72bcd729790fa3e1646d68c365a9
 and used `when` instead of `if` in c0dffc22ccdee66ace502f026fc29c5f3f4e5ae8.

jmercouris:
Nice work!

Thanks a lot Artyom :-)

PR #800, opened 2020-06-21
GTK: Switch to gtk:gtk-main for FreeBSD

wahjava:
Launching `nyxt` on FreeBSD (`12.1-RELEASE` in my case), gives a white/gray/empty window, although in _verbose_ mode, i can see keypress events on console.

After applying this patch, `nyxt` works as expected. Although, I'm not sure if it's any other side-effects, but for now, it causes browser to work in FreeBSD.

Thanks!

### DIFF
--- a/source/renderer-gtk.lisp
+++ b/source/renderer-gtk.lisp
@@ -67,7 +67,7 @@ See https://github.com/atlas-engineer/nyxt/issues/740")
           (gdk:gdk-set-program-class "nyxt")
           (finalize browser urls startup-timestamp))
         (unless *keep-alive*
-          (gtk:join-gtk-main)))
+          (#-freebsd gtk:join-gtk-main #+freebsd gtk:gtk-main)))
       #+darwin
       (progn
         (setf gtk-running-p t)


Ambrevar:
Thanks for testing Nyxt on FreeBSD, it's great you managed to get it to work on an additional platform! :)

Instead of you fix, could you try to us the same branch as darwin?  E.g. something like this:

```lisp
(defmethod ffi-initialize ((browser gtk-browser) urls startup-timestamp)
  "gtk:within-main-loop handles all the GTK initialization. On
   GNU/Linux, Nyxt could hang after 10 minutes if it's not
   used. Conversely, on Darwin, if gtk:within-main-loop is used, no
   drawing happens. Drawing operations on Darwin MUST originate from
   the main thread, which the GTK main loop is not guaranteed to be
   on."
  (log:debug "Initializing GTK Interface")
  (if gtk-running-p
      (ffi-within-renderer-thread
       browser
       (lambda ()
         (finalize browser urls startup-timestamp)))
      #-(or darwin freebsd)
      (progn
        (setf gtk-running-p t)
        (gtk:within-main-loop
          (gdk:gdk-set-program-class "nyxt")
          (finalize browser urls startup-timestamp))
        (unless *keep-alive*
          (gtk:join-gtk-main)))
      #+(or darwin freebsd)
      (progn
        (setf gtk-running-p t)
        (gdk:gdk-set-program-class "nyxt")
        (finalize browser urls startup-timestamp)
        (gtk:gtk-main))))
```

jmercouris:
I had also hinted on IRC about probably needing some of the same darwin fixes. I guess we'll find out :-)

jmercouris:
The user contacted me on IRC, and if I remember correctly, they noted that the Darwin specific fixes did not seem to help...

jmercouris:
@Ambrevar is it okay for you to merge this in?

Ambrevar:
It's unfortunate we didn't hear back from the original submitter.  Let's add a comment point to this discussion and merge.

wahjava:
It only helps in rendering the webkit widget. The main issue is 100% CPU usage after few minutes, and application becoming unusable. On `truss` (equivalent of `strace` on FreeBSD), SBCL debugger gets invoked:

```
debugger invoked on a SIMPLE-ERROR in thread
#<THREAD "main thread" RUNNING {10006D8083}>:
  internal error 176: NIL; args=NIL

  0: [ABORT] Exit from the current thread.

("bogus stack frame")
0] backtrace

Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {10006D8083}>
0: ("bogus stack frame")
1: ("foreign function: _pthread_cond_wait")
2: ("foreign function: WTF::ThreadCondition::timedWait(WTF::Mutex&, WTF::WallTime)")
3: ("foreign function: WTF::ParkingLot::parkConditionallyImpl(void const*, WTF::ScopedLambda<bool ()> const&, WTF::ScopedLambda<void ()> const&, WTF::TimeWithDynamicClockType const&)")
4: ("foreign function: JSC::Heap::acquireAccessSlow()")
5: ("foreign function: JSC::JSLock::didAcquireLock()")
6: ("foreign function: JSC::JSRunLoopTimer::timerDidFire()")
7: ("foreign function: JSC::JSRunLoopTimer::Manager::timerDidFire()")
8: ("foreign function: WTF::RunLoop::TimerBase::secondsUntilFire() const")
9: ("foreign function: g_main_context_dispatch")
10: ("foreign function: g_main_context_pending")
11: ("foreign function: g_main_loop_run")
12: ("foreign function: gtk_main")
13: (GTK::%GTK-MAIN)
14: (GTK:GTK-MAIN)
15: (START NIL)
16: ((LAMBDA NIL :IN UIOP/IMAGE:RESTORE-IMAGE))
17: (UIOP/IMAGE:CALL-WITH-FATAL-CONDITION-HANDLER #<CLOSURE (LAMBDA NIL :IN UIOP/IMAGE:RESTORE-IMAGE) {1005ED0D2B}>)
18: ((FLET SB-UNIX::BODY :IN SB-EXT:SAVE-LISP-AND-DIE))
19: ((FLET "WITHOUT-INTERRUPTS-BODY-14" :IN SB-EXT:SAVE-LISP-AND-DIE))
20: ((LABELS SB-IMPL::RESTART-LISP :IN SB-EXT:SAVE-LISP-AND-DIE))
```

Ambrevar:
We had a similar issue before with Guix: it was resolved by building
using asdf:make instead of compile-bundle-op (in particular for the CFFI
deps like cl-webkit and cl-cffi-gtk).

How do you start Nyxt?  From the REPL or from a binary?
Could you try both and report?

How did you install the Common Lisp dependencies?

wahjava:
I used `ql:quickload` to fetch the dependencies, and then `(asdf:make :nyxt/gtk-application)` to build the binary as mentioned in the Makefile. Started Nyxt with binary. I will tryout the REPL method and report back.

Thanks for the hint.

wahjava:
Assuming by running from REPL, you meant:

```
* (asdf:load-system :nyxt/gtk-application)
* (nyxt:entry-point)
....
```

It still results in same. Could you point me to the Guix issue ?

Thanks!

Ambrevar:
There are you are: https://github.com/atlas-engineer/nyxt/issues/680

wahjava:
Thank you, I'll take a look, and see if I can figure something out.

jmercouris:
ping! :-)

wahjava:
Sorry for the delay in follow up. Unfortunately, none of the advice in the issue #680 worked, although the seems like the same, so nothing new to report here.

Ambrevar:
To recap: the original patch works for you but Nyxt hangs after 10 minutes, is that right?

If so, it's an issue with the FFI and it will be probably very hard to debug.
However there is hope that the bug will go away when we implement the cl-gobject-introspection renderer.  Stay tuned!

wahjava:
> To recap: the original patch works for you but Nyxt hangs after 10 minutes, is that right?

Right, may not be 10 minutes, sometimes under a minute. But without the diff it's a white window. Only the diff gets it to render the Gtk+ widgets.

> 
> If so, it's an issue with the FFI and it will be probably very hard to debug.
> However there is hope that the bug will go away when we implement the cl-gobject-introspection renderer. Stay tuned!

Looking forward to it. If there is anything you like me to test on FreeBSD, please feel free to ask.

Thanks!

Issue #798, opened 2020-06-20
add-domain-to-certificate-whitelist does not work


hsanjuan:
**Describe the bug**

When accessing a domain with a self-signed certificate, I get a TLS error page : `TLS Error: <no_url/name>`. It explains that I should run `add-domain-to-certificate-whitelist` but doing that does nothing. Refreshing the buffer brings me to DuckDuckGo.

**Precise recipe to reproduce the issue**

Nyxt Version: 1.5.0-d0d9c552 
Lisp Implementation: SBCL 
Lisp Version: 2.0.5-115.5-suse 
Operating System: Linux 5.6.14-1-default 
Features: (WEBKIT2 WEBKIT2-2.28 WEBKIT2-EMOJI WEBKIT2-MEDIA WEBKIT2-SANDBOXING
           GTK-3-22 GTK-3-20 GTK-3-18 GTK-3-16 GTK-3-14 GTK-3-12 GTK-3-10
           GTK-3-8 GTK-3-6 GTK-3-4 GTK GDK-3-22 GDK-3-20 GDK-3-18 GDK-3-16
           GDK-3-14 GDK-3-12 GDK-3-10 GDK-3-8 GDK-3-6 GDK-3-4 CAIRO-1-10
           CAIRO-1-12 GDK-PIXBUF GLIB-2-30 GLIB-2-32 GLIB-2-34 GLIB-2-36
           GLIB-2-38 GLIB-2-40 GLIB-2-42 GLIB-2-44 GLIB-2-46 GLIB-2-48
           GLIB-2-50 GLIB-2-52 GLIB-2-54 GLIB-2-56 GLIB-2-58 GLIB
           FSET-EXT-STRINGS SWANK PLUMP-UTF-32 GLOBAL-VARS DECLARE-TYPES
           PARENSCRIPT NAMED-READTABLES OSICAT-FD-STREAMS LPARALLEL 21BIT-CHARS
           CL-FAD CHUNGA FLEXI-STREAMS CLOSER-MOP CL-PPCRE-UNICODE CL-UNICODE
           CL-PPCRE BORDEAUX-THREADS LPARALLEL.WITH-CLTL2 LPARALLEL.WITH-CAS
           LPARALLEL.WITH-STEALING-SCHEDULER SPLIT-SEQUENCE GRAY-STREAMS
           FLAT-NAMESPACE X86-64 UNIX CFFI FLAT-NAMESPACE SEQUENCE-EMPTYP
[...]

PR #806, opened 2020-06-23
Tls error in same buffer

Ambrevar:
This supersedes #710 which purpose actually was to print the TLS error in the same buffer, not to handle `:load-failed`.

This seems to work now, I don't know why it failed in #710.  Maybe a GTK / WebKitGTK update.

@jmercouris Can you test and see if it works for you with https://intra.ec.krugercorp.com/ and when you try to load a page while being offline?


Ambrevar:
This should close #710, #687 , #798.

jmercouris:
Yes, it functions! Yay!

When trying to connect to a page while offline, I now get a message in english: "error resolving X nodename nor servname not provided or known".

PR #807, opened 2020-06-24
Fix relative paths to symlinked quicklisp libs

tviti:
The resulting symlinks were pointing to the wrong source locations, resulting in `system-not-found` errors during build.

Off topic: Would it be worth pointing out in the docs somewhere that, when cloning nyxt, users must now use the `--recursive` flag in order to clone quicklisp-client, and that quicklisp-client is wiped during a `make clean` (hence you must call `git submodule update --init ...` afterterwards to re-clone it)?

### DIFF
--- a/libraries/text-buffer/text-buffer.lisp
+++ b/libraries/text-buffer/text-buffer.lisp
@@ -5,7 +5,7 @@
 (defclass cursor (cluffer-simple-line::right-sticky-cursor)
   ((word-separation-characters
     :accessor word-separation-characters
-    :initform '(":" "/" "-" "." " " " "))))
+    :initform '(":" "/" "." " " " "))))
 
 (defmethod string-representation ((buffer text-buffer))
   (with-output-to-string (out)
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -279,35 +279,19 @@ URL."
 (define-command insert-suggestion-or-tag (&optional (minibuffer (current-minibuffer)))
   "Paste selected suggestion or tag in input.
 If character before cursor is '+' or '-' complete against tag."
-  (cond
-    ;; Complete a tag.
-    ((and (not (str:emptyp (input-buffer minibuffer)))
-          (let ((current-word (word-at-cursor minibuffer)))
-            (or (str:starts-with? "-" current-word)
-                (str:starts-with? "+" current-word))))
-     (with-result (tag (read-from-minibuffer
-                        (make-minibuffer
-                         :input-prompt "Tag"
-                         :input-buffer (subseq (word-at-cursor minibuffer) 1)
-                         :suggestion-function (tag-suggestion-filter))))
-       (when tag
-         (let ((before (subseq (input-buffer minibuffer)
-                               0
-                               (word-start (input-buffer minibuffer)
-                                           (input-cursor-position minibuffer))))
-               (after (subseq (input-buffer minibuffer)
-                              (word-end (input-buffer minibuffer)
-                                        (input-cursor-position minibuffer))))
-               (prefix (elt (word-at-cursor minibuffer) 0)))
-           (log:info (input-buffer minibuffer) before after)
-           (nyxt/minibuffer-mode:kill-whole-line minibuffer)
-           (insert (str:concat
-                    before
-                    (string prefix) (tag-name tag) " "
-                    after)
-                   minibuffer)))))
-    (t
-     (nyxt/minibuffer-mode:insert-suggestion minibuffer))))
+  (let* ((current-word (text-buffer::word-at-cursor (input-cursor minibuffer)))
+         (operand? (unless (str:emptyp current-word) (subseq current-word 0 1))))
+    (if (or (equal "-"  operand?)
+            (equal "+" operand?))
+        (with-result (tag (read-from-minibuffer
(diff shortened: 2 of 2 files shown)


Ambrevar:
Can you explain why it's pointing the the wrong source location?  It works for me here.

Sadly, `realpath` is not portable.  If I recall correctly, it's not on BSD systems by default.

About the off-topic question: No need to use the recursive flag since the submodules are fetched in a Makefile rule.

You are right about the wiping of Quicklisp, we should probably not do it.  Thoughts?

tviti:
Starting from the question about using the `--recursive` flag, here's what I get from a fresh, non-recursive clone:

```
[nix-shell:~/Source/nyxt]$ make all
make application
make[1]: Entering directory '/mnt/data/tviti/Source/nyxt'
true && \
env NYXT_INTERNAL_QUICKLISP=true sbcl --no-userinit --non-interactive \
	--eval '(require "asdf")' \
	--eval '(when (string= (uiop:getenv "NYXT_INTERNAL_QUICKLISP") "true") (load "quicklisp-client/setup.lisp"))' \
	--eval '(ql:quickload :swank)' \
	--eval '(load (merge-pathnames  "contrib/swank-asdf.lisp" swank-loader:*source-directory*))' \
	--eval '(swank:delete-system-fasls "nyxt")' \
	--eval '(uiop:quit)' || true
This is SBCL 2.0.0.nixos, an implementation of ANSI Common Lisp.
More information about SBCL is available at <http://www.sbcl.org/>.

SBCL is free software, provided as is, with absolutely no warranty.
It is mostly in the public domain; some portions are provided under
BSD-style licenses.  See the CREDITS and COPYING files in the
distribution for more information.
Unhandled SB-INT:SIMPLE-FILE-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                                {10005184C3}>:
  Couldn't load "quicklisp-client/setup.lisp": file does not exist.

<further output snipped>
```

So it looks like quicklisp-client isn't being cloned by the `all` target. I'm wondering if the intention was to have a target for `setup.lisp` that clones it in?

> Can you explain why it's pointing the the wrong source location? It works for me here.

In `.gitmodules`, the submodule names reference `quicklisp/local-projects`, but the `path`s are set to `quicklisp-libraries/<project-name>`, so the repo contents are getting cloned into the latter. Hence, when `make` hits the symlinking command in `quicklisp-extra-libs`, the sources don't actually exist, yielding the following error:

```
[nix-shell:~/Source/nyxt]$ make quicklisp-extra-libs
true && git submodule update --init --remote || true
Submodule 'quicklisp/local-projects/cl-webkit' (https://github.com/joachifm/cl-webkit) registered for path 'quicklisp-libraries/cl-webkit'
Submodule 'quicklisp/local-projects/cluffer' (git@github.com:robert-strandh/Cluffer.git) registered for path 'quicklisp-libraries/cluffer'
Cloning into '/mnt/data/tviti/Source/nyxt/quicklisp-libraries/cl-webkit'...
Cloning into '/mnt/data/tviti/Source/nyxt/quicklisp-libraries/cluffer'...
Enter passphrase for key '/home/tviti/.ssh/id_rsa': 
Submodule path 'quicklisp-libraries/cl-webkit': checked out 'c023dad817e8ff031073bf2c76119707e73992a4'
Submodule path 'quicklisp-libraries/cluffer': checked out '4aad29c276a58a593064e79972ee4d77cae0af4a'
for i in quicklisp-libraries/local-projects/*; do ln -sf "$i" "quicklisp-client/local-projects/$(basename "$i")"; done
ln: failed to create symbolic link 'quicklisp-client/local-projects/*': No such file or directory
make: *** [Makefile:113: quicklisp-extra-libs] Error 1
```

Taking a look in `quicklisp-libraries` also reveals the submodules:

```
[nix-shell:~/Source/nyxt]$ ls ./quicklisp-libraries/
cluffer  cl-webkit  README.org
```

> Sadly, realpath is not portable. If I recall correctly, it's not on BSD systems by default.

I originally just added the prefix `../../`  to the `ln` targets to fix this, so that's one option. The other as I see it would be to change the paths in `.gitmodules` to point to the correct locations.

> You are right about the wiping of Quicklisp, we should probably not do it. Thoughts?

I think it would be ok to wipe it, so long as one of the make targets re-clones it (which would also remove the need for the recursive flag). It definitely seems "cleaner" to have the `clean` target kill it off.

Ambrevar:
tviti <notifications@github.com> writes:

> I originally just added the prefix `../../`  to the `ln` targets to fix this, so that's one option. The other as I see it would be to change the paths in `.gitmodules` to point to the correct locations.

I don't know if it's possible to clone submodules inside another submodule.

tviti:
I was actually thinking to just move them to where the `ln` call currently
expects them to be.

For example, in .gitmodules, `path = quicklisp-libraries/cl-webkit` would
become `path = quicklisp-libraries/local-projects/cl-webkit`, then the
Makefile can be left as-is.

Ambrevar:
Should be fixed with 29ee66edfab9ce359aafe626436787b5b2884b97.
Let me know if it works for you.

tviti:
Still having this problem, but I submitted an issue ticket instead of re-opening, since I just realized I submitted this without branching.

PR #812, opened 2020-06-25
Update status view when modifying (buffers *browser*).

Ambrevar:
At the moment, deleting and creating new buffers does not update the status view, which is a bug.

I tried to override (setf gethash) for the browser `buffers` slot but SBCL returned an error.

Defining (setf buffers) is dangerous because the call site might not be aware they need to set `buffers` to the new value (calling remhash modifies the hash-table in-place, so there is no need to set `buffers`).

With Fset, since it's a functional map, we make it explicit that `buffers` must be set to the new value.

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -16,6 +16,7 @@
                :moptilities
                :dexador
                :enchant
+               :fset
                :iolib
                :local-time
                :log4cl
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -610,7 +610,10 @@ useful when no Nyxt window is focused and we still want `ffi-window-active' to
 return something.
 See `current-window' for the user-facing function.")
    (last-active-buffer :accessor last-active-buffer :initform nil)
-   (buffers :accessor buffers :initform (make-hash-table :test #'equal))
+   (buffers :accessor buffers :initform (fset:empty-map)
+            :documentation "Hash table of (ID LIVE-BUFFER).
+We leverage the Fset library to enforce the call to (setf (buffers *browser*) ...)
+so that we can customize the setf method for this slot.")
    (total-buffer-count :initform 0
                        :type integer
                        :documentation "This is used to generate unique buffer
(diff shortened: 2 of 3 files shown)


jmercouris:
I have gone back and forth over this, and in the end introducing fset seems significant for the scope of this problem. I believe this could be handled by removing the accessor to buffers and adding a couple of methods. Thoughts?

Ambrevar:
fset is a very nice library in general.  It advantageously supersedes CL
hash-tables, beside adding sets and a few more functional structures.

Anyways, we already depend on it through the keymap library.

That said, it's true that we don't really need fset here.  It's just one
way to solve the problem.
I need to sleep over it.

Ambrevar:
@jmercouris To answer your suggestion more precisely, using custom settings for BUFFERS won't work.

See https://github.com/atlas-engineer/nyxt/pull/812/files#diff-7ac17e04b23c7965f556e8e3e3fa8ff5L86 for instance, where we set the value of the hash table, not of BUFFERS, and as such the BUFFER setter is never called.

Ambrevar:
Another way to do it is to abstract the hash-table away in a buffer-list class or something.

jmercouris:
That sounds like something I would be in favor of. Do we need a new class though? Could we utilize mutator methods?

Ambrevar:
Not sure what you mean.  Can you provide an example?

jmercouris:
You can see an example in the text-analysis library:

```
(defmethod add-document ((document-collection document-collection) document)
  (push document (documents document-collection)))
```

Of course the user COULD add a document by pushing themselves, but I've provided a method for that to perform any other side-effects/actions that may be necessary at that time. In my case it involves potentially pre-processing some document data.

Ambrevar:
Oh, I understand now, you are saying
- let's remove the BUFFERS accessor
- let's add methods to add and remove buffers.

Yes, this seems to be the simplest.  Let's do that.

Ambrevar:
We already have (buffer-list).  Let's add some methods:
- buffers-set (both to add and modify)
- buffers-delete

Not quite happy with the naming though.

Ambrevar:
Done in d62cbfd7f07fd27b707fb712d126dcf7033d997b.
The functions are not exported for now.  We can choose better names in due time.

Issue #695, opened 2020-04-24
Decode percent encoded characters in URIs


arunisaac:
URIs with percent encoded Unicode characters should be decoded before being displayed to the user on the mode line, history, etc. Percent encoded characters are not human readable and this makes reading URIs in non-English languages painful.

Thank you!

Ambrevar:
Will do!

Can you give me a URL to test with?

Ambrevar:
When pressing C-w in the minibuffer (copy-url), should we copy the
decoded URL or the raw URL?

At the moment we use the OBJECT-STRING method both to tab-complete and
copy.

For URLs, tab-completion must insert the decoded URLs otherwise it would
not match the candidates which should also be decoded.

arunisaac:
Here are a couple of examples, one Tamil and the other French.

https://ta.wikipedia.org/wiki/%E0%AE%B5%E0%AE%BF%E0%AE%95%E0%AF%8D%E0%AE%95%E0%AE%BF%E0%AE%AA%E0%AF%8D%E0%AE%AA%E0%AF%80%E0%AE%9F%E0%AE%BF%E0%AE%AF%E0%AE%BE

https://fr.wikipedia.org/wiki/Wikip%C3%A9dia

> When pressing C-w in the minibuffer (copy-url), should we copy the
> decoded URL or the raw URL?

I think copy-url should copy the raw URL (the one with the percent
encoded characters). That way, the copied URL will still be valid when
pasted into other programs. This is also what Firefox does. I just
checked.

Ambrevar:
Thanks!

Does this mean we need 3 string representation methods?

- object-display to display the candidates,
- object-string for tab completion,
- object-copy for copy?

Or can we be smarter?

Maybe tab-completion could insert object-display until the first space?
I'll try that.

Ambrevar:
Fixed in 8f394a5e7655b9d85823e96101627b13b2dfb7a2.

> Maybe tab-completion could insert object-display until the first space?
> I'll try that.

This does not work because the URL is not always on first position
(e.g. switch-buffer).  Instead I decided to insert the object-string and
if it is a URL then we decode it.

Ambrevar:
Feel free to reopen if you can spot URLs that should be decoded.

arunisaac:
Thanks, works nicely! :-)

> Feel free to reopen if you can spot URLs that should be decoded.

Sure!

arunisaac:
[...]

Issue #803, opened 2020-06-22
Search-engine tab-completion stopped working


Ambrevar:
Recipe:
- Have 2 search engines that start with the same letter, e.g. "wiki" and "wifoo".
- Type `w<TAB>`.

Expected result: A list of search engines suggestions including "wiki" and "wifoo".

Actual result: Nothing happens.  When run in a buffer, I get a condition:

```lisp
The variable NYXT::MATCH-COUNT is unbound.
   [Condition of type UNBOUND-VARIABLE]

Restarts:
 0: [CONTINUE] Retry using NYXT::MATCH-COUNT.
 1: [USE-VALUE] Use specified value.
 2: [STORE-VALUE] Set specified value and use it.
 3: [RETURN-FROM-G-CLOSURE] Return value from closure
 4: [ABORT] abort thread (#<THREAD "cl-cffi-gtk main thread" RUNNING {100BAEEE33}>)

Backtrace:
 0: (NYXT:INSERT-SUGGESTION-OR-SEARCH-ENGINE #<NYXT:MINIBUFFER {100B3DFDD3}>)
 1: (NYXT:DISPATCH-INPUT-EVENT #S(GDK:GDK-EVENT-KEY :TYPE :KEY-PRESS :WINDOW #<GDK:GDK-WINDOW {1005949693}> :SEND-EVENT NIL :TIME 139450083 :STATE (:MOD2-MASK) :KEYVAL 65289 ...) #<NYXT-USER::USER-BUFFER ..
 2: (GOBJECT::CALL-WITH-RESTARTS #<CLOSURE (LAMBDA (NYXT::WIDGET NYXT::EVENT) :IN INITIALIZE-INSTANCE) {100DF6259B}> (#<GTK:GTK-WINDOW {1005D7B9E3}> #S(GDK:GDK-EVENT-KEY :TYPE :KEY-PRESS :WINDOW #<GDK:GDK..
 3: ((LAMBDA (GOBJECT::CLOSURE GOBJECT::RETURN-VALUE GOBJECT::COUNT-OF-ARGS GOBJECT::ARGS GOBJECT::INVOCATION-HINT GOBJECT::MARSHAL-DATA) :IN "/gnu/store/pglf1avn8dan3zdlijxv3pcljjwm7zxa-cl-cffi-gtk-0.11...
[...]

PR #814, opened 2020-06-25
Change URL representation to QURI:URI

### LINKED ISSUE #794: Systematically decode URLs
#695 bumped up again.
Our current fix is too ad-hoc.

I suggest we change how we store URLs: instead of strings, let's store them as `quri:uri` objects.  

Then we can specialize object-display on `quri:uri` and all URLs will be displayed properly in the minibuffer once and for all.

As a bonus, we would not have to call (quri:uri ...) all over the place before manipulating the URL.

Thoughts?

aartaka:
This changes the representation of URLs wherever necessary to `quri:uri`, more stable, standardized and convenient representation.

## Motivation

There was #695, with the issues of URL decoding that motivated this change. But, outside of this decoding issue, `quri:uri` representation is better in that it:
- Is standardized (complies on the RFCs for the URLs -- implementing a home-grown version of it could have been a colossal labor)
- Is integrateable into CLOS and `s-serialization` (with the help of `object-print` method definition on `quri:uri` and `quri:uri` being a mere structure)
- Has a good API (e.g. `quri:uri-host`, `quri:uri-scheme`, `quri:uri-authority`, `quri:uri-http-p`, etc.) that make both maintainers' and users' URL-handling easier.

Given all that, even partially migrating to `quri:uri` can be useful and can ease both maintainers' and users' lives :)

## Caveats

This change is not exhaustively tested, and some minor bugs may emerge after merging. But, given the narrow domain of change, these bugs will be relatively easy to catch and fix -- they all will be connected to type-conflicts between `quri:uri` and `string`/`vector` types. To provide a possible debugging help, I've added some `ftype` declaration for the functions I considered as these bug-bottlenecks.  

Anyway, additional testing is welcome before merging!

## What's Gone
- `host`, `domain`, `url-sans-scheme`, and `url-display` functions are no longer there.

## What's Changed
- Small utility function `url-empty-p` and local utility function bindings were added, especially to bookmark-related code where string-representation of URLs was used for bookmark sorting before.
- `requested-uri` and `resolved-uri` slots of `download` class have `quri:uri` class.
- `url` slot of `hostlist` class is `quri:uri` now, for the sake of simplifying the filtering.
- `url` slot of `bookmark-entry` class is `quri:uri` now.
- New hook type, `hook-uri->uri` is added, mainly to make `set-url-hook` work with `quri:uri` structures.
- `url`, `url-at-point`, `default-new-duffer-url` and slots of `buffer` class are `quri:uri` now.
- `set-url-hook` slot of `buffer` class is of `hook-uri->uri` type now.
- `start-page-url` slot of `browser` class uses `quri:uri`.
- `url` slot of `history-entry` class is `quri:uri` now.
- `parse-url` returns `quri:uri` now. 
- `quri:uri-domain` and `quri:quri-host` are used in place of home-grown `host` and `domain` functions.
- `uri` and `original-uri` fields of the `webkit-history-entry` structure are `quri:uri` now too.
- URLs taken from renderer get wrapped in `quri:uri` constructor for convenience.
- Manual is adjusted to the changes made (although it may need proofreading)

## What's Retained and Why
- `hosts` slot of the `hostlist` class was left as it is because it's a parsing-intensive piece of the browser and having dozens of `quri:uri` objects would neither make it faster nor conceptually simpler.
- `shortcut` and `search-url` fields of `bookmark` class are left in their initial form of `string`s, because former is intended to be matched from the input, and the latter is intended to be used as the `format` control-string.
- `search-engine` class is using `string`s exclusively, due to its substitute/format/search nature.
- `proxy` class remains with `string`-encoded URLs (`server-address` slot), although I'm open to discussion about that. My idea is that it's both easier to set by the user as a string, and easier to pass to lower-level APIs (be it OS-level or some networking libraries) as a string.
- `buffer-description` class is left with the `string` type of its `url` slot because it is used for serialization only and has no necessity to use `quri:uri`.
- `open-urls`, `set-url*` and `parse-url` functions still accept strings as arguments, but `parse-url` returns `quri:uri` structure now, and `set-url*` uses this representation and passes it to renderer. That's because someone still needs to parse string-representation of URL and return a nice n' shiny `quri:uri` for all the other Nyxt parts to use.

## How Has This Been Tested
- I've started the GTK version from REPL and spent two hours browsing the web and calling random commands. This revealed some problems in places I haven't been attentive enough to. However, last half an hour was bug-less and I was unable to cause any errors to emerge.
- I've installed the GTK version from the `build-scripts/guix.scm` package definition. Seems to work the same relatively flawless way, but it certainly lacked testing, because I've spent less than an hour with it.
- I was unable to run the Qt version due to CFFI error about "NewQApplication" having improper arguments, so no testing happened on the Qt version.

Closes #794


source/bookmark.lisp
```diff
@@ -54,39 +54,35 @@ SEARCH-URL maybe either be a full URL or a path.  If the latter, the path is
...
               (format nil " (~{~a~^, ~})" (tags entry))
               "")))
 
-(defun url-sans-scheme (url)
-  (apply #'str:concat
-         (mapcar (alex:curry #'format nil "~a")
-                 (delete nil
-                         (rest (multiple-value-list (quri:parse-uri url)))))))
-
+(declaim (ftype (function (quri:uri quri:uri) t) equal-url))
```
Ambrevar, line 72: The return type should be `boolean` here I believe.

source/bookmark.lisp
```diff
@@ -54,39 +54,35 @@ SEARCH-URL maybe either be a full URL or a path.  If the latter, the path is
...
 - http://example.org
 - https://example.org/"
-  (flet ((normalize-path (uri)
-           (if (nth-value 4 (quri:parse-uri uri))
-               uri
-               (str:concat uri "/"))))
-    (string= (url-sans-scheme (normalize-path url1))
-             (url-sans-scheme (normalize-path url2)))))
+  (flet ((url-sans-scheme (url)
+           (quri:copy-uri url :scheme nil :port nil)))
```
Ambrevar, line 80: Why `:port nil`?

source/bookmark.lisp
```diff
@@ -373,8 +376,11 @@ If character before cursor is '+' or '-' complete against tag."
     (setf (slot-value *browser* 'bookmarks-data)
           (sort (slot-value *browser* 'bookmarks-data)
                 (lambda (e1 e2)
-                  (string< (url-sans-scheme (url e1))
-                           (url-sans-scheme (url e2))))))
+                  (flet ((url-sans-scheme (url)
+                           (object-display
+                            (quri:copy-uri url :scheme nil :port nil))))
```
Ambrevar, line 381: If you are using this twice, it's worth extracting to a top-level function.

source/browser.lisp
```diff
@@ -19,6 +19,8 @@
                                                    &optional (or (eql :stop)
                                                                  (eql :forward)))))
 (export-always '(make-hook-resource make-handler-resource))
+;; TODO: Should we export this? It's used in unexported SET-URL-HOOK only.
```
Ambrevar, line 22: Then no, because it's unclear whether set-url-hook will ever be needed.  request-resource-hook largely supersedes set-url-hook.

source/browser.lisp
```diff
@@ -909,7 +912,7 @@ when `proxied-downloads-p' is true."
 (defun download (url &key
                      cookies
                      (proxy-address :auto))
-  "Download URI.
+  "Download URI (it should be of QURI:URI type).
```
Ambrevar, line 915: The original docstring had a typo, it's `URL` here.
Instead of commenting about the type, I suggest you declaim the function type instead.  It serves as documentation while enforcing some compiler type checking.

source/buffer.lisp
```diff
@@ -242,9 +243,9 @@ complete against a search engine."
                                  :suggestion-function #'search-engine-suggestion-filter)))
             (when engine
               (nyxt/minibuffer-mode:kill-whole-line minibuffer)
-              (insert minibuffer (str:concat (shortcut engine) " "))))))))
+              (insert minibuffer (str:concat (shortcut engine) " ")))))))
     (t
-     (nyxt/minibuffer-mode:insert-suggestion minibuffer)))
+     (nyxt/minibuffer-mode:insert-suggestion minibuffer))))
```
Ambrevar, line 248: Good catch! ;)

libraries/download-manager/engine.lisp
```diff
@@ -44,12 +44,12 @@ beforehand in order to specify different initialization argument."
 
 (defclass download ()
   ((requested-uri :accessor requested-uri :initarg :requested-uri
-                  :initform ""
+                  :initform (quri:uri "") :type quri:uri
```
jmercouris, line 47: Do we need an initform here? Will we ever instantiate a DOWNLOAD object without a requested-uri?

libraries/download-manager/engine.lisp
```diff
@@ -44,12 +44,12 @@ beforehand in order to specify different initialization argument."
...
                   :documentation "The URI that the user requested.  This may be
 different from the actual location of the download, e.g. in case of automatic
 redirection.  See RESOLVED-URI.")
    (resolved-uri :accessor resolved-uri :initarg :resolved-uri
-                 :initform ""
+                 :initform (quri:uri "") :type quri:uri
```
jmercouris, line 52: Same as above: 

> Do we need an initform here? Will we ever instantiate a DOWNLOAD object without a requested-uri?

source/history.lisp
```diff
@@ -35,23 +35,25 @@ Entry for the global history.
...
 (defun history-add (url &key title explicit)
   "Add URL to the global history.
 The `implicit-visits' count is incremented unless EXPLICIT is non-nil, in which
 case `explicit-visits'.
 The history is sorted by last access."
-  (unless (str:emptyp url)
+  (unless (url-empty-p url)
     (unless (history-data *browser*)
       (setf (slot-value *browser* 'history-data)
+            ;; TODO: Use equal-url from bookmark.lisp?
```
Ambrevar, line 56: I'd say not because for history the differentiation between HTTPS and HTTP matters.

source/parse-urls.lisp
```diff
@@ -1,7 +1,8 @@
 (in-package :nyxt)
 
+(declaim (ftype (function (string) (values (or quri:uri null) t &optional)) parse-url))
 (defun parse-url (input-url)
-  "From user input, return the full url to visit.
+  "From user input, return the full url (as a QURI:URI) to visit.
```
Ambrevar, line 5: No need to specify the type since it's already in the function type declaration.
While you are at it, you can upcase `url` to `URL` since they are initials.

libraries/download-manager/native.lisp
```diff
@@ -16,23 +16,24 @@
...
-            (dex:get requested-uri :want-stream t :force-binary t :keep-alive nil
-                                   :proxy proxy
-                                   :cookie-jar cookies-jar)
+            (dex:get (quri:render-uri requested-uri)
+                     :want-stream t :force-binary t :keep-alive nil
+                     :proxy proxy :cookie-jar cookies-jar)
           ;; TODO: Allow caller to set the target filename?
-          (let* ((file (merge-pathnames directory (extract-filename requested-uri
-                                                                    response-headers))))
+          (let* ((file (merge-pathnames
```
jmercouris, line 30: Formatting changes, without changing the logic don't really need to be in a pull request, you can typically just commit them/propose them separately. Here is OK too though :-)

source/blocker-mode.lisp
```diff
@@ -12,8 +12,8 @@
 
 (defclass-export hostlist ()
   ((url :accessor url :initarg :url
-        :initform ""
-        :type string
+        :initform (quri:uri "")
```
jmercouris, line 15: Same as above, do we need an initform?

Ambrevar:
Excellent, thank you so much!  And thanks for the great explanatory abstract ;)

I see that you've merged from master.  Could you rebase against it instead?

source/bookmark.lisp
```diff
@@ -15,8 +15,8 @@
 (defclass-export bookmark-entry ()
   ((url :initarg :url
         :accessor url
-        :type string
-        :initform "")
+        :type quri:uri
+        :initform (quri:uri ""))
```
jmercouris, line 19: Same as before :-)

It may seem to be not very important, but by declaring an initform, you are implicitly saying that we rely on the existence of some object to check and/or mutate. You are also implicitly saying that we may make a bookmark-entry object without specifying a URL.

source/bookmark.lisp
```diff
@@ -281,16 +281,16 @@ URL."
...
         (with-result (tag (read-from-minibuffer
                            (make-minibuffer
-                            :input-prompt "Tag"
+                            :input-prompt "Tags"
                             :suggestion-function (tag-suggestion-filter))))
           (when tag
-            (text-buffer::replace-word-at-cursor 
+            (text-buffer::replace-word-at-cursor
              (input-cursor minibuffer)
-             (concatenate 'string operand? (tag-name tag)))))
```
jmercouris, line 293: Why this change? What benefits does using the STR library here confer?

source/browser.lisp
jmercouris, line 915: Instead of a comment, would be possible to do a type declaration?

Ambrevar:
About the proxy: Yes, I think we should use quri:uri there too.  It's easier to set and manipulate the different parts, like the ports, etc.

Ambrevar:
In terms of vocabulary, I suggest the following convention: let's use URI for value of the `quri:uri` type, and URL for values of the `string` type.  Thoughts?

source/element-hint.lisp
```diff
@@ -322,7 +322,7 @@ identifier for every hinted element."
 (define-command follow-hint (&key annotate-full-document)
   "Show a set of element hints, and go to the user inputted one in the
 currently active buffer."
-  (query-hints "Go to element:" '%follow-hint
+  (query-hints "Go to element:" #'%follow-hint
```
jmercouris, line 325: Why not leave as quoted form so that we can recompile in image?

source/history.lisp
jmercouris, line 56: Yes :-)

Ambrevar:
About testing: as you've noticed our test suite is crucially lacking.  You can invoke (asdf:test-system :nyxt) and see if the few tests we have there still pass.

When the minibuffer get's overhauled to a separate library, we will be able to expand on the tests, in particular when it comes to URL handling.

It's normal that you couldn't test the QtWebengine renderer, it hasn't been updated in a few month :p  By the way, if you are interested in updating it, help is welcome! :)  Expect a medium challenge ;)

source/history.lisp
jmercouris: Then equal-url should accept an optional to consider the protocol, no?

Ambrevar:
Another good test is to compile with CCL.  It will complain about a few forward declarations which need to be fixed.  Other warnings might indicate by typing.  Indeed, CCL is very good at type checking.

source/browser.lisp
jmercouris: Copycat :-P that's exactly what I wrote!

libraries/download-manager/engine.lisp
Ambrevar: In doubt, let's leave it, otherwise if it ever is instantiated without a requested-uri it would hang the browser.

jmercouris:
> In terms of vocabulary, I suggest the following convention: let's use URI for value of the `quri:uri` type, and URL for values of the `string` type. Thoughts?

Yes, this is a good idea!

jmercouris:
It seems both Pierre and I decided to review this at the same time :-D. We've said many of the same things... :-D. From what I can see, there is no conflict in advice. Enjoy :-)

source/bookmark.lisp
Ambrevar: The caller may instantiate the object before having access to the URI for some reason.

As @jmercouris once told me: In the vast majority of cases, either we specify the type zero value for the initform or we put `(error "Must provide value for this slot")` to force the user to provide something at instantiation time.

It's very rare that we would make use of empty slots and usually it makes for inconvenient code because all callers must check if the slot is bound, etc.

source/bookmark.lisp
jmercouris: I believe ANYTHING can be boolean as everything inherits from `t`, I could very well be wrong though.

source/bookmark.lisp
Ambrevar: It's shorter and consistent with the rest of the code base.  By the way, this is my change, @aartaka merged instead of rebasing.

source/element-hint.lisp
Ambrevar: I don't understand what you mean.
Using the sharp-quote here allows for better type checking.  We should type the `query-hints` function for that.

source/history.lisp
Ambrevar: As I said in my review, I'd opt for no here, because it's important to tell the difference between HTTP and HTTPS in the history.  

Otherwise we could change the logic: If a page is available both in HTTP and HTTPS, keep the HTTPS one.

source/bookmark.lisp
Ambrevar: `t` is correct but it's not very useful for type-checking.  `boolean` is more specific and more helpful to the human reader.

source/element-hint.lisp
jmercouris: If you want to test `%follow-hint` after recompiling it by running `follow-hint`, you will not be able to.

source/history.lisp
Ambrevar: But then it would be the same as `quri:uri=`, no?

Nit: Don't forget it's "scheme", not "protocol" ;)

jmercouris: I think equal-url should accept a flag to consider the protocol, then, problem solved!

libraries/download-manager/engine.lisp
jmercouris: I would consider that a bug though, and we are basically just hiding the bug by instantiating it :-D

source/bookmark.lisp
jmercouris: Well, fair enough :-)

source/bookmark.lisp
jmercouris: Oh jeez, my own advice coming back to haunt me :-D

source/history.lisp
Ambrevar: Then why not use `quri:uri=` directly?

source/element-hint.lisp
Ambrevar: In general, to test changes in functions it's better to recompile the package.

libraries/download-manager/engine.lisp
Ambrevar: As we discussed below, not using `:initform` does not trigger an error immediately and indeed, a bug may be lurking for a while before we notice it.  Better to use the `(error ...)` initform to raise the condition at instantiation time.

source/bookmark.lisp
aartaka: `port` slot is implicitly set for any constructed `quri:uri` structure, based on its `scheme` slot. If we will `quri:copy-uri` any `quri:uri`, then the port will stay in this inferred-inherited state of either 80 (HTTP) or 443 (HTTPS) and it may influence the comparison. To avoid this, I'm explicitly setting `port` to `nil`.

source/element-hint.lisp
aartaka: Oh, yes, I thought that the absence of the sharp-quote was a typo, based on the functions after this one. I've looked at the code of `query-hints` and thought that it should have been a `function` object that gets passed to it. That's why I set the sharp-quote here. I guess, I misunderstood the code...

source/bookmark.lisp
Ambrevar: Good point, but this won't give the right result if the URLs indeed have different ports.

Internally, quri uses quri:uri-authority to print the hostname+port in a smart way: if it's the default port, then don't print it, otherwise do.

I suggest you compare the URLs authority and the rest of the elements (path, etc.).

source/element-hint.lisp
Ambrevar: You didn't misunderstand.  The function eventually gets called with `apply` which supports both functions and function symbols.

source/bookmark.lisp
aartaka: You're right. I just wasn't sure whether it's really boolean and not a *generalized boolean* that this function returns. Now, after reparsing it and reading through the standard, I'm sure it's boolean that `quri:uri=` returns!

libraries/download-manager/engine.lisp
aartaka: Having `(error ...)` forms in the initialization sounds like a good practice to me, but for now I'll refrain from adding these, because that's a whole new layer of analysis and testing. Let's get `quri:uri` transition done and then we can work on better initforms.

source/browser.lisp
aartaka: Side question: was there any hypothetical application for `set-url-hook`? 

I can vaguely think of changing things on the just-after-parsing phase, but all this can be done on `request-resource` phase either, because it still happens early enough.

libraries/download-manager/engine.lisp
Ambrevar: No problem!

source/browser.lisp
Ambrevar: Indeed, `set-url-hook` was there before we had `request-resource`.  Now it's mostly deprecated.  We are keeping it around in case we find out a use case for it.

aartaka:
Sorry for force-pushing and awkward merging -- my lack of Git knowledge have manifested itself in terrible corruption that have stopped me from doing anything except for force-pushing. Lesson learned :)

The changes I've made are:
- Add several `ftype` declarations to the functions that had the `quri:uri` mentioned, and remove this mention.
- Change `server-address` slot of `proxy` class and `proxy-uri` slot of `gtk-buffer` class to `quri:uri`
- Other small changes based on the review.

CCL haven't complained about anything during compilation. `asdf:test-system` throws error messages about the `match-count` variable being unbound in `insert-suggestion-or-search-engine` (this may be another reason for #803) and nothing else. I guess all the bugs are either covered or are too deep to be easily found :)

aartaka:
> It's normal that you couldn't test the QtWebengine renderer, it hasn't been updated in a few month :p By the way, if you are interested in updating it, help is welcome! :) Expect a medium challenge ;)

Sounds interesting, but, I guess, I'll stick to the ephemeral natures of WebKitGTK-land for now, while implementing incognito-mode. But this Qt port binding-work is the experience that I'll be glad to learn from when everything else will be done!

source/bookmark.lisp
aartaka: I can't imagine the case of purposefully bookmarking two similar URLs with only ports differing. Can you, please, provide one? It seems that I had little to no experience of bookmarking anything except HTPP(s) links :)

Ambrevar:
About match-count: Did you test against your changes?  Because it seemed to me that you did fix insert-suggestion-or-search-engine indeed.

About Git: I don't think you made any mistake.  When you rebase, you rewrite the history so you can't just push to the same branch on your repo without force-pushing.
Since any history rewrite is irreversible, it's good practice to back up your branch before rebasing it (by simply creating a new branch at the tip of the branch you are going to rebase).

Ambrevar:
SBCL reports a few warnings, could you fix them?

For the type declaration that's too complex to check, you can use this trick:
`(the (values YOUR-RETURN_TYPE &optional) RETURN-VALUE))`

Ambrevar:
(To wrap around your return value.)

aartaka:
It seems that the STYLE-WARNINGS that SBCL emits are all about the generic functions whose `ftype` (somewhy) gets `declaim`-ed instead of their methods' `ftype`. It seems I either don't know a way to specify the `ftype` for a specific method, or it's just impossible to specify `ftype` of methods for the generic function with lots of methods. I doubt the latter, but HyperSpec says nothing about methods' `ftype`, so I just don't know where else I can find the answer.

aartaka:
> About match-count: Did you test against your changes? Because it seemed to me that you did fix insert-suggestion-or-search-engine indeed.

Hmmmm. It seems working now, and no errors appear... I guess it's really fixed then.

aartaka:
Ok, this time #803 is finally resolved. There was yet another misplaced parentheses and one of the `match` clauses was outside `match`. This was the reason for `match-count` being unbound. Weird concentration of parentheses shift though...

About STYLE-WARNINGS: I've removed theconflicting `declaim`s, because it seemed that they influenced the generic function itself and not the separate methods. Given that these definitions were not crucial and problems of improper type of the argument would be relatively easy to observe without them (`string`s and `quri:uri`s don't have a lot of methods in common). 

However, if you will hint me on a way around these issues, I'll be glad to know and implement it -- `declaim`s feel important to me!

Ambrevar:
Indeed, I don't know how to declaim method types either.  Might be something worth asking on Reddit.

This issue could be non-standard and SBCL-specific.

Alternatively, we can declare the types or check-type inside the function body.  Type errors won't be caught at compile-time though.

Ambrevar:
I think you misunderstood the warning I was referring to about the type assertion.
I'll paste it here:

```

/home/ambrevar/common-lisp/nyxt/source/bookmark.lisp:76:1:
  note: 
    Type assertion too complex to check:
    (VALUES &OPTIONAL BOOLEAN &REST T).
    It allows an unknown number of values, consider using
    (VALUES &OPTIONAL BOOLEAN).
    --> PROGN SB-IMPL::%DEFUN SB-IMPL::%DEFUN SB-INT:NAMED-LAMBDA 
    ==>
      #'(SB-INT:NAMED-LAMBDA NYXT::EQUAL-URL
            (NYXT::URL1 NYXT::URL2)
          (DECLARE (SB-C::TOP-LEVEL-FORM))
          "URLs are equal if the URIs are equal, scheme excluded.
    Empty paths are also excluded from the comparison.
    For instance, these are equal:
    - http://example.org
    - https://example.org/"
          (BLOCK NYXT::EQUAL-URL
            (IF (AND (QURI.URI.HTTP:URI-HTTP-P NYXT::URL1)
                     (QURI.URI.HTTP:URI-HTTP-P NYXT::URL2))
                (QURI:URI= (NYXT::URL-SANS-SCHEME NYXT::URL1)
                           (NYXT::URL-SANS-SCHEME NYXT::URL2))
                (QURI:URI= NYXT::URL1 NYXT::URL2))))
    

/home/ambrevar/common-lisp/nyxt/source/bookmark.lisp:206:1:
  note: 
    Type assertion too complex to check:
    (VALUES STRING &REST T).
    It allows an unknown number of values, consider using
    (VALUES STRING &OPTIONAL).
    --> PROGN SB-IMPL::%DEFUN SB-IMPL::%DEFUN SB-INT:NAMED-LAMBDA 
    ==>
      #'(SB-INT:NAMED-LAMBDA NYXT:URL-BOOKMARK-TAGS
            (NYXT:URL)
          (DECLARE (SB-C::TOP-LEVEL-FORM))
          "Return the space-separated string of tags of the bookmark corresponding to
    URL."
          (BLOCK NYXT:URL-BOOKMARK-TAGS
            (LET ((NYXT::EXISTING-BM #))
              (IF NYXT::EXISTING-BM
                  (STR:JOIN " " #)
                  ""))))
    

/home/ambrevar/common-lisp/nyxt/source/bookmark.lisp:381:28:
  style-warning: undefined function: NYXT::OBJECT-DSIPLAY

/home/ambrevar/common-lisp/nyxt/source/browser.lisp:917:1:
  note: 
    Assignment to argument: PROXY-ADDRESS
      prevents use of assertion from function type proclamation:
     (OR QURI.URI:URI KEYWORD)
    
    --> PROGN SB-IMPL::%DEFUN SB-IMPL::%DEFUN SB-INT:NAMED-LAMBDA 
    ==>
      #'(SB-INT:NAMED-LAMBDA NYXT:DOWNLOAD
            (NYXT:URL &KEY NYXT::COOKIES (NYXT::PROXY-ADDRESS :AUTO))
          (DECLARE (SB-C::TOP-LEVEL-FORM))
          "Download URL.
    When PROXY-ADDRESS is :AUTO (the default), the proxy address is guessed from the
    current buffer."
          (BLOCK NYXT:DOWNLOAD
            (HOOKS:RUN-HOOK (NYXT:BEFORE-DOWNLOAD-HOOK NYXT:*BROWSER*) NYXT:URL)
            (WHEN (EQ NYXT::PROXY-ADDRESS :AUTO) (SETF NYXT::PROXY-ADDRESS #))
            (LET ((NYXT::DOWNLOAD-DIR #))
              (WHEN NYXT::DOWNLOAD-DIR
                (LET* #
                  #)))))
    

Compilation successful.

```

You should see this too.  How do you compile Nyxt?  From SLY, if you run `,load-system RET nyxt` it should prompt a `*sly-compilation*` buffer with all the warning.

To fix the warnings, use the `(the ...)` trick I suggested above.

aartaka:
Oh, I haven't even seen these warnings and, maybe, misunderstood what you meant. I've compiled Nyxt using `make` and there were warnings about the types of methods/generics. And I haven't looked at `sly-compilation` buffer either, but now it shows none of the warnings you mention... Thanks for pasting the warnings, now I understand where the problems are! 

Applying this trick with `the` (thanks for that too!), I can now work on the first two warnings, but what to do with third? I'm trying to `declare` it's type after the assignment that SBCL complains about, and it even seems to work, but is it enough?

aartaka:
I guess it is, because, if anything non-`quri:uri`-like wil be there after assignment, it will clearly mean something's not right. I'll push the changes in a minute!

aartaka:
And what about this issue of port absence in `url-sans-scheme`? The concern about bookmarking the address with a non-standart port sounds valid to me, but I just cannot imagine a situation of bookmarking two absolutely similar HTTP(S) addresses (because scheme and port get thrown away only in the `quri:uri`s that satisfy `quri:uri-http-p`) with only the ports differing.

Ambrevar:
The `make` command does not report many warnings.  Use the REPL to see them.

The 3rd warning is a typo: it's "object-dISplay", not "object-dSIplay" :p

About bookmarks: We should bookmark that URL that's given by the user _as is_.  `url-sans-scheme` is only used for comparison.
As mentioned above, I believe we should use quri:uri-authority in `url-sans-scheme` so that default ports are ignored, but non-default ports are kept.

aartaka:
> The 3rd warning is a typo: it's "object-dISplay", not "object-dSIplay" :p

Oh, yes. Awkward typo...
 
> As mentioned above, I believe we should use quri:uri-authority in `url-sans-scheme` so that default ports are ignored, but non-default ports are kept.

Then, I guess, there's no other way around this except for re-implementing `quri:uri=` without scheme comparison...

aartaka:
Okay, typo is fixed and URLs are now compared in a truly schemeless way :sparkles:

aartaka:
And, unfortunatley, `url-sans-scheme` is no longer there -- one of the use-cases is covered by the new `schemeless-uri=`, and the other was pretty much covered by `quri:copy-uri` with a `nil`-ed `:scheme`.

Ambrevar:
It fails to compile with

```
/home/ambrevar/common-lisp/nyxt/source/renderer-gtk.lisp:712:1:
  warning: 
    Constant NIL conflicts with its asserted type QURI.URI:URI.
```

Why is it unfortunate that "`url-sans-scheme` is no longer
there"?
Do we miss url-sans-scheme anywhere?

source/bookmark.lisp
```diff
@@ -54,39 +54,42 @@ SEARCH-URL maybe either be a full URL or a path.  If the latter, the path is
...
-                 (delete nil
-                         (rest (multiple-value-list (quri:parse-uri url)))))))
 
+(declaim (ftype (function (quri:uri quri:uri) boolean) schemeless-uri=))
+(defun schemeless-uri= (uri1 uri2)
+  "A copy of QURI:URI=, except for the scheme comparison and FTP-specific code"
+ (and (equal  (or (quri:uri-path uri1) "/") (or (quri:uri-path uri2) "/"))
+      (equal  (quri:uri-query uri1)     (quri:uri-query uri2))
+      (equal  (quri:uri-fragment uri1)  (quri:uri-fragment uri2))
+      (equalp (quri:uri-authority uri1) (quri:uri-authority uri2))))
```
Ambrevar, line 79: I suppose that you are using `equalp` here because case does not matter in the URI's authority?
In anycase, I suggest you make this explicit in the docstring (or as a comment if this is an implementation detail).

The docstring is a bit confusing.  Besides, it must end with a ".".
I suggest this:

    Like `quri:uri=' but ignore scheme in comparison.

What do you mean with FTP-specific code?

source/browser.lisp
```diff
@@ -368,12 +369,12 @@ renderers might support this.")
...
-                 :initform (hooks:make-hook-string->string
+                 :initform (make-hook-uri->uri
                             :combination #'hooks:combine-composed-hook)
-                 :type hooks:hook-string->string
+                 :type hook-uri->uri
                  :documentation "Hook run in `set-url*' after `parse-url' was processed.
-The handlers take the URL going to be loaded as argument and must return a
-(possibly new) URL.")
+The handlers take the URL going to be loaded as argument (as QURI:URI)
+and must return a (possibly new) URL (of type QURI:URI).")
```
Ambrevar, line 377: Nit: capitals are use to refer to lambda list arguments.
If you refer to a declared symbol, use the "`foo'" syntax.  Emacs should highlight the symbol.

source/bookmark.lisp
```diff
@@ -54,39 +54,42 @@ SEARCH-URL maybe either be a full URL or a path.  If the latter, the path is
...
-           (if (nth-value 4 (quri:parse-uri uri))
-               uri
-               (str:concat uri "/"))))
-    (string= (url-sans-scheme (normalize-path url1))
-             (url-sans-scheme (normalize-path url2)))))
+  (the (values boolean &optional)
+       (if (and (quri:uri-http-p url1)
+                (quri:uri-http-p url2))
+           (schemeless-uri= url1 url2)
+           (quri:uri= url1 url2))))
```
Ambrevar, line 92: If schemeless-uri= already returns a boolean, then you only need to use the `(the ...)` trick around `(quri:uri= ...)`.

source/bookmark.lisp
aartaka: > I suppose that you are using `equalp` here because case does not matter in the URI's authority?
> In anycase, I suggest you make this explicit in the docstring (or as a comment if this is an implementation detail).

It was used in the original `quri:uri=`, and the reason behind that is the case insensitivity of the authority and the case-sensitivity of all the other parts. It's regulated by RFC 3986 (section 3.2.2):
> The host subcomponent is case-insensitive.

I'll ad the reference to RFC 3986 in the docstring.

> The docstring is a bit confusing. Besides, it must end with a ".".
> I suggest this:
> 
> ```
> Like `quri:uri=' but ignore scheme in comparison.
> ```

Yes, sounds good to me.

> What do you mean with FTP-specific code?

Well, the code of the `schemeless-uri=` was just a slightly altered code of `quri:uri=`. One of the cases that `quri:uri=` specifies and treats differently was the case of FTP URIs. I thought that schemelessness and absence of FTP-related comparison code are differences that should be underlined. But, I guess, it's not that important for the current use of `schemeless-uri=`.

source/browser.lisp
aartaka: Oh, that's interesting. Somewhy these unnecessary docstring type requirements got back. I'll remove them.

I didn't know about this sintax though, thank you for pointing that out!

aartaka:
> It fails to compile with ``` /home/ambrevar/common-lisp/nyxt/source/renderer-gtk.lisp:712:1: warning: Constant NIL conflicts with its asserted type QURI.URI:URI. ``` 

Oh, sorry for letting it through. This type declaration revealed the fact that the defaulted `proxy-uri` argument of `ffi-buffer-set-proxy` can lead to the `nil`-ed `proxy-uri` slot for `gtk-buffer` with some undesirable consequences. I'll add a default form of `(quri:uri "")`.

> Why is it unfortunate that "`url-sans-scheme` is no longer there"? Do we miss url-sans-scheme anywhere?

No, it's not used anywhere. I am just missing this fancy piece of French :)

aartaka:
I must warn you in advance -- there's one formatting change, together with these fixes. It was just impossible for me to look at 115-column long lambda-list of `ffi-buffer-set-proxy`.

jmercouris:
@Ambrevar do you see anything else wrong? I believe this is OK to merge.

Ambrevar:
I'll test again.

Ambrevar:
Perfect, thanks a lot for your patience and thorough follow-up!

Issue #817, opened 2020-06-26
Overhaul URL dispatcher (magnet support, customizable)


Ambrevar:
Now that our request-resource-hook is almighty, we can move the URL dispatch from `parse-url` to `request-resource-hook`.

Ideally, we should make it in a way that's easily customizable.
For instance, we need to derive a way to let the user customize which Torrent client to use to open magnet links.
Maybe there is nothing to do beside defining a handler that checks which Torrent client is available and loads the magnet link appropriately.

In the end, the following expression can be removed from `parse-url`:
```lisp
((str:starts-with? "magnet:" input-url)
             (log:debug "Open magnet link with external application.")
             (ignore-errors
              (uiop:launch-program (list "xdg-open" input-url))
              (nyxt/minibuffer-mode:cancel-input)))
```

aartaka:
I'm thinking of a slot in the `browser` class, but I'm not sure about it's type:
- List, like the `search-engines` slot, with the same idea of the user-appendable list of `protocol` objects. But the problem is that one may want to redefine the built-in `protocol`s (whatever the reason may be) and struggle with that because there will be a need in calling `(remove-if (lambda (protocol) (string= "http" (scheme protocol))) %slot-default)` or something like that.
- Hash table, having the essential scheme names (e.g. `"http"`, `"https"`, `"mailto"`, etc.) as keys and `protocol` objects as values. This will ensure that there's exactly one action for the given protocol and store them in one place. The downside is quirky customization -- I can't imagine an easy way of redefining this slot without the knowledge of how has tables work.

The first looks better to me, given that lists of `protocol` objects (the name is not set in stone, so you can suggest a better name, actually!) has an analogous one (`search-engines`) and is easier to manipulate than a hash table, even with the non-uniqueness problem (is there a set-like data structure in `fset` that we can use implicitly to ensure uniqueness, or should we define a list-based uniqueness-preserving type?).

I imagine `protocol` object with 3 slots so far:
- `scheme` slot with a literal `string` to compare the URL scheme against
- `type` slot with either `:nyxt-command`, `:shell-command` or `:lambda` value:
    - `:nyxt-command` can be a short name of the command as either string or symbol (e.g. `"tutorial"`, `"manual"`).
    - `:shell-command` is a list of strings or string to pass to `uiop:run-program`, with the consequence of request cancellation and opening the URL in the external application
[...]

PR #829, opened 2020-07-04
Simplify request-resource-hook's handlers return value

aartaka:
This changes the ambiguous format of the `request-resource-hook`'s handlers return value from `(values request-data &optional (or (eql :stop) (eql :forward)))` to a simpler `(or request-data null)` while still retaining all the power of the `request-resource-hook` and making it more readable and understandable, both for maintainers and users. Documentation was adjusted accordingly.

### Motivation

The format of what `request-resource-hook` return values was relatively complex for me as a project newcomer and there seemed to be no strong use-case for such a complexity of the return value format. Bringing the format of the `request-resource-hook` handlers to a conventional lispy format of "either it's nil, or something meaningful" seemed as a good step towards readability, ease of use and conventionality of the `request-resource-hook`.

### How Has This Been Tested

Because the most breaking change was the change from the expression `(values request-data :stop)` to `nil`, the testing mainly employed enabling several handlers together with the `request-resource-block` handler of `blocker-mode` that returned `nil` instead of the two-valued expression now. Testing revealed that there's a need in failure-testing for each handler, and `hooks:run-hook` was replaced with `hooks:run-hook-with-args-until-failure` the result of which was unwrapped after it's call.

### Other things

- The absence of log notes about blocking the hosts from the default blocking list emerged during testing, but this bug seems to be present in the upstream version, so it's unrelated to this patch.
- The name of the `force-https-handler` argument was changed from `resource` to `request-data` to unify it with other similar handlers.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -14,10 +14,7 @@
 (hooks:define-hook-type keymaps-buffer (function (list-of-keymaps buffer)
                                                  (values &optional list-of-keymaps buffer)))
 (export-always '(make-hook-keymaps-buffer make-handler-keymaps-buffer))
-(hooks:define-hook-type resource (function (request-data)
-                                           (values request-data
-                                                   &optional (or (eql :stop)
-                                                                 (eql :forward)))))
+(hooks:define-hook-type resource (function (request-data) (or request-data null)))
 (export-always '(make-hook-resource make-handler-resource))
 (hooks:define-hook-type uri->uri (function (quri:uri) quri:uri))
 
@@ -170,21 +167,19 @@ This can apply to specific buffer."))
 (export-always '*proxy-class*)
 (defvar *proxy-class* 'proxy)
 
-(export-always 'combine-composed-hook-until-non-nil-second-value)
-(defmethod combine-composed-hook-until-non-nil-second-value ((hook hooks:hook) &rest args)
+(export-always 'combine-composed-hook-until-nil)
+(defmethod combine-composed-hook-until-nil ((hook hooks:hook) &optional arg)
   "Return the result of the composition of the HOOK handlers on ARGS, from
-oldest to youngest.  Stop processsing when a handler returns a non-nil second value.
-Without handler, return ARGS as values.  This is an acceptable `combination' for
+oldest to youngest.  Stop processsing when a handler returns nil.
+Without handler, return ARGS.  This is an acceptable `combination' for
 `hook'."
   (labels ((compose-handlers (handlers result)
              (if handlers
-                 (multiple-value-bind (new-result second-value)
-                     (apply (first handlers) result)
-                   (if second-value
-                       (values new-result second-value)
-                       (compose-handlers (rest handlers) (list new-result))))
-                 (values-list result))))
-    (compose-handlers (mapcar #'hooks:fn (hooks:handlers hook)) args)))
+                 (let ((new-result (funcall (first handlers) result)))
+                   (when new-result
+                     (compose-handlers (rest handlers) new-result)))
+                 result)))
+    (compose-handlers (mapcar #'hooks:fn (hooks:handlers hook)) arg)))
 
 (eval-when (:compile-toplevel :load-toplevel :execute)
   (export 'buffer)
@@ -299,15 +294,18 @@ The functions are expected to take key arguments like `:url'.")
    (request-resource-hook :accessor request-resource-hook
                           :initarg :request-resource-hook
                           :initform (make-hook-resource
-                                     :combination #'combine-composed-hook-until-non-nil-second-value
+                                     :combination #'combine-composed-hook-until-nil
                                      :handlers (list #'request-resource))
                           :documentation "Hook run on every resource load.
-The handlers are composed, passing a `request-data' until one returns a non-nil
-second value.
+The handlers are composed, passing a `request-data'
+until one of them returns nil or all handlers apply successfully.
+
 Newest hook is run first.
-If :FORWARD is returned, the resource loading of the returned `request-data' is
-deferred to the renderer.
-If :STOP is returned, stop the the hook and cancel the resource load.
+If a `request-data' object is returned, it gets passed to other handlers
+or right to the renderer if there are no more handlers.
+If nil is returned, stop the hook and cancel the resource load.
+
+There's no more ability to pass the results to the renderer with :FORWARD.
 
 Example:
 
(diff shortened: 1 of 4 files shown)


aartaka:
Related to the recent discussion of `request-resource-hook` handlers' return value in #817.

source/browser.lisp
```diff
@@ -302,12 +299,13 @@ The functions are expected to take key arguments like `:url'.")
...
+The handlers are composed, passing a `request-data'
+until one of them returns nil or all handlers apply successfully.
+
 Newest hook is run first.
-If :FORWARD is returned, the resource loading of the returned `request-data' is
-deferred to the renderer.
-If :STOP is returned, stop the the hook and cancel the resource load.
+If `request-data' object is returned, it gets passed to other handlers
+or right to renderer if there are no more handlers.
+If nil is returned, stop the the hook and cancel the resource load.
```
Ambrevar, line 308: Typos: "If a `request-data' object... to the renderer"

source/force-https-mode.lisp
```diff
@@ -3,17 +3,17 @@
...
     (when (string= (quri:uri-scheme uri) "http")
       (log:info "HTTPS enforced on \"~a\"" (object-display uri))
       ;; FIXME: http-only websites are displayed as "https://foo.bar"
       ;; FIXME: some websites (e.g., go.com) simply time-out
       (setf (quri:uri-scheme uri) "https"
             (quri:uri-port uri) (quri.port:scheme-default-port "https")
-            (url resource) uri)))
-  (values resource :forward))
+            (url request-data) uri)))
+  request-data)
```
Ambrevar, line 16: By the way, `:forward` here was a mistake that we failed to stop :)  Another good reason to merge this patch!

source/renderer-gtk.lisp
```diff
@@ -484,8 +484,13 @@ Warning: This behaviour may change in the future."
...
           nil)
-        (multiple-value-bind (request-data status)
-            (hooks:run-hook (request-resource-hook buffer)
+        (let ((request-data
+                ;; Last element in the list of handlers' results is either nil
+                ;; (in case request is blocked) or the final request-data object.
+                ;;
+                ;; Is there a better way of getting the last element than (car (last ...))?
+                (car (last (hooks:run-hook-with-args-until-failure
+                            (request-resource-hook buffer)
```
Ambrevar, line 493: Prefer `run-hook` and set the `:combination` slot of the hook to a function that is like `combine-composed-hook` but stops when it encounters nil.

`run-hook-with-args-until-failure` is wrong because it does not compose the hooks.

In the end, you won't need the (car (last ...)) business.  By the way, prefer `first` over `car`.

source/renderer-gtk.lisp
```diff
@@ -495,31 +500,28 @@ Warning: This behaviour may change in the future."
...
-                   ;; confused about which URL to load.
-                   (webkit:webkit-policy-decision-ignore response-policy-decision)
-                   (webkit:webkit-web-view-load-request (gtk-object buffer) request)
-                   nil)))
-            (_
-             (log:debug "Don't forward to renderer (handler stop).")
-             (webkit:webkit-policy-decision-ignore response-policy-decision)
-             nil))))))
+                                           :known-type-p is-known-type))))))
+          ;; Replace with cond for readability?
```
Ambrevar, line 504: Yes! :)

source/browser.lisp
```diff
@@ -302,12 +299,13 @@ The functions are expected to take key arguments like `:url'.")
                                      :combination #'combine-composed-hook-until-non-nil-second-value
```
Ambrevar, line 299: This combination function should be updated (see below about `run-hook`).

Ambrevar:
> while still retaining all the power of the request-resource-hook 

To be exact, I assumed that this change would _not_ be equivalent and lose in expressiveness.
The initial reason why I wanted to pass either :stop or :forward as a second value is because I wanted handlers to be able to express "Stop the hook and load the return value", which is different from "stop the hook and do nothing" (`:stop`).

However, I just figured out that there is another way to do is without second value: before the handler returns `nil`, it can call `(set-url* ...)`, which does exactly what I want.

Note that this seems to be a bit of an undefined behaviour, because we are asking WebKit to load a URI before it has decided to cancel on the current URI.  There is room for a race condition here.  But I've tested and it works.

All that said, I think this patchset should include a comment explaining all this.

Ambrevar:
> The absence of log notes about blocking the hosts from the default blocking list emerged during testing, but this bug seems to be present in the upstream version, so it's unrelated to this patch.

Can you provide a recipe?  I'm not sure I get what you mean.

aartaka:
> Can you provide a recipe? I'm not sure I get what you mean.

Sure, I'll open an issue and provide a recipe there.

aartaka:
> To be exact, I assumed that this change would _not_ be equivalent and lose in expressiveness.
> The initial reason why I wanted to pass either :stop or :forward as a second value is because I wanted handlers to be able to express "Stop the hook and load the return value", which is different from "stop the hook and do nothing" (`:stop`).

Yes, you're right. I guess I just got carried away while writing the pull request description, given that there seemed to be no obvious use-case for this feature :)
> All that said, I think this patchset should include a comment explaining all this.

Hmm, I'm not sure where to put it. Is `request-resource-hook`'s `:documentation` slot a suitable place?

aartaka:
Changes:
- Use `cond` in `on-signal-decide-policy` in the segment handling `request-resource-hook` return value.
- Use `run-hook` in `let` just before the same segment.
- Rewrite `combine-hook-until-non-nil-second-value` as `combine-hook-until-nil`. It is placed in the `request-resource-hook` `:combination` slot. Same in `blocker-mode`.
- Fix typos.

Ambrevar:

> Hmm, I'm not sure where to put it. Is `request-resource-hook`'s `:documentation` slot a suitable place?

Yes!

aartaka:
Here you go:
- Fix a small typo in `compose-hook-until-nil` docstring.
- Add a comment on inability to `:forward` request to `request-resource-hook` documentation.

source/browser.lisp
```diff
@@ -299,15 +294,18 @@ The functions are expected to take key arguments like `:url'.")
...
+The handlers are composed, passing a `request-data'
+until one of them returns nil or all handlers apply successfully.
+
 Newest hook is run first.
-If :FORWARD is returned, the resource loading of the returned `request-data' is
-deferred to the renderer.
-If :STOP is returned, stop the the hook and cancel the resource load.
+If a `request-data' object is returned, it gets passed to other handlers
+or right to the renderer if there are no more handlers.
+If nil is returned, stop the the hook and cancel the resource load.
```
Ambrevar, line 306: Typo: "the the"

aartaka:
Had some misunderstanding of `compose-hook-until-nil` that caused a type mismatch and hooks miscomposition. I've slightly changed its signature to be more in line with the new one-return-value format of hooks.

Ambrevar:
Thank you!  Merged with 043636fcf588c406429bde9bbdca121276d72e94.

PR #830, opened 2020-07-05
Restore sessions lazily.

Ambrevar:
Previously we would tell GTK to load all restored buffers on startup.  This
would hang the browser for a while before it displayed the first window because
the queries are done within a single GTK main loop cycle.

It's hard to load more than very few URIs in a single GTK loop cycle because of
how long it takes.

The easiest (and possibly best) option is to restore the buffers on demand.
This is what this patch does.

Another option would be to restore about one buffer per second, or when the
browser is idle.  The benefit is that this would make all the buffers ready to
browse after a certain time and the user would not have to wait on every buffer switch.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -230,9 +230,13 @@ Dead buffers (i.e. those not associated with a web view) have an empty ID.")
    (load-status ;; :accessor load-status ; TODO: Need to decide if we want progress / errors before exposing to the user.
                 :initarg :load-status
                 :type (or (eql :loading)
-                          (eql :finished))
-                :initform :finished
-                :documentation "Whether the buffer is loading or finished loading.")
+                          (eql :finished)
+                          (eql :unloaded)
+                          (eql :void))
+                :initform :void
+                :documentation "Whether the buffer is loading or finished loading.
+`:unloaded' is for buffers that have not been loaded yet, like session-restored
+buffers.")
    (last-access :accessor last-access
                 :initform (local-time:now)
                 :type local-time:timestamp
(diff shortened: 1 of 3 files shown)


Ambrevar:
@jmercouris Thoughts on this?

jmercouris:
Ah yes, I'm sorry, I will review today! I read this and then forgot!

source/browser.lisp
```diff
@@ -230,9 +230,13 @@ Dead buffers (i.e. those not associated with a web view) have an empty ID.")
    (load-status ;; :accessor load-status ; TODO: Need to decide if we want progress / errors before exposing to the user.
                 :initarg :load-status
                 :type (or (eql :loading)
-                          (eql :finished))
-                :initform :finished
-                :documentation "Whether the buffer is loading or finished loading.")
+                          (eql :finished)
+                          (eql :unloaded)
+                          (eql :void))
```
jmercouris, line 235: what is a load status of void?

jmercouris: I see it is used as a flag for the predicate `replacement-buffer-p`, but what does void mean?

Ambrevar: For now it's only used as a marker for dummy pages.  It means that nothing was set to be loaded, meaning the page is void.

Do you suggest a better term?  Or maybe a better way to mark replacement buffers?

jmercouris: Your explanation makes sense! I think it is perfect as is.

Ambrevar:
If you had to ask, it probably means the docstring is not clear enough ;)

PR #832, opened 2020-07-06
Add customizable URL-dispatching

### LINKED ISSUE #817: Overhaul URL dispatcher (magnet support, customizable)

aartaka:
This adds a mechanism for a customizable URL-dispatching and a simple way to customize it.

### Motivation

There can be a necessity to change browser behavior while encountering a certain type of link (e.g., open torrent client on "magnet:" link, redirect "doi:" protocol link to the DOI website). Using handlers and `request-resource-hook` is too much of a work for a simple task of doing something depending on domain/scheme/host/whatever in a URL. There's a need in a simpler way to act on URLs while having robustness of hooks.

### What's Not There (Yet)

- Shorter syntax for shell command invocations on matching URLs. Although it can be built on top of this URL dispatching mechanism easily, it needs more discussion.

### How Has This Been Tested

Browser was started with this config file:
``` lisp
(in-package #:nyxt-user)

(define-configuration browser
  ((url-dispatchers `(,(url-dispatching-handler
                        (match-scheme "doi")
                        (lambda (url)
                          (quri:uri (format nil "https://doi.org/~a"
                                            (quri:uri-path url))))
                        "Resolve DOI links")
                      ,(url-dispatching-handler
                        (match-host "reddit.com")
                        (lambda (url)
                          (quri:copy-uri url :host "old.reddit.com"))
                        "Open old Reddit")
                      ,(url-dispatching-handler
                        (match-scheme "javascript")
                        (lambda (url)
                          (ffi-buffer-evaluate-javascript
                           (current-buffer) (quri:uri-path url))
                          nil))
                      ,@%slot-default))))
```
And the given handlers worked in the appropriate situations. No bugs were noticed.

I'll be glad to see your feedback on that!

Closes #817

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -1102,6 +1102,67 @@ Deal with REQUEST-DATA with the following rules:
          (log:debug "Forwarding: ~a" (object-display url))
          request-data)))))
 
+(export-always 'url-dispatching-handler)
+(defun url-dispatching-handler (name test action)
+  "Return a `resource' handler that, if `add-hook'ed to the `request-resource-hook',
+will automatically apply its ACTION on the URLs that conform to TEST.
+
+TEST and ACTION should be functions of one argument, the requested URL.
+In case ACTION returns nil, URL request is aborted.
+The new URL returned by ACTION is loaded otherwise.
+
+`match-host', `match-scheme', `match-domain' and `match-file-extension'
+can be used to create TEST-functions, but any other function of one argument
+would fit the TEST slot as well.
+
+Example:
+
+\(define-configuration buffer
+    ((request-resource-hook (reduce #'hooks:add-hook
+                                    (list (url-dispatching-handler
+                                           'doi-link-dispatcher
+                                           (match-scheme \"doi\")
+                                           (lambda (url)
+                                             (quri:uri (format nil \"https://doi.org/~a\"
+                                                               (quri:uri-path url)))))
+                                          (url-dispatching-handler
+                                           'transmission-magnet-links
+                                           (match-scheme \"magnet\")
+                                           (lambda (url)
+                                             (uiop:launch-program
+                                              (list \"transmission-remote\" \"--add\"
+                                                    (object-string url))
+                                              nil))))
+                                    :initial-value %slot-default))))"
+  (make-handler-resource
+    #'(lambda (request-data)
+        (let ((url (url request-data)))
+          (if (funcall-safely test url)
+              (let ((new-url (funcall-safely action url)))
+                (log:info "Applied \"~a\" URL-dispatcher on \"~a\" and got \"~a\""
+                          (symbol-name name) (object-display url) (object-display new-url))
+                (when new-url (setf (url request-data) new-url) request-data))
+              request-data)))
+    :name name))
+
+(declaim (ftype (function (string) (function (quri:uri) boolean))
+                match-scheme match-host match-domain match-file-extension))
+(export-always 'match-scheme)
+(defun match-scheme (scheme)
+  #'(lambda (url) (string= scheme (quri:uri-scheme url))))
+
+(export-always 'match-host)
+(defun match-host (host)
+  #'(lambda (url) (string= host (quri:uri-host url))))
+
+(export-always 'match-domain)
+(defun match-domain (domain)
+  #'(lambda (url) (string= domain (quri:uri-domain url))))
+
+(export-always 'match-file-extension)
+(defun match-file-extension (extension)
+  #'(lambda (url) (string= extension (pathname-type (quri:uri-path url)))))
+
 (defun javascript-error-handler (condition)
   (echo-warning "JavaScript error: ~a" condition))
 
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -67,6 +67,9 @@ have priorities over the other modes key bindings.")
     (:p "See the " (:code "search-engines") " browser slot documentation.
 Bookmarks can also be used as search engines, see the corresponding section.")
 
+    (:h3 "URL-dispatchers")
+    (:p "See the" (:code "url-dispatching-handler") " function documentation.")
+
     (:h3 "Downloads")
     (:p "See the " (:code "download-list") " command and the "
         (:code "download-path") " browser slot documentation.")


Ambrevar:
Thanks, it's a great start.

This raises a design question though. Now I understand what you intention was in your last comment, but I don't understand the reason: why create a new hook?
Why not create a "handler contruction helper"?
Something like that (untested):

```lisp
(define-configuration browser
    ((request-resource-hook
      (reduce #'hooks:add-hook
              (url-dispatching-handler
               (match-scheme "doi")
               (lambda (url)
                 (quri:uri (format nil "https://doi.org/~a"
                                   (quri:uri-path url))))
               :initial-value %slot-default)))))
```

Ambrevar:
(The `url-dispatching-handler` helpder function would then return a resource handler.)

aartaka:
There are two weak reasons:
- User convenience: it's easier to make a list of simple handlers and to leave composing them to some other function.
- Narrow focus: if this function is focused on the URL transformation and redirection only, then it should be `uri->uri` handler. If it will be `resource` handler, it'll lose this focus and idea. Something can go wrong because of that, either conceptually (user intentionally would use it to modify `request-data`) or implementationally (it will be implemented in a way that would be leaky and allow for these same `request-data` side-effects).

Again, these reasons are weak and rewriting it for `request-resource-hook` sounds better to me.

However, I don't have a clear idea of how it could look (at least yet). Your example is valid, but having 3+ dispatching handlers and some other `resource` handlers would be hard to read. You can prove me wrong though.

Maybe add a dedicated function to abstract this `(reduce #'hooks:add-hook ... )` or am I trying to abstract too much? :)

Ambrevar:
Some counter-points:

- The `(reduce ...)` only adds one line.
- The `url-dispatching-handler` helper is the center of the abstraction here.  It abstracts away the request-data and leaves the user only with the quri:uri.
- We can store the list of URI dispatches in a separate list.  It's a full-blown programming language after all! :)  Example


```lisp
(defvar my-dispatchers
  (list
   (url-dispatching-handler
    (match-scheme "doi")
    (lambda (url)
      (quri:uri (format nil "https://doi.org/~a"
                        (quri:uri-path url)))))
   (url-dispatching-handler
    (match-scheme "magnet")
    (lambda (url)
      ...))))

(define-configuration browser
    ((request-resource-hook
      (reduce #'hooks:add-hook
              (append
               my-dispatchers
               my-non-dispatchers)
              :initial-value %slot-default))))
```

Do you like it?

aartaka:
Yes, you're right. I've pushed the changes you suggested. 

I've added a `hyphenize-dispatching-test` function -- it can seem unnecessary now, but it can be handy for smart per domain mode enabling/disabling from #376. Specifically, it can allow us to distinguish the hosts/domains that either the browser or the user have set rules for -- we can search them by names!

Ambrevar:
Sorry, I forgot that our handlers had to be named :)

Instead of the macro and gensym, what about adding a `:name` key argument to the  `url-dispatching-handler` helper?

Ambrevar:
Then we would not need a macro nor `hyphenize-dispatching-test`.

source/browser.lisp
```diff
@@ -1102,6 +1102,69 @@ Deal with REQUEST-DATA with the following rules:
...
+
+Example:
+\(define-configuration buffer
+   ((request-resource-hook (hooks:add-hook
+                            (url-dispatching-handler
+                                (match-scheme \"doi\")
+                                (lambda (url)
+                                  (quri:uri (format nil \"https://doi.org/~a\"
+                                                    (quri:uri-path url))))
+                                \"Resolve DOI links\")
```
Ambrevar, line 1134: I'd add the magnet example since it shows how to open URI with external programs.

aartaka: Sure.

aartaka:
> Instead of the macro and gensym, what about adding a `:name` key argument to the `url-dispatching-handler` helper?

Yes, I thought about it myself. I'll push it in an hour!

source/browser.lisp
```diff
@@ -1102,6 +1102,66 @@ Deal with REQUEST-DATA with the following rules:
          (log:debug "Forwarding: ~a" (object-display url))
          request-data)))))
 
+(export-always 'url-dispatching-handler)
+(defun url-dispatching-handler (test action &optional (description "") (name (gensym)))
```
Ambrevar, line 1106: The philosophy of our hook system is to be legible, unlike Emacs hooks which accept lambda.
`gensym` hides the intention away from the user.  This is why I suggest that we do like the rest of the Nyxt hook system and we force the use of a name.

Actually, if names are mandatory, aren't the description and the name redundant?  If so, then feel free to remove `description`.

source/browser.lisp
```diff
@@ -1102,6 +1102,66 @@ Deal with REQUEST-DATA with the following rules:
...
+  "Returns a `resource' handler that, if `add-hook'ed to the `request-resource-hook',
```
Ambrevar, line 1107: Use the imperative form `Return`.

source/browser.lisp
```diff
@@ -1102,6 +1102,66 @@ Deal with REQUEST-DATA with the following rules:
...
+will automatically apply its ACTION on the URLs that conform to TEST.
+
+TEST and ACTION should be the functions of one argument -- the requested URL.
```
Ambrevar, line 1110: Typo: "should be functions".
Nit: I would replace the `--` by a comma.  As you prefer.

source/browser.lisp
```diff
@@ -1102,6 +1102,66 @@ Deal with REQUEST-DATA with the following rules:
...
+In case ACTION returns nil, URL request is aborted.
+New URL returned by ACTION is loaded otherwise.
```
Ambrevar, line 1112: The new URL

source/browser.lisp
```diff
@@ -1102,6 +1102,66 @@ Deal with REQUEST-DATA with the following rules:
...
+
+`match-host', `match-scheme', `match-domain' and `match-file-extension'
+can be used to create TEST-functions.
```
Ambrevar, line 1115: This sounds like they are the only valid test functions.  I would rephrase this and list them as example functions.

Ambrevar:
I guess we are almost good to merge (sorry for being picky :p).

Previously we discussed the convention that the `URI` symbol would refer to `QURI:URI` objects, while `URL` would refer to strings.  Do you like this convention?  If so, please adapt your code to follow it.  Thanks! :)

aartaka:
> I guess we are almost good to merge (sorry for being picky :p).

It makes for good code, so it's nothing to be sorry about :)

> Previously we discussed the convention that the `URI` symbol would refer to `QURI:URI` objects, while `URL` would refer to strings. Do you like this convention? If so, please adapt your code to follow it. Thanks! :)

Hmmmm, I'm not sure there. Several reasons to doubt this distinction:
- Terms URI and URL are ill-defined and are frequently used interchangeably, so distinguishing between these is hard.
- Having type-tied names is cool, but type `declare`ations would work better. That's what I remember from the talk on my habit to put type requirements in docstrings.
- While `quri:uri` has it's name, we're mainly using it to store the website links, URLs. A meaningful name would be URL in this situation, while URI is the format, not meaning, and type delivers the format idea it well enough for it be in the name.

jmercouris:
Artyom has made a very good point here, I'm starting to doubt my earlier position :-D. There are however inconsistencies here, we use `-p` for example to signify that a symbol is pointing to a predicate. Type declarations are great, but it is also nice to have a symbol that makes things clear without having to reference them. What about this?:

URI when using URI
and URL-string when using a string representation

aartaka:
Hmmm, given that string representation is an exception now, it's sounds good to make it stand out. But, outside this, we use `quri:uri` to store locations on the web by default and in most cases. I think this default behavior doesn't need explicit signs outside type declarations and documentation.

aartaka:
So I'm still on the side of `url` name for URLs, no matter what their format is, and `-string` postfix for the string-represented ones.

Ambrevar:
We might be bike-shedding here a little :)

aartaka:
So, what stops the merging?

Ambrevar:
Nothing, I just need to find time to test and merge today.  Will do, no worries!

Ambrevar:
This is a great change, thanks a lot for your awesome contribution again!

jmercouris:
+1 Well done!

PR #833, opened 2020-07-07
Fix compilation of Qt renderer

KrzysiekJ:
This change reflects commits 85f9ca3a44d0c0b and 19c673879539c408.

While the Qt renderer is compilable after this commit, on my machine it results in a blank screen without a title, so this fix may well be incorrect/incomplete.


Ambrevar:
Indeed, the Qt renderer does not work (yet), it is still at the stage of a proof of concept.  We are focusing on WebKitGTK at the moment but hopefully the Qt renderer development will resume soon enough to bring it up  to speed! :)

Anyways, thanks for the fixes, it's a start :)

PR #835, opened 2020-07-08
Accept &rest arguments in match-* helpers for url-dispatching-handler.

Ambrevar:
@aartaka What do you think of this?
It makes the syntax a bit shorter, e.g.

```lisp
(defparameter old-reddit-handler
  (url-dispatching-handler
   'old-reddit-dispatcher
   (match-host  "www2.reddit.com" "www.reddit.com")
   (lambda (url)
     (quri:copy-uri url :host "old.reddit.com"))))
```

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -1145,23 +1145,35 @@ Example:
               request-data)))
     :name name))
 
-(declaim (ftype (function (string) (function (quri:uri) boolean))
+(declaim (ftype (function (string &rest string) (function (quri:uri) boolean))
                 match-scheme match-host match-domain match-file-extension))
 (export-always 'match-scheme)
-(defun match-scheme (scheme)
-  #'(lambda (url) (string= scheme (quri:uri-scheme url))))
+(defun match-scheme (scheme &rest other-schemes)
+  "Return a predicate for URLs matching one of SCHEME or OTHER-SCHEMES."
+  #'(lambda (url)
+      (some (alex:curry #'string= (quri:uri-scheme url))
+            (cons scheme other-schemes))))
 
 (export-always 'match-host)
-(defun match-host (host)
-  #'(lambda (url) (string= host (quri:uri-host url))))
+(defun match-host (host &rest other-hosts)
+  "Return a predicate for URLs matching one of HOST or OTHER-HOSTS."
+  #'(lambda (url)
+      (some (alex:curry #'string= (quri:uri-host url))
+            (cons host other-hosts))))
 
 (export-always 'match-domain)
-(defun match-domain (domain)
-  #'(lambda (url) (string= domain (quri:uri-domain url))))
+(defun match-domain (domain &rest other-domains)
+  "Return a predicate for URLs matching one of DOMAIN or OTHER-DOMAINS."
+  #'(lambda (url)
+      (some (alex:curry #'string= (quri:uri-domain url))
+            (cons domain other-domains))))
 
 (export-always 'match-file-extension)
-(defun match-file-extension (extension)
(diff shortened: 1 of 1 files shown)


source/browser.lisp
```diff
@@ -1145,23 +1145,35 @@ Example:
               request-data)))
     :name name))
 
-(declaim (ftype (function (string) (function (quri:uri) boolean))
+(declaim (ftype (function (string &rest string) (function (quri:uri) boolean))
                 match-scheme match-host match-domain match-file-extension))
 (export-always 'match-scheme)
-(defun match-scheme (scheme)
-  #'(lambda (url) (string= scheme (quri:uri-scheme url))))
+(defun match-scheme (scheme &rest other-schemes)
```
aartaka, line 1151: Having a `(&rest schemes)` in lambda list will make args `cons`-ing unnecessary (i.e. `(cons scheme other-schemes)` would be a simple `schemes`).

source/browser.lisp
```diff
@@ -1145,23 +1145,35 @@ Example:
...
+  "Return a predicate for URLs matching one of SCHEME or OTHER-SCHEMES."
+  #'(lambda (url)
+      (some (alex:curry #'string= (quri:uri-scheme url))
+            (cons scheme other-schemes))))
 
 (export-always 'match-host)
-(defun match-host (host)
-  #'(lambda (url) (string= host (quri:uri-host url))))
+(defun match-host (host &rest other-hosts)
```
aartaka, line 1158: Same there.

source/browser.lisp
```diff
@@ -1145,23 +1145,35 @@ Example:
...
+  "Return a predicate for URLs matching one of HOST or OTHER-HOSTS."
+  #'(lambda (url)
+      (some (alex:curry #'string= (quri:uri-host url))
+            (cons host other-hosts))))
 
 (export-always 'match-domain)
-(defun match-domain (domain)
-  #'(lambda (url) (string= domain (quri:uri-domain url))))
+(defun match-domain (domain &rest other-domains)
```
aartaka, line 1165: Same.

source/browser.lisp
```diff
@@ -1145,23 +1145,35 @@ Example:
...
+  "Return a predicate for URLs matching one of DOMAIN or OTHER-DOMAINS."
+  #'(lambda (url)
+      (some (alex:curry #'string= (quri:uri-domain url))
+            (cons domain other-domains))))
 
 (export-always 'match-file-extension)
-(defun match-file-extension (extension)
-  #'(lambda (url) (string= extension (pathname-type (quri:uri-path url)))))
+(defun match-file-extension (extension &rest other-extensions)
```
aartaka, line 1172: And there too :)

source/browser.lisp
Ambrevar: This is on purpose!  This way we make it explicit (and a compile error) to call these functions with at least one argument.  Otherwise `(make-scheme)` would pass but blow up at run-time, which ain't pretty.

aartaka: It'll not blow, because `string=` accepts `nil` as argument and validly returns `nil`. But I agree it isn't pretty to have an unexpected `nil` there :)

Ambrevar:
The point is that `(match-scheme)` (without argument) can only be a
programming mistake, so it's good if the compiler can warn us about it.
With only a `&rest` argument, we won't get any warning.  Hence my "trick".

aartaka:
Yes, that makes sense. Sorry for this change request then :)

Ambrevar:
Thanks for your review!

Issue #717, opened 2020-05-04
Can't log in to various sites


khinsen:
**Describe the bug**

Login for many sites just doesn't seem to work with Next.

Example: https://org-roam.discourse.group/

The exact same login/password combination that works with Firefox does not work with Next. I always get back to a dialog box asking me for login plus password. Other sites work fine though (GitHub and Twitter, for example).

**Precise recipe to reproduce the issue**

**Information**
- OS name+version: macOS 10.14.6
- Next version (from =M-x next-version= or =next --version=): 1.5.0 (but that's wrong, I am on a recent build from source, commit f454a9827d673b62fa86e06948d68e76ff1d3fad)
- How you installed Next (Guix pack, package manager, build from source): built from source
- Lisp implementation/version (if built from source): SBCL 2.0.4 from MacPorts
- WebKitGTK+ | QtWebEngine version:

Ambrevar:
Can you reproduce with Epiphany?

Might be the same as #644.

khinsen:
Epiphany works fine for all those logins.

Ambrevar:
Can you reproduce on commit 817223e050f5cce265780c98446cfcb9850dcef4 ?
(There is a regression which makes Next crash much more, so you might
have to insist a little bit.  But logins might be fixed.)

Ambrevar:
I can login to org-roam on the aforementioned commit (built from source, on Guix System).

Ambrevar:
> Next version (from =M-x next-version= or =next --version=): 1.5.0 (but that's wrong, I am on a recent build from source, commit f454a98)

@khinsen Do you have `git` in your `PATH`?  This is needed to set the `+version+` variable in `global.lisp`.

khinsen:
With commit 817223e050f5cce265780c98446cfcb9850dcef4, I can indeed log in under Guix. Under macOS, I get a highly reproducible WebKit Internal Error when loading https://org-roam.discourse.group/.

khinsen:
As for `git`, yes, it's in my path, but under macOS, `$PATH` is not passed to GUI apps launched from the desktop, so Next won't see it.

Ambrevar:
@jmercouris ?

azzamsa:
I can't login with twitter (even with the same password) the `login` button doesn't shows up. So I change my password for the longer one. It works, but several feature didn't work. e.g update email. It's seems like I also can't paste image here (like usually I did in FF)

jmercouris:
[...]

Issue #644, opened 2020-04-10
Background buffer: Logged out as soon as buffer is switched


hsanjuan:
**Describe the bug**

Soon after I switch buffers I get logged out of beta.protonmail.com.

**Precise recipe to reproduce the issue**

1) Login
2) Switch buffer to something else
3) Come back and we're logged out.

I think I saw an "invalid token" popup when I switched back. 

There is an XHR request every few seconds from the web application that probably keeps the session alive. `Next` must lose some context on tab switching or something so that this request is either not happening or is happening with the wrong information. Any ideas? Tabs in the background should still be able to do their things.

Ambrevar:
I can reproduce with Epiphany.  I guess this is a WebKitGTK+ bug.
Which version do you have?  Would like to report the bug upstream?

hsanjuan:
I have 2.28. I cannot seem to reproduce it with Epiphany though. I tried again with `Next` and it seemed to need a bit longer, but happened fast enough.

Ambrevar:
I did reproduce with Epiphany, so it might be a heisenbug :p

hsanjuan:
How would you go about debugging this a bit? `--verbose` does not show much. I guess what I am asking is something akin to the web developer console network tab so that I get a log of requests made and parameters and figure out if it was a request not happening, or a request happening with the wrong data etc...

Ambrevar:
I'd like to add support for the WebKit Inspector soon, stay tuned!

Ambrevar:
Done in d570f80002b73f3f67cd61db66a86d45101628bb.
You can now call the `open-inspector` command to fire up the inspector.

jmercouris:
Yay!

jmercouris:
![Screen Shot 2020-04-13 at 17 41 45](https://user-images.githubusercontent.com/1691662/79134479-11860a00-7dae-11ea-8d86-3d079616a772.png)

Ambrevar:
I'm realizing there might be a misunderstanding with the bug description:
@hsanjuan Did you mean that you switched to another _buffer_ then went back to the one with protonmail to find it logged out?

In my case, I could reproduce the issue by loading a different domain in the _current_ protonmail buffer, then browsing back to protonmail (from the history).  All in the same buffer.

hsanjuan:
> I'm realizing there might be a misunderstanding with the bug description:
> @hsanjuan Did you mean that you switched to another buffer then went back to the one with protonmail to find it logged out?

[...]

PR #839, opened 2020-07-09
Per buffer contexts

### LINKED ISSUE #826: Security: proxy-mode is misleadingly global
Our proxy-mode is global, but as a mode it's meant to be buffer-local.
We are dangerously misleading our users here :(

The fix is to use per-buffer WebContexts.
The problem is that on doing so,

1. It generates hundreds of

```
(process:23271): libsoup-WARNING **: 11:40:55.141: Failed to execute query: table soup_hsts_policies already exists

(process:23271): libsoup-WARNING **: 11:40:55.141: Failed to execute query: database is locked
```

2. The browser tends to crash after about 1 minute, with

```
CORRUPTION WARNING in SBCL pid 23231(tid 0x7fd373667700):
Memory fault at (nil) (pc=0x7fd3796214e0, fp=0x7fd3640a4b48, sp=0x7fd3736661c0) tid 0x7fd373667700
The integrity of this image is possibly compromised.
Continuing with fingers crossed.

debugger invoked on a SB-SYS:MEMORY-FAULT-ERROR in thread
#<THREAD "cl-cffi-gtk main thread" RUNNING {1005E73CE3}>:
  Unhandled memory fault at #x0.

The current thread is not at the foreground,
SB-THREAD:RELEASE-FOREGROUND has to be called in #<SB-THREAD:THREAD "main thread" waiting on: #<MUTEX "thread result lock" owner: #<SB-THREAD:THREAD "cl-cffi-gtk main thread" RUNNING {1005E73CE3}>> {1000650083}>
for this thread to enter the debugger.
```

The 2 issues could be related.  Let's get started with the first one.

Ambrevar:
Great news!  Since @aartaka implemented data-manager support in cl-webkit, we can now use per-buffer contexts without crashing Nyxt (I've ran it for half an hour so far).

This fixes #826.
Can you test on your site and confirm the browser does not crash (leave it up for at least 10 minutes).

@hsanjuan This might also fix the Protonmail logouts (#644).  Could you check if it works for you?  I've tried it here, Protonmail is still logged in after 30 minutes and many buffer switches.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -61,7 +61,10 @@ Parent directories are created if necessary."
      status-formatter
      window-delete-hook)))
 (defclass window ()
-  ((id :accessor id :initarg :id)
+  ((id :accessor id
+       :initarg :id
+       :type string
+       :initform "")
    (active-buffer :reader active-buffer :initform nil)
    (active-minibuffers :accessor active-minibuffers :initform nil
                        :documentation "The stack of currently active minibuffers.")
@@ -215,7 +218,10 @@ Without handler, return ARG.  This is an acceptable `combination' for
      buffer-delete-hook
      default-cookie-policy)))
 (defclass buffer ()
-  ((id :accessor id :initarg :id :initform ""
+  ((id :accessor id
+       :initarg :id
+       :type string
+       :initform ""
        :documentation "Unique identifier for a buffer.
 Dead buffers (i.e. those not associated with a web view) have an empty ID.")
    ;; TODO: Or maybe a dead-buffer should just be a buffer history?
@@ -946,10 +952,10 @@ current buffer."
             nil))))))
 
 (defmethod get-unique-window-identifier ((browser browser))
-  (incf (slot-value browser 'total-window-count)))
+  (format nil "~s" (incf (slot-value browser 'total-window-count))))
 
 (defmethod get-unique-buffer-identifier ((browser browser))
-  (incf (slot-value browser 'total-buffer-count)))
+  (format nil "~s" (incf (slot-value browser 'total-buffer-count))))
 
 (declaim (ftype (function (&optional window buffer)) set-window-title))
 (export-always 'set-window-title)
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -50,11 +50,14 @@ If URL is `:default', use `default-new-buffer-url'."
 Run `*browser*'s `buffer-make-hook' over the created buffer before returning it.
 If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
   (let* ((buffer (if dead-buffer
-                     (ffi-buffer-make dead-buffer)
+                     (progn
+                       ;; Dead buffer ID must be renewed before calling `ffi-buffer-make'.
+                       (setf (id dead-buffer) (get-unique-buffer-identifier *browser*))
+                       (ffi-buffer-make dead-buffer))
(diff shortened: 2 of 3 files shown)


Ambrevar:
This could fix #717 as well. @khinsen Mind checking?

Ambrevar:
This patch does not fix the libsoup warnings though, I wonder why.

source/browser.lisp
```diff
@@ -946,10 +952,10 @@ current buffer."
             nil))))))
 
 (defmethod get-unique-window-identifier ((browser browser))
-  (incf (slot-value browser 'total-window-count)))
+  (format nil "~s" (incf (slot-value browser 'total-window-count))))
```
jmercouris, line 955: Why format? You want a string specifically?

Ambrevar:
Yes, our identifiers have always been strings!  (Look at the commit message.)

khinsen:
> This could fix #717 as well. @khinsen Mind checking?

I will check, but not before August, sorry!

hsanjuan:
I will need a couple of days, but my issue was already not happening in recent versions.

source/browser.lisp
Ambrevar: Yes, our identifiers have always been strings!  (Look at the commit message.)

hsanjuan:
I'm getting 

```
; caught ERROR:
;   
;     Symbol "WEBKIT-WEBSITE-DATA-MANAGER" not found in the CL-WEBKIT2 package.
;   
;       Line: 421, Column: -1, File-Position: 17818
;   
;       Stream: #<SB-INT:FORM-TRACKING-STREAM for "file /home/hector/common-lisp/nyxt/source/renderer-gtk.lisp" {1007079673}>

; compilation aborted after 0:00:00.126
Unhandled UIOP/LISP-BUILD:COMPILE-FILE-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                                          {10009F0083}>:
  COMPILE-FILE-ERROR while compiling #<CL-SOURCE-FILE "nyxt/gtk" "renderer-gtk">
```

With both the commited version of submodules and the latest one.

Ambrevar:
Does #853 fix it for you?
cl-webkit master has the webkit-website-data-manager symbol, it seems that your setup is not picking the right cl-webkit.  How are you building Nyxt?

Ambrevar:
@hsanjuan Does it build now for you?

hsanjuan:
I don't manage, I build with `make all`.

I am using latest cl-webkit from master and it fails after running `make clean`, with or without NYXT_INTERNAL_QUICKLISP=true... I am not sure it is using the right cl-webkit as I have it checked out? Could it be cached on an old version somewhere?

hsanjuan:
Of course, the moment I sent the message above I realized I had `cl-webkit` in `~/common-lisp` because that was the thing before submodules were added. Removing that fixed the build.

hsanjuan:
Very preliminary tests show no issues with Protonmail.

Ambrevar:
Great! Thanks for confirming!

About the ~/common-lisp issue, I wonder if we can add a flag to tell ASDF to ignore this folder.  This is a recurring issue.  But now that we've got submodules, newcomers should not be hit by this again.

Ambrevar:
Merged with acca0d81cee102830ff05c5c6fccbb0fc87bec46.  @khinsen Let us know when you can if this fixes #717 (anyways, we are not closing that issue just yet).

PR #841, opened 2020-07-10
Add shell-command strings support in url-dispatching-handler

### LINKED ISSUE #834: Add support for direct program execution in url-dispatching-handler
It'd be neat if we could call external programs more easily, e.g. instead of

```lisp
(url-dispatching-handler
   'transmission-magnet-links
   (match-scheme "magnet")
   (lambda (url)
     (uiop:launch-program
      (list "transmission-remote" "--add"
            (object-string url))
      nil)))
```

allow

```lisp
(url-dispatching-handler
   'transmission-magnet-links
   (match-scheme "magnet")
   '("transmission-remote" "--add")))
```

aartaka:
This adds the ability to pass shell-commands as strings to `url-dispatching-handler`.

### Motivation

While a lot of things can be done with high-level mechanisms, like lambdas, shell isn't going anywhere, and having a simple and convenient way to call shell commands is a great extensibility advantage. Having a short and straight-to-the-point string-encoded shell-command syntax in `url-dispatching-handler` `action` argument is a step towards this convenience.

### What's New
- `run-shell-command` helper function, inspired by StumpWM's `run-shell-command`. It supports only `sh` for now, but can be easily extended to support other shells.
- Possibility to pass strings (actually, `format`'s `control-string`s, so `"transmission-remote --add ~s"` is a valid `action` there) as `action`s to `url-dispatching-handler`.
- New optional `url-formatter` argument to `url-dispatching-handler`, whose sole purpose is to format URL properly before passing them to shell-commands.
- Documentation of `url-dispatching-handler` was changed accordingly.

### How Has This Been Tested

- I've added new `url-dispatching-handler` to `request-resource-hook` and ran Nyxt with it. No bugs were noticed, except for the fact that the directory Nyxt runs shell scripts in is "~/.config/nyxt/", but that's not a bug. That's what this handler looked like.
``` lisp 
(url-dispatching-handler
  'emacs-file
  (match-scheme "emacs")
  "emacs ~s"
  #'quri:uri-path)
```

I am conscious that this testing is not exhaustive (sorry, I haven't come up with the cases to use this syntax for), and you can test it yourself to make sure that it works properly. I'll be happy to see your feedback and testing notes on that!

Closes #834.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -1103,12 +1103,17 @@ Deal with REQUEST-DATA with the following rules:
          request-data)))))
 
 (export-always 'url-dispatching-handler)
+(declaim (ftype (function (symbol
+                           (function (quri:uri) boolean)
+                           (or string (function (quri:uri) (or quri:uri null)))))
+                url-dispatching-handler))
 (defun url-dispatching-handler (name test action)
   "Return a `resource' handler that, if `add-hook'ed to the `request-resource-hook',
 will automatically apply its ACTION on the URLs that conform to TEST.
 
-TEST and ACTION should be functions of one argument, the requested URL.
-In case ACTION returns nil, URL request is aborted.
+TEST should be function of one argument, the requested URL.
+ACTION can be either a shell command as a string, or a function taking a URL as argument.
+In case ACTION returns nil (always the case for shell command), URL request is aborted.
 The new URL returned by ACTION is loaded otherwise.
 
 `match-host', `match-scheme', `match-domain' and `match-file-extension'
@@ -1128,22 +1133,40 @@ Example:
                                           (url-dispatching-handler
                                            'transmission-magnet-links
                                            (match-scheme \"magnet\")
+                                           \"transmission-remote --add ~a\")
+                                          (url-dispatching-handler
+                                           'emacs-file
+                                           (match-scheme \"file\")
                                            (lambda (url)
                                              (uiop:launch-program
-                                              (list \"transmission-remote\" \"--add\"
-                                                    (object-string url))
-                                              nil))))
+                                              `(\"emacs\" ,(quri:uri-path url)))
+                                             nil)))
                                     :initial-value %slot-default))))"
   (make-handler-resource
-    #'(lambda (request-data)
-        (let ((url (url request-data)))
-          (if (funcall-safely test url)
-              (let ((new-url (funcall-safely action url)))
-                (log:info "Applied \"~a\" URL-dispatcher on \"~a\" and got \"~a\""
-                          (symbol-name name) (object-display url) (object-display new-url))
-                (when new-url (setf (url request-data) new-url) request-data))
-              request-data)))
-    :name name))
+   #'(lambda (request-data)
+       (let ((url (url request-data)))
+         (if (funcall-safely test url)
+             (etypecase action
+               (function
+                (let* ((new-url (funcall-safely action url)))
+                  (log:info "Applied ~s URL-dispatcher on ~s and got ~s"
+                            (symbol-name name)
+                            (object-display url)
+                            (object-display new-url))
+                  (when new-url
+                    (setf (url request-data) new-url)
+                    request-data)))
+               (string (let ((action #'(lambda (url)
+                                         (uiop:launch-program
+                                          (format nil action
+                                                  (object-string url)))
+                                         nil)))
+                         (funcall-safely action url)
+                         (log:info "Applied ~s shell-command URL-dispatcher on ~s"
+                            (symbol-name name)
+                            (object-display url)))))
+             request-data)))
+   :name name))
 
 (declaim (ftype (function (string) (function (quri:uri) boolean))
                 match-scheme match-host match-domain match-file-extension))


source/utility.lisp
```diff
@@ -82,6 +82,13 @@ before running this command."
                        (uiop:launch-program command))))
        (notify (if (zerop exit-code) success-msg error-msg))))))
 
+(export-always 'run-shell-command)
+(declaim (ftype (function (string) t) run-shell-command))
+(defun run-shell-command (command)
+  ;; TODO: support other shells?
+  ;; StumpWM uses *shell-program* special variable
+  (uiop:launch-program `("sh" "-c" ,command)))
```
Ambrevar, line 90: I believe we don't need `run-shell-command` at all since `uiop:launch-program` takes a string as well, which is equivalent and does not hard-code the shell.

source/browser.lisp
```diff
@@ -1128,22 +1137,31 @@ Example:
...
+                            (object-display new-url))
+                  (when new-url (setf (url request-data) new-url) request-data)))
+               (string (let ((action #'(lambda (url)
+                                         (run-shell-command
+                                          (format nil action
+                                                  (funcall-safely url-formatter url)))
+                                         nil)))
+                         (funcall-safely action url)
+                         (log:info "Applied \"~a\" shell-command URL-dispatcher on \"~a\""
+                            (symbol-name name)
```
Ambrevar, line 1161: Nit: You can use `~s` to print the string with surrounding quotes.

source/browser.lisp
```diff
@@ -1128,22 +1137,31 @@ Example:
...
+       (let ((url (url request-data)))
+         (if (funcall-safely test url)
+             (etypecase action
+               (function
+                (let* ((new-url (funcall-safely action url)))
+                  (log:info "Applied \"~a\" URL-dispatcher on \"~a\" and got \"~a\""
+                            (symbol-name name)
+                            (object-display url)
+                            (object-display new-url))
+                  (when new-url (setf (url request-data) new-url) request-data)))
```
Ambrevar, line 1153: I'd insert a line break here for readability.

source/browser.lisp
```diff
@@ -1128,22 +1137,31 @@ Example:
...
-                (when new-url (setf (url request-data) new-url) request-data))
-              request-data)))
-    :name name))
+   #'(lambda (request-data)
+       (let ((url (url request-data)))
+         (if (funcall-safely test url)
+             (etypecase action
+               (function
+                (let* ((new-url (funcall-safely action url)))
+                  (log:info "Applied \"~a\" URL-dispatcher on \"~a\" and got \"~a\""
```
Ambrevar, line 1149: Seem comment below about `~s` instead of `~a`.

source/browser.lisp
```diff
@@ -1103,14 +1103,23 @@ Deal with REQUEST-DATA with the following rules:
...
+                           &optional (function (quri:uri) string)))
+                url-dispatching-handler))
+(defun url-dispatching-handler (name test action &optional (url-formatter #'object-string))
   "Return a `resource' handler that, if `add-hook'ed to the `request-resource-hook',
 will automatically apply its ACTION on the URLs that conform to TEST.
 
-TEST and ACTION should be functions of one argument, the requested URL.
-In case ACTION returns nil, URL request is aborted.
+TEST should be function of one argument, the requested URL.
+ACTION can be either string with a shell command, or function taking URL as argument.
```
Ambrevar, line 1116: "ACTION can be either a shell command as a string"
"or a function taking a/the URL as argument"

source/browser.lisp
```diff
@@ -1103,14 +1103,23 @@ Deal with REQUEST-DATA with the following rules:
...
+In case ACTION returns nil (always the case for shell command), URL request is aborted.
 The new URL returned by ACTION is loaded otherwise.
 
+URL-FORMATTER can be used to format url before passing it to the shell-command.
+It should take URL and return the formatted one as string.
```
Ambrevar, line 1121: Can be shortened.  What about
`URL-FORMATTER takes the URL and returns the URL string to be passed to the shell command`

I think this formatter needs an example, I don't see a use case.

aartaka: Well, one of the use cases is in the `emacs-file` dispatcher in the pull-request description. `#'quri:uri-path` is used there to get only the path from url. 

But I agree that docstring should have more examples.

Ambrevar:
Is there a specific reason why you sharp-quote the lambdas?  It seems that it's not needed in any modern compilers.
https://stackoverflow.com/questions/29338004/why-sharp-quote-lambda-expressions

About the url-formatter: isn't it simpler to use a lambda then?

Compare:

```lisp
(url-dispatching-handler
 'emacs-file
 (match-scheme "emacs")
 "emacs ~a"
 #'quri:uri-path)

(url-dispatching-handler
 'emacs-file
 (match-scheme "emacs")
 (lambda (url)
   (uiop:launch-program `("emacs" ,(quri:uri-path url)))
   nil))

```

The url-formatter version is a bit shorter, but really not much, at the cost of an extra argument, which means more complex specifications.

Worse case, should we use the url-formatter, I would use a keyword argument to help with readability.

Ambrevar:
About the Emacs example: Why pass the uri-path to Emacs?  I'm not sure I understand the example.

aartaka:
> About the Emacs example: Why pass the uri-path to Emacs? I'm not sure I understand the example.

To open a local file in emacs :) It's just a variation of the `file:` scheme with a certain program. 

But, again, I had no use-cases for this syntax, and this example was the best I could come up with. I'll be grateful if you would provide a better use-case :)

Ambrevar:
> To open a local file in emacs :) It's just a variation of the `file:` scheme with a certain program. 

This would only make sense with a file:// URL, right?  If so, I think
the predicate should match the file scheme as well :)

Then we've got a good example!

To be honest, I can't think of any other example and this one is really
not too verbose with a lambda, so I'm still a bit reluctant to add this keyword.

aartaka:
> Is there a specific reason why you sharp-quote the lambdas? It seems that it's not needed in any modern compilers.
> https://stackoverflow.com/questions/29338004/why-sharp-quote-lambda-expressions

Well, it's not needed but it shows that I want exactly a `function` object. Just a namespace marking)

> About the url-formatter: isn't it simpler to use a lambda then?

Yes, that's what I was saying about the non-trivial shell-command calls, when you've (suggested)[https://github.com/atlas-engineer/nyxt/issues/817#issuecomment-652997768] to use transformer function (that's where the `url-formatter` idea comes from).

> The url-formatter version is a bit shorter, but really not much, at the cost of an extra argument, which means more complex specifications.

I agree there. Let's try to live without it -- we can always add it back if there will be a solid use-case.

Ambrevar:
Looking good, thanks a lot!

PR #843, opened 2020-07-11
Serialize quri:uri as string in session.

Ambrevar:


aartaka:
There's a `#'str:emptyp` remaining in the `previous-history-urls-suggestion-filter` (certificate-whitelist-mode.lisp) that won't work for `quri:uri`. Please, replace it with `#'url-empty-p`, while you're at it.

Ambrevar:
Good catch!

Fixed in 780837c28815b364fe413e2aa7171a6d59b186d4.

Ambrevar:
@jmercouris OK to merge?

Note: last commit is missing from this pull request.

jmercouris:
Oui, merci beaucoup!

Issue #838, opened 2020-07-09
Add a `web-buffer` class


Ambrevar:
There is some inconsistency throughout the code base when it comes to how we initialize buffers.

- Should buffers always have an ID?  I'd say no, replacement / temp buffers don't need one.  Otherwise it would seem to the user that buffer ID numbers would often get skipped, which seems like a bug.  After all, temp buffers are mostly an implementation detail: they serve as a place holder for the webview.

- In `window-set-active-buffer` we generate temporary buffers with `buffer-make`: we should not do this since this calls the `buffer-make-hook`, initializes modes, etc.  Either we use `(make-instance *buffer-class*)` or a new `make-temp-buffer` function.

- If we choose the latter, we need to update the call to `(make-instance *buffer-class*)` in the gtk-window :after method.

- We should stick to one term, so which one should it be?  `temp`, `replacement`, `dummy`, `void`, `placeholder`?

- We need a predicate that distinguishes a temp buffer from a non-temp buffer.  I suggest `reified-buffer-p`.  If temp buffers don't have an ID, then this checks suffices.

Thoughts?

EDIT: Mention "placeholder".

Ambrevar:
Furthermore, the `load-status` conflates 2 concepts: the load status and the type of buffers.

We can have the following types of buffers:

- Live buffers (has a web page loaded with a URL).
- Local buffers (e.g. help pages, it's loaded but has no URL).
- Dead buffers (has history, URL, etc. but not loaded).
- Placeholder buffers (for window creation plus temp buffer when switching
  buffers).

The distinction between them can be done in various ways.

- We can have a subclass for each one of them.
- Placeholder buffers could be identified by their lack of ID.
  Dead buffers have no ID either, but that's not a problem because an ID is re-attributed as soon as their are resurected.

Ambrevar:
I'm leaning towards adding a `web-buffer` class and we could move the following slots there:

- `load-status`
- `request-resource-scheme`
- `request-resource-hook`
- `default-new-buffer-url`
- `cookies-path`
- `proxy`
- `certificate-whitelist`
-  `default-cookie-policy`

This would make a lot of sense with regard to how we decided to separate base-mode from web-mode.
The `buffer` default modes would be `base-mode` while web-buffers would have `web-mode base-mode...`.

jmercouris:
I agree with your reasoning and conclusion! very valid points!

jmercouris:
to be clear, I mean your latest comment about the new classes

[...]

PR #846, opened 2020-07-12
Add auto-mode for automatic mode white- and blacklisting

### LINKED ISSUE #376: Automatic proxy/tor/noscript/adblocker black- and whitelisting

aartaka:
This adds a new mode that remembers the current buffer mode setup for given URLs and recovers it when the matching URL is visited.

### Motivation

Having such a mode can simplify a routine of enabling/disabling some modes (e.g., `proxy-mode`) on the websites where you need them, while not needing these modes in other places. This mode then resembles the idea of `default-modes` and make it more flexible with the ability to define the included/excluded (i.e. whitelisted/blacklisted) modes for even a small subdomain/directory of a website!

My personal pain that this mode solves is the necessity to constantly toggle proxy-mode when I need to access some academic/software-related resources blocked in Russia. Including the `proxy-mode` to the newly introduced `auto-mode-list` solves this problem and proxy gets enabled at the moment I request the addresses of the blocked websites without any additional action on my side.

### What's New

- `enable-mode-hook` and `disable-mode-hook` `buffer` slots, whose appearance was necessary for one of the use scenarios of this mode.
- `auto-mode-list` and `auto-mode-list-data-path` slots in `browser` that store the `auto-model-list` itself and the path to its serialized version respectively.
- `auto-mode` mode and two main commands of this mode `include-modes-to-auto-mode-list` and `disable-and-exclude-modes-from-auto-mode-list` that help in manipulation on the white- and blacklisted modes and seem sufficient to me (and, probably, that's not the case for everyone, so your feedback on user experience can help me a lot!).
- `*prompt-on-mode-toggle*` and `*non-rememberable-modes*` special variables that change the behavior of auto-mode:
    - `*prompt-on-mode-toggle*` defines whether `auto-mode` asks the user if they want to permanently enable/disable the mode they do enable/disable, for the URL they are visiting. That's what new hooks were necessary for!
    - `*non-rememberable-modes*` lists modes that (possibly) no user would want to white- or blacklist. There are `web-mode` that is planned to transform to `web-buffer` soon (#838); `help-mode`, whitelisting which could break Nyxt behavior given that most buffers using `help-mode` are inner and URL-less ones; and `auto-mode` itself, because it is illogical to put the `auto-mode-list`-moderating mode into `auto-mode-list` and it can cause some bugs or illogical behaviors.
- A lot of helper functions that belong to `auto-mode` and are not intended to be exported.
- auto-mode-list.lisp data-storing file intended to be easily readable and editable.
- Documentation to this mode, scattered around the `include-modes-to-auto-mode-list`, auto-mode-list.lisp and several other functions and classes.

### Use scenarios

We had a discussion on use experience with this mode is #376 with @Ambrevar. As a consequence, there are two use scenarios that I made way for:

#### Manual and undisturbing

This is the default scenario, because the `*prompt-on-mode-toggle*` variable is `nil` by default. This mode means that user manages modes themselves and adds them to the `auto-mode-list` only when there's a necessity. These additions are done through the introduced commands (`include-modes-to-auto-mode-list` and `disable-and-exclude-modes-from-auto-mode-list`) only.

#### "Smart" and annoying

This needs `*prompt-on-mode-toggle*` set to `t`. In this use scenario, on every not `non-remembered-mode` toggle on the website that is not in `auto-mode-list` yet, user is asked `"Permanently (enable|disable) %MODE-NAME for this URL?"`. Mode is included in either `included-modes` or `excluded-modes` (on mode enable/disable respectively) of suitable `auto-mode-list` entry on user consent then. The commands of the `auto-mode` are accessible in this scenario too, but they are used less because of these "smart" prompts.

### How has this been tested

- I've added `auto-mode` to the `default-modes` and ran Nyxt. There are 8 entries in my auto-mode-list.lisp now, as a result of testing, and no problems were noticed with `auto-mode` itself, but there were difficulties with other modes:
    - `proxy-mode` starts working only after page reload, but that's not a bug, that's a natural behavior of this mode :)
    - `reading-line-mode` needs toggling after you `follow-hint` the link with `reading-line-mode` enabled. It causes lots of JS errors like this otherwise:
```
<WARN> [22:27:03] Warning: JavaScript error: GError: Domain: "WebKitJavascriptError", Code: 699, Message: https://habr.com/ru/post/510588/:1:47: TypeError: null is not an object (evaluating 'document.querySelector('#reading-line-cursor').scrollIntoViewIfNeeded')
```

I'll be glad to see your feedback on this and your ideas on how to enhance the user experience with this mode!

Closes #376

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -1,0 +1,306 @@
+(uiop:define-package :nyxt/auto-mode
+  (:use :common-lisp :nyxt)
+  (:import-from #:serapeum #:export-always)
+  (:documentation "Mode for automatic URL-based mode toggling."))
+(in-package :nyxt/auto-mode)
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (trivial-package-local-nicknames:add-package-local-nickname :alex :alexandria))
+
+(export-always '*prompt-on-mode-toggle*)
+(defparameter *prompt-on-mode-toggle* nil
+  "Whether user will be asked about adding the mode to included/excluded modes
+in auto-mode-list on mode activation/deactivation.")
+
+(export-always '*non-rememberable-modes*)
+(defparameter *non-rememberable-modes*
+  ;; Base mode conflicts with its Nyxt symbol if it's not prefixed
+  '(help-mode web-mode auto-mode nyxt::base-mode)
+  "Modes that AUTO-MODE won't even try to save.")
+
+(declaim (ftype (function ((or symbol root-mode)) (values symbol &optional)) maybe-mode-name))
+(defun maybe-mode-name (mode)
+  (if (symbolp mode) mode (mode-name mode)))
+
+(declaim (ftype (function ((or symbol root-mode) (or symbol root-mode)) boolean)
+                mode-equal))
+(defun mode-equal (mode1 mode2)
+  (string-equal (symbol-name (maybe-mode-name mode1))
+                (symbol-name (maybe-mode-name mode2))))
+
+(defun non-rememberable-mode-p (mode)
+  (member mode *non-rememberable-modes* :test #'mode-equal))
+
+(defun rememberable-of (modes)
+  (remove-if #'non-rememberable-mode-p modes))
+
+(defclass-export auto-mode-rule ()
+  ((test :initarg :test
+         :accessor test
+         :type list
+         :initform (error "Slot `test' should be set."))
+   (modes :initarg :modes
+          :accessor modes
+          :type list-of-symbols
+          :initform '())))
+
+(declaim (ftype (function (quri:uri) (or auto-mode-rule null))
+                matching-auto-mode-rule))
+(defun matching-auto-mode-rule (url)
+  (flet ((priority (test1 test2)
+           (let ((priority-list '(match-regex match-url match-host match-domain)))
+             (< (or (position (first test1) priority-list) 4)
+                (or (position (first test2) priority-list) 4)))))
+    (first (sort (remove-if-not
+                  #'(lambda (rule)
+                      (funcall-safely
+                       (apply (first (test rule)) (rest (test rule))) url))
+                  (or (auto-mode-list *browser*) (restore-auto-mode-list)))
+                 #'priority :key #'test))))
+
+(declaim (ftype (function (quri:uri buffer)) enable-matching-modes))
+(defun enable-matching-modes (url buffer)
+  (let ((rule (matching-auto-mode-rule url)))
+    (enable-modes (rememberable-of (set-difference
+                                    (modes rule)
+                                    (mapcar #'mode-name (modes buffer))
+                                    :test #'mode-equal))
+                  buffer)
+    (disable-modes (rememberable-of (set-difference
+                                     (mapcar #'mode-name (modes buffer))
+                                     (modes rule) :test #'mode-equal))
+                   buffer)))
+
+(defun clean-up-auto-mode (mode)
+  (dolist (handler-name '(auto-mode-request-handler
+                          enable-mode-prompting-handler
+                          disable-mode-prompting-handler))
+    (hooks:remove-hook (request-resource-hook (buffer mode)) handler-name)))
+
+(defun store-last-active-modes (auto-mode)
+  (setf (last-active-modes auto-mode)
+        (modes (buffer auto-mode))))
+
+(defun restore-last-active-modes (auto-mode)
+  (disable-modes
+   (mapcar #'maybe-mode-name
+           (set-difference (modes (buffer auto-mode))
+                           (last-active-modes auto-mode)
+                           :test #'mode-equal))
+   (buffer auto-mode))
+  (enable-modes
+   (mapcar #'maybe-mode-name
+           (set-difference (last-active-modes auto-mode)
+                           (modes (buffer auto-mode))
+                           :test #'mode-equal))
+   (buffer auto-mode)))
+
+(defun auto-mode-request-handler (request-data)
+  (let* ((auto-mode (find-submode (buffer request-data) 'auto-mode))
+         ;; We need to supress prompting when auto-mode modifies modes.
+         (*prompt-on-mode-toggle* nil)
+         (web-mode (find-submode (buffer request-data) 'web-mode))
+         (prev-url
+           (unless (eq (htree:root (history web-mode))
+                       (htree:current (history web-mode)))
+             (url (htree:data
+                   (htree:parent (htree:current (history web-mode))))))))
+    (if (matching-auto-mode-rule (url request-data))
+        (progn
+          (when (and prev-url (not (matching-auto-mode-rule prev-url)))
+            (store-last-active-modes auto-mode))
+          (enable-matching-modes (url request-data) (buffer request-data)))
+        ;; Apply previous saved modes only on buffer-loads, not anytime else.
+        ;;
+        ;; TODO: Relies on somewhat illogical behavior of (url (current-buffer))
+        ;; and (url request-data) being equal after buffer-load is called.
+        ;; Replace with more stable and logical condition?
+        (when (quri:uri= (url request-data) (url (buffer auto-mode)))
+          (restore-last-active-modes auto-mode))))
+  request-data)
+
+(defun mode-covered-by-auto-mode-p (mode auto-mode-instance)
+  (or (non-rememberable-mode-p mode)
+      (let ((matching-rule (matching-auto-mode-rule
+                            (url (buffer auto-mode-instance)))))
+        (member mode (or (and matching-rule (modes matching-rule))
+                         (last-active-modes auto-mode-instance))
+                :test #'mode-equal))))
+
+(declaim (ftype (function (boolean t) (function (root-mode)))
+                make-mode-toggle-auto-mode-handler))
+(defun make-mode-toggle-prompting-handler (enable-p auto-mode-instance)
+  #'(lambda (mode)
+      (when (and (not (mode-covered-by-auto-mode-p mode auto-mode-instance))
+                 *prompt-on-mode-toggle*)
+        (nyxt::with-confirm ("Permanently ~a ~a for this URL?"
+                             (if enable-p "enable" "disable")
+                             (mode-name mode))
+          (with-result (url (read-from-minibuffer
+                             (make-minibuffer
+                              :input-prompt "URL:"
+                              :input-buffer (object-string (url (buffer mode)))
+                              :must-match-p nil)))
+            (let* ((test (make-dwim-match url))
+                   (rule (find test (auto-mode-list *browser*)
+                               :key #'test :test #'equal))
+                   (rule-modes (if rule (modes rule) (modes (buffer mode))))
+                   (modes (if enable-p
+                              (union (list mode) rule-modes :test #'mode-equal)
+                              (set-difference rule-modes (list mode)
+                                              :test #'mode-equal))))
+              (add-modes-to-auto-mode-list test modes)))))))
+
+(defun initialize-auto-mode (mode)
+  (unless (last-active-modes mode)
+    (setf (last-active-modes mode)
+          (default-modes (buffer mode))))
+  (hooks:add-hook (enable-mode-hook (buffer mode))
+                  (nyxt::make-handler-mode
+                   (make-mode-toggle-prompting-handler t mode)
+                   :name 'enable-mode-auto-mode-handler))
+  (hooks:add-hook (disable-mode-hook (buffer mode))
+                  (nyxt::make-handler-mode
+                   (make-mode-toggle-prompting-handler nil mode)
+                   :name 'disable-mode-auto-mode-handler))
+  (hooks:add-hook (request-resource-hook (buffer mode))
+                  (make-handler-resource #'auto-mode-request-handler)))
+
+(define-mode auto-mode ()
+  "Remember the modes setup for given domain/host/URL and store it in an editable form.
+These modes will then be activated on every visit to this domain/host/URL."
+  ((last-active-modes :accessor last-active-modes
+                      :type list
+                      :initform '())
+   (destructor :initform #'clean-up-auto-mode)
+   (constructor :initform #'initialize-auto-mode)))
+
+(declaim (ftype (function (string) list) make-dwim-match))
+(defun make-dwim-match (url)
+  (let ((url (or (ignore-errors (quri:uri url)) url)))
+    (if (and (quri:uri-p url)
+             (empty-path-url-p url)
+             (host-only-url-p url))
+        (if (string= (quri:uri-domain url)
+                     (cl-ppcre:regex-replace "www[0-9]?\\." (quri:uri-host url) ""))
+            `(match-domain ,(quri:uri-domain url))
+            `(match-host ,(quri:uri-host url)))
+        `(match-url ,(object-display url)))))
+
+(define-command save-modes-to-auto-mode-list ()
+  "Store the enabled modes to auto-mode-list for all the future visits of this
+domain/host/URL/group of websites inferring the suitable matching condition by user input.
+The rules are:
+- If it's a plain domain-only URL (i.e. \"http://domain.com\") -- then use `match-domain'.
+- If it's host with subdomains (\"http://whatever.subdomain.domain.com\") -- use `match-host'.
+- Use `match-url' otherwise.
+
+For the storage format see the comment in the head of your `auto-mode-list-data-path' file."
+  (if (find-submode (current-buffer) 'auto-mode)
+      (with-result (url (read-from-minibuffer
+                         (make-minibuffer
+                          :input-prompt "URL:"
+                          :input-buffer (object-string (url (current-buffer)))
+                          :suggestion-function (nyxt::history-suggestion-filter
+                                                :prefix-urls (list
+                                                              (object-string
+                                                               (url (current-buffer)))))
+                          :history (minibuffer-set-url-history *browser*)
+                          :must-match-p nil)))
+        (when (typep url 'nyxt::history-entry)
+          (setf url (url url)))
+        (add-modes-to-auto-mode-list (make-dwim-match url)
+                                     (modes (current-buffer))))
+      (echo "Enable auto-mode first.")))
+
+(declaim (ftype (function (list (or null list)) (values list &optional))
+                add-modes-to-auto-mode-list))
+(defun add-modes-to-auto-mode-list (test modes)
+  (let ((rule (or (find test (auto-mode-list *browser*) :key #'test :test #'equal)
+                  (make-instance 'auto-mode-rule :test test))))
+    (when modes
+      (setf (modes rule) (rememberable-of (mapcar #'maybe-mode-name modes))
+            (auto-mode-list *browser*) (remove-duplicates
+                                        (push rule (auto-mode-list *browser*))
+                                        :key #'test :test #'equal))
+      (store-auto-mode-list)
+      (auto-mode-list *browser*))))
+
+(defmethod serialize-object ((rule auto-mode-rule) stream)
+  (let ((*standard-output* stream)
+        (*print-case* :downcase))
+    (write-string "(" stream)
+    (if (and (eq (first (test rule)) 'match-url)
+             (= 2 (length (test rule))))
+        (format t "~s " (second (test rule)))
+        (format t "~s " (test rule)))
+    (format t ":MODES ~a"
+            (mapcar (alex:compose #'string-downcase #'symbol-name)
+                    (modes rule)))
+    (write-string ")" stream)))
+
+(defmethod deserialize-auto-mode-rules (stream)
+  (handler-case
+      (let ((*standard-input* stream))
+        (let ((rules (read stream)))
+          (mapcar #'(lambda (rule)
+                      (let ((rule (append '(:test) rule)))
+                        (when (stringp (getf rule :test))
+                          (setf (getf rule :test) `(match-url ,(getf rule :test))))
+                        (apply #'make-instance 'auto-mode-rule rule)))
+                  rules)))
+    (error (c)
+      (log:error "During auto-mode rules deserialization: ~a" c)
+      nil)))
+
+(defun store-auto-mode-list ()
+  (with-data-file (file (auto-mode-list-data-path *browser*)
+                        :direction :output
+                        :if-does-not-exist :create
+                        :if-exists :supersede)
+    (write-string ";; List of auto-mode rules.
+;; It is made to be easily readable and editable, but you still need to remember some things:
+;;
+;; Every rule starts on a new line and consists of two parts:
+;; - Condition for rule activation. It is either (match-domain ...),
+;;   (match-host ...), (match-regex ...) or a string.
+;; - :MODES -- List of the modes to enable on condition. Edit carefully:
+;;   all the modes except these and nyxt/auto-mode:*non-rememberable-modes*
+;;   will be disabled on matching URL.
+;;
+;; Conditions work this way:
+;; - match-domain matches the URL domains only.
+;;   Example: (match-domain \"reddit.com\") will work for all of Reddit.
+;; - match-host is more specific -- it activates only on certain subdomains of the website.
+;;   Example: (match-host \"old.reddit.com\") will work on old Reddit only.
+;; - match-regex works for any address that matches a given regex. You can add these manually,
+;;   but remember: with great regex comes great responsibility!
+;;   Example: \"https://github.com/.*/.*\" will activate only in repos on GitHub.
+;; - String format matches the exact URL and nothing else
+;;   Example: \"https://lispcookbook.github.io/cl-cookbook/pattern_matching.html\"
+;;            will work on the Pattern Matching article of CL Cookbook, and nowhere else.
+;;
+;; You can write additional URLs in the bracketed conditions, to reuse the rule for other URL
+;; Example: (match-host \"reddit.com\" \"old.reddit.com\" \"www6.reddit.com\")
+" file)
+    (write-string "(" file)
+    (dolist (rule (slot-value *browser* 'auto-mode-list))
+      (write-char #\newline file)
+      (serialize-object rule file))
+    (format file "~%)~%")
+    (echo "Saved ~a auto-mode rules to ~s."
+          (length (slot-value *browser* 'auto-mode-list))
+          (expand-path (auto-mode-list-data-path *browser*)))))
+
+(defun restore-auto-mode-list ()
+  (handler-case
+      (let ((data (with-data-file (file (auto-mode-list-data-path *browser*)
+                                        :direction :input
+                                        :if-does-not-exist nil)
+                    (when file (deserialize-auto-mode-rules file)))))
+        (when data
+          (echo "Loading ~a auto-mode rules from ~s."
+                (length data) (expand-path (auto-mode-list-data-path *browser*)))
+          (setf (slot-value *browser* 'auto-mode-list) data)))
+    (error (c)
+      (echo-warning "Failed to load auto-mode-list from ~s: ~a"
+                    (expand-path (auto-mode-list-data-path *browser*)) c))))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -200,6 +200,8 @@ Without handler, return ARG.  This is an acceptable `combination' for
      load-status
      modes
      default-modes
+     enable-mode-hook
+     disable-mode-hook
      keymap-scheme-name
      current-keymaps-hook
      override-map
@@ -256,6 +258,18 @@ have an empty ID.")
                   :initform '(certificate-whitelist-mode web-mode base-mode)
                   :type list-of-symbols
                   :documentation "The symbols of the classes to instantiate on buffer creation.")
+   (enable-mode-hook :accessor enable-mode-hook
+                     :initarg :enable-mode-hook
+                     :initform (make-hook-mode)
+                     :type hook-mode
+                     :documentation "Hook run on every mode activation,
+after the mode-specific hook.")
+   (disable-mode-hook :accessor disable-mode-hook
+                      :initarg :disable-mode-hook
+                      :initform (make-hook-mode)
+                      :type hook-mode
+                      :documentation "Hook run on every mode deactivation,
+after the mode-specific hook.")
    (keymap-scheme-name
     :accessor keymap-scheme-name
     :initarg :keymap-scheme-name
@@ -559,6 +573,8 @@ the empty string.")))
             session-store-function
             session-restore-function
             session-restore-prompt
+            auto-mode-list
+            auto-mode-list-data-path
             standard-output-path
             error-output-path
             before-exit-hook
@@ -756,6 +772,14 @@ from `session-path'.")
                            :initform :always-ask
                            :documentation "Ask whether to restore the
 session. Possible values are :always-ask :always-restore :never-restore.")
+   (auto-mode-list :accessor auto-mode-list
+                   :type list
+                   :initform '()
+                   :documentation "The auto-mode rules kept in memory.")
+   (auto-mode-list-data-path :accessor auto-mode-list-data-path
+                             :type data-path
+                             :initform (make-instance 'auto-mode-list-data-path :basename "auto-mode-list")
+                 :documentation "The path where the auto-mode rules are saved.")
    (standard-output-path :accessor standard-output-path
                          :type data-path
                          :initform (make-instance 'data-path :basename "standard-out.txt")
@@ -1153,7 +1177,8 @@ The following example does a few things:
    :name name))
 
 (declaim (ftype (function (string &rest string) (function (quri:uri) boolean))
-                match-scheme match-host match-domain match-file-extension))
+                match-scheme match-host match-domain
+                match-file-extension match-regex match-url))
 (export-always 'match-scheme)
 (defun match-scheme (scheme &rest other-schemes)
   "Return a predicate for URLs matching one of SCHEME or OTHER-SCHEMES."
@@ -1181,6 +1206,20 @@ The following example does a few things:
   #'(lambda (url)
       (some (alex:curry #'string= (pathname-type (or (quri:uri-path url) "")))
             (cons extension other-extensions))))
+
+(export-always 'match-regex)
+(defun match-regex (regex &rest other-regex)
+  "Return a predicate for URLs matching one of REGES or OTHER-REGEX."
+  #'(lambda (url)
+      (some (alex:rcurry #'cl-ppcre:scan (object-display url))
+            (cons regex other-regex))))
+
+(export-always 'match-url)
+(defun match-url (one-url &rest other-urls)
+  "Return a predicate for URLs exactly matching ONE-URL or OTHER-URLS."
+  #'(lambda (url)
+      (some (alex:rcurry #'string= (object-display url))
+            (cons one-url other-urls))))
 
 (defun javascript-error-handler (condition)
   (echo-warning "JavaScript error: ~a" condition))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -421,11 +421,31 @@ The current buffer access time is set to be the last."
                               oldest-buffer)))
     (set-current-buffer oldest-buffer)))
 
+(export-always 'mode-name)
+(defun mode-name (mode)
+  (class-name (class-of mode)))
+
+(declaim (ftype (function (list-of-symbols &optional buffer)) disable-modes enable-modes))
+(export-always 'disable-modes)
+(defun disable-modes (modes &optional (buffer (current-buffer)))
+  "Disable MODES for BUFFER.
+MODES should be a list symbols, each possibly returned by `mode-name'."
+  (dolist (mode modes)
+    (funcall-safely (sym (mode-command mode))
+                    :buffer buffer :activate nil)))
+
+(export-always 'enable-modes)
+(defun enable-modes (modes &optional (buffer (current-buffer)))
+  "Enable MODES for BUFFER.
+MODES should be a list of symbols, each possibly returned by `mode-name'."
+  (dolist (mode modes)
+    (funcall-safely (sym (mode-command mode))
+                    :buffer buffer :activate t)))
+
 (defun active-mode-suggestion-filter (buffers)
   "Return the union of the active modes in BUFFERS."
-  (let ((modes (delete-duplicates (mapcar (lambda (m)
-                                            (class-name (class-of m)))
-                                          (apply #'append (mapcar #'modes buffers))))))
+  (let ((modes (delete-duplicates (mapcar #'mode-name
+                                          (alex:mappend #'modes buffers)))))
     (lambda (minibuffer)
       (fuzzy-match (input-buffer minibuffer) modes))))
 
@@ -438,8 +458,7 @@ The current buffer access time is set to be the last."
                     (mode-list)))
         (common-modes (reduce #'intersection
                               (mapcar (lambda (b)
-                                        (mapcar (lambda (m) (class-name (class-of m)))
-                                                (modes b)))
+                                        (mapcar #'mode-name (modes b)))
                                       buffers))))
     (lambda (minibuffer)
       (fuzzy-match (input-buffer minibuffer) (set-difference all-non-minibuffer-modes common-modes)))))
@@ -452,9 +471,7 @@ The current buffer access time is set to be the last."
                         :multi-selection-p t
                         :suggestion-function (active-mode-suggestion-filter buffers))))
     (dolist (buffer buffers)
-      (dolist (mode modes)
-        (funcall (sym (mode-command mode))
-                 :buffer buffer :activate nil)))))
+      (disable-modes modes buffer))))
 
 (define-command disable-mode-for-buffer ()
   "Disable queried mode(s) for select buffer(s)."
@@ -473,9 +490,7 @@ The current buffer access time is set to be the last."
                         :multi-selection-p t
                         :suggestion-function (inactive-mode-suggestion-filter buffers))))
     (dolist (buffer buffers)
-      (dolist (mode modes)
-        (funcall (sym (mode-command mode))
-                 :buffer buffer :activate t)))))
+      (enable-modes modes buffer))))
 
 (define-command enable-mode-for-buffer ()
   "Enable queried mode(s) for select buffer(s)."
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -72,6 +72,7 @@
                (:file "bookmark")
                (:file "history")
                (:file "autofill")
+               (:file "auto-mode")
                (:file "external-editor")
                (:file "url-group")
                ;; Core Modes
--- a/source/data-storage.lisp
+++ b/source/data-storage.lisp
@@ -71,6 +71,8 @@ This can be used to set the path from command line.  See
   ((ref :initform "history")))
 (defclass-export download-data-path (data-path) ; TODO: Rename to downloads-data-path?
   ((ref :initform "download")))
+(defclass-export auto-mode-list-data-path (data-path)
+  ((ref :initform "auto-mode-list")))
 
 (declaim (ftype (function (string) (or string null)) find-ref-path))
 (defun find-ref-path (ref)
--- a/source/mode.lisp
+++ b/source/mode.lisp
@@ -72,11 +72,13 @@ Example:
                        (when (constructor new-mode)
                          (funcall-safely (constructor new-mode) new-mode))
                        (push new-mode (modes buffer))
-                       (hooks:run-hook (enable-hook new-mode) new-mode))
+                       (hooks:run-hook (enable-hook new-mode) new-mode)
+                       (hooks:run-hook (enable-mode-hook buffer) new-mode))
                      (print-status)
                      (log:debug "~a enabled." ',name))
                    (when existing-instance
                      (hooks:run-hook (disable-hook existing-instance) existing-instance)
+                     (hooks:run-hook (disable-mode-hook buffer) existing-instance)
                      (when (destructor existing-instance)
                        (funcall-safely (destructor existing-instance) existing-instance))
                      (setf (modes buffer) (delete existing-instance
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -63,6 +63,20 @@ If it cannot be derived, return an empty `quri:uri'."
        (when url
          (uiop:emptyp (object-string url)))))
 
+(declaim (ftype (function (quri:uri) boolean)
+                empty-path-url-p host-only-url-p))
+(export-always 'empty-path-url-p)
+(defun empty-path-url-p (url)
+  (or (string= (quri:uri-path url) "/")
+      (null (quri:uri-path url))))
+
+(export-always 'host-only-url-p)
+(defun host-only-url-p (url)
+  (every #'null
+         (list (quri:uri-query url)
+               (quri:uri-fragment url)
+               (quri:uri-userinfo url))))
+
 (declaim (ftype (function (quri:uri quri:uri) boolean) schemeless-uri=))
 (defun schemeless-uri= (uri1 uri2)
   "Like `quri:uri=' but ignore scheme in comparison.


Ambrevar:
> There are web-mode that is planned to transform to web-buffer soon

Err, no, to clarify we would like to create a new buffer class for web buffers, but we are not going to touch web-mode :)

aartaka:
I misunderstood it then, sorry. However, should `web-mode` be saveable by `auto-mode`?

source/browser.lisp
```diff
@@ -742,6 +756,16 @@ from `session-path'.")
...
+   (auto-mode-list :accessor auto-mode-list
+                   :type list
+                   :initform nil
+                   :documentation "
+The auto-mode mode list kept in memory.")
+   (auto-mode-list-data-path :accessor auto-mode-list-data-path
+                             :type data-path
+                             :initform (make-instance 'auto-mode-list-data-path :basename "auto-mode-list")
+                 :documentation "
+The path where the system will create/save the auto-mode mode lists.")
```
Ambrevar, line 768: Same.

"create/save" -> "persist", or just "save".

"auto-mode mode lists" may look like a typo to the user :p  I suggest "audo-mode entries" or "auto-mode rules".  What about using the term "rules" systematically in this change?

"the system" is a generic word that's not too meaningful.  Maybe use the passive form instead" The path where to save the auto-mode lists.

source/auto-mode.lisp
```diff
@@ -0,0 +1,301 @@
...
+     (url-dispatching-handler
+      (name ,entry)
+      (apply (first test) (rest test))
+      #'(lambda (url)
+          (dolist (mode (included-modes ,entry))
+            (funcall-safely (nyxt::sym (nyxt::mode-command mode))
+                            :buffer (current-buffer) :activate t))
+          (dolist (mode (excluded-modes ,entry))
+            (funcall-safely (nyxt::sym (nyxt::mode-command mode))
+                            :buffer (current-buffer) :activate nil))
```
Ambrevar, line 61: Don't we want to exclude some modes before including the others?  This way the user can set the exact modes they want to a URL.

This is only possible if we have a way to specify "all modes" (without listing them all of course).  I think this feature is missing in this patch.

One option would be to match the modes against a regexp.

Ambrevar:
Wow, what a contribution!  This is very impressive! :)

I only had a cursory look at the patch, I'll test more in-depth later.
For now, I've only spotted some nits, everything else looks perfect except for one thing: see the comment about the activation order and the possibility to match modes with regexp.

Ambrevar:

> I misunderstood it then, sorry. However, should `web-mode` be saveable by `auto-mode`? 

Hmmm... I'd say no, because for now every buffer should have it.
In the future we may have buffers without web-mode, but for them it
would probably not make sense to ever enable web-mdoe either.

So I think web-mode has effectively the same status as base-mode.

source/auto-mode.lisp
```diff
@@ -0,0 +1,301 @@
...
+
+(defclass-export auto-mode-entry ()
+  ((test :initarg :test
+         :accessor test
+         :type list
+         :initform (error "slot `test' should be set!"))
+   (included-modes :initarg :included-modes
+                   :accessor included-modes
+                   :type list-of-symbols
+                   :initform nil)
```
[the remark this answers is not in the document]
jmercouris: Alternative viewpoint, I prefer to write (list) :-D

source/auto-mode.lisp
```diff
@@ -0,0 +1,301 @@
...
+(defun make-dwim-match (url)
+  (let ((url (or (ignore-errors (quri:uri url)) url)))
+    (if (and (quri:uri-p url)
+             (or (string= (quri:uri-path url) "/") (null (quri:uri-path url)))
+             (every #'null
+                    (list (quri:uri-query url)
+                          (quri:uri-fragment url)
+                          (quri:uri-userinfo url))))
+        (if (string= (quri:uri-domain url)
+                     (str:replace-first "www." "" (quri:uri-host url)))
```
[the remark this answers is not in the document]
aartaka: How about matching against regexp? Something like `"www[0-9]?\."` would do the job there, if the variations are only numbers after "www".

It's a matching condition that is made to work for most cases in an intuitive manner, and in most cases "www" doesn't matter much. The exceptions can be covered by `match-regex`.

source/auto-mode.lisp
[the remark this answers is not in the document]
aartaka: And I prefer `nil` because it seemed unambiguous to me that `nil` is an empty list. But we need to stick to one style in the end, so I'd go with `'()` (even though it reminds me of Scheme more that CL) :)

source/auto-mode.lisp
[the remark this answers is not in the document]
Ambrevar: Maybe that would work.  To be honest I'm unsure here.  Well, it's a rather low-level detail that can be tweaked later if need be, so this is definitely not a blocker.

source/auto-mode.lisp
aartaka: > Don't we want to exclude some modes before including the others? This way the user can set the exact modes they want to a URL.

Well, exact mode setup is possible now, because when mode is excluded, it is removed from the `included-modes` automatically and vice versa.

> This is only possible if we have a way to specify "all modes" (without listing them all of course). I think this feature is missing in this patch.
>
> One option would be to match the modes against a regexp.

I believe we don't need it for the same reason we don't need `match-regex` as default case fro URL-matching -- it's too easy to break things with it. It's safer to ask user directly what they mean and exclude/include the overlapping modes when necessary, based on the priority of the last user action (e.g. if I exclude `proxy-mode` now, but had it before in the `included-modes` then I probably know what I'm doing).

Having these simple checks of mode presence in `auto-mode-list` will be enough.

aartaka:
Here you go. I'll miss regex, but it's true that they are bad as a default choice.

Things that I haven't changed:
- There's still no ability to specify "all modes", because I believe the current set-based exclusion and inclusion intelligent enough.
- The bodies of the `disable-and-exclude-modes-from-auto-mode-list` and `include-modes-to-auto-mode-list` are still similar, because I'm not sure in the way of abstracting these.

And, as a side-note: there is some code duplication in the mode-related pieces. Maybe it can be worth to look through the mode.lisp and auto-mode.lisp and abstract some reoccuring things away?

jmercouris:
The way you can answer uncertain questions about code duplication is with a test: will the duplicated snippets of code change in lockstep, or are they likely to diverge? If lockstep- consolidate. If diverge- leave alone.

aartaka:
I guess it's okay for now. What else is left to do until merge?

aartaka:
Fixing the typo that prevented deserialization from happening :)

source/auto-mode.lisp
Ambrevar: To clarify: when a URL is set, we go through enable / disable modes _starting from_ the list of currently active modes, right?

Say I'm on example.org with mode FOO on.  I go to abc.xyz which only has the (:include BAR) rule.  When the page is loaded, I end up with modes FOO and BAR.
Is this correct?

If so, is there a way to say "set exactly BAR and nothing else"?

Ambrevar:
Looking great! I'll play around a little bit.
The only question to answer is whether we can set the exact set of modes for a URL.

source/auto-mode.lisp
```diff
@@ -0,0 +1,295 @@
...
+                   :type list-of-symbols
+                   :initform nil)
+   (excluded-modes :initarg :excluded-modes
+                   :accessor excluded-modes
+                   :type list-of-symbols
+                   :initform '())
+   (name :initarg :name
+         :accessor name
+         :type symbol
+         :initform (gensym))))
```
Ambrevar, line 38: I'm realizing that you never set the name.  We should not just use `gensym`: the point of the name is that it should be readable so that the user can know which handler is which when introspecting.

Another important point is that handler names must be unique in a hook.  A handler with an already existing name won't be added.  `gensym` is brittle because if it generates a name that was already used (e.g. when restoring the rules), this will prevent the handler from being added.

I suggest the following:
- Prefix the name with, say, "auto-mode-DOMAIN".
- Create a function that keeps generating a random name (say with `gensym`) until the final name does not conflict.

All this considered, I wonder if we should move the auto-mode handlers to a subhook, like you originally suggested when you implemented the url-dispatching handlers.  (Note that this is not necessary for url-dispatching handlers itself  since the handlers are not generated at this level.)  This would avoid potential collisions with other modes.  If we end up having many hooks with generated handlers, this could become necessary.
But this consideration may be premature for this change set.

jmercouris: I'm not sure I understand the statement about gensym. Gensym accepts an optional

`(gensym "salmon") ;; -> #:|salmon639|`

would that not suffice?

Ambrevar:
We can use it but we also need to make sure no handler with the same
name already exists.  It may exist because gensym generates names unique
the the current Lisp image, however it does not know about the names
that were persisted then restored.

jmercouris:
> We can use it but we also need to make sure no handler with the same name already exists. It may exist because gensym generates names unique the the current Lisp image, however it does not know about the names that were persisted then restored.

Then, could the solution be to intern names that are restored?

Ambrevar:

> Then, could the solution be to intern names that are restored?

Smart!  I think that would work indeed!

source/auto-mode.lisp
aartaka: No, it's not about the collision of new gensyms with the existyng symbols, it's about the previous session gensyms with the current session ones. 

Say, you've generated rules with name-gensyms from `#:G100` to `#:G115` in your previous session. When you start a new session, gensyms can start counting from `#:G100` again, because symbols from auto-mode rules are uninterned. If you'll happen to add a new auto-mode rule with, for example, `#:G107` in a name, then it will conflict an existing rule and the hook won't be created. 

Adding a domain sounds good, but I'd go with even more strict option -- domain plus `match-*` (e.g., `MATCH-DOMAIN-REDDIT-COM`) rule, as I tried to do this in `url-dispatching-handler` pull request.

aartaka:
> The only question to answer is whether we can set the exact set of modes for a URL.

Not yet. I'll look into that.

aartaka:
Here you go:
- Readable and interned names for the rules.
- `exact` boolean slot of the `auto-mode-rule` to set the exact mode setup.
- More readable and package-prefix-less names of the modes in the `auto-mode-list-data-path` file.
- Several helper functions.

aartaka:
Adding to that, useful mode-related functions are now moved to buffer.lisp and exported.

source/auto-mode.lisp
```diff
@@ -0,0 +1,312 @@
...
+       (url-dispatching-handler
+        (intern (symbol-name (name ,rule)))
+        (apply (first ,test) (rest ,test))
+        #'(lambda (url)
+            (when (exact ,rule)
+              (dolist (mode (remove-if
+                             #'non-rememberable-mode-p
+                             (mapcar #'mode-name (modes (current-buffer)))))
+                (when (not (non-rememberable-mode-p mode))
+                  (disable-mode mode))))
```
Ambrevar, line 64: Why the `(when ...)` here if `mode` is already rememberable?

source/auto-mode.lisp
```diff
@@ -0,0 +1,312 @@
...
+                       (name rule)))
+  (when *prompt-on-mode-toggle*
+      (hooks:remove-hook (enable-mode-hook (buffer mode))
+                         'enable-mode-prompting-handler)
+      (hooks:remove-hook (disable-mode-hook (buffer mode))
+                         'disable-mode-prompting-handler)))
+
+(defun initialize-auto-mode (mode)
+  (dolist (rule (or (auto-mode-list *browser*)
+                     (restore-auto-mode-list)))
```
Ambrevar, line 89: Nit: Indentation.

source/auto-mode.lisp
```diff
@@ -0,0 +1,312 @@
...
+;; It is made to be easily readable and editable, but you still need to remember some things.
+;;
+;; Every rule starts on a new line and consists of five parts (three are optional):
+;; - Condition for rule activation. It is either (match-domain ...),
+;;   (match-host ...), (match-regex ...) or a string.
+;; - OPTIONAL: List of the modes to enable on condition, it starts with :INCLUDED-MODES keyword.
+;; - OPTIONAL: List of the modes to disable on condition -- :EXCLUDED-MODES keyword.
+;; - OPTIONAL: Whether the included/excluded modes of the rule define exact mode
+;;   setup -- :EXACT keyword. You need to add this yourself in case you need it.
+;; - :NAME of the rule -- this arbitrary name is only used during logging.
```
Ambrevar, line 275: The items are not consistently formatted.  I'd write "(optional)" instead.  Example:

```
- :INCLUDED-MODES (optional): List of the ...
```

source/auto-mode.lisp
jmercouris: One other comment, instead of (when (not ...)) write (unless ...)

source/auto-mode.lisp
jmercouris: I usually do a `C-x C-h i` on the whole file, saves some time :-)

Ambrevar:
I've tested it, it seems to be working very well so far!
This is awesome!!!

- I'd print the the auto-mode-list in lower case, like we do with
  bookmarks.  This is meant to be edited by hand and writing in upper
  case is painful.

- I think the most "dwim" behaviour when auto-mode and the user
  navigates to a URL without rules is to restore the default modes or
  the last active modes before a rule was matches (this can be saved in
  the mode slot.)
  At the moment, the modes from the last URL remain, which does not
  sound like something we'd want.

- Actually, do we really need 2 commands to save the modes for a
  website?  It sounds like 1 would suffice if we exclude the modes that
  are not there but are present in the defaults.

Thoughts?

Ambrevar:
Are you sure about the binding? O.o

aartaka:
> - I'd print the the auto-mode-list in lower case, like we do with bookmarks. This is meant to be edited by hand and writing in upper case is painful.

Yes, you're right there.

> - I think the most "dwim" behaviour when auto-mode and the user navigates to a URL without rules is to restore the default modes or the last active modes before a rule was matches (this can be saved in the mode slot.) At the moment, the modes from the last URL remain, which does not sound like something we'd want. 

Hmmm, should it be a part of core functionality or a new hook? It sounds more like a built-in option, but then it shouldn't work if auto-mode is disabled. I'll go with hook for now.

> - Actually, do we really need 2 commands to save the modes for a website? It sounds like 1 would suffice if we exclude the modes that are not there but are present in the defaults. Thoughts?

I haven't thought about it before, but it sounds perfectly reasonable now!

Unfortunately, I won't be able to work on these until Saturday :( But these are great suggestions, thatnks for pointing them out!

Ambrevar:

>> - I think the most "dwim" behaviour when auto-mode and the user
>> navigates to a URL without rules is to restore the default modes or
>> the last active modes before a rule was matches (this can be saved in
>> the mode slot.) At the moment, the modes from the last URL remain,
>> which does not sound like something we'd want.
>
> Hmmm, should it be a part of core functionality or a new hook? It sounds more like a built-in option, but then it shouldn't work if auto-mode is disabled. I'll go with hook for now.

Sorry, I didn't get what you mean.  Can you give an example?

To clarify, I think the default behaviour (not even sure this should be
customizable) should be as follows:

_while auto-mode is enabled_, navigating to a URL without rules should
set the modes that were active in for the last URL without rules.

> Unfortunately, I won't be able to work on these until Saturday :( But these are great suggestions, thatnks for pointing them out!

No stress, your contributions are already invaluable! :)

aartaka:
> Sorry, I didn't get what you mean. Can you give an example? 

Sure. The current behavior of Nyxt (without this patch) is to stay with whatever modes are there. In auto-mode, however, it's more intuitive to restore dafault modes on any URL that don't have any matching `auto-mode-rule`. I was confused, should this behavior be hard-coded somewhere, or it's better to have another hook that restores the default modes is `auto-mode` specifically.

aartaka:
> - Actually, do we really need 2 commands to save the modes for a website? It sounds like 1 would suffice if we exclude the modes that are not there but are present in the defaults. Thoughts?

Yes, we do. It seems we're having a use scenario split again there:
- You seem to be pretty okay with storing all the modes for a URL as an exact mode reproduction list, and **disabling** all the other `default-modes` that were not mentioned.
- I'm okay with storing only what I need in addition to `default-modes`, and what I don't need from `default-modes` (as `:included-modes` and `:excluded-modes` respectively), while leaving all the other `default-modes` I haven't mentioned **enabled**. I don't want my `auto-mode-list-data-path` file to be full of lists of repeating modes there.

Don't forget, this file is meant to be easily editable, so I'd better go with the `:exact` flag for the exact mode reproduction (first use case), and exclusion/inclusion for other cases. And, to enable both use scenarios, we need two command -- one for exclusion and one for inclusion.

Ambrevar:
Makes sense, thanks for the clarification.

Ambrevar:

> Don't forget, this file is meant to be easily editable, so I'd better go with the `:exact` flag for the exact mode reproduction (first use case), and exclusion/inclusion for other cases. And, to enable both use scenarios, we need two command -- one for exclusion and one for inclusion.

Actually I'm not sure we need an exclusion command and an inclusion
command. 

If I understood you right, your workflow is to save the mode state for
the URL every time you enable or disable a mode.  In this case, there is
no need for commands.

With my workflow where I call a command to explicitly save the state,
since I've already set the modes I want for the present URL, there is no
need to prompt for the modes and no need for multiple commands.

What do you think?

Ambrevar:
Something else: we need to be able to temporarily override auto-mode.

Say I browse the example.org domain with Tor.  But this one subpage does
not work, so I'd like to temporarily disable Tor (without persisting).

If I disable proxy mode and reload the page, the URL will be matched in
the handler and proxy-mode will be reenabled.

Maybe one way to proceed would be to have handlers be a no-op if
auto-mode is disabled.  This way, disabling auto-mode would allow the
user to tweak the modes until auto-mode is reenabled.

Thoughts?

aartaka:
> Maybe one way to proceed would be to have handlers be a no-op if auto-mode is disabled. This way, disabling auto-mode would allow the user to tweak the modes until auto-mode is reenabled. Thoughts?

Disabling `auto-mode` deletes all the `auto-mode`-related handlers already, so there's no problem with false-positive  handlers (i.e. handlers that activate when they shouldn't) there.

If you've actually noticed handlers still working when `auto-mode` is disabled, I'll be glad to know more details, because it sounds like a bug!

Ambrevar:
Oh, no problem then, I didn't notice that you had implemented this
already :)

aartaka:
Hmmm. it seems that storing last active modes is not a good thing, because they are highly context-dependent and some bugs lurk around when you enable them blindly. 

Is there a solid way to get the list of default modes that will be context-neutral (i.e. independent of the buffer contents and/or class), except for creating new buffer? 

I'll push the updated version that suits all the requirement above, except for default modes reenablement -- I haven't come up with a perfect way to restore modes yet.

Will it be better if I'll rebase against master to fix the merge conflicts, or you'll better fix conflicts yourself (you know recent code changes much better than I do, so you can filter these conflicts better, I think :) )?

aartaka:
- `auto-mode-list` now haves three slots. The main one is `:modes` and it includes exact list of modes to enable on matching URL, disabling all the others.
- Documentation is changed accordingly.
- 1 command (`save-modes-to-auto-mode-list`), instead of 2.

- And still no ability to restore non-obvious default modes.

Ambrevar:

> Hmmm. it seems that storing last active modes is not a good thing, because they are highly context-dependent and some bugs lurk around when you enable them blindly. 

Can you give an example?

> Is there a solid way to get the list of default modes that will be context-neutral (i.e. independent of the buffer contents and/or class), except for creating new buffer? 

(default-modes (current-buffer)) should do.  If you want the default
modes of a non-user-defined buffer class, look at default-mode-symbols
on master.

> Will it be better if I'll rebase against master to fix the merge conflicts, or you'll better fix conflicts yourself (you know recent code changes much better than I do, so you can filter these conflicts better, I think :) )?

I can rebase, no problem.

aartaka:
Okay, I found the reason of the problem. `base-mode`, when disabled, disables all the keyboards schemes, except for CUA, and this was the bug I encountered and was talking about.

What was making it more difficult was that `nyxt::base-mode` was conflicting with the `base-mode` interned in `auto-mode` and I was unable to include it into `*non-rememberable-modes*` as `base-mode` without package prefix (and I was really reluctant to include it with the package prefix). Buto now, when it's there as `nyxt::base-mode`, it matches well and doesn't get disabled, so the bug is no more there.

I'll push the changes in a minute.

aartaka:
Here you go, now default modes get restored on non-matching URLs.

Ambrevar:
There is a compilation warning about disable-mode and enable-mode being
undefined.

Besides, this seems wrong:


    (dolist (buffer buffers)
      (dolist (mode modes)
        (disable-mode mode buffer)))

It this was supposed to be disable-modes, the function takes a list of
modes, so the call should be 

    (dolist (buffer buffers)
      (disable-modes (modes buffer)))

Can you confirm?  Let me know if you don't have time to fix it, I'll do
it then.

source/browser.lisp
```diff
@@ -256,6 +258,18 @@ have an empty ID.")
...
+                     :initform (make-hook-mode)
+                     :type hook-mode
+                     :documentation "Hook run on every mode activation,
+afterwards the mode-specific hook.")
+   (disable-mode-hook :accessor disable-mode-hook
+                      :initarg :disable-mode-hook
+                      :initform (make-hook-mode)
+                      :type hook-mode
+                      :documentation "Hook run on every mode deactivation,
+afterwards the mode-specific hook.")
```
Ambrevar, line 272: Typo: "after"

source/browser.lisp
```diff
@@ -256,6 +258,18 @@ have an empty ID.")
                   :initform '(certificate-whitelist-mode web-mode base-mode)
                   :type list-of-symbols
                   :documentation "The symbols of the classes to instantiate on buffer creation.")
+   (enable-mode-hook :accessor enable-mode-hook
+                     :initarg :enable-mode-hook
+                     :initform (make-hook-mode)
+                     :type hook-mode
+                     :documentation "Hook run on every mode activation,
+afterwards the mode-specific hook.")
```
Ambrevar, line 266: Typo: "after"

source/buffer.lisp
```diff
@@ -421,11 +421,36 @@ The current buffer access time is set to be the last."
...
+(export-always 'mode-short-name)
+(defun mode-short-name (mode)
+  "Return an interned and prettily-printable mode name."
+  (intern (symbol-name (mode-name mode))))
+
+(export-always 'mode-name)
+(defun mode-name (mode)
+  (class-name (class-of mode)))
+
+(declaim (ftype (function (t &optional buffer) t) disable-modes enable-modes))
```
Ambrevar, line 433: If you don't return anything, I prefer specifying nothing as the return type.
Specifying `t` implies that you return something of unspecified type.

Also I'd prefer `list` instead of `t` for the MODES type.

source/buffer.lisp
```diff
@@ -421,11 +421,36 @@ The current buffer access time is set to be the last."
...
+(export-always 'disable-modes)
+(defun disable-modes (modes &optional (buffer (current-buffer)))
+  "Disable MODES for BUFFER. MODES should be a list symbols,
```
Ambrevar, line 436: It's customary to have a single sentence of the first line.  You can start the second sentence on a new line.

source/buffer.lisp
```diff
@@ -421,11 +421,36 @@ The current buffer access time is set to be the last."
...
+each possibly returned by `mode-name'"
```
Ambrevar, line 437: Missing trailing period.

source/buffer.lisp
```diff
@@ -421,11 +421,36 @@ The current buffer access time is set to be the last."
...
+  (dolist (mode modes)
+    (funcall-safely (sym (mode-command mode))
+                    :buffer buffer :activate nil)))
+
+(export-always 'enable-modes)
+(defun enable-modes (modes &optional (buffer (current-buffer)))
+  "Enable MODES for BUFFER. MODES should be a list of symbols,
+each possibly returned by `mode-name'"
```
Ambrevar, line 445: Same as above.

source/buffer.lisp
```diff
@@ -453,8 +477,7 @@ The current buffer access time is set to be the last."
                         :suggestion-function (active-mode-suggestion-filter buffers))))
     (dolist (buffer buffers)
       (dolist (mode modes)
-        (funcall (sym (mode-command mode))
-                 :buffer buffer :activate nil)))))
+        (disable-mode mode buffer)))))
```
Ambrevar, line 480: Wrong function name? (Missing 's'?)
If using the plural version, then you should pass `modes` directly.

source/buffer.lisp
```diff
@@ -474,8 +497,7 @@ The current buffer access time is set to be the last."
                         :suggestion-function (inactive-mode-suggestion-filter buffers))))
     (dolist (buffer buffers)
       (dolist (mode modes)
-        (funcall (sym (mode-command mode))
-                 :buffer buffer :activate t)))))
+        (enable-mode mode buffer)))))
```
Ambrevar, line 500: Same as above.

aartaka:
That's weird — I haven't noticed these function names staying unchanged there, and somehow haven't seen the compilation warnings...

Should be fixed now!

jmercouris:
If you don't see compilation bugs, sometimes it is a good idea to invoke "force-recompilation" in slime.

Ambrevar:
You may not see the warning if you didn't restart the Lisp image and the
former enable-mode was still defined.

To get reliable compilation messages, compile on a fresh image.
You don't have the restart your REPL, you can start a new Lisp in a new REPL.

aartaka:
> You may not see the warning if you didn't restart the Lisp image and the former enable-mode was still defined. To get reliable compilation messages, compile on a fresh image.

This was the case, I guess. Thanks for reminding!

Ambrevar:
Sadly it does not work for me anymore.  I suspect something went wrong when you introduce enable-modes / disable-modes.

Recipe:

- Enable auto-mode.
- Have proxy-mode disabled.
- Navigate to check.torproject.org.
- Enable proxy-mode.
- Call save-modes-to-auto-mode-list
- Reload the page.

Result: proxy mode is disabled, while it should have remained enabled.

source/auto-mode.lisp
```diff
@@ -0,0 +1,272 @@
...
+    (flet ((covered-by-auto-mode-p (candidate-mode url)
+             (or (non-rememberable-mode-p candidate-mode)
+                 (let ((matching-rules (matching-auto-mode-rules url)))
+                   (if matching-rules
+                       (member-mode candidate-mode
+                                    (union (alexandria:mappend #'modes matching-rules)
+                                           (default-modes (buffer mode))))
+                       (member-mode candidate-mode
+                                    (default-modes (buffer mode))))))))
+      (flet ((prompt-user-handler (prompt include)
```
Ambrevar, line 96: Use `labels` instead of nested `flet`.  `labels` is to `flet` what `let*` is to `let`.

source/auto-mode.lisp
```diff
@@ -0,0 +1,272 @@
...
+  (hooks:add-hook (request-resource-hook (buffer mode))
+                  (make-handler-resource
+                   #'(lambda (request-data)
+                       (unless (matching-auto-mode-rules (url request-data))
+                         (disable-modes (remove-if #'non-rememberable-mode-p
+                                                   (mapcar #'mode-short-name (modes (buffer mode)))))
+                         (enable-modes (remove-if #'non-rememberable-mode-p
+                                                  (default-modes (buffer mode)))))
+                       request-data)
+                   :name 'restore-default-modes-handler))
```
Ambrevar, line 85: I don't think we need a lambda here.  The buffer is part of the request-data, so you can replace `(buffer mode)` with `(request-data)` then mode everything to a separate function.  This would help with debugging as well ;)

source/auto-mode.lisp
```diff
@@ -0,0 +1,272 @@
...
+                                            :key #'test :test #'equal))
+                                (rule-modes
+                                  (if rule
+                                      (modes rule)
+                                      (mapcar #'mode-short-name (modes (buffer new-mode)))))
+                                (modes (if include
+                                           (union (list (mode-short-name new-mode)) rule-modes)
+                                           (set-difference rule-modes
+                                                           (list (mode-short-name new-mode))))))
+                           (add-modes-to-auto-mode-list test modes))))))))
```
Ambrevar, line 117: If I understand correctly, the above flets can be extracted to a separate function.

It's good practice to keep functions short.  It's also better to have named, top-level functions whenever possible since it helps with debugging.

Ambrevar:
Let me know whenever you start losing patience on this, I can help ironing out the last issues.  (This should remain fun and not become a frustration :p)

source/auto-mode.lisp
```diff
@@ -0,0 +1,272 @@
...
+
+(defun initialize-auto-mode (mode)
+  (hooks:add-hook (request-resource-hook (buffer mode))
+                  (make-handler-resource
+                   #'(lambda (request-data)
+                       (unless (matching-auto-mode-rules (url request-data))
+                         (disable-modes (remove-if #'non-rememberable-mode-p
+                                                   (mapcar #'mode-short-name (modes (buffer mode)))))
+                         (enable-modes (remove-if #'non-rememberable-mode-p
+                                                  (default-modes (buffer mode)))))
```
Ambrevar, line 83: I don't understand this chunk: 

- When there is no result, this disables the rememberable modes, then it enables the default modes.  
I suppose the goal was to restore the default modes.  But as we discussed, I think it's better to save the modes that were active for the last unmatched URL  (it's easy to save these modes in the auto-mode class object for instance).  Furthermore, you should avoid toggling a mode since it can have side effect.
To sum up:  Disable the modes that are not present in the "saved mode list" and enable those that are missing.  You can use the `set-difference` standard function.

- And when there is a result, this does nothing.  I believe this is why it does not work for me.

Ambrevar:
I believe I found out why it does not work:
https://github.com/atlas-engineer/nyxt/pull/846/files#r458629048

aartaka:
One moment I didn't quite get: when should we save these modes for non-matching pages? 

- If we save them at page load, then they can interfere the moment of restoration of themselves.
- If we save them on every mode toggle (the best option I can think of), then we need to add even more handlers (should they go before or after prompting handlers?) and, possibly, look if user (in case `*prompt-on-mode-toggle*` is enabled) have added the modes to `auto-mode-list`, because then we'd need to discard the saved modes and stay with the previous set of saved modes.
- If we save them unconditionally, we can miss on some important events.

aartaka:
> I believe I found out why it does not work:
> https://github.com/atlas-engineer/nyxt/pull/846/files#r458629048

I found yet another reason for it not working: the `add-hook`-ing of auto-mode handlers in `initialize-auto-mode` was inside the `(when *prompt-on-mode-toggle* ... )` block, and was evaluated only if this variable was set to non-nil, which was not the case for you. That's why the newly-added proxy-mode wasn't enabled. 

I'll push the fix (and, possibly, the feature of restoring previous list of modes) in three hours.

Ambrevar:

> One moment I didn't quite get: when should we save these modes for non-matching pages? 
>
> - If we save them at page load, then they can interfere the moment of restoration of themselves.

Not if we do it withing the same handler that makes sure the modes are first saved, then restored.

> - If we save them on every mode toggle (the best option I can think of), then we need to add even more handlers (should they go before or after prompting handlers?) and, possibly, look if user (in case `*prompt-on-mode-toggle*` is enabled) have added the modes to `auto-mode-list`, because then we'd need to discard the saved modes and stay with the previous set of saved modes.



> - If we save them unconditionally, we can miss on some important events.

What is "unconditionally" and "important events"?


What I had in mind is something like this:

```lisp
(defun initialize-auto-mode (mode)
  (hooks:add-hook (request-resource-hook (buffer mode))
                  (make-handler-resource
                   #'(lambda (request-data)
                       (if (matching-auto-mode-rules (url request-data))
                           (progn
                             ;; CHANGE: Save current modes if current buffer has no rule:
                             (unless (matching-auto-mode-rules (url (current-buffer)))
                               (save-current-modes ...))
                             (disable-modes (remove-if #'non-rememberable-mode-p
                                                       (mapcar #'mode-short-name (modes (buffer mode)))))
                             (enable-modes (remove-if #'non-rememberable-mode-p
                                                      (default-modes (buffer mode)))))
                           ;; CHANGE: Restore saved modes
                           (restore-current-modes ...)
                           )
                       request-data)
                   :name 'restore-default-modes-handler))
  ...)
```

aartaka:
> What is "unconditionally" and "important events"?

I mean saving them independently of what user does (saving modes each five minutes, for example). This can miss mode toggling that should influence this default mode setup, and disable some mode that was enabled but wasn't saved yet.

Ambrevar:
Agreed, a timer is not a good idea to save modes.

aartaka:
But isn't `request-resource-hook` firing after the request to ffi from `buffer-load`? `buffer-load` sets `url` of the `buffer` to the new URL and calls ffi only after that. If so, there's no difference between `(url request-data)` and `(url (current-buffer))` is your example. 

I've tried it, and it saves nothing.

aartaka:
The problem is not in the `buffer-load`, sorry. It's okay, and it's not causing the problem I said it does.

Ambrevar:
I haven't tested it :p 

Try extracting the functions as I suggested, it should help with debugging.

aartaka:
Strange behavior: `url` of `request-data` and `url` of `current-buffer` contains the same URL at the time of `request-resource-hook` activation, unless it's a third-party request. Is this deliberate?

You can test it with this `url-dispatching-handler` 

``` lisp
(url-dispatching-handler
 'log-urls-of-request-and-buffer
 (match-regex ".*")
 (lambda (url)
   (log:info "The current-buffer URL is ~s"
             (object-string (url (current-buffer))))
   (log:info "The request URL is ~s"
             (object-string url))
   url))
```

It logs "reddit.com" and "reddit.com" for both `current-buffer` and `request-data` if you request Reddit. But then, when Reddit starts requesting media/ads/whatever, URLs start diverging:
```
<INFO> [12:56:36] The current-buffer URL is "https://www.reddit.com/"
<INFO> [12:56:36] The request URL is "https://www.reddit.com/"
<INFO> [12:56:36] Applied "LOG-URLS-OF-REQUEST-AND-BUFFER" URL-dispatcher on "https://www.reddit.com/" and got "https://www.reddit.com/"
...
<INFO> [12:56:41] The current-buffer URL is "https://www.reddit.com/"
<INFO> [12:56:41] The request URL is "about:blank"
<INFO> [12:56:41] Applied "LOG-URLS-OF-REQUEST-AND-BUFFER" URL-dispatcher on "about:blank" and got "about:blank"
<INFO> [12:56:43] The current-buffer URL is "https://www.reddit.com/"
<INFO> [12:56:43] The request URL is "https://www.redditmedia.com/gtm/jail?id=GTM-5XVNS82"
<INFO> [12:56:43] Applied "LOG-URLS-OF-REQUEST-AND-BUFFER" URL-dispatcher on "https://www.redditmedia.com/gtm/jail?id=GTM-5XVNS82" and got "https://www.redditmedia.com/gtm/jail?id=GTM-5XVNS82"
<INFO> [12:56:43] The current-buffer URL is "https://www.reddit.com/"
<INFO> [12:56:43] The request URL is "https://www.redditmedia.com/mediaembed/hw33o3?responsive=true&is_nightmode=false"
```

aartaka:
I mean, your suggested code impied that the URLs of `buffer` and `request-data` should differ when `request-resource-hook` runs (unless it's a reload request), but it's not the case in current implementation somewhy.

aartaka:
Okay, I've found a workaround: we can look at the `htree:parent` of `htree:current` node in the history of `web-mode` and check it against the `auto-mode-list`. If it'll have no rule mathing it, we can safely save the current mode list as default.

But then this hook should definitely run before the mode-changing hooks generated from `auto-mode-list`. Is `add-hooking` this "non-auto-mode-hook" after these rules enough for that?

Ambrevar:
OK for the htree workaround, although I'd like to understand why `(url (current-buffer))` is already updated at this point.  This does not seem right.  I'll investigate.

In your last paragraph, did you mean "handler" instead of "hook"?
If we have a single handler that applies to unmatched URLs, then it does not matter where we put it in the resquest-resource-hook, or am I mistaken?

aartaka:
Well, the problem is that saving happens in the same handler. If a request is made for the page that has an associated rule, and modes are changed before this handler activates, it will save the changed modes upon activation.

aartaka:
> In your last paragraph, did you mean "handler" instead of "hook"?

Yes, it was the "handler" I meant :)

Ambrevar:
Can you provide a precise recipe of the problem?  Which is the current page, which is the loading page?  Which modes are currently active?  Which modes are we going to load?  Etc.

I found out about `(url (current-buffer))`.  It's set in `on-signal-notify-uri (buffer)` and there is no way around it, because some pages (like clicking on an anchor) update the URI via this signal only.
The request-resource-hook is always run _after_ the notify::uri signal (this is how WebKitGTK works).
As such, the `(url (current-buffer))` will always be the same as the one of the request-data.  I'll document this for posterity.

aartaka:
One more issue: is it necessary to represent the `modes` of `auto-mode-rule` as symbols? 

I used symbols originally, because I didn't know how to compare and filter mode instances, and this workaround stayed there, but it's quirky and less straightforward to use. Moreover, now, when I came up with a good way of mode comparison, it's no longer necessary to store symbols.

Pros of symbols:
+ it's easy to call them with `(funcall (sym (mode-command foo)))`.
+ easy to serialize.

Cons:
- `(mapcar #'mode-short-name ...)` is everywhere in the code, making it less readable.
- `mode-short-name` reinterns symbols into `nyxt-user`, and it can harm us later.
- we can't trace exact settings of saved modes.

Storing mode instances is a good alternative:

Pros:
+ Almost as easy to call, because we have `mode-name`.
+ Almost as easy to serialize, for the same reason.
+ Code is cleaner.
+ No unnecessary symbol interning.
+ We can reuse instances and possibly lose less data.

Cons:
- It's a bit harder to filter them, and pick the non-rememberable ones, but it's not impossible.
- It may be necessary to still support the symbol representation to store the not-yet-enabled modes right after desealization.

aartaka:
> Can you provide a precise recipe of the problem? Which is the current page, which is the loading page? Which modes are currently active? Which modes are we going to load? Etc.

I'll try to think how to make handlers independent of their order. It's not a bug yet, because it is a part of current implementation that can be rewritten :)

Ambrevar:
OK!

About the mode representation: it makes sense to me to handle the mode
objects instead of the mode symbols.  This is generally what we do in
the rest of the code.

If I'm not mistaken, the only place where we deal with mode symbols is
in buffer's `default-modes' slot: the reason for this is that this
configuration slot can only provide names since mode objects are
instantiated after these names.

Ambrevar:
If you wonder "but why don't we instantiate the modes in the `modes'
slot initform?", the answer is that modes need to have access to the
instantiated buffer (including its webview) when instantiated.

aartaka:
What should the first value of this list of last non-auto-mode-enabled modes be? 

I'm leaning towards `default-modes` for this value, but we can't derive them at `auto-mode` initialization. The reason is that at the time of mode initialization not every mode could be initialized and `find-submode` wouldn't work. So I set it to `default-modes` the first time it's needed, i.e. on the first requests of the URL not covered by auto-mode. Such a request usually happens quite soon after browser startup, so nothing extremely bad should happen...

It's not bulletproof, but I haven't found a better option yet. You're welcome to suggest a better one!

Ambrevar:

> What should the first value of this list of last non-auto-mode-enabled modes be? 

Which list?  Sorry, I don't understand.

> I'm leaning towards `default-modes` for this value, but we can't
> derive them at `auto-mode` initialization. The reason is that at the
> time of mode initialization not every mode could be initialized and
> `find-submode` wouldn't work.

Which mode initialization?
Did you mean auto-mode initialization?

If so, maybe you meant there could be a problem when we initialize the
modes of a buffer, say (mode-A auto-mode mode-B), if auto-mode happens
to tweak the modes, we might end up with inconsistent active modes.  For
instance, it could be (auto-mode mode-B mode-C) or (mode-A auto-mode
mode-D).

I don't think the above is a problem since auto-mode matching happens
during resource-request-hook execution, which happens after all modes
have initialized.

If I misunderstood you, can you provide a precise receipe?
This is a bit difficult to discuss in writing, ideally code snippets
would make it clearer.

aartaka:
Okay, I've overhauled the way rules are activated and stored.

### What's New
- Now `auto-mode` rules don't get transformed into hook. We can't contorl 
  the amount (and sometimes order) of hooks that shoot on a matching URL, and 
  we don't need more than one `auto-mode` rule applied to the same URL -- 
  it's unpredictable.
- As a consequence of this divergence from hooks, there's no more `name` 
  slot in `auto-mode-rule`. It makes it hard to log, but do we really 
  need to log it?
- Strictly one `auto-mode-rule` can be called on a matching domain -- the 
  most specific one, with the help of `matching-auto-mode-rule` (the actual 
  priority is `match-regex` > `match-url` > `match-host` > `match-domain`).
- `mode-equal` that compares modes and their names (as symbols) and vice versa.
- Several suspicious helper-functions connected to `mode-equal`, 
  like `maybe-mode-name` and `rememberable-of`. They introduce type uncertainty,
  and I'll be happy to get rid of them in case we can find a worthy alternative!
- `add-modes-to-auto-modes` now accepts almost anything as modes 
  and implicitly converts it to symbols with `maybe-mode-name`.
- The modes of last non-matching URL are saved conveniently -- 
  in `last-active-modes` slot of `auto-mode`.
- Almost everything (except the optional prompting handlers) is now in one 
  handler -- `auto-mode-request-handler`, because this way we can state that 
  `last-active-modes` restoration and saving goes before activation 
  of whatever matched rule.
- No more unnecessary interns and no more `mode-short-name`.
- More function type declarations and small debuggable functions.

### What Haven't Changed
- Modes are still stored as their names (symbols) in the `modes`
  slot of `auto-mode-rule`. 
  
  Problematizing example: after desearialization of 
  `auto-mode-list-data-path`, we need to store something to the `modes` of 
  `auto-mode-list` rules. Ideally we should store mode instances, but, for 
  example, `proxy-mode` is not enabled in any buffer. Enabling it just to
  store its instance is not good -- mode activation can cause side effects 
  that are undesirable for the user, if they didn't asked for that. Thus, 
  we can store only `NYXT/PROXY-MODE:PROXY-MODE` as symbol. And being 
  coherent is better that having two different types in one list, so
  I've chosen to represent _all_ modes as symbols is `modes` of `auto-mode-rule`.

### Caveats:
- Should we have the `name` slot of `auto-mode-rule` back and how should we 
  log `auto-mode` activity?
- Having type uncertainty is bad, so ideally there shouldn't be `maybe-mode-name`
  and even necessity for it. Hovewer, it's the best I was able to come with on 
  my own. We can  (and, probably, need to) discuss a better design.
- There should be a safe side-effect-less way to represent modes as
  mode instanses in `modes` of `auto-mode-rule`. I haven't found it
  yet, but maybe we can discuss it and come up with something.
- There is a dependence (in `auto-mode-request-handler`) on this strange 
  behavior of `(url (current-buffer))` and `(url request-data)` being equal.
  It's used to make sure that the request is request is a `buffer-load`-
  initiated one, because other requests shouldn't concern `auto-mode`.
  To see this how other requests influence this, remove the
  ``` lisp
    (quri:uri= (url request-data) (url (buffer auto-mode)))
  ```
  condition form `auto-mode-request-handler`, save some modes for 
  old.reddit.com, and then load it -- all the requests to the third 
  party ads/fonts/media providers will just override the `auto-mode-rule` 
  associated to old.reddit.com. This shouldn't be the case, but using
  the "strange behavior" against it is too fragile.
  
### How Has This Been Tested
- I've ran Nyxt form the fresh REPL with this set of changes and tried 
  both use scenarios just browsing around some random URLs deliberately 
  testing different sequences of matching and non-matching URLs to see 
  whether the restoration of `last-active-modes` is working. It was okay 
  for me, but more testing is welcome, especially given the differences 
  of our uses for `auto-mode` :)
  
I'll be glad to see your feedback and testing on this change!

source/auto-mode.lisp
```diff
@@ -0,0 +1,306 @@
...
+  (:import-from #:serapeum #:export-always)
+  (:documentation "Mode for automatic URL-based mode toggling."))
+(in-package :nyxt/auto-mode)
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (trivial-package-local-nicknames:add-package-local-nickname :alex :alexandria))
+
+(export-always '*prompt-on-mode-toggle*)
+(defparameter *prompt-on-mode-toggle* nil
+  "Whether user will be asked about adding the mode to included/excluded modes
+in auto-mode-list on mode activation/deactivation.")
```
Ambrevar, line 12: Should be a defvar, no?

source/auto-mode.lisp
```diff
@@ -0,0 +1,306 @@
...
+
+(export-always '*non-rememberable-modes*)
+(defparameter *non-rememberable-modes*
+  ;; Base mode conflicts with its Nyxt symbol if it's not prefixed
+  '(help-mode web-mode auto-mode nyxt::base-mode)
+  "Modes that AUTO-MODE won't even try to save.")
```
Ambrevar, line 18: Same, should be a defvar I think.

source/auto-mode.lisp
```diff
@@ -0,0 +1,306 @@
...
+                           (modes (buffer auto-mode))
+                           :test #'mode-equal))
+   (buffer auto-mode)))
+
+(defun auto-mode-request-handler (request-data)
+  (let* ((auto-mode (find-submode (buffer request-data) 'auto-mode))
+         ;; We need to supress prompting when auto-mode modifies modes.
+         (*prompt-on-mode-toggle* nil)
+         (web-mode (find-submode (buffer request-data) 'web-mode))
+         (prev-url
```
Ambrevar, line 102: Nit: previous-url is better :)
Otherwise, `old-url` is shorter, but less accurate.

source/auto-mode.lisp
```diff
@@ -0,0 +1,306 @@
...
+          (when (and prev-url (not (matching-auto-mode-rule prev-url)))
+            (store-last-active-modes auto-mode))
+          (enable-matching-modes (url request-data) (buffer request-data)))
+        ;; Apply previous saved modes only on buffer-loads, not anytime else.
+        ;;
+        ;; TODO: Relies on somewhat illogical behavior of (url (current-buffer))
+        ;; and (url request-data) being equal after buffer-load is called.
+        ;; Replace with more stable and logical condition?
+        (when (quri:uri= (url request-data) (url (buffer auto-mode)))
+          (restore-last-active-modes auto-mode))))
```
Ambrevar, line 118: To clarify, when don't we want to restore the last active modes?
buffer-load is not called when clicking a URL, so are you sure about the condition?

source/auto-mode.lisp
```diff
@@ -0,0 +1,306 @@
...
+
+(declaim (ftype (function (boolean t) (function (root-mode)))
+                make-mode-toggle-auto-mode-handler))
+(defun make-mode-toggle-prompting-handler (enable-p auto-mode-instance)
+  #'(lambda (mode)
+      (when (and (not (mode-covered-by-auto-mode-p mode auto-mode-instance))
+                 *prompt-on-mode-toggle*)
+        (nyxt::with-confirm ("Permanently ~a ~a for this URL?"
+                             (if enable-p "enable" "disable")
+                             (mode-name mode))
```
Ambrevar, line 137: Note to myself: Remove nyxt:: when rebasing on master.

source/auto-mode.lisp
```diff
@@ -0,0 +1,306 @@
...
+      (echo "Enable auto-mode first.")))
+
+(declaim (ftype (function (list (or null list)) (values list &optional))
+                add-modes-to-auto-mode-list))
+(defun add-modes-to-auto-mode-list (test modes)
+  (let ((rule (or (find test (auto-mode-list *browser*) :key #'test :test #'equal)
+                  (make-instance 'auto-mode-rule :test test))))
+    (when modes
+      (setf (modes rule) (rememberable-of (mapcar #'maybe-mode-name modes))
+            (auto-mode-list *browser*) (remove-duplicates
```
Ambrevar, line 222: delete-duplicates

source/auto-mode.lisp
```diff
@@ -0,0 +1,306 @@
...
+
+(defmethod serialize-object ((rule auto-mode-rule) stream)
+  (let ((*standard-output* stream)
+        (*print-case* :downcase))
+    (write-string "(" stream)
+    (if (and (eq (first (test rule)) 'match-url)
+             (= 2 (length (test rule))))
+        (format t "~s " (second (test rule)))
+        (format t "~s " (test rule)))
+    (format t ":MODES ~a"
```
Ambrevar, line 236: Why MODES in all capitals?  Should probably be downcased.

Ambrevar:
Excellent!  
I feel that I've nagged you enough on this feature, so after testing I'll test and fix the few commends I've left in my review.  (If that's OK with you.)

We can iron out the remaining details later.

I'm sorry if this has been dragging, but I believe it was worth the effort: I really like the last implementation, it is much clearer and makes a lot more sense in my opinion.  The code is very well written!
I believe this is the right direction!

Ambrevar:
To answer some of your questions:

> - Now `auto-mode` rules don't get transformed into hook. We can't contorl 
>   the amount (and sometimes order) of hooks that shoot on a matching URL, and 
>   we don't need more than one `auto-mode` rule applied to the same URL -- 
>   it's unpredictable.

Indeed, I think auto-mode needs not allow multiple rules per URL.  I
can't think of a use case.  If the user really wants a more
sophisticated URL rule, they can always add their own in request-resource-hook.

> - As a consequence of this divergence from hooks, there's no more `name` 
>   slot in `auto-mode-rule`. It makes it hard to log, but do we really 
>   need to log it?

Considering it's much simpler than before, maybe not.

> - Strictly one `auto-mode-rule` can be called on a matching domain -- the 
>   most specific one, with the help of `matching-auto-mode-rule` (the actual 
>   priority is `match-regex` > `match-url` > `match-host` > `match-domain`).

Makes lots of sense!

> - Several suspicious helper-functions connected to `mode-equal`, 
>   like `maybe-mode-name` and `rememberable-of`. They introduce type uncertainty,
>   and I'll be happy to get rid of them in case we can find a worthy alternative!

By "type uncertainty", you mean (or symbol root-mode)?
I believe it's OK, but maybe we could introduce a mode-symbol type.

Note that mode symbols are necessarily unique, because by contract each
mode must be in their own package of the same name.

(base-mode and root-mode being the only exceptions.)

>   Problematizing example: after desearialization of 
>   `auto-mode-list-data-path`, we need to store something to the `modes` of 
>   `auto-mode-list` rules. Ideally we should store mode instances, but, for 
>   example, `proxy-mode` is not enabled in any buffer. Enabling it just to
>   store its instance is not good -- mode activation can cause side effects 
>   that are undesirable for the user, if they didn't asked for that. Thus, 
>   we can store only `NYXT/PROXY-MODE:PROXY-MODE` as symbol. And being 
>   coherent is better that having two different types in one list, so
>   I've chosen to represent _all_ modes as symbols is `modes` of `auto-mode-rule`.

I believe this is sane.  As I said above, mode symbols are unique.
So if we want to make this a bit more robust type-wise, we could
introduce a mode-symbol type I think.

Thoughts?

> ### Caveats:
> - Should we have the `name` slot of `auto-mode-rule` back and how should we 
>   log `auto-mode` activity?

I'll test more and see if it's a problem.

> - Having type uncertainty is bad, so ideally there shouldn't be `maybe-mode-name`
>   and even necessity for it. Hovewer, it's the best I was able to come with on 
>   my own. We can  (and, probably, need to) discuss a better design.

Maybe the mode-symbol type is a solution.

> - There should be a safe side-effect-less way to represent modes as
>   mode instanses in `modes` of `auto-mode-rule`. I haven't found it
>   yet, but maybe we can discuss it and come up with something.

The problem is that modes can have side effects by nature.  This is
mostly because web renderers are not functional, e.g. enabling the proxy
is an impure operation.  I don't think there is a way around this.

> - There is a dependence (in `auto-mode-request-handler`) on this strange 
>   behavior of `(url (current-buffer))` and `(url request-data)` being equal.
>   It's used to make sure that the request is request is a `buffer-load`-
>   initiated one, because other requests shouldn't concern `auto-mode`.
>   To see this how other requests influence this, remove the
>   ``` lisp
>     (quri:uri= (url request-data) (url (buffer auto-mode)))
>   ```
>   condition form `auto-mode-request-handler`, save some modes for 
>   old.reddit.com, and then load it -- all the requests to the third 
>   party ads/fonts/media providers will just override the `auto-mode-rule` 
>   associated to old.reddit.com. This shouldn't be the case, but using
>   the "strange behavior" against it is too fragile.

What do you mean with "override the auto-mode-rule"?  You mean they will
match the rule as well and lead Nyxt to set the modes again and again?

I need to think a bit more about this, but at first glance it seems that
the condition is fine.

As far as I understand, request-resource-hook is the only place where we
can capture the load of a URL (including URL clicks) without capturing
anchor changes (e.g. clicking on a TOC anchor).

So testing if the requested resource matches the buffer URL seems like a
proper way to capture the load of the "main URL".

I suggest we extract the condition to a well-named predicate and explain
this detail in the docstring.

Ambrevar:
Merged with 7a0382e30ef9ea05b05d9123212eb5a81f59531d.  I've fixed all the comments in subsequent commits.

See #868 for a continued discussion.
Feel free to post here if there is anything else I've missed.

aartaka:
> What do you mean with "override the auto-mode-rule"? You mean they will match the rule as well and lead Nyxt to set the modes again and again? 

I mean they will override the modes set by matching auto-mode-rule, i.e. restore `last-active-modes` even on a matching URL if there are non-"main" URL requests, like ads/fonts/whatever loads.

PR #847, opened 2020-07-14
Allow quri:uri in set-url* and rename it to buffer-load.

Ambrevar:
set-url* was a wrapper around the low-level ffi-buffer-load, so to be consistent
with our other wrappers, we rename it like its low-level function minus "ffi-".

Now that our URLs are quri:uri objects, we can easily make the distinction
between raw URLs and user input that needs to be passed to parse-url: we simply
check if the argument is a string or a quri:uri.
This rids us of all the object-string conversions.

While we are at it, we unexport set-url-hook which was mistakenly exported.  We
also rename it to buffer-load-hook for consistency.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -63,10 +63,12 @@ MODES is a list of mode symbols.
 If URL is `:default', use `default-new-buffer-url'."
   (let* ((buffer (buffer-make *browser* :title title :default-modes modes))
          (url (if (eq url :default)
-                  (object-string (default-new-buffer-url buffer))
+                  (default-new-buffer-url buffer)
                   url)))
-    (unless (str:emptyp url)
-      (set-url* url :buffer buffer))
+    (unless (if (quri:uri-p url)
+                (url-empty-p url)
+                (str:emptyp url))
+      (buffer-load url :buffer buffer))
     buffer))
 
 (define-class-type buffer)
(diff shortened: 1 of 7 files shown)


source/buffer.lisp
```diff
@@ -63,10 +63,12 @@ MODES is a list of mode symbols.
 If URL is `:default', use `default-new-buffer-url'."
   (let* ((buffer (buffer-make *browser* :title title :default-modes modes))
          (url (if (eq url :default)
-                  (object-string (default-new-buffer-url buffer))
+                  (default-new-buffer-url buffer)
                   url)))
-    (unless (str:emptyp url)
-      (set-url* url :buffer buffer))
+    (unless (if (quri:uri-p url)
```
jmercouris, line 68: (url-empty-p "") -> T

so, I believe it can just be (unless (url-empty-p url), no?

jmercouris:
Other than my comment about the branch condition, it looks good to me! Nice aesthetic improvement :-)

source/buffer.lisp
Ambrevar: Indeed!

Ambrevar:
Merged with 316121ce44594625f3841098ecaba2460495c6a9.

source/buffer.lisp
Ambrevar: Actually (url-empty-p nil) would return nil, which was wrong.  I've fixed url-empty-p.  The type declaration was on the wrong symbol.

jmercouris:
Woohoo!

Issue #581, opened 2020-03-04
Move buffer and window classes + methods to their respective files


Ambrevar:
Current `browser.lisp` is huge and cluttered with definitions for `buffer` and `window`.  It would make more sense and it would probably make it easier to find our way around if those definitions were move to `buffer.lisp` and `window.lisp`.

Thoughts?

jmercouris:
Agreed!

Ambrevar:
What about moving the proxy code to a separate file? Say buffer-proxy.lisp.
We also have quite a few functions that relate to the request-resource-hook. We could also move these to a separate file, say buffer-request.lisp.
This way we would have all buffer related functions grouped in the buffer*.lisp files.

jmercouris:
Wait a minute, didn't you say the exact same thing in the pull request :-D?

Ambrevar:
I did, I'm leaving this here so that we don't forget about it ;)

Ambrevar:
While most files were rearranged, we haven't cleaned up the "request-resource" and proxy.  Shall we do it?

jmercouris:
I don't think it is necessary. Personally I don't touch those pieces of code very often. Plus, they can be moved at any time without breaking API.

Ambrevar:
Fair enough.  If I feel like moving them someday, I just will, no need
for an open issue.

PR #852, opened 2020-07-15
Move buffer and window classes to correct files

jmercouris:
This is with regards to https://github.com/atlas-engineer/nyxt/issues/581

In a subsequent pull request, the methods can be moved. This is the minimum necessary to move the classes.


source/package.lisp
```diff
@@ -27,4 +27,12 @@ It's recommended to use this package in the Nyxt configuration file, instead of
...
 (uiop:define-package parenscript-user
-  (:use :common-lisp :nyxt :parenscript))
+    (:use :common-lisp :nyxt :parenscript))
+
+(uiop:define-package :nyxt/minibuffer-mode
+  (:use :common-lisp :trivia :nyxt)
+  (:import-from #:keymap #:define-key #:define-scheme)
+  (:import-from #:serapeum #:export-always)
+  (:export :minibuffer-mode)
+  (:documentation "Mode for minibuffer"))
```
Ambrevar, line 37: Can you add a comment explaining why this mode must be here?
Otherwise the reader may not understand why this is inconsistently declared here, unlike other modes.

source/minibuffer-mode.lisp
```diff
@@ -63,6 +58,58 @@
...
+          (with-result (engine (read-from-minibuffer
+                                (make-minibuffer
+                                 :input-prompt "Search engine"
+                                 :input-buffer (if (zerop match-count) "" (input-buffer minibuffer))
+                                 :suggestion-function #'search-engine-suggestion-filter)))
+            (when engine
+              (kill-whole-line minibuffer)
+              (insert minibuffer (str:concat (shortcut engine) " "))))))))
+    (t
+     (insert-suggestion minibuffer))))
```
Ambrevar, line 111: We have a lot of search-engine related code.  What about moving it to a new search-engines.lisp file?

Ambrevar:
Thank you so much for taking a shot at this, it's been itching me for way too long! :)

Looking good beside the search-engine suggestion.

source/minibuffer-mode.lisp
jmercouris: we can do that, it is outside of the scope of this pull request, so I will do it in a separate commit

source/package.lisp
jmercouris: will do!

jmercouris:
Yay!

Ambrevar:
> we can do that, it is outside of the scope of this pull request, so I will do it in a separate commit

But you've already moved them.
The downside of moving code around too many times is that it's
harder to track the Git history of the related code.

jmercouris:
It sure is, luckily we are only going to do this once :-D

jmercouris:
Ah, I just understood what you meant by "I've already moved them"! True, but I moved them for compilation purposes, which I think does justify a discrete commit.

PR #853, opened 2020-07-15
Makefile: update submodules without --remote

hsanjuan:
Remote causes the update to use the latest HEAD in the submodule's remote,
rather than the version tied to the nyxt's project current commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ endif
 
 .PHONY: quicklisp-extra-libs
 quicklisp-extra-libs:
-	$(NYXT_INTERNAL_QUICKLISP) && git submodule update --init --remote || true
+	$(NYXT_INTERNAL_QUICKLISP) && git submodule update --init || true
 	mkdir -p "$(QUICKLISP_DIR)/local-projects"
 	for i in $(QUICKLISP_LIBRARIES)/*; do ln -sf "$$(readlink -f "$$i")" "$(QUICKLISP_DIR)/local-projects/$$(basename "$$i")"; done
 


hsanjuan:
Current `make all` leaves a dirty state because it checks out (sometimes, when a difference exists) newer versions of the submodules than those commited with the code.

Ambrevar:
Sorry, I don't understand.  We indeed want to use the submodule's master, not the commit that were commited in Nyxt.  This is on purpose.  Maybe it's a bad decision, but then we need to update the submodule commits if we stop using `--remote`.

hsanjuan:
>  Maybe it's a bad decision, but then we need to update the submodule commits if we stop using --remote

That is the whole point of submodules, for every commit that I check out I must know exactly what version of the submodules I'm supposed to use (a working one). It makes no sense that checking out an old commit and trying to build that fails because it uses newer version of submodules than what it used in the past.

jmercouris:
That is a good point, however we only change cl-webkit in backwards compatible ways.

The only downside to accepting this change would be that we'd have to update the sub module commit- I'm okay with that if Pierre agrees.

hsanjuan:
Think of trying to git bisect to find when an issue was introduced... it would be impossible if you don't have information on what version of dependencies you were building with. Simply going back to a Nyxt version that worked if there is some problem becomes way more difficult if it was an issue within the dependencies in one of the submodules. Really difficult to figure out for the end user.

Ambrevar:
Héctor is right, storing commits is much better practice for reproducibility.

PR #858, opened 2020-07-20
refactor: Move methods/functions from browser to window|buffer

jmercouris:
Methods that are specific to buffers and windows are moved to their
appropriate files.

This will close https://github.com/atlas-engineer/nyxt/issues/581

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -82,19 +82,6 @@ Without handler, return ARG.  This is an acceptable `combination' for
                  result)))
     (compose-handlers (mapcar #'hooks:fn (hooks:handlers hook)) arg)))
 
-(defmethod proxy ((buffer buffer))
-  (slot-value buffer 'proxy))
-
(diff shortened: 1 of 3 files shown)


Ambrevar:
Not sure it closes #581, there are still some file arrangement we can do.  But I guess we can remove the 2.0 tag at least.

What about moving the proxy code to a separate file?  Say buffer-proxy.lisp.
We also have quite a few functions that relate to the request-resource-hook.  We could also move these to a separate file, say buffer-request.lisp.
This way we would have all buffer related functions grouped in the buffer*.lisp files.

Ambrevar:
Otherwise looks good to me.
I haven't tested, make sure it compiles without warnings even on `,reload-system`.

jmercouris:
I think we should talk about the separate files in an abstract sense. When do we break apart files, what are the reasons? I prefer longer files over many files, personally. I used to prefer many short files when I was a java developer, but things have changed :-D

P.S. no warnings

PR #864, opened 2020-07-23
browser.lisp: add button to default status formatter

jmercouris:
window.lisp: add more styles to the status line for button usage

this commit adds buttons to the status mode line so that the user can
click and the buffer will go backwards, forwards, or refresh

this relies on the proper resolution of lisp:// URLs.


jmercouris:
This is what it looks like: 
Screen Shot 2020-07-23 at 14 26 52

Ambrevar:
This would be a great UI improvement indeed.

To clarify: Did you mean to merge this or just show-case how this would look like?
Because we don't support lisp:// URIs at the moment, so I suppose this would not work as is.

Why the hex encoding in the href though?  Would could use quri here for more code readability if need  be.

jmercouris:
> This would be a great UI improvement indeed.
> 
> To clarify: Did you mean to merge this or just show-case how this would look like?

Not ready to merge yet, still needs lisp:// support, just wanted to show that it is ready.

> Because we don't support lisp:// URIs at the moment, so I suppose this would not work as is.
> 
> Why the hex encoding in the href though? Would could use quri here for more code readability if need be.

Good point! I could have used (quri:uri-encode ...) inline!

jmercouris:
I have now using an FLET which should make it clearer what each URL is supposed to do.

Ambrevar:
Yup, the FLET seems to be necessary :)

jmercouris:
Now that we have internal buffer support, we can make the status line inherit from internal buffer.

jmercouris:
Merged and working! So many strange behaviors from macOS... I will have to go into detail later.

Ambrevar:
Just tested, seems to be working great so far for me!
Thanks a lot for this, it's a fantastic improvement in my opinion!

The only issue I ran into is that the BUFFERS command display a blank
button instead of wrapping over the URL.

Ambrevar:
Oh, got it: it's when the title is empty!
We had this issue before: you should never use the title as the first
display option because it may be missing, in particular when we restart
the browser.

I'll fix it.

jmercouris:
Glad you like it!

PR #873, opened 2020-07-27
paste-or-set-url no longer loads empty buffers when middle clicking on page background

walseb:
Currently, if you open github.com for example and middleclick on the background or anywhere there aren't any links or you don't have any text-boxes selected, nyxt will open a empty buffer. This new check fixes that and I haven't seen any negative side-effects.


Ambrevar:
Looks good to me!  Can you just fix the indentation before I merge?

walseb:
Is this better? I'm not very familiar with common lisp

Ambrevar:
Perfect, thanks!

PR #874, opened 2020-07-30
Define default search engine via default-search-engine slot

xelxebar:
Previously, the magic value "default" in a search-engine signified it as
the default search engine. This commit factors that hard-coded value
into the browser slot default-search-engine instead.

The motivating use case for this change is to provide modes an easy
method of changing the default search engine. In particular, users may
wish to default to an onion URL when proxy-mode is enabled with Tor.

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -113,8 +113,8 @@ bookmarks."
             (bookmark-search-engines))))
 
 (defun default-search-engine (&optional (search-engines (all-search-engines)))
-  "Return the search engine with the 'default' shortcut, or the first one if
-there is none."
-  (or (find "default"
+  "Return the search-engine that matches default-search-engine, or the first
+one if there is none."
+  (or (find (slot-value *browser* 'default-search-engine)
             search-engines :test #'string= :key #'shortcut)
       (first search-engines)))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -275,6 +275,12 @@ if there are errors, they will be reported by this function.")
                                        :documentation "When open links from an external program, or
 when C-cliking on a URL, decide whether to open in a new
 window or not.")
+   (default-search-engine :accessor default-search-engine
+                          :initform "default"
+                          :type string
+                          :documentation "Specify the default seacrh engine.
+The search-engine with this shortcut is used when none are explicitly specified
+and the query is not a valid URL.")
    (search-engines :accessor search-engines
                    :initform (list (make-instance 'search-engine
                                                   :shortcut "default"
@@ -290,9 +296,9 @@ You can invoke them from the minibuffer by prefixing your query with SHORTCUT.
 If the query is empty, FALLBACK-URL is loaded instead.  If
 FALLBACK-URL is empty, SEARCH-URL is used on an empty search.
 
-The engine with the \"default\" shortcut (or the first engine if there is no
-\"default\") is used when the query is not a valid URL, or the first keyword is
-not recognized.")
+The engine specified by default-search-engine (or the first engine if there is
+no \"default\") is used when the query is not a valid URL, or the first keyword
+is not recognized.")
    (key-stack :accessor key-stack :initform '()
               :documentation "A stack that keeps track of the key chords a user has pressed.")
    (downloads :accessor downloads :initform '()


xelxebar:
This is mostly just me getting familiar with the code base. Latent questions:

1) Do we even want this? Or is there already a good way of changing the default `search-engine` dynamically?
2) Is a new slot actually what we want, or is there a better, more general solution?
3) What are you thoughts on the name `default-search-engine`? (It unfortunately collides with the function of the same name in `source/urls.lisp`.)

source/urls.lisp
```diff
@@ -113,8 +113,8 @@ bookmarks."
             (bookmark-search-engines))))
 
 (defun default-search-engine (&optional (search-engines (all-search-engines)))
-  "Return the search engine with the 'default' shortcut, or the first one if
-there is none."
-  (or (find "default"
+  "Return the search-engine that matches default-search-engine, or the first
+one if there is none."
+  (or (find (slot-value *browser* 'default-search-engine)
```
jmercouris, line 118: since you made an accessor for the slot, you can use that in place of slot-value if you wish

jmercouris: there are good reasons for either approach

jmercouris:
> This is mostly just me getting familiar with the code base. Latent questions:
> 
> 1. Do we even want this? Or is there already a good way of changing the default `search-engine` dynamically?

You can change the default search engine dynamically by let binding or shadowing the default-search-engine function.

> 2. Is a new slot actually what we want, or is there a better, more general solution?

A new slot IS a good solution. The only other approach I can think of involves a magical keyword in a list like '(:default "url.com")

> 3. What are you thoughts on the name `default-search-engine`? (It unfortunately collides with the function of the same name in `source/urls.lisp`.)

It is not necessarily unfortunate. If you remove the accessor for default-search-engine, then you can make a method called `default-search-engine` which accepts a browser object and returns the default search engine with whatever logic you like. This way, you write a custom accessor.

jmercouris:
Firstly: thank you very much for this pull request!

My biggest suggestion would be to turn the `default-search-engine` function into a method and remove the slot accessor for `default-search-engine`. Interested in your feedback and thoughts!

source/urls.lisp
xelxebar: Unfortunately, the accessor `default-search-engine` collides with the parent function of the same name, so trying to use `(default-search-engine *browser*)` ends up in unbounded recursion.

xelxebar:
> My biggest suggestion would be to turn the `default-search-engine` function into a method and remove the slot accessor for `default-search-engine`. Interested in your feedback and thoughts!

Oh! This is nice. I will try making the change!

Ambrevar:
@jmercouris: A method specializing on what?  I think @xelxebar intention was to allow easy customization within modes, so I'm not sure a method would help much in that regard (against which mode would we specialize then?).

I'm not sure a new slot is the right approach.
- It creates a redundancy in the search engines.
- I don't see how it would allow customization from a mode.
- It adds a new slot which seems to be avoidable.  The less exported symbols, the better.

The original post raises a more general question: how do we customize the behaviour of the browser, windows, and buffers with modes?
I think it makes little sense to change the browser or windows settings with modes, since modes are buffer-local by definition.  Instead, we should use commands.  But maybe we can introduce window-local modes or global modes if the user wants to toggle settings and go back to a previous state.

For buffer settings, the current way to do it is to store the previous value in a mode slot, which is restored when the mode is disabled.  We do this in a few modes like vi-insert-mode.
The drawback is that this relies on mode activation and deactivation order.  For instance if modes A and B tweak setting S with original value `s` set to `sa` and `sb` respectively, then disabling A before B we will end up with value `sa` instead of `s`.
This looks like a difficult problem to solve.  We would need some form of stack of values, but I don't know how to do this elegantly for all slots.

This discussion makes me realize that search engines could be a buffer slot instead of a browser slot.  This way we could modify the search engines depending on the buffer-local modes like proxy-mode.

Finally, our current search-engines structure has the drawback that it can have multiple engines with the same keyword.  But maybe that's OK and we should just warn the user when they call a search engine with a duplicate keyword.

Ambrevar:
With the work that has been recently done on master and in #889, #890, we can now (or will soon be able to) locally override classes and have a history of class changes that's accessible at all times.  It's thus possible for a mode to modify the buffer defaults on activation, and restore the previous settings on deactivation.  

**TODO:** Document an example, such as the default search engine change when proxy-mode is on.

To finalize this patch, I suggest the following:
- Don't add any new slot.
- Move the `search-engines` slot to the buffer class.

Thoughts?

Ambrevar:
Additionally, maybe it would make more sense to make the _last_ search engine the default, while removing the "default" exception.

Ambrevar:
(It would also be easier to customize since the user would not have to deal with 2 rules to modify the default search engine.)

Ambrevar:
I've moved the search-engines slot to the BUFFER class in 17903a2e61e54482408968fc31a216f87d598841.

You can now customize this slot from the mode constructor / destructor, see `vi-normal-mode` for an example.

Besides, the default search engine is now the last one in the list, which should make it easier to hack.

I'm closing this since this should answer all the issues raised in the original post.  Feel free to reopen if you think I'm missing something.

Issue #868, opened 2020-07-27
auto-mode polishing: "main URL" predicate, mode-symbol type


Ambrevar:
As discussed in #846:

1. The typing of serialized mode is a bit ad-hoc.  It'd be nice to add a mode-symbol type.  We already have much of the logic implemented.

2. In `auto-mode-request-handler` we rely on the `(quri:uri= (url request-data) (url (buffer auto-mode)))` condition which is a bit obscure.  As far as I understand, this is the only way to test if the current request is the one the "main" one (the request for the current URL).  If so, we'd need to extract the condition to a well-named predicate to make this clearer.

3. Maybe improve logging.

Thoughts? @aartaka

Ambrevar:
4. Now the auto-mode-list file can only persist the exact list of modes unless I'm mistaken.  We need to add a way to persist the modes to include or exclude as was intiallly discussed.

The practical example is when we want to disable just one mode, say force-https-mode, but leave the rest of the modes to the default.  So if later we change the defaults modes, we won't have to update the whole list of 

I believe `save-modes-to-auto-mode-list` should use `:include-modes` and `:excludes-modes` by default.  We can leave the exact list of modes for manual editing of auto-mode-list.lisp.

Thoughts?

aartaka:
>  1. The typing of serialized mode is a bit ad-hoc. It'd be nice to add a mode-symbol type. We already have much of the logic implemented.

I don't quite get what this `mode-symbol` type should look like. In other words, what should its `satisfied` predicate look like?

>  2. In `auto-mode-request-handler` we rely on the `(quri:uri= (url request-data) (url (buffer auto-mode)))` condition which is a bit obscure.  As far as I understand, this is the only way to test if the current request is the one the "main" one (the request for the current URL).  If so, we'd need to extract the condition to a well-named predicate to make this clearer. 

That's a good idea! It can be useful for the end-user too, because then custom handlers for `request-resource-hook` can be more flexible.

You've mentioned in [this comment](https://github.com/atlas-engineer/nyxt/pull/846#discussion_r460760813) that link clicks are somewhat outside the scope of this condition. Is it because all the clicks are passed to the renderer (not sure abour that, need to look it up in the codebase)? Should it be included in this predicate for "main" URL requests then? 

[...]

PR #875, opened 2020-07-30
Fix match-url non-matching in auto-mode

aartaka:
This fixes the problem with `match-url` never wofking on `auto-mode`-saved URLs, mentioned in #868.

### Roots

The root of the problem was `object-string` used instead of `object-display` in `auto-mode` code.

### Caveats

As a precaution against the decoding issues, all the URLs passed to `match-url` are now decoded. Can it cause any problem?


jmercouris:
I don't believe it can cause any problems. Thanks for the fix!

PR #876, opened 2020-07-30
Inheritance rework

jmercouris:
Architectural simplifications:

+ removes need for special variables


source/buffer.lisp
```diff
@@ -238,6 +238,9 @@ The handlers take the buffer as argument.")
 Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
 `:no-third-party' (accept cookies for current website only).")))
 
+(defclass internal-buffer (gtk-buffer) ()
```
jmercouris, line 241: This works, but is problematic. It is not agnostic of the renderer. I thought about renaming both classes in both files to `renderer-buffer` and then just inheriting from `renderer-buffer`, there must be a better way to do this.

source/utility.lisp
```diff
@@ -185,45 +184,29 @@ In the above, `%slot-default' will be substituted with the default value of
...
 
 To discover the default value of a slot or all slots of a class, use the
-`describe-slot' or `describe-class' commands respectively.
-
-Since modes are classes with class variables (the `*MODE-class*'), the same applies.
-
-Example to get the `blocker-mode' command to use a new default hostlists:
-
-\(define-configuration nyxt/blocker-mode:blocker-mode
-  ((nyxt/blocker-mode:hostlists (append (list *my-blocked-hosts*) %slot-default))))
```
Ambrevar, line 195: I think you should keep the example.

Ambrevar:
Lots of good stuff in there!  Actually, if I understand correctly, this pull request regroups 3 unrelated changes:

- The internal buffer class.
- The defgeneric trick.
- The define-configuration update.

It's easier if you send these changes separately ;)

The defgeneric trick can be merge on master right away, good catch!

About define-configuration: this is smart indeed, I hadn't thought about replacing the class.  In Next 1.5 we were editing the initforms of the slots which raised many issues, but replacing the class is indeed the clean way to do this!
Plus this fixes a long standing issue: we can now use `define-configuration` over the same class multiple times in the user init file!

The main drawback is that we lose the default (vanilla) values.  So I suggest we backup the vanilla classes (and only them) in `default-CLASS` for instance.

Ambrevar:
Something else that we lose with the define-configuration change is the ability to override the class locally.
So I suggest we keep the `*CLASS-class*` variable and use it to instantiate the classes, but we don't use `*user-CLASS*`.

Thoughts?

Ambrevar:
Maybe there is no need for a `default-CLASS`: the parent class is recorded in the class, so we can access the default class by going up the list of `mopu-:direct-superclasses` until the last buffer|window|browser class, which should be guaranteed to be the default.

source/buffer.lisp
jmercouris: I've thought about this and I believe the internal buffer needs to be extended in the gtk-renderer.lisp file as is done with the regular buffer.

source/utility.lisp
jmercouris: OK! will do!

jmercouris:
> Lots of good stuff in there! Actually, if I understand correctly, this pull request regroups 3 unrelated changes:
>
> * The internal buffer class.
> * The defgeneric trick.
> * The define-configuration update.

Correct!
 
> It's easier if you send these changes separately ;)
> 
> The defgeneric trick can be merged on master right away, good catch!

OK!

> About define-configuration: this is smart indeed, I hadn't thought about replacing the class. In Next 1.5 we were editing the initforms of the slots which raised many issues, but replacing the class is indeed the clean way to do this!
> Plus this fixes a long standing issue: we can now use `define-configuration` over the same class multiple times in the user init file!
> 
> The main drawback is that we lose the default (vanilla) values. So I suggest we backup the vanilla classes (and only them) in `default-CLASS` for instance.

As you have hinted in later comments, you can just keep referencing the super class to get the original "vanilla classes", so there is no need to maintain a reference to them.

source/buffer.lisp
Ambrevar: Yes, after all it's a one liner and we probably won't have more than 3-4 renderers ever.

Ambrevar:
What do you think about https://github.com/atlas-engineer/nyxt/pull/876#issuecomment-667515028 ?

jmercouris:
I don't think we have a use case for overriding the class locally. Typically one doesn't shadow classes... as it could be very dangerous if a fish object is pretending to be a seagull object- for example.

Ambrevar:
I don't think it's dangerous: the methods would be correctly dispatched
anyways.

The use case would be for command/mode-local settings such as in #874:
allow a command/mode to specify the default search-engine within its
context.

Another one is the upcoming ephemeral buffer support: The related
commands will want to locally set the buffer class to that of the
ephemeral class.

jmercouris:
If I understand correctly, that could be accomplished by a macro like with-class. This class could set and reset the class for the @body of the macro. Thoughts?

Ambrevar:
Sounds like a great idea indeed!
I wonder why I've never seen such a macro before.

This macro should cut it:

```lisp
(defmacro with-class ((class-sym override-sym) &body body)
  (alex:with-gensyms (old-class)
    `(let ((,old-class (find-class ',class-sym)))
       (unwind-protect
            (progn
              (setf (find-class ',class-sym) (find-class ',override-sym))
              ,@body)
         (setf (find-class ',class-sym) ,old-class)))))
```

Test:

```lisp
(defclass foo () ((name :initarg :name :initform "fooname")))
(defclass bar () ((name :initarg :name :initform "barname")))

(with-class (foo bar)
 (let ((b (make-instance 'foo)))
  (slot-value b 'name)))
; "barname"

(let ((b (make-instance 'foo)))
  (slot-value b 'name))
; "fooname"
```

jmercouris:
Great!

then, I shall:

+ add the macro
+ remove the unrelated internal buffer code
+ add the internal buffer code to another branch for development and review
+ merge this branch

Ambrevar:
Please do! :)

jmercouris:
OK! Done! I will do the internal-buffer changes on a separate branch! Now I at least have clarity on their implementation, so it should be relatively quick. Then we'll be able to add user clickable buttons into our bookmarks page for example! :-) that will be so cool!!!

Ambrevar:
Thanks!

Ambrevar:
Actually the merge triggers many warnings...  Any idea what's going on?
I'll investigate.

jmercouris:
They are because of the way we load the systems, if you define standard methods, they can go away.

jmercouris:
That is, the defgenerics cause the warnings because I removed a default "implementation".

jmercouris:
If you revert the defgeneric commit you can see what I mean :-)

We could still provide a default implementation if the warnings bother you...

jmercouris:
Sorry, it is because we do not export the symbols and use the defgeneric only!

jmercouris:
fixed! good catch, I thought I had fixed the warnings but didn't check after reverting my changes!

PR #881, opened 2020-08-04
Rename auto-mode-list to auto-mode-rules

aartaka:
This changes the names of the `auto-mode-list`, `auto-mode-list-data-path` and related functions&commands (ending with `-auto-mode-list`) in `auto-mode`, to `auto-mode-rules`, `auto-mode-rules-data-patch` and `*-auto-mode-rules` (i.e. names ending with `-auto-mode-rules`).

Additionally, there are several minor code/formatting changes like small docstring fixes and serialization of modes' list without `symbol-name` and `string-downcase`.

### Motivation

`auto-mode-list` can sound similar to Emacs' `auto-mode-alist`, but not all the Nyxt userbase is using Emacs, so it's better to stick to a more straightforward name. `*-auto-mode-rules` names fit the situation better, because these are `auto-mode-rule` objects that we manipulate with these variables.

### Caveats

You need to manually rename auto-mode-list.lisp to auto-mode-rules.lisp in your Nyxt data directory to keep the `auto-mode` rules accumulated there for future sessions.

### How Has It Been Tested

I renamed the rules file and ran Nyxt. I've added several rules for different domains, and they were stored. I was browsing the matching and non-matching websites, and it all was working well. 

Given that this change is mainly just a variable renaming, it's not going to be extremely buggy (and I've `grep`-d it to make sure that there are no more `*-auto-mode-list` names in the code).

I'll be glad if you test it and say what you think!

Related to discussion in #868.

EDIT: add the reference to the issue.

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -3,19 +3,19 @@
   (:import-from #:serapeum #:export-always)
   (:documentation "Mode for automatic URL-based mode toggling."))
 (in-package :nyxt/auto-mode)
-(eval-when (:compile-toplevel :load-toplevel :execute)
-  (trivial-package-local-nicknames:add-package-local-nickname :alex :alexandria))
 
 (export-always '*prompt-on-mode-toggle*)
 (defvar *prompt-on-mode-toggle* nil
   "Whether user will be asked about adding the mode to included/excluded modes
-in auto-mode-list on mode activation/deactivation.")
+in `auto-mode-rules' on mode activation/deactivation.")
 
 (export-always '*non-rememberable-modes*)
 (defvar *non-rememberable-modes*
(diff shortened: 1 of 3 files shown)


Ambrevar:
Excellent, thanks!

Talking about grepping, do you know about the `sly-who-*` commands?  They allow to semantically find the call sites, setting sites, etc.  This is a good way to refactor semantically.

Another useful Emacs package is wgrep combined with Helm or Ivy: you can grep a term, save the resulting minibuffer and edit the results directly.  This is awesome in particular when some changes must be done contextually.

aartaka:
I've used `sly-who-*` commands, but haven't heard of wgrep. I'll give it a look, thanks!

PR #883, opened 2020-08-06
auto-mode: Make auto-mode dynamic & extract auto-mode-handler to a new hook

### LINKED ISSUE #868: auto-mode polishing: "main URL" predicate, mode-symbol type

aartaka:
This refactors `auto-mode` to make it more stable and enables somewhat dynamic default behavior at the cost of minor backwards-incompatibility.

### What's New (numbered as in #868):
- [**7**](https://github.com/atlas-engineer/nyxt/issues/868#issuecomment-664440162): `pre-request-hook` made exclusively for `auto-mode` and ran right before `request-resource-hook`. Putting `auto-mode-handler` into `request-resource-hook` was dangerous, because `auto-mode` modifies the contents of `request-resource-hook` and cannot control the order of handlers in it. This makes for the unpredictable behavior, as described in #868.
- [**4**](https://github.com/atlas-engineer/nyxt/issues/868#issuecomment-664311865): New commands, `enable-modes-for-future-visits` and `disable-modes-for-future-visits`, to enable the dynamic control of the enabled\disabled modes (e.g. modes' inclusion and exclusion that doesn't define the exact mode list, but only states that some modes should be there and some shouldn't).
- [**2**](https://github.com/atlas-engineer/nyxt/issues/868#issue-666172472): Several conditions are now extracted to the dedicated predicates for debugging and ease of reading. Examples are `can-store-last-active-modes`, `new-page-request-p`, and `history-empty-p`.
- `save-modes-to-auto-mode-list` is now called `save-exact-modes-for-future-visits` and saves an exact list of modes, much like it was in the initial version of auto-mode.
- `add-modes-to-auto-mode-rules` is now a Swiss knife of mode storage and has a lot of keyword arguments that help to put most repetitive mode control tasks in one place.
- User-oriented rename:
  - `save-modes-to-auto-mode-list` → `save-exact-modes-for-future-visits` (new commands are named in the same style).
  - `make-dwim-match` → `url-infer-match`.
- Documentation fixes, examples and additions.

### What's Not There (Yet) (numbered as in #868)
- [**1**](https://github.com/atlas-engineer/nyxt/issues/868#issue-666172472): There's no `mode-symbol` type yet, because I've seen no strict necessity in it while making these changes — the existing set of functions is comfortable enough. I don't deny the possible necessity of `mode-symbol` though. We need to discuss it a bit more, because now the advantages and implementation are unclear to me.
- [**5**](https://github.com/atlas-engineer/nyxt/issues/868#issuecomment-664332178): There's no mode confirmation dialog, because it's still unclear to me what should trigger this dialog and why should it be there. Once I understand the value of this dialog (possibly due to discussion with use cases and examples), I'll be glad to implement it!

### Caveats and Workarounds

Storage format have changed significantly. Now, If you want to use your old `auto-mode` rules with new `auto-mode`, you need to make some changes to your auto-mode-rules.lisp:
- Replace `:modes` keyword with `:included` keyword.
- Add `:exact-p t` phrase to the end of every rule.

These changes will reproduce the same behavior as previous version of `auto-mode`.

However, if you want to use new features, like  dynamic mode control, you'll need to either delete/rewrite some of the rules by browsing (I'm not sure it's going to be smooth...), or fix them manually:
- Replace `:modes` keyword with `:included` keyword. Delete modes you already have in your `default-modes`, from the list. 
  - E.g., if your rule was `("foo.bar" :modes (emacs-mode blocker-mode proxy-mode))`, but `emacs-mode` and `blocker-mode` are already in your defaults, then your rule can be shortened to `("foo.bar" :included (proxy-mode))`!
- (optional) Add `:excluded` keyword with a list of modes you don't want to ever see enabled for a given rule.
- (optional) Add `:exact-p t` if you want the rule to enable exactly the `:included` modes and nothing else.

### How Has This Been Tested

- I've loaded Nyxt to a fresh REPL and ran it for an hour, creating rules and testing edge-cases discussed in #868. It seems to work pretty well. Some behaviors have changed, so feedback on usability and/or bugs is welcome!

Possibly closes #868. Look Caveats for something to discuss before merging and closing the issue.

EDIT: add link to the possibly solved issue.

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -11,7 +11,7 @@ in `auto-mode-rules' on mode activation/deactivation.")
 
 (export-always '*non-rememberable-modes*)
 (defvar *non-rememberable-modes*
-  ;; Base mode conflicts with its Nyxt symbol if it's not prefixed
+  ;; base-mode conflicts with its Nyxt symbol if it's not prefixed
   '(help-mode web-mode auto-mode nyxt::base-mode)
   "Modes that `auto-mode' won't even try to save.
 Append names of modes you want to always control manually to this list.
@@ -38,10 +38,22 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
          :accessor test
          :type list
          :initform (error "Slot `test' should be set."))
-   (modes :initarg :modes
-          :accessor modes
-          :type list-of-symbols
-          :initform '())))
+   (included :initarg :included
+             :accessor included
+             :type list-of-symbols
+             :initform '()
+             :documentation "The list of mode symbols to enable on rule activation.")
+   (excluded :initarg :excluded
+             :accessor excluded
+             :type list-of-symbols
+             :initform '()
+             :documentation "The list of mode symbols to disable on rule activation.")
+   (exact-p :initarg :exact-p
+            :accessor exact-p
+            :type boolean
+            :initform nil
+            :documentation "If non-nil, enable the INCLUDED modes exclusively.
+Enable INCLUDED modes plus the already present ones, and disable EXCLUDED modes, if nil.")))
 
 (declaim (ftype (function (quri:uri) (or auto-mode-rule null))
                 matching-auto-mode-rule))
@@ -60,27 +72,26 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
 (declaim (ftype (function (quri:uri buffer)) enable-matching-modes))
 (defun enable-matching-modes (url buffer)
   (let ((rule (matching-auto-mode-rule url)))
-    (enable-modes (rememberable-of (set-difference
-                                    (modes rule)
-                                    (mapcar #'mode-name (modes buffer))
-                                    :test #'mode-equal))
+    (enable-modes (set-difference
+                   (included rule)
+                   (rememberable-of (mapcar #'mode-name (modes buffer)))
+                   :test #'mode-equal)
                   buffer)
-    (disable-modes (rememberable-of (set-difference
-                                     (mapcar #'mode-name (modes buffer))
-                                     (modes rule) :test #'mode-equal))
+    (disable-modes (if (exact-p rule)
+                       (set-difference
+                        (rememberable-of (mapcar #'mode-name (modes buffer)))
+                        (included rule) :test #'mode-equal)
+                       (excluded rule))
                    buffer)))
 
-(defun clean-up-auto-mode (mode)
-  (hooks:remove-hook (request-resource-hook (buffer mode))
-                     'auto-mode-request-handler)
-  (hooks:remove-hook (enable-mode-hook (buffer mode))
-                     'enable-mode-auto-mode-handler)
-  (hooks:remove-hook (disable-mode-hook (buffer mode))
-                     'disable-mode-auto-mode-handler))
+(defun can-store-last-active-modes (auto-mode url)
+  (or (null (last-active-modes-url auto-mode))
+      (not (quri:uri= url (last-active-modes-url auto-mode)))))
 
-(defun store-last-active-modes (auto-mode)
-  (setf (last-active-modes auto-mode)
-        (modes (buffer auto-mode))))
+(defun store-last-active-modes (auto-mode url)
+  (when (can-store-last-active-modes auto-mode url)
+      (setf (last-active-modes auto-mode) (modes (buffer auto-mode))
+            (last-non-matching-url auto-mode) url)))
 
 (defun restore-last-active-modes (auto-mode)
   (disable-modes
@@ -96,61 +107,92 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
                            :test #'mode-equal))
    (buffer auto-mode)))
 
-(defun auto-mode-request-handler (request-data)
+(defun new-page-request-p (request-data)
+  "Whether the REQUEST-DATA is a request for a new page load.
+Resource/font/ads/anchor loads are safely ignored.
+
+It relies on the fact that, due to the WebKit limitations, we store the loaded
+URL in the buffer slot when we need to load a new page, while, for
+non-new-page requests, buffer URL is not altered."
+  (quri:uri= (url request-data) (url (buffer request-data))))
+
+(defun history-empty-p (history)
+  (eq (htree:root history) (htree:current history)))
+
+(defun auto-mode-handler (request-data)
   (let* ((auto-mode (find-submode (buffer request-data) 'auto-mode))
-         ;; We need to supress prompting when auto-mode modifies modes.
+         ;; We need to suppress prompting when auto-mode modifies modes.
          (*prompt-on-mode-toggle* nil)
          (web-mode (find-submode (buffer request-data) 'web-mode))
          (previous-url
-           (unless (eq (htree:root (history web-mode))
-                       (htree:current (history web-mode)))
+           (unless (history-empty-p (history web-mode))
              (url (htree:data
-                   (htree:parent (htree:current (history web-mode))))))))
-    (if (matching-auto-mode-rule (url request-data))
-        (progn
-          (when (and previous-url (not (matching-auto-mode-rule previous-url)))
-            (store-last-active-modes auto-mode))
-          (enable-matching-modes (url request-data) (buffer request-data)))
-        ;; Apply previous saved modes only on buffer-loads, not anytime else.
-        ;;
-        ;; TODO: Relies on somewhat illogical behavior of (url (current-buffer))
-        ;; and (url request-data) being equal after buffer-load is called.
-        ;; Replace with more stable and logical condition?
-        (when (quri:uri= (url request-data) (url (buffer auto-mode)))
-          (restore-last-active-modes auto-mode))))
+                   (htree:parent (htree:current (history web-mode)))))))
+         (rule (matching-auto-mode-rule (url request-data)))
+         (previous-rule (when previous-url (matching-auto-mode-rule previous-url))))
+    (when (and rule previous-url (not previous-rule))
+      (store-last-active-modes auto-mode previous-url))
+    (cond
+      ((and (not rule) (new-page-request-p request-data))
+       (restore-last-active-modes auto-mode))
+      ((and rule (not (equal rule previous-rule)))
+       (enable-matching-modes (url request-data) (buffer request-data)))))
   request-data)
 
-(defun mode-covered-by-auto-mode-p (mode auto-mode-instance)
+(defun mode-covered-by-auto-mode-p (mode auto-mode enable-p)
   (or (non-rememberable-mode-p mode)
       (let ((matching-rule (matching-auto-mode-rule
-                            (url (buffer auto-mode-instance)))))
-        (member mode (or (and matching-rule (modes matching-rule))
-                         (last-active-modes auto-mode-instance))
+                            (url (buffer auto-mode)))))
+        (member mode (or (and matching-rule (union (included matching-rule)
+                                                   (excluded matching-rule)))
+                         ;; Mode is covered by auto-mode only if it is
+                         ;; in last-active-modes and gets enabled.
+                         ;; If it gets disabled, user should be prompted,
+                         ;; because they may want to persist it.
+                         (and enable-p (last-active-modes auto-mode)))
                 :test #'mode-equal))))
+
+(export-always 'url-infer-match)
+(declaim (ftype (function (string) list) url-infer-match))
+(defun url-infer-match (url)
+  "Infer the best `test' for `auto-mode-rule', based on the form of URL.
+The rules are:
+- If it's a plain domain-only URL (i.e. \"http://domain.com\") -- then use `match-domain'.
+- If it's host with subdomains (\"http://whatever.subdomain.domain.com\") -- use `match-host'.
+- Use `match-url' otherwise."
+  (let ((url (or (ignore-errors (quri:uri url)) url)))
+    (if (and (quri:uri-p url)
+             (empty-path-url-p url)
+             (host-only-url-p url))
+        (if (string= (quri:uri-domain url)
+                     (cl-ppcre:regex-replace "www[0-9]?\\." (quri:uri-host url) ""))
+            `(match-domain ,(quri:uri-domain url))
+            `(match-host ,(quri:uri-host url)))
+        `(match-url ,(object-display url)))))
 
 (declaim (ftype (function (boolean t) (function (root-mode)))
                 make-mode-toggle-prompting-handler))
-(defun make-mode-toggle-prompting-handler (enable-p auto-mode-instance)
+(defun make-mode-toggle-prompting-handler (enable-p auto-mode)
   #'(lambda (mode)
-      (when (and (not (mode-covered-by-auto-mode-p mode auto-mode-instance))
+      (when (and (not (mode-covered-by-auto-mode-p mode auto-mode enable-p))
                  *prompt-on-mode-toggle*)
-        (with-confirm ("Permanently ~a ~a for this URL?"
-                       (if enable-p "enable" "disable")
-                       (mode-name mode))
+        (if-confirm ("Permanently ~:[disable~;enable~] ~a for this URL?"
+                       enable-p (mode-name mode))
           (with-result (url (read-from-minibuffer
                              (make-minibuffer
                               :input-prompt "URL:"
                               :input-buffer (object-display (url (buffer mode)))
                               :must-match-p nil)))
-            (let* ((test (make-dwim-match url))
-                   (rule (find test (auto-mode-rules *browser*)
-                               :key #'test :test #'equal))
-                   (rule-modes (if rule (modes rule) (modes (buffer mode))))
-                   (modes (if enable-p
-                              (union (list mode) rule-modes :test #'mode-equal)
-                              (set-difference rule-modes (list mode)
-                                              :test #'mode-equal))))
-              (add-modes-to-auto-mode-rules test modes)))))))
+            (add-modes-to-auto-mode-rules (url-infer-match url)
+                                          :append-p t
+                                          :include (when enable-p (list mode))
+                                          :exclude (unless enable-p (list mode))))
+          (setf (last-active-modes auto-mode)
+                (if enable-p
+                    (union (list mode) (last-active-modes auto-mode)
+                           :test #'mode-equal)
+                    (remove mode (last-active-modes auto-mode)
+                            :test #'mode-equal)))))))
 
 (defun initialize-auto-mode (mode)
   (unless (last-active-modes mode)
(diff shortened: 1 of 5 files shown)


source/auto-mode.lisp
```diff
@@ -96,43 +111,81 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
...
-        ;; TODO: Relies on somewhat illogical behavior of (url (current-buffer))
-        ;; and (url request-data) being equal after buffer-load is called.
-        ;; Replace with more stable and logical condition?
-        (when (quri:uri= (url request-data) (url (buffer auto-mode)))
+          (when (and previous-url (not previous-rule))
+            (store-last-active-modes auto-mode previous-url)
+            (log:info "Stored last active modes: ~a. Previous link: ~s. Current link: ~s."
+                      (last-active-modes auto-mode)
+                      previous-url
+                      (url request-data)))
```
Ambrevar, line 144: This should probably be a log:debug, no?

source/auto-mode.lisp
```diff
@@ -96,43 +111,81 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
...
+          (unless (equal rule previous-rule)
+            (log:info "Enabled matching rules: ~a." (test rule))
+            (enable-matching-modes (url request-data) (buffer request-data))))
+        (when (new-page-request-p request-data)
+          (log:info "Restored last active modes: ~a." (last-active-modes auto-mode))
```
Ambrevar, line 149: Same.

source/auto-mode.lisp
```diff
@@ -260,12 +381,14 @@ For the storage format see the comment in the head of your `auto-mode-rules-data
     (write-string ";; List of auto-mode rules.
 ;; It is made to be easily readable and editable, but you still need to remember some things:
 ;;
-;; Every rule starts on a new line and consists of two parts:
+;; Every rule starts on a new line and consists of one to four parts:
```
Ambrevar, line 384: More future-proof: "consists of one or more of the following elements".

source/buffer.lisp
```diff
@@ -128,6 +129,14 @@ forwarded when no binding is found.")
...
                ;; TODO: Store multiple events?  Maybe when implementing keyboard macros.
                :documentation "The last event that was received for the current buffer.")
+   (pre-request-hook :accessor pre-request-hook
+                     :initarg :pre-request-hook
+                     :initform (make-hook-resource
+                                :combination #'combine-composed-hook-until-nil)
+                     :type hook-resource
+                     :documentation "Hook run before the `request-resource-hook'.
+For now, its only application is in `auto-mode', but any action on modes that can
+possibly change the handlers in `request-resource-hook' should find its place there.")
```
Ambrevar, line 139: More future-proof: give `auto-mode' as an example, no need to say it's the only application.

source/auto-mode.lisp
```diff
@@ -38,10 +38,22 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
...
+   (excluded :initarg :excluded
+             :accessor excluded
+             :type list-of-symbols
+             :initform '()
+             :documentation "The list of mode symbols to disable on rule activation.")
+   (exact-p :initarg :exact-p
+            :accessor exact-p
+            :type boolean
+            :initform nil
+            :documentation "Enable the INCLUDED modes exclusively if non-nil.
```
Ambrevar, line 55: Maybe move "if non-nil" to avoid a possible ambiguity with that the modes are enabled only if exact-p is non-nil.

source/auto-mode.lisp
```diff
@@ -164,79 +212,152 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
...
+  (hooks:remove-hook (disable-mode-hook (buffer mode))
+                     'disable-mode-auto-mode-handler))
 
 (define-mode auto-mode ()
   "Remember the modes setup for given domain/host/URL and store it in an editable form.
 These modes will then be activated on every visit to this domain/host/URL."
-  ((last-active-modes :accessor last-active-modes
+  ((last-non-matching-url :accessor last-non-matching-url
+                          :type (or quri:uri null)
+                          :initform nil)
```
Ambrevar, line 231: Maybe document why we need to store this.

source/auto-mode.lisp
```diff
@@ -96,43 +111,81 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
...
-        (member mode (or (and matching-rule (modes matching-rule))
-                         (last-active-modes auto-mode-instance))
+                            (url (buffer auto-mode)))))
+        (member mode (or (and matching-rule (union (included matching-rule)
+                                                   (excluded matching-rule)))
+                         ;; Mode is covered by auto-mode only if it is
+                         ;; in last-active-modes and gets enabled.
+                         ;; If it gets disabled, user should be prompted,
+                         ;; because they may want to persist it.
+                         (when enable-p (last-active-modes auto-mode)))
```
Ambrevar, line 163: To better match the comment, maybe change `when` for `and`.

source/auto-mode.lisp
```diff
@@ -164,79 +212,152 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
...
-        (if (string= (quri:uri-domain url)
-                     (cl-ppcre:regex-replace "www[0-9]?\\." (quri:uri-host url) ""))
-            `(match-domain ,(quri:uri-domain url))
-            `(match-host ,(quri:uri-host url)))
-        `(match-url ,(object-display url)))))
+(defun active-rememberable-mode-suggestion-filter (buffer)
+  (let ((modes (delete-duplicates
+                (rememberable-of
+                 (mapcar #'mode-name
+                         (modes buffer))))))
```
Ambrevar, line 242: We can save the newline in the mapcar here.

source/auto-mode.lisp
```diff
@@ -164,79 +212,152 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
...
+  (with-result (url (read-from-minibuffer
+                     (make-minibuffer
+                      :input-prompt "URL:"
+                      :input-buffer (object-string (url (current-buffer)))
+                      :suggestion-function (nyxt::history-suggestion-filter
+                                            :prefix-urls (list
+                                                          (object-string
+                                                           (url (current-buffer)))))
+                      :history (minibuffer-set-url-history *browser*)
+                      :must-match-p nil)))
```
Ambrevar, line 315: Why not prompting for the list of modes here?

Ambrevar:
Fantastic work!

Have you tested with force-https-mode enabled by default and disabling it for select URLs?

About `enable-modes-for-future-visits` and `disable-modes-for-future-visits`: Why not have a single command instead that saves the difference?  I.e. the missing default modes get added to `:exclude` and the active non-default modes get added to `:include`.

Ambrevar:
About the other points:

1\.: For now I don't see any pressing reason to have a mode-symbol type.  This can wait.
5\.: This can wait for the next pull request.  I'll give you an example right away anyways it this can help understand what I mean.

When `*prompt-on-mode-toggle*` is nil, modes are enabled and disabled automatically without any user intervention.  But humans make mistake and are prone to forget which mode is they've toggled manually and might make the assumption that some mode is going to be enabled when clicking on a URL.  Critical examples include force-https-mode and proxy-mode: if these modes happen to be disabled without the user realizing it, it could violate the user's privacy.

Example of undesirable behiour.
- User enables proxy-mode because they want to browse privately _from now on_.
- User browses a URL which has a rule that disables proxy-mode.  Proxy-mode gets disabled without prompting the user.  Even if the user realizes it by looking at the status view, it's too late, their privacy has already been violated.

Example of possibly better behaviour.
- User enables proxy-mode as above.
- User browses a URL which has a rule that disables proxy-mode.  This time the browser stop during the request and prompt the user for confirmation.  The user can choose to
  - ignore the rule
  - go with the rule
  - choose the modes they want from the full list of modes, with the rule modes being pre-selected.

aartaka:
> Have you tested with force-https-mode enabled by default and disabling it for select URLs?

Yes, I tested it. You still need to edit "https" to "http" manually if the site doesn't have the TLS-enabled version, but nothing else stops things. That's how my sbcl.org rule looks like (and it works okay if I don't forget to type exactly http://sbcl.org/):
``` lisp
((match-domain "sbcl.org")  :excluded (proxy-mode force-https-mode))
```
 
> About `enable-modes-for-future-visits` and `disable-modes-for-future-visits`: Why not have a single command instead that saves the difference? I.e. the missing default modes get added to `:exclude` and the active non-default modes get added to `:include`.

Oh, that's smart! I haven't thought about that. 

However, I doubt it's exactly `default-modes` that we need to match included/excluded modes against. For me, matching against `last-active-modes` is more intuitive, because it's basically what user have chosen for their usual browsing for now. 

`last-active-modes` are too dynamic and hard to track though. I'll think more on how to make it work.

aartaka:
This example with `proxy-mode` makes sense, but do we really want to prompt the user on every match? Depending on the browsing style and number of `auto-mode` rules, prompting can happen quite frequently and irritate the user. 

Adding yet another special variable (`*prompt-modes-on-match*`?) sounds like the way here, because this behavior is important (thanks for the example illuminating the importance of it!), but not everyone would want to use it.

source/auto-mode.lisp
aartaka: It saves exact list of modes here, so there's no need in prompting.

aartaka:
`last-non-matching-url` have appeared because I discovered that `pre-request-hook` and `request-resource-hook` can activate more than one time per page load. 

When the first `pre-request-hook` fires with `auto-mode-handler`, it saves the `last-active-modes` and possibly enables new modes. However, when the second/third/N-th `pre-request-hook` fires, it saves previously enabled modes as `last-active-modes`. This destroys all the purpose of `last-active-modes`.

I added the condition that it should not be the same URL for mode storage to `last-active-modes`, as a workaround. I needed to store it somewhere, that's why `last-non-matching-url` appeared.

Ambrevar:
> I added the condition that it should not be the same URL for mode storage to `last-active-modes`, as a workaround. I needed to store it somewhere, that's why `last-non-matching-url` appeared.

Makes sense now.  Can you add this explanation to the docstring?

source/auto-mode.lisp
Ambrevar: Maybe I'm thinking a bit too sophisticated here, but the way I see it the current list of modes is not logically bound to the exact list of persisted modes.  The user could very well want to exclude some FOO mode from the `exact-p` rule while keeping FOO for the current browsing.

But that's a detail, I'm not sure anyone will ever need this.  Maybe add a TODO: comment about this.

Ambrevar:
> Adding yet another special variable (`*prompt-modes-on-match*`?) sounds like the way here, because this behavior is important (thanks for the example illuminating the importance of it!), but not everyone would want to use it.

Absolutely.

Ambrevar:
>> About `enable-modes-for-future-visits` and `disable-modes-for-future-visits`: Why not have a single command instead that saves the difference? I.e. the missing default modes get added to `:exclude` and the active non-default modes get added to `:include`.
>
> Oh, that's smart! I haven't thought about that. 
>
> However, I doubt it's exactly `default-modes` that we need to match included/excluded modes against. For me, matching against `last-active-modes` is more intuitive, because it's basically what user have chosen for their usual browsing for now. 
>
> `last-active-modes` are too dynamic and hard to track though. I'll think more on how to make it work.

Both scenarios make sense, but I'm leaning towards comparing with
default modes because it translates the following: "replay my current
settings next time I start the browser".

Example:

- With last-active-modes, proxy-mode (not default) is active at foo.org,
  then the user goes to bar.org, enables noscript-mode and saves the
  rule.  noscript-mode is persisted but not proxy-mode.  Next time the
  user opens the browser and goes to bar.org, only noscript-mode will be
  active.

- With default modes and the above example, both proxy-mode and
  noscript-mode are persisted.  Next time the user opens the browser and
  goes to bar.org, both proxy-mode and noscript-mode will be active.

In the second case, the user gets the exact same mode settings, and thus
_reproduces_ the user experience from one browse to the next.  I think
it's an important property.

aartaka:
>     * User browses a URL which has a rule that disables proxy-mode.  This time the browser stop during the request and prompt the user for confirmation.  The user can choose to
>       
>       * ignore the rule
>       * go with the rule
>       * choose the modes they want from the full list of modes, with the rule modes being pre-selected.

Okay, this example is totally valid, but it exposes that `with-confirm` is somewhat lacking the ability to have several branches, depending on the use decision. Sometimes you need to do some actions exactly because user answers "no", and this case is not covered by `with-confirm`. I'm reproducing the body of `with-confirm` for the second time in a row, and it already looks fishy.

Should we expand `with-confirm` to support at least two branches?

aartaka:
Hmmmm... 
``` lisp
(if (with-confirm ("Do FOO on yes, do BAR on no.") t)
    foo
    bar)
```

Seems to do the trick, but is it clear enough?

Ambrevar:
Instead I would let-bind the choices to (:yes "Do FOO" :no "Do BAR").

 > Should we expand `with-confirm` to support at least two branches?

It already has 2 branches.  Did you mean more than 2?
It's trivial to pass a list of string as minibuffer suggestions, than we
can match against them.

Untested:

```
(with-result (result (read-from-minibuffer
                       (make-minibuffer
                        :suggestion-function (lambda () '("foo" "bar" "baz")))))
  (match result
    ("foo" ...)
    ("bar" ...)
    ("baz" ...)))
```

What do you think?

aartaka:
Yes, `with-confirm` has two branches, but one of them is `nil` in any case, which doesn't add much of a meaning. What I'm trying to suggest is to have two meaningful branches in `with-confirm`, like in this imaginary example with `*prompt-modes-on-match*`:
``` lisp 
(if *prompt-modes-on-match*
    (with-confirm ("Set ~a as the modes of the current URL?"
                   (modes rule))
      ;; Yes-branch -- apply the rule
      (enable-matching-modes (url request-data) (buffer request-data))
      ;; No-branch -- prompt the user for the modes to activate
      (with-result
          (modes (read-from-minibuffer
                  (make-minibuffer
                   :input-prompt "Enable mode(s)"
                   :multi-selection-p t
                   :suggestion-function (nyxt::inactive-mode-suggestion-filter
                                         (list (buffer request-data))))))
        (enable-modes modes)))
    ;; Apply the rule silently  if *prompt-modes-on-match* is nil
    (enable-matching-modes (url request-data) (buffer request-data)))
```

While writing a custom minibuffer query can be a way out, it is more straightforward to have the choice between "yes" and "no" processed by `with-confirm`.

Ambrevar:
Oh, I understand now!

Indeed, we could have an optional `else` branch in the with-confirm.
Feel free to update it.

You mentioned another copy of the body of with-confirm: where did it occur?

aartaka:
It occured in the `make-mode-toggle-prompting-handler`-generated function. I needed to add the non-persisted mode to the `last-active-modes` here, because it's DWIM-ish and user would probably want to stay with the mode they have permanently enabled. Example:
- One navigates to foo.org, where they have a rule enabling `proxy-mode`.
- They enable `noimage-mode`, because their connection is bad today.
- But they don't want to persist it, because it's just for today or even for one browsing session. They choose "no" in the prompting handler.
- Then, there is a split of behaviors:
   - In the current version, `noimage-mode` is just added to `(modes (current-buffer))`, but it is disabled on any non-matching URL, because it's not on `last-active-modes`.
   - In my dream version, `noimage-mode` is stored in `last-active-modes` because it's enabled for the same reason as all the `last-active-modes` -- just for temporary rule-less browsing.

Does it sound intuitive enough to be a default behavior?

aartaka:
Regarding update to `with-confirm`: is notation like
``` lisp
(with-confirm ("Do you want to go to then-branch?")
  (foo)
  (bar)
  :else (baz) (quux))
   
```

elegant enough?

Ambrevar:
> Regarding update to `with-confirm`: is notation like
> ``` lisp
> (with-confirm ("Do you want to go to then-branch?")
>   (foo)
>   (bar)
>   :else (baz) (quux))
>    
> ```
>
> elegant enough?

Sorry, I had forgotten that our with-confirm has an implicit `progn`.
While the `:else` should mostly be backward compatibly, I don't find it
very elegant.

A more elegant way would be to remove the implicit progn and have
`with-confirm` behave like `if`, e.g.

(with-confirm ("...")
  yes-form
  no-form)

This means that with-confirm calls that have only 1 "yes" branch with
multiple subforms would look like this:

(with-confirm ("...")
  (progn
    form1
    ...
    formn))

If we want to keep the previous behaviour, maybe we can rename the
with-confirm macro to `when-confirm`.

Thoughts?

Ambrevar:
> Does it sound intuitive enough to be a default behavior?

Yes, I think it's a good default! :)

aartaka:
Well, if we split `with-confirm` into two macros, it sounds better to name them after their general synchronous counterparts — `when-confirm` and `if-confirm`. But `with-*` is an important macro marker referring tk `with-result` and it's almost necessary here. All that said, I'd stay with the good old `with-confirm` and just fix its one-branched versions to use explicit `progn`.

I can fix all the uses of one-branched `with-confirm`, if this is the only problem with it. Backwards-compatibility is not much of a problem, because it was not exported untill recently, and almost all the code using `with-confirm` is in the Nyxt core :)

Ambrevar:
@jmercouris: Thoughts on this?

aartaka:
By the way, all the uses of `with-confirm` except the ones in auto-mode.lisp have only one expression in their body. There are no explicit `progn`s to add :)

jmercouris:
It is a branch condition, so I like the use of IF and WHEN. I find that the with-* macros are used to expose access to some resource within the body. Since we are not doing that, but rather branching, my vote is for the IF and WHEN.

jmercouris:
Either direction is good though.

Ambrevar:
@aartaka Then I propose you replace with-confirm with an `if-confirm` that has 2 branches.  If we ever need `when-confirm` we will implement it in due time.

aartaka:
Fine by me. I'll push it in an hour.

aartaka:
What's new:
- Simplified and shortenmed code wherever possible. If something goes wrong, this commit can be a regressive one :) 
- Dwim-ish behavior of `last-active-modes` even on the matching URLs.
- `if-confirm` as a two-branched replacement for `with-confirm`. All the places where `with-confirm` was used are changed accodingly.
- Renaming `last-non-matching-url` -> `last-active-modes-url`, because it's used on matching URLs too. I don't like this name either, because `active-modes` is too vague.
- `save-non-default-modes-for-future-visits` (rename to `save-modes-for-future-visits-dwim`?) command that uses set differences of `default-modes` and `(modes (current-buffer))` to store the enabled/disabled modes where they belong.
- Little documentation, comments and code adjustments.

aartaka:
I've been close to implementing the mode-prompting, but the synchronous nature of Nyxt hooks have stopped me. 

The problem is that `pre-request-hook`, same as `request-resource-hook`, is syncronous. It takes the result of whatever is there, not waiting for the `if-confirm` and `with-result` to finish what they do, resulting in `nil` as their value. This can bring two possible problems:
- Website request can time-out/cancel.
- Website will load with no problems, but the necessary modes won't be activated, because the loading have happened before user have inputed "yes" or "no" on mode application.

Both consequences are undesirable (expecially the latter one), that's why I don't see this feature as easily implementable. To properly implement it, we need to make hooks asynchronous or at least make them wait for the results of asynchronous constructs.

What do you think?

Ambrevar:
> - `if-confirm` as a two-branched replacement for `with-confirm`. All the places where `with-confirm` was used are changed accodingly.

Great!  One thing though: this change belong to a separate commit ;)

> - Renaming `last-non-matching-url` -> `last-active-modes-url`, because
> it's used on matching URLs too. I don't like this name either, because
> `active-modes` is too vague.

The problem here seems to be the word "active", which is not well
defined.

What about "non-auto" or "manual" instead of "active"?

Thus we would have:

- "last-manual-modes"
- "last-manual-modes-url"

> - `save-non-default-modes-for-future-visits` (rename to
> `save-modes-for-future-visits-dwim`?) command that uses set
> differences of `default-modes` and `(modes (current-buffer))` to store
> the enabled/disabled modes where they belong.

I prefer `save-non-default-modes-for-future-visits`.
As you found out as well, I think "infer" is better than the "dwim"
acronym.
Here it could be `save-inferred-modes-for-future-visits`, but
"non-default" is better in my opinion.

Ambrevar:
> - Website request can time-out/cancel.

This would be a show-stopper indeed.
Are you sure it times out though?

Are you saying that if we add a handler to request-resource-hook that
sleeps for, say, 1 minute, then the resource may time out and not get loaded?

> - Website will load with no problems, but the necessary modes won't be
> activated, because the loading have happened before user have inputed
> "yes" or "no" on mode application.

We've been discussing the question of synchronous minibuffer interaction
before.  It's an important issue.

If we are not going to implement this feature here, we can keep this
discussion for later.

> Both consequences are undesirable (expecially the latter one), that's why I don't see this feature as easily implementable. To properly implement it, we need to make hooks asynchronous or at least make them wait for the results of asynchronous constructs.

If the timeout is an issue, we would need to work around it first.

Maybe we can replace the request by a new final one, thus avoiding any
possible delay:

```
             (webkit:webkit-policy-decision-ignore response-policy-decision)
             (webkit:webkit-web-view-load-request (gtk-object buffer) request)
```

Ambrevar:
Merged with cba32e8ef473f2b6a5ecd212a50ba4d4075cb453.

aartaka:
> This would be a show-stopper indeed. Are you sure it times out though? Are you saying that if we add a handler to request-resource-hook that sleeps for, say, 1 minute, then the resource may time out and not get loaded?

I'm not sure about that. It was just one case of ISP-blocked website that the cancellation message have appeared for. I don't have an exact recipe, but I'll try to tinker with asynchronous/timed constructs in hooks and see how it works. If it's more general than blocked websites, then it's a problem.

Issue #24, opened 2017-11-28
C-g not generating link hints


jonathanwilner:
On Mac OS X High Sierra, C-g does not generate link hints on any page - just an audio error chime.

jmercouris:
Hi, thank you for the report!

which keys are you literally pressing? The chime indicates an unrecognized action

jonathanwilner:
Hi- I am hitting Control & G. Either the Apple Control or CapsLock mapped to Control.

jmercouris:
Hi thank you for the update! Did you compile it yourself or download the pre-built binary? What happens when you launch the application from the terminal like this "/Applications/nEXT.app/Contents/MacOS/nEXT"? Does it give any more output?

tinder-jonathanwilner:
I am using chunkwm and skhd.  - good call! However, when I disable them, there's no difference.

Using the prebuilt binary - I failed trying to build from source with something going wrong with QT. 

There's no console output when run from the terminal.

jmercouris:
Hey, thank you for the follow up!

That's really strange that you had some issues with QT, as there is no QT in the project used at all :D

Could you please post the build error? I would like to determine whether it is an issue with building and running on different versions of OSX, or something different altogether

Thank you for your time!

tinder-jonathanwilner:
Hi there - sorry for the bad report re: QT. I meant QL. It was the same as #25 -
https://github.com/nEXT-Browser/nEXT/issues/25 . I followed the workaround and got it compiled. 

Unfortunately, same issue & no more output from running from nEXT.app/Contents/MacOS. 👎

jmercouris:
Hi, I've been thinking about what may be causing this behavior as it works on most people's computers. It must be something tied to configuration somehow. What does your keyboard look like? Are you using the American keyboard layout? Do you have any issues with any other keybindings?  What does the modifier keys section look like on your system?

<img width="780" alt="screen shot 2017-12-06 at 16 35 01" src="https://user-images.githubusercontent.com/1691662/33669695-730da626-daa3-11e7-88f5-d6ddb74f938b.png">

seanfarley:
I'm going to guess all of us have something along the lines of caps lock being mapped to control (I know I do).

jonathanwilner:
This works for me in the latest HEAD.

Layout is US English, caps lock mapped to control.

jmercouris:
I'm glad to hear that it is working now! I'll close this issue then!

PR #885, opened 2020-08-06
terminology: change to allow/deny-list terminology

tychoish:
Following a conversation on IRC yesterday, I went ahead and made a relatively trivial change to use allow/deny terminally in appropriate contexts. 

I think the cost/risk of this change is pretty minimal, the impact to the interface is small, and I think the potential for impact may be worthwhile. I'm particularly thinking about: 
- potential contributors who might struggle the legacy language for any number of reasons, who's contribution we'd miss as a result. 
- potential confusion, particularly across language barriers. The new language is clear and explicit, and consistent with the way that other projects are thinking about these constructs (e.g. the linux kernel made a similar change to its language earlier this year.)

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -82,7 +82,7 @@
                (:file "application-mode")
                (:file "web-mode")
                (:file "reading-line-mode")
-               (:file "certificate-allowlist-mode")
+               (:file "certificate-exception-mode")
                (:file "emacs-mode")
                (:file "vi-mode")
                (:file "blocker-mode")
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -82,7 +82,7 @@
                (:file "application-mode")
                (:file "web-mode")
                (:file "reading-line-mode")
-               (:file "certificate-whitelist-mode")
+               (:file "certificate-allowlist-mode")
                (:file "emacs-mode")
                (:file "vi-mode")
                (:file "blocker-mode")
--- a/source/blocker-mode.lisp
+++ b/source/blocker-mode.lisp
@@ -154,7 +154,7 @@ Example:
              :combination #'combine-composed-hook-until-nil
              :handlers (list #'request-resource-block)))))))
 
-(defmethod denylisted-host-p ((mode blocker-mode) host)
+(defmethod blocklisted-host-p ((mode blocker-mode) host)
   "Return non-nil of HOST if found in the hostlists of MODE.
 Return nil if MODE's hostlist cannot be parsed."
   (when host
@@ -162,11 +162,11 @@ Return nil if MODE's hostlist cannot be parsed."
                never (member-string host (parse hostlist))))))
 
 (defun request-resource-block (request-data)
-  "Block resource queries from denylisted hosts.
+  "Block resource queries from blocklisted hosts.
 This is an acceptable handler for `request-resource-hook'."
   (let ((mode (find-submode (buffer request-data) 'blocker-mode)))
     (if (and mode
-             (denylisted-host-p
+             (blocklisted-host-p
               mode
               (quri:uri-host (url request-data))))
         (progn
--- a/source/blocker-mode.lisp
+++ b/source/blocker-mode.lisp
@@ -1,6 +1,6 @@
 (uiop:define-package :nyxt/blocker-mode
   (:use :common-lisp :trivia :nyxt)
-  (:documentation "Block resource queries blacklisted hosts."))
+  (:documentation "Block resource queries for listed hosts."))
 (in-package :nyxt/blocker-mode)
 (eval-when (:compile-toplevel :load-toplevel :execute)
   (trivial-package-local-nicknames:add-package-local-nickname :alex :alexandria)
@@ -119,7 +119,7 @@ Return nil if hostlist cannot be parsed."
   "Default hostlist for `blocker-mode'.")
 
 (define-mode blocker-mode ()
-    "Enable blocking of blacklisted hosts.
+    "Enable blocking of listed hosts.
 To customize the list of blocked hosts, set the `hostlists' slot.
 See the `hostlist' class documentation.
 
@@ -154,7 +154,7 @@ Example:
              :combination #'combine-composed-hook-until-nil
              :handlers (list #'request-resource-block)))))))
 
-(defmethod blacklisted-host-p ((mode blocker-mode) host)
+(defmethod denylisted-host-p ((mode blocker-mode) host)
   "Return non-nil of HOST if found in the hostlists of MODE.
 Return nil if MODE's hostlist cannot be parsed."
   (when host
@@ -162,11 +162,11 @@ Return nil if MODE's hostlist cannot be parsed."
                never (member-string host (parse hostlist))))))
 
 (defun request-resource-block (request-data)
-  "Block resource queries from blacklisted hosts.
+  "Block resource queries from denylisted hosts.
 This is an acceptable handler for `request-resource-hook'."
   (let ((mode (find-submode (buffer request-data) 'blocker-mode)))
     (if (and mode
-             (blacklisted-host-p
+             (denylisted-host-p
               mode
               (quri:uri-host (url request-data))))
         (progn
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -53,7 +53,7 @@ Parent directories are created if necessary."
                    :documentation "The address of the proxy server.
 It's made of three components: protocol, host and port.
 Example: \"http://192.168.1.254:8080\".")
-   (whitelist :accessor whitelist :initarg :whitelist
+   (allowlist :accessor allowlist :initarg :allowlist
               :initform '("localhost" "localhost:8080")
               :type list-of-strings
               :documentation "A list of URIs not to forward to the proxy.")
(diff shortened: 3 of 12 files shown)


Ambrevar:
Maybe I'm missing some context because I don't really understand your points.
Can you name some of the "any number of reasons"?

"whitelist" and "blacklist" are both in the dictionary and have an unambiguous use in the English language.
"allowlist" and "denylist" are not in the dictionary however.

https://en.wiktionary.org/wiki/Special:Search?search=allowlist&ns0=1
https://en.wiktionary.org/wiki/whitelist

tychoish:
The legacy terms work by associating "black" with bad and "white" with good, which I think draws a pretty clear line from racism in western society. While allowlist isn't in the dictionary "allow" certainly is, and I think they conclusion is pretty straightforward from there, certainly more straightforward than passing through the indirection of color associations.

Not only does it seem like a fair change, particularly the minimal costs, but I suspect that the convention is also changing and evolving (e.g. the linux kernel is hardly the vanguard of change.), as language does, and:
- it seems like making a 50 line patch now is easier than making a larger one one in the future, and
- the primary (to my mind) risk is (silently) loosing potential contributors.

jmercouris:
Here is a quote from Wikipedia:

Though the terms do not have a racial origin, some have started avoiding the usage of blacklist/whitelist due to perceived racial connotations, and use alternate words like allowlist/blocklist.

jmercouris:
So here is the question, since there is no technical justification for doing this- is there a social one? Given that these terms have a non racial origin seems to indicate: no. That said, could someone _reasonable_ be offended? I find it hard to believe that they could.

jmercouris:
The cost is minimal, though non-zero.

Certificate-allowlist-mode doesn't readily make sense to me like certificate-whitelist-mode.

Ambrevar:
Indeed, certificate-allowlist-mode sounds like broken English and does
not read well.

Can you think of a better term?
inclusion-list is a bit lengthy.
What about `in-list` and `out-list`?  Or `inclusions` and `exclusions`?

certificate-whitelist-mode could be renamed to certificate-exception-mode.

tychoish:
I think blocklist is an improvement over deny-list. I'm happy and/or inclined to make that change.

I think think debating the merits or reason-ability of someone's offense doesn't feel like a great place to start a conversation with someone who we'd imagine we don't already know.

jmercouris:
I think it is, otherwise what are we discussing exactly? We must first discuss IF there exists a reasonable probability of someone taking offense. Otherwise, everything should be axiomatically accepted as potentially being offensive as soon as anyone brings it up. We must operate with reason.

jmercouris:
The only thing I can think of is certificate-passlist-mode...

tychoish:
Right, I guess I'm saying that there are lots of examples of folks saying "maybe it's best if we change this?" in the world right now, lots of bigger projects that have risked much larger diffs to make changes along these lines, so I think it's fair to say "this is a choice that impacts/upsets people." Not making a change, then puts us down as saying "we don't think it's reasonable for you to be impacted or upset," to people who might otherwise be inclined to contribute/participate in this project. 

I'm not particularly worried (nor would I be if I were you) about making this change as opening a floodgate for any number of arbitrary terminology changes in the future. It's clear that the project evaluates each change very seriously and I don't see any reason why that should change. 

I think pass-list isn't great, because it's not clear if "pass" means "to do nothing, and let it pass through" or "pass and deny the request." Allow is unambiguous in this regard.

Ambrevar:

> The only thing I can think of is certificate-passlist-mode...

Don't you like certificate-exception-mode?

Blocklist is good in my opinion.

Now we only need to come up with the antonym.

phil-hands:
As a long-time Debian Developer, I've seen quite a few of these discussions over the decades, and certainly would have been on the "What's the technical justification?" side of this 20 years ago, but I've learnt to take a wider view since then.

It seems pretty clear where the drift of history is on this subject is leading.  Even high-inertia projects which are packed with stick-in-the-mud folks like myself have started to make moves in the direction of inclusive language.

That being the case, if we jump forward 10 years, I think it is reasonable to assume that by then, (if Nyxt is still a live project) it will either have made this sort of language change, or it will still be regularly having newbies enthusiastically suggesting that they could make the world a better place by making such a change for you.

That being the case, you are presented with the option to either change this now, and never have to think about it again, or to have regular pointless discussions about it.

If you chose the latter option, you then run the risk every time it happens that one of the people witnessing the discussion concludes that if you're willing devote effort to repeatedly reject such a trivial change then you're also likely to reject things they care more about, so perhaps they should devote their enthusiasm to a more welcoming project.

Trying to separate the technical from the social in this situation strikes me as unhelpful, because you're talking about the way you are deciding to present your project to the world (to a small extent), and those decisions are used by people to judge the viability of your project, which then can actually feed back and have real effects on your project.

As a case in point, you recently decided to on a name-change for the project, which strikes me in many ways as a similar decision.  I presume you started out thinking that the name made no technical difference so couldn't see why you needed to change it.  When I first came across "next-browser" my first thought was that it was a foolish name that rendered the project mostly unsearchable-for on the Internet, and thus doomed it to failure unless it was changed.  That feeling was confirmed when I saw that your Logo was sufficiently close to the NeXT Computer Logo to land you in legal trouble if the browser ever managed to achieve any sort of popularity.  I've no idea if the name change was prompted by a realisation of those problems on your part, or by the arrival of a Cease&Desist letter, but when I saw it I revised my assessment of the future viability of the project upwards, pulled the latest version, and resumed trying to track down why video playback on Debian didn't work (which I've now fixed).  I would not have bothered with that if the old name had remained.  You'd have never seen me on your IRC channel.  You'd not know why I'd decided not to give 'next' a second look, and you'd still be cheerfully stating that you don't believe that anyone could possibly be put of using a great program by its choice of name ... but you'd be wrong.

Even though (as a privileged white male) the language has no direct emotional impact on me, the refusal to change it would cause me to again re-evaluate the projects viability.  I would see it as a very minor nudge in the direction of discouraging contributions.  The problem is that maintaining the viability of a Free Software project can sometimes be on a knife edge.  If someone turns away from your project, they may be the one that was going to write the killer feature, or find the subtle bug.  Understanding that, it makes onlookers like me slightly doubt the viability of the project, which may cause some of us to go elsewhere too.  The vast majority of people that make those decisions will not tell you why they went elsewhere, and may not even notice themselves that they made such a decision.

Anyway, I seem to have written about ten times as much as this deserves, so I'll stop now -- have a nice day :-)

Cheers, Phil.

P.S. if a purely technical justification helps you accept this change, note that:

  https://github.com/atlas-engineer/nyxt/pull/885/files#diff-5ba4e57a3b8ac46b5365e7d862ab7570L30

it fixes a typo in a comment :-)

PlasmaShift:
Some Thoughts on this issue

-   Resons to keep whitelist/denylist
    -   whitelist and blacklist have well establehed meanings that everybody know.
    -   They are in the dictionary
-   Reasons to change to allow/deny
    -   Don't want to deal with arguing
        -   Policicly correct activist will keep on buging you
-   Non Reasons
    -   Racism
        -   Only racisit at a very surface level
-   Other people talking about this
    -   <https://www.youtube.com/watch?v=tx2v0Y9Ci7I> Politically Correct Tech by Mental Outlaw
    -   <https://www.youtube.com/watch?v=n_HzEmGOVJ4> Linux Kernel blacklists "blacklist" by Bryan Lunduke
-   Political Reasons to keep the changes
    -   Goes agenst freedom of speech or dulutes speach
        -   <https://www.youtube.com/watch?v=K-58HoTHWQk> Control the Words, Control the Culture by prageru
    -   Racism &#x2013; Put your energy in the right place &#x2013; placing it here wastes energy and dosn't solve anything at best, prevents people through misderction from helping by inderection at worst
        -   The energy to help should be spent helping poor families or single <https://www.youtube.com/watch?v=c8PEv5SV4sU> Social Class & Poverty in the US: Crash Course Sociology #24 by CrashCourse

tychoish:
Certificate Exception Mode is actually really good! 

How do we feel about keeping allow-list (or allowlist, preferences?) in, say, the proxy-mode and proxy class in browser.lisp?

Ambrevar:
Sam Kleinman <notifications@github.com> writes:

> How do we feel about keeping allow-list (or allowlist, preferences?) in, say, the proxy-mode and proxy class in browser.lisp? 

I prefer allow-list to allowlist, but I'm still not very happy with it.

Thinking a bit more about it, I am starting to like the terms
`inclusions` and `exclusions`.  It's short, meaningful and it does not
imply that the value is a list, which may not be true (could very well
be a set or a hash table).

jmercouris:
Anything can be serialized to a list of elements, the underlying data structure is irrelevant. If we wish to be short and universal we may use green and red. What do you think Pierre?

Ambrevar:
You mean greenlist and redlist?
Seems too unfamiliar in my opinion.

My point about the "*list" naming is that it implies a list _type_.
Calling a hash table "foo-list" is confusing.

tychoish:
I think colors are harder than other words. I worry a little bit about the -list convention being confusing, but that wouldn't be a change, and I avoided doing hyphen separation to draw some kind of distinction.

PR #888, opened 2020-08-07
Use defclass-export instead of defining user classes in define-config…

Ambrevar:
…uration.

This greatly simplifies the define-configuration macro.
The creation of a user-CLASS is poor practice since it may clash with existing
classes.
Relying on define-configuration is foolproof since it uses gensym to generate
unique names.

While we are at it, we remove the unused *CLASS-class* global.

### DIFF
--- a/source/utility.lisp
+++ b/source/utility.lisp
@@ -153,9 +153,8 @@ Return most recent entry in RING."
 
 (export-always '%slot-default)
 (export-always 'define-configuration)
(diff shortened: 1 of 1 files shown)


jmercouris:
I can't test this currently, but reviewing the changes in the source, the logical implications are sound. I agree with this change.

Ambrevar:
Merged in 7db1b0835b3c2f4ffa33ba704e0345bd59562aa6.

PR #890, opened 2020-08-07
Replace classes using replace-class macro and fix CCL support.

Ambrevar:
Should be good for merging.
@jmercouris Can you test on macOS to make sure I didn't break anything?

### DIFF
--- a/source/macro.lisp
+++ b/source/macro.lisp
@@ -1,5 +1,25 @@
 (in-package :nyxt)
 
+(defmacro replace-class (from to)
+  "Set class corresponding to FROM symbol to that of TO.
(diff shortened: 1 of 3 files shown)


jmercouris:
#I will test when I can, don't have easy computer access at the current moment. I put it on my calendar for Monday, but will try to get to it sooner

Ambrevar:
On second though it should not break anything, the macro should expand
to the same code than that on master, so this should be rather safe to merge.

jmercouris:
OK!

Ambrevar:
Merged with 4a2ddfc48255da2c7019c716bf237cf126ec7461.

PR #896, opened 2020-08-12
Rearrange files

Ambrevar:
This massively cleans up the code base.

Stated goals:

- Move all the browser-unrelated code away from browser.lisp.
- Get rid of utility.lisp and macro.lisp.  These unfortunate file names have the drawback that they tend to endlessly accumulate code.
- Disentangle load order.

I'm quite happy with the result.  I'm only a little dissatisfied with the "command-commands.lisp" file name.


Ambrevar:
I believe this should close #581.

Ambrevar:
Note: I haven't edited any form, I've only moved code around, so the browser should work just the same as before.

source/javascript.lisp
```diff
@@ -0,0 +1,45 @@
+(in-package :nyxt)
```
jmercouris, line 1: I think this file should be called parenscript.lisp

Ambrevar: I was hesitant.  I was thinking that the purpose of this file is to deal with javascript (the end result), while Parenscript is an implementation detail.  What do you think?

jmercouris:
Bonjour Pierre,

thank you for this nice clean up! It moves things to much more logical locations!

I have only the following differences of opinion:

+ The javascript.lisp file can be called parenscript.lisp. It is parenscript after all.

+ I believe that minibuffer-helper.lisp can just be in the minibuffer.lisp, it is not such a big file after all. Scattering the functions around more files will make it harder to find things.

+ I have the same idea about command-commands.lisp. I understand what you are trying to do in the separation, I just don't think it is wholly necessary. In order to avoid this confusion, just keep it in a single command.lisp file.

source/javascript.lisp
jmercouris: I actually think of it the other way around :-D. Javascript is an implementation detail, as that is what parenscript compiles to. I imagine a user would be quite surprised to open up javascript.lisp and find no javascript, and only parenscript and lisp :-D

Ambrevar:
> + I believe that minibuffer-helper.lisp can just be in the minibuffer.lisp, it is not such a big file after all. Scattering the functions around more files will make it harder to find things.
>
> + I have the same idea about command-commands.lisp. I understand what you are trying to do in the separation, I just don't think it is wholly necessary. In order to avoid this confusion, just keep it in a single command.lisp file.

None of the above is possible because of the load order.  If you look at
nyxt.asd, you'll see that the files are loaded at a particular moment.

For instance, much of the minibuffer helpers are used by buffer
functions.
And it's not possible to load minibuffer.lisp buffer buffer.lisp because
MINIBUFFER inherits from BUFFER.

source/javascript.lisp
Ambrevar: It has a `lisp` extension, so I'd expect to find Lisp code, not Javascript.

Some more arguments:
- If we ever use another Javascript generator, or plain javascript, it would make less sense to store it in a parenscript.lisp file.
- If we had code related to HTML, I would store it in `html.lisp`, not in `markup.lisp`.

All that said, maybe both parenscript.lisp and javascript.lisp are poor names.  What about `renderer-script`, `scripting`, webview-scripts`?

jmercouris: renderer-script is the best in this case. It is agnostic and obvious what it does

jmercouris:
Well then, in that case let us merge :-)

in the future we can move things around if we don't like them

Ambrevar:
> in the future we can move things around if we don't like them 

I'm not sure what you mean by that, can you detail? :)

source/javascript.lisp
Ambrevar: OK!

jmercouris:
I'm saying no point in sweating things like command-commands because the solutions to these problems will come to us in due time

Ambrevar:
Merged with 098d7359e0648ce5745031c93d5ccc0caec63010.

PR #897, opened 2020-08-12
Internal buffer support/example usage

jmercouris:
Please try the MESSAGES command to see an example of usage.

Summary:

+ we have internal buffers
+ if there is an internal buffer resource request, evaluate it assuming it is a Lisp form
+ the Lisp forms of internal buffers take the above form as embedded HREFs with the target lisp://(some-lisp-code-to-evaluate)

Should you desire to navigate to a URL on the WWW from an internal buffer, you will have to use code that creates a new buffer or otherwise sets the URL of the internal buffer manually. This is intentional as a security measure.

I think this will do a fair bit in lowering the barrier to entry for Nyxt users :-)

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -556,7 +556,7 @@ Deal with REQUEST-DATA with the following rules:
       (declare (type quri:uri url))
       (cond
         ((internal-buffer-p buffer)
-         (evaluate (quri:url-decode (quri:uri-domain url)))
+         (evaluate (quri:url-decode (schemeless-url url)))
          nil)
         (bound-function
          (log:debug "Resource request key sequence ~a" (keyspecs-with-optional-keycode keys))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -555,6 +555,9 @@ Deal with REQUEST-DATA with the following rules:
                                 (keymap:lookup-key keys keymap))))
       (declare (type quri:uri url))
       (cond
+        ((internal-buffer-p buffer)
+         (evaluate (quri:url-decode (quri:uri-domain url)))
+         nil)
         (bound-function
          (log:debug "Resource request key sequence ~a" (keyspecs-with-optional-keycode keys))
          (funcall-safely bound-function :url url)
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -365,13 +365,14 @@ The version number is stored in the clipboard."
                          (buffer-list))))
     (unless buffer
       (setf buffer (nyxt/help-mode:help-mode :activate t
-                                             :buffer (make-buffer :title "*Messages*"))))
+                                             :buffer (make-internal-buffer :title "*Messages*"))))
     (let* ((content
-             (apply #'markup:markup*
-                    '(:h1 "Messages")
-                    (mapcar (lambda (message)
-                              (list :p message))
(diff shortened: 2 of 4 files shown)


source/browser.lisp
```diff
@@ -676,6 +676,9 @@ Deal with REQUEST-DATA with the following rules:
                                 (keymap:lookup-key keys keymap))))
       (declare (type quri:uri url))
       (cond
+        ((internal-buffer-p buffer)
+         (evaluate (quri:url-decode (quri:uri-domain url)))
```
Ambrevar, line 680: I think you should use `schemeless-url` here, otherwise code containing a slash might get trimmed.

source/urls.lisp
```diff
@@ -119,3 +119,8 @@ there is none."
   (or (find "default"
             search-engines :test #'string= :key #'shortcut)
       (first search-engines)))
+
+(defun lisp-url (data)
```
Ambrevar, line 123: `data` is not very meaningful here.  What about "lisp-code"?
Also declaim the ftype of the function to make the argument type explicit.

About that: I suggest we accept an _s-exp form_ instead of a string: this way we would have syntax highlighting in the argument of `lisp-url`, instead of an obscure string which might have unbalanced parentheses ;)

source/help.lisp
```diff
@@ -365,13 +365,14 @@ The version number is stored in the clipboard."
...
+                                             :buffer (make-internal-buffer :title "*Messages*"))))
     (let* ((content
-             (apply #'markup:markup*
-                    '(:h1 "Messages")
-                    (mapcar (lambda (message)
-                              (list :p message))
-                            (reverse (messages-content *browser*)))))
+             (markup:markup
+              (:h1 "Messages")
+              (:a :href (lisp-url "(nyxt::clear-messages)(nyxt::messages)") "Clear Messages")
```
Ambrevar, line 372: Are the double colons really needed?  Both commands are exported.

Ambrevar: Nit: I would spell it "Clear messages" (no title case)

Ambrevar:
Just tried it: this is awesome!!! :D

source/urls.lisp
jmercouris: I've changed the name to `lisp-data`. Code is data after all :-)

As per accepting SEXPs, I agree that would be nice. I have a bit of a challenge however. I believe I will have to either make it a macro or accept a quoted list. In which case passing something like

```
(fish are cool)(salmon are the coolest)
```

would become a little bit cumbersome. Do you have any ideas?

source/help.lisp
jmercouris: Complete.

source/browser.lisp
jmercouris: Complete.

source/urls.lisp
jmercouris: You know what, I totally forgot about `&rest` for a second, that would of course work... :-D

Ambrevar: Indeed, I believe we don't have to make it a macro (which could be confusing to read) and just pass &rest quoted lists. Example:

```lisp
(lisp-url '(form1 arg1) '(form2 arg2))
```

PR #900, opened 2020-08-20
Composite search

jmercouris:
This is a demonstration of the composite search behavior we have spoken about. A single minibuffer shows the result from several mini buffers. Each result can have different behavior upon selection. This allows us to compose the results from buffer switch, bookmark select etc into one cohesive minibuffer, and still perform the appropriate action with the selection.

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -79,6 +79,7 @@
                (:file "auto-mode")
                (:file "external-editor")
                (:file "url-group")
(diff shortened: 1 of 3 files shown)


Ambrevar:
Just tested and it works for me! :)

That said, I think this belongs to the set of changes I intended to do
when moving the minibuffer to a separate library.

Maybe we put this on hold until the library gets into shape.  What do
you think?

PR #903, opened 2020-08-24
Lisp url with forms

Ambrevar:


source/help.lisp
```diff
@@ -358,6 +358,11 @@ The version number is stored in the clipboard."
   (trivial-clipboard:text +version+)
   (echo "Version ~a" +version+))
 
+(define-command clear-messages ()       ; TODO: Move to lod-mode?
```
jmercouris, line 361: what is lod-mode?

Ambrevar: I was wondering the same thing!  The TODO is not from this change.  git-blame indicates that I wrote this, but I can't remember.  It's probably a typo.

That said, clear-messages should only be available from the message buffer.  So we need a message-mode, and move clear-messages there.  Thoughts?

jmercouris: I have wondered for some time since I first spotted 'lod-mode' :-D

Do we need a message-mode? In my opinion, not really. I presume it is because you don't want clear-messages to appear in the minibuffer when not viewing the messages buffer?

Ambrevar: Correct.  There is another benefit for a message-mode, as well as a dedicated mode for all internal buffers: it allows us to "type" these buffers.
Right now, to look up the *Messages* buffer or the *Buffers* buffer we check the name.  But it's brittle, because the name can be changed (for now from the REPL, maybe interacticaly in the future -- like Emacs).

In Emacs, it's idiomatic to identify a buffer by its mode.  Seems like a good strategy and I propose we do the same.

jmercouris: OK, your second line of reasoning has convinced me. It is indeed brittle to rely on the name! It is much easier to mutate the name than it is to mutate the object type.

Ambrevar:
I'll open an issue about the per-special-buffer mode.

OK to merge otherwise?

jmercouris:
OK to merge! :-)

Ambrevar:
Merged with a6c89eeadb7465d527eb4a9e050ddb1ebcd81ffe.

Issue #909, opened 2020-08-25
Per-internal-buffer modes


Ambrevar:
We've got many internal buffers:

- Downloads
- Help buffers
- List buffers
- Messages
- REPL

Please share if you can think of special buffers I forgot.

Currently we have no way to identify these buffers.  Making a new buffer class for each one of them is a bad idea, because we want to be able to navigate across any of the internal buffers (e.g. a help page could point to the downloads buffer).

Instead, we can do like Emacs does it: have a "major-mode" for each type of  internal buffers.

Another issue is that we currently cannot have multiple buffers for some of these types.  For instance there can be only one *Downloads* buffer because we match against its name.  If we start matching against the mode however, then we can have multiple downloads buffer.

Finally, some commands like `clear-message` should only be visible from the *Messages* buffer.  Having a dedicated mode would fix this.

Questions:

- Should these modes be mutually exclusive?  If a buffer has both downloads-mode and messages-mode, then which one is it?  If we want mutual exclusivity, then we need a system like Emacs' major-modes.

- How do we instantiate these buffers?  Shall we just pass the mode and the name to buffer-make?  Do we want an additional abstraction? 

Also see discussion in #903.

Ambrevar:
a6c22473eb4a186416f13fc07a1e37901afccf76
adds message-mode and 7c14c34216394aa8bad8b8fc00c028d86e580ba0
adds buffer-listing-mode.

jmercouris:
We have solved this, every internal buffer has a special mode when necessary :-)

Ambrevar:
Have we answered all the questions from the original post though?

I believe we still don't have a good abstraction for these special modes / buffers.  Should we reopen this issue to keep discussing this?

jmercouris:
I would suggest a new issue.

Ambrevar:
Done in #1790.

Issue #837, opened 2020-07-09
Cookies isolation (e.g. per-domain)


Ambrevar:
Cookies have been the bane of privacy for long, so Nyxt gotta do something about it! :)

A nice feature would be to isolate cookie per-domain.
Currently, Nyxt can set the cookies path.  Changing it for a live buffer might not be very useful because I don't think it changes in-memory cookies (this would need verification).

So we need to see if WebKitGTK has some options to isolate cookies within a live buffer.
If not, we might have to create a new buffer when navigating to a domain for which we want isolated cookies.
This is similar to how we need to do "private / incognito" buffers (#247).

aartaka:
Can [Intelligent Tracking Prevention](https://webkit.org/blog/7675/intelligent-tracking-prevention/) be the thing we look for? It isolates the cookies that are not used daily, purges all the third-party tracking cookies, while retaining the cookies that user needs. The downside is that it's a black box Machine Learning-based solution that tracks all the activity of the user (it's for the good end, but still...)

Alternatively, WebKit has the ability to manipulate cookies manually (`WebKitCookieManager` and `WebKitDataManager` have functions for that, e.g., [webkit_website_data_manager_remove()](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebsiteDataManager.html#webkit-website-data-manager-remove), [webkit_website_data_manager_clear()](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebsiteDataManager.html#webkit-website-data-manager-clear), [webkit_cookie_manager_delete_cookie ()](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitCookieManager.html#webkit-cookie-manager-delete-cookie)), so we can build the cookie-watching mechanism on top of that. I'm not sure about isolation, but per-domain cookie cleaning is certainly possible.

Intelligent Tracking Prevention seems to be the best way, given that it's relatively mature and well-designed.

Ambrevar:
It's a great idea and it was released in WebKitGTK very recently:
https://blogs.gnome.org/mcatanzaro/2020/09/16/epiphany-3-38-and-webkitgtk-2-30/

We could very well offer both options: Intelligent Tracking Prevention
_and_ home-rolled solution to manipulated the cookies in Common Lisp.

@aartaka: Interested in adding bindings for Intelligent Tracking Prevention?

aartaka:
Yes, I'm already looking into that!

[...]

Issue #247, opened 2019-06-22
incognito mode


easbarbosa:
Hi,

May be it still too early to an incognito mode to arrive, but at last I will make it known! 

Thanks!

jmercouris:
It is not too early, you could implement it quite easily at the moment. Maybe a fun exercise to do ;)?

Ambrevar:
I've pushed a fix on master which allows a workaround.  Export XDG_DATA_HOME to something temporary to create a folder for cookies, history, bookmarks, sessions, etc.

Example:

```sh
$ XDG_DATA_HOME=/tmp next
```
If your temp folder gets reset when the machine starts, then you can leave it there.  Otherwise erase all tracks with
```
rm -rf /tmp/next
```

Ambrevar:
Of course a proper incognito mode would be nicer.

A see a couple of approaches:

- Incognito process: With an `--incognito` flag, disable history, bookmarks, sessions, etc.
Maybe it's as easy as setting XDG_DATA_HOME to /dev/null.  This is the least flexible approach.

- Incognito buffer mode: The history is not saved, the cookies are not persisted, the session won't remember this buffer.  The bookmarks are still accessible though.  This is more flexible as it does not force the user to user 2 difference processes.

With the second approach, we could still start a process with an `--incognito` flag that would force-enable the buffer mode.  But maybe that's too fragile and we should implement both approaches anyways.

Ambrevar:
And... another enhancement on master: we now have data profiles.

Starting with `next --data-profile private` will disable all file persistence.

Next step: allow reading the history, bookmarks, etc. without persisting them.  Can this be done with a data profile?

jmercouris:
The first thing that comes to mind is redefining the store sexp functions to be no ops, that may suffice

Ambrevar:
Yes but how do we make a "profile" to set this?

jmercouris:
rebind symbols to point to noop functions?

Ambrevar:
We can just bind them to nil, no problem.

My question was about the user interface.

Do we implement it this kind of "profiles" as modes?
How do we drive them from the command line?
How do we customize them?
How do they relate to data profiles?

jmercouris:
Ah, I see, yes that is a much bigger question than I thought.

Ambrevar:
About the name:

- Chrome uses "incognito".
- Firefox uses "private".
- WebKit uses "ephemeral".

[...]

PR #910, opened 2020-08-26
Add per-buffer data storage

aartaka:
This moves all the storage-related data to buffers and makes for a flexible per-buffer storage settings.

### Motivation
It's necessary to separate some data (e.g., cookies, local-storage) for different buffers to preserve users privacy, when necessary. Nyxt had no such mechanism, except for rebinding browser-based `(history|download|bookmarks)-path` slots, which is awkward. Another alternative, the `data-profile`s, was good for disabling all the storage, but was inflexible and uncomfortable to work with (however, it just can be my subjective repulsion from signatures like `(defmethod expand-data-path ((profile (eql +default-data-profile+)) (path session-data-path))...)` — it just doesn't look right).

This patch aims to make it easier to customize storage settings and isolate them for certain buffers. This opens the way for per-domain data storage settings and private mode (outlined below in "What's Not There (Yet)") and, possibly, some other conveniences I can't even imagine yet!

### What's New
- `data-profile` is now the place for some data-storing slots:
  - `downloads`,
  - `download-watcher`,
  - `history-data`,
  - `bookmarks-data`,
  - `auto-mode-rules`
- `buffer` is the home for previously `browser`-based slots:
  - `download-path`,
  - `history-path`,
  - `bookmarks-path`,
  - `session-path`,
  - `auto-mode-rules-path`,
  - `cookies-path`,
  - `standard-output-path`,
  - `error-output-path`.
- `buffer`s now have a `data-profile` slot, so storage can be adjusted on a per-buffer basis.
- `restore` and `store` generic functions to (re)store the data from the given paths, depending on the paths themselves.
- `expand-path` and, consequently, `with-data-file` now require additional buffer argument. Several functions/commands changed their signatures to match this change.
- `default-data-profile` and `private-data-profile` classes with (respectively) class-allocated data and instance-allocated data. In other words: now all the data of the buffers with default profiles will be shared, while buffers with private profile will be isolated even from each other.
- `*global-data-profile` comes as an alternative to the `+default-data-profile+`, because the latter had two use cases:
  - providing default data-management behavior;
  - dispatching the path on browser-less paths (e.g., `*init-file-path*`);
  While the first goal is now covered by `default-data-profile` class, the second needs some handling, and global variable seems to be necessary here, as something neither `browser`- nor `buffer`-based.
- data-storage.lisp is now less dependent on other files and it's easier to extract to a separate library. Some work still needs to be done though.
- Manual and documentation are adjusted to the changes.
- Some typo and syntax fixes.

### What's Not There (Yet)
- The per-buffer `data-profile`-conscious `*Downloads*` buffer is not there yet, because it's better to implement it together with #909.
- Ephemeral/private/nodata buffers (#247) -- these can be built on top of `private-data-profile`, path redefinition, and `is-ephemeral` `WebKitWebView` property, but this is somewhat out of scope of this patch.
- Per-domain cookies (and whatever else) isolation (#837) -- this can be implemented with introduction of several `isolate-(downloads|history|bookmarks|cookies)-mode` modes that will alter the buffer-local `data-profile` (introduce `isolate-cookies-data-profile`?) and with `auto-mode` to toggle them. Doesn't seem hard, but it's outside the scope of this pull request too.

I've assigned to implement the latter two, but I won't be able to do this properly, so this pull request is my best shot at making #247 and #837 implementable.

### Caveats
- The look of data-concerned (`buffer`, `data-profile`) class definition asks for new `defclass` (#889) — the number and similarity of slots is horrible xD
- It's currently hard to know what profile is used by the browser at browser startup. Inspecting the `buffer` class with the help of MOP seems to be the only solid option, but it's not perfect either.
- Race conditions can happen on file access and we don't have any safety mechanism against it yet.

### Possible Future Directions
- `:read-only`/`:readonly` storage keyword. It can be useful for obvious reasons of reading what's stored on disk without persisting the changes. However, it would require non-obvious changes and, possibly, even more strict storage policy.
- Ability to store data in browser instance to duplicate less data (although it's solved by `*global-data-profile*` already)?
- Lock file access while `restore`-ing and `store`-ing data separately?
- Maybe making alias methods for data-storing slots specialized on buffers (`bookmarks-data` and friends), so that `(bookmarks-data buffer) = (bookmarks-data (data-profile buffer))`? I think it's faulty to use such aliases, but it looks prettier and somewhat more intuitive (alternative: actually move data-storing slots to `buffer`, instead of creating aliases -- not really useful though).
- Make the storage safe from race conditions. I've sketched it, but it needs more thought and redesign. Here's the macro I came up with and that should be used to wrap the bodies of the data-altering functions (and it's a pattern you can see in the current code of the functions referred to in the docstring). Importantly, it's unclear where should we store locks and what operations should be blocking.
``` lisp
(export-always 'with-parallel-data-access)
(defmacro with-parallel-data-access (path buffer &body body)
  "Macro to conveniently access persisted data in the data-altering functions.
Exemplary applications are `history-add', `add-modes-to-auto-mode-rules'.
This macro spawns a thread that locks the data, restores the fresh data,
and stores the (possibly) altered data back to the BUFFER's PATH-NAME."
  (alex:once-only (buffer path)
    (alex:with-gensyms (profile path-type lock)
      `(let* ((,profile (if ,buffer (data-profile ,buffer) *global-data-profile*))
              (,path-type (serapeum:class-name-of ,path))
              ;; Locks are stored in a slot of `data-profile' in the 
              ;; initial design iteration, but it's not perfect
              (,lock (gethash ,path-type (locks ,profile)
                              (setf (gethash ,path-type (locks ,profile))
                                    (bt:make-lock)))))
         (bt:join-thread
          (bt:make-thread
           (lambda ()
             (bt:with-lock-held (,lock)
               (restore ,path ,profile ,buffer)
               ;; Returning result of the body is better than implicit `store' result.
               (prog1
                   (progn ,@body)
                 (store ,path ,profile ,buffer))))))))))
```

### How Has This Been Tested

I've created a toy-mode for bookmarks isolation (don't take it serious, it has some faulty pieces but was sufficient for feature testing):
``` lisp 
(uiop:define-package :nyxt/isolate-bookmarks-mode
  (:use :common-lisp :nyxt)
  (:import-from #:serapeum #:export-always)
  (:documentation "Mode for per-buffer bookmark isolation."))
(in-package :nyxt/isolate-bookmarks-mode)

(defclass isolate-bookmarks-profile (default-data-profile)
  ((name :initform "isolate-bookmarks")
   (bookmarks-data :allocation :instance)))

(define-mode isolate-bookmarks-mode ()
  "Isolate bookmarks for one buffer and persist them to a separate file."
  ((previous-bookmarks-path :accessor previous-bookmarks-path
                            :type (or null data-path)
                            :initform nil
                            :documentation "The path that bookmarks were stored to
before `isolate-bookmarks-mode' was activated.")
   (previous-data-profile :accessor previous-data-profile
                          :type (or null data-profile)
                          :initform nil
                          :documentation "The profile that the buffer had
before `isolate-bookmarks-mode' was activated.")
   (constructor
    :initform #'(lambda (mode)
                  (let ((buffer (buffer mode)))
                    (setf (previous-bookmarks-path mode) (bookmarks-path buffer)
                          (previous-data-profile mode) (data-profile buffer)
                          (data-profile buffer) (make-instance
                                                 'isolate-bookmarks-profile)
                          (bookmarks-path buffer) (make-instance
                                                   'nyxt:bookmarks-data-path
                                                   :basename (format nil "bookmarks-~a"
                                                                     (gensym "isolate")))
                          (bookmarks-data (data-profile buffer)) nil)
                    (store (bookmarks-path buffer) (data-profile buffer) buffer))))
   (destructor
    :initform #'(lambda (mode)
                  (let ((buffer (buffer mode)))
                    (uiop:delete-file-if-exists (expand-path (bookmarks-path buffer) buffer))
                    (setf (bookmarks-path buffer) (previous-bookmarks-path mode)
                          (data-profile buffer) (previous-data-profile mode)
                          (bookmarks-data (data-profile buffer)) nil)
                    (restore (bookmarks-path buffer) (data-profile buffer) buffer))))))

```
and ran Nyxt with it. Bookmarks were isolated and saved to a dedicated file! 

I've also ran Nyxt for some time to test whether the changes could have broken anything. Doesn't seem like so.

Additional testing is welcome though, because I may not notice some bugs due to my Nyxt use-style.

I'll be glad to know what you think of it!

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -46,11 +46,10 @@ Without handler, return ARG.  This is an acceptable `combination' for
 You must understand the risks before enabling this: a privliged user with access
 to your system can then take control of the browser and execute arbitrary code
 under your user profile.")
-   (data-profile (or (find-data-profile (getf *options* :data-profile))
-                     +default-data-profile+)
-                 :type data-profile
-                 :documentation "Profile to use for all persisted files.
-See the `data-path' class and the `expand-path' function.")
+;; TODO: is it always user's data? Better name maybe?
+   (user-data-cache (make-hash-table :test #'equal)
+                    :type hash-table
+                    :documentation "Table that maps the expanded `data-path's to the `user-data' (possibly) stored there.")
    (socket-thread nil                   ; TODO: Unexport?
                   :type t
                   :documentation "Thread that listens on socket.
@@ -132,93 +131,21 @@ FALLBACK-URL is empty, SEARCH-URL is used on an empty search.
 The engine with the \"default\" shortcut (or the first engine if there is no
 \"default\") is used when the query is not a valid URL, or the first keyword is
 not recognized.")
-   (key-stack '()
-              :export nil
-              :documentation "A stack that keeps track of the key chords a user has pressed.")
-   (downloads '()
-              :export nil
-              :documentation "List of downloads.")
    (download-watcher nil
                      :type t
                      :export nil
                      :documentation "List of downloads.")
-   (download-path :accessor download-path
-                  :type data-path
-                  :initform (make-instance 'download-data-path
-                                           :dirname (xdg-download-dir))
-                  :documentation "Path of directory where downloads will be
-stored.  Nil means use system default.")
+   (key-stack '()
+              :documentation "A stack that keeps track of the key chords a user has pressed.")
    (startup-timestamp (local-time:now)
                       :export nil
                       :documentation "`local-time:timestamp' of when Nyxt was started.")
    (init-time 0.0
               :export nil
               :documentation "Init time in seconds.")
-   (history-data nil
-                 :accessor nil
-                 :type t
-                 :export nil
-                 :documentation "The history data kept in memory.")
-   (history-path (make-instance 'history-data-path :basename "history")
-                 :type data-path
-                 :documentation "
-The path where the system will create/save the global history.")
-   (history-store-function :initarg :history-store-function
-                           :accessor history-store-function
-                           :type (or function null)
-                           :initform #'store-sexp-history
-                           :documentation "
-The function which stores the global history into `history-path'.")
-   (history-restore-function #'restore-sexp-history
-                             :type (or function null)
-                             :documentation "
-The function which restores the global history from `history-path'.")
-   (bookmarks-data nil
-                   :accessor nil
-                   :type t
-                   :export nil
-                   :documentation "The bookmarks kept in memory.")
-   (bookmarks-path (make-instance 'bookmarks-data-path :basename "bookmarks")
-                   :type data-path
-                   :documentation "
-The path where the system will create/save the bookmarks.")
-   (bookmarks-store-function #'store-sexp-bookmarks
-                             :type (or function null)
-                             :documentation "
-The function which stores the bookmarks into `bookmarks-path'.")
-   (bookmarks-restore-function #'restore-sexp-bookmarks
-                               :type (or function null)
-                               :documentation "
-The function which restores the bookmarks from `bookmarks-path'.")
-   (session-path (make-instance 'session-data-path
-                                :basename "default"
-                                :dirname (uiop:xdg-data-home +data-root+ "sessions"))
-                 :type data-path
-                 :documentation "
-The path where the system will create/save the session.")
-   (session-store-function #'store-sexp-session
-                           :type (or function null)
-                           :documentation "The function which stores the session
-into `session-path'.")
-   (session-restore-function #'restore-sexp-session
-                             :type (or function null)
-                             :documentation "The function which restores the session
-from `session-path'.")
    (session-restore-prompt :always-ask
                            :documentation "Ask whether to restore the
 session. Possible values are :always-ask :always-restore :never-restore.")
-   (auto-mode-rules '()
-                    :documentation "The list of auto-mode rules kept in memory.")
-   (auto-mode-rules-data-path (make-instance 'auto-mode-rules-data-path
-                                             :basename "auto-mode-rules")
-                              :type data-path
-                              :documentation "The path where the auto-mode rules are saved.")
-   (standard-output-path (make-instance 'data-path :basename "standard-out.txt")
-                         :type data-path
-                         :documentation "Path where `*standard-output*' can be written to.")
-   (error-output-path (make-instance 'data-path :basename "standard-error.txt")
-                      :type data-path
-                      :documentation "Path where `*error-output*' can be written to.")
    ;; Hooks follow:
    (before-exit-hook (hooks:make-hook-void)
                      :type hooks:hook-void
@@ -279,6 +206,11 @@ editor executable."))
 
 (define-user-class browser)
 
+(defmethod get-user-data ((profile default-data-profile) (path data-path))
+  (sera:and-let* ((expanded-path (expand-path path)))
+    (alex:ensure-gethash expanded-path (user-data-cache *browser*)
+                         (make-instance 'user-data))))
+
 (defmethod get-containing-window-for-buffer ((buffer buffer)
                                              (browser browser))
   "Get the window containing a buffer."
@@ -303,37 +235,6 @@ editor executable."))
   (when (null browser)
     (error "There is no current *browser*. Is Nyxt started?")))
 
-(defmethod history-data ((browser browser))
-  "Return the `history-data' slot from BROWSER.
-If empty, the history data is initialized with `history-restore-function'."
-  (when (and (null (slot-value browser 'history-data))
-             (history-restore-function browser))
-    (funcall-safely (history-restore-function browser)))
-  (slot-value browser 'history-data))
-
-(defmethod (setf history-data) (value (browser browser))
-  "Set `history-data' to VALUE.
-Persist the `history-data' slot from BROWSER to `history-path' with
-`history-store-function'."
-  (setf (slot-value browser 'history-data) value)
-  (match (history-store-function browser)
-    ((guard f f) (funcall-safely f))))
-
-(defmethod bookmarks-data ((browser browser))
-  "Return the `bookmarks-data' slot from BROWSER.
-If empty, the bookmarks data is initialized with `bookmarks-restore-function'."
-  (when (and (null (slot-value browser 'bookmarks-data))
-             (bookmarks-restore-function browser))
-    (funcall-safely (bookmarks-restore-function browser)))
-  (slot-value browser 'bookmarks-data))
-
-(defmethod (setf bookmarks-data) (value (browser browser))
-  "Set `bookmarks-data' to VALUE.
-Persist the `bookmarks-data' slot from BROWSER to `bookmarks-path' with
-`bookmarks-store-function'."
-  (setf (slot-value browser 'bookmarks-data) value)
-  (match (bookmarks-store-function browser)
-    ((guard f f) (funcall-safely f))))
 
 (defun download-watch ()
   "Update the download-list buffer.
@@ -364,8 +265,8 @@ This function is meant to be run in the background."
                 download))
 (export-always 'download)
 (defun download (url &key
-                     cookies
-                     (proxy-address :auto))
+                       cookies
+                       (proxy-address :auto))
   "Download URL.
 When PROXY-ADDRESS is :AUTO (the default), the proxy address is guessed from the
 current buffer."
@@ -373,7 +274,8 @@ current buffer."
   (when (eq proxy-address :auto)
     (setf proxy-address (proxy-address (current-buffer)
                                        :downloads-only t)))
-  (let ((download-dir (expand-path (download-path *browser*))))
+  (let* ((path (download-path (current-buffer)))
+         (download-dir (expand-path path)))
     (declare (type (or quri:uri null) proxy-address))
     (when download-dir
       (let* ((download nil))
@@ -383,7 +285,7 @@ current buffer."
                                  :directory download-dir
                                  :cookies cookies
                                  :proxy proxy-address))
-                 (push download (downloads *browser*))
+                 (push download (get-data path))
                  download))
           (if *keep-alive*
               (unsafe-download)
@@ -689,17 +591,20 @@ sometimes yields the wrong reasult."
 
 (defmethod write-output-to-log ((browser browser))
   "Set the *standard-output* and *error-output* to write to a log file."
-  (values
-   (setf *standard-output*
-         (open (expand-path (standard-output-path browser))
-               :direction :output
-               :if-does-not-exist :create
-               :if-exists :append))
-   (setf *error-output*
-         (open (expand-path (error-output-path browser))
-               :direction :output
-               :if-does-not-exist :create
-               :if-exists :append))))
+  (let ((buffer (current-buffer)))
+    (values
+     (sera:and-let* ((path (expand-path (standard-output-path buffer))))
+       (setf *standard-output*
+             (open path
+                   :direction :output
+                   :if-does-not-exist :create
+                   :if-exists :append)))
+     (sera:and-let* ((path (expand-path (error-output-path buffer))))
+       (setf *error-output*
+             (open path
+                   :direction :output
+                   :if-does-not-exist :create
+                   :if-exists :append))))))
 
 (defmacro define-ffi-method (name arguments)
   `(progn
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -13,6 +13,11 @@
        :documentation "Unique identifier for a buffer.
 Dead buffers or placeholder buffers (i.e. those not associated with a web view)
 have an empty ID.")
+   (data-profile (make-instance (or (find-data-profile (getf *options* :data-profile))
+                                              'default-data-profile))
+                 :type data-profile
+                 :documentation "Profile to use for all persisted files.
+See the `data-path' class and the `expand-path' function.")
    ;; TODO: Or maybe a dead-buffer should just be a buffer history?
    (url (quri:uri ""))
    (url-at-point (quri:uri ""))
@@ -144,9 +149,6 @@ distance scroll-left or scroll-right will scroll.")
                       :documentation "The ratio of the page to scroll.
 A value of 0.95 means that the bottom 5% will be the top 5% when scrolling
 down.")
-   (cookies-path (make-instance 'cookies-data-path :basename "cookies.txt")
-                 :documentation "The path where cookies are stored.  Not all
-renderers might support this.")
    (box-style (cl-css:css
                '((".nyxt-hint"
                   :background "linear-gradient(#fcff9e, #efcc00)"
@@ -180,6 +182,43 @@ and must return a (possibly new) URL.")
                        :type hook-buffer
                        :documentation "Hook run before `buffer-delete' takes effect.
 The handlers take the buffer as argument.")
+   (session-path (make-instance 'session-data-path
+                                          :basename "default"
+                                          :dirname (uiop:xdg-data-home +data-root+ "sessions"))
+                 :type data-path
+                 :documentation "
+The path where the system will create/save the session.")
+   (download-path (make-instance 'download-data-path
+                                 :dirname (xdg-download-dir))
+                  :type data-path
+                  :documentation "Path of directory where downloads will be
+stored.  Nil means use system default.
+Downloads are kept in browser's `user-data', keyed by the expanded `download-path'.")
+   (history-path (make-instance 'history-data-path :basename "history")
+                 :type data-path
+                 :documentation "
+The path where the system will create/save the global history.
+History data is kept in browser's `user-data', keyed by the expanded `history-path'.")
+   (bookmarks-path (make-instance 'bookmarks-data-path :basename "bookmarks")
+                   :type data-path
+                   :documentation "
+The path where the system will create/save the bookmarks.
+Bookmarks' data is kept in browser's `user-data', keyed by the expanded `bookmarks-path'.")
+   (cookies-path (make-instance 'cookies-data-path :basename "cookies.txt")
+                 :type data-path
+                 :documentation "The path where cookies are stored.  Not all
+renderers might support this.")
+   (auto-mode-rules-path (make-instance 'auto-mode-rules-data-path
+                                        :basename "auto-mode-rules")
+                         :type data-path
+                         :documentation "The path where the auto-mode rules are saved.
+Rules are kept in browser's `user-data', keyed by the expanded `auto-mode-rules-path'.")
+   (standard-output-path (make-instance 'standard-output-data-path :basename "standard-out.txt")
+                         :type data-path
+                         :documentation "Path where `*standard-output*' can be written to.")
+   (error-output-path (make-instance 'error-output-data-path :basename "standard-error.txt")
+                      :type data-path
+                      :documentation "Path where `*error-output*' can be written to.")
    (default-cookie-policy :no-third-party
                           :type cookie-policy
                           :documentation "Cookie policy of new buffers.
@@ -412,11 +451,12 @@ If URL is `:default', use `default-new-buffer-url'."
   (buffer-make *browser* :title title :default-modes modes :internal-buffer-p t))
 
 (declaim (ftype (function (browser &key (:title string)
+                                   (:data-profile data-profile)
                                    (:default-modes list)
                                    (:dead-buffer buffer)
                                    (:internal-buffer-p boolean)))
                 buffer-make))
-(defun buffer-make (browser &key title default-modes dead-buffer internal-buffer-p)
+(defun buffer-make (browser &key data-profile title default-modes dead-buffer internal-buffer-p)
   "Make buffer with title TITLE and modes DEFAULT-MODES.
 Run `*browser*'s `buffer-make-hook' over the created buffer before returning it.
 If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
@@ -428,7 +468,8 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
                      (apply #'make-instance (if internal-buffer-p 'user-internal-buffer 'user-buffer)
                             :id (get-unique-buffer-identifier *browser*)
                             (append (when title `(:title ,title))
-                                    (when default-modes `(:default-modes ,default-modes)))))))
+                                    (when default-modes `(:default-modes ,default-modes))
+                                    (when data-profile `(:data-profile ,data-profile)))))))
     (hooks:run-hook (buffer-before-make-hook *browser*) buffer)
     ;; Modes might require that buffer exists, so we need to initialize them
     ;; after the view has been created.
@@ -468,9 +509,7 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
     (buffers-delete (id buffer))
     (setf (id buffer) "")
     (add-to-recent-buffers buffer)
-    (match (session-store-function *browser*)
-      ((guard f f)
-       (funcall-safely f)))))
+    (store (data-profile buffer) (session-path buffer))))
 
 (export-always 'buffer-list)
 (defun buffer-list (&key sort-by-time domain)
--- a/source/configuration.lisp
+++ b/source/configuration.lisp
@@ -143,31 +143,3 @@ Initialization file use case:
     (error ()
       ;; Improper list.
       list)))
-
-(export-always 'expand-path)
-(declaim (ftype (function ((or null data-path)) (or string null)) expand-path))
-(defun expand-path (data-path)
-  "Return the expanded path of DATA-PATH or nil if there is none.
-`expand-data-path' is dispatched against `data-path' and `*browser*'s
-`data-profile' if `*browser*' is instantiated, `+default-data-profile+'
-otherwise.
-This function can be used on browser-less globals like `*init-file-path*'."
-  (when data-path
-    (the (values (or string null) &optional)
-         (match (if *browser*
-                    (expand-data-path (data-profile *browser*) data-path)
-                    (expand-data-path +default-data-profile+ data-path))
-           ("" nil)
-           (m (uiop:native-namestring m))))))
-
-(export-always 'with-data-file)
-(defmacro with-data-file ((stream data-path &rest options) &body body)
-  "Evaluate BODY with STREAM bound to DATA-PATH.
-DATA-PATH can be a GPG-encrypted file if it ends with a .gpg extension.
-If DATA-PATH expands to NIL or the empty string, do nothing.
-OPTIONS are as for `open'.
-Parent directories are created if necessary."
-  `(let ((path (expand-path ,data-path)))
-     (when path
-       (with-maybe-gpg-file (,stream path ,@options)
-         ,@body))))
--- a/source/data-storage.lisp
+++ b/source/data-storage.lisp
@@ -9,39 +9,6 @@
 
 (defvar *gpg-program* "gpg")
 
-(define-class data-profile ()
-  ((name ""
-         :documentation "The name of the profile to refer it with."))
-  (:export-class-name-p t)
-  (:export-accessor-names-p t)
-  (:accessor-name-transformer #'class*:name-identity))
-
-(export-always '+default-data-profile+)
-(defvar +default-data-profile+ (make-instance 'data-profile :name "default")
-  "With the default profile all data is persisted to the standard locations.")
-
-(export-always '+private-data-profile+)
-(defvar +private-data-profile+ (make-instance 'data-profile :name "private")
-  "With the private profile no data should be persisted to disk.")
-
-(defun package-data-profiles ()
-  "Return the list of data profiles a (DATA-PROFILE NAME DOCSTRING) tuples."
-  (sort (mapcar (lambda (profile)
-                  (list (symbol-value profile)
-                        (name (symbol-value profile))
-                        (documentation profile 'variable)))
-                (delete-if (lambda (sym)
-                             (not (ignore-errors
-                                   (mopu:subclassp (class-of (symbol-value sym))
-                                                   'data-profile))))
-                           (package-variables)))
-        #'string< :key #'second))
-
-(export-always 'find-data-profile)
-(defun find-data-profile (name)
-  "Return profile matching NAME.
-Return NIL on no match."
-  (first (find name (package-data-profiles) :test #'string= :key #'second)))
 
 (define-class data-path ()
   ((dirname ""
@@ -84,6 +51,82 @@ This can be used to set the path from command line.  See
   ((ref :initform "auto-mode-rules"))
   (:export-class-name-p t)
   (:accessor-name-transformer #'class*:name-identity))
+(define-class standard-output-data-path (data-path)
+  ((ref :initform "standard-output"))
+  (:export-class-name-p t)
+  (:accessor-name-transformer #'class*:name-identity))
+(define-class error-output-data-path (data-path)
+  ((ref :initform "error-output"))
+  (:export-class-name-p t)
+  (:accessor-name-transformer #'class*:name-identity))
+
+(export-always 'xdg-download-dir)
+(defun xdg-download-dir ()
+  (let ((dir (ignore-errors (uiop:run-program '("xdg-user-dir" "DOWNLOAD")
+                                              :output '(:string :stripped t)))))
+    (when (or (null dir) (string= dir (uiop:getenv "HOME")))
+      (setf dir (uiop:getenv "XDG_DOWNLOAD_DIR")))
+    (unless dir
+      (setf dir (str:concat (uiop:getenv "HOME") "/Downloads/")))
+    (unless (str:ends-with? "/" dir)
+      (setf dir (str:concat dir "/")))
+    dir))
+
+(define-class data-profile ()
+  ((name ""
+         :documentation "The name of the profile to refer it with."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer #'class*:name-identity))
+
+(define-class default-data-profile (data-profile)
+  ((name :initform "default"))
+  (:export-class-name-p t)
+  (:documentation "With the default profile all data is persisted to the standard locations."))
+
+(define-class private-data-profile (data-profile)
+  ((name :initform "private")
+   (user-data-cache (make-hash-table :test #'equal)
+                    :type hash-table
+                    :documentation "Buffer-local `user-data-cache' to isolate the data of a private buffer."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer #'class*:name-identity)
+  (:documentation "With the private profile no data should be persisted to disk.
+No data should be shared with other buffers either."))
+
+(export-always '*global-data-profile*)
+(defvar *global-data-profile* (make-instance 'default-data-profile)
+  "The profile to use in the absence of buffers and on browser-less variables.")
+
+(export-always 'current-data-profile)
+(defun current-data-profile ()
+  "If `%buffer' is non-nil, return its data-profile.
+Return `*global-data-profile*' otherwise."
+  ;; TODO: %BUFFER is not defined yet. Move %BUFFER there?
+  (let ((buffer (current-buffer)))
+    (or (and buffer (data-profile buffer))
+        *global-data-profile*)))
+
+(defun package-data-profiles ()
+  "Return the list of data profiles as a (DATA-PROFILE-SYM NAME DOCSTRING) tuples."
+  (sort (mapcar (lambda (profile)
+                  (list profile
+                        (getf (mopu:slot-properties profile 'name)
+                              :initform)
+                        (documentation profile 'type)))
+                (delete-if (lambda (sym)
+                             (not (ignore-errors
+                                   (and (not (eq sym 'data-profile))
+                                        (mopu:subclassp sym 'data-profile)))))
+                           (package-classes)))
+        #'string< :key #'second))
+
+(export-always 'find-data-profile)
+(defun find-data-profile (name)
+  "Return profile matching NAME.
+Return NIL on no match."
+  (first (find name (package-data-profiles) :test #'string= :key #'second)))
 
 (declaim (ftype (function (string) (or string null)) find-ref-path))
 (defun find-ref-path (ref)
@@ -108,7 +151,7 @@ place of PATH `basename'.
 - When PATH has no basename, return ROOT.  This is useful to refer to
   directories.
 - If basename has a slash, return basename.
-- Otherwise expand to 'ROOT/basename.lisp' or 'ROOT/basname' if the basename
+- Otherwise expand to 'ROOT/basename.lisp' or 'ROOT/basename' if the basename
   already contains a period."
   (let ((name (match (find-ref-path (ref path))
                 (nil (namestring (basename path)))
@@ -131,20 +174,16 @@ Return NIL when path must not be used.  This makes it possible to use the
 function result as a boolean in conditions."
   (expand-default-path path))
 
-(export-always 'xdg-download-dir)
-(defun xdg-download-dir ()
-  (let ((dir (ignore-errors (uiop:run-program '("xdg-user-dir" "DOWNLOAD")
-                                              :output '(:string :stripped t)))))
-    (when (or (null dir) (string= dir (uiop:getenv "HOME")))
-      (setf dir (uiop:getenv "XDG_DOWNLOAD_DIR")))
-    (unless dir
-      (setf dir (str:concat (uiop:getenv "HOME") "/Downloads/")))
-    (unless (str:ends-with? "/" dir)
-      (setf dir (str:concat dir "/")))
-    dir))
+(defmethod expand-data-path ((profile private-data-profile) (path cookies-data-path))
+  "We shouldn't store cookies for `private-data-profile'."
+  nil)
 
-(defmethod expand-data-path ((profile (eql +private-data-profile+)) (path data-path))
-  "Don't persist anything in private mode."
+(defmethod expand-data-path ((profile private-data-profile) (path standard-output-data-path))
+  "We shouldn't store `*standard-output*' for `private-data-profile'."
+  nil)
+
+(defmethod expand-data-path ((profile private-data-profile) (path error-output-data-path))
+  "We shouldn't store `*error-output*' for `private-data-profile'."
   nil)
 
 (export-always 'ensure-parent-exists)
@@ -154,6 +193,88 @@ function result as a boolean in conditions."
   "Create parent directories of PATH if they don't exist and return PATH."
   (ensure-directories-exist (directory-namestring path))
   path)
+
+(export-always 'store)
+(defgeneric store (profile path)
+  (:method ((profile data-profile) (path data-path))
+    (error "No store method defined for ~a and ~a."
+           (serapeum:class-name-of path)
+           (serapeum:class-name-of profile)))
+  (:documentation "The generic way to store data to the given path type.
+Define a method for your `data-path' type to make it storable."))
+
+(export-always 'restore)
+(defgeneric restore (profile path)
+  (:method ((profile data-profile) (path data-path))
+    (error "No restore method defined for ~a and ~a."
+           (serapeum:class-name-of path)
+           (serapeum:class-name-of profile)))
+  (:documentation "The generic way to restore data from the given path type.
+Define a method for your `data-path' type to make it restorable."))
+
+(defmethod store ((profile private-data-profile) (path data-path))
+  nil)
+
+(defmethod restore ((profile private-data-profile) (path data-path))
+  nil)
+
+(export-always 'expand-path)
+(declaim (ftype (function ((or null data-path)) (or string null)) expand-path))
+(defun expand-path (data-path)
+  "Return the expanded path of DATA-PATH or nil if there is none.
+`expand-data-path' is dispatched against `data-path' and `current-data-profile'
+`data-profile' if there are buffers, `*global-data-profile*' otherwise.
+This function can be used on browser-less globals like `*init-file-path*'."
+  (when data-path
+    (the (values (or string null) &optional)
+         (match (expand-data-path (current-data-profile) data-path)
+           ("" nil)
+           (m (uiop:native-namestring m))))))
+
+;; TODO: create subclasses for history, session, bookmark, auto-mode data to ensure typing?
+(define-class user-data ()
+  ((data nil
+         :documentation "The meaningful data to store. Examples: history data, session, bookmarks.")
+   (lock (bt:make-recursive-lock)
+         :type bt:lock
+         :documentation "The lock to guard from race conditions on the access to this data."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer #'class*:name-identity)
+  (:documentation "The class to mediate the data keeping."))
+
+(export-always 'get-user-data)
+(defgeneric get-user-data (profile path)
+  (:documentation "Access the browsing-related data depending on the `data-profile'."))
+
+(defmethod get-user-data ((profile private-data-profile) (path data-path))
+  "Look up the buffer-local data in case of `private-data-profile'."
+  (sera:and-let* ((expanded-path (expand-path path)))
+    (alex:ensure-gethash expanded-path (user-data-cache profile)
+                         (make-instance 'user-data))))
+
+(export-always 'get-data)
+;; TODO: Better name? Isn't it too wide?
+(defmethod get-data ((path data-path))
+  (data (get-user-data (current-data-profile) path)))
+
+(defmethod (setf get-data) (value (path data-path))
+  (setf (data (get-user-data (current-data-profile) path)) value))
+
+(export-always 'with-data-access)
+(defmacro with-data-access (data-var data-path &body body)
+  "Lock the data for the BODY to avoid race conditions.
+Bind the DATA-VAR to the value of the data from DATA-PATH to reuse it."
+  (alex:with-gensyms (lock path-name)
+    `(let* ((,path-name ,data-path)
+            (,lock (lock (get-user-data (current-data-profile) ,path-name))))
+       (bt:with-recursive-lock-held (,lock)
+         (restore (current-data-profile) ,path-name)
+         (let ((,data-var (get-data ,path-name)))
+           (unwind-protect
+                (progn ,@body)
+             (setf (get-data ,path-name) ,data-var)
+             (store (current-data-profile) ,path-name)))))))
 
 (defvar *gpg-default-recipient* nil)
 
@@ -301,3 +422,16 @@ Parent directories are created if necessary."
             `(ensure-parent-exists ,filespec))
          (with-open-file (,stream ,filespec ,@options)
            ,@body))))
+
+(export-always 'with-data-file)
+(defmacro with-data-file ((stream data-path &rest options) &body body)
+  "Evaluate BODY with STREAM bound to DATA-PATH.
+DATA-PATH can be a GPG-encrypted file if it ends with a .gpg extension.
+If DATA-PATH expands to NIL or the empty string, do nothing.
+OPTIONS are as for `open'.
+Parent directories are created if necessary."
+  (alex:with-gensyms (path)
+    `(let ((,path (expand-path ,data-path)))
+       (when ,path
+         (with-maybe-gpg-file (,stream ,path ,@options)
+           ,@body)))))
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -49,39 +49,36 @@ The total number of visit for a given URL is (+ explicit-visits implicit-visits)
 (declaim (ftype (function (quri:uri &key (:title string) (:explicit t)) t) history-add))
 (export-always 'history-add)
 (defun history-add (uri &key title explicit)
-  "Add URL to the global history.
+  "Add URL to the global/buffer-local history.
 The `implicit-visits' count is incremented unless EXPLICIT is non-nil, in which
 case `explicit-visits'.
 The history is sorted by last access."
-  (unless (url-empty-p uri)
-    (unless (history-data *browser*)
-      (setf (slot-value *browser* 'history-data)
-            (make-hash-table :test #'equal)))
-    (let ((entry (gethash (object-string uri) (history-data *browser*))))
-      (unless entry
-        (setf entry (make-instance 'history-entry :url uri)))
-      (if explicit
-          (incf (explicit-visits entry))
-          (incf (implicit-visits entry)))
-      (setf (last-access entry) (local-time:now))
-      ;; Always update the title since it may have changed since last visit.
-      (setf (title entry) title)
-      (setf (gethash (object-string uri) (history-data *browser*)) entry)
-      ;; Use accessor to ensure store function is called.
-      (setf (history-data *browser*) (history-data *browser*)))))
+  (with-data-access history (history-path (current-buffer))
+    (unless (url-empty-p uri)
+      (unless history
+        (setf history (make-hash-table :test #'equal)))
+      (let ((entry (gethash (object-string uri) history)))
+        (unless entry
+          (setf entry (make-instance 'history-entry :url uri)))
+        (if explicit
+            (incf (explicit-visits entry))
+            (incf (implicit-visits entry)))
+        (setf (last-access entry) (local-time:now))
+        ;; Always update the title since it may have changed since last visit.
+        (setf (title entry) title)
+        (setf (gethash (object-string uri) history) entry)))))
 
 (define-command delete-history-entry ()
   "Delete queried history entries."
-  (with-result (entries (read-from-minibuffer
-                         (make-minibuffer
-                          :input-prompt "Delete entries"
-                          :suggestion-function (history-suggestion-filter)
-                          :history (minibuffer-set-url-history *browser*)
-                          :multi-selection-p t)))
-    (dolist (entry entries)
-      (remhash (object-string (url entry)) (history-data *browser*)))
-    ;; Use accessor to ensure store function is called.
-    (setf (history-data *browser*) (history-data *browser*))))
+  (with-data-access history (history-path (current-buffer))
+    (with-result (entries (read-from-minibuffer
+                           (make-minibuffer
+                            :input-prompt "Delete entries"
+                            :suggestion-function (history-suggestion-filter)
+                            :history (minibuffer-set-url-history *browser*)
+                            :multi-selection-p t)))
+      (dolist (entry entries)
+        (remhash (object-string (url entry)) history)))))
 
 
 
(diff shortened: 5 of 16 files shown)


aartaka:
Small bugfix: `start-browser` fas failing if no profile was provided as CLI argument. The reason was that the result of `find-data-profile` wasn't null-checked.

source/configuration.lisp
```diff
@@ -127,29 +127,40 @@ Initialization file use case:
...
+`data-profile' if BUFFER is present, `*global-data-profile*' otherwise.
 This function can be used on browser-less globals like `*init-file-path*'."
   (when data-path
     (the (values (or string null) &optional)
-         (match (if *browser*
-                    (expand-data-path (data-profile *browser*) data-path)
-                    (expand-data-path +default-data-profile+ data-path))
+         (match (if buffer
+                    (expand-data-path (data-profile buffer) data-path)
+                    (expand-data-path *global-data-profile* data-path))
```
Ambrevar, line 150: You can move the `if` to the 1st argument of `expand-data-path.

source/configuration.lisp
```diff
@@ -127,29 +127,40 @@ Initialization file use case:
...
 DATA-PATH can be a GPG-encrypted file if it ends with a .gpg extension.
 If DATA-PATH expands to NIL or the empty string, do nothing.
+BUFFER is the buffer to take current `data-profile' settings from.
 OPTIONS are as for `open'.
 Parent directories are created if necessary."
-  `(let ((path (expand-path ,data-path)))
-     (when path
-       (with-maybe-gpg-file (,stream path ,@options)
-         ,@body))))
+  (alex:with-gensyms (path)
```
Ambrevar, line 162: Good catch!

source/data-storage.lisp
```diff
@@ -93,6 +59,88 @@ This can be used to set the path from command line.  See
...
+   (bookmarks-data :allocation :instance)
+   (auto-mode-rules :allocation :instance))
+  (:documentation "With the private profile no data should be persisted to disk."))
+
+(export-always '*global-data-profile*)
+(defvar *global-data-profile* (make-instance 'default-data-profile)
+  "The profile to use in the absence of buffers and on browser-less variables.")
+
+(defun package-data-profiles ()
+  "Return the list of data profiles a (DATA-PROFILE-SYM NAME DOCSTRING) tuples."
```
Ambrevar, line 125: Good opportunity to fix the typo here: "as a" ;)

source/data-storage.lisp
```diff
@@ -137,21 +185,12 @@ place of PATH `basename'.
   "Return finalized path.
 Return NIL when path must not be used.  This makes it possible to use the
 function result as a boolean in conditions."
+  ;; TODO: this makes `default-data-profile' somewhat unnecessary.
+  ;; Maybe throw errors here instead of expanding
+  ;; and expand for `default-data-profile' only?
```
Ambrevar, line 190: I don't understand why the `default-data-profile` class would be unnecessary.  Can you give an example?

source/data-storage.lisp
```diff
@@ -163,6 +202,29 @@ function result as a boolean in conditions."
   (ensure-directories-exist (directory-namestring path))
   path)
 
+(export-always 'store)
+;; TODO: BUFFER is almost always used in `store'/`restore' methods. Make it positional?
```
Ambrevar, line 206: Positional?  You mean non-optional?

source/data-storage.lisp
```diff
@@ -163,6 +202,29 @@ function result as a boolean in conditions."
...
+(defgeneric store (path profile &optional buffer)
+  (:method ((path data-path) (profile data-profile) &optional buffer)
+    (declare (ignore buffer))
+    (error "No store method defined for ~a" (serapeum:class-name-of path)))
+  (:method ((path data-path) (profile private-data-profile) &optional buffer)
+    (declare (ignore buffer))
+    nil)
```
Ambrevar, line 213: I think it would be more explicit and readable to define the private-data-profile method separately, and together with the `restore` method.

source/data-storage.lisp
```diff
@@ -163,6 +202,29 @@ function result as a boolean in conditions."
...
+Define a method for your `data-path' type to make it storable."))
+
+(export-always 'restore)
+(defgeneric restore (path profile &optional buffer)
+  (:method ((path data-path) (profile data-profile) &optional buffer)
+    (declare (ignore buffer))
+    (error "No restore method defined for ~a" (serapeum:class-name-of path)))
+  (:method ((path data-path) (profile private-data-profile) &optional buffer)
+    (declare (ignore buffer))
+    nil)
```
Ambrevar, line 224: See above.

source/history.lisp
```diff
@@ -43,42 +43,43 @@ The total number of visit for a given URL is (+ explicit-visits implicit-visits)
...
-          (incf (explicit-visits entry))
-          (incf (implicit-visits entry)))
-      (setf (last-access entry) (local-time:now))
-      ;; Always update the title since it may have changed since last visit.
-      (setf (title entry) title)
-      (setf (gethash (object-string uri) (history-data *browser*)) entry)
-      ;; Use accessor to ensure store function is called.
-      (setf (history-data *browser*) (history-data *browser*)))))
+  (restore (history-path buffer) (data-profile buffer) buffer)
+  (let ((profile (if buffer (data-profile buffer) *global-data-profile*)))
```
Ambrevar, line 54: The `(if buffer ...)` idiom seems quite frequent.  Add a `(get-profile buffer)` helper?  (Choose a better name.)

source/renderer-gtk.lisp
```diff
@@ -107,7 +99,15 @@ See https://github.com/atlas-engineer/nyxt/issues/740")
    (proxy-ignored-hosts :accessor proxy-ignored-hosts
                         :initarg :proxy-ignored-hosts
                         :type list
-                        :initform '())))
+                        :initform '())
+   ;; TODO: define per-buffer directory names now, when it's a buffer slot?
```
Ambrevar, line 103: What do you mean with "now"?  What's the matter?

source/tutorial.lisp
```diff
@@ -136,8 +136,8 @@ place rather than having to jump around on a buffer (or multiple buffers).")
     (:li (command-markup 'nyxt/web-mode:remove-search-hints) ": Remove the highlighting around the search hits."))
    (:h3 "Bookmarks")
    (:p "The bookmark file "
-       (:code (expand-path (bookmarks-path (or *browser*
-                                               (make-instance 'browser)))))
+       ;; TODO: Can be incorrect if non-default.
+       (:code (expand-path (bookmarks-path (current-buffer))))
```
Ambrevar, line 140: This won't do because the tutorial / manual must be generable without a browser.
You need to make a dummy buffer here.  I'm doing this in other places in the manual / tutorial.

tests/test-global-history.lisp
```diff
@@ -9,14 +9,18 @@
...
-    (let ((*browser* (make-instance 'browser)))
-      (setf (history-restore-function *browser*) nil)
-      (history-add (quri:uri "http://example.org"))
-      (is (hash-table-count (nyxt::history-data *browser*))
+    (let ((*browser* (make-instance 'browser))
+          (buffer (nyxt::make-buffer)))
+      ;; We need to make sure nothing will be restored.
+      (setf (history-path buffer) (make-instance 'history-data-path
+                                                 :basename "history"
+                                                 :dirname "/tmp/history-test"))
```
Ambrevar, line 17: Make test temp data easily customizable to avoid breaking tests for packagers.
The dirname is a bit too generic.  I'd suggest it contains "nyxt" and maybe use an atomically created random temp dir (UIOP has a function for that i believe).

source/data-storage.lisp
jmercouris: in other words, required

source/data-storage.lisp
aartaka: I mean, it's uncertain what should we do when we have an `expand-data-path` on some new `data-profile` that is neither `default-data-profile` nor `private-data-profile`. The behavior of expanding it right away seems to be intuitive in case of `default-data-profile`, but do we want to have this behavior for any `data-profile`, unless specified otherwise?

Strictly speaking, it's not making `default-data-profile` unnecessary, but it takes some of its characteristic features and applies to any new profile. That's why it can be better to expand paths only on `default-data-profile` and throw errors in case of general `data-profile`, like in `store` and `restore`.

source/data-storage.lisp
aartaka: Yes!

Ambrevar:
Excellent!  Wow, this is behemoth work!  I didn't expect the change to
be so impactful! :p  Thank you so much for taking care of this!

Comments below:

> - The look of data-concerned (`buffer`, `data-profile`) class definition asks for new `defclass` (#889) — the number and similarity of slots is horrible xD

Coming very soon!

> - It&#39;s currently hard to know what profile is used by the browser at browser startup. Inspecting the `buffer` class with the help of MOP seems to be the only solid option, but it&#39;s not perfect either.

Why?  When exactly?  For the first buffer?  Can you give a more precise example?

> - Race conditions can happen on file access and we don&#39;t have any safety mechanism against it yet.

Indeed, let's fix it right after this is merged.

> - Ability to store data in browser instance to duplicate less data (although it&#39;s solved by `*global-data-profile*` already)?

All buffers that share the
same profile will share the same data.  All good here.

So what about buffers that have different profiles which expand to the
same data nonetheless?

I think we need a global hash-table that maps an expanded path to the
corresponding data.  Thoughts?



On a similar topic: I don't think using slots in data-profile is a good
idea.  What about user-defined or third-party-defined data that
want to use the default profile?

Instead, I think we need a hash-table which stores the data of the
expanded path.  As hinted above, this hash-table could be global to
avoid duplication across data-profiles.

> - Lock file access while `restore`-ing and `store`-ing data separately?

Isn't it the same as the 2nd item below "Make the storage safe from race conditions"?

> - Maybe making alias methods for data-storing slots specialized on
> buffers (`bookmarks-data` and friends), so that `(bookmarks-data
> buffer) = (bookmarks-data (data-profile buffer))`? I think it&#39;s
> faulty to use such aliases, but it looks prettier and somewhat more
> intuitive (alternative: actually move data-storing slots to `buffer`,
> instead of creating aliases -- not really useful though).

Let me take a step back in a slightly different direction.  Initially I
had introduced `expand-path` as a wrapper around `expand-data-path` that
would infer the data profile for the user.  The point was that we could
do

  (expand-path PATH)

instead of specifying the data-profile manually:

  (expand-data-path PROFILE PATH)

But if we are going to pass the BUFFER anyways, then there is no need
for expand-path.

Instead, we can add a helper function to retrieve the profile from a
potential buffer (may be nil).

All that said, do we have to pass the buffer all the time?  It's a bit
cumbersome.
Can we use the `with-current-buffer` macro instead?


Back to bookmarks-data: The problem of aliases is that they would be
define for each one of the default data-storing slots, but not for user
ones.  So it's probably a better idea to have a `expand-path`-like
function that gets the data-profile for you.

But do we really need that?  In my understanding, data-storing slot
manipulation is rather rare.


> (defmacro with-parallel-data-access (path buffer &amp;body body)
>   &quot;Macro to conveniently access persisted data in the data-altering functions.
> Exemplary applications are `history-add&#39;, `add-modes-to-auto-mode-rules&#39;.
> This macro spawns a thread that locks the data, restores the fresh data,
> and stores the (possibly) altered data back to the BUFFER&#39;s PATH-NAME.&quot;
>   (alex:once-only (buffer path)
>     (alex:with-gensyms (profile path-type lock)
>       `(let* ((,profile (if ,buffer (data-profile ,buffer) *global-data-profile*))
>               (,path-type (serapeum:class-name-of ,path))
>               ;; Locks are stored in a slot of `data-profile&#39; in the 
>               ;; initial design iteration, but it&#39;s not perfect
>               (,lock (gethash ,path-type (locks ,profile)
>                               (setf (gethash ,path-type (locks ,profile))
>                                     (bt:make-lock)))))
>          (bt:join-thread
>           (bt:make-thread
>            (lambda ()
>              (bt:with-lock-held (,lock)
>                (restore ,path ,profile ,buffer)
>                ;; Returning result of the body is better than implicit `store&#39; result.
>                (prog1
>                    (progn ,@body)
>                  (store ,path ,profile ,buffer))))))))))
> ```

You don't need to start a thread here.  You can simply use `bt:with-lock-held`.

I would also make this the default behaviour and replace
`with-data-access` with your draft.



A few other points:

- `store` could be renamed to the more unique and explicit `persist`.
But what about `restore`?  `persist` does not have an obvious antonym.

- You've removed the (setf DATA-STORING-SLOTS) method.  The drawback is
  that now we need to be careful to store / restore at the right time
  when getting/setting the data-storing slots.  Maybe we can restore the
  setf-method by using `with-current-buffer`?

Ambrevar:
I'll do more testing later.

source/data-storage.lisp
Ambrevar: I'm confused.  expanda-data-path should accept any data-profile, that's the point.  Why wouldn't you want to do it?

To clarify, `default-data-profile` and `private-data-profile` should not be considered special.  All data-profiles should be treated the same.

source/data-storage.lisp
Ambrevar: Maybe we can use `current-buffer` here and `with-current-buffer` from the caller.

source/renderer-gtk.lisp
aartaka: I mean, before `data-manager-path` was moved to `gtk-buffer` it was intuitive to define the `:base-data-directory` as 
``` lisp
:base-data-directory
(str:concat (expand-path (data-manager-path buffer))
                         "/nyxt-data-manager-" (id buffer))
```
in `make-data-manager`, because there was no other way to know which buffer the directory would belong to. But now, when `data-manager-path` is already a part of a `buffer`, we can initialize the buffer with the final `:base-data-directory` as the `:directory` of `data-manager-path`. Or can't we? It requires anaphoric `defclass`, and I heard that standard `defclass` is anaphoric, but I can be mistaken, because I can't find any reference to it.

source/history.lisp
aartaka: `maybe-data-profile`? It sounds more suitable for "either data-profile or nil" type of function though.

aartaka:
> Why? When exactly? For the first buffer? Can you give a more precise example?

Sure. It's browser initialization, namely `start-browser`. We can't know the profile user set without inspecting the `buffer` slots. Here's relatively working (but imperfect) code from `start-browser`:
``` lisp
;; TODO: Ideally, we need to know the initform that user could've set for
;; `data-profile' slot of `buffer'. Use `mopu:slot-properties'?
(log:info "Using data profile ~s." (name (make-instance
                                           (or (find-data-profile
                                                 (getf *options* :data-profile))
                                               'default-data-profile))))
```
 
> I think we need a global hash-table that maps an expanded path to the corresponding data. Thoughts? 
> On a similar topic: I don't think using slots in data-profile is a good idea. What about user-defined or third-party-defined data that want to use the default profile? Instead, I think we need a hash-table which stores the data of the expanded path. As hinted above, this hash-table could be global to avoid duplication across data-profiles.

This sounds good, although I feel there can be problems, especially when the buffer with `private-data-profile` and default paths will try to access and store data somewhere it shouldn't.

> - Lock file access while `restore`-ing and `store`-ing data separately?
> Isn't it the same as the 2nd item below "Make the storage safe from race conditions"?

It is related. The difference is that `with-parallel-data-access` is locking the data for a series of `restore`-`do-something`-`store` only, while `restore` and `store`, appearing in the code independently of `with-parallel-data-access` should lock data too.

> Let me take a step back in a slightly different direction. Initially I had introduced `expand-path` as a wrapper around `expand-data-path` that would infer the data profile for the user. The point was that we could do (expand-path PATH) instead of specifying the data-profile manually: (expand-data-path PROFILE PATH) But if we are going to pass the BUFFER anyways, then there is no need for expand-path. 

Fair point!

> Instead, we can add a helper function to retrieve the profile from a potential buffer (may be nil). 

Is it going to be the same as this `(if buffer (data-profile buffer) *global-data-profile*)` idiom?

> All that said, do we have to pass the buffer all the time? It's a bit cumbersome. Can we use the `with-current-buffer` macro instead? 

I don't yet understand how `with-current-buffer` can help us here. Can you expand on that, please?

> Back to bookmarks-data: The problem of aliases is that they would be define for each one of the default data-storing slots, but not for user ones. So it's probably a better idea to have a `expand-path`-like function that gets the data-profile for you. But do we really need that? In my understanding, data-storing slot manipulation is rather rare.

Agreed then. I'll be glad to know what you meant to do with `with-current-buffer`, after that I'll implement the alternative to `expand-path` (or just remove it xD).

> You don't need to start a thread here. You can simply use `bt:with-lock-held`. I would also make this the default behaviour and replace `with-data-access` with your draft.

But, again, where should we store locks? Global hash-table from expanded filenames to locks?

>  - `store` could be renamed to the more unique and explicit `persist`. But what about `restore`? `persist` does not have an obvious antonym. 

Don't know about that. Is there a necessity? Can we possibly have other processes to disambiguate these from? 

`nyxt/auto-mode:(re)store-last-active-modes` seem to collide with `(re)store` already.

> - You've removed the (setf DATA-STORING-SLOTS) method. The drawback is that now we need to be careful to store / restore at the right time when getting/setting the data-storing slots. Maybe we can restore the setf-method by using `with-current-buffer`?

Hmmmm. I'm not sure implicit storage on `setf` was a good idea in this context (it created several leaky-abstraction snippets like `(setf (history-data *browser*) (history-data *browser*))`), so I doubt it should be back. If we'll implement the `with-parallel-data-access` or incorporate it into `with-data-file`, we won't miss these implicitly stored slots :)

source/data-storage.lisp
aartaka: Makes sense :)

source/data-storage.lisp
aartaka: Isn't it too much responsibility to burden the caller with? 

Adding to that, it's easier to write aditional argument to `store` than `with-current-buffer` form, both in terms of lines of code and cognitive load (because macros are a complicated matter).

tests/test-global-history.lisp
aartaka: `uiop:temporary-directory` returns "/tmp/" on *nix systems, so it's not of help there. 

File is deleted right after the test, so adding "nyxt" to `:dirname` should suffice for almost any test setup.

aartaka:
> Instead, I think we need a hash-table which stores the data of the
expanded path.  As hinted above, this hash-table could be global to
avoid duplication across data-profiles.

Then we'd need to either make new hash-table for every buffer with `private-data-profile` or set the `*-path`s of this buffer to a random value. Both options sound fishy to me, although I understand that storing data in `data-profile` can seem bad too. I still think it's easier to work with the instanses/class-stored data than with hash-tables, because CL (and some Nyxt macros) allows you to redefine the classes. In other words, you can do something like:
``` lisp
(defclass foo-data-profile (default-data-profile)
  ((foo-data :accessor foo-data
             :type data-path
             :allocation :class
             :initform (make-instance 'data-path :basename "foo")
             :documentation "Keep FOO there.")))

(replace-class default-data-profile foo-data-profile)
```
and all the new instances of `default-data-profile` will have this data slot.

aartaka:
Changes:
- Add `maybe-data-profile` function that returns either buffer's `data-profile` or `*global-data-profile*`.
- Extract `store` and `restore` methods on `private-data-profile` to the separate method definitions.
- Use dummy buffer in tutorial.
- Have a better dirname for the history-testing directory.
- Fix typos.

source/data-storage.lisp
Ambrevar: The question is, will we ever need to store/restore or expand-data-path for a buffer that's not the current one?

If it happens to be rare, we would use with-current-buffer for these cases.  That would make much of the code much lighter.

About your last point: `with-current-buffer` is a common idiom in Emacs (and also when dealing with monads).  The resulting code woule look like this for the caller:

```lisp
(with-current-buffer my-buffer
  (expand-data-path PROFILE PATH))
```

source/renderer-gtk.lisp
Ambrevar: Instead of an anaphoric class, you can use `initialize-instance :after` to accomplish the same effect I believe.

Ambrevar:
> - Add `maybe-data-profile` function that returns either buffer's `data-profile` or `*global-data-profile*`.

I suggest `current-data-profile` which is consistent with our
`current-buffer`, `current-window`, etc. functions.

tests/test-global-history.lisp
Ambrevar: No, that's not the one.  There is a Common Lisp version of POSIX' mktemp.
Osicat has mktemp, mkstemp, mkdtemp.

Ambrevar:
> ``` lisp
> ;; TODO: Ideally, we need to know the initform that user could've set for
> ;; `data-profile' slot of `buffer'. Use `mopu:slot-properties'?
> (log:info "Using data profile ~s." (name (make-instance
>                                            (or (find-data-profile
>                                                  (getf *options* :data-profile))
>                                                'default-data-profile))))
> ```

You could do as for the manual: instantiate a dummy buffer and get the
profile from there.

But really, just for a log:info, it's not worth it.  Also now it makes
little sense semantically because previously, the data-profile was
unique per run.  Now we can have a data-profile per buffer.  I suggest
moving the log message to the buffer creation.

Problem solved! :)

>> I think we need a global hash-table that maps an expanded path to the corresponding data. Thoughts? 
>> On a similar topic: I don't think using slots in data-profile is a good idea. What about user-defined or third-party-defined data that want to use the default profile? Instead, I think we need a hash-table which stores the data of the expanded path. As hinted above, this hash-table could be global to avoid duplication across data-profiles.
>
> This sounds good, although I feel there can be problems, especially when the buffer with `private-data-profile` and default paths will try to access and store data somewhere it shouldn't.

If the hash-table key is the path, then the hash-table value is the
content.  Thus if different profiles work on the same key, they are
necessarily dealing with the same file and thus the same content.

There won't be more contention between different profiles than
between different threads storing/restoring the same data path.  We just
need to be careful with locks.

>> Instead, we can add a helper function to retrieve the profile from a potential buffer (may be nil). 
>
> Is it going to be the same as this `(if buffer (data-profile buffer) *global-data-profile*)` idiom?

Yes.

>> All that said, do we have to pass the buffer all the time? It's a bit cumbersome. Can we use the `with-current-buffer` macro instead? 
>
> I don't yet understand how `with-current-buffer` can help us here. Can you expand on that, please?

See my example above.  And see Emacs' function of the same name.  Let me
know if you want more details.

> But, again, where should we store locks? Global hash-table from expanded filenames to locks?

The locks could be stored in the same hash-table mentioned above that
would hold the data.  It's a "global" hash-table indeed, but it could be
stored in BROWSER.

> `nyxt/auto-mode:(re)store-last-active-modes` seem to collide with `(re)store` already.

You mean that the terms are already used for a different purpose in auto-mode?

>> - You've removed the (setf DATA-STORING-SLOTS) method. The drawback is that now we need to be careful to store / restore at the right time when getting/setting the data-storing slots. Maybe we can restore the setf-method by using `with-current-buffer`?
>
> Hmmmm. I'm not sure implicit storage on `setf` was a good idea in this context (it created several leaky-abstraction snippets like `(setf (history-data *browser*) (history-data *browser*))`), so I doubt it should be back. If we'll implement the `with-parallel-data-access` or incorporate it into `with-data-file`, we won't miss these implicitly stored slots :)

Actually, there is a better reason why the setf-method won't do here:
it's because when the slot is a composite structure like a hash-table,
setf-ing the hash-table _key_ won't trigger the setf-method for storing,
because the slot is unchanged.

So I agree that `with-data-access` is the right way to go here!

Ambrevar:

>> Instead, I think we need a hash-table which stores the data of the
> expanded path.  As hinted above, this hash-table could be global to
> avoid duplication across data-profiles.
>
> Then we'd need to either make new hash-table for every buffer with
> `private-data-profile` or set the `*-path`s of this buffer to a random
> value.

I don't understand.  `private-data-profile` don't persist anything, so
they don't need to rely on the global hash-table for the data.  Misunderstanding?

Maybe because we talked twice of hash-tables, in different context.

The solution I suggest is to _not store_ the hash table of data in
data-profile, but to move it to, say, a browser slot.

aartaka:
> I don't understand. `private-data-profile` don't persist anything, so they don't need to rely on the global hash-table for the data.

True, `private-data-profile` doesn't need to persist anything, but it can be useful to keep browsing history or `auto-mode-rules` for the time of `private-data-profile`-owning buffer existence and clear it afterwards. Currently, `private-data-profile` can contain some data in its data-slots and even use it while browsing, even though it never persists it. 

That's why having only one hash-table for data management seems unsatisfactory for me -- we will either put `private-data-profile` data in one pot with everything else of we'll need to create separate hash-tables for every private buffer.

Ambrevar:

> True, `private-data-profile` doesn't need to persist anything, but it can be useful to keep browsing history or `auto-mode-rules` for the time of `private-data-profile`-owning buffer existence and clear it afterwards. Currently, `private-data-profile` can contain some data in its data-slots and even use it while browsing, even though it never persists it. 
>
> That's why having only one hash-table for data management seems unsatisfactory for me -- we will either put `private-data-profile` data in one pot with everything else of we'll need to create separate hash-tables for every private buffer.

Good point.  What about this then: index the data hash-table by
(DATA-PROFILE DATA-PATH)?  This is awkwardly reminiscent of the method
signature.

aartaka:
Another solution: we can keep no data for `private-data-profile` _at all_ and introduce third `data-profile` class that'll have well-expanding `expand-data-path` (`private-data-profile`'s one always returns nil), while having disabled `(re)store` (as in `private-data-profile`). 

Given the buffer-isolated paths (and, consequently, data), this third profile will give us best of both `default-data-profile` and `private-data-profile` -- an ability to keep some data for session duration without persistence!

However, this could better be be a part of `isolate-*` modes' patchset or proper incognito mode one, because it isn't necessarily related to the per-buffer-data-storage patchset.

Ambrevar:
I don't understand your last suggestion.  What's the usage difference between private-data-profile and the third profile?  How is data-path expansion useful if you don't store anything?  

Are you suggesting we have a profile that does not even store session data (e.g. history for the current buffer only)?  Do you think this can be desirable for some users?

aartaka:
> Are you suggesting we have a profile that does not even store session data (e.g. history for the current buffer only)? Do you think this can be desirable for some users?

I doubt it can be desirable. It's not the way then. 

How about generating unique paths for all the `*-path` slots of `private-data-profile`-bearing buffer on its initialization? This way we can isolate and keep the data without the need to persist it. This will demand making `expand-data-path` on `private-data-profile` same as the one for `default-data-profile` to enable `gethash`-ing of expanded unique paths. It makes `expand-data-path` a one-method-generic. 

Do we even need to have `expand-data-path` then, when most of data precautions now are in `store`/`restore` and `get-data` (the new method on `data-path` that'd `gethash` the data from the global data storage (the prospective name is `user-data`, but it's not good enough))?

Ambrevar:

>> Are you suggesting we have a profile that does not even store session data (e.g. history for the current buffer only)? Do you think this can be desirable for some users?
>
> I doubt it can be desirable. It's not the way then. 
>
> How about generating unique paths for all the `*-path` slots of
> `private-data-profile`-bearing buffer on its initialization? This way
> we can isolate and keep the data without the need to persist it.

So this means that all private buffers would share history, bookmarks,
etc., right?

This could work.

> This will demand making `expand-data-path` on `private-data-profile` same as the one for `default-data-profile` to enable `gethash`-ing of expanded unique paths. It makes `expand-data-path` a one-method-generic. 

I don't understand this part.  Can you give an example?

> Do we even need to have `expand-data-path` then, when most of data precautions now are in `store`/`restore` and `get-data` (the new method on `data-path` that'd `gethash` the data from the global data storage (the prospective name is `user-data`, but it's not good enough))?

Callers may want to know where data is stored for whatever reason.
Think of expand-data-path as and overpowered version of Emacs' `expand-path'.

Also `store' and `restore' themselves need to use `expand-data-path'! ;)

aartaka:
> So this means that all private buffers would share history, bookmarks, etc., right? This could work.

No, I meant that every private buffer will have its own unique set of paths that it will expand an use as keys for data-access in the global storage hash-table.

> This will demand making `expand-data-path` on `private-data-profile` same as the one for `default-data-profile` to enable `gethash`-ing of expanded unique paths. It makes `expand-data-path` a one-method-generic.
> I don't understand this part. Can you give an example?

Sure. I'll try to cover the previous point too, because I could have been unclear about it. Imagine two private buffers (for now it means only that they have `private-data-profile` instance as their `data-profile`) -- A and B. A gets created with it's unique `history-path` (and any other path is unique for every `private-data-profile` too!) -- "buffer-A-history.lisp", B has it's own "buffer-B-history.lisp". None of the buffers will store anything in these paths, because they are private, but these paths will be used to index the data of these buffers.

Now, say, A navigates to "example.org", then "foo.bar", while B goes to "example.org", but then deletes the last entry. Then, if we 
``` lisp
;; A is the %buffer, expand-path results in "buffer-A-history.lisp"
(gethash (expand-path (history-path A))
   (user-data *browser*))
=> ... hash-table containing "example.org" and "foo.bar" ...

;; B is the %buffer, expand-path results in "buffer-B-history.lisp"
(gethash (expand-path (history-path B)) 
   (user-data *browser*))
=> ... hash-table containing "example.org" only ...
```
After entry deletion:
``` lisp
;; A is the %buffer, expand-path results in "buffer-A-history.lisp"
(gethash (expand-path (history-path A))
   (user-data *browser*))
=> ... hash-table containing "example.org" and "foo.bar" ...

;; B is the %buffer, expand-path results in "buffer-B-history.lisp"
(gethash (expand-path (history-path B)) 
   (user-data *browser*))
=> ... nil
```

> Callers may want to know where data is stored for whatever reason. Think of expand-data-path as and overpowered version of Emacs' `expand-path'. Also `store' and `restore' themselves need to use `expand-data-path'! ;)

I'll need some time to look up the Emacs' `expand-path` and how we can use `expand-data-path`, so no answer yet :)

Ambrevar:
It seems that you are being inconsistent: you are saying that `expand-data-path` should expand on the private profile just like on the default profile, but  actually we need to expand private data in a buffer-unique fashion as in your example.

For instance:

```lisp
(defmethod expand-data-path ((private-data-profile ...) (data-path ...))
  ;; Generate a unique string
  (str:concat  (basename data-path) (id buffer)))
```

I think it would work.

Random string generation is a bit brittle though.   Besides, what do we do with the stale data in (user-data *browser*) once a buffer is deleted? Maybe there is simpler and safer.

What about adding a method that'd return the storage location of the data-path depending on the profile?
This would allow us to store the user data in the profile, thus removing the need for complicated indexing.
Another benefit is that no stale data would be left over.

Example
```lisp
(defmethod expand-user-data ((public-data-profile ...) (data-path ...))
  (gethash (user-data *browser) (expand-data-path data-path)))

(defmethod expand-user-data ((private-data-profile ...) (data-path ...))
  (gethash (user-data private-data-profile) (expand-data-path data-path)))
```
Feel free to pick a name better than `expand-user-data`.

Thoughts?

aartaka:
> It seems that you are being inconsistent: you are saying that `expand-data-path` should expand on the private profile just like on the default profile, but actually we need to expand private data in a buffer-unique fashion as in your example.

I guess I just haven't thought that we can generate these paths in `expand-data-path` and was locked on the idea that we need to compute things as buffer initialization. Your idea sounds totally intuitive though!

> What about adding a method that'd return the storage location of the data-path depending on the profile?
> This would allow us to store the user data in the profile, thus removing the need for complicated indexing.
> Another benefit is that no stale data would be left over.

I thought about making the data-returning methods on profiles, but couldn't wrap my head around the way to do so. Thanks for hinting at this way!

aartaka:
What's new there:
- `browser` slots:
  - `user-data` (I doubt meaningfulness of the name) -- a place to store everything previously stored in `history-data`, `session-data` etc.
  - `data-locks` -- the locks corresponding to the given paths, to make sure no race condition will happen on read/write.
  - `download-watcher` -- not a new one, but it's back in `browser`, because it wasn't really useful to keep in browser, given its global character.
- `current-data-profile` -- the new name for `maybe-data-profile`.
- `expand-path` needs buffer argument no more -- it infers the current profile with `current-data-profile`. Same for other functions that initial version of this patch demanded buffer argument from -- it's not necessary in most of them now.
- `get-data` and `get-profile-data` -- data-related counterparts of `expand-path` and `expand-data-path`.
- `store` and `restore` have swapped the order of arguments to make sure that profile-specializing methods will be favored over path-specializing ones.
- `save-last-active-modes`, `reapply-last-active-modes` are new names for `store-last-active-modes` and `restore-last-active-modes` that collided with `store`/`restore`.
- `with-data-file` and `expand-path` are now in data-storage.lisp, where (I thought) they belong.
- `get-data-lock`, `with-data-lock`, `with-data-access` that build on each other to enable the path-locking in an intuitive way.
- `private-data-profile` isolates the data in it's instances!
- The buffers with `private-data-profile` are now behaving somewhat differently on creation/deletion. This could've been a part of incognito-mode patch, but here it is.
- `data-manager-path` is now generated on `gtk-buffer` initialization.
- `data-profile`-naming log message is removed from `start-browser`, but no new place for it was found yet.
- Minor code movements.


Caveats:
- Still no per-buffer downloads' lists.
- The addition of different behavior for the `private-data-profile`-owning buffers looks awkward. I guess introduction of `ephemeral-buffer` and specializing creation/deletion actions on it will make it clearer.
- `%buffer` appears in data-storage.lisp, although it's defined much later, in browser.lisp. Should we move it?
- Merge conflict between two fixes of the same bug xD

Ambrevar:
- > with-data-file and expand-path are now in data-storage.lisp, where (I thought) they belong.

  The reason I didn't put them there is that I though we would eventually make an independent library out of `data-storage` and thus we can't depend on a Nyxt-specific feature like `%buffer`.  If you think it looks better this way for now, we can always move it back later.  Should you keep it in `data-storage.lisp`, please separate the Nyxt-specific code from the generic code.

- From just the name `get-data` it's not clear how it relates to `data-path`.  `data-path-data` is more logical but sounds odd to a human :)  Maybe the term `data-path` was poorly chosen.  What about `user-path` and `user-profile` instead?  Could `user-profile` be confused for something else?  Anyways, let's not fix this issue in this change set and keep it for later.

- On data locks: I think separating the locks from the data is a dangerous practice, it involves too much manual setup.  What about this instead: create a `user-data` class which contains the data as well as a mutex, both in slots without accessors.  Then you can make a custom `user-data` accessor which retrieves the data after unlocking the mutex.  You should be able to get rid of `get-data-lock` and `with-data-lock` after that. Another benefit is that it saves a slot in `browser`.

I think we are looking good!
As soon as the class-star patch is merged and my last few suggestions considered, we will be good to merge!  Hooray!

source/data-storage.lisp
```diff
@@ -163,6 +191,98 @@ function result as a boolean in conditions."
...
+           (m (uiop:native-namestring m))))))
+
+(export-always 'get-profile-data)
+(defgeneric get-profile-data (profile path)
+  (:documentation "Access the browsing-related data depending on the `data-profile'."))
+
+(defmethod get-profile-data ((profile private-data-profile) (path data-path))
+  "Look up the buffer-local data in case of `private-data-profile'."
+  (let ((expanded-path (expand-path path)))
+    (when expanded-path
```
Ambrevar, line 238: Nit: You could use sera:and-let* to save this when.

source/data-storage.lisp
```diff
@@ -163,6 +191,98 @@ function result as a boolean in conditions."
...
+      (gethash expanded-path (user-data profile)))))
+
+(defmethod (setf get-profile-data) (value (profile private-data-profile) (path data-path))
+  (let ((expanded-path (expand-path path)))
+    (when expanded-path
```
Ambrevar, line 243: Same.

source/data-storage.lisp
```diff
@@ -163,6 +191,98 @@ function result as a boolean in conditions."
...
+
+(export-always 'with-data-access)
+(defmacro with-data-access ((path-name data-path) &body body)
+  "Lock the data for the BODY to avoid race conditions.
+Bind the PATH-NAME to the value of DATA-PATH for to reuse it."
+  `(let ((,path-name ,data-path))
+     (with-data-lock ,path-name
+       (restore (current-data-profile) ,path-name)
+       ;; We need to return the possibly meaningful result of BODY.
+       ;; Isn't `unwind-protect' an overkill there? `prog1'-`progn' is enough.
```
Ambrevar, line 281: The questions is, do we want to store when there is an error?  This is what `unwind-protect` is for.

Let's consider a body along these lines:

```
Set A in data.
Run operation that results in an error.
Set B in data.
```

If we don't use `unwind-protect`, the in-memory data would contain A, but not the one on disk.
If we use `unwind-protect`, both the in-memory data and the data on disk have A.

So yes, I think you need `unwind-protect`.

source/data-storage.lisp
```diff
@@ -93,6 +59,75 @@ This can be used to set the path from command line.  See
...
+(export-always '*global-data-profile*)
+(defvar *global-data-profile* (make-instance 'default-data-profile)
+  "The profile to use in the absence of buffers and on browser-less variables.")
+
+(export-always 'current-data-profile)
+(defun current-data-profile ()
+  "If `%buffer' is non-nil, return its data-profile.
+Return `*global-data-profile*' otherwise."
+  ;; TODO: %BUFFER is not defined yet. Move %BUFFER there?
+  (or (and %buffer (data-profile %buffer))
```
Ambrevar, line 108: You should use `(current-buffer)` instead of `%buffer`.

Ambrevar:
There is a pitfall with the (current-buffer) technique: when a buffer triggers `store` / `restore` in the background, the `(current-data-profile)` will return the data-profile of the buffer in the foreground, which will return the wrong data-path.

Current I think the only such data-path is `history-path` which is modified in `history-add`, called in `nyxt:on-signal-notify-uri` in `web-mode.lisp`.  I suggest you wrap this call with `(with-current-buffer (buffer mode) ...)`.

source/data-storage.lisp
aartaka: The problem with `(current-buffer)` is that sometimes it returns a leftover buffer from the previous session. I feel that it can violate user's privacy if the permissive data-profile from the previous session will be used instead of the new buffer's one. Can it actually be the case?

source/browser.lisp
```diff
@@ -443,7 +329,7 @@ current buffer."
                                  :directory download-dir
                                  :cookies cookies
                                  :proxy proxy-address))
-                 (push download (downloads *browser*))
+                 (push download (get-data path))
```
Ambrevar, line 332: If `path` is only used once, maybe no need to let-bind it.  Besides, `(get-data (download-path (current-buffer)))` is a bit more explicit than `(get-data path)`.

source/browser.lisp
```diff
@@ -775,17 +661,20 @@ sometimes yields the wrong reasult."
...
-   (setf *error-output*
-         (open (expand-path (error-output-path browser))
-               :direction :output
-               :if-does-not-exist :create
-               :if-exists :append))))
+  (let ((buffer (current-buffer)))
+    (values
+     (when (expand-path (standard-output-path buffer))
+       (setf *standard-output*
+             (open (expand-path (standard-output-path buffer))
```
Ambrevar, line 668: Use `sera:andlet-*` to bind the expanded path.  This is a pitfall: calling expand-path twice may return different results, e.g. if the `expand-data-path` method depends on the time or any external changing factor.

source/buffer.lisp
```diff
@@ -510,6 +576,22 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
       (setf (url buffer) (url dead-buffer)))
     (when (expand-path (cookies-path buffer))
       (ensure-parent-exists (expand-path (cookies-path buffer))))
+    ;; Creating isolated paths for private-data-profile-owning buffers
+    (when (typep (data-profile buffer) 'private-data-profile)
+      ;; TODO: rebind standard-output-path and error-output-path too?
+      (labels ((buffer-local-pathname (name id)
+             (let* ((pathname (pathname name))
```
Ambrevar, line 583: Indentation.

source/buffer.lisp
```diff
@@ -510,6 +576,22 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
       (setf (url buffer) (url dead-buffer)))
     (when (expand-path (cookies-path buffer))
       (ensure-parent-exists (expand-path (cookies-path buffer))))
+    ;; Creating isolated paths for private-data-profile-owning buffers
```
Ambrevar, line 579: Typo: "Create" and missing period.

source/buffer.lisp
```diff
@@ -531,6 +613,13 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
 (declaim (ftype (function (buffer)) buffer-delete))
 (defun buffer-delete (buffer)
   (hooks:run-hook (buffer-delete-hook buffer) buffer)
+  (when (typep (data-profile buffer) 'private-data-profile)
+    (dolist (path (list (session-path buffer)
+                        (history-path buffer)
+                        (bookmarks-path buffer)
+                        (auto-mode-rules-path buffer)
+                        (cookies-path buffer)))
+      (uiop:delete-file-if-exists (expand-path path))))
```
Ambrevar, line 622: Wait, why are these files existing in the first place?  'private-data-profile should not persist anything, right?

source/buffer.lisp
```diff
@@ -510,6 +576,22 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
...
+               (if (null type)
+                   (format nil "~a-~a" pathname id)
+                   (let* ((printed-pathname (object-string pathname))
+                          (new-pathname (subseq printed-pathname
+                                                0 (- (length printed-pathname) (length type) 1))))
+                     (format nil "~a.~a" (buffer-local-pathname new-pathname id) type))))))
+        (dolist (path (list (session-path buffer) (history-path buffer) (bookmarks-path buffer)
+                            (auto-mode-rules-path buffer) (cookies-path buffer)))
+          (setf (basename path)
+                (buffer-local-pathname (basename path) (id buffer))))))
```
Ambrevar, line 594: We can not hard-code the list of data-path.  As soon as the user uses their own data-path or some third-party, this will fail to do the job.

Instead, we should define the a method on (DATA-PROFILE DATA-PATH) that returns the user data.

source/buffer.lisp
aartaka: Right :)

source/data-storage.lisp
Ambrevar: Indeed, but the %buffer technique suffers from the converse problem since it will return the global data profile for all buffers unless `with-current-buffer` is used.

In all cases, wrapping the call in `with-current-buffer` is the only sure way to get the right data profile.

source/browser.lisp
aartaka: It's also used in the `download-dir` binding above, but I agree that being explicit is more important here than saving two calls to `download-path`.

Ambrevar:
There might be a misunderstanding about what `private-data-profile` ought to do: it should not persist any file!  It should only keep the data in memory, when that makes sense.

This is why we need to methods: one method that tells where the in-memory data is kept, and one method that tells where the disk data is going to be stored.

Does that make sense?

aartaka:
It does! I don't remember why I added file deletion on the obviously non-existent-files, but it looks strange to me now xD

source/browser.lisp
Ambrevar: Oh, then leave it, because file expansion should happen only once in a single context.  Expanding multiple times may yield different results if the expansion depends on external factors such as time.

aartaka:
> * On data locks: I think separating the locks from the data is a dangerous practice, it involves too much manual setup.  What about this instead: create a `user-data` class which contains the data as well as a mutex, both in slots without accessors.  Then you can make a custom `user-data` accessor which retrieves the data after unlocking the mutex.  You should be able to get rid of `get-data-lock` and `with-data-lock` after that. Another benefit is that it saves a slot in `browser`.

When should we lock the mutex in this situation? How do we incorporate it into `with-data-access`? Isn't it simpler to just make accessors and alter the `user-data` in `get-profile-data` (with `uiop:ensure-gethash`, for example) and `get-data`? I'll push a version of data-blocking based on the `ensure-gethash` in an hour.

aartaka:
Okay, I've added a `user-data` class and altered the code according to it. 
- `get-profile-data` -> `get-user-data`. It returns a matching instance of `user-data`, creating it if necessary.
- No more `data-locks`, locks are part of `user-data`. No more `get-data-lock` and `with-data-locks` either.
Other changes:
- `sera:and-let*` all over the place :) It's really convenient, thanks!
- `current-data-profile` uses `current-buffer`.
- Non-existent `private-data-profile` files don't get deleted anymore.

One problematic thing: we can control the `store`/`restore` for bookmarks, auto-mode rules, session, but we cannot control persistence for WebKit-provided abstractions, like `cookie-manager` and `data-manager`. To disable storage to the `data-manager-path`/`cookies-path` we need to redefine them to be empty or test whether the `current-data-profile` is the `private-...` one (the latter is used for now, but it's not really future-proof). Should we get back to the nil-returning `expand-data-path` on `data-manager-path`/`cookies-path` and `private-data-profile`?

Another problematic thing: we cannot nest two `with-data-access` because they'd try lock the same resource and err terribly. Should we do something about it?

Ambrevar:
You lock then mutex in the accessor.  Untested code:

(defmethod get-data ((user-data user-data))
  (with-held-lock (slot-value user-data 'mutex)
    (slot-value user-data 'data))

Same for the setf method.

I was wrong about the removal of get-lock and with-data-lock.
However, with-data-access would bind the _data_ instead of the
path-name.  This is also more convenient.

P.S.:
You might also want to not squash the commits while working on them.  It
makes it easier to compare the different implementations for the sake of discussion.

source/data-storage.lisp
```diff
@@ -163,6 +180,85 @@ function result as a boolean in conditions."
...
+(defmethod get-data ((path data-path))
+  (data (get-user-data (current-data-profile) path)))
+
+(defmethod (setf get-data) (value (path data-path))
+  (setf (data (get-user-data (current-data-profile) path)) value))
+
+(export-always 'with-data-access)
+(defmacro with-data-access ((path-name data-path) &body body)
+  "Lock the data for the BODY to avoid race conditions.
+Bind the PATH-NAME to the value of DATA-PATH for to reuse it."
```
Ambrevar, line 252: Typo: "for to"

aartaka: Fixed, thanks!

aartaka:
> You lock then mutex in the accessor. 
>
>``` lisp
>(defmethod get-data ((user-data user-data))
>   (with-held-lock (slot-value user-data 'mutex)
>     (slot-value user-data 'data))
> ```

But wait, all this business with locks is to make data-to-disk storage safer. We don't need to lock the data itself, because it's usually changed inside the `with-data-access` block and is guaranteed to be changed in one place at a time. If it's changed outside the `with-data-access` block, it will be overwritten by the next `restore`. 

In a way, we lock both the on-disk and in-browser data, when in `with-data-access`. Do we need to separate these, for example, in the `user-data` accessor? I can't think of a case where this could be both necessary and safe. I'll be glad to know about the possible use-cases, if you have some in mind!

> I was wrong about the removal of get-lock and with-data-lock. 

Why so? `with-data-lock` could've been used in some restricted domain, but `get-lock` was just an implementation detail and it became unnecessary once the implementation changed.

> However, with-data-access would bind the _data_ instead of the path-name. This is also more convenient.

You're right, it sounds much more convenient. I'll look into that!

aartaka:
`with-data-access` now binds the data stored in it's second argument, to the name of its first argument. I've removed the brackets around the arguments of `with-data-access` to make it less `let`-like (it was resembling `let` until now).

What looks fishy to me is the `(setf (get-data ,path-name) ,data-var)` expression in the end of `with-data-access`. Can it bring any trouble?

source/data-storage.lisp
```diff
@@ -163,6 +180,87 @@ function result as a boolean in conditions."
...
+
+(export-always 'get-user-data)
+(defgeneric get-user-data (profile path)
+  (:documentation "Access the browsing-related data depending on the `data-profile'."))
+
+(defmethod get-user-data ((profile private-data-profile) (path data-path))
+  "Look up the buffer-local data in case of `private-data-profile'."
+  (sera:and-let* ((expanded-path (expand-path path)))
+    (alex:ensure-gethash expanded-path (user-data profile)
+                         (make-instance 'user-data))))
```
Ambrevar, line 239: How is this "buffer-local"?  It seems to be the same as with the default profile.

Ambrevar:

> One problematic thing: we can control the `store`/`restore` for bookmarks, auto-mode rules, session, but we cannot control persistence for WebKit-provided abstractions, like `cookie-manager` and `data-manager`. To disable storage to the `data-manager-path`/`cookies-path` we need to redefine them to be empty or test whether the `current-data-profile` is the `private-...` one (the latter is used for now, but it's not really future-proof). Should we get back to the nil-returning `expand-data-path` on `data-manager-path`/`cookies-path` and `private-data-profile`?

There may be a misunderstanding.

Cookies and the data-manager don't have any Lisp in-browser data.  The
get-data methods should return nil for them.

Each data-path in-browser data is different and needs to be specialized accordingly.

`expand-path` should return nil for private-data-profile for all data-paths.

Does this answer your question?

> Another problematic thing: we cannot nest two `with-data-access` because they'd try lock the same resource and err terribly. Should we do something about it?

You could use a recursive-lock instead.

Ambrevar:
> What looks fishy to me is the `(setf (get-data ,path-name) ,data-var)` expression in the end of `with-data-access`. Can it bring any trouble?

What does it look fishy in your opinion?

Ambrevar:

>> You lock then mutex in the accessor. 
>>
>>``` lisp
>>(defmethod get-data ((user-data user-data))
>>   (with-held-lock (slot-value user-data 'mutex)
>>     (slot-value user-data 'data))
>> ```

> But wait, all this business with locks is to make data-to-disk storage safer. We don't need to lock the data itself, because it's usually changed inside the `with-data-access` block and is guaranteed to be changed in one place at a time. If it's changed outside the `with-data-access` block, it will be overwritten by the next `restore`. 

Nyxt runs multiple threads and there may be more in the future.  Since
the user-data is stored in *BROWSER*, it's essentially a global and thus
a source of race conditions.

Multiple buffers may change the global history at the same time, see `history-add`.


> In a way, we lock both the on-disk and in-browser data, when in `with-data-access`. Do we need to separate these, for example, in the `user-data` accessor? I can't think of a case where this could be both necessary and safe. I'll be glad to know about the possible use-cases, if you have some in mind!

Separate the locks?  I don't think that's necessary.  To recap, we have
2 kinds of persisted data:

- The web renderer data that's out of our control.  This one needs no
  lock at all, and no user data is kept, we only need expand-path.

- The Lisp data that the user is free to access and modify.  This data
  must be kept in sync between the browser and the file on disk.  This
  synchronization only happens when the data is accessed.  The
  file is never directly edited by the user, only the data is.

  The bridge between user-data and on-disk data are the `store` and
  `restore` methods.

  The bridge between the user and the user-data is the
  `with-data-access` macro.

Am I getting it right?

source/data-storage.lisp
aartaka: No, the one on `default-data-profile` (it's in the browser.lisp) gethashes `(user-data *browser*)`, not `(user-data profile)`. Given that every `private-data-profile` instance has a `user-data` slot with different contents between instances, we can say that it's buffer-local (because data-profile belongs to buffer).

However, it can sound ambiguous, because it's not directly local. "profile-local" maybe?

aartaka:
> > What looks fishy to me is the `(setf (get-data ,path-name) ,data-var)` expression in the end of `with-data-access`. Can it bring any trouble?

> What does it look fishy in your opinion?

It's located in the cleanup branch of the `unwind-protect` and I'm not sure we should `setf` data to a new value in any case, including error case. Erring computations can corrupt the data, and assigning it blindly can be harmful. 

However, we can't predict anything. I may be over-engineering here :)

Ambrevar:
> It's located in the cleanup branch of the `unwind-protect` and I'm not sure we should `setf` data to a new value in any case, including error case. Erring computations can corrupt the data, and assigning it blindly can be harmful. 
>
> However, we can't predict anything. I may be over-engineering here :)

The `setf` is only useful for list-like data whose pointer needs to
follow the head.

In case of error, unwind-protect or not, the data-structure will remain
modified nonetheless.

If you want to avoid that,  we would need a functional data
structure.  We already use FSet in keymap, so we could switch our
bookmarks, history, etc. to FSet data structures.
This is out of the scope of this patch however.

That said, do we want to really want to discard storage on errors?
What if we performed a lengthy operation over many entries, then the
last one would fail, would we really want to revert all the work that
was done?  Maybe I'm nitpicking here with a case that's never going to
happen.

However, I can see why _not reverting_ could corrupt the data.
Say we save a (URL DETAILS) entry.  We first set URL in the user-data,
then the DETAILS, but an error happens then.  Because of the
unwind-protect, we may end up persisting (URL nil) which may not be
legal.

aartaka:
Fair points. Let's stick to the current version with modifying it anyway and move to the functional data types if the need arises.

aartaka:
> Nyxt runs multiple threads and there may be more in the future. Since the user-data is stored in \*BROWSER\*, it's essentially a global and thus a source of race conditions. Multiple buffers may change the global history at the same time, see `history-add`.

`history-add` uses `with-data-access` already and it binds data inside `bt:with(-recursive)-lock-held`, so it's safe from race conditions. I guess same can be said about any function that uses `with-data-access` -- it will get the data in the newest state, after all the possible modifications other threads could do. 

However, I may be biased about current implementation and wouldn't see its flaws. If there is a possibility of race conditions in the current version -- I'll be grateful if you point at these!

Ambrevar:
>> Nyxt runs multiple threads and there may be more in the future. Since the user-data is stored in \*BROWSER\*, it's essentially a global and thus a source of race conditions. Multiple buffers may change the global history at the same time, see `history-add`.
>
> `history-add` uses `with-data-access` already and it binds data inside
> `bt:with(-recursive)-lock-held`, so it's safe from race conditions.

I realize I didn't answer your question properly.  Your question was

> But wait, all this business with locks is to make data-to-disk storage
> safer. We don't need to lock the data itself, because it's usually
> changed inside the `with-data-access` block and is guaranteed to be
> changed in one place at a time.  If it's changed outside the
> `with-data-access` block, it will be overwritten by the next
> `restore`.

Your last sentence is making a good point.  Our problem is that we can't
know when the data is modified so we are forced to always store, even if
in really we just read some data.

This can be needlessly costly.

It looks like we really need a functional data structure, it seems to be
the only way to make the distinction between "reads" and "writes".

One we have that, then we won't need with-data-access if all we want is
to read data, or a one-liner-modification like

```
(push <HISTORY-ENTRY> <HISTORY-DATA>)
```

If we want to allow the above syntax, we will need to use the mutex in
the accessor.

jmercouris:
@aartaka 

Artyom, I just merged the new class model onto master, so you should be able to merge into your branch now.

aartaka:
Thanks! I've brought this branch up to date with master, so it should be safe to merge.

One thing that I've noticed working improperly though (it bugs on master too, so it's not relater to this patch) is that `auto-mode` stopped recognising modes because of introduction of `user-*-mode` classes (especially noticeable when `*prompt-on-mode-toggle*` is enabled). I believe we need to start an Issue about it :)

jmercouris:
Oh, that is a bit of an issue. I don't imagine it will be too hard to figure out!

Ambrevar:
I know this issue, I thought I had fixed it, I'll take a second look.

Ambrevar:
@aartaka:

Does origin/master 19675f54481133fdcf4238ef2d8d785fa16fdf93
fix it for you?

If not, can you provide a recipe?

I've also noticed that modes don't get reset on unmatched sites.  There
may be another bug.

Ambrevar:
> I've also noticed that modes don't get reset on unmatched sites.  There
> may be another bug.

Fixed with 8735b757b1029451047669e8702fa9b578213034.

Ambrevar:
I still don't like the name `get-data`.  Any idea for a better name?
@jmercouris?

source/browser.lisp
```diff
@@ -132,93 +131,21 @@ FALLBACK-URL is empty, SEARCH-URL is used on an empty search.
...
                      :export nil
                      :documentation "List of downloads.")
-   (download-path :accessor download-path
-                  :type data-path
-                  :initform (make-instance 'download-data-path
-                                           :dirname (xdg-download-dir))
-                  :documentation "Path of directory where downloads will be
-stored.  Nil means use system default.")
+   (key-stack :accessor key-stack :initform '()
+              :documentation "A stack that keeps track of the key chords a user has pressed.")
```
Ambrevar, line 139: Mistake in rebase here: you can use the define-class syntax for key-stack.

source/browser.lisp
```diff
@@ -279,6 +206,11 @@ editor executable."))
 
 (define-user-class browser)
 
+(defmethod get-user-data ((profile default-data-profile) (path data-path))
+  (sera:and-let* ((expanded-path (expand-path path)))
+    (alex:ensure-gethash expanded-path (user-data *browser*)
+                         (make-instance 'user-data))))
```
Ambrevar, line 212: This looks wrong: The default value of `ensure-gethash` is when the key is not found, not when the hash-table does not exist.

jmercouris:
I have no better ideas :-\

On the bright side, we can always rename later!

source/browser.lisp
aartaka: It's a naming problem too: `user-data` is the `browser` slot, while it's also the class that is used to store data and locks. Instances of `user-data` are stored as values of `user-data` of browser. When there's no such instance, we need to create one (with default arg of `ensure-gethash`), because we usually need to have locks for any requested path.

So, the code is doing the right thing, but the names are really misleading.

source/data-storage.lisp
```diff
@@ -131,20 +174,16 @@ Return NIL when path must not be used.  This makes it possible to use the
...
+  nil)
+
+(defmethod expand-data-path ((profile private-data-profile) (path standard-output-data-path))
+  "We shouldn't store `*standard-output*' for `private-data-profile'."
+  nil)
 
-(defmethod expand-data-path ((profile (eql +private-data-profile+)) (path data-path))
-  "Don't persist anything in private mode."
+(defmethod expand-data-path ((profile private-data-profile) (path error-output-data-path))
+  "We shouldn't store `*error-output*' for `private-data-profile'."
```
Ambrevar, line 186: All private data path should expand to `nil`, shouldn't they?
If so, the above 3 methods can be summed up into 1 by specializing against `data-path`.

aartaka: No, they shouldn't. We:
- Redefine `store`/`restore` for the Nyxt-specific data, like bookmarks, session, history, if we don't want to store these. We don't redefine `expand-default-path` here, because it's used for data indexing in `user-data`.
- Redefine `expand-data-path` to return nil on path to the external data that we don't control, like cookies or `data-manager` data, so that these paths (and related properties) won't even be initialized in the FFI code.

source/browser.lisp
Ambrevar: Indeed!  What about `data-cache` as the browser slot?

Ambrevar: Or `user-data-cache` if we want to clarify the relation with `user-data`.

source/data-storage.lisp
Ambrevar: Got it, my bad!

Ambrevar:
@aartaka Is it reading for merging?

aartaka:
Yes, it is.

Ambrevar:
Rebased and merged as 88d0c5e3757f25566d0a2fe52db006402ae33138.

Thanks one more time for this tremendous effort!

jmercouris:
Thank you Artyom, well done!

PR #913, opened 2020-08-27
command-commands.lisp: execute-extended-command

jmercouris:
ability to execute an extended command with arguments :-)

How do we evaluate the arguments? Do we process an input of NIL as "NIL" or do we treat it as 'NIL?

### DIFF
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -51,10 +51,11 @@
                               (make-minibuffer
                                :input-prompt ,string
                                :input-buffer ,default-value))))))
-             (let ((bindings (mapcar #'make-binding arguments))
-                   (bound-values (mapcar #'first arguments)))
+             (let* ((bindings (mapcar #'make-binding arguments))
+                    (bound-value1 (mapcar #'first arguments))
+                    (bound-value2 (mapcar (lambda (i) `(read-from-string ,i)) bound-value1)))
                `(with-result* ,bindings
-                  (run ,command ,@bound-values))))))
+                  (run ,command ,@bound-value2))))))
     (with-result (command (read-from-minibuffer
                            (make-minibuffer
                             :input-prompt "Execute extended command"
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -35,8 +35,7 @@
                            (make-minibuffer
                             :input-prompt "Execute command"
                             :suggestion-function (command-suggestion-filter
-                                                  (mapcar (alex:compose #'class-name #'class-of)
-                                                          (modes (current-buffer))))
+                                                  (mapcar #'mode-name (modes (current-buffer))))
                             :hide-suggestion-count-p t)))
       (setf (access-time command) (get-internal-real-time))
       (run command))))
@@ -60,8 +59,7 @@
                            (make-minibuffer
                             :input-prompt "Execute extended command"
                             :suggestion-function (command-suggestion-filter
-                                                  (mapcar (alex:compose #'class-name #'class-of)
-                                                          (modes (current-buffer))))
+                                                  (mapcar #'mode-name (modes (current-buffer))))
                             :hide-suggestion-count-p t)))
       (setf (access-time command) (get-internal-real-time))
       (let ((optional-arguments (nth-value 1 (alex:parse-ordinary-lambda-list 
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -58,7 +58,7 @@
                       (run ,command ,@bound-values))))))
     (with-result (command (read-from-minibuffer
                            (make-minibuffer
-                            :input-prompt "Execute command"
+                            :input-prompt "Execute extended command"
                             :suggestion-function (command-suggestion-filter
                                                   (mapcar (alex:compose #'class-name #'class-of)
                                                           (modes (current-buffer))))
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -44,18 +44,18 @@
 (define-command execute-extended-command ()
   "Execute a command by name, also supply optional parameters."
   (flet ((query-arguments (arguments command)
-               (flet ((make-binding (argument)
-                        (let* ((sym (first argument))
-                               (default-value (format nil "~a" (second argument)))
-                               (string (string-downcase (symbol-name sym))))
-                          `(,sym (read-from-minibuffer
-                                  (make-minibuffer
-                                   :input-prompt ,string
-                                   :input-buffer ,default-value))))))
-                 (let ((bindings (mapcar #'make-binding arguments))
-                       (bound-values (mapcar #'first arguments)))
-                   `(with-result* ,bindings
-                      (run ,command ,@bound-values))))))
+           (flet ((make-binding (argument)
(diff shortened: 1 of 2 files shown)


source/command-commands.lisp
```diff
@@ -41,6 +41,32 @@
       (setf (access-time command) (get-internal-real-time))
       (run command))))
 
+(define-command execute-extended-command ()
+  "Execute a command by name, also supply optional parameters."
+  (flet ((query-arguments (arguments command)
+               (flet ((make-binding (argument)
```
Ambrevar, line 47: Indention issue?

source/command-commands.lisp
```diff
@@ -41,6 +41,32 @@
...
+                        (let* ((sym (first argument))
+                               (default-value (format nil "~a" (second argument)))
```
Ambrevar, line 49: In terms of style, I'm starting to wonder if we should replace all the `(format nil "~a" ...)` with `write-to-string`.  The latter is more readable and has the benefit of being more mapcar-able.

source/command-commands.lisp
```diff
@@ -41,6 +41,32 @@
...
+                                   :input-buffer ,default-value))))))
+                 (let ((bindings (mapcar #'make-binding arguments))
+                       (bound-values (mapcar #'first arguments)))
+                   `(with-result* ,bindings
+                      (run ,command ,@bound-values))))))
+    (with-result (command (read-from-minibuffer
+                           (make-minibuffer
+                            :input-prompt "Execute command"
+                            :suggestion-function (command-suggestion-filter
+                                                  (mapcar (alex:compose #'class-name #'class-of)
```
Ambrevar, line 63: Replace the composition by `#'mode-name`.

source/command-commands.lisp
```diff
@@ -41,6 +41,32 @@
...
+                                  (make-minibuffer
+                                   :input-prompt ,string
+                                   :input-buffer ,default-value))))))
+                 (let ((bindings (mapcar #'make-binding arguments))
+                       (bound-values (mapcar #'first arguments)))
+                   `(with-result* ,bindings
+                      (run ,command ,@bound-values))))))
+    (with-result (command (read-from-minibuffer
+                           (make-minibuffer
+                            :input-prompt "Execute command"
```
Ambrevar, line 61: Maybe "Execut extended command" or better "Execute command with arguments"?

Ambrevar:
I would use a slightly different strategy: don't eval the whole call to (run...) but eval all argument individually.  This would be safer I think.

Ambrevar:
About the `nil`, etc.  I think the answer is simple: use `read` :) 
This means that the user will have to surround strings with `"`.

source/command-commands.lisp
```diff
@@ -41,6 +41,32 @@
...
+                            :suggestion-function (command-suggestion-filter
+                                                  (mapcar (alex:compose #'class-name #'class-of)
+                                                          (modes (current-buffer))))
+                            :hide-suggestion-count-p t)))
+      (setf (access-time command) (get-internal-real-time))
+      (let ((optional-arguments (nth-value 1 (alex:parse-ordinary-lambda-list 
+                                              (swank::arglist (sym command))))))
```
Ambrevar, line 68: We also need the key arguments (nth-value 3 ...).
Many commands use them.

Ambrevar:
Additionally, we could use the function type declaration (at least with SBCL) to infer the argument.
If the type is not found or too complex, then fall back to `read`.  We can modify the minibuffer prompt to let know the user when type is infered or not.

Ambrevar:
You can get the function argument types with
`sb-introspect:function-type`.
See `describe-function` for example.

source/command-commands.lisp
jmercouris: Agreed, I just wanted to do a single step/first proof of concept :-)

source/command-commands.lisp
jmercouris: Shall I also change it in `EXECUTE-COMMAND`?

source/command-commands.lisp
jmercouris: `Execute extended command` should be OK. Eventually users will learn our terminology :-D

source/command-commands.lisp
jmercouris: Are they in fact equivalent? Will the aesthetic control character always produce write-to-string?

jmercouris:
About the argument types, MAYBE. If you can find out how to do it cross implementation, I will be much more in favor!

source/command-commands.lisp
jmercouris: Changed in both locations.

source/command-commands.lisp
Ambrevar: If I get the specs right, the equivalent is princ-to-string.  

```
princ-to-string acts like write-to-string with :escape nil :readably nil. Thus no escape
characters are written. 

```

Example:

```lisp
(write-to-string "foo")
; "\"foo\""
(princ-to-string "foo")
; "foo"
```

So I suggest using `princ-to-string` when we want aesthetic, `write-to-string` when we want Lisp-readability.

Ambrevar:
> About the argument types, MAYBE. If you can find out how to do it cross implementation, I will be much more in favor!

I could not find it for CCL.  In fact, declaiming the type in CCL does
not change the function description (e.g. `(describe #'foo)`).

That said, I'm in favour of adding this option for SBCL.  After all,
- it's the only compiler we officially support for now,
- it would not break CCL support,
- we already have a similar SBCL-specific feature in describe-function.

jmercouris:
Can you please help me figure how to READ the results?

Let's say we have a command X Y Z.
In the code base, X Y and Z will be bound to:

```
(with-result* (x (read-from-minibuffer))
                    (y (read-from-minibuffer))
                    (z read-from-minibuffer))
  (some-body)
```
I then need to READ X Y and Z and then bind the values of (READ X) to X. For some reason or another, I cannot seem to figure this out...

Ambrevar:
`(read-from-string x)` should parse X and return the correspond Lisp
form.

Ex:

```
(read-from-string "17")
; => 17

(read-from-string "(format t \"Hello ~a\" 17)")
; (FORMAT T "Hello ~a" 17)

jmercouris:
My question was more along the lines of if there is a less convoluted way to achieve what I want rather than: a097e563b6e1f09a1449b0e2481dd71265f90535

source/command-commands.lisp
jmercouris: OK! :-)

Ambrevar:
Why do you need to quote `read-from-string` ib the `mapcar`?

jmercouris:
Because we need to take the result and read it from string, the minibuffer only returns string results.

Ambrevar:
I tried to hack your function and much of the complexity seems to stem
from the "asynchronousness" of the minibuffer.

I suggest we first rework the minibuffer then give this a better shot.

jmercouris:
Yes, hence why I am using a macro to generate the forms rather than just iterating through them :-D

jmercouris:
Closing this pull request in favor of a new version using the synchronous minibuffer.

PR #921, opened 2020-08-31
start.lisp: use CLI utilities on macOS instead of osicat

jmercouris:
Currently, osicat has problems with compilation in pkg files, this helps temporarily work around them.

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -21,6 +21,7 @@
                :log4cl
                :lparallel
                :mk-string-metrics
+               #-darwin
                :osicat
                :parenscript
                :quri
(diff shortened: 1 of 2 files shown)


jmercouris:
P.S. I knew there was a reason I kept thinking of chmod! :-D

Ambrevar:
Looks good to me, but I can only test tomorrow!

Ambrevar:
Tested and approved!

Ambrevar:
Ideally the functions would be more generic (e.g. pass the permission type as argument) but since this is only a workaround, let's fix it properly later.

Issue #908, opened 2020-08-25
Synchronous minibuffer


Ambrevar:
As mentioned in #906 and a couple of times before, the current asynchronousness of our minibuffer is problematic:

- it's counter-intuitive;
- it's a source of programming errors (e.g. when trying to apply a procedure to the return value of `read-from-minibuffer`);
- it makes use of the dynamically bound `%callback` variable which is referentially non-transparent, which can lead to subtle bugs;
- chaining minibuffer calls is impractical with just callbacks, which forces us to use a macro like `with-result*`.

On top of that, the `(with-result (foo (read-from-minibuffer (make-minibuffer ...))))` idiom is cumbersome.

Ideally, we would just have

```lisp
(let* ((result1 (read-from-minibuffer ...))
       (result2 (read-from-minibuffer ... <whatever which uses result1>))
  body)
```

When sending javascript to web views we also make use of an asynchronous mechanism which `with-result` can also leverage.  But note that minibuffer does not need Javascript so technically we _don't have to_ resort to callbacks.

If we want the minibuffer to draw on screen, it means that the command which called the minibuffer must run in a separate thread.

So here is what I suggest:

- Run all commands in their own thread (in the `run` method of commands).
- `read-from-minibuffer` starts a new thread and returns the result from a channel (using ChanL).

That's it!  Note that we only have to change only two sites, we don't have to alter all commands to start new threads.

Then the minibuffer result is fetched by reading from the channel.  In terms of concurrency, it seems dead simple.  It also removes the need for `with-result*` and `%callback`.

At the moment we don't really handle command results, mostly because of with-result.  WIth the channel strategy we could actually return a channel (or from a channel, I'm not sure at this point).

Note that threads don't have to be OS threads (which can be expensive).  We could resort to using green threads (e.g. https://github.com/thezerobit/green-threads
) of which we can fire up millions with little performance impact.
That said, there will probably not be more than a few nested minibuffers and a couple of commands running at the same time, so the threading setup is probably not an issue.

jmercouris:
If each new command is invoked in a new thread, how do we chain commands together?

Ambrevar:
With channels.

Say command A starts command B.
A waits on the channel result of B.
The B thread completes and writes the result to the channel (last instruction).
A can now resume, gets the value from the channel and procedes.

Note that our current implementation cannot chain commands if they use callbacks (e.g. they contain a with-result).

jmercouris:
[...]

PR #928, opened 2020-09-04
Synchronous minibuffer: Proof of concept

Ambrevar:
This is what `set-url2` looks like:

```lisp
(let ((url (prompt-minibuffer
                :input-prompt (format nil "Open URL in ~A buffer"
                                      (if new-buffer-p
                                          "new"
                                          "current"))
                :input-buffer (if prefill-current-url-p
                                  (object-string (url (current-buffer))) "")
                :default-modes '(set-url-mode minibuffer-mode)
                :suggestion-function (history-suggestion-filter
                                      :prefix-urls (list (object-string
                                                          (url (current-buffer)))))
                :history history
                :must-match-p nil)))

      (when (typep url 'history-entry)
        ;; In case read-from-minibuffer returned a string upon
        ;; must-match-p.
        (setf url (url url)))
      (ffi-within-renderer-thread       ; TODO: Wrap all ffi-* method content with ffi-within-renderer-thread.
       *browser*
       (lambda ()
         (buffer-load url :buffer (if new-buffer-p
                                      (make-buffer-focus)
                                      (current-buffer))))))
```

No more `with-result` or `with-result*` or `%callback`, just a regular `let`! :)

To test, you need to press `C-return` on the URL.  It won't work with `return` in this demo because I didn't want to break other commands.

The new `prompt-minibuffer` takes all minibuffer args directly, no need to call `make-minibuffer` anymore.

`chanl` is great!  After merging this, I propose we use ChanL instead of lparallel everywhere (it's used by the download manager).

The user (or the devs) should not need to bother with `ffi-within-renderer-thread`.  I suggest we move it inside all `ffi-*` methods.
Example:

```lisp
(defmethod initialize-instance :after ((buffer status-buffer) &key)
  (ffi-within-renderer-thread
   *browser*
    (lambda ()
  (with-slots (gtk-object) buffer
    (setf gtk-object (make-web-view buffer))
    (gobject:g-signal-connect
     gtk-object "decide-policy"
     (lambda (web-view response-policy-decision policy-decision-type-response)
       (declare (ignore web-view))
       (on-signal-decide-policy buffer response-policy-decision policy-decision-type-response))))))
```

Maybe make a macro. 

Once set up, we can remove all `ffi-within-renderer-thread` calls in the program as well as `within-renderer-thread`.
A great benefit is that this approach guarantees us that calling functions from the REPL will never hang Nyxt anymore!

Waiting for #910 before proceeding.

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -5,6 +5,7 @@
   :serial t
   :depends-on (:alexandria
                :bordeaux-threads
+               :chanl
                :cl-css
                :cl-json
                :cl-markup
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -620,6 +620,7 @@ sometimes yields the wrong reasult."
 (define-ffi-method ffi-buffer-delete (buffer))
 (define-ffi-method ffi-buffer-load (buffer uri))
 (define-ffi-method ffi-buffer-evaluate-javascript (buffer javascript))
+(define-ffi-method ffi-buffer-evaluate-javascript-SYNC (buffer javascript))
 (define-ffi-method ffi-minibuffer-evaluate-javascript (window javascript))
 (define-ffi-method ffi-buffer-enable-javascript (buffer value))
 (define-ffi-method ffi-buffer-enable-javascript-markup (buffer value))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -746,6 +746,36 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
                                    (make-buffer-focus)
                                    (current-buffer))))))
 
+(define-command set-url2 (&key new-buffer-p prefill-current-url-p)
+  "Set the URL for the current buffer, completing with history."
+  (let ((history (minibuffer-set-url-history *browser*)))
+    (when history
+      (containers:insert-item history (url (current-buffer))))
+    (let ((url (prompt-minibuffer
+                :input-prompt (format nil "Open URL in ~A buffer"
+                                      (if new-buffer-p
+                                          "new"
+                                          "current"))
+                :input-buffer (if prefill-current-url-p
+                                  (object-string (url (current-buffer))) "")
+                :default-modes '(set-url-mode minibuffer-mode)
+                :suggestion-function (history-suggestion-filter
+                                      :prefix-urls (list (object-string
+                                                          (url (current-buffer)))))
+                :history history
+                :must-match-p nil)))
+
+      (when (typep url 'history-entry)
+        ;; In case read-from-minibuffer returned a string upon
+        ;; must-match-p.
+        (setf url (url url)))
+      (ffi-within-renderer-thread       ; TODO: Wrap all ffi-* method content with ffi-within-renderer-thread.
+       *browser*
+       (lambda ()
+         (buffer-load url :buffer (if new-buffer-p
+                                      (make-buffer-focus)
+                                      (current-buffer))))))))
+
 (define-command set-url-from-current-url ()
   "Set the URL for the current buffer, pre-filling in the current URL."
   (set-url :prefill-current-url-p t))
--- a/source/command.lisp
+++ b/source/command.lisp
@@ -284,7 +284,8 @@ This function can be `funcall'ed."
 
 (defmethod run ((command command) &rest args)
   "Run COMMAND over ARGS."
-  (apply #'funcall-safely (command-function command) args))
+  (chanl:pexec ()
+    (apply #'funcall-safely (command-function command) args)))
 
 (define-command noop ()                 ; TODO: Replace with ESCAPE special command that allows dispatched to cancel current key stack.
   "A command that does nothing.
--- a/source/element-hint.lisp
+++ b/source/element-hint.lisp
@@ -7,6 +7,120 @@
   (defun qs (context selector)
     "Alias of document.querySelector"
     (ps:chain context (query-selector selector)))
+
+  (defun qsa (context selector)
+    "Alias of document.querySelectorAll"
+    (ps:chain context (query-selector-all selector)))
+
+  (defun code-char (n)
+    "Alias of String.fromCharCode"
+    (ps:chain -string (from-char-code n)))
+
+  (defun add-stylesheet ()
+    (unless (qs document "#nyxt-stylesheet")
+      (ps:try
+       (ps:let* ((style-element (ps:chain document (create-element "style")))
+                 (box-style (ps:lisp (box-style (current-buffer))))
+                 (highlighted-style (ps:lisp (highlighted-box-style (current-buffer)))))
+         (setf (ps:@ style-element id) "nyxt-stylesheet")
+         (ps:chain document head (append-child style-element))
+         (ps:chain style-element sheet (insert-rule box-style 0))
+         (ps:chain style-element sheet (insert-rule highlighted-style 1)))
+       (:catch (error)))))
+
+  (defun hint-determine-position (rect)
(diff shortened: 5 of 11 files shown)


Ambrevar:
Note to reviewers: `return-selection` (I like the name better than `return-input` which is less accurate) is a copy-paste of `return-input` except that instead of calling the callback, it passes the selection to the channel.

In other words, the implementation is really simple!

Ambrevar:
To do: Try the same technique with Parenscripts.

Currently, our parenscripts are defined this way

```lisp
(defmacro define-parenscript (script-name args &body script-body)
  "Define parenscript function SCRIPT-NAME.
SCRIPT-BODY must be a valid parenscript and will be wrapped in (PS:PS ...).
Any Lisp expression must be wrapped in (PS:LISP ...).

The returned function is called with the ARGS lambda-list over the current
buffer."
  `(progn
     (defun ,script-name ,args
       (ffi-buffer-evaluate-javascript (current-buffer)
                                       (ps:ps ,@script-body)))))
```

We could edit the macro to pass a callback which just writes the result to a channel.
This way our parenscripts would be synchronous too, which means they would be first class functions that don't have race conditions and don't need to run within `with-result`.

Thoughts?

Ambrevar:
Fixes #908.

Ambrevar:
I'd like to retract my suggestion of moving ffi-within-renderer-thread inside all FFI methods: `ffi-within-renderer-thread` appends the thunk to a queue of tasks in the GTK idle loop.  Call ffi-within-renderer-thread multiple times is not the same as calling it one time over multiple things.

Maybe we should ask upstream or roll out our own benchmark.

Ambrevar:
I've just pushed a demo for parenscript :)

jmercouris:
I've tested this, I like it! Makes it much simpler to do a lot of things! I've been playing around with it, and have had a couple of questions:

+ error occurs on separate thread, we don't get debugger unless it joins on main thread, how might we handle this?
+ how to handle javascript that never returns for parenscript? is there a way to do timeouts?

Ambrevar:

> I've tested this, I like it! Makes it much simpler to do a lot of things! I've been playing around with it, and have had a couple of questions:
>
> + error occurs on separate thread, we don't get debugger unless it joins on main thread, how might we handle this?

I do get a the debugger of separate threads, at least with SLY.
Do you have a precise example to share?

> + how to handle javascript that never returns for parenscript? is there a way to do timeouts?

Timeouts are probably not a good idea since it would break script beyond
the timeout and hang Nyxt for the duration of the timeout.

Instead, I suggest we add an option to `define-parenscripts` to tell
whether it should be synchronous or asynchronous.

jmercouris:
What if we gave the caller the option to define whether the parenscript should run synchronously or asynchronously?

jmercouris:
I wasn't running in slime, I was just using the SLDB in a terminal, if it shows up in slime, then that is OK too.

Ambrevar:
> What if we gave the caller the option to define whether the parenscript should run synchronously or asynchronously?

Maybe.  We can do this in a second time when the need arises.

jmercouris:
Is this branch almost ready? I am working on the configuration and would like to take advantage of the new minibuffer :-D

Ambrevar:
I'll finish it tomorrow, sorry, I don't have much time for Nyxt this week.

jmercouris:
OK, I will probably just merge onto master then with my I M P R O V E M E N T S

jmercouris:
The second iteration will wait for this B R A N C H

Ambrevar:
I'm on it.
First pass pruned... 140 lines! O.o

Ambrevar:
Problem: It's a bit hard to control which thread is executing when
writing code.

Using ffi-within-renderer-thread has drawbacks:

- It may slow down execution (needs benchmark to confirm).
- Worse: It needs to be inserted everywhere and it's hard to guarantee
  that a piece of code is properly wrapped by
  ffi-within-renderer-thread.

I'll try a bit harder but this could be a dead end.

Alternative: Always stay on the GTK thread (as is the case on master)
and have prompt-minibuffer return a channel instead.

The caller (usually commands), then calls chanl:recv over it to actually
get the value.  The problem is that the recv operation may block the GTK
thread if it's run from there, which is again hard to control.


If neither is satisfactory, we need to devise a better plan for our
threading that gives us _control_.


@jmercouris  Thoughts?

jmercouris:
Some early thoughts:

There is no silver bullet. Mayhaps we will have to support both approaches. Sometimes it is quite useful to have synchronous minibuffer.

I can confirm a performance penalty from within-renderer-thread. If we can speed it up, it should be no great issue to wrap whatever is necessary.

jmercouris:
I think the best case scenario approach would be: responsibility of the user to wrap within-renderer-thread. Keep with-result macros for users that don't want to deal with threading and would prefer continuation passing style.

thoughts?

Ambrevar:
I'll experiment.  This will need some work for sure.

Ambrevar:
Continuing in #977.

Ambrevar:
@aartaka About your REPL issue, can you try this branch if it works for you?
It should start as usual, then you can call set-url2 and confirm with `C-return` (`return` won't work).

aartaka:
> @aartaka About your REPL issue, can you try this branch if it works for you?
> It should start as usual, then you can call set-url2 and confirm with `C-return` (`return` won't work).

It does start, and the page set when calling `set-url2` is even loaded. It hangs just after loading the requested page, though.

Here's the relevant excerpt from logs (Nyxt hangs after the last message):
```
<DEBUG> [16:41:49] nyxt browser.lisp (request-resource) -
  Forwarding: https://nyxt.atlas.engineer/start
<DEBUG> [16:41:49] nyxt renderer-gtk.lisp (on-signal-decide-policy gtk-buffer) -
  Forward to renderer (unchanged URL).
<DEBUG> [16:41:49] nyxt browser.lisp (request-resource) -
  Forwarding: https://nyxt.atlas.engineer/start
<DEBUG> [16:41:49] nyxt renderer-gtk.lisp (on-signal-decide-policy gtk-buffer) -
  Forward to renderer (unchanged URL).
 <INFO> [16:41:49] nyxt message.lisp (echo form-fun-6) -
  Loading "https://nyxt.atlas.engineer/start".
<DEBUG> [16:41:50] nyxt browser.lisp (request-resource) -
  Forwarding: https://nyxt.atlas.engineer/start
<DEBUG> [16:41:50] nyxt renderer-gtk.lisp (on-signal-decide-policy gtk-buffer) -
  Forward to renderer (unchanged URL).
 <INFO> [16:41:50] nyxt session.lisp (store data-profile session-data-path) -
  Saving session to "/home/aartaka/.local/share/nyxt/sessions/default.lisp".
<DEBUG> [16:41:50] nyxt browser.lisp (request-resource) -
  Forwarding: https://duckduckgo.com/search.html?width=768&duck=yes&prefill=Search%20DuckDuckGo&focus=yes
<DEBUG> [16:41:50] nyxt renderer-gtk.lisp (on-signal-decide-policy gtk-buffer) -
  Forward to renderer (unchanged URL).
<DEBUG> [16:41:51] nyxt browser.lisp (request-resource) -
  Forwarding: https://duckduckgo.com/search.html?width=768&duck=yes&prefill=Search%20DuckDuckGo&focus=yes
<DEBUG> [16:41:51] nyxt renderer-gtk.lisp (on-signal-decide-policy gtk-buffer) -
  Forward to renderer (unchanged URL).
<DEBUG> [16:41:51] nyxt/web-mode zoom.lisp (unzoom-page form-fun-8) -
  Calling command UNZOOM-PAGE.
```

PR #931, opened 2020-09-05
Define web-buffer class and move load-status, proxy, cookies and cert…

Ambrevar:
…ificates there.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -22,18 +22,6 @@ See the `data-path' class and the `expand-path' function.")
    (url (quri:uri ""))
    (url-at-point (quri:uri ""))
    (title "")
-   (load-status :unloaded
-                :type (or (eql :loading)
-                          (eql :finished)
-                          (eql :unloaded))
-                :accessor nil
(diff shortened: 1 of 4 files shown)


source/buffer.lisp
```diff
@@ -244,7 +255,15 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
...
 
-(define-user-class buffer)
+(define-user-class web-buffer)
+
+(defmethod initialize-instance :after ((buffer web-buffer) &key)
+  (when (expand-path (cookies-path buffer))
+    (ensure-parent-exists (expand-path (cookies-path buffer)))))
+
+(export-always 'web-buffer-p)
+(defun web-buffer-p (buffer)
```
jmercouris, line 265: you could use a method that specializes on web-buffer and another one that just specializes on buffer as has been done for internal buffers

Ambrevar: I saw you did that, I just wanted to play with a different implementation style.  This one is shorter, I'm don't know if there is any drawback.

Ambrevar:
@jmercouris Good to merge?

jmercouris:
Indeed!

Issue #934, opened 2020-09-07
Feature Request: Tabs Group Profiles alla Guix's Profiles


ccao001:
@Ambrevar 

## Question 

How hard would it be to implement profiles (alla guix, nix, flatpak, etc...) but for your current/previous collection(s) of open tabs?

## Further Explanation

Firefox has the option `Restore Previous Session` which brings back all the user's opened tabs from the previous session.

The problem is that the user only has the ability to restore the __previous session__. If the user happens to open the browser again and starts browsing only to subsequently close the session, then the previous previous session would be lost. Maybe there's a similarity to what [history as a tree](https://github.com/atlas-engineer/nyxt/blob/1.5.0/documents/MANUAL.org#history) solves in nyxt?

[Bookmark Tags](https://github.com/atlas-engineer/nyxt/blob/1.5.0/documents/MANUAL.org#bookmarks) seem to help with this but I am thinking here of grouped tab sessions as a semantic object that has a version controlled traversable history.

## Proposal 

What I am proposing is that at the end of a browsing session the user can save all tabs into an `open-tabs-manifest` and that manifest can subsequently be put under a version control system.

Maybe this tabs manifest can be represented with a similar syntax to a guix manifest or just plain S-expressions to be used for later parsing by nyxt. 

Maybe json would be a more portable representation here? Or there can be levels of [intermediary representations](https://www.gnu.org/software/guile/manual/html_node/Tree_002dIL.html) for tab groups.

This would allow the user to `rollback` to any previous `session-tabs-profile`. These `tabs-profiles` contents can be browsed and chosen from nyxt's minibuffer.

I'm just thinking out loud but I thought that this might be a nice feature that nyxt can implement in the future. 

What are your thoughts? 

## Mailing List or Wiki or Not?

Would nyxt consider having a mailing list or wiki? Maybe that is the appropriate place to document and discuss these ideas in a more persistent and public manner.

Sorry if this does not fit the bill of what is expected in an issue request on github.

Ambrevar:
Nyxt supports multiple sessions.
From a shell, you can start Nyxt with

  nyxt --with-path session=foo

to start the foo session.
The session is stored in ~/.local/share/nyxt/sessions by default.  It's
in a s-exp format, so it's versionable.  It preserves the buffer histories.

What's mostly missing is the ability to select your session on startup,
although that wouldn't be too hard to implement.

Would that cover your use case?

About the mailing list: We thought of having one, but we are currently a
[...]

Issue #631, opened 2020-04-04
Synchronising session/bookmarks/history


bijang:
Currently I mainly use Firefox with a few addons including
https://tab-session-manager.sienori.com/
https://github.com/piroor/treestyletab
https://www.xbrowsersync.org/
It's an incomplete solution, not everything is synced yet, but I chose this because it was the best combo for my demands at this time.
The  next browser can do everything already with relation to advanced bookmark management, history as tree etc

it does raise a few questions for me, (forgive me I don't have next installed on this machine at present). 

- How is history managed in next, does it erase after a time or can it be erased? 
- This is where history and 'session' overlap, ie history is made up of sessions. Can one 'save' a session (meaning the open windows/tabs), to then come back to the same, or even a previous session (in the way managed by the tab session manager extension)? 
In this extension, the session is tagged automatically with metadate including time and date, number  of windows open and number of tabs within windows open, as well as optionally giving other tags or titles to the session. [I understand the concept of tabs in next is different]. 
Please have a look at tab session extension if you have time, it can be easier than my descibing it. 
Or am I barking up the wrong tree, and the way the tab paradigm works in next and the history, with ability to fuzzy search between places, provides a 'fluidity' in a way that means there is no such thing as a well defined 'session' you may have in a traditional browser (again I havn't used next enough yet myself to answer these questions). 
I think all this probably depends on how you define a session. I'm glad there is tree history in next at least allow some intropection into the work flow that went before. Is the history tree populated as a plain text list or do they conisit of working hyperlinks? (if not would this be possible?)
- Synchronisation, of bookmarks and related metadata, history and session (whatever this may be defined as); ie portability. So if I move to a different laptop can I access the history and bookmarks etc, and therefore modify them, and have these synced up. Could this be done without the need for stting up a central server instance (like xbrowsersync)? Would the data set up be like a git where can push and pull into it? The way pass does for password management. Would security factor into it (in terms of holding the browser hisotry and bookmarks in an encrypted file)

I think this last one is important as next gets more popular and esp when next gets it's inevitable companion mobile app ;)

Apologies is so long, I hope this is ok as an issue, or perhaps you want me to put it into a few seperate issues?

tviti:
For your last bullet: I wound up making the bookmark dir a git repo, and
rolled a few commands for pushing/pulling. The implementation is pretty
shitty, and every so often I have to go in and manually merge (when I
forget to sync for a while), but maybe this will give you some ideas.

[...]

PR #937, opened 2020-09-07
Add session-dumping commands

aartaka:
This adds two commands (`store-session-by-name` and `restore-session-by-name`, namely) to store/restore session data to/from a dedicated user-named file.

### Motivation

As said in #934, It'd be nice to have an  ability to have `profile`-like (as used in GNU Guix) buffer collections that one can store and restore at arbitrary moment.

### Caveats

- A `;; Delete the old buffers?` commentary from session-specified `restore` would fit this changeset perfectly, although such a radical thing needs some discussion.
- Should we add some name suggestions for sessions, if there are no sessions but default one?
- Should we even list default session as option for storage/restoration?
- Name clash with `store` and `restore` generic functions. Is it harmful? I think it's okay, given that `(re)store-session-by-name` and `(re)store` are connected.

### Future directions:

- Maybe make session serialization more readable.

I'll be glad to know what you think of it :)

Closes: #934.

### DIFF
--- a/source/session.lisp
+++ b/source/session.lisp
@@ -65,11 +65,6 @@ Currently we store the list of current URLs of all buffers."
            (log:info "Restoring ~a."
                      (mapcar (alex:compose #'object-string #'htree:data #'htree:current)
                              buffer-histories))
-           ;; Delete the old buffers?
-           ;; (maphash (lambda (id buffer)
-           ;;            (declare (ignore id))
-           ;;            (buffer-delete buffer))
-           ;;          buffers)
            ;; Make the new ones.
            ;; TODO: Replace the loop with a function `make-buffer-from-history' in web-mode?
            (loop for history in buffer-histories
@@ -86,3 +81,51 @@ Currently we store the list of current URLs of all buffers."
            )))
     (error (c)
       (log:warn "Failed to restore session from ~a: ~a" (expand-path path) c))))
+
+(defun session-name-suggestion-filter (minibuffer)
+  (fuzzy-match (input-buffer minibuffer)
+               (mapcar #'pathname-name
+                       (uiop:directory-files
+                        (uiop:pathname-directory-pathname
+                         (expand-path (session-path (current-buffer))))))))
+
+(define-command store-session-by-name ()
+  "Store the session data (i.e. all the opened buffers) in the file named by user input."
+  (with-result (name (read-from-minibuffer
+                      (make-minibuffer
+                       :input-prompt "The name to store session with"
+                       :history (minibuffer-session-restore-history *browser*)
+                       :suggestion-function #'session-name-suggestion-filter)))
+    (when name
+      (store (data-profile (current-buffer))
+             (make-instance 'session-data-path
+                            :dirname (dirname (session-path (current-buffer)))
+                            :basename name)))))
+
+(define-command restore-session-by-name ()
+  "Restore the session data from the file named by user input."
+  (with-result (name (read-from-minibuffer
+                      (make-minibuffer
+                       :input-prompt "The name of the session to restore"
+                       :history (minibuffer-session-restore-history *browser*)
+                       :suggestion-function #'session-name-suggestion-filter)))
+    (when name
+      (restore (data-profile (current-buffer))
+               (make-instance 'session-data-path
+                              :basename name
+                              :dirname (dirname (session-path (current-buffer))))))))
+
+(define-command replace-session-by-name ()
+  "Delete all the buffers of the current session and restore the one chosen by user."
+  (with-result (name (read-from-minibuffer
+                      (make-minibuffer
+                       :input-prompt "The name of the session to replace the current one"
+                       :history (minibuffer-session-restore-history *browser*)
+                       :suggestion-function #'session-name-suggestion-filter)))
+    (when name
+      (dolist (buffer (buffer-list))
+        (buffer-delete buffer))
+      (restore (data-profile (current-buffer))
+               (make-instance 'session-data-path
+                              :basename name
+                              :dirname (dirname (session-path (current-buffer))))))))
--- a/source/start.lisp
+++ b/source/start.lisp
@@ -255,15 +255,10 @@ To change the default buffer, e.g. set it to a given URL:
       ;; will be overwritten because if-confirm is asynchronous.
       ;; Implement synchronous minibuffer to fix this.
       (when (expand-path (session-path buffer))
-        (flet ((restore-session ()
-                 (when (uiop:file-exists-p (expand-path (session-path buffer)))
-                   (log:info "Restoring session ~s." (expand-path (session-path buffer)))
-                   (restore (data-profile buffer) (session-path buffer)))))
-          (match (session-restore-prompt *browser*)
-            (:always-ask (if-confirm ("Restore previous session?")
-                                     (restore-session)))
-            (:always-restore (restore-session))
-            (:never-restore (log:info "Not restoring session.")))))
+        (match (session-restore-prompt *browser*)
+               (:always-ask (restore-session-by-name))
+               (:always-restore (restore (data-profile buffer) (session-path buffer)))
+               (:never-restore (log:info "Not restoring session."))))
       (if urls
           (open-urls urls)
           (window-set-active-buffer window (funcall-safely buffer-fn))))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -66,6 +66,7 @@ Most recent messages are first.")
    (minibuffer-generic-history (make-ring))
    (minibuffer-search-history (make-ring))
    (minibuffer-set-url-history (make-ring))
+   (minibuffer-session-restore-history (make-ring))
    (recent-buffers (make-ring :size 50)
                    :export nil
                    :documentation "A ring that keeps track of deleted buffers.")


source/session.lisp
```diff
@@ -86,3 +86,34 @@ Currently we store the list of current URLs of all buffers."
...
+               ;; TODO: add suggested names, like "work", "reading", if empty?
+               (mapcar #'pathname-name
+                       (uiop:directory-files
+                        (dirname (session-path (current-buffer)))))))
+
+(define-command store-session-by-name ()
+  "Store the session data (i.e. all the opened buffers) in the file named by user input."
+  (with-result (name (read-from-minibuffer
+                      (make-minibuffer
+                       :input-prompt "The name to store session with:"
```
Ambrevar, line 101: Maybe default the input to the current session basename.

Ambrevar: No need for the trailing ":", it's automatically added.

source/session.lisp
```diff
@@ -86,3 +86,34 @@ Currently we store the list of current URLs of all buffers."
...
+      (store (data-profile (current-buffer))
+             (make-instance 'session-data-path
+                            :dirname (dirname (session-path (current-buffer)))
+                            :basename name)))))
+
+(define-command restore-session-by-name ()
+  "Restore the session data from the file named by user input."
+  (with-result (name (read-from-minibuffer
+                      (make-minibuffer
+                       :input-prompt "The name to store session with:"
```
Ambrevar, line 113: Same comment about `:`.

source/session.lisp
```diff
@@ -86,3 +86,34 @@ Currently we store the list of current URLs of all buffers."
            )))
     (error (c)
       (log:warn "Failed to restore session from ~a: ~a" (expand-path path) c))))
+
+(defun session-name-suggestion-function (minibuffer)
+  (fuzzy-match (input-buffer minibuffer)
+               ;; TODO: add suggested names, like "work", "reading", if empty?
```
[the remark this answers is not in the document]
aartaka: To help the user with coming up with some name :)

source/session.lisp
aartaka: > Maybe default the input to the current session basename.

Isn't it always "default" unless overriden? Additionally, current session is saved really often, so no need to re-save it manually.

aartaka:
All fixed now.

Ambrevar:
> - A `;; Delete the old buffers?` commentary from session-specified `restore` would fit this changeset perfectly, although such a radical thing needs some discussion.

Yes, this needs to be resolved!  Maybe make an option?  Or provide 2
commands:

- restore-by-name
- restore-by-name-and-replace

> - Should we add some name suggestions for sessions, if there are no sessions but default one?

I don't know, can you think of a use case?

> - Should we even list default session as option for storage/restoration?

To be clear, I think you meant the "current" session.  The default one
might not be the current one.  I think there is no need to list the
current session since it's already stored / restored.

> - Name clash with `store` and `restore` generic functions. Is it harmful? I think it&#39;s okay, given that `(re)store-session-by-name` and `(re)store` are connected.

I think it's OK.

Ambrevar:
> Actually it closes #631 too :)

While it addresses some points in #631, I think it does not address the question of synchronization.  #631 is about adding sync-related commands, e.g. a wrapper around Git with a service to GitHub or the like.

Ambrevar:
While we are at it, as a second commit, what about added an option on startup to choose which session file to restore?  Should be trivial now that we have restore-by-name.

source/session.lisp
Ambrevar: Sorry, I was confused, you are right.
Better: add a history so that when pressing M-h the user only gets the history of sessions and not the generic history.

source/session.lisp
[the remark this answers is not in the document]
Ambrevar: Hmm... Well, it does not cost anything but I'm not sure it's really helpful!  As you wish :)

aartaka:
> While we are at it, as a second commit, what about added an option on startup to choose which session file to restore? Should be trivial now that we have restore-by-name.

Sounds good!

aartaka:
> Yes, this needs to be resolved! Maybe make an option? Or provide 2 commands: - restore-by-name - restore-by-name-and-replace

Yes, sounds good to me.

>> - Should we add some name suggestions for sessions, if there are no sessions but default one?
> I don't know, can you think of a use case?

No, it just can be convenient to make it context-dependent, like bookmark tags. Not really necessary though, because saving sessions implies some context that can be summarized in the name.

> To be clear, I think you meant the "current" session. The default one might not be the current one. I think there is no need to list the current session since it's already stored / restored.

Yes, this make sense.

aartaka:
Here you go:
- `replace-session-by-name` deletes all the buffers and then restores the session.
- `make-startup-function` now uses `restore-session-by-name`. It seems to chatty though. Maybe define a browser slot with default session name and `session-restore-prompt` option like `:always-restore-default-session`?

aartaka:
- `session-name-suggestion-filter` no longer deletes the current session name from the suggestions.
- `minibuffer-session-restore-history` slot of `browser`.

Ambrevar:
About the last commit: chaining 2 minibuffers, or always prompting a minibuffer when the users just wants the default session does not seem right.

Instead, I suggest 
- Leave `:always-restore` as it is.
- Move `:always-ask` out of `if-confirm`.  After all, if the user escapes the `restore-session-by-name` minibuffer, no session will be restored.

Thoughts?

aartaka:
>     * Leave `:always-restore` as it is.
> 
>     * Move `:always-ask` out of `if-confirm`.  After all, if the user escapes the `restore-session-by-name` minibuffer, no session will be restored.

You mean something like:
``` lisp
(when (expand-path (session-path buffer))
        (match (session-restore-prompt *browser*)
               (:always-ask (restore-session-by-name))
               (:always-restore
                (when (uiop:file-exists-p (expand-path (session-path buffer)))
                  (log:info "Restoring session ~s." (expand-path (session-path buffer)))
                  (restore (data-profile buffer) (session-path buffer))))
               (:never-restore (log:info "Not restoring session."))))
```

aartaka:
Seems to be pretty complete now.

However, I experience minibuffer showing suggestions in `replace-session-by-name`, but not in two other commands. Recompiling them manually makes suggestions show up. That's strange.

source/start.lisp
```diff
@@ -255,15 +255,13 @@ To change the default buffer, e.g. set it to a given URL:
...
-          (match (session-restore-prompt *browser*)
-            (:always-ask (if-confirm ("Restore previous session?")
-                                     (restore-session)))
-            (:always-restore (restore-session))
-            (:never-restore (log:info "Not restoring session.")))))
+        (match (session-restore-prompt *browser*)
+               (:always-ask (restore-session-by-name))
+               (:always-restore
+                (when (uiop:file-exists-p (expand-path (session-path buffer)))
+                  (log:info "Restoring session ~s." (expand-path (session-path buffer)))
```
Ambrevar, line 262: Maybe we can move this `log:info` into the `restore` method.  Then there would be no need for the `when`.
But that would also make the standard output much more verbose because the session gets stored/restored all the time.
What do you think?  If verbosity is a problem, let's leave this `when` and add a comment explaining why we do it this way.

Ambrevar:

> However, I experience minibuffer showing suggestions in `replace-session-by-name`, but not in two other commands. Recompiling them manually makes suggestions show up. That's strange.

Can't tell why at first glance.  I'll test later.

source/session.lisp
```diff
@@ -86,3 +81,54 @@ Currently we store the list of current URLs of all buffers."
...
+
+(defun session-name-suggestion-filter (minibuffer)
+  (fuzzy-match (input-buffer minibuffer)
+               (mapcar #'pathname-name
+                       (uiop:directory-files
+                        (reduce #'(lambda (a b) (str:concat a "/" b))
+                                (pathname-directory
+                                 (expand-path (session-path (current-buffer))))
+                                :start 1
+                                :initial-value "")))))
```
Ambrevar, line 93: What are you trying to do with this `reduce`?  Convert the pathname to a string?  If so, use `namestring` ;)

source/start.lisp
aartaka: Good catch! It's already logged in the `restore`, so we don't need it.

Session storage output is quite verbose sometimes, but I have no strong opinion about whether we should remove these.

source/session.lisp
aartaka: `pathname-directory` returns a list of strings (e.g., `(:ABSOLUTE "home" "aartaka" ".local" "share" "nyxt" "sessions")`). `namestring` seems to be incapable of parsing these. 

Maybe there's a better way to parse these directory specs, but I'm unaware of it :)

aartaka:
Deleted the unnecessary `log:info` in `make-startup-function`.

source/session.lisp
aartaka: OK, `fad:pathname-directory-pathname` seems like a good way to do this.

aartaka: Sorry for being unclear. What I wanted to get with this `reduce` was the parent directory of the current session file.

aartaka:
Prettify the code with `cl-fad` functions and put things into respective commits (the `session-name-suggestion-filter` change didn't belong to the startup session restoration commit).

source/session.lisp
Ambrevar: Don't use FAD, it's been mostly superseded by UIOP.
Use uiop:pathname-parent-directory-pathname instead.

source/session.lisp
```diff
@@ -86,3 +81,51 @@ Currently we store the list of current URLs of all buffers."
            )))
     (error (c)
       (log:warn "Failed to restore session from ~a: ~a" (expand-path path) c))))
+
+(defun session-name-suggestion-filter (minibuffer)
+  (fuzzy-match (input-buffer minibuffer)
+               (mapcar #'pathname-name
+                       (uiop:directory-files
+                        (fad:pathname-directory-pathname
```
Ambrevar, line 89: We don't use FAD in Nyxt ;)

source/start.lisp
```diff
@@ -255,15 +255,12 @@ To change the default buffer, e.g. set it to a given URL:
...
-                   (restore (data-profile buffer) (session-path buffer)))))
-          (match (session-restore-prompt *browser*)
-            (:always-ask (if-confirm ("Restore previous session?")
-                                     (restore-session)))
-            (:always-restore (restore-session))
-            (:never-restore (log:info "Not restoring session.")))))
+        (match (session-restore-prompt *browser*)
+               (:always-ask (restore-session-by-name))
+               (:always-restore
+                (when (uiop:file-exists-p (expand-path (session-path buffer)))
```
Ambrevar, line 261: Why do we need the `when`?  If there is no file, `restore` should not fail.

Ambrevar:
> However, I experience minibuffer showing suggestions in `replace-session-by-name`, but not in two other commands. Recompiling them manually makes suggestions show up. That's strange.

I can't reproduce this.  Can you provide a recipe?

jmercouris:
> However, I experience minibuffer showing suggestions in `replace-session-by-name`, but not in two other commands. Recompiling them manually makes suggestions show up. That's strange.
> I can't reproduce this. Can you provide a recipe?

I don't think it is worth worrying about, it is probably just a caching issue.

Ambrevar:
Something else: If I start Nyxt by selecting a FOO session that was
previously saved,

(session-path (current-buffer))

returns the "default" session still.  This means that loading a new
buffer will save the FOO session + the new buffer to "default".

Is this natural?  Should we set the default session to FOO?

aartaka:
> Something else: If I start Nyxt by selecting a FOO session that was previously saved, (session-path (current-buffer)) returns the "default" session still. This means that loading a new buffer will save the FOO session + the new buffer to "default". Is this natural? Should we set the default session to FOO?

Hmmmm... Yes, it sound like we need to set it to a new path. This is also consistent with the idea of changing sessions -- the contexts will be fully separated.

But then, sessions are something beyond buffers, because they are about the collections of buffers inside the browser, so changing `session-path` in one buffer will not change the path for all the other buffers and will duplicate the same data to different paths. We need to either move `session-path` back to browser, change default paths when restoring session, or stay with the current version.

I'd leave it as it is now, because it will preserve per-buffer storage and private profiles, while still allowing to load and store session snapshots. Unfortunately, default session will be somewhat cluttered with the other sessions' data. But do we really need to keep it clean for future restoration, if we now have other sessions to restore?

jmercouris:
I agree with Artyom's assessment. While it is a bit annoying, it doesn't prevent you from constantly using named sessions. The only collateral damage is that the default session is polluted.

aartaka:
Somewhat off-topic: I've realized now, that the buffers with the `private-data-profile` will be saved to default `session-path`, if at least one buffer with non-`private` data profile will call `store(-session-by-name)`. It shouldn't be this way, I believe.

Then moving `session-path` to `buffer` could have been a bad idea, because session is a browser-wide data. We need to think more about how we handle sessions in the context of private buffers :)

One radical idea would be to remove session at all and make history tree able to have several roots (or just use level-1 nodes for this?) and mark nodes by buffers they belong to, to allow several buffers share one history tree, one root per buffer (start buffer history as a branch of another buffer history sounds quite weird, but I can make sense of it). Storing sessions, then, will be a plain history storage. This way we'll be able to safely isolate the data of private buffers and have some other cool things based on the global tree-based history :sparkles:.

But it's just one option. There could be better way to resolve it. Does this deserve a separate issue or should discussion on sessions be part of #247?

aartaka:
> The only collateral damage is that the default session is polluted.

Well, sometimes you need to have a chaotic side-effectful grab-bag of things to be productive ;) 

I may be mistaken in this, but that's how I use my default Guix profile -- I install throwaway programs there and can sacrifice it anytime, because I have all the important packages in the manifest files for other profiles. Seems to be similar to how we want to store sessions to dedicated files.

Ambrevar:

> One radical idea would be to remove session at all and make history tree able to have several roots (or just use level-1 nodes for this?) and mark nodes by buffers they belong to, to allow several buffers share one history tree, one root per buffer (start buffer history as a branch of another buffer history sounds quite weird, but I can make sense of it). Storing sessions, then, will be a plain history storage. This way we'll be able to safely isolate the data of private buffers and have some other cool things based on the global tree-based history :sparkles:.

At first glance this seems too complex or unintuitive.  But if you get
convinced this is a good idea, please develop it with examples, I'm
curious about where this could lead us.

Ambrevar:
Here is what I think is the most intuitive: My default session can be
specified on startup.  This session should be shared by all new buffers,
unless the buffer is specialized to use another session (e.g. private
buffers).

We can add a command to change the default session.  This would
essentially redefine the BUFFER user class with define-configuration on
the session-path slot.

As @aartaka mentioned, changing the session path in just one buffer
makes little sense, so I think there is no need to do add a command for
that.

Thoughts?

aartaka:
> Here is what I think is the most intuitive: My default session can be specified on startup. This session should be shared by all new buffers, unless the buffer is specialized to use another session (e.g. private buffers).

What do you mean? Should private buffers have an empty/differing `session-path` or something like that then? Or should we have some other session-identifying slot to look up which buffers we can/cannot store as session?

aartaka:
>> However, I experience minibuffer showing suggestions in `replace-session-by-name`, but not in two other commands. Recompiling them manually makes suggestions show up. That's strange.

> I can't reproduce this. Can you provide a recipe?

I had first-iteration definitions of `(re)store-session-by-name` in my init file that were shadowing the ones from the patch but not `replace-session-by-name` xD

Ambrevar:

> What do you mean? Should private buffers have an empty/differing
> `session-path` or something like that then?

Exactly.  It's trivial, it's just one method definition.
Do you foresee a pitfall?

aartaka:
I don't yet understand how to implement it, so I don't see pitfalls :)

In any way, is it directly connected to this patchset? Maybe it will be faster/easier if you'll merge and implement the session separation in a separate commit.

Ambrevar:
I'm a little worried about merging this just now because if the first
thing Nyxt does is prompt the user for sessions which may overwrite
their previous session unexpectedly, this could frustrate many!

jmercouris:
Running master on the bleeding edge means that users should expect this. If they want more stable behavior the ragged releases are available. For that reason I think it is OK!

jmercouris:
ragged-> tagged

aartaka:
Hmmmm... That's a valid cause for worrying. 

> It's trivial, it's just one method definition.

Then, can you expand on this idea or, maybe, even push the changes (I've opened the push access to `session-snapshots` for maintainers) you meant?

aartaka:
Oh, the quoted message was about private buffers being stored in session. It won't solve the problem of session rewriting, then. 

On the subject of session rewriting: do we have any alternative to making default session a chaotic grab-bag of histories here? I'm not sure there's one. Default session rewriting seems to be unavoidable.

Ambrevar:
@jmercouris: Feel free to merge this if there is no other objection.

@aartaka: If this gets merged, please open an issue detailing the
current behavioural issues with multiple session management.
If you can't think of an alternative I'll try to put myself to it in a
couple of days, see if I can help ;)

aartaka:
>@aartaka: If this gets merged, please open an issue detailing the current behavioural issues with multiple session management. If you can't think of an alternative I'll try to put myself to it in a couple of days, see if I can help ;)

Sure :)

jmercouris:
Et voila! Great work Artyom!!!

PR #945, opened 2020-09-09
Easy Configuration

jmercouris:
This allows describe-slot/describe-class to set configuration. When the user types in describe-slot or describe-class there is a button appearing that says "Configure". The user can click on that to set the value. The value will also be persisted in a config.lisp file. This config.lisp file, like the init.lisp file can be controlled via command line flags.

This is the first draft implementation. The second implementation will feature:

+ type checking
+ more efficient usage of define-configuration in config.lisp, instead of appending new define-configuration forms, it will only persist define-configuration forms with all slots that are different than the default

Type checking is waiting on the new synchronous minibuffer (it will loop querying the user until the result satisfies the type information).

### DIFF
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -217,8 +217,9 @@ A command is a special kind of function that can be called with
   (with-result (input (read-from-minibuffer
                        (make-minibuffer
                         :input-prompt (format nil "Configure slot value ~a" slot))))
-    (eval `(define-configuration ,class
-             ((,slot (read-from-string ,input)))))))
+    (let ((configuration-form `(define-configuration ,class
+                                 ((,slot (read-from-string ,input))))))
+      (eval configuration-form))))
 
 (define-command describe-slot ()
   "Inspect a slot and show it in a help buffer."
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -217,7 +217,7 @@ A command is a special kind of function that can be called with
       (set-current-buffer help-buffer))))
 
 (defun configure-slot (slot class)
-  "Set the value of a slot in a users config.lisp."
+  "Set the value of a slot in a users autoconfig.lisp."
   (with-result (input (read-from-minibuffer
                        (make-minibuffer
                         :input-prompt (format nil "Configure slot value ~a" slot))))
@@ -227,11 +227,11 @@ A command is a special kind of function that can be called with
                              ((,slot ,input))))))
 
 (defun append-configuration (form)
-  (with-data-file (file *config-file-path*
+  (with-data-file (file *autoconfig-file-path*
                         :direction :output
                         :if-does-not-exist :create
                         :if-exists :append)
-    (log:info "Appending configuration form ~a to ~s." form (expand-path *config-file-path*))
+    (log:info "Appending configuration form ~a to ~s." form (expand-path *autoconfig-file-path*))
     (format file "~&~a~%" form)))
 
 (define-command describe-slot ()
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -217,7 +217,7 @@ A command is a special kind of function that can be called with
       (set-current-buffer help-buffer))))
 
 (defun configure-slot (slot class)
-  "Set the value of a slot in a users autoconfig.lisp."
+  "Set the value of a slot in a users auto-config.lisp."
   (with-result (input (read-from-minibuffer
                        (make-minibuffer
                         :input-prompt (format nil "Configure slot value ~a" slot))))
@@ -227,11 +227,11 @@ A command is a special kind of function that can be called with
                              ((,slot ,input))))))
 
 (defun append-configuration (form)
-  (with-data-file (file *autoconfig-file-path*
+  (with-data-file (file *auto-config-file-path*
                         :direction :output
                         :if-does-not-exist :create
                         :if-exists :append)
-    (log:info "Appending configuration form ~a to ~s." form (expand-path *autoconfig-file-path*))
+    (log:info "Appending configuration form ~a to ~s." form (expand-path *auto-config-file-path*))
     (format file "~&~a~%" form)))
 
 (define-command describe-slot ()
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -217,7 +217,8 @@ A command is a special kind of function that can be called with
   (with-result (input (read-from-minibuffer
                        (make-minibuffer
                         :input-prompt (format nil "Configure slot value ~a" slot))))
-    (print input)))
+    (eval `(define-configuration ,class
+             ((,slot (read-from-string ,input)))))))
 
 (define-command describe-slot ()
   "Inspect a slot and show it in a help buffer."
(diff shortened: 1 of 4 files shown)


jmercouris:
@Ambrevar OK to merge?

source/help.lisp
```diff
@@ -207,6 +213,24 @@ A command is a special kind of function that can be called with
       (ffi-buffer-evaluate-javascript help-buffer insert-help)
       (set-current-buffer help-buffer))))
 
+(defun configure-slot (slot class)
+  "Set the value of a slot in a users config.lisp"
```
Ambrevar, line 217: Trailing period missing.
"config.lisp" is not a well defined term.  How shall we call it?  "Configuration file" is too easy to confuse with the initialization file (we might already been using the two terms interchangeably).

source/help.lisp
```diff
@@ -207,6 +213,24 @@ A command is a special kind of function that can be called with
...
+    (append-configuration `(define-configuration ,class
+                             ((,slot ,input))))))
+
+(defun append-configuration (form)
+  (with-data-file (file *config-file-path*
+                        :direction :output
+                        :if-does-not-exist :create
+                        :if-exists :append)
+    (log:info "Appending configuration form ~a to ~s." form (expand-path *config-file-path*))
+    (format file "~%~a" form)))
```
Ambrevar, line 232: Maybe optional newline first, then mandatory newline? `~&~a~%`

source/start.lisp
```diff
@@ -7,6 +7,10 @@
 (defvar *init-file-path* (make-instance 'data-path :basename "init")
   "The path of the initialization file.")
 
+(export-always '*config-file-path*)
+(defvar *config-file-path* (make-instance 'data-path :basename "config")
+  "The path of the generated configuration file.")
```
Ambrevar, line 12: See above about the naming.
Having something in the name that implies it's generated would be a good idea.

source/start.lisp
```diff
@@ -55,6 +71,15 @@ use the socket without parsing any init file.")
            :short #\I
            :long "no-init"
            :description "Do not load the user init file.")
+    (:name :init
+           :short #\c
+           :long "config"
+           :arg-parser #'identity
+           :description "Set path to config file.")
+    (:name :no-init
```
Ambrevar, line 79: The `:name` is wrong ;)

source/start.lisp
```diff
@@ -216,6 +241,10 @@ Return the short error message and the full error message as second value."
   "Load or reload the init file."
   (load-lisp init-file :package (find-package :nyxt-user)))
 
+(define-command load-config-file (&key (config-file (expand-path *config-file-path*)))
+  "Load or reload the config file."
+  (load-lisp config-file :package (find-package :nyxt-user)))
+
```
Ambrevar, line 247: Do we need this command if the file is never supposed to be edited manually?

Ambrevar:
We also need to update the manual to explain about the new generated configurations, the files, the load order, the associated commands.

source/help.lisp
jmercouris: The term config.lisp is certainly most logical. It lives in `~/.config`. I had thought about calling it generated-config, but stuck with config.lisp. Unlike in emacs, there is no consequence to manually editing this file yourself.

source/help.lisp
jmercouris: I was thinking about this as well, that's a good idea!

source/start.lisp
jmercouris: I don't think it is useful. What does knowing that it is generated do for you?

source/start.lisp
jmercouris: That is an assumption that you have, that I do not have :-D

My reasoning was always that it is safe to edit manually as well.

jmercouris:
Manual information added.

source/start.lisp
Ambrevar: It can be safe, but is there a usecase for it?  If the user is going to write Lisp, they'd rather to it in init.lisp.  Exposing this command is inciting the user to edit the config file which is not by design.

source/help.lisp
Ambrevar: The problem is that a newcomer cannot tell the semantic difference between init.lisp and config.lisp.  We need names that make their respective purposes obvious.

source/start.lisp
jmercouris: I was planning on using it, but you make a good point!

source/help.lisp
jmercouris: I think the names are clear:

config.lisp: configuration
init.lisp: initialization

configuration = generated configuration
initialization = manually written initialization

I believe this is one of those cases where we will have to introduce this terminology that people will have to just learn. There is no obvious one-word distinction for these terms. In other words, we must invent them.

Ambrevar: Qutebrowser uses the term `autoconfig` for exactly that: https://qutebrowser.org/doc/help/configuring.html
What do you think?
Otherwise "config-autosave"?

If you really want to stick to config.lisp, we need to edit a few spots in the code that refer to init.lisp as the "configuration file".

source/manual.lisp
```diff
@@ -15,6 +15,16 @@ of Nyxt."))
     (:p "Nyxt is written in the Common Lisp programming language which offers a
 great perk: everything in the browser can be customized by the user, even while
 it's running!")
+    (:p "Nyxt provides a mechanism for new users unfamiliar with Lisp
+to customize Nyxt. To begin, start by invoking the
```
Ambrevar, line 19: "To begin" is redundant with "start", you can remove it.

source/manual.lisp
```diff
@@ -15,6 +15,16 @@ of Nyxt."))
...
+commands " (:code "describe-class") " or " (:code "describe-slot") ".
+When you find a value that you would like to change, you can press the
+button marked 'Configure' to change it. The settings will be applied
```
Ambrevar, line 22: "When you find a value that you would like to change" is not very informative.  We can shorten the sentence (remember we are trying to make the manual as short as possible ;p), e.g.

```
You can press the 'Configure' button to change the value of a setting.
```

source/manual.lisp
```diff
@@ -15,6 +15,16 @@ of Nyxt."))
...
+immediately and saved for future sessions.")
+    (:p "Settings created by Nyxt are stored in "
+        (:code (expand-path *config-file-path*)) ".")
+    (:p "Any settings can be overridden by users manually via "
```
Ambrevar, line 26: Superfluous "by users".

source/start.lisp
```diff
@@ -453,20 +478,27 @@ REPL examples:
...
+              (not (expand-path *config-file-path*)))
+    (load-lisp (expand-path *config-file-path*) :package (find-package :nyxt-user)))
   (unless (or (getf *options* :no-init)
               (not (expand-path *init-file-path*)))
     (load-lisp (expand-path *init-file-path*) :package (find-package :nyxt-user)))
   (load-or-eval :remote (getf *options* :remote)))
 
 (defun start-browser (free-args)
   "Load INIT-FILE if non-nil.
+Load CONFIG-FILE.
```
Ambrevar, line 491: This docstring is misleading.  We should say something like "First load the config-file, then the init-file, finally start the browser."

jmercouris: I just moved Load CONFIG-FILE above so that the order is as it is in the codebase.

source/help.lisp
jmercouris: autoconfig is a fine term.

PR #949, opened 2020-09-10
Fixed minor typos

aadcg:
Minor documentation review.

### DIFF
--- a/source/base-mode.lisp
+++ b/source/base-mode.lisp
@@ -19,7 +19,7 @@
                      "C-tab" 'switch-buffer-next
                      "C-pageup" 'switch-buffer-previous
                      "C-pagedown" 'switch-buffer-next
-                     "C-l" 'set-url-from-current-url
+                     "C-l" 'set-url
                      "M-l" 'set-url-new-buffer
                      "f5" 'reload-current-buffer
                      "C-r" 'reload-current-buffer
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -284,8 +284,6 @@ export GDK_SCALE=2
 export GDK_DPI_SCALE=0.5
 nyxt
 "))
-    (:p "Note that if you're running sway, you'll also need this
-    environment variable " (:code "export GDK_BACKEND=x11") ".")
     (:h3 "StumpWM mouse scroll")
     (:p "If the mouse scroll does not work for you, see the "
         (:a
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -284,6 +284,8 @@ export GDK_SCALE=2
 export GDK_DPI_SCALE=0.5
 nyxt
 "))
+    (:p "Note that if you're running sway, you'll also need this
+    environment variable " (:code "export GDK_BACKEND=x11") ".")
     (:h3 "StumpWM mouse scroll")
     (:p "If the mouse scroll does not work for you, see the "
         (:a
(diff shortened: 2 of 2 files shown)


aadcg:
The latest commit is to fix the fact that when a user starts nyxt with the default cua-mode, he's greeted with `UNBOUND: Load URL`.

aadcg:
Regarding the last commit, 10fbec6, the same could perhaps hold for other wm running on wayland but I haven't tested.

aadcg:
On 10fbec6, I thought I had found a way to get bigger font sizes on sway. Though fonts indeed scale at startup, nyxt doesn't run very well with the GDK backend set to x11 on sway.

I don't understand why the scaling doesn't apply when the backend is wayland...

With that said, this commit should be ignored.

source/manual.lisp
```diff
@@ -90,7 +90,7 @@ documentation.")
     (:p "See the " (:code "proxy-mode") " documentation.")
 
     (:h3 "Custom commands")
-    (:p "Creating your own invokable commands is similar to creating a Common
+    (:p "Creating your own invocable commands is similar to creating a Common
```
Ambrevar, line 93: With a `k` is correct: https://en.wiktionary.org/wiki/invokable

source/base-mode.lisp
```diff
@@ -19,7 +19,7 @@
                      "C-tab" 'switch-buffer-next
                      "C-pageup" 'switch-buffer-previous
                      "C-pagedown" 'switch-buffer-next
-                     "C-l" 'set-url-from-current-url
+                     "C-l" 'set-url
```
Ambrevar, line 22: @jmercouris Wasn't this set to `set-url-from-current-url` on purpose?  I can't recall.

Ambrevar: See 0bb2c1f46d517ca9ef080d0f881c13cbed8b7a7b

jmercouris: Indeed it was the behavior that new users would expect.

source/manual.lisp
aadcg: Ok, fine for me

aadcg: Is it possible to discard this commit from github?

source/base-mode.lisp
aadcg: Then I suggest changing mirroring these changes to emacs-mode and vim-mode and then show C-l as Load URL on the quickstart keys.

jmercouris: That's probably a good idea. I'm just thinking about whether a more advanced user would want to have to C-k the whole input every time they set-url.

Ambrevar: I find it more convenient to start with an empty input and press tab.

source/manual.lisp
Ambrevar: Yes, you can squash the commits locally, then force-push.
But I can also squash when merging on your behalf.

source/base-mode.lisp
jmercouris: Myself as well. Perhaps we can just bind set-url-from-current-url as well so it doesn't show up as undefined in Vi/Emacs mode.

Ambrevar: To what should we bind set-url-from-current-url?

Ambrevar:
Fixed the `even` typo in 188f74bd253f7e640d8b902cc581637b0ff958c5.
I've open #969 to continue the discussion on set-url vs. set-url-from-current-buffer.

PR #954, opened 2020-09-14
Synchronous minibuffer cleanup

Ambrevar:
I've tried to apply #928 to all of Nyxt.  The result is that it does not start :( 

The problem with starting command in a new thread is that if a command calls a renderer instruction (like making a buffer) it needs to switch to the renderer thread.  

One way to fix it is to wrap all FFI functions with `ffi-within-renderer-thread`.  But what about the FFI functions for which the command needs a return value?

We can juggle with threads but it seems hard to guarantee the absence of dead locks.

### DIFF
--- a/build-scripts/guix.scm
+++ b/build-scripts/guix.scm
@@ -133,6 +133,7 @@
      ;; See https://github.com/atlas-engineer/nyxt/issues/680.
      `(("alexandria" ,cl-alexandria)
        ("bordeaux-threads" ,cl-bordeaux-threads)
+       ("cl-chanl" ,cl-chanl)
        ("cl-containers" ,cl-containers)
        ("cl-css" ,cl-css)
        ("cl-json" ,cl-json)
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -72,6 +72,7 @@
                (:file "message")
                (:file "input")
                (:file "minibuffer")
+               (:file "minibuffer-prompt")
                (:file "minibuffer-mode")
                (:file "command-commands")
                (:file "recent-buffers")
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -5,6 +5,7 @@
   :serial t
   :depends-on (:alexandria
                :bordeaux-threads
+               :chanl
                :cl-css
                :cl-json
                :cl-markup
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -180,11 +180,10 @@ The rules are:
                  *prompt-on-mode-toggle*)
         (if-confirm ("Permanently ~:[disable~;enable~] ~a for this URL?"
                        enable-p (mode-name mode))
-          (with-result (url (read-from-minibuffer
-                             (make-minibuffer
-                              :input-prompt "URL:"
-                              :input-buffer (object-display (url (buffer mode)))
-                              :must-match-p nil)))
+                    (let ((url (prompt-minibuffer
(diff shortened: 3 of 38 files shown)


jmercouris:
I think the only way to fix it is to have ffi-within-renderer-thread sprinkled when necessary. The way I see it we have two options:

1. Awkward with-result handling of inputs
2. Awkward necessity of ffi-within-renderer-thread

I would prefer number 2. There is unfortunately no perfect solution (yet :-D), and we must make a trade.

jmercouris:
We cannot guarantee the absence of deadlocks UNLESS we perform no blocking operations on the main thread.

jmercouris:
The only resource that any thread could currently starve over would be the main thread.

Ambrevar:
But when is it "necessary" to use ffi-within-renderer-thread?  It's not
so clear.

The only place where it's clear is around all the ffi-* functions.  I
tried to do this in the latest commit, but it hangs the browser on
startup.

The problem is that we don't know where the ffi-* function is called
from.
If it's called from the command thread, it works.  If called from the
renderer thread, it blocks.

Ambrevar:
I've got an idea: the main issue here is that we don't know when the
caller is on the GTK thread or another thread.  If we knew, we could
decide whether to call ffi-within-renderer-thread or not (no need if
already on the GTK thread).

We can obtain the renderer thread with this code:

(defvar *renderer-thread* nil)
(ffi-within-renderer-thread *browser*
 (lambda ()
  (setf *renderer-thread* (bt:current-thread))))

(defun renderer-thread-p ()
  (eq *renderer-thread* (bt:current-thread)))

Then we update ffi-within-renderer-thread:

(defmethod ffi-within-renderer-thread ((browser gtk-browser) thunk)
  (declare (ignore browser))
  (if (renderer-thread-p)
      (funcall thunk)
      (gtk:within-gtk-thread
        (funcall thunk))))

@jmercouris: Do you think that would work?

Ambrevar:
Actually this is wrong:

> --8<---------------cut here---------------start------------->8---
> (ffi-within-renderer-thread *browser*
>  (lambda ()
>   (setf *renderer-thread* (bt:current-thread))))
> --8<---------------cut here---------------end--------------->8---

since we modify ffi-within-renderer-thread.  Instead we should

(gtk:within-gtk-thread
    (lambda ()
      (setf *renderer-thread* (bt:current-thread))))

jmercouris:
I don't see why that should not work. Though GUI thread semantics have bitten me in the past... :-D

Ambrevar:
Progress: I managed to start Nyxt and get the session restore prompt.
Every else fails after that, the web view is blank.

There is hope :)

Ambrevar:
Progress!  This branch works for most things: loading pages, new windows, minibuffer inputs, history, hinting, etc.

Left to do:

- Pasting hangs again.
- web-mode's nyxt:on-signal-load-finished is supposed to unzoom, but when it does it hangs the browser.  Not sure why.

@jmercouris Can you test again on macOS?  You might have to play with the ffi-initialize and finalize functions to set up the threading properly on macOS.

On GNU/Linux, I run the startup-function in a chanl thread, _not_ in the GTK thread.  This is because the startup function, like any command, can prompt the minibuffer which is a blocking operation, so it cannot be done from the GTK thread otherwise it would hang the browser.

jmercouris:
No luck, no window, I will play around with it.

Ambrevar:
Continuing in #977.

PR #955, opened 2020-09-14
Common settings

jmercouris:
This branch exposes some common settings for new users.

Not sure how to handle the setting of VI and Emacs, specifically:

```
(nyxt::configure-slot
                                      'default-modes
                                      'web-buffer
                                      :value (append '(vi-mode) %slot-default))
```

### DIFF
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -477,8 +477,8 @@ The version number is stored in the clipboard."
     (let* ((help-contents
              (markup:markup
               (:style (style help-buffer))
-              (:h1 "Getting started")
-              (:p (:b "Warning: ") "Nyxt is under active development. Feel free to "
+              (:h1 "Welcome to Nyxt " (:span :style "font-weight:normal" "☺"))
+              (:p (:b "Attention: ") "Nyxt is under active development. Feel free to "
                   (:a :href "https://github.com/atlas-engineer/nyxt/issues"
                       "report")
                   " bugs, instabilities or feature wishes.")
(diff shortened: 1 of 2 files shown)


jmercouris:
I decided to handle it by hard-coding the value. I figured it was the most robust solution.

source/help.lisp
```diff
@@ -216,16 +216,23 @@ A command is a special kind of function that can be called with
...
-  (with-result (input (read-from-minibuffer
-                       (make-minibuffer
-                        :input-prompt (format nil "Configure slot value ~a" slot))))
-    (echo "Slot ~a updated with value ~a." slot input)
-    (eval `(define-configuration ,class
-             ((,slot (read-from-string ,input)))))
-    (append-configuration `(define-configuration ,class
-                             ((,slot ,input))))))
+  (flet ((set-slot (slot class input)
+           (echo "Slot ~a updated with value ~a." slot input)
```
Ambrevar, line 222: `~s` instead of `~a` is clearer.

source/help.lisp
```diff
@@ -254,6 +261,58 @@ A command is a special kind of function that can be called with
...
+                    "Use Emacs"))
+            (:p (:a :class "button"
+                    :href (lisp-url `(nyxt::configure-slot
+                                      'default-modes
+                                      'web-buffer
+                                      :value ''(vi-normal-mode
+                                                certificate-exception-mode
+                                                web-mode
+                                                base-mode)))
+                    "Use vi"))
```
Ambrevar, line 302: "VI" (all caps)

source/help.lisp
```diff
@@ -254,6 +261,58 @@ A command is a special kind of function that can be called with
...
+            buffers.")
+            (:h2 "Keybinding style")
+            (:p (:a :class "button"
+                    :href (lisp-url `(nyxt::configure-slot
+                                      'default-modes
+                                      'web-buffer
+                                      :value ''(certificate-exception-mode
+                                                web-mode
+                                                base-mode)))
+                    "Use Standard (CUA)"))
```
Ambrevar, line 284: No uppercase here: "standard".
What about `Use CUA (default)`?
More generally, shouldn't we automatically mention the defaults?

source/help.lisp
jmercouris: I looked it up on Wikipedia, it is all lowercase

source/help.lisp
jmercouris: I was using the word "standard" as "default" here, you are right, default is more clear, I will change it!

source/help.lisp
jmercouris: I assume you mean `(echo "Slot ~a updated with value ~s." slot input)`?

source/help.lisp
Ambrevar: Oops!  Today I learned something!

source/help.lisp
Ambrevar: Yes!

PR #907, opened 2020-08-25
Import bookmarks from a html file

C-Entropy:

PR #960, opened 2020-09-21
Import bookmarks from an HTML file

### LINKED ISSUE #465: Import bookmarks from HTML files

kssytsrk:
Fixes #465 
Would probably close #907 too?

### DIFF
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -85,9 +85,9 @@ In particular, we ignore the protocol (e.g. HTTP or HTTPS does not matter)."
               (tag-description tag))
       (object-string tag)))
 
-(declaim (ftype (function (quri:uri &key (:title string) (:tags t)) t) bookmark-add))
+(declaim (ftype (function (quri:uri &key (:title string) (:date local-time:timestamp) (:tags t)) t) bookmark-add))
 (export-always 'bookmark-add)
-(defun bookmark-add (url &key title tags)
+(defun bookmark-add (url &key date title tags)
   (with-data-access bookmarks (bookmarks-path (current-buffer))
     (unless (or (url-empty-p url)
                 (string= "about:blank" (object-string url)))
@@ -419,30 +419,32 @@ rest in background buffers."
 (define-command import-bookmarks-from-html ()
   "Import bookmarks from an HTML file."
   (with-result (html-file (read-from-minibuffer
-			               (make-minibuffer
-                            :default-modes '(nyxt/file-manager-mode::file-manager-mode
+                           (make-minibuffer
+                            :default-modes '(nyxt/file-manager-mode:file-manager-mode
                                              minibuffer-mode)
-			                :input-prompt "Path to the HTML file: "
-                            :suggestion-function #'nyxt/file-manager-mode::open-file-from-directory-suggestion-filter
+                            :input-prompt "Path to the HTML file: "
+                            ;; the suggestion filter allows non-html files right
+                            ;; now, as with :must-match-p set to nil the user
+                            ;; would still be able to select a non-html file;
+                            ;; this should be fixed (and :must-match-p nil -
+                            ;; removed) once the file-manager has better navigation
+                            ;; support
+                            :suggestion-function #'nyxt/file-manager-mode:open-file-from-directory-suggestion-filter
                             :must-match-p nil)))
-    (unless (and (probe-file html-file)
-		         (equal (pathname-type html-file) "html"))
-      (import-bookmarks-from-html))
-    (with-open-file (in-html html-file
-   			                 :external-format :utf-8)
-      (let ((a-tags (plump:get-elements-by-tag-name (plump:parse in-html) "a")))
-        (loop for a-tag in a-tags
-           do (let ((extracted
-                     (list (plump:attribute a-tag "href")
-                           (plump:render-text a-tag)
-                           (plump:attribute a-tag "tags"))))
-                (apply #'(lambda (url title tags)
-                            (let ((url-uri (quri:uri url)))
-   		                      (when (string= (quri:uri-scheme url-uri)
-                                             "http"
-                                             :end1 4)
-                                (bookmark-add url-uri
-   				                              :title title
-   				                              :tags (unless (null tags)
-                                                      (str:split "," tags))))))
-                        extracted)))))))
+    (if (and (probe-file html-file)
+             (equal (pathname-type html-file) "html"))
+        (with-open-file (in-html html-file :external-format :utf-8)
+          (let ((a-tags (plump:get-elements-by-tag-name (plump:parse in-html) "a")))
+            (dolist (a-tag a-tags)
+              (let* ((url (plump:attribute a-tag "href"))
+                     (title (plump:render-text a-tag))
+                     (date (plump:attribute a-tag "add_date"))
+                     (tags (plump:attribute a-tag "tags"))
+                     (url-uri (quri:uri url)))
+                (when (str:starts-with? "http" (quri:uri-scheme url-uri))
+                  (bookmark-add url-uri
+                                :title title
+                                :date (local-time:unix-to-timestamp (parse-integer date))
+                                :tags (when tags
+                                        (str:split "," tags))))))))
+        (echo "The file doesn't exist or is not an HTML file."))))
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -422,7 +422,7 @@ rest in background buffers."
                            (make-minibuffer
                             :default-modes '(nyxt/file-manager-mode:file-manager-mode
                                              minibuffer-mode)
-                            :input-prompt "Path to the HTML file: "
+                            :input-prompt "Path to the HTML file"
                             ;; the suggestion filter allows non-html files right
                             ;; now, as with :must-match-p set to nil the user
                             ;; would still be able to select a non-html file;
@@ -431,7 +431,7 @@ rest in background buffers."
                             ;; support
                             :suggestion-function #'nyxt/file-manager-mode:open-file-from-directory-suggestion-filter
                             :must-match-p nil)))
-    (if (and (probe-file html-file)
+    (if (and (uiop:file-exists-p html-file)
              (equal (pathname-type html-file) "html"))
         (with-open-file (in-html html-file :external-format :utf-8)
           (let ((a-tags (plump:get-elements-by-tag-name (plump:parse in-html) "a")))
@@ -444,7 +444,7 @@ rest in background buffers."
                 (when (str:starts-with? "http" (quri:uri-scheme url-uri))
                   (bookmark-add url-uri
                                 :title title
-                                :date (local-time:unix-to-timestamp (parse-integer date))
+                                :date (ignore-errors (local-time:unix-to-timestamp (parse-integer date)))
                                 :tags (when tags
                                         (str:split "," tags))))))))
         (echo "The file doesn't exist or is not an HTML file."))))
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -415,3 +415,34 @@ rest in background buffers."
     (error (c)
       (echo-warning "Failed to load bookmarks from ~s: ~a"
                     (expand-path path) c))))
+
+(define-command import-bookmarks-from-html ()
+  "Import bookmarks from an HTML file."
+  (with-result (html-file (read-from-minibuffer
+			               (make-minibuffer
+                            :default-modes '(nyxt/file-manager-mode::file-manager-mode
+                                             minibuffer-mode)
+			                :input-prompt "Path to the HTML file: "
+                            :suggestion-function #'nyxt/file-manager-mode::open-file-from-directory-suggestion-filter
+                            :must-match-p nil)))
+    (unless (and (probe-file html-file)
+		         (equal (pathname-type html-file) "html"))
+      (import-bookmarks-from-html))
+    (with-open-file (in-html html-file
+   			                 :external-format :utf-8)
+      (let ((a-tags (plump:get-elements-by-tag-name (plump:parse in-html) "a")))
+        (loop for a-tag in a-tags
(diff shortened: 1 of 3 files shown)


source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
...
       (echo-warning "Failed to load bookmarks from ~s: ~a"
                     (expand-path path) c))))
+
+(define-command import-bookmarks-from-html ()
+  "Import bookmarks from an HTML file."
+  (with-result (html-file (read-from-minibuffer
+			               (make-minibuffer
+                            :default-modes '(nyxt/file-manager-mode::file-manager-mode
+                                             minibuffer-mode)
+			                :input-prompt "Path to the HTML file: "
```
Ambrevar, line 425: Make sure to use spaces for indentation, not tabs.

source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
     (error (c)
       (echo-warning "Failed to load bookmarks from ~s: ~a"
                     (expand-path path) c))))
+
+(define-command import-bookmarks-from-html ()
+  "Import bookmarks from an HTML file."
+  (with-result (html-file (read-from-minibuffer
+			               (make-minibuffer
+                            :default-modes '(nyxt/file-manager-mode::file-manager-mode
```
Ambrevar, line 423: Replace `::` with `:` since the symbol is exported.

source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
...
+                                             minibuffer-mode)
+			                :input-prompt "Path to the HTML file: "
+                            :suggestion-function #'nyxt/file-manager-mode::open-file-from-directory-suggestion-filter
+                            :must-match-p nil)))
+    (unless (and (probe-file html-file)
+		         (equal (pathname-type html-file) "html"))
+      (import-bookmarks-from-html))
```
Ambrevar, line 430: Are you looping when the file does not have an html extension?
If so, I suggest to define a new filter that only display .html files, then you can remove this `unless` expression.

source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
...
+    (with-open-file (in-html html-file
+   			                 :external-format :utf-8)
```
Ambrevar, line 432: You can leave this on the same line.

source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
...
+      (let ((a-tags (plump:get-elements-by-tag-name (plump:parse in-html) "a")))
+        (loop for a-tag in a-tags
```
Ambrevar, line 434: Replace this `loop` with `dolist`.

source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
...
+           do (let ((extracted
+                     (list (plump:attribute a-tag "href")
+                           (plump:render-text a-tag)
+                           (plump:attribute a-tag "tags"))))
+                (apply #'(lambda (url title tags)
```
Ambrevar, line 439: Style: You don't need the `#'`.  If you prefer it for clarity, no problem, you can leave it.

source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
...
+                            (let ((url-uri (quri:uri url)))
+   		                      (when (string= (quri:uri-scheme url-uri)
+                                             "http"
+                                             :end1 4)
```
Ambrevar, line 443: Replace this `sting=` with `str:starts-with?`.

source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
...
+                                (bookmark-add url-uri
+   				                              :title title
+   				                              :tags (unless (null tags)
```
Ambrevar, line 446: Replace `(unless (null tags) ...)` by `(when tags ...)`.

source/bookmark.lisp
```diff
@@ -415,3 +415,34 @@ rest in background buffers."
...
+                                                      (str:split "," tags))))))
+                        extracted)))))))
```
Ambrevar, line 448: The `extracted` and `apply` are not needed.  Instead, you can let-bind `url`, `title` and `tags` to the respective plump calls, then run the body of the lambda.  Does that make sense?

Ambrevar:
Great, thanks for your effort!

Please make sure to indent with spaces, not tabs.
Can you please try to import the bookmark `:date` like #907 does?

source/bookmark.lisp
kssytsrk: Wouldn't this just change the suggestions? I have :must-match-p set to nil due to (sorta) brokenness of the suggestion function (won't let me choose a file from another directory) so this wouldn't do anything, the user would still be able to select a file with an extension that isn't .html. 
I left a loop in this place because #907 had it that way and it didn't seem to be a problem. Should I replace it with an error message or something like that?

Ambrevar: An error message would be better indeed, otherwise it's confusing.

Also add a comment to that explains why the suggestion filter allows non-HTML files for now, and that we should update it once the file manager has better navigation support.

kssytsrk:
I think the code of the ```BOOKMARK-ADD``` function has changed since #907 pull-request, now it seems it doesn't have the date argument, so I have nowhere to pass it. Are there any other ways of adding a bookmark that would allow specifying the date?

source/bookmark.lisp
jmercouris: why?

source/bookmark.lisp
kssytsrk: May be a really basic question, but how do I "quit" the function after showing the error message (I suppose I echo it? or should I log:info to the console?)? ```(return-from import-bookmarks-from-html nil)``` seems to not work, gives me an ```attempt to RETURN-FROM a block or GO to a tag that no longer exists```. Any pointers?

jmercouris: You can use return-from, though in this case, it is not necessary. That is more typically used to break out of loops. So what you would do then is wrap everything in a conditional, `(when )` or `(if )`.

and if that conditional is not satisfied, don't execute any code. For example

```
(if file-is-html
  ;; true case
  do-something
  ;; false case
)
```

becuase you are only interested in the true case, you can just use a `when`. So, 

```
(when file-is-html
  ;; true
  do-something)
```

then, your function will automatically "exit' when the file is not HTML. Does that help?

source/bookmark.lisp
Ambrevar: In this trivial case, `dolist` saves one line :)

Ambrevar:
If you look closely, the #907 did add a `:date` &key argument to the bookmark-add function.  You can reuse that.

source/bookmark.lisp
kssytsrk: I see, just thought that there was some other "cleaner" way to do it. That helps, thanks for pointing it out

kssytsrk:
I suppose I fixed everything that needed a fix, so please check it once more.

source/bookmark.lisp
```diff
@@ -415,3 +415,36 @@ rest in background buffers."
...
       (echo-warning "Failed to load bookmarks from ~s: ~a"
                     (expand-path path) c))))
+
+(define-command import-bookmarks-from-html ()
+  "Import bookmarks from an HTML file."
+  (with-result (html-file (read-from-minibuffer
+                           (make-minibuffer
+                            :default-modes '(nyxt/file-manager-mode:file-manager-mode
+                                             minibuffer-mode)
+                            :input-prompt "Path to the HTML file: "
```
Ambrevar, line 425: Remove the trailing `: `.

source/bookmark.lisp
```diff
@@ -415,3 +415,36 @@ rest in background buffers."
...
+                            ;; the suggestion filter allows non-html files right
+                            ;; now, as with :must-match-p set to nil the user
+                            ;; would still be able to select a non-html file;
+                            ;; this should be fixed (and :must-match-p nil -
+                            ;; removed) once the file-manager has better navigation
+                            ;; support
+                            :suggestion-function #'nyxt/file-manager-mode:open-file-from-directory-suggestion-filter
+                            :must-match-p nil)))
+    (if (and (probe-file html-file)
```
Ambrevar, line 434: `probe-file` will raise an error if the argument contains a wild card (e.g. `*`).  Use `uiop:file-exists-p` instead.

source/bookmark.lisp
```diff
@@ -415,3 +415,36 @@ rest in background buffers."
...
+            (dolist (a-tag a-tags)
+              (let* ((url (plump:attribute a-tag "href"))
+                     (title (plump:render-text a-tag))
+                     (date (plump:attribute a-tag "add_date"))
+                     (tags (plump:attribute a-tag "tags"))
+                     (url-uri (quri:uri url)))
+                (when (str:starts-with? "http" (quri:uri-scheme url-uri))
+                  (bookmark-add url-uri
+                                :title title
+                                :date (local-time:unix-to-timestamp (parse-integer date))
```
Ambrevar, line 447: Surroung the `unix-to-timestamp` call with `ignore-errors` since an error may occur if `date` is not well formatted.

Ambrevar:
Just a few more nits and we are good to merge!

source/bookmark.lisp
jmercouris: No problem!

kssytsrk:
Did it. Also, could you please check if it compiles for you? It errored out (about not being able to find the symbols in nyxt/file-manager-mode) for me a couple times, but went away after a few Emacs restarts. Still, a bit scared that something like that might happen to someone else.

jmercouris:
I'll give it a try, thank you for your persistence :-)

jmercouris:
OK, I know what is going on here. You've done the following (per Pierre's recommendation):

` :default-modes '(nyxt/file-manager-mode:file-manager-mode`

which is all fine and dandy IF, that symbol is declared in that package. BUT, it is not. If you remember, we moved the definition into package.lisp. What we did not do is say that we are exporting that symbol. Typically, this exportation of that symbol is handled by our macro. So you have two options:

1. Use "::" again
2. add a clause like `(:export :file-manager-mode)` into package.lisp

kssytsrk:
Thanks for the detailed explanation! I get it now, added (:export :file-manager-mode).

...or should I have exported the suggestion filter function, too?

jmercouris:
depends, are there warnings if you don't export that symbol? If not, no problem.

jmercouris:
I just checked, it seems to still have problems, it probably needs to be exported as well.

kssytsrk:
seems to compile for me now...

jmercouris:
Compiles, a couple of warnings, if you like, I can take care of them and merge this. OR I can tell you what they are and how to fix them, either is OK with me :-)

kssytsrk:
Please tell me about the warnings, I'd love to fix them!

jmercouris:
OK! Then, here are the warnings:

On line 90 of bookmark.lisp:  "The variable DATE is defined but never used."

You can put a `(declare (ignore date))` within the body of `bookmark-add` since it is not currently used.

Then, on line 447 of bookmark.lisp: "Result is a (VALUES NULL T &OPTIONAL), not a LOCAL-TIME:TIMESTAMP.".

The problem is that `:date` is expected to be a `local-time:timestamp`, but since we've now wrapped everything in `ignore-errors`, we are not guaranteed to get this type of result. To fix this, we'll have to update the signature of `bookmark-add`.

That said, it seems to me that we should just remove the signature altogether. If we accept literally anything for the `:date`, then the signature has no value.

Ambrevar:
> On line 90 of bookmark.lisp:  "The variable DATE is defined but never used."
>
> You can put a `(declare (ignore date))` within the body of `bookmark-add` since it is not currently used.

bookmark-add should set the date of the created bookmark, otherwise
passing the date is pointless :)

> Then, on line 447 of bookmark.lisp: "Result is a (VALUES NULL T &OPTIONAL), not a LOCAL-TIME:TIMESTAMP.".
>
> The problem is that `:date` is expected to be a `local-time:timestamp`, but since we've now wrapped everything in `ignore-errors`, we are not guaranteed to get this type of result. To fix this, we'll have to update the signature of `bookmark-add`.

Removing the signature of bookmark-add means that all calls of
bookmark-add will now be type-less, which would be too bad.

Better: Accept nil as date (means "no date") by changing it's type in
the bookmark-add signature to

(or local-time:timestamp null)

jmercouris:
Well yes, that's what it should do, but that is outside of the scope of this pull request. I believe that the deficiencies of the bookmark-add function are not related to the facility that properly parses and imports the bookmarks from an html file.

As per the signature of the declaim, I agree with your assessment!

Ambrevar:

> Well yes, that's what it should do, but that is outside of the scope of this pull request. I believe that the deficiencies of the bookmark-add function are not related to the facility that properly parses and imports the bookmarks from an html file.

I don't understand what you mean.

This pull request is for importing HTML bookmarks.  Those bookmarks
often contain a date, which we want to import as well.

Finally, it's trivial to specify the date in the bookmark-add function,
just add

(when date
 (setf (date entry) date))

jmercouris:
I am trying to say that this pull request is about importing book marks from an HTML file. It is not about fixing the `bookmark-add` function.

I don't think it is a good idea to expect a new contributor/user to understand these other functions. I think we should make it easy for new users to get started and make a short, but valuable contribution to Nyxt. Afterwards, we can fix up whatever needs to be done. In this case, fixing `bookmark-add`.

kssytsrk:
It doesn't seem to signal any warnings now for me anymore, how's this for you?

jmercouris:
It looks good! Thank you! I will merge this on Monday, if it is not gotten to before then.

Ambrevar:
Ping :)

jmercouris:
Vielen dank!!!! Thank you for your persistence!

PR #965, opened 2020-09-24
check SBCL version

phil-hands:
I can imagine that one could make this slightly neater using make rather than shell for the test and conditional, but this does work as it is.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,11 @@ build-deps: quicklisp-extra-libs
 
 .PHONY: deps
 deps:
+	SBCL_VER="$$(sbcl --version)" ; \
+	if [ "sbcl" = "$(LISP)" ] && expr "$$SBCL_VER" : "SBCL 1" > /dev/null ; then \
+		printf "\nERROR: SBCL version (%s) too old -- needs version 2 or later\n\n" "$$SBCL_VER" >&2 ; \
+		false ; \
+	fi
 	$(NYXT_INTERNAL_QUICKLISP) && $(MAKE) build-deps || true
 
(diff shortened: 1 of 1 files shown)


jmercouris:
Thank you! It looks/works good on my system. I would like @Ambrevar to confirm it works on his system before merging.

Ambrevar:
As per the developer's readme, we support SBCL 1.4.2.
Some reported that it does not work with SBCL <=2.0.0, I wonder if this can be confirmed.

Also intead of parsing the SBCL output, we could run

```sh
sbcl --no-userinit --non-interactive --eval '(assert-version->= 1 4 2)'
```

jmercouris:
Phil just had problems on that version, so it doesn't support something we are doing.

Ambrevar:
Which version? 1.4.2?

jmercouris:
1.4.16, maybe 1.4.2 is still OK.

phil-hands:
I was running the version that is in Debian stable, the changelog for which is here:

  https://metadata.ftp-master.debian.org/changelogs//main/s/sbcl/sbcl_1.4.16-2_changelog

Ambrevar:
Thanks.  I'll try to fix Nyxt for 1.4.16 when I find the time.

jmercouris:
In the meantime shall we not apply this patch to prevent people from compiling with a version which does not work?

Ambrevar:
No, let's use the line I suggested instead, it's simpler, shorter, more reliable.

Ambrevar:
Done in a23a2124e18ed71168398028fb4284b06d16a144.
Thanks for reporting!

PR #970, opened 2020-09-25
Add reduce-tracking-mode.

### LINKED ISSUE #941: Disable Accept Language header
Nyxt sends out the Accept-Language header, presumably based on the locale. This is a leak of private information and aids browser fingerprinting. Is there a way to disable the Accept-Language header? Perhaps the Accept-Language header should even be disabled by default, and enabled only if the user explicitly does so in their configuration.

Ambrevar:
Closes #941.

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -104,6 +104,7 @@
                (:file "download-mode")
                (:file "vcs-mode")
                (:file "force-https-mode")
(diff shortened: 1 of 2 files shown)


jmercouris:
Looks ready to me! Thanks for the work!

PR #977, opened 2020-09-28
Synchronous minibuffer working

Ambrevar:


Ambrevar:
Everything is implemented and everything seems to work.

Left to review:

- lisp-repl
- meta search
- macOS
- Make sure all the on-signal methods are running on the right thread.

Do we still need `define-parenscript` though?  Maybe ffi-evaluate-javascript is enough?

@jmercouris Can you test this again?

Ambrevar:
Bugs:

- Sometimes the keys are misunderstood.  For instance pressing C-v then another key like C-n may paste text twice.

- In one occurrence the minibuffer became black; I had to restart the browser.

jmercouris:
I'm not understanding why you feel we do not need define-parenscript? Because it is now simpler to invoke parenscript synchronously? or?

I think there is still a case for being able to easily wrap parenscript in a function call that the user can make. Of course the macro can change substantially, but the macro itself is still quite useful, no?

Ambrevar:
Yes, we can leave it for now, I'm just thinking out loud.

jmercouris:
My performance issues have been resolved. I have no idea why. I compiled with my SBCLRC disabled. Perhaps my settings were introducing just enough latency for me to notice, no longer noticable.

jmercouris:
Bug: M-i in element hint mode no longer triggers the transparency.

Ambrevar:
That's great news!

> Bug: M-i in element hint mode no longer triggers the transparency.

I'm on it.

Ambrevar:
Works with CCL.

Ambrevar:
> Bug: M-i in element hint mode no longer triggers the transparency.

Fixed.

Ambrevar:
@jmercouris: meta-search is broken because it relies on the old
minibuffer style.  Do you see a way to fix it?

Otherwise I suggest we wait for
the minibuffer library with multi-source support.

Ambrevar:
Bug: search-buffers has no selections.

jmercouris:
> Works with CCL.

Yay! :-)

jmercouris:
> @jmercouris: meta-search is broken because it relies on the old minibuffer style. Do you see a way to fix it? Otherwise I suggest we wait for the minibuffer library with multi-source support.

I will investigate!

jmercouris:
I've removed it for now. I see what you mean about there being no point in working with it until the minibuffer rewrite. It is the right decision.

Shall we begin architecting the new minibuffer design in a new issue?

Ambrevar:

> Shall we begin architecting the new minibuffer design in a new issue?

Sure!  We can collect all the issues with the `minibuffer' tag.

Ambrevar:
Bug: After some time, the minibuffer takes about a second to pop up.

Ambrevar:
I've fixed the binding issues.  The recipe was:

- Turn on VI normal mode.
- Click on a link.
- After the page has loaded, the next key chord will be prefixed with
  "button1" and thus invalidate the binding.

The only severe issue that's left as far as I can tell is the minibuffer
which (randomly?) slows down after a while.

jmercouris:
Perhaps the threads are not being cleaned up?

Ambrevar:

> Perhaps the threads are not being cleaned up?

Yes, that's what I'm starting to think.  Indeed, when we call
`channel:recv` the channel never receives anything, the calling thread
will never be terminated.  But sometimes we can't be sure in advance
whether the other end of the channel will ever send anything.

Any idea?

Ambrevar:
Could also be a red herring.

jmercouris:
This is starting to feel like the halting problem. My only reasonable conclusion would be to use a thread pool, and kill the oldest threads running. Maybe we can limit it to something like 512 threads. If you are still running by the time 512 threads have been created, well, I'm sorry, but goodbye! :-D

jmercouris:
What do you think?

Ambrevar:
Chanl already has a thread pool, maybe there is something that can be
done there.

But what about this: command threads are usually run in a user context,
i.e. upon key pressed or execute-command calls.  If the thread hasn't
returned upon the next user interaction, we could kill it at this point.

All that said, let's first confirm where the issue comes from.  First of
all: can you reproduce?

Ambrevar:
Bug: Sometimes on startup the window name will be SBCL for a little
while before everything loads.

Some other times, the mode list in the status view won't be populated.

It seems that out startup logic is too parallel and the sequence of
things a bit out of control.

jmercouris:
Is the minibuffer slowdown caused by the amount of times the minibuffer appears, OR how long the program has been running? OR some other factor?

jmercouris:
Well, here's a bit of an issue.... I keep opening the minibuffer and closing it, not doing anything with the command. I did "open link hint" and then closed the minibuffer. I also held down "M-i" to toggle the transparency of the link hints over time (triggering hundreds of parenscript calls/threads)

here's my result:

```
(length (bt:all-threads)) -> 195
```

I think it is fair to say this would be a reasonable cause for slow down.

jmercouris:
I do see: `chanl:%thread-pool-soft-limit`

jmercouris:
here's a comment from the source code with regards to the thread pool limit :-D

```
(soft-limit :accessor pool-soft-limit :initform 1000) ; this seems like a sane-ish default
```

jmercouris:
I think if we were to set it to 32 or 64 that would be reasonable. If the user has queued up 32 actions in a row and your oldest result has not yet been resolved... well, we have a bit of a problemchen.

Ambrevar:
Thanks for investigating.  I'll try that.
Did you notice any slow-down yourself?

jmercouris:
Yes, I noticed a slow down after toggling the minibuffer many times. It is certainly related to the amount of live threads.

Ambrevar:
See commit 
0e4242ac9027c7edd0edf7d00db3b4858fed2ee6 which raises a condition on
cancel.  This seems simpler.  Thought?

jmercouris:
Would that cancel all active minibuffers if a single one is cancelled?

Ambrevar:
I think I found an elegant fix: when killing the minibuffer, send nil
to the channel, this will terminated the thread.

I've tried it for 5 minutes and the number of threads remained stable.

See 28c600bc1c7320815237227996002303a8df5eaa.

The issue now is that exiting the minibuffer will make prompt-minibuffer
return nil, and many commands must be updated to make sure they don't
continue with a nil value (see the commit).

An alternative would be to raise a condition and to catch this condition
in `run` and `run-async`.  This way we leave all the commands as they
are, which is also neater for the user since they don't need to always
check prompt-minibuffer return value against nil.

What do you think?

jmercouris:
Would there be a way to guard commands against nil with a macro instead of updating every command?

Ambrevar:
> Would that cancel all active minibuffers if a single one is cancelled?

No, it works as it should.

Ambrevar:
> Would there be a way to guard commands against nil with a macro instead of updating every command?

I don't think a macro would do.
You don't like my condition approach?

jmercouris:
The approach is fine, I was just wondering if there were any side effects we didn't think of

jmercouris:
I guess it is a perfectly valid usage of the signal and condition system, it is not just for errors :-D

jmercouris:
After some testing I am still able to increase the amount of threads slowly over time. I am not sure how I am doing this. I did too many discrete actions to be sure.

Ambrevar:
`evaluate` might be the culprit.  Try these actions.

jmercouris:
Which actions?

Ambrevar:
Evaluate anything in the REPL.

jmercouris:
Screen Shot 2020-10-01 at 15 50 24

it does not appear so, unfortunately :-\

Ambrevar:
In commit 4294f23512d1c72c03c0080b9cb5f8424840e93a I've implemented a
way to interrupt minibuffer without making NIL special (which would be a
problem if "NIL" was a valid selection).

Ambrevar:
Let's use this branch for a couple of days.  If we notice the number of
threads is steadily rising, I'll add a trace that should identify which
thread started but didn't exit.  Hopefully that will help.

Ambrevar:
Please copy-paste the content, it's easier (and lighter) ;)

Ambrevar:
Good news!  I've fixed the startup issue in
30dda771ee7ec2d33b0e0bbe1ed3d347a288d783.

This possibly fixes other dangling threads.

jmercouris:
> Good news! I've fixed the startup issue in [30dda77](https://github.com/atlas-engineer/nyxt/commit/30dda771ee7ec2d33b0e0bbe1ed3d347a288d783). This possibly fixes other dangling threads.

Nice! I'll replace my daily driver with this one for the next week then :-D

jmercouris:
Well, after I rebase master on it...

Ambrevar:
The last bug that remains, it seems, is search-buffer.
I can't figure it out: query-buffer always returns nil.  Any idea?

Any tip on how to debug Javascript?

jmercouris:
If you want to debug Javascript M-x `open-inspector`, you can then use JS to log to the console as you wish!

Ambrevar:
I understand part of the problem with search-buffer, the story is in
this comment from match-suggestion-function:

  ;; return NIL, the suggestions will be updated asynchronously by the
  ;; callback from query-buffer

Now that we are synchronous, this can't work indeed.
If I return the suggestions in match-suggestion-function, then I get the
suggestions but only when a delete a character.  For some reason, the
suggestions are nil when I insert a character.

Ambrevar:
Merged as 94373058733a14c3b837bf668c8699121d00b7ff.

jmercouris:
Woohoo!!!

jmercouris:
Congratulations on this great undertaking monsieur!

Ambrevar:
And thanks for testing!

PR #979, opened 2020-09-30
System interface

jmercouris:
This provides a simple interface to Quicklisp

You can:

1. add distributions
2. load a system
3. list all systems on all distributions

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -83,6 +83,7 @@
                (:file "external-editor")
                (:file "url-group")
                (:file "minibuffer-composite")
+               (:file "system")
                ;; Core Modes
                (:file "buffer-listing-mode")
                (:file "base-mode")
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -83,7 +83,7 @@
                (:file "external-editor")
                (:file "url-group")
                (:file "minibuffer-composite")
-               (:file "system")
+               (:file "lisp-system")
                ;; Core Modes
                (:file "buffer-listing-mode")
                (:file "base-mode")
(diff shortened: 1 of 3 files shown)


nyxt.asd
```diff
@@ -83,6 +83,7 @@
                (:file "external-editor")
                (:file "url-group")
                (:file "minibuffer-composite")
+               (:file "system")
```
Ambrevar, line 86: "system" is too vague I think.  What about "lisp-system"?  Or "package-manager"?
If we are going to have both a package manager for Nyxt extensions and for system packages, we need a good naming conventions to disambiguate the 2.

Ambrevar: Add a #+quicklisp feature macro since this depends on Quicklisp (which is not included in the Guix build for instance).

Ambrevar:
This depends on Quicklisp.  What about users who don't have Quicklisp?
Should we offer to install Quicklisp for them?

nyxt.asd
jmercouris: lisp-system it is! :-)

jmercouris:
For users who don't have quicklisp... hm, maybe just include quicklisp in the Guix pack?

Ambrevar:
quicklisp will probably never be accepted as a Guix package for a few
reasons, the most obvious one being that it's redundant with Guix.

Guix users don't need Quicklisp and if we provide an OS package manager
then all Guix packages, including the Common Lisp ones, will be
available.

But the problem is not limited to Guix: users may compile Nyxt without
Quicklisp using the means they want, we can't predict that.  So if we
want to provide a lisp-system for everyone, we need a way to include
Quicklisp.

When listing systems, can we display whether the system is installed or not?

jmercouris:
I don't think we can display whether it is installed, I think we might be able to display if it is loaded, I will check.

PR #985, opened 2020-10-05
Merge present-current-window and unpresent-current-window into toggle…

Ambrevar:
…-toolbars.

Replace ffi-window-set-status-buffer-height and
ffi-window-set-message-buffer-height with ffi-window-toggle-status-view and
ffi-window-toggle-message-view respectively.


jmercouris:
I have a proposal that we can get and set the height instead of just toggle it. Since we can already set the height, we can just add a get for the height, and we are good to go.

Why?: This gives us more control.

PR #986, opened 2020-10-05
Hide toolbars

jmercouris:
Expose FFI methods to hide toolbars.


source/window.lisp
```diff
@@ -152,12 +152,19 @@ The handlers take the window as argument."))
   "Unfullscreen WINDOW, or the currently active window if unspecified."
   (ffi-window-unfullscreen window))
 
-(define-command present-current-window (&optional (window (current-window)))
+(define-command toggle-toolbar-visibility (&optional (window (current-window)))
+  "Toggle the visibility of the message and status buffer areas."
+  (if (and (zerop (ffi-window-get-status-buffer-height window))
+           (zerop (ffi-window-get-message-buffer-height window)))
```
Ambrevar, line 158: Nit: 
```
(= 0
   (ffi-window-get-status-buffer-height window)
   (ffi-window-get-message-buffer-height window))
```

(No need for the `and` then.)

Ambrevar: `toogle-toolbars` is a shorter name.

jmercouris: OK I change to toggle-toolbars!

Ambrevar:
Other than that, I agree with your statement, setting the height gives us more flexibility.

What about getting and setting the size (width+height) instead with the same (renamed) FFI functions?  This would be even more generic for those would want vertical toolbars.

jmercouris:
It would be, but it is probably a bit premature since they cannot actually change the layout of the window yet. We have some hard-coded assumptions about what goes where. Therefore, I think we should do it later. Thoughts?

Ambrevar:
Sure, thanks!

Issue #935, opened 2020-09-07
Remove *default-proxy*?


Ambrevar:
Does not seem it's useful anymore now that it's so easy to customize the `proxy` slot of `proxy-mode`.

Ambrevar:
Similarly we should probably move the `auto-mode:*prompt-on-mode-toggle*` `auto-mode:*non-rememberable-modes*` globals to auto-mode slots.  @aartaka Thoughts on this?

jmercouris:
@aartaka ?

aartaka:
Sorry, I somewhy disabled notifications for mentions xD 

It sounds good to me. Given that the only use-case for modifying these auto-mode globals is in init file, it won't hurt anyone in runtime, I believe.

Ambrevar:
I've moved *default-proxy* to a slot in 
6582d615d484037cee3c360c6d8ae25ba0492c71.

However I haven't move the globals in auto-mode because they are
semantically truly
global, not mode-local.

For instance `*non-rememberable-modes*` is used in
`non-rememberable-mode-p` which is used in `rememberable-of` which is
used in the `enable-matching-modes`...  There is no auto-mode all the
way.

Regarding `*prompt-on-mode-toggle*`, it's dynamically bound at some
point.  It's possible to simulate dynamic slot bindings with
unwind-protect, but that's not very convenient.  Maybe there is a
generic, clean way to do this.

I can see two solutions:

- Turn `*non-rememberable-modes*` and `*prompt-on-mode-toggle*` into a
  user class with a single slot.  This gives us graphical configuration
  and access to the original value.

- Or add a `current-auto-mode` function which returns the auto-mode in
  context.

  This function could be implemented as (find-submode (current-buffer) 'auto-mode).

Thoughts?

aartaka:
> However I haven't move the globals in auto-mode because they are semantically truly global, not mode-local. For instance `*non-rememberable-modes*` is used in `non-rememberable-mode-p` which is used in `rememberable-of` which is used in the `enable-matching-modes`... There is no auto-mode all the way. 

That's true, but both `non-rememberable-mode-p` and `rememberable-of` are one-liners and can be easily rewritten/`flet`-bound (latter appeals more to me, because domain of application is quite restricted for these functions) to use `auto-mode` instances.

[...]

PR #987, opened 2020-10-05
Move auto-mode global variables (*prompt-on-mode-toggle*, *non-rememberable-modes*) to auto-mode slots

aartaka:
This moves the global variables of `auto-mode` to the `auto-mode` slots, to integrate it into the current configuration system.

### Caveats

To make configuration work, I need to put 
``` lisp
(define-configuration nyxt/auto-mode:auto-mode
  ((nyxt/auto-mode:prompt-on-mode-toggle t)))
```
instead of 
``` lisp
(define-configuration nyxt/auto-mode:auto-mode
  ((prompt-on-mode-toggle t)))
```
in my init file. Is it intended? `proxy-mode` doesn't require one to reference package explicitly, but there seem to be little difference between `proxy-mode` and `auto-mode` in terms of definition and exporting.

### Testing 
I've ran Nyxt for half an hour and tested different typical `auto-mode` use scenarios. It seems to behave the same as the previous version.

Let me know what you think of it! (By the way, should've it been a Pull Request? Or is it a small change that I could've pushed without review?)

Closes: #935 (?)

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -3,23 +3,9 @@
 
 (uiop:define-package :nyxt/auto-mode
   (:use :common-lisp :nyxt)
-  (:import-from #:serapeum #:export-always)
   (:import-from #:class-star #:define-class)
   (:documentation "Mode for automatic URL-based mode toggling."))
 (in-package :nyxt/auto-mode)
-
-(export-always '*prompt-on-mode-toggle*)
-(defvar *prompt-on-mode-toggle* nil
-  "Whether user will be asked about adding the mode to included/excluded modes
-in the auto-mode rules on mode activation/deactivation.")
-
-(export-always '*non-rememberable-modes*)
-(defvar *non-rememberable-modes*
-  ;; base-mode conflicts with its Nyxt symbol if it's not prefixed
-  '(help-mode web-mode auto-mode nyxt::base-mode)
-  "Modes that `auto-mode' won't even try to save.
-Append names of modes you want to always control manually to this list.
-Be careful with deleting the defaults -- it can be harmful for your browsing.")
 
 (declaim (ftype (function ((or symbol root-mode)) (values symbol &optional)) maybe-mode-name))
 (defun maybe-mode-name (mode)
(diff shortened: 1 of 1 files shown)


source/auto-mode.lisp
```diff
@@ -3,24 +3,10 @@
 
 (uiop:define-package :nyxt/auto-mode
   (:use :common-lisp :nyxt)
-  (:import-from #:serapeum #:export-always)
-  (:import-from #:class-star #:define-class)
+  (:import-from :class-star :define-class)
```
jmercouris, line 6: I too prefer this style, though many still write `#:`

jmercouris:
This is about on the edge of what I would consider large enough for a pull request, and small enough to just merge. There are some significant-enough logic changes, so I think a pull request is OK.

jmercouris:
I am OK with this change. With regards to the define-configuration not working as you expect, I am not surprised. The slot is named in another package that is not used by the nyxt-user. What's the best way to fix this? Not sure, do we in-package, automatically add package prefixes?

aartaka:
> I am OK with this change. With regards to the define-configuration not working as you expect, I am not surprised. The slot is named in another package that is not used by the nyxt-user. What's the best way to fix this? Not sure, do we in-package, automatically add package prefixes?

Adding package prefixes sounds quite intuitive for me. 

Given that docstring of `proxy-mode` mentions using `define-configuration` on slots without prefixing them, this also seems to be the behavior that is to be expected from `define-configuration`. The reason it works with `proxy-mode` is that there is a function with the same name in `nyxt-user`. This name clash is not really pleasant, so using package prefix can ensure that things are properly isolated.

source/auto-mode.lisp
aartaka: Yes, I've brought the `#:` style back, it's more consistent with other Nyxt code.

Ambrevar: If I understand correctly, `#:` is uninterned (it's used by `gensym` for instance) while `:` in interned as a keyword.

Ambrevar:
About whether to send a pull request or not: if in doubt, then send a pull request :)

Thanks for the quick changes!

Ambrevar:
About the proxy example: You are totally right, this was a mistake on my end.  Fixed with a3d7788d4a01cd5d3d2929c90354a1868ad9a704.

PR #990, opened 2020-10-06
Discussion: Per-domain autofill helper

phil-hands:
I thought that it might be worth adding links to relevant articles, given that you've kindly written them :-)

Cheers, Phil.

### DIFF
--- a/source/tutorial.lisp
+++ b/source/tutorial.lisp
@@ -185,7 +185,9 @@ query inside parenthesis in which you can use 'and', 'or' and 'not'. Examples:")
     (:li (command-markup 'nyxt/web-mode:jump-to-heading) ": Query a heading (a
 section) of the current page and jump to it.")
     (:li (command-markup 'nyxt/web-mode:autofill) ": See the "
-         (:code "autofills") " browser slot.")
+         (:code "autofills") " browser slot or "
+         (:a :href "https://nyxt.atlas.engineer/article/autofills.org"
+                   "this article on autofills"))
     (:li (command-markup 'vcs-clone) ": Clone version control repository
 matching current URL.")
     (:li (command-markup 'open-file) ", " (command-markup 'download-open-file)


Ambrevar:
Thanks for the suggestion.  I'm not sure such links belong to the code base because:
- They might go out-of-sync (e.g. we update the autofill code but not the article)
- The article could be moved or removed, which would result in a broken link.

@jmercouris Thoughts?

phil-hands:

> Thanks for the suggestion.  I'm not sure such links belong to the code base because:

I mostly did the pull request to kick off a discussion, rather than in
the expectation it would be immediately merged, so I won't be even
slightly upset if it is not merged.

The problem I was trying to address was that the current documentation
in several places effectively boils down to "read the source code, and
then imagine what you might do with this feature".

I am sure that it makes sense to keep the in-codebase documentation
brief, but doing so makes it inaccessable to those of us that are not
completely fluent in common lisp.

Even having read the linked article, I'm not really sure how I'm
supposed to get Nyxt to suggest the right username to use on a
particular site.  I guess I'm supposed to write a function that consults
a list of tuples, and somehow calls out to pass to get hold of the
matching password, but I don't really know where to start I'm afraid.

Should documentation covering this be in the codebase? I've no idea, but
if you can do it a literate lisp, say, then that would at least provide
some hope that the code and the documentation would say in sync.

If it doesn't fit in the codebase, then perhaps settling on a stable
scheme for a documentation site that one could populate links into as
documentation becomes available would do the trick.

Cheers, Phil.

Ambrevar:
The documentation is supposed to be in 2 places:

- The manual.
- The docstrings.

You can access the docstrings from the browser using the `describe-*`
functions.

We want the documentation to be explorable and dynamic.

If something is missing are unclear (which is definitely the case as of
today!) then it should be considered a bug.

About your issue:

> Even having read the linked article, I'm not really sure how I'm
> supposed to get Nyxt to suggest the right username to use on a
> particular site.  I guess I'm supposed to write a function that consults
> a list of tuples, and somehow calls out to pass to get hold of the
> matching password, but I don't really know where to start I'm afraid.

An autofill can be a lambda, so you'd write a function that checks
whether the current URI matches and return the corresponding URL.

Untested example:

```lisp
(defvar *my-user-names*
  '(("wikipedia.org" "wikiuser")
    ("example.org" "exampleuser")))

(define-configuration browser
  ((autofills
    (append (list (nyxt::make-autofill
                   :key "Usernames"
                   (lambda ()
                     (second (assoc (quri:uri-host (url (current-buffer))) *my-user-names* :test #'string=)))))
            %slot-default))))
```

What you are trying to do seems to be a common use case, so it would
make sense to include it as an example in the autofill docstring.



> Should documentation covering this be in the codebase? I've no idea, but
> if you can do it a literate lisp, say, then that would at least provide
> some hope that the code and the documentation would say in sync.

Since the documentation is almost exclusively docstrings, we ensure that
it stays in sync with the code.

Ambrevar:
@jmercouris: What's the `name` slot for?  It's not documented.  Using a `define-class` and documenting the slots would be a good idea.

Could you incorporate the above example (after testing and confirming it's the right way to do it) in the doc?  Thanks!

jmercouris:
I will take a look and try to do it, yes :-)

jmercouris:
Reading through this again, even if I do what we have concluded, we still have not addressed the main problem with our documentaiton which @phil-hands has pointed out. It is very brief and assumes a lot of CL knowledge. We could maybe address this by putting in CL documentation directly into Nyxt. We can't link the hyperspec, but we might be able to utilize the Ultrapsec by @phoe.

With regards to brevity, it is the most important. We are doing very well on this front. Perhaps however we could also have an optional extended documentation for some topics. Autofills can be explained in one or two sentences. BUT there could be a button that says "More" or something that goes into a longer discussion about how one might utilize autofills and how they could be used in a common case of filling in passwords. Thoughts?

phoe:
CLUS is not yet done and won't be done for the next year perhaps.

You actually **can** link the Hyperspec - you simply need to copy and distribute it as-is without any modifications, since that's what the CLHS license states.

jmercouris:
Thank you for the clarification @phoe, I still don't feel like the license lends itself to be easilly integrated into Nyxt. For example, I believe we couldn't inline it into our documentation without the user going to a complete hyperspec page.

phoe:
> For example, I believe we couldn't inline it into our documentation without the user going to a complete hyperspec page.

Correct; this is undoable.

Ambrevar:
@jmercouris Did you get the change to update the autofill documentation?

jmercouris:
Not yet!

jmercouris:
The example as listed is broken. Will take a look again later... may also just make a macro for setting usernames for websites, it is probably a common enough use case.

Ambrevar:
I gave this another thought: what if we added a helper, say `make-autofill-per-host`, that takes a key and automatically derives the value from a user-provided map.  Example:

```lisp
(defvar *autofill-map*
 '("fish.org" '(:name "salmon" :color "pink")
   "aquarium.org" '(:name "bloop" :color "purple")))

(defineconfiguration browser 
  ((autofilles (list (make-autofill-per-host "name")
                             (make-autofill-per-host "color")))
```

Ambrevar:
Still feels like I'm repeating the keys, maybe we just need `make-autofill-from-host-map` that takes the map and return the list of autofill directly.

Ambrevar:
Why close this?

jmercouris:
I think that the discussion could be in a issue instead. Additionally, I thought we did add autofills for usernames, so it seemed solved, no?

Ambrevar:
The discussion is about per-domain autofills (the title :p).
We don't have anything like this (i.e. we don't offer a convenient way
to customize per-domain autofills).

Ambrevar:
True that a discussion should rather belong to an issue, but now that
it's here, I'd keep it rather than duplicating the content in a new issue.

Ambrevar:
Reopen then?

jmercouris:
I'm closing this. It is too long in the tooth! We need to clean up our pull requests. If there is an issue to be made, let us make one.

Issue #757, opened 2020-05-25
Add `load-after-system` helper?


Ambrevar:
I think a common pattern that we are to see when users want to depend on third-party Common Lisp systems is to load the system and if it's found, load a configuration file.  Which could itself be another system.

I suggest the following helper:

```
;; If :foo-extension is not found, this does nothing.
(load-after-system :foo-extension :bar-config)
;; Also works:
(load-after-system :foo-extension "path/to/bar/config")
```

The crucial point is that third-party system configuration can only be done in another file because the reader will not know about the loaded system package prefixes.

Ambrevar:
This would supersede #713.

jmercouris:
That is a good point you make... probably easier than doing the macro. Unless we force loading of the whole world before actually executing someone's init file.

jmercouris:
I'm not sure what is best.

PR #998, opened 2020-10-17
Ci warnings

Ambrevar:
Note the elegance of the process: I'm not parsing some the output of some executable.  Instead, I'm leveraging ASDF directly in pure Common Lisp.  This is awesome!


Ambrevar:
And this pull-request is meta-useful: It already highlights a problem with the double-definition of load-system!

What shall we do about it?  Rename one to `load-lisp-system`?  Which one?

Ambrevar:
See the error on GitHub: "Some checks were not succesful".

phoe:
Side note: you should write a separate article about it and publish it somewhere; possibly have it included on Planet Lisp as well.

Ambrevar:
Will do!

Ambrevar:
I suggest we replace the load-system helper with `load-after-system` as per #757.

jmercouris:
Tested, works fine on my machine!

PR #1003, opened 2020-10-19
Wip ospama

Ambrevar:


Ambrevar:
Left to do:
- Handle case when no package manager is found.
- Add manifest management for Guix / Nix.
- Add tests.
- Add macports support.

Ambrevar:
This is mostly functional by now.
@jmercouris Wanna give macports a try?

Ambrevar:
Thoughts:

- There seems to be a race condition the first time the buffer is created: the title appears briefly then gets overwritten by the command output.  I can't reproduce reliably and it seems to only happen the first time.

- We could add a `os-package-with-output` class and have `guix-package` (and `nix-package` in the future) inherit from it.  Then, from the client side, we can either check the type of the package object, or write methods for `os-package` and `os-package-with-output` (e.g. when describing the package or  when listing the files.

Ambrevar:
Tests done.

jmercouris:
I will give MacPorts a try :-)

Ambrevar:
OK to merge?  The macports / Nix backends can come later.

jmercouris:
OK to merge!

Ambrevar:
Done with e5d87559145a743776d592fd16e249b2cb8941b6.

jmercouris:
Thank you! This is really great, well done monsieur!

PR #1006, opened 2020-10-20
Disable WebGL in reduce-tracking mode.

aartaka:
WebGL allows for easy user identification, because the hash of WebGL canvas is almost unique, given that it depends on lost of variables in user's setup. Disabling WebGL is the only solid way to escape fingerprinting via WebGL. Breaks some websites that use WebGL, of course.

I cannot test whether it breaks Nyxt (because I cannot run Nyxt from master since approximately a week ago), so I need help in testing that. Shouldn't be hard -- run Nyxt with this patchset, go to https://panopticlick.eff.org/ , test the browser and see whether there's a WebGL hash captured. If hash is not captured and Nyxt doesn't crash, then we've made `reduce-tracking-mode` better at fingerprinting-avoidance :)

Let me know what you think of it!

### DIFF
--- a/source/reduce-tracking-mode.lisp
+++ b/source/reduce-tracking-mode.lisp
@@ -15,6 +15,8 @@
 part of the Accept-Language HTTP header.")
    (destructor
     (lambda (mode)
+      ;; What if user wants to have WebGL disabled by default?
+      (ffi-buffer-enable-webgl (buffer mode) t)
       (ffi-set-preferred-languages (buffer mode)
                                    (list (first
                                           (str:split
@@ -22,5 +24,6 @@ part of the Accept-Language HTTP header.")
                                            (or (uiop:getenv "LANG") "")))))))
    (constructor
     (lambda (mode)
+      (ffi-buffer-enable-webgl (buffer mode) nil)
       (ffi-set-preferred-languages (buffer mode)
                                    (preferred-languages mode))))))
--- a/source/reduce-tracking-mode.lisp
+++ b/source/reduce-tracking-mode.lisp
@@ -13,10 +13,12 @@
                         :type list-of-strings
                         :documentation "The list of languages that will be sent as
 part of the Accept-Language HTTP header.")
+   (previous-webgl-setting nil
+                           :documentation "The state of WebGL
+before reduce-tracking-mode was enabled.")
    (destructor
     (lambda (mode)
-      ;; What if user wants to have WebGL disabled by default?
-      (ffi-buffer-enable-webgl (buffer mode) t)
+      (ffi-buffer-enable-webgl (buffer mode) (previous-webgl-setting mode))
       (ffi-set-preferred-languages (buffer mode)
                                    (list (first
                                           (str:split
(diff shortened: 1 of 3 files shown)


source/reduce-tracking-mode.lisp
```diff
@@ -15,12 +15,15 @@
 part of the Accept-Language HTTP header.")
    (destructor
     (lambda (mode)
+      ;; What if user wants to have WebGL disabled by default?
+      (ffi-buffer-enable-webgl (buffer mode) t)
```
Ambrevar, line 19: Store the "previous-webgl-setting" in a mode slot from the constructor, then restore this value in the destructor.

Ambrevar:
What's blocking you from running Nyxt on master?
Works fine for me here on Guix.  Report your Guix commit, the Nyxt output, maybe the offending Nyxt commit if you have time to bisect.

Let me know how I can help.

aartaka:
> What's blocking you from running Nyxt on master?
> Works fine for me here on Guix. Report your Guix commit, the Nyxt output, maybe the offending Nyxt commit if you have time to bisect.

I'm investigating it. I'll send a proper Bug Report in a day, I guess. 

So far it looks like the problem is either with recently introduced `chanl`, or with WebKit. I'm uncertain, though.

source/reduce-tracking-mode.lisp
aartaka: Sounds fair :)

Ambrevar:
The synchronous minibuffer is a disruptive change indeed, a race
condition could be the culprit.

Ambrevar:
Tested: it properly disabled WebGL, however the hash is still visible on
panopticlick.

(Great website by the way, thanks for sharing!)

aartaka:
> Tested: it properly disabled WebGL, however the hash is still visible on panopticlick.

Yes, this could have happened too. Can you compare the number of identifying bits with and without WebGL? Reduction of identifying information is better than nothing :)

aartaka:
OK, WebGL state is now saved in a dedicated slot of `reduce-tracking-mode`.

Ambrevar:
I get the same result:

bits of identifying information: 14.98
one in x browsers have this value: 32345.57
value: 3c908d2a5cf3025e3ca103dc92d63975

aartaka:
I believe we should merge it nonetheless, because disabling WebGL is still a safer bet :)

Ambrevar:
But then I wouldn't like to include it in reduce-tracking since it would mislead the user into thinking that disabling WebGL the way we do it reduces tracking, while the only measure that we have shows that it doesn't.

Alternatively, I propose to add a nowebgl-mode.

jmercouris:
I would add a nowebgl-mode, which can be hooked into reduce-tracking mode if we do find out that by some measures it does reduce ones trackability.

aartaka:
That's reasonable :thinking:  I'll implement it.

aartaka:
I believe this PR can be closed in favor of #1013.

PR #1009, opened 2020-10-21
Current buffer last buffer

Ambrevar:
Description in the various commits.

### DIFF
--- a/source/configuration.lisp
+++ b/source/configuration.lisp
@@ -24,6 +24,28 @@ from a binary) then any condition is logged instead of triggering the debugger."
           (log:error "In ~a: ~a" f c)
           nil))))
 
+(defparameter %buffer nil)              ; TODO: Make a monad?
+
+(export-always 'current-buffer)
+(defun current-buffer (&optional window)
+  "Get the active buffer for WINDOW, or the active window otherwise."
+  (or %buffer
+      (match (or window (current-window))
+        ((guard w w) (active-buffer w))
+        (_ (when *browser*
+             (log:debug "No active window, picking last active buffer.")
+             (last-active-buffer))))))
+
+(export-always 'with-current-buffer)
+(defmacro with-current-buffer (buffer &body body)
+  "Execute BODY in a context in which `current-buffer' returns BUFFER."
+  ;; We `unwind-protect' to restore the right buffer when nesting this macro.
+  `(let ((old-%buffer %buffer))
+     (unwind-protect
+          (let ((%buffer ,buffer))
+            ,@body)
+       (setf %buffer old-%buffer))))
+
 (defun user-class-name (class-sym)
   (intern (str:concat "USER-" (string class-sym))
           (symbol-package class-sym)))
--- a/source/search-engine.lisp
+++ b/source/search-engine.lisp
@@ -37,7 +37,8 @@ the empty string."))
           (search-url engine)))
 
 (defun bookmark-search-engines (&optional (bookmarks (get-data (bookmarks-path
-                                                                (current-buffer)))))
+                                                                (or (current-buffer)
+                                                                    (make-instance 'user-buffer))))))
   (mapcar (lambda (b)
             (make-instance 'search-engine
                            :shortcut (shortcut b)
@@ -52,7 +53,8 @@ the empty string."))
 (defun all-search-engines ()
   "Return the `search-engines' from the `browser' instance plus those in
 bookmarks."
-  (sera:and-let* ((buffer (current-buffer)))
+  (let ((buffer (or (current-buffer)
+                    (make-instance 'user-buffer))))
     ;; Make sure `default-search-engine' returns the same value after the append.
     (append (bookmark-search-engines)
             (search-engines buffer))))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -544,28 +544,6 @@ sometimes yields the wrong reasult."
         ;; No window when browser is not started or does not implement `ffi-window-active'.
         (ignore-errors (ffi-window-active *browser*)))))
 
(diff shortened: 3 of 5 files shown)


source/configuration.lisp
```diff
@@ -24,6 +24,28 @@ from a binary) then any condition is logged instead of triggering the debugger."
           (log:error "In ~a: ~a" f c)
           nil))))
 
+(defparameter %buffer nil)              ; TODO: Make a monad?
```
jmercouris, line 27: I don't understand what a monad is still :-D, can you please explain how it would function and what the benefits are?

source/search-engine.lisp
```diff
@@ -37,7 +37,8 @@ the empty string."))
           (search-url engine)))
 
 (defun bookmark-search-engines (&optional (bookmarks (get-data (bookmarks-path
-                                                                (current-buffer)))))
+                                                                (or (current-buffer)
```
jmercouris, line 40: from my understanding, =(current-buffer)= will always return a buffer because it will resort to =(last-active-buffer)= which will just go through all buffers. Therefore, will there ever be an instance of a user-buffer created here?

jmercouris:
I like it! Now we can be sure that (current-buffer) will return EXACTLY what we expect within a command body! :-)

Very neatly done too!

source/configuration.lisp
Ambrevar: See https://github.com/atlas-engineer/nyxt/issues/868#issuecomment-673024703
Not that this comment is from before, nothing has changed here.

source/search-engine.lisp
Ambrevar: There is, when you instantiate a browser but you didn't start it!  It's a corner case but it's an important one for tests!

source/configuration.lisp
jmercouris: I still don't see why we need a monad when we can use a closure

source/search-engine.lisp
jmercouris: Ah, OK, in that case there will be no buffers!

jmercouris: We can make a user buffer object without starting the browser, but we can't actually interact with it, so this function won't actually work, will it?

source/search-engine.lisp
```diff
@@ -52,7 +53,8 @@ the empty string."))
 (defun all-search-engines ()
   "Return the `search-engines' from the `browser' instance plus those in
 bookmarks."
-  (sera:and-let* ((buffer (current-buffer)))
+  (sera:and-let* ((buffer (or (current-buffer)
```
aartaka, line 56: We probably don't need `sera:and-let*` here anymore, because there always will be an instance of `user-buffer` bound to `buffer`.

Ambrevar: Good catch!

source/search-engine.lisp
Ambrevar: It will, because bookmarks-path will return the default value.
This technique of making a buffer to access the defaults is used a couple of times, e.g. in the tutorial / manual.

source/configuration.lisp
Ambrevar: Because we need to reset %buffer when we change buffer within a `with-current-buffer`.  This is stateful, global and thus brittle.

A monad would give us full control over the side effects.

Ambrevar: Note: currently we _do_ reset %buffer in the only place (I think) that changes the current buffer.  So at the moment our code should be correct.

source/search-engine.lisp
jmercouris: Aha! I see!

jmercouris:
Please merge whenever you are ready!

source/configuration.lisp
jmercouris: I think the reason we have this problem is because %buffer is serving two purposes:

- result of active buffer
- place to save current-buffer when wrapping in with-current-buffer

jmercouris: therefore we have to update %buffer when we switch buffer within a within-current-buffer form, whereas we shouldn't have to, they should be discrete (I think)

Ambrevar: The goal of `with-current-buffer` is to ensure that `(current-buffer)` returns a given buffer.

%buffer is used in only one spot: `current-buffer`.

What is the other purpose that you see?

jmercouris: but as we just said, if we change the current buffer then %buffer is no longer valid, and (current-buffer) must return a different buffer. there are two concepts here

(with-current-buffer ...) should have a symbol/variable for the buffer that is the current buffer at the time of creation, and (current-buffer) should always just return the current buffer.

am I making sense? :-D I feel like it is a bit hard to explain the subtle difference here.

Ambrevar: OK, but this is not what we want.  We effectively want to locally override the return value of (current-buffer), and we don't want to resort to another function/variable to get the possibly-modified current buffer.

In other words, we don't want to write code that knows in advance whether it should use `current-buffer` or `buffer-from-with-current-buffer` (I made this last one up).

jmercouris: well, in that case, then we must deal with the state :-D

Ambrevar: This is exactly what monads allow us to do cleanly ;)

Ambrevar:
OK to merge then?

jmercouris:
OK to merge :-)

Ambrevar:
Done in ccde4fbab58e94092a0c2d44dfcda8e9e33b914c.

jmercouris:
Thank you monsieur!

PR #1010, opened 2020-10-21
Fix invert-color

### LINKED ISSUE #362: Custom stylesheets (CSS), e.g. Dark Mode support
Hi

is there any plan to support dark mode for website ala [stylish](https://github.com/StylishThemes/GitHub-Dark) or ala [dark reader](https://darkreader.org/help/en/) ?

Since I think its a very basic need for average users.

Thanks

brongulus:
invert-color bookmarklet was unreliable, didn't work and wasn't toggle-able, this new JS code makes it toggleable and reliable. Might fix #362


Ambrevar:
Thanks!

Did you write this Javascript yourself or did you copy-paste it from somewhere else?
If the former, would you like to write it in Parenscript instead?
If the latter, can you mention the source?

#362 is more about per-domain stylesheets and support for existing Dark CSS sheets.

brongulus:
I copied it from here: 
https://gist.github.com/frontdevops/8aea1e0252dd826488dad63319e3ec88#file-darkthemeswitcher-inline-js

jmercouris:
I don't think we can use it in any case because it is published without a license from that source. If the author has published it elsewhere with a license, then we can take a look.

jmercouris:
Thank you for looking into this!

brongulus:
> If the author has published it elsewhere with a license, then we can take a look.

I could try to contact them and ask for a licensed version? Would that be acceptable, since the current version of invert-color doesn't work at all.

jmercouris:
Yes, that would be wonderful!

frontdevops:
Hi! Use my code however you like. If necessary, I can make a repository where I will write that the free license.
No problem, use at your own discretion, I will only be glad. Free license - take whoever you want, do what you want

frontdevops:
I create it

https://github.com/frontdevops/darkthemeswitcher-inline

jmercouris:
Thanks so much :-)

Ambrevar:
@brongulus Can you edit your patch to include a comment referencing the origin of the bookmarklet then?
Thanks!

brongulus:
Done :)

jmercouris:
Thank you!

Issue #1012, opened 2020-10-22
Nyxt doesn't start from REPL


aartaka:
**Describe the bug**

Nyxt doesn't start properly when started from REPL -- it just hangs after logging "Initializing GTK Interface" to debug. It seems that it's a problem with `chanl:pexec` that starts soon after the "Initializing GTK Interface" message.

I haven't yet tested commits before the ones introducing `chanl`, though.

**Precise recipe to reproduce the issue**

Run Nyxt from master from the REPL :)

I haven't found other circumstances to narrow it down, unfortunately.

**Information**
- OS name+version: Guix System
```
$ guix describe
]8;;file://paranoidal/var/guix/profiles/per-user/aartaka/current-guix-56-link\Generation 56	Oct 22 2020 22:06:39]8;;\	(current)
  nonguix 02cf886
    repository URL: https://gitlab.com/nonguix/nonguix
    branch: master
    commit: ]8;;https://gitlab.com/nonguix/nonguix/commit/02cf886efac46de75a1226128a7cc4534ef426d2\02cf886efac46de75a1226128a7cc4534ef426d2]8;;\
  guix-gaming-games 3c9a3c4
    repository URL: https://gitlab.com/guix-gaming-channels/games.git
    branch: master
    commit: ]8;;https://gitlab.com/guix-gaming-channels/games/commit/3c9a3c450343b24fb7c27e4986b5da402347c1e3\3c9a3c450343b24fb7c27e4986b5da402347c1e3]8;;\
  chur 8fcb5dd
    repository URL: https://github.com/aartaka/chur-guix
    branch: master
    commit: ]8;;https://github.com/aartaka/chur-guix/commit/8fcb5ddaa15fd10d71c1e50950405bb6d48e4393\8fcb5ddaa15fd10d71c1e50950405bb6d48e4393]8;;\
  guix 3e09453
    repository URL: https://git.savannah.gnu.org/git/guix.git
    branch: master
    commit: ]8;;https://git.savannah.gnu.org/cgit/guix.git/commit/?id=3e09453884efa82ef97b8ec6e34470c67a1206a7\3e09453884efa82ef97b8ec6e34470c67a1206a7]8;;\

$ guix --version
guix (GNU Guix) 3e09453884efa82ef97b8ec6e34470c67a1206a7
Copyright (C) 2020 the Guix authors
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
[...]

PR #1013, opened 2020-10-22
Add nowebgl-mode

aartaka:
This adds the new mode to toggle WebGL state. 

Motivation (although an untested one) can be found in #1006. 

Untested (because of #1012), but the changes are trivial, so it shouldn't break anything.

Let me know what you think of it!


Ambrevar:
Thanks!

PR #1016, opened 2020-10-23
User status formatter

Ambrevar:
My own status bar is now this:

```lisp
(setf nyxt::*invisible-modes* '("base-mode" "my-mode" "certificate-exception-mode" "web-mode"))
(defun my-format-status (window)
  (let ((buffer (current-buffer window)))
    (if (or (internal-buffer-p buffer)
            (find-submode buffer 'proxy-mode))
        (setf (style (status-buffer window))
              (getf (mopu:slot-properties 'status-buffer 'style) :initform))
        (setf (style (status-buffer window))
              my-status-style))
    (str:concat
     (nyxt::format-status-modes buffer)
     (nyxt::format-status-buttons)
     (nyxt::format-status-load-status buffer)
     (nyxt::format-status-url buffer))))
```

Which is some 30 lines shorter than before! :)

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -328,7 +328,6 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
                 :line-height "16px"
                 :margin 0)
                ("#modes"
-                :border-right "2px solid rgb(120, 120, 120)"
                 :display "inline-block"
                 :padding "0 5px 0 5px"
                 :margin "0 5px 0 0")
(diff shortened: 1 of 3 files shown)


jmercouris:
so the advantage is that the user can more easily override specific functions instead of starting from scratch when customizing?

Ambrevar:
Override or simply define their own while reusing just some of the
vanilla functions.

jmercouris:
Also yeah, maybe we don't need that vertical bar. I have been thinking maybe it should be a different color though or something. I had been experimenting with making it a scrollable area so that it doesn't take up too much width. I also had made it hide the overflow with ellipses.

PR #1019, opened 2020-10-30
Minor review of the documentation

aadcg:
Hi Nyxt,

I'm sorry if some of these changes seem very picky. Feel free not to take the changes you think unnecessary  .

I tried to make things a bit more uniform across most of the .org files.

I also tried to improve the developer documentation, in particular the section relative to SLIME and SLY as they seemed a bit confusing to me.

Thank you!

### DIFF
--- a/documents/README.org
+++ b/documents/README.org
@@ -1,96 +1,111 @@
-* Nyxt Developer Readme                                                 :TOC:
+#+TITLE: Nyxt Developer Readme
+
+# If you edit this file, please make sure that you install the Emacs
+# package org-make-toc so that the table of contents below gets updated
+# automatically
+
+* Table of Contents                                                     :TOC:
 :PROPERTIES:
-:TOC:      :include all
+:TOC:      :include all :ignore this
 :END:
 :CONTENTS:
-- [[#nyxt-developer-readme][Nyxt Developer Readme]]
 - [[#developer-installation-with-guix][Developer Installation with Guix]]
 - [[#standard-developer-installation][Standard Developer Installation]]
   - [[#installing-sbcl][Installing SBCL]]
   - [[#installing-quicklisp-optional][Installing Quicklisp (optional)]]
-  - [[#gtk-dependencies][GTK dependencies]]
+  - [[#gtk-dependencies][GTK Dependencies]]
     - [[#lisp-dependencies][Lisp Dependencies]]
-    - [[#gnulinux-gtk-dependencies][GNU/Linux GTK dependencies]]
-    - [[#macos-specific-dependencies][macOS specific dependencies]]
-  - [[#qt-dependencies][Qt dependencies]]
+    - [[#gnulinux-gtk-dependencies][GNU/Linux GTK Dependencies]]
+    - [[#macos-specific-dependencies][macOS Specific Dependencies]]
+  - [[#qt-dependencies][Qt Dependencies]]
     - [[#lisp-dependencies][Lisp Dependencies]]
-    - [[#gnulinux-dependencies][GNU/Linux dependencies]]
-    - [[#macos-dependencies][macOS dependencies]]
-  - [[#fetch-the-source-code][Fetch the source code]]
+    - [[#gnulinux-dependencies][GNU/Linux Dependencies]]
+    - [[#macos-dependencies][macOS Dependencies]]
+  - [[#fetch-the-source-code][Fetch the Source Code]]
   - [[#compile][Compile]]
     - [[#using-the-makefile][Using the Makefile]]
     - [[#using-lisp][Using Lisp]]
-  - [[#update-local-lisp-libraries][Update local Lisp libraries]]
-- [[#run-nyxt-in-a-security-sandbox][Run Nyxt in a security sandbox]]
+  - [[#update-local-lisp-libraries][Update Local Lisp Libraries]]
+- [[#run-nyxt-in-a-security-sandbox][Run Nyxt in a Security Sandbox]]
 - [[#hacking][Hacking]]
   - [[#run-from-source][Run from Source]]
-  - [[#slime-or-sly-with-a-compiled-version-of-nyxt][SLIME (or SLY) with a compiled version of Nyxt]]
+  - [[#interacting-with-a-compiled-version-of-nyxt][Interacting with a Compiled Version of Nyxt]]
+    - [[#slime][SLIME]]
+    - [[#sly][SLY]]
 - [[#help--community][Help & Community]]
   - [[#learning-common-lisp][Learning Common Lisp]]
-  - [[#contribution][Contribution]]
+  - [[#contributing][Contributing]]
   - [[#irc-channel][IRC Channel]]
 :END:
 
 * Developer Installation with Guix
-First, clone the Nyxt repository.  It is most convenient to clone the Nyxt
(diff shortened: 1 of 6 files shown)


jmercouris:
Some preliminary comments:

+ Thank you for putting the time into this. Our documentation could definitely use extension and improvement!
+ We use sentence case throughout, it is more easy to be consistent than in title case. In sentence case only proper nouns and the first word are capitalized.

I am okay with sticking to a convention for lists being either '-' or '+'. Personally I favor '+', but I am probably the minority, so I am OK with using '-'.

The SLY documentation could be a bit more formal. Please avoid the usage of casual language. Do you think you could update this pull request with those changes?

Interested in your feedback,

thanks,

-John

documents/README.org
```diff
@@ -291,90 +317,122 @@ Then in a shell execute the following:
...
+From the SLY manual:
+#+begin_quote
+SLY is a direct fork of SLIME, the "Superior Lisp Interaction Mode
+for Emacs", which itself derived from previous Emacs programs such as
+SLIM and ILISP.  If you already know SLIME, SLY's closeness to is
+immediately apparent.  However, where SLIME has traditionally focused on
+the stability of its core functionality, SLY aims for a richer feature
+set, a more consistent user interface, and an experience generally
+closer to Emacs' own.
+#+end_quote
```
Ambrevar, line 373: Probably enough to link to the SLY web page instead of repeating it here.

documents/README.org
```diff
@@ -291,90 +317,122 @@ Then in a shell execute the following:
...
+
+First, follow the following steps in Nyxt:
+
+1. Install =cl-slynk= (available on Guix)
```
Ambrevar, line 377: For most users this is not needed, they just need to install SLY.

documents/README.org
```diff
@@ -291,90 +317,122 @@ Then in a shell execute the following:
...
+2. Add the following line to your Nyxt init file (usually located at
+   =~/.config/nyxt/init.lisp=)
```
Ambrevar, line 379: No need to repeat the path.

documents/README.org
```diff
@@ -291,90 +317,122 @@ Then in a shell execute the following:
...
+        (slynk:create-server :port slynk-port :dont-close t)
+        (echo "Slynk server started at port ~a" slynk-port))
+  #+end_src
+4. Run the command =start-lynk= in Nyxt
 
-Use like for =SLIME=, substituting =start-slynk= for =start-swank= in
-Nyxt, and =sly-connect= for =slime-connect= in Emacs.
+Then, in Emacs:
+
+1. Make sure you have =SLY= installed in Emacs
```
Ambrevar, line 402: No need for this since the above implies SLY was installed.

documents/README.org
```diff
@@ -291,90 +317,122 @@ Then in a shell execute the following:
...
+2. ~M-x sly-connect~
+3. Enter ~127.0.0.1~ for the host
+4. Enter the port number set in the Nyxt variable ~*swank-port*~
+   (e.g. ~4006~)
```
Ambrevar, line 406: We can shorten these 3 points into one: "Run `sly-connect` over port 4006", since we use the default host which is localhost.

Ambrevar:
Hi @aadcg, thanks you so much for these changes!
Consistency and rigour are always welcome here!

I agree with John's comments:
- Let's use "Sentence case", it's easier.
- Let's use `-` as you did.
- The SLY section could be more concise.  I've posted some suggestions.

About the other changes:

- Specifying `fill-column` is a good idea, it helps with consistency , thanks!
- You've used double-spacing for sentence separation.  I also use it but John does not.  This is a nit, but while we are at it let's pick one :)  The Emacs variable that controls this behaviour is `sentence-end-double-space` (`t` by default).  I believe John has set it to nil.  If we want to stick to double-spacing in Nyxt and not disturb other users configuration, we can add `sentence-end-double-space t` as a file-local variable.  Thoughts?

The rational behind double-spacing is that it allows the editor to tell the difference between an abbreviation period and a sentence period.  Compare:
- `This is a sentence.  This is another sentence.`
- `I'm listing this, that, etc. and now I go on within the same sentence.`

In the second sentence, Emacs knows that it's a single sentence because the etc period is followed by a single space only.  This is how `M-e` (`forward-sentence`) works.  Without double-spacing, forward-sentence would not work as expected.

jmercouris:
Single spaces are better because that is the way I was taught :-D. Deficiencies of the editor should not be a prevailing reason for using double spaces (the origin of double spaces for typewriters).

Ambrevar:
> Single spaces are better because that is the way I was taught :-D. Deficiencies of the editor should not be a prevailing reason for using double spaces (the origin of double spaces for typewriters).

It's not a deficiency of the editor, this is a deficiency in English
typography.  This is valid for all editors.

jmercouris:
Well, it is a deficiency of the editor... it should be able to tell what is a abbreviation, no :-D?

documents/README.org
aadcg: Perhaps there's something that I'm missing here, but I really need to install it in order to be able to run `start-slynk`. Notice that `cl-slynk` isn't an input to the nyxt package definition at `build-script/guix.scm`.

documents/README.org
aadcg: Will also remove the analog line at the slime section.

documents/README.org
aadcg: Makes sense. 

Btw `slime-connect` isn't working for me anymore for some obscure reason (the repl doesn't pop up), whereas `sly-connect` does.

aadcg:
I will not take sides in the single/double space discussion. I'm partial to double space because of what @Ambrevar mentioned about `M-e` and also because I find it easier to tell apart different sentences with monospaced fonts.

Regardless, I'm happy to follow any approach :)

aadcg:
Hopefully, I've made all the changes following your input. Let me know what else needs to be done from my side.

I think the only "open" issues are those concerning single/double spaces and fill-column. Happy to follow your lead!

documents/README.org
Ambrevar: cl-slynk contains the Common Lisp slynk code which is also in Emacs SLY.  Most users will install SLY from MELPA and slynk should be made automatically available.

documents/README.org
```diff
@@ -1,96 +1,111 @@
...
-    - [[#lisp-dependencies][Lisp Dependencies]]
+    - [[#lisp-dependencies][Lisp dependencies]]
     - [[#gnulinux-dependencies][GNU/Linux dependencies]]
     - [[#macos-dependencies][macOS dependencies]]
   - [[#fetch-the-source-code][Fetch the source code]]
   - [[#compile][Compile]]
     - [[#using-the-makefile][Using the Makefile]]
     - [[#using-lisp][Using Lisp]]
-  - [[#update-local-lisp-libraries][Update local Lisp libraries]]
+  - [[#update-local-lisp-libraries][Update local lisp libraries]]
```
Ambrevar, line 28: Should be "Lisp".

Ambrevar:
I gave it another read, it's looking good to me beside the last few typos! Can you fix them? Thanks! :)

aadcg:
@Ambrevar, fixed :)

Is there smth else I should do? Thanks.

Ambrevar:
@jmercouris: OK to merge?  What's your take on the sentence spacing point?

jmercouris:
My take is that it does not matter, it will not be that which makes us succeed or fail, so I am OK with either in the end.

Ambrevar:
Merged as 480ef005f47c467f334f228bbdc391b94a7db45e.  Thanks @aadcg !

PR #1022, opened 2020-11-02
CI: Build .deb package on tag push.

Ambrevar:
Note that I have to include the system definition in `nyxt.asd` if I want to inherit from the `:author`, `:version`, etc.

The build completes in less than 10 minutes, this is really good!

Notes:

- It'd be neat if we could automatically update the specialized SBCL version used by linux-packaging.
- I've deleted the old RPM .asd since it's largely obsoleted by linux-packaging.  I suggest that we enable the .rpm production once the .deb has proven to be mature.
- The GitHub artifact is a .zip file.  It seems that upstream forces us to wrap our result in a zip, which is a bit dumb.  But maybe there is a way around it.

@ralt What do you think?
I'm planning to write an article on Nyxt CI, I'll make sure to mention your amazing work!


Ambrevar:
Shoot!

Actually using the same .asd has the big drawback that linux-packaging must be loaded, which is an extra dependency that's useless for non-deb packagers.

So I should move the system to another .asd.  The drawback is that I need to repeat the metadata like `:author`.

ralt:
This looks great! Are you sure the modern ASDF is required?

ralt:
Are you sure the .asd file matters? Isn't linux-packaging only needed when the system is loaded?

Ambrevar:
> Are you sure the modern ASDF is required?

system-version is undefined in SBCL's asdf.

Florian Margaine <notifications@github.com> writes:

> Are you sure the .asd file matters? Isn't linux-packaging only needed when the system is loaded?

Yes it does because ASDF parses the whole .asd on `load-asd`.  It will
fail on 

  `:defsystem-depends-on ("linux-packaging")`

and the special keywords like `:additional-dependencies` if
linux-packaging is not found.

I'll send a fix soon.

ralt:
Ok, that's a shame.

Ambrevar:
Fixed moving the system to a separate file.

ralt:
I wanted to check in the artifacts but haven't found them... have you verified with e.g. `dpkg-deb -I` whether the resulting `.deb` depends on the correct libraries? (gtk etc)

ralt:
I also just realized: you're removing the rpm build and adding a deb build?

Ambrevar:
> I wanted to check in the artifacts but haven't found them... have you verified with e.g. `dpkg-deb -I` whether the resulting `.deb` depends on the correct libraries? (gtk etc)

Artifacts are created on tag push.  So there will only be an artifact
for the next pre-release.

I have tested the resulting .deb in a virtual machine running Ubuntu.

Is there a pitfall with regard to library dependencies?  Can `dpkg-deb
-I` validate something beyond actually starting Nyxt in a virtual machine?

> I also just realized: you're removing the rpm build and adding a deb build?

Indeed, the RPM code is useless I presume now that we depend on
linux-packaging.
I suggest to also create an RPM artifact once the .deb has been validated.

ralt:
> Is there a pitfall with regard to library dependencies?  Can `dpkg-deb
-I` validate something beyond actually starting Nyxt in a virtual machine?

Potentially, if that VM already has what Nyxt depends on, without explicitly declaring it. The command will show the `Depends` field, which should list things like libgtk. When a .deb package is installed, dpkg will ensure that the dependencies are installed as well.

Ambrevar:
I can access this "depends" field in the "control" file by extracting
the .deb archive with `ar x`.

Here is the list:

    Depends: glib-networking, gsettings-desktop-schemas, xclip, enchant, notify-osd, libc6, zlib1g, libwebkit2gtk-4.0-dev, libgtk-3-0, libgtk-3-0, libpangocairo-1.0-0, libpango-1.0-0, libcairo2, libgdk-pixbuf2.0-0, libglib2.0-0, libglib2.0-0, libglib2.0-0, libglib2.0-0, libc6-i386, libfixposix-dev, libssl1.1

Looks good to me!

ralt:
Awesome, thanks!

jmercouris:
This looks ready for me. Please merge as soon as you wish!

jmercouris:
As per the ASD duplication/credentials, what about this: read the original ASD, spit out a modified ASD that includes linux packaging for usage in build systems?

Ambrevar:
I found a way to do it: use read-time expansion with `#.` :)

Merged with 5913d2f4393045b82b83901b38e060b55d4e98fe.

Issue #1001, opened 2020-10-19
Add buffer HTML generation helper


Ambrevar:
We have the following pattern all over the place:

```lisp
(let* ((buffer (or (find-buffer 'whatever-mode)
                   (whatever-mode
                    :activate t
                    :buffer (make-internal-buffer :title "*My special buffer*"))))
         (content
           (markup:markup
            (:style (style buffer))
            (:h1 "Packages")
            ;; ...
            ))
         (insert-content (ps:ps (setf (ps:@ document body |innerHTML|)
                                      (ps:lisp content)))))
    (ffi-buffer-evaluate-javascript-async buffer insert-content)
    (set-current-buffer buffer)
    buffer)
```

I suggest we factor this into a helper function, e.g. `make-html-buffer`.  Thoughts?

Ambrevar:
@jmercouris: Thoughts?

jmercouris:
Yes, this would be a good candidate for an abstraction, agreed!

aartaka:
It seems we found the first thing to abstract into #946 :) After all, these auto-generated pages are UI that we're trying to build.

Ambrevar:
Indeed!

I also suggest the following helpers, since they keep popping up:

```lisp
(defun html-write (content &optiona (buffer (current-buffer)))
  (ffi-buffer-evaluate-javascript-async
   buffer
   (ps:ps (ps:chain document
                    (write (ps:lisp content))))))

(defun html-set (content &optiona (buffer (current-buffer)))
  (ffi-buffer-evaluate-javascript-async
   buffer
   (ps:ps (setf (ps:@ document body |innerHTML|)
                (ps:lisp content)))))
```

jmercouris:
That is a good start I think! What do you think of specializing a method on buffer?

Ambrevar:
Which use case can you foresee?

jmercouris:
I can anticipate that internal and external buffers may have different ways of writing, we may for example use renderer methods to set the HTML directly in future iterations (without JavaScript)

Ambrevar:
Are we ever _writing HTML_ to web-buffer?

jmercouris:
I still think we should use methods, it is a kind of type checking in a way if you think about it

Ambrevar:
It's not because it's not caught at compilation time.
Type checking must be done with (declaim (ftype ...)).

jmercouris:
well, yes, but IS caught at run time :-D

[...]

PR #1023, opened 2020-11-04
Html helpers

Ambrevar:
This saves almost 100 lines of code in help.lisp alone!
It also reduce indentation, factors out the redundant code thus removing room for typos.

Thoughts?
If this looks OK, we can use the helpers in the entire code base afterwards.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -593,6 +593,22 @@ sometimes yields the wrong reasult."
    (ps:ps (setf (ps:@ document body |innerHTML|)
                 (ps:lisp content)))))
 
+(defmacro with-current-html-buffer ((buffer-var title mode) &body body)
+  "Create and switch to a buffer in MODE displaying CONTENT.
+BUFFER-VAR is bound to the new bufer in BODY.
+MODE is a mode symbol.
+BODY must return the HTML markup as a string. "
+  `(let* ((,buffer-var (or (find-buffer ,mode)
+                           (funcall (symbol-function ,mode)
+                            :activate t
+                            :buffer (make-internal-buffer :title ,title)))))
+     (html-set
+      (progn
+        ,@body)
+      ,buffer-var)
+     (set-current-buffer ,buffer-var)
+     ,buffer-var))
+
 (defmacro define-ffi-generic (name arguments)
   `(progn
      (export-always ',name)
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -581,6 +581,18 @@ sometimes yields the wrong reasult."
                    :if-does-not-exist :create
                    :if-exists :append))))))
 
+(defun html-write (content &optional (buffer (current-buffer)))
+  (ffi-buffer-evaluate-javascript-async
+   buffer
+   (ps:ps (ps:chain document
+                    (write (ps:lisp content))))))
+
+(defun html-set (content &optional (buffer (current-buffer)))
+  (ffi-buffer-evaluate-javascript-async
+   buffer
+   (ps:ps (setf (ps:@ document body |innerHTML|)
+                (ps:lisp content)))))
+
 (defmacro define-ffi-generic (name arguments)
   `(progn
      (export-always ',name)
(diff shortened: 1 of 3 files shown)


Ambrevar:
Closes #1001.

jmercouris:
This is great! I love it! I have one concern though: must we make a unique mode for every type of buffer? I say this because it appears it will look for a buffer with a matching that already exists and replace its content. Consider two different help buffers one about a command, and one about a function, if they do not have distinct modes, one could overwrite the other during creation no?

Ambrevar:
Indeed, this is an issue we've had since ever, now is the time to fix
it!

Here is what I suggest, trying to mimick the behaviour of Emacs' Info-mode.

- Find existing buffer using the following algorithm:
  - If current buffer is of the right mode, reuse it.
  - Otherwise look for a buffer with the right mode _and_ title.

- Rename the various *help-...* titles to *help*.

- Add history support to internal-buffers.

- When reusing an existing buffer, first set the URL to about:blank to
  that it starts a new history entry.

- Add a rename-buffer command so that the user can create multiple help
  buffers.

Use cases:

- No help buffer.  `F1-C buffer` describes the BUFFER class in a new
  *help* buffer.

- Now `F1-c set-url` describes SET-URL in the same buffer.

- `history-backwards` goes back to BUFFER.

- `rename-buffer` to `*help-2*`.

- `F1-s cookies-path` describes the COOKIES-PATH in `*help-2*`.

- Now switch to another buffer.

- `F1-f whatever` desribes WHATEVER in a new `*help*` buffer, without
  overriding `*help-2*`.

Does that make sense?


Note to myself: remove the spurious double colon (`::`).

jmercouris:
That is an interesting solution in the long term. it does make sense, is it the best model for user interaction? That I am not sure of. I think the behavior of it may be confusing to some users.

Since it requires a significant change, and many decisions, I suggest the following change in the interim so that we can utilize this new patch. Check if the mode AND the name is the same as the new buffer. That way, you will only override existing versions of the buffer. For example, `*Tutorial*` will only overwrite `*Tutorial*`, even though `*Help*` may also have the same modes active. What do you think?

Ambrevar:
There may be a misunderstanding because what I'm suggesting is very easy
to implement and actually almost exactly what you suggest :)

Should I update this pull request with what I have in mind, then we
discuss it?

jmercouris:
Well, there is a difference here, I am not suggesting that the user rename buffers to avoid them being overwritten :-D

If you think it is a five minute thing though, feel free to make changes to this PR. What I am suggesting is just changing the predicate for reusing a buffer to consider the title of the buffer AND the mode (which woud work in our existing model without any other changes).

Ambrevar:

> Well, there is a difference here, I am not suggesting that the user rename buffers to avoid them being overwritten :-D
>
> If you think it is a five minute thing though, feel free to make changes to this PR. What I am suggesting is just changing the predicate for reusing a buffer to consider the title of the buffer AND the mode (which woud work in our existing model without any other changes).

So 2 strategies here:

- Either we make all help buffer titles unique.  (Yours.)
- Or we reuse them.  (Mine.)

Is this correct?

Your strategy is a 1 minute change, mine is a 5 minute change :)

jmercouris:
That is more or less, correct, yes :-)

jmercouris:
I like the unique titles as well becuase they make things easily searchable in switch-buffer

jmercouris:
So... is my strategy OK then? Shall I implement it?

Ambrevar:
Don't bother, I'll update the patch in a few minutes while I'm still on
a roll ;)

Ambrevar:
Done in 54fdfff6348e0ad8279a7e920afecacfec107d70, is this what you have
in mind?

jmercouris:
That is exactly it, feel free to merge when you are ready, and I can do some of the translation work/line deletion as well if you want. Unless you really enjoy doing it :-D

Ambrevar:
You mean using with-current-html-buffer in the rest of the code base?

CI is complaining, need to fix it before merging.

jmercouris:
Yes, I mean using it in the rest of the codebase

Ambrevar:
Merged with ed879430cdf13b5dcc233eb7ac2bd02b2367bef2.

Feel free to go berserk with with-current-html-buffer, html-write and html-set! :)

PR #1031, opened 2020-11-11
Style mode (and auto-mode) redesign

Ambrevar:
This reduces the length of style-mode.lisp from 127 lines to 84 :) 

This mainly removes the URL-style association, leveraging auto-mode instead.

To use a specific CSS for a given site, simply inherit from style-mode or dark-mode and provide the CSS in one of the slots.

The drawback is that this can't be configured graphically.  But anyways to association list could not really be configured graphically before either.
We need some interface to create modes.


Ambrevar:
A drawback of the previous code is that there is a logic conflict in the URL/predicate association when used together with auto-mode.

This patch removes this issue since only auto-mode drives the URL/predicate matching.

Ambrevar:
If we merge this, then we need to update the style-mode article on the homepage.

Ambrevar:
Here is what I've added to my config:

```lisp
(define-mode dark-mode-github (nyxt/style-mode:style-mode)
  "Dark-mode for GitHub."
  ((nyxt/style-mode:style-file "/home/ambrevar/projects/solarized-everything-css/css/solarized-dark/solarized-dark-github.css")))
```

Now I can use auto-mode to use dark-mode-github on all github pages.

Ambrevar:
Considering how this alters the style-mode API, should we take down the https://nyxt.atlas.engineer/article/dark-mode.org article in the meantime? @jmercouris

jmercouris:
No, let's not take it down just yet, I am not really sold on this design, hence why I have hesitated commenting. I am hoping that it grows on me :-D

Ambrevar:
Can you elaborate on what bothers you with this design?

In my opinion:

- It's shorter.

- It's simpler, it has no level of indirection since there is only one
  class and no more association list.

- It's less for the user to learn since the URL rules would work the
  same of for every other mode.

- An issue with the previous design, as I mentioned, is that the end
  user ends up with 2 orthogonal (and incompatible) ways of enable dark
  mode for select URLs.

Ambrevar:
This is my auto-mode rule for Nyxt's GitHub page:

```lisp
((match-regex "https://github.com/atlas-engineer/nyxt/")  :included (dark-mode-github) :excluded (noscript-mode))
```

Ambrevar:
I could also have a rule that specifies other style modes to be excluded:


```lisp
((match-regex "https://github.com/atlas-engineer/nyxt/")  :included (dark-mode-github) :excluded (dark-mode noscript-mode))
```

bqv:
+1 for this design

Ambrevar:
@jmercouris suggested that if we were to keep this design, it would be nice if we could have a way to easily associate a mode with multiple URLs, like the original design allows.

https://nyxt.atlas.engineer/article/dark-mode.org

With the new design, if we want to associate a style with a URL we would do something like this (untested):

```lisp
(define-mode dark-mode-mediawiki (nyxt/style-mode:style-mode)
  ((nyxt/style-mode:style-url "https://raw.githubusercontent.com/alphapapa/solarized-everything-css/master/css/solarized-dark/solarized-dark-mediawiki.org.css")))

(add-modes-to-auto-mode-rules
 '(match-domain "example.org") :include '(dark-mode-mediawiki))
```

which is a bit more verbose than the original

```lisp
(make-style-association
  :url "https://example.org"
  :style-url "https://raw.githubusercontent.com/alphapapa/solarized-everything-css/master/css/solarized-dark/solarized-dark-mediawiki.org.css")
``` 

The problem is more general than with just style-mode.  It's about the relationship between mode specialization and URLs.

Currently auto-mode can associate URLs with a specific _mode list_, but we can't easily associate a specifically configured mode with URLs.

So... What if we added _slot specialization_ support to auto-mode?
In short, instead of just storing the mode name, we would also store it's initialization slot values.

The user config would then become:

```lisp
(add-modes-to-auto-mode-rules
 '(match-domain "example.org") :include '(foo-mode
                                          (dark-mode :style-url "http://...")
                                          bar-mode))
```

and the resulting rule would be very similar:

```lisp
((match-domain "example")  :included (foo-mode (dark-mode :style-url "http://...") bar-mode))
```

I believe with this we are  fully getting back the behaviour of the original design, while making it applicable to all modes _and_ well integrated with auto-mode.

@jmercouris  @aartaka Thoughts?

jmercouris:
Aha, a very good point about expanding auto-mode's behavior! I think this is a very good design, I like it!

Ambrevar:
I need to sleep on it.  I'd like to ponder what we could otherwise do with _user classes_ and _mode hooks_.  

If we have overlapping features I'd like to factor them out.

jmercouris:
Indeed, this design decision need not be rushed.

aartaka:
> So... What if we added _slot specialization_ support to auto-mode?
> In short, instead of just storing the mode name, we would also store it's initialization slot values.

This sounds really good! One disadvantage I can see is that serialization of complex data structures will make auto-mode rules file less readable. One will need to understand much more implementation details to confidently edit the rules file (imagine if one of the mode slot is `buffer` instance -- it will take several line on its own!).

Ambrevar:

> This sounds really good! One disadvantage I can see is that serialization of complex data structures will make auto-mode rules file less readable. One will need to understand much more implementation details to confidently edit the rules file (imagine if one of the mode slot is `buffer` instance -- it will take several line on its own!).

Yeah, that's a concern.  Unless... We don't store serialized data, but
Lisp forms to execute on mode initialization.  What do you think?

aartaka:
> Yeah, that's a concern. Unless... We don't store serialized data, but Lisp forms to execute on mode initialization. What do you think?

Yes, that sounds good :)

Ambrevar:
Commit 37d471c01ccefce62b325f79a57c2a6dbd0ce552 adds argument support to
auto-mode.

@jmercouris @aartaka Can you test it?

I've put this in my init file:

``` lisp
(nyxt/auto-mode::add-modes-to-auto-mode-rules
 '(match-domain "example.org") :include '((dark-mode :style-url "https://raw.githubusercontent.com/alphapapa/solarized-everything-css/master/css/solarized-dark/solarized-dark-github.css")))
```

Warning: Make sure you Backup your rule file before proceeding since a code
mistake may erase all your rules :(

Ambrevar:
Commit 37d471c01ccefce62b325f79a57c2a6dbd0ce552 adds argument support to
auto-mode.

@jmercouris @aartaka Can you test it?

I've put this in my init file:

``` lisp
(nyxt/auto-mode::add-modes-to-auto-mode-rules
 '(match-domain "example.org") :include '((dark-mode :style-url "https://raw.githubusercontent.com/alphapapa/solarized-everything-css/master/css/solarized-dark/solarized-dark-github.css")))
```

Warning: Make sure you Backup your rule file before proceeding since a code
mistake may erase all your rules :(

Ambrevar:
Note that the above snippet _erase_ your current rule file.
If you already have a rule file, simply add a rule to it, like

```lisp
 ((match-domain "example.org")  :included ((dark-mode :style-url "https://raw.githubusercontent.com/alphapapa/solarized-everything-css/master/css/solarized-dark/solarized-dark-github.css")))
 ```

**Comment by aartaka** [comment]
Just tested it and this seems to work fine enough! I noticed that the rule does not always apply and you need to reload the page to make is re-styled. That, I think, is not a style-mode problem -- it's, probably, loading a file from somewhere on the Internet that takes time.

But it's all good otherwise :)

File: source/renderer-script.lisp [hunk 1]
```diff
@@ -70,6 +70,14 @@ The function can be passed ARGS."
    (ps:ps (setf (ps:@ document body |innerHTML|)
                 (ps:lisp content)))))
 
+(defun html-set-style (style-string &optional (buffer (current-buffer)))
+  (let ((style (markup:markup (:style style-string))))
+    (ffi-buffer-evaluate-javascript-async
```
aartaka, line 75: How about using a recently added `webkit:webkit-user-content-manager-add-style-sheet` on WebKit port? It should be more reliable that plain stylesheet insertion via JS.

Abstracting this stylesheet injection into `ffi-buffer-inject-css` (arbitrary name, feel free to use a better one) then sounds intuitive for that. We can call the `webkit:webkit-user-content-manager-add-style-sheet` on WebKit while using JS calls on Qt, for example.

Ambrevar:
> How about using a recently added `webkit:webkit-user-content-manager-add-style-sheet` on WebKit port? It should be more reliable that plain stylesheet insertion via JS.

Yes, it's probably a good idea.  I suggest we do it in a subsequent pull
request :)

Ambrevar:
OK to merge then?

jmercouris:
OK to merge, let's not forget to update the article!

Ambrevar:
Let's update it for pre-release 5 then?

I've pushed an update on the =dark-mode-update= branch.
It's almost 40 lines shorter! :p

jmercouris:
Agreed!

Ambrevar:
Merged in ac5323f0dbf30d812aa60029a759172f0e1a8651.

jmercouris:
Thank you!!!

PR #1035, opened 2020-11-12
buffer.lisp: refresh list-buffers buffer after deletion

aadcg:
Hi Nyxt,

This is just a very dirty attempt to fix the issue. I'm not sure if you're fine about this "solution" for the time being. 

I'm happy to follow any of your advice, as always.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -738,7 +738,9 @@ See `make-buffer'."
 (define-command delete-buffer (&key id)
   "Delete the buffer(s) via minibuffer input."
   (if id
-      (buffer-delete (gethash id (slot-value *browser* 'buffers)))
+      (progn
+        (buffer-delete (gethash id (slot-value *browser* 'buffers)))
+        (nyxt::list-buffers))
       (let ((buffers (prompt-minibuffer
                       :input-prompt "Delete buffer(s)"
                       :multi-selection-p t
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -738,14 +738,13 @@ See `make-buffer'."
 (define-command delete-buffer (&key id)
   "Delete the buffer(s) via minibuffer input."
   (if id
-      (progn
-        (buffer-delete (gethash id (slot-value *browser* 'buffers)))
-        (when (current-buffer) 'buffer-listing-mode (list-buffers)))
+      (buffer-delete (gethash id (slot-value *browser* 'buffers)))
       (let ((buffers (prompt-minibuffer
                       :input-prompt "Delete buffer(s)"
                       :multi-selection-p t
                       :suggestion-function (buffer-suggestion-filter))))
-        (mapcar #'buffer-delete buffers))))
+        (mapcar #'buffer-delete buffers)))
+  (when (find-submode (current-buffer) 'buffer-listing-mode) (list-buffers)))
 
 (define-command reduce-to-buffer (&key (delete t))
   "Reduce the buffer(s) via minibuffer input and copy their titles/URLs to a
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -740,7 +740,7 @@ See `make-buffer'."
   (if id
       (progn
         (buffer-delete (gethash id (slot-value *browser* 'buffers)))
-        (nyxt::list-buffers))
+        (when (current-buffer) 'buffer-listing-mode (list-buffers)))
       (let ((buffers (prompt-minibuffer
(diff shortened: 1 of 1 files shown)


Ambrevar:
I guess you want to refresh the list if you were showing `list-buffers`.
Your fix would not do the right thing if the user is not showing `list-buffers` however.

Instead, I suggest to only update the list if the current buffer is in buffer-listing-mode.

Untested:

```lisp
(when
  (find-submode (current-buffer) 'buffer-listing-mode)
  (list-buffers))
```

aadcg:
You're right. I left it like that because I noticed that the only place where the buffers are being deleted with an id is precisely in `list-buffers`.  

It will make the necessary changes.

Ambrevar:
`current-buffer` returns the current buffer :) So the `when` test here is always true (unless there is no buffer at all, which should never be the case once the browser has started).

The ` 'buffer-listing-mode` is ignored in your code.

You want to test if the 'buffer-listing-mode is present in the current buffer.
For this, you can usei either

- `(find-mode (current-buffer) 'buffer-listing-mode)`: return the object that's an instance of class `buffer-listing-mode` in the current buffer.  Return `nil` if there is no such object.
- `find-submode`: same as above but also matches object that are instance of a _subclass_ of `buffer-listing-mode'.  This is useful to match custom modes.

In general, you should use `find-submode`.

Does that make sense?

aadcg:

> Does that make sense?

Of course, my brain was clearly somewhere else... sorry!

I'm wondering if this is beneficial.  It would give us a way to update
the list-buffers when a buffers is deleted, but it would not do anything
when buffers are created.  From my cursory look, there are several
functions that triggers the creation of new buffers, and it seems not a
great idea to append such a check at the end of all of them.

I only noticed the "Update" button now (typical me :D).  Only thing that
would make lots of sense (imo) would be to update the list-buffers when
we switch to it, which doesn't happen as of now.

Thoughts?

Ambrevar:
André Alexandre Gomes <notifications@github.com> writes:

> I'm wondering if this is beneficial.  It would give us a way to update
> the list-buffers when a buffers is deleted, but it would not do anything
> when buffers are created.  From my cursory look, there are several
> functions that triggers the creation of new buffers, and it seems not a
> great idea to append such a check at the end of all of them.

You are right, the current patch does not fix the problem at the right
place.

We have only 2 entry points for buffer list modifications:
`buffers-set` and `buffers-delete`.

Here is what I suggest: in the above two functions, we broadcast a
signal.  (I don't know how to broadcast signals in Common Lisp, we have
to find out.)

Then in `list-buffers` we listen to this signal and refresh the buffer
on every ping.

Thoughts?

aadcg:
@Ambrevar, sounds great. I will investigate more about broacasting signals. Should I close this one and open another when I have smth meaninful?

Ambrevar:
You can leave it open so that we don't forget, no problem!

Ambrevar:
Couldn't `watch-mode` be used for this? :)
Anyways, it's probably not the right fix... :p 

I believe the we need to set the internal buffer URLs.  @aartaka Thoughts on this?

aartaka:
> Couldn't `watch-mode` be used for this? :)
> Anyways, it's probably not the right fix... :p

Hooks work better there, it seems.

> I believe the we need to set the internal buffer URLs. @aartaka Thoughts on this?

I agree with that :) But we need to finally move to the WebKit-provided mechanism for custom schemes to do this. Out current approach is not safe (see https://github.com/atlas-engineer/nyxt/issues/1094).

What blocks the WebKit-provided approach (see https://github.com/joachifm/cl-webkit/pull/53) is just a necessity to test and debug it. I hadn't a chance to do this -- it was too low on my priority list :)

Next month, when I have some more time, I could finally polish it. After that, we'll be able to leverage WebKit for `lisp:` URLs.

aadcg:

> Hooks work better there, it seems.

I agree that hooks are a better approach.

What are custom schemes?

aartaka:
> What are custom schemes?

Think `about:` scheme of Firefox & others. Basically, custom schemes are browser-specific URI prefixes dedicated to a certain goal, like browser configuration, or special features, like `reader-view:` (or some similar) scheme in Epiphany.

Ambrevar:
@jmercouris Why did you close this?

jmercouris:
I spoke with Andre on IRC and he said OK to close.

Ambrevar:
OK, let's upon an issue about it then.

PR #1036, opened 2020-11-12
help.lisp: minor bug fix

aadcg:
Hi Nyxt!

This one should be straightforward.


Ambrevar:
Ouch, this made it to the pre-release-4 :/
Anyways, good catch!

PR #1042, opened 2020-11-17
Start args

Ambrevar:
Benefit: we now have keyword argument completion from the REPL for the start function!

Drawback: if we want to start the browser over no URL, we still need to pass NIL.

### DIFF
--- a/source/minibuffer-prompt.lisp
+++ b/source/minibuffer-prompt.lisp
@@ -1,10 +1,13 @@
 (in-package :nyxt)
 
-(defmacro define-function (name args &body body)
-  "Eval ARGS then define function over the resulting lambda list.
+(defmacro define-function (name args docstring &body body)
+  "Eval ARGS and DOCSTRING then define function over the resulting lambda list
+and string.
 All ARGS are declared as `ignorable'."
-  (let ((evaluated-args (eval args)))
-    `(defun ,name  ,evaluated-args
+  (let ((evaluated-args (eval args))
+        (evaluated-docstring (eval docstring)))
+    `(defun ,name ,evaluated-args
+       ,evaluated-docstring
        (declare (ignorable ,@(set-difference (mapcar (lambda (arg) (if (listp arg) (first arg) arg))
                                                      evaluated-args)
                                              lambda-list-keywords)))
@@ -17,6 +20,7 @@ All ARGS are declared as `ignorable'."
             (delete 'input-buffer
                     (public-initargs 'minibuffer))
             '((input-buffer nil explicit-input-buffer)))
+    "Return a new minibuffer instance."
   (let ((tmp-input-buffer (make-instance 'text-buffer:text-buffer))
         (tmp-input-cursor (make-instance 'text-buffer:cursor)))
     (cluffer:attach-cursor tmp-input-cursor tmp-input-buffer)
--- a/source/start.lisp
+++ b/source/start.lisp
@@ -397,9 +397,11 @@ Otherwise bind socket."
                                                            #'symbol-name
                                                            #'opts::name)
                                              opts::*options*))
-  "Start the browser, loading URLs if any.
+    (format nil "Start the browser, loading URLs if any.
 URLs is a list of strings.
 The OPTIONS are the same as the command line options.
+
+~a
 
 Examples:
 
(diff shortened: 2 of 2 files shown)


Ambrevar:
There are warnings with CCL, but I don't get why:

```
(In OSICAT::%OPEN-TEMPORARY-FILE/FD-STREAMS: Undefined function OSICAT::MAKE-FD-STREAM
19
;   In OSICAT::%OPEN-TEMPORARY-FILE/FD-STREAMS: Undefined function OSICAT::MAKE-FD-STREAM
20
;Compiler warnings for "home:.roswell;lisp;quicklisp;dists;quicklisp;software;osicat-20200925-git;src;osicat.lisp.newest" :
21
;   In OSICAT::%OPEN-TEMPORARY-FILE/FD-STREAMS: Undefined function OSICAT::MAKE-FD-STREAM
22
)
```

Ambrevar:
I've reported the Osicat warning here: https://github.com/osicat/osicat/issues/37
Maybe we can ignore conditions that match the "osicat" regexp.

Other than that, @jmercouris what do you think?  OK to merge?

jmercouris:
About the osicat failures, I don't know why they are sometimes appearing. I agree upon removing them, they are just a red herring.

source/minibuffer-prompt.lisp
```diff
@@ -1,10 +1,13 @@
...
+(defmacro define-function (name args docstring &body body)
+  "Eval ARGS and DOCSTRING then define function over the resulting lambda list
+and string.
 All ARGS are declared as `ignorable'."
-  (let ((evaluated-args (eval args)))
-    `(defun ,name  ,evaluated-args
+  (let ((evaluated-args (eval args))
+        (evaluated-docstring (eval docstring)))
+    `(defun ,name ,evaluated-args
+       ,evaluated-docstring
```
jmercouris, line 10: why are we evaluating docstrings?

jmercouris: so we can have dynamic docstrings?

jmercouris:
It looks like a nice improvement to me. This idea of evaluating docstrings has me thinking about the possibilities for define-command... maybe we can make a mini docstring language- maybe it would be overengineering. There is something there for sure. Feel free to merge whenever you are ready.

source/minibuffer-prompt.lisp
Ambrevar: Yes, now the `start` function embeds the help message displayed by `nyxt --help` :)

Ambrevar:
To be sure, you OK with starting nyxt with `(nyxt:start nil)` instead of `(nyxt:start)`?

jmercouris:
I assume you mean an empty list rather than nil? I am OK with it, but there must be a way to avoid it...

Ambrevar:
> I assume you mean an empty list rather than nil?

Yes, there are equivalent in Common Lisp, but `nil` can be faster to type
than `()` :)

> I am OK with it, but there must be a way to avoid it...

Having both &optional and &key is illegal.

One option would be to change the URLS positional argument to a key
argument.  So we would do this:

```
;; Start with default URL
(nyxt:start)
;; 
(nyxt:start
 :verbose t
 :urls '("http://..." "http://other.example")
 :no-init t)
```

The only constraint is that we can't have a "--urls" command line flag then.

Thoughts?

jmercouris:
I would prefer that I think. I don't ever use the command line arguments, and typically people only use them to open up one URL at a time from an external program, no?

Ambrevar:
Merged in 9c4121555943fc7d0ce7840a9249be53877cae2e.

jmercouris:
thank you!

jmercouris:
I just fixed the case of options and free args being nil failing on startup. (nyxt:start nil nil) does not work, so I handle it in entry-function. However, `(nyxt:start :no-init)` does not actually work. I also tried the following:

```
(if options
  (list (list options) :urls free-args)
  (list :urls free-args))
```

not sure why it is still complaining. It at least starts and accepts free args, so I am going to sleep for now.

Ambrevar:
The problem is that opts::*options* is set when compiled, but not when
the file is read.

Since `start' keyword were set when the file was read, they were empty
beside `:urls`.

Fixed now on master, sorry for the trouble :(

Issue #1040, opened 2020-11-14
Replace fullscreen commands with a toggler


Ambrevar:
I would save one more command in the list! :) 
It's also more consistent with `toogle-toolbars`.

@jmercouris OK with this change?

aadcg:
Though no one directly asked my opinion, I like this idea :)

jmercouris:
I agree as well!

PR #1045, opened 2020-11-17
Refactor minibuffer-(un)mark-all.

aadcg:
In the vein of #1040, I had this idea.

I know @Ambrevar is working on new minibuffer features so this might not be pertinent at the moment.

### DIFF
--- a/source/minibuffer-mode.lisp
+++ b/source/minibuffer-mode.lisp
@@ -52,7 +52,8 @@
        "shift-space" 'minibuffer-toggle-mark-backwards
        "M-space" 'minibuffer-toggle-mark
        "M-a" 'minibuffer-mark-all
-       "M-u" 'minibuffer-unmark-all))
+       "M-u" 'minibuffer-unmark-all
+       "M-m" 'minibuffer-toggle-mark-all))
     ;; TODO: We could have VI bindings for the minibuffer too.
     ;; But we need to make sure it's optional + to have an indicator
     ;; for the mode.
@@ -346,3 +347,24 @@ Only available if minibuffer `multi-selection-p' is non-nil."
       (setf nyxt::marked-suggestions (set-difference nyxt::marked-suggestions nyxt::suggestions)))
     (state-changed minibuffer)
     (update-display minibuffer)))
+
+(define-command minibuffer-toggle-mark-all (&optional
+                                            (minibuffer (current-minibuffer)))
+  "Toggle the mark over all visible suggestions.
+Only available if minibuffer `multi-selection-p' is non-nil."
+  (when (multi-selection-p minibuffer)
+    (with-slots ((suggestions nyxt::suggestions)
+                 (marked-suggestions nyxt::marked-suggestions))
+        minibuffer
+      (cond ((subsetp marked-suggestions suggestions)
+             (setf marked-suggestions
+                   (set-difference suggestions marked-suggestions)))
+            ((subsetp suggestions marked-suggestions)
+             (setf marked-suggestions
+                   (set-difference marked-suggestions suggestions)))
+            (t ; the intersection of suggestions and marked-suggestions
+               ; is non-trivial
+             (setf marked-suggestions
+                   (union suggestions marked-suggestions))))))
+  (state-changed minibuffer)
+  (update-display minibuffer))
--- a/source/tutorial.lisp
+++ b/source/tutorial.lisp
@@ -81,7 +81,10 @@ upon return.  The suggestion under the cursor is not processed if not marked.")
          ": Select all currently-displayed suggestions.")
     (:li (command-markup 'nyxt/minibuffer-mode:minibuffer-unmark-all
                          :modes (list (make-instance 'nyxt/minibuffer-mode:minibuffer-mode)))
-         ": Deselect all currently-displayed suggestions."))
+         ": Deselect all currently-displayed suggestions.")
+    (:li (command-markup 'nyxt/minibuffer-mode:minibuffer-toggle-mark-all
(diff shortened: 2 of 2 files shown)


jmercouris:
Hello, it is a most interesting idea! It makes sense from a consistency perspective. Here is what I percieve to be the biggest challenge:

Imagine this scenario:

+ The user opens the kill-buffers prompt
+ The user types "xyz"
+ The user toggles all of the matches for "xyz"
+ The user replaces the string "xyz" with "qwe"

At this point, how does the user additionally select all matches for "qwe"? If they press the toggle, it would logically "unselect" all entries, whereas if they could "add" and "remove" as discrete functions, the user can again press "select all" to also match all "qwe" entries.

If the above scenario is confusing, you may not be aware that the user can select some entries, type in more strings, select more entries, and ALL of the entries remain selected. Said another way; even if something is NOT visible in the current minibuffer results, it can remain selected.

Does that make sense? How could we overcome that?

aadcg:
Ahhh! Now finally I understand the use of `union` and `set-difference` in the previous implementation. 

Let me give some thought to it.

aadcg:
I came up with something that I find quite interesting actually.

The beauty is that we able able to match just like before (with multiple string replacements) with only one key, and at each iteration the user can take the complement. For instance:

- Open the kill-buffers prompt
- Type a string, say "github"
- Toggle all matches
- Delete the string
- Toggle and you get all buffers that were not matched by "github" (optional)
- Type another string
- Toggle all matches
- Take the complement of the previous 2 matches (optional)
etc.

Give it a try

source/minibuffer-mode.lisp
```diff
@@ -329,20 +328,22 @@ Only available if minibuffer `multi-selection-p' is non-nil.  DIRECTION can be
...
+      (if (subsetp marked-suggestions suggestions)
+          (setf marked-suggestions
+                (set-difference suggestions marked-suggestions))
+          (if (subsetp suggestions marked-suggestions)
+              (setf marked-suggestions
+                    (set-difference marked-suggestions suggestions))
+              ;; suggestions and marked-suggestions are disjoint sets at
+              ;; this point so take their union
+              (setf marked-suggestions
+                    (union suggestions marked-suggestions)))))
```
Ambrevar, line 347: I recomment you replace the nested `if` with a `cond`, it makes conditional switches more readable :)

Ambrevar:
Hi André, thanks for this feature, it's had been missing for surpisingly long, so this is welcome indeed!

Note that you can't remove `mark-all` however: say you select 3-4 entries by hand, then change your mind, how would you mark everything?  You'd have to unmark the 3-4 entries... :/  

With both `toggle-mark` and `mark-all`, `unmark-all` would technically be optional since it can be emulated with `mark-all + toggle-mark`.  But I believe some users would still love to have it as a shortcut, as a way to start from a clean slate.

Here is what I suggest: fix the `cond` nit, restore the mark-all and unmark-all commands, then I'll merge!

aadcg:

> Note that you can't remove mark-all however: say you select 3-4 entries
> by hand, then change your mind, how would you mark everything? You'd
> have to unmark the 3-4 entries... :/
>
> With both toggle-mark and mark-all, unmark-all would technically be
> optional since it can be emulated with mark-all + toggle-mark. But I
> believe some users would still love to have it as a shortcut, as a way
> to start from a clean slate.

I tried to do the most possible with a single toggle, but I agree with
you.

What about keeping toggle-mark and unmark-all, and dump mark-all?  The
former would be binded to M-a and the latter to M-e.  Indeed, an extra
keypress would be needed to emulate mark-all (unmark-all followed by
toggle-mark).  On the other hand, toggle-mark would do smth more complex
than mark-all.

> Here is what I suggest: fix the cond nit, restore the mark-all and
> unmark-all commands, then I'll merge!

What I mentioned above would be my own config , so I'd be happy to
follow your lead.

What about keybindings?  Or maybe toggle-mark shouldn't be binded by
default at all? @ambrevar @jmercouris



P.S. - I've been going down the Emacs rabbit hole lately---I switched
to EXWM.  Sorry for my late reply and overall silence.

Ambrevar:
Offering commands and bindings comes mostly at a single cost: we need to fulfill
a backward compatibility promise.

In this particular case, I believe exposing the 3 marking commands is
unambiguous and won't cause any backward compatibility trouble in the future.

Since the 3rd command can be implemented by calling the other 2, it comes for
free.

Conclusion: I'd rather keep the 3 commands :)  After all, users can disable any
binding they want, so this is not a big deal.

Regarding the bindings: in Helm, toggle-all-marks is bound to `M-m`.

P.S.: Welcome to EXWM! :)

aadcg:

> Since the 3rd command can be implemented by calling the other 2, it
> comes for
> free.

That would be an option indeed.  That wasn't what I had in mind tho.  I
wanted to easily take complements.  For instance, I can easily kill all
buffers that don't match some strings.

> Conclusion: I'd rather keep the 3 commands :) After all, users can
> disable any
> binding they want, so this is not a big deal.

Sure!

> Regarding the bindings: in Helm, toggle-all-marks is bound to M-m.

Ah, so such a feature exists (I never used Helm; Ivy user here)!  I took
your suggestion to bind the toggle to M-m.

Let me know if what I did looks fine.  I forced-pushed my commit so that
my previous work got overwritten.  I'm not sure if that goes against
good practices, so please let me know!  My naive (perhaps wrong) idea
was to make it easier for you.

> P.S.: Welcome to EXWM! :)

Thanks!  Once again re-learning and "re-muscle memorizing" a bunch of
stuff :D

Ambrevar:
Merged as 8bab7db41a5020782bf4576981ecabfae210d88a, thanks a lot!

PR #1047, opened 2020-11-19
Calispel: Proper concurrency in Nyxt at last!

Ambrevar:
I've replaced the ChanL library with Calispel which seems much, much better!

See https://github.com/CodyReichert/awesome-cl/issues/290#issuecomment-726032117 for a discussion.

I took the opportunity to (hopefully) fix the startup race condition.  @aartaka Can you make sure this branch starts properly for you?

Finally, I've made use of `fair-alt` instead of the interrupt-channel juggling business which was ugly and hard to understand.
This `fair-alt` will become essential in the new minibuffer library.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -215,12 +215,11 @@ editor executable."))
       (hooks:run-hook *after-init-hook*)
     (error (c)
       (log:error "In *after-init-hook*: ~a" c)))
-  (eager-future2:pexec ()
-    (funcall-safely (startup-function browser) urls)
-    ;; Set 'init-time at the end of finalize to take the complete startup time
-    ;; into account.
-    (setf (slot-value *browser* 'init-time)
-          (local-time:timestamp-difference (local-time:now) startup-timestamp))))
+  (funcall-safely (startup-function browser) urls)
+  ;; Set 'init-time at the end of finalize to take the complete startup time
+  ;; into account.
+  (setf (slot-value *browser* 'init-time)
+        (local-time:timestamp-difference (local-time:now) startup-timestamp)))
 
 ;; Catch a common case for a better error message.
 (defmethod buffers :before ((browser t))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -215,7 +215,7 @@ editor executable."))
       (hooks:run-hook *after-init-hook*)
     (error (c)
       (log:error "In *after-init-hook*: ~a" c)))
-  (chanl:pexec ()
+  (eager-future2:pexec ()
     (funcall-safely (startup-function browser) urls)
     ;; Set 'init-time at the end of finalize to take the complete startup time
     ;; into account.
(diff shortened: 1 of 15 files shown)


aartaka:
This branch starts properly for me.

Ambrevar:
Yay!

jmercouris:
Branch appears to work for me too! woohoo!

First bug report: Reduce to buffer command will activate, but will not show the minibuffer. I have to cancel the command, try to reactivate it, and then sometimes it shows. I am not sure what the sequence of events are to get it work.

Ambrevar:
It's the same issue with all chained minibuffers.

Try:

- M-x open-file RET

The current workaround: Bind the command ;)

This issue is also on master, but the race condition happens way less
often.

It's actually good we have Calispel now so we can more reliable test if
we've fixed this issue properly.

My guess is that `hide` is run too late and hides the new minibuffer.
It's probably easy to fix this, I'll work on it later.

jmercouris:
Ah, yes! I remember we had discussed a similar bug at some point in the recent past.

jmercouris:
Perhaps it was the very same bug.

Ambrevar:
To clarify, this race condition only occurs _frequently_ (about 3 times
out of 4 for me) with Calispel.

Ambrevar:
Before marging:

- Fix aforementioned minibuffer issue.
- Package Calispel for Guix.
- Test for potential global dead lock when multiple threads are listening.
- Test Calispel on #1051.

Ambrevar:
- [X] Global dead lock does not seem to be a problem.

- [X] I've just switched to Calispel in #1051.

Ambrevar:
- [X] Minibuffer race condition should now be fixed.

source/browser.lisp
```diff
@@ -215,12 +215,11 @@ editor executable."))
       (hooks:run-hook *after-init-hook*)
     (error (c)
       (log:error "In *after-init-hook*: ~a" c)))
-  (chanl:pexec ()
```
aartaka, line 218: Another use of `chanl:pexec` is replaced with new `nyxt:pexec`. Why isn't it the case there? Should it be documented?

Ambrevar: Sorry, I don't understand.  Where isn't it the case?  `nyxt:pexec` is just a dumb convenience wrapper around `bt:make-thread`.

Ambrevar: Oh!  I see what you mean: this is `pexec` was the race condition you suffered from for weeks! :)  It was wrongly introduced during the sync minibuffer experiment and I had forgotten to remove it.

Ambrevar:
- [X] Package Calispel for Guix.

All done.

Has anyone experienced any other issues with this patch?

Ambrevar:
Any other review?  Did anyone have the chance to look at the code?
Let me know if this is OK to merge.

Ambrevar:
@jmercouris: Can you test this on macOS?  Thanks!

jmercouris:
Works on macOS! Merge when ready!

Ambrevar:
Merged with c0b6b9e8be7536638196fef86d3aeaef3b342d1a
Thank you all for testing!

PR #1007, opened 2020-10-20
Global history tree

aartaka:
This changes the history to a more coherent and buffer-oriented model, turning browser history into one huge tree by default.

### Motivation

This change is to make history handling coherent with recent changes (#910) and to open a way for new features (#247 (when we store session, we won't capture the history of the incognito buffers anymore, because these will have buffer-isolated history), #854, #958).

Adding to that, there is a change in user experience -- buffer-local histories are now connected to each other and do branch out from each other. This means that if you open a link in a new buffer, you'll be able to return to original page even in this new buffer. It's not conventional, but it's productive, and there are examples of extensions to other browsers ([Firefox](https://addons.mozilla.org/en-US/firefox/addon/tree-style-tab/) and [Chrome](https://chrome.google.com/webstore/detail/tree-style-tab/oicakdoenlelpjnkoljnaakdofplkgnd)) that organize tabs/histories into a tree for the sake of productive navigation.

### What's New

- History data (i.e. the thing accessed by `(get-data (history-path (current-buffer)))`) is a `htree`.
- `buffer-description` is deleted in favor of `history-entry`.
- `web-mode` relies on the global history, instead of having its own.
- Some utilities (e.g. `map-tree`, `do-tree`, `remove-node`, `find-node`, `find-data`, `delete-data`, `add-children`) are added to the `htree` code.
- Old history storage format is supported with close-to-intuitive importers (session is not imported because lots of deleted code it relies on).
- `(re)store` methods can have arbitrary keyword arguments.
- Finally, `history-tree` command to see the full tree in it's grotesque beauty xD
- `history-tree` and `buffer-history-tree` are now colored/glyphed (with `display-buffer-id-glyphs-p` set to non-nil) to highlight the current buffer.
- `history-list` to see the history as a list.
- `web-mode` configuration slots for history navigation:
    - `history-forwards-prompting-p` -- whether `history-forwards` prompts the user on which branch to take in case there are several. Defaults to t.
    - `history-forwards-to-dead-history-p` -- whether `history-forwars` considers the dead history nodes worth visiting. Defaults to nil.
    - `conservative-history-movement-p` -- whether history nodes that don't belong to the current buffer are considerend visitable, i.e. conservative buffer-bounded navigation that we had before the patch. Defaults to nil.

### Caveats

- Our format for history serialization becomes unreadable with this patch. Maybe write a custom (de)serialization code for `htree`? 

### How Have This Been Tested.

- Nyxt runs well on this branch.

How does that look to you?

[...]

PR #1049, opened 2020-11-21
Avoid tautological buffer-load.

aartaka:
This rewrites most places of `buffer-load` usage with `:buffer` keyword as
tautological (because `make-buffer` creating the buffer passed with
`:buffer` keyword calls `buffer-load`). `make-buffer` and `make-buffer-focus`
are used instead.

### Motivation

While this doesn't change things dramatically in any way, it make for
a better code and architecture with less nested and repetitive
function calls :)

This also helps in maintaining a coherent history tree in #1007.

### How Has This Been Tested

I've ran most of the functions covered (except for the ones I don't
quite get the use-case of, like `paste-or-set-url`), so it should be
safe to merge.

How does it look to you?

EDIT: mark function names and keywords with backticks.

### DIFF
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -321,8 +321,7 @@ rest in background buffers."
                   :multi-selection-p t)))
     (dolist (entry (rest entries))
       (make-buffer :url (object-string (url entry))))
-    (buffer-load (url (first entries))
-                 :buffer (make-buffer-focus))))
+    (make-buffer-focus :url (url (first entries)))))
 
 (defmethod serialize-object ((entry bookmark-entry) stream)
   (unless (url-empty-p (url entry))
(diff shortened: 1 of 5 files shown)


aartaka:
Hmmmmmmm... Build checks look weird, but I'll test it on CCL to make sure things are working.

aartaka:
What's more weird is that re-running the checks have resolved the warnings. I guess it was some dependency (`osicat`) problem rather than the patch breakage. In any way, feel free to merge it anytime :)

Ambrevar:
We've noticed the Osicat issue a few times, it's annoying indeed.
I've reported to Osicat: https://github.com/osicat/osicat/issues/37.
Maybe we should ask the CCL people.
A quick workaround would be to filter out the Osicat warning.

Ambrevar:
Excellent, thanks!

PR #1050, opened 2020-11-21
Add ensured default value to with-data-access.

aartaka:
This adds the optional parameter to `with-data-access`. This parameter
specifies what value to assign to the given variable if data is not
found.

### Motivation

It can be useful to set the value to some initial value after getting
empty data, like it can be useful to set the `gethash`-ed value in a
hash-table to some sensible default. All in all, this change makes
`with-data-access` capable of `ensure-gethash`-like functionality on
user-data.

### Caveats (?)

One important addition is that all the arguments of `with-data-access`
are now parenthesised -- to ease argument parsing. This, however,
looks similar to `let`-bindings and other `with-*` macros, so it shouldn't
be hard to grasp for a human reader.

### How Has This Been Tested

- `asdf:test-system` 
- I've also ran Nyxt for a short time to check whether the functions
  relying on `with-data-access` do work. They do.

What do you think of this?

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -282,7 +282,7 @@ For the storage format see the comment in the head of your `auto-mode-rules-data
                           (values list &optional))
                 add-modes-to-auto-mode-rules))
 (defun add-modes-to-auto-mode-rules (test &key (append-p nil) exclude include (exact-p nil))
-  (with-data-access rules (auto-mode-rules-path (current-buffer))
+  (with-data-access (rules (auto-mode-rules-path (current-buffer)))
     (let* ((rule (or (find test rules
                            :key #'test :test #'equal)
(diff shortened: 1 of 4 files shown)


aartaka:
Maybe this initial value should be a keyword argument? This way it will be clear what the value means and the `with-data-access` will look like `with-open-file` (that we can mimic to reinforce understanding through association).

Adding to that, altering new keyword arguments will be easier than altering positional ones, if we'd need to.

Ambrevar:
I agree that a keyword argument would be more readable.
Can you make the change?

Ambrevar:
Looking good otherwise! :)

aartaka:
Here you go. `:default` is the keyword. It looks readable and short enough to me, while reminding of default argument to gethash.

Ambrevar:
Thanks!

PR #1051, opened 2020-11-21
Ospama: Guix server + full Scheme support

Ambrevar:
While this patch does not add anything user-facing, it's very interesting on the development front.

- I manage the `guix repl` command as a server.  Check out how I combine `uiop:launch-program`, `bt:with-timeout` and channels to do this.  The timeout ensures we don't leave a dangling Guix process behind.  The benefit of a "server"?  Performance. 
  - Before: 50ms overhead per Guix request.  
  - After: 50ms on startup, 0ms overhead afterwards.
  In short, if I run 10 requests in a row, previously it would take 500ms while it only takes 50ms now, assuming the requests are trivial.
Bonus: no more need for intermediary temp files.

- I now parse the return values of the "guix server" instead of parsing the standard output.  So we don't have to `(format t ...)` the results anymore, which makes for much more idiomatic Scheme.

- I use named-readtable to read and write Scheme directly from Common Lisp.  No more need to escape `#t` or `#f` and we can use Scheme keywords directly (e.g. `#:foo`).  This also helps with portability since we don't have need compiler-dependent cl->scheme string transformations anymore.

### DIFF
--- a/libraries/ospama/ospama-guix.lisp
+++ b/libraries/ospama/ospama-guix.lisp
@@ -27,47 +27,120 @@ package managers like Nix or Guix."))
 
 (detect-manager "guix" 'guix-manager)
 
-;; TODO: Call read-from-string in `guix-eval'?
 (declaim (ftype (function (t) string)))
 (defun cl->scheme-syntax (form)
-  #+ccl
-  ;; Escape symbols are printed as \\NAME while they should be printed as NAME.
-  (str:replace-all "\\" "" (format nil "~s" form))
-  #+(not ccl)
-  ;; Escaped symbols (e.g. '\#t) are printed as '|NAME| but should be
-  ;; printed as NAME.
-  (ppcre:regex-replace-all "'\\|([^|]*)\\|" (format nil "~s" form) "\\1"))
+  (str:replace-all
+   ;; Backslashes in Common Lisp are doubled, unlike Guile.
+   "\\\\" "\\"
+   #+ccl
+   ;; Escape symbols are printed as \\NAME while they should be printed as NAME.
+   (str:replace-all "\\" "" (format nil "~s" form))
+   #+(not ccl)
+   ;; Escaped symbols (e.g. '\#t) are printed as '|NAME| but should be
+   ;; printed as NAME.
+   (ppcre:regex-replace-all "'\\|([^|]*)\\|" (format nil "~s" form) "\\1")))
+
+(defvar *guix-repl-idle-timeout* 60
+  "Time in seconds of being idle before the `guix repl` is exited.")
+
+(defvar *guix-command* "guix"
+  "Name or path to the `guix' executable.")
+
+(defvar %guix-listener-channel nil)
+(defvar %guix-result-channel nil)
+
+;; TODO: Make sure Guix process is closed when Lisp shuts down.
+
+(defun guix-listener ()
+  "Automatically start and quit the Guix REPL after
+`*guix-repl-idle-timeout*'.
+For each inputs on `%guix-listener-channel' a result is returned on
+`%guix-result-channel'."
+  (flet ((maybe-start-guix (guix-process)
+           (unless (and guix-process
+                        (uiop:process-alive-p guix-process))
+             (setf guix-process
+                   (uiop:launch-program `(,*guix-command* "repl" "--type=machine")
+                                        :input :stream :output :stream))
+             ;; Skip REPL header.
+             ;; We could use `read' but CCL won't swallow the linebreak, while
+             ;; SBCL will.
+             (read-line (uiop:process-info-output guix-process) nil :eof))
+           guix-process))
+    (do ((guix-process nil)) (nil)
+      (handler-case
+          (do () (nil)
+            (let ((input (bt:with-timeout (*guix-repl-idle-timeout*)
+                           (chanl:recv %guix-listener-channel))))
+              (setf guix-process (maybe-start-guix guix-process))
+              ;; Append a newline so that REPL proceeds.
+              (format (uiop:process-info-input guix-process) "~a~%" input)
+              (finish-output (uiop:process-info-input guix-process))
+              ;; We set the package to current so that symbols in
+              ;;   (values (value ...) (value ...) ...)
+              ;; do not get returned with a package prefix.
+              (let* ((*package* (find-package :ospama))
+                     (output (read-line (uiop:process-info-output guix-process)
+                                        nil :eof)))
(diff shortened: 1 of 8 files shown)


Ambrevar:
I forgot to highlight one big collateral progress: the "guix server" I've implemented here is essentially the same as having a "guix comint mode" except that we only have commands and no interactive buffers.
In other words, this paves the way for a future `comint-mode` in Nyxt, where we can run REPLs for languages like Scheme, Ruby, Python...

Ambrevar:
Did anyone have time to review this?

jmercouris:
Not yet! I can review this today if you like

Ambrevar:
Please do, thanks!

libraries/ospama/scheme-syntax.lisp
```diff
@@ -0,0 +1,49 @@
...
+
+(in-package :named-readtables)
+
+(defreadtable ospama::scheme-reader-syntax
+  (:merge :standard)
+  ;; TODO: While Scheme is case sensitive, preserving the case would mean we'd
+  ;; have to upcase all return value symbols.  Or is there a smarter way to "do
+  ;; what I mean"?
+  ;; (:case :preserve)
+  (:macro-char #\[ #'(lambda (stream char)
```
jmercouris, line 12: Perhaps for the context of this library it is best to pretend that scheme is always upcased?

libraries/ospama/ospama-guix.lisp
```diff
@@ -27,241 +27,122 @@ package managers like Nix or Guix."))
...
+(defvar *guix-repl-idle-timeout* 60
+  "Time in seconds of being idle before the `guix repl` is exited.")
+
+(defvar *guix-command* "guix"
+  "Name or path to the `guix' executable.")
+
+(defvar %guix-listener-channel nil)
+(defvar %guix-result-channel nil)
+
+;; TODO: Make sure Guix process is closed when Lisp shuts down.
```
jmercouris, line 54: How could this be achieved? A hook to our quit command? What if Nyxt is killed ungracefully? A timeout?

jmercouris: The timeout would have to be within the Guix process, how could that be done?

libraries/ospama/ospama-guix.lisp
```diff
@@ -27,241 +27,122 @@ package managers like Nix or Guix."))
...
+Return the REPL result of the last form.
+On REPL exception (or unexpected form), return NIL and the form as second
+value.
+#<unspecified> objects are returns as NIL."
+  (unless (and %guix-listener-channel %guix-result-channel)
+    (setf %guix-listener-channel (make-instance 'calispel:channel))
+    (setf %guix-result-channel (make-instance 'calispel:channel))
+    (bt:make-thread #'guix-listener))
+  ;; Need to be in this package to avoid prefixing symbols with current package.
+  (let* ((*package* (find-package :ospama))
```
jmercouris, line 112: I don't believe this works. The first thing I would change is: `(find-package 'ospama)

jmercouris:
```
CL-USER> (defpackage :salmon (:use :common-lisp))
#<PACKAGE "SALMON">
CL-USER> (in-package :salmon)
#<PACKAGE "SALMON">
SALMON> (defun swim () (print "I am swimming"))
SWIM
SALMON> (swim)

"I am swimming" 
"I am swimming"
SALMON> (in-package :cl-user)
#<PACKAGE "COMMON-LISP-USER">
CL-USER> (salmon::swim)

"I am swimming" 
"I am swimming"
CL-USER> (let ((*package* (find-package 'salmon)))
           (print *package*)
           (swim))
; in: LET ((*PACKAGE* (FIND-PACKAGE 'SALMON)))
;     (SWIM)
; 
; caught STYLE-WARNING:
;   undefined function: COMMON-LISP-USER::SWIM
; 
; compilation unit finished
;   Undefined function:
;     SWIM

#<PACKAGE "SALMON"> 

; No value
CL-USER> 
```

jmercouris:
It looks quite close to me! I can't make any comments on the Scheme code, but I have reviewed the Lisp code as much as is possible without being able to run it.

libraries/ospama/scheme-syntax.lisp
Ambrevar: `(:case :preserve)` is commented, so we are indeed not case-sensitive here.

libraries/ospama/ospama-guix.lisp
Ambrevar: First question: this is a general Lisp issue, `uiop:launch-program` has the drawback of forking to PID 1 when the Lisp parent dies.

Second question: this is really smart, I hadn't thought of it!  Indeed, I can add a timer to Guix so the REPL automatically kills itself on timeout.

libraries/ospama/ospama-guix.lisp
Ambrevar: It works indeed, we are already using this technique in auto-mode.lisp, history.lisp, session.lisp.

Let-binding `*package*` only has an effect on the _reader_ (and the printer), not on the evaluator.  In your example, `swim` was read in the cl-user package, which is why `cl-user::swim` is evaluated even when `*package*` is changed locally.

To see an example of the impact this let-binding has on the reader, consider this:

```lisp
CL-USER> (let* ((*package* (find-package :nyxt)))
					 (read-from-string "foo"))
NYXT::FOO
3 (2 bits, #x3, #o3, #b11)

```

jmercouris: Ah, OK, I misunderstood how you were using it!

libraries/ospama/ospama-guix.lisp
Ambrevar: Found a better way thanks to @phoe: https://old.reddit.com/r/Common_Lisp/comments/k231ai/kill_uioplaunchprogram_process_on_exit/

In short, we've got to wrap the `guix repl ...` command in a shell call, like this:

```lisp
  (uiop:launch-program "set -o monitor; sleep 22 & read dummy; kill %1" :input :stream)
```

Ambrevar: But actually, since `guix repl` read on a pipe which is closed when the parent process dies, it quits automatically.  So there is no need for the above trick, it all works as expected.

That said, this trick is good to know!

Ambrevar:
Merged in d00e426800aa62394a0e1a1875fa6bdef41a7c33.

PR #1052, opened 2020-11-23
Nix

bqv:
@jmercouris

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ Nyxt.app
 
 # Generated doc
 manual.html
+
+# Ignore flake lockfile
+flake.lock
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,15 @@
+{ pkgs ? import <nixpkgs> {}
+, src ? ./.
+, version ? "nix"
+, ... }:
+
+pkgs.nyxt.overrideAttrs (drv: {
+  src = drv.src.overrideAttrs (drv: {
+    inherit src;
+    name = pkgs.lib.replaceStrings [drv.meta.version] [version] drv.name;
+
+    # Pending nixpkgs changes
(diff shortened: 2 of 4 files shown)


jmercouris:
Thank you, will take a look!

jmercouris:
Would it be possible to move this into the build-scripts?

bqv:
All three of those files are meant to be in the root, really. If you were desperate you can delete shell.nix, it's not really necessary here

jmercouris:
I was afraid that would be the answer... I tried moving it myself and experienced some problems.

aartaka:
I may be out of context here, but can symlinking help in keeping things in the `build-scripts/`? Like
- Have all the three new files in the `build-scripts/`.
- Have a `nix-install.sh` script that will:
    - Symlink/copy all the three files to root.
    - Run `nix-env` (or whatever it is that installs packages on Nix) on these files.
    - Do cleanup when necessary.

Does that fit the Nix style? Shell scripts are discouraged in functional world, so I might be suggesting nonsense :)

bqv:
The main issue is that for the flake.nix to work, it needs to be committed at the root of the git repository, so there's no real messing with that. The default.nix is less of an issue, but given people tend to like to `import ./.` and `nix build` in the root, it breaks a bit of ergonomicity. The shell.nix is pretty unnecessary as i mentioned, especially given none of you use nix to dev

jmercouris:
I use NixOS to develop, I am only on the system about 1 month now though, so I am quite ignorant still. I use the shell.nix file located in our build-scripts directory.

bqv:
Fixed the missing calispel dependency by using lisp-overlay

jmercouris:
Will test again today, thank you!

jmercouris:
Thank you!

Issue #777, opened 2020-06-04
Minibuffer overhaul


Ambrevar:
The minibuffer needs a rewrite to fix a couple of current issues and implement
many core features.

The code needs to be extracted:

- A text editing library (see what Lem has to offer).
- A "readline" library, that does the candidate suggesting, fuzzy matching,
  sorting, etc.  It would be heavily inspired by Emacs Helm.
- Finally, the minibuffer rendering in Next.

One of the biggest flaws of our current implementation is that it uses
Javascript to know which candidate is selected.  If I'm not mistaken, there is
no need for that.  Instead, would could do all the logic in the readline
library, only using HTML to render.
A great benefit of this is that the minibuffer queries would not be asynchronous
anymore.  In particular, instead of the convoluted

```lisp
(with-result (result (read-from-minibuffer
                      (make-minibuffer ...))
  (use result here)))
```

we could have

```lisp
(let ((result (minibuffer-read ...)))
  (use result here))
```

Another thing that can be improved is the vocabulary:

- `candidate` or `suggestion`?
- `mark` or `select`?
- The currently highlighted item could be called `current candidate` or `current
  suggestion`.
- Now we have `select-next` and `select-previous` to shift the current
  suggestion to the nexet.  Since it's ambiguous (`select` has the meaning of
  `mark`), I suggest to simply name it `next`.  With the package prefix, that
  would be `readline:next`.

New features:

- Support paging (https://github.com/atlas-engineer/next/issues/454).  This
  means we need to know the number of suggestions currently displayed.  It's not
  trivial, especially if the suggestion contains a picture for instance.  This
  is why each suggestion item needs to report how much space it takes, but
  horizontally and vertically.

- Support multi-columns: currently `execute-command` has 3 columns and we want
  to add a 4th one (https://github.com/atlas-engineer/next/issues/624).
  For the sake a clear display and better control, we need to allow for
  tabulated display and add a command to allow the user to choose which column
  to toggle.
  Also add a column to choose which columns (yes, plural!) to sort by.
  
- Support filters.  Find a way or an input syntax to filter suggestions
  depending on their characteristic.  The filtering should be done by a callback
  in the suggestion itself, so that each different suggestion type understands
  their own filtering.
  
- Support multiple actions.  Each suggestion should specify a list of actions to
  operate on it.  Hitting `return` runs the default action, but hitting some
  other key allows for choosing the action.
  
- Support multiple sources.  With multiple action support we could mix-match
[...]

Issue #1005, opened 2020-10-20
Make mode constructors/destructors composable


aartaka:
One may need to run the constructors of parent modes in while initializing child classes. 

For example: 
It is a good practice in Nyxt to create small modes for settings-toggling, like `noimage-mode` and `noscript-mode`. I'm thinking about `noclipboard-mode` to disable clipboard access for JS. This should be a separate mode. But it should be part of `reduce-tracking-mode` too, because clipboard access is a way of tracking and breaching one's privacy and security.

noclipboard-mode.lisp (not an actual code, no renderer abstraction):
``` lisp
(define-mode noclipboard-mode ()
  "Disable JavaScript access to clipboard."
  ((destructor
    (lambda (mode)
      (setf (webkit:webkit-settings-javascript-can-access-clipboard
             (webkit:webkit-web-view-get-settings (gtk-object buffer)))
            t)))
   (constructor
    (lambda (mode)
      (setf (webkit:webkit-settings-javascript-can-access-clipboard
             (webkit:webkit-web-view-get-settings (gtk-object buffer)))
            nil)))))
```

reduce-tracking-mode.lisp:
``` lisp
(define-mode reduce-tracking-mode (noclipboard-mode root-mode)
  "Set specific settings in the web view in order to mitigate fingerprinting,
(how third-party trackers attempt to indentify you."
  ((preferred-languages '("en_US")
                        :type list-of-strings
                        :documentation "The list of languages that will be sent as
part of the Accept-Language HTTP header.")
   (destructor
   ;;; noclipboard-mode's destructor should run too, this code shouldn't be there
   ;; (setf (webkit:webkit-settings-javascript-can-access-clipboard
   ;;        (webkit:webkit-web-view-get-settings (gtk-object buffer)))
   ;;       t)))
    (lambda (mode)
      (ffi-set-preferred-languages (buffer mode)
                                   (list (first
                                          (str:split
                                           "."
                                           (or (uiop:getenv "LANG") "")))))))
   (constructor
   ;;; noclipboard-mode's constructor should run too, this code shouldn't be there
   ;; (setf (webkit:webkit-settings-javascript-can-access-clipboard
   ;;        (webkit:webkit-web-view-get-settings (gtk-object buffer)))
   ;;       nil)))
    (lambda (mode)
      (ffi-set-preferred-languages (buffer mode)
                                   (preferred-languages mode))))))
```

[...]

PR #1054, opened 2020-11-27
WIP: Minibuffer overhaul

Ambrevar:
This is an attempt to address #777 and most (all?) minibuffer-related issues.

First step is to implement the "backend": anything that's not display related, roughly reimplementing Emacs Helm.

Second step is to integrate it in Nyxt by implementing the frontend.

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -86,13 +86,11 @@ another.")
                   "Function called with the source as argument.
 It is useful for instance to create the list of `initial-suggestions'.")
 
-     (destructor nil                    ; TODO: Use it.
+     (destructor nil
                  :type (or null function)
                  :documentation
                  "Function called with the source as parameter to clean it up.
-It's called when `destructor' is called over `prompter'.
-
-Note that the function is executed *before* performing any action.")
+It's called when `destroy' is called over `prompter'.")
 
      (initial-suggestions '()
                           :reader initial-suggestions
@@ -162,7 +160,7 @@ filtered.  The predicate works the same as the `sort' predicate.")
               :documentation "List of functions that can be run on suggestions
 of this source.")
 
-     (persistent-action nil             ; TODO: Implement.
+     (persistent-action nil ; TODO: Should be a list so we can support as many persistent actions as we want.
                         :type (or null function)
                         :documentation
                         "Function called over the selection without returning
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -155,8 +155,8 @@ when input is modified, after filtering the suggestions.")
                      :documentation "A predicate used to sort the suggestions once
 filtered.  The predicate works the same as the `sort' predicate.")
 
-     (actions '()
-              :type (or null (cons function)) ; TODO: Accept function symbols?  Commands?
+     (actions '(identity)
+              :type (cons (or function symbol)) ; TODO: Accept function symbols?  Commands?
               :accessor nil
               :export nil
               :documentation "List of functions that can be run on suggestions
@@ -250,7 +250,7 @@ Also see `follow-delay'.")
                    "Execute `persistent-action' after this delay when `follow' is
 non-nil.")
 
-     (must-match-p :always
+     (must-match-p :always ; TODO: Remove and use dedicated source instead?  Then remove `return-input'.
                    :type (or must-match-choices null)
                    :documentation
                    "Control what to do when input does not match anything.
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -327,39 +327,43 @@ If a previous suggestion computation was not finished, it is forcefully terminat
              ;; TODO: This is prone to a race condition.
              (bt:thread-alive-p (update-thread source)))
     (bt:destroy-thread (update-thread source)))
-  (setf (update-thread source)
-        (bt:make-thread
-         (lambda ()
-           (let ((last-notification-time (get-internal-real-time))
-                 (preprocessed-suggestions (mapcar #'copy-object
-                                                   (initial-suggestions source))))
-             (setf preprocessed-suggestions
-                   (if (filter-preprocessor source)
-                       (maybe-funcall (filter-preprocessor source)
-                                      preprocessed-suggestions source input)
-                       (list (make-instance 'suggestion
-                                            :value input
-                                            :properties (maybe-funcall (suggestion-property-function source)
-                                                                       input)
-                                            :match-data ""))))
-             ;; TODO: Should we really reset the suggestions here?
-             (setf (slot-value source 'suggestions) '())
-             (if (or (str:empty? input)
-                     (not (functionp (filter source))))
-                 (setf (slot-value source 'suggestions) preprocessed-suggestions)
-                 (dolist (suggestion preprocessed-suggestions)
-                   (sera:and-let* ((processed-suggestion
-                                    (funcall (filter source) input suggestion)))
-                     (setf (slot-value source 'suggestions)
-                           (insert-item-at suggestion (sort-predicate source)
-                                           (suggestions source)))
-                     (let* ((now (get-internal-real-time))
-                            (duration (/ (- now last-notification-time)
-                                         internal-time-units-per-second)))
-                       (when (> duration (notification-delay source))
-                         (calispel:! (update-notifier source) t)
-                         (setf last-notification-time now)))))))
+  ;; We let-bind the channel so that if PROMPTER's channel is reset, we keep
+  ;; using the old one in this (outdated) thread.
+  ;; See (setf input).
+  (let ((ready-channel (ready-channel prompter)))
+    (setf (update-thread source)
+          (bt:make-thread
+           (lambda ()
+             (let ((last-notification-time (get-internal-real-time))
+                   (preprocessed-suggestions (mapcar #'copy-object
+                                                     (initial-suggestions source))))
+               (setf preprocessed-suggestions
+                     (if (filter-preprocessor source)
+                         (maybe-funcall (filter-preprocessor source)
+                                        preprocessed-suggestions source input)
+                         (list (make-instance 'suggestion
+                                              :value input
+                                              :properties (maybe-funcall (suggestion-property-function source)
+                                                                         input)
+                                              :match-data ""))))
+               ;; TODO: Should we really reset the suggestions here?
+               (setf (slot-value source 'suggestions) '())
+               (if (or (str:empty? input)
+                       (not (functionp (filter source))))
+                   (setf (slot-value source 'suggestions) preprocessed-suggestions)
+                   (dolist (suggestion preprocessed-suggestions)
+                     (sera:and-let* ((processed-suggestion
+                                      (funcall (filter source) input suggestion)))
+                       (setf (slot-value source 'suggestions)
+                             (insert-item-at suggestion (sort-predicate source)
+                                             (suggestions source)))
+                       (let* ((now (get-internal-real-time))
+                              (duration (/ (- now last-notification-time)
+                                           internal-time-units-per-second)))
+                         (when (> duration (notification-delay source))
+                           (calispel:! (update-notifier source) t)
+                           (setf last-notification-time now)))))))
 
-           ;; TODO: Pass `filter-preprocessor' result to source in case filter is not run?
-           (maybe-funcall (filter-postprocessor source) source input)
-           (calispel:! (ready-channel prompter) source)))))
+             ;; TODO: Pass `filter-preprocessor' result to source in case filter is not run?
+             (maybe-funcall (filter-postprocessor source) source input)
+             (calispel:! ready-channel source))))))
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -25,6 +25,7 @@
 (export-always 'object-properties)
 (defmethod object-properties ((object t))
   "Suitable as a `prompter-source' `suggestion-property-function'."
+  ;; TODO: Support objects and structs and expand their exported slot to a plist.
   (list :default (write-to-string object)))
 
 (define-class suggestion ()
@@ -51,7 +52,7 @@ The other slots are optional."))
 ;; the initargs of the class.
 (sera:eval-always
   (define-class prompter-source ()      ; TODO: Rename `source'?
-    ((name ""
+    ((name (error "Source must have a name")
            :documentation
            "Name which can be used to differentiate sources from one
 another.")
@@ -258,7 +259,8 @@ It can be a function of one argument, the prompter, which returns a string."))
   (apply #'make-instance 'prompter-source args))
 
 (define-class yes-no-source (prompter-source)
-  ((initial-suggestions '("yes" "no")))
+  ((name "Confirm")
+   (initial-suggestions '("yes" "no")))
   (:export-class-name-p t)
   (:accessor-name-transformer #'class*:name-identity)
   (:documentation "Prompt for yes-no questions."))
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -237,15 +237,15 @@ present.")
              "Function called with the source as argument when the prompter is
 resumed.")
 
-     (follow nil                        ; TODO: Implement.
-             :type boolean
-             :documentation
-             "When non-nil, automatically execute `persistent-action'.
+     (follow-p nil
+               :type boolean
+               :documentation
+               "When non-nil, automatically execute `persistent-action'.
 Also see `follow-delay'.")
 
-     (follow-delay 0.0                  ; TODO: Implement.
+     (follow-delay 0.0
                    :documentation
-                   "Execute `persistent-action' after this delay when `follow' is
+                   "Execute `persistent-action' after this delay when `follow-p' is
 non-nil.")
 
      (must-match-p :always ; TODO: Remove and use dedicated source instead?  Then remove `return-input'.
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -9,7 +9,6 @@
 ;; TODO: User classes?  Probably useful mostly for `prompter-source' since
 ;; they may be defined globally.  Conversely, `prompter' is mostly used
 ;; locally.
-;; TODO: Implement prompter actions.
 
 (deftype must-match-choices ()
   `(or (eql :always)
@@ -87,7 +86,7 @@ another.")
                   "Function called with the source as argument.
 It is useful for instance to create the list of `initial-suggestions'.")
 
-     (destructor nil
+     (destructor nil                    ; TODO: Use it.
                  :type (or null function)
                  :documentation
                  "Function called with the source as parameter to clean it up.
@@ -116,7 +115,7 @@ See `ready-notifier' to know when the list is final.
 See `update-notifier' to know when it has been updated, to avoid polling the
 list.")
 
-     (marked-suggestions '()            ; TODO: Implement.
+     (marked-suggestions '()
                          :documentation
                          "The list of suggestions which have been marked by the user.
 Marking is only allowed when `multi-selection-p' is non-nil.
@@ -156,7 +155,7 @@ when input is modified, after filtering the suggestions.")
                      :documentation "A predicate used to sort the suggestions once
 filtered.  The predicate works the same as the `sort' predicate.")
 
-     (actions '()                       ; TODO: Implement!
+     (actions '()
               :type (or null (cons function)) ; TODO: Accept function symbols?  Commands?
               :accessor nil
               :export nil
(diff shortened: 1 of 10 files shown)


Ambrevar:
This is based on the Calispel patchset (#1047) because I need it here.
Minibuffer-related changes start November 27.

See the tests for a nice use of Calispel with `fair-alt` to track suggestion computation progress.

Ambrevar:
For now there is nothing that can be used in Nyxt, however the backend data structures are ready for review.

Next steps:

- [ ] Backport `test-fuzzy.lisp` to make sure we have at least the same matching quality as master.
- [ ] Implement a new Nyxt command to use this new minibuffer, e.g. `set-url2`.

Ambrevar:
Any early code review is already welcome!  In particular, I'm happy to hear feedback about the class and slot names.

libraries/minibuffer/minibuffer-source.lisp
```diff
@@ -0,0 +1,356 @@
...
+used by the `sort-predicate'."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer #'class*:name-identity)
+  (:documentation "Suggestions are processed and listed in `minibuffer-source'.
+It wraps arbitrary object stored in the `value' slot.
+The other slots are optional."))
+
+;; We must eval the class at read-time because `make-source' is generated using
+;; the initargs of the class.
```
jmercouris, line 52: why do we need a make-source? what are the benefits over make-instance 'minibuffer-source and then having initialize-instance after?

libraries/minibuffer/minibuffer-source.lisp
```diff
@@ -0,0 +1,356 @@
...
+
+     (sort-predicate #'score>
+                     :type (or null function)
+                     :documentation "A predicate used to sort the suggestions once
+filtered.  The predicate works the same as the `sort' predicate.")
+
+     (actions '()                       ; TODO: Implement!
+              :type list)
+
+     (persistent-action nil
```
jmercouris, line 136: would just =action= be OK too? I think that would make more sense personally

libraries/minibuffer/minibuffer-source.lisp
```diff
@@ -0,0 +1,356 @@
...
+
+(export-always 'make-source)
+(define-function make-source
+    (append '(&rest args)
+            `(&key ,@(initargs 'minibuffer-source)))
+  "Return minibuffer source object."
+  (apply #'make-instance 'minibuffer-source args))
+
+(defmethod initialize-instance :after ((source minibuffer-source) &key)
+  ;; TODO: Should we always do this?  What if initial-suggestions are already
```
jmercouris, line 245: We can just check if they are or not, no? I guess I don't understand the problem indicated by the TODO here :-D

libraries/minibuffer/minibuffer.lisp
```diff
@@ -0,0 +1,111 @@
...
+(export-always 'cleanup)
+(defmethod cleanup ((minibuffer minibuffer))
+  "First call `before-cleanup', then clean up all sources, finally call
+`after-cleanup'."
+  (maybe-funcall (before-cleanup minibuffer))
+  (mapc #'cleanup (sources minibuffer))
+  (maybe-funcall (after-cleanup minibuffer)))
+
+(export-always 'ready?)
+(defun ready? (minibuffer &optional timeout)
```
jmercouris, line 99: I would prefer `readyp`

Ambrevar: Forgot to update this one.  I prefer `ready-p` over `readyp` though.

jmercouris:
Just some surface level comments for now, I haven't dug into the code yet :-)

libraries/minibuffer/minibuffer.lisp
jmercouris: Thats fine, the convention is a bit silly to me that single words don't get a dash and multiple words get a dash. I think it should just always be a dash.

libraries/minibuffer/minibuffer-source.lisp
Ambrevar: `initialize-instance' can be used regardless.
`make-source` is a tiny shorthand for `make-instance 'minibuffer-source`.  I figured that the user is going to use it a lot, so let's make this as short as possible.

libraries/minibuffer/minibuffer-source.lisp
Ambrevar: But having both `actions` and `action` is confusing.
Otherwise I can change `actions` to be a structure of

- list of actions
- default action
- persistent action

jmercouris: I had already assumed that actions was a list of action objects, what else would it be?

libraries/minibuffer/minibuffer-source.lisp
Ambrevar: The problem is that we change initarg data at initialization time, which may be surprising to the user.  This is not very good practice in terms of data consistency.

An alternative would be to have a separate slot, e.g. `initial-data`, and leave `initial-suggestions` unexported.

Another solution is to remove the initarg for `initial-suggestions` and add a new keyword argument `initial-data` to `make-source`.

libraries/minibuffer/minibuffer-source.lisp
jmercouris: OK, that is a good point. I assume you also want to have the key arguments for the user hence why we have to read the slots?

Ambrevar: Yes, that's the point of the `define-function` macro, it allows us to automatically generate the keyword arguments.

libraries/minibuffer/minibuffer-source.lisp
Ambrevar: It is, but it's a bit limited because a simple list of actions does not specify:

- the default action (when the user presses `return`)
- the persistent action (when follow-mode is on, or when the user presses a dedicated key).

So either we add slots to specify these two, or we set `actions` to be of the `action-list` type which has the aforementioned 3 slots.

jmercouris: or we could use a plist?

jmercouris: OR actions is actually an object and not a list, with methods to return the default action, the persistent action, etc.

Ambrevar: A plist could do, but the benefit of a dedicated object is that it catches error (plist accept arbitrary keys) which is very useful to catch typos in the user config.

About methods: the drawback is that it forces the user to define the action list globally to be able to decide which action is the default, which is much more verbose.  Compare this (untested) config sample:

```lisp
;; Without methods
(minibuffer:make
  :sources (list (minibuffer:make-source
                  :actions (minibuffer:make-actions
                            :default default-action
                            :persistent persistent-action
                            :others (list action-a action-b)))))

```

```lisp
;; With methods
(defvar *my-actions* (minibuffer:make-actions default-action
                                              persistent-action
                                              action-a action-b))

(defmethod default-action ((actions (eql *my-actions*)))
  (first (minibuffer:actions actions)))
(defmethod persistent-action ((actions (eql *my-actions*)))
  (second (minibuffer:actions actions)))

(minibuffer:make
 :sources (list minibuffer:make-source
                :actions *my-actions*))

```

Ambrevar:
I'm thinking of renaming `initializer` and `cleanup` to `constructor` and
`destructor` respectively.  This would be consistent with modes.

While we are at it, we can make methods for them as suggested in #1005.
We could still keep the slots which would allow for instance-specific overrides.

jmercouris:
I agree with that change

Ambrevar:
Note: make-source should return a lambda that creates minibuffer-source
objects.  Indeed, we don't want global, non-thread-safe minibuffer-source
object.
The source object should be instantiated with the minibuffer.

Ambrevar:
On second though, do we really need a macro to create top-level sources?
What about just using define-class?  For instance

``` common-lisp
(define-class minibuffer-foo-source (minibuffer-source)
  ((initial-suggestions '("foo" "bar"))
   (filter #'slow-identity-match)))

(minibuffer:make
 :sources (list (make-instance 'minibuffer-foo-source)))
```

Thought?

jmercouris:
define-class may suffice for the purpose, if you find yourself limited by it, we can always create a macro :-)

Ambrevar:
Acknowledged.

I tried getting input methods to work in the minibuffer.
The easiest way to test is with unicode input, e.g. press
"control-shift-u 1 2 3 4 space" and see if it inserts a unicode
character.

It does not work because in our gtk-entry workaround because we only have
gtk:gtk-entry-im-context-filter-keypress and no way to tell that some
input preedit is happening.

One way around it would be to define a custom GTK input method.  This
seems difficult to get right and it's unclear to me whether it will work
well with the various user settings (UIM, IBUS, etc.).

https://developer.gnome.org/gtk3/stable/GtkIMContext.html

So for now I'm leaning towards using a GtkTextInput widget instead of
our home-rolled HTML input.
The main reason we had this HTML input initially was because we used
Javascript to control the text editing features.  But now that we have
input-edit.lisp, we can perfectly edit any GTK text input.

Another thing to keep in mind is that GTK is capable of applying CSS
anywhere, so styling would not be an issue.

Thoughts?

jmercouris:
I am not a huge fan of it because one part of the minibuffer would be native widget, while the other would be a GTK widget. Also, the HTML affords us many benefits that a normal GtkTextInput widget does not. For example, we could implement Hydra really easily within the minibuffer right now. If we use a GtkTextInput, then we won't be able to. What if we had an invisible GtkTextInput that is focused and recieves input, and we just mirror the contents of that into our HTML view?

Ambrevar:
Can you explain why Hydra would be a problem with GtkTextInput?

> What if we had an invisible GtkTextInput that is focused and recieves input, and we just mirror the contents of that into our HTML view?

I'll try that.

Another benefit of using a native text widget is that it gives us
regions for free.

jmercouris:
A hydra would be a problem because:

1. it would be renderer dependent implementation
2. It would take up unresolvable space in the minibuffer

jmercouris:
Also, regions not existing is a limitation of our text-buffer library, we will have them eventually :-)!

Ambrevar:
> 1. it would be renderer dependent implementation
> 2. It would take up unresolvable space in the minibuffer

We can render the Hydra as part of the minibuffer suggestions, so this
won't be a problem.

Ambrevar:
See commit d5204f088fdee2b092ea4c4999335556ce65ef1f.

I tried what you suggested, it was a good idea but sdaly it does not seem to
work :/

I think it's because intercepting key events breaks the IM context.

Can you think of a way around this?

jmercouris:
If intercepting key events breaks the IM context, then it won't work no matter what. How does Emacs allow insertion of special symbols? We don't necessarily have to use the input manager, is there another way?

jmercouris:
Could we perahps toggle/untoggle intercepting the IM context? Have a sort of "raw input" button?

Ambrevar:

> If intercepting key events breaks the IM context, then it won't work
> no matter what.

Yes there is.  If we use a GTK widget, then instead of intercepting key
presses when the minibuffer input is focused, we hook on the widget
changes (like any other GUI would do) to update the candidates.

https://developer.gnome.org/gtk2/stable/GtkEntry.html#GtkEntry-insert-at-cursor

> How does Emacs allow insertion of special symbols? We don't necessarily have to use the input manager, is there another way?

Emacs has its own input system.  It's probably big :p

Rolling out our own costs time, it's hard to get right and finally it's
incompatible with what users settings.

Ambrevar:
> Could we perahps toggle/untoggle intercepting the IM context? Have a sort of "raw input" button?

Maybe there is a misunderstanding.  We don't intercept the IM context,
only the input events.

If I understand you right, you are saying that when we enter "preedit"
mode (e.g. when the input displays an underlined "u") then we stop
intercepting the input.

The problem is there is no way to distinguish a "preedit" IM input from
a regular input.

Ambrevar:
Beside the Hydra which is not really a problem, can you foresee another
issue with using GtkEntry as a minibuffer input?

jmercouris:
If a GTK input solves the problem, it shouldn't matter whether it is visible or not, only whether it is focused. Therefore it should be possible to mirror its contents into a string.

jmercouris:
Perhaps I don't fully understand the problem.

Ambrevar:
I see what you mean.  I'll try to stop intercepting keyboard events when
the minibuffer is up, see if that works.

jmercouris:
Yes, that is what I was suggesting :-)

Ambrevar:
Progress!  I found out why IM was broken before: it's because we are
also intercepting with on-signal-key-release-event.  Returning nil there
I can correctly input unicode with control+shift+u!

Now one obstacle remains: if we don't show the GtkEntry widget, we need
a way to display the "preedit" characters.

For instance if you press "control-shift-u 12" in any GTK application,
it should display "u12" underlined, waiting for you to complete the hex
string.

We want to also display this underlined "u12".  The problem is that the
GtkEntry text does not contain it, it seems to be only virtual.

NXYT> (gtk:gtk-entry-text   (dummy-text-input (current-window)))
""

I looked everywhere in
https://developer.gnome.org/gtk2/stable/GtkEntry.html
but I can't seem to find a way to access the preedit text.

Any clue?

jmercouris:
I think this is an OK bug to have (i.e., it is not implementation stopping). Do we have a way to sense a "preedit" char? If so, we can just use a universal "?" after the current char or something.

jmercouris:
Also, congratulations on the discovery!

Ambrevar:
Actually I think it's a dealbreaker.  When you write in CJK, the
character compose as you type and for those among us who are not
touch-typers, they won't be able to type anything :(
(Not mentioning newcomers won't understand what's going on if they type
and nothing happens beside a "?".)

So using a native widget seems to be the only way forward as I
understand it.

Anyways, I don't see any drawback.  Is there something else that bugs
you with native widgets?

jmercouris:
There must be a way... the native widget will cause many problems. Please do not move forward with a native widget, we can worry about input handling later. OK?

Ambrevar:
Which problems?
Can you elaborate?

Can you please help me with this? I'm running out of ideas :(

Ambrevar:
The new UI depends on how we deal with input, so we need to fix this
first before we can do anything else with the minibuffer overhaul.

jmercouris:
We can proceed with imperfect CJK input. Using the native UI widget will compromise the design in many ways (even though it does accept CSS), it will primarily compromise the code design and the concept of everything being a buffer.

jmercouris:
Would it be OK if we troubleshoot this together on Monday?

Ambrevar:
Can you give a precise example of what could be compromised?

After all, the message area and the status area are not buffers. 
The only time we leverage the fact that the minibuffer is a buffer is
for modes, and that's not a problem, mode can perfectly be added to a
minibuffer that has a native widget.

So yes, we can keep the minibuffer as a child of the "buffer" class.

If you want, I can work on a proof of concept if that can convince you.

There is something else we could do: use an actual HTML input tag!
We would get input method support for free since it's WebKitGTK.
I just don't know how to capture an event when the input has been modified.

jmercouris:
I would like the buffer to *only* be a webview. When you add more things into the mix, it starts getting complex. A HTML Input tag may be the solution...

jmercouris:
A problem of the native view is that we cannot easiily displace it without making a bunch of exceptions in the codebase. For example, the hydra still does *not* work because there would be a undeletable section of the top of the minibuffer which would be a minibuffer input area.

Ambrevar:
> A problem of the native view is that we cannot easiily displace it without making a bunch of exceptions in the codebase. For example, the hydra still does *not* work because there would be a undeletable section of the top of the minibuffer which would be a minibuffer input area.

You mean something like Magit' transcient where there is no input line?

We can make the input line a togglable option, that would fix it.  It's
that easy! :)

jmercouris:
> A problem of the native view is that we cannot easiily displace it without making a bunch of exceptions in the codebase. For example, the hydra still does *not* work because there would be a undeletable section of the top of the minibuffer which would be a minibuffer input area.
> You mean something like Magit' transcient where there is no input line? We can make the input line a togglable option, that would fix it. It's that easy! :)

That just doesn't feel right to me. It would feel like a kludge of our design. A work-around...

jmercouris:
To trigger a LISP event when an HTML input tag is edited set a listener in JS, and then invoke a lisp:// when the listener gets invoked.

Ambrevar:
Thanks for the tip, I'll try it.

>> You mean something like Magit' transcient where there is no input line? We can make the input line a togglable option, that would fix it. It's that easy! :)
>
> That just doesn't feel right to me. It would feel like a kludge of our design. A work-around...

Indeed, it would be nicer to have a consistent interface, either all
GTK, or all HTML. 

- All GTK: Doable, the main drawback being that it's more demanding on
  the FFI interface, and as such the Qt renderer interface will have to
  be larger too.

- All HTML: I think the aforementioned HTML input tag approach could
  do.  I'll go this way for now.

That said, having just 1 exception to the consistency is not a
deal-breaker in my opinion.  This inconsistency can be contained in just
one or two spots (construction and toggler).  All current features of
Nyxt would still be supported.

Ambrevar:
Renaming to do:

- [ ] Prompt to

Ambrevar:
See commit 83fd8d9d5d79f4d1519b68a1d11b143a84cde5bb: I've added an HTML input
tag and I'm logging it's content to the REPL.  This confirms that we can input
arbitrary text in the minibuffer input and act upon it.

I didn't bind Javascript to the input tag: opening a lisp:// URL while the
minibuffer is running maybe be brittle, not sure.

Anyways, instead I'm reading the input tag content on key release, which seems
to do the job perfectly (and it's simpler).

@jmercouris: Can you confirm you like this strategy?

jmercouris:
I confirm insofar as I know the strategy is sound. Some potential caveats:

1. processing action on keystroke up introduces latency for the user (a lot)
2. what happens when you set the visibility of the input to hidden? are you still able to type?

Ambrevar:
Thanks!
I'll start integrating it then!

> 1. processing action on keystroke up introduces latency for the user (a lot)

While we never do it, the "keystroke up" event is fire up instantly, I
can confirm that when I type "control-shift u 20ac" and release, the €
symbol appears instantly as well as the log in the REPL.

> 2. what happens when you set the visibility of the input to hidden? are you still able to type?

My guess is that we can use the regular HTML password input.

jmercouris:
I'm not talking about using a input type of password, I'm talking about setting it's visibility to hidden; is it still interactable?

Ambrevar:
Sorry, I don't understand.

Are you talking about the `invisible-input-p' minibuffer slot?

If so, we can implement it by using an HTML input tag with the password
option instead of a regular one.

jmercouris:
I'm talking about setting a style on the input with the following parameters:

<input style="visibility: hidden">

Ambrevar:
Yes, this is what we can do!  I think we are talking about the same thing!

jmercouris:
I think you may be thinking of `<input type="password">`

which is slightly different

Ambrevar:
Forgive my ignorance :)
Can you detail the difference and how that would
matter for the minibuffer?

jmercouris:
In one case the text appears as a password field, obfuscated, in the other case, the view doesn't appear at all, it is invisible.

jmercouris:
Why it matters? May wish to provide entirely custom view of the input area (to support block cursors for example)

Ambrevar:
Many updates on this branch, it's starting to be quite usable!
Feedback welcome!

You can try out 2 commands that use the new _prompt buffer_:

- `set-url2`
- `switch-buffer2`

Highlights:

- Paging is supported (can be improved in the future).

  - `M-<` and `M->` go to first, resp. last suggestion.
  - `C-v` and `M-v` go one page up, resp. down.

- Multiple column support.  It's fully customizable via the `object-properties`
  method and the `suggestion-property-function` slot.

- Multiple source support.  See `set-url2`.

- Multiple action support.  See `set-url`, press `M-return` and stand in awe :)

- Persistent action support.  See `switch-buffer2` and navigate the suggestions:
  the current buffer is automatically switched to the suggestion.  On `escape`,
  the original buffer is restored.  `C-c C-f` toggles "follow mode".
  
The `prompter-library` is almost finished.

The `prompt-buffer.lisp` file is the main spot that needs refactoring.  Will do
when I'm done with the rest.

What remains to be done:

- [ ] Have `prompt-buffer` inherit from `prompter` instead of having a
  `prompter` slot.  This would make the Nyxt side of things way less verbose, it
  will also be easier for the end user to invoke the prompt buffer.
- [ ] Rename the `prompter:prompter-source` class to `prompter:source`.
- [ ] Contextual help display.
- [ ] Resume previous prompt-buffers.
- [ ] Highlight portion of matching text.
- [ ] VI bindings
- [ ] Implement `search-buffer2` command using `prompt-buffer`.
- [ ] Restore multi-buffer search leveraging the new multiple source feature.
- [ ] Make meta command which allows for selecting sources, then drop into new
  prompt buffer with given sources.

The good news is that most of the above tasks are really easy to implement.  The
only one I'm unsure about is the `search-buffer` command.  Will see.

Then after a final review, I'll replace all the minibuffer instances with
prompt-buffers.

There are also some cosmetic tweaks I would need help with (pinging @jmercouris :p):

- [ ] Auto-focus on HTML input.  Maybe focusing on the GTK widget would fix it.
- [ ] Display HTML input and prompt on the same line.
- [ ] Fit columns to width, and keep their width constant when navigating the
      suggestions.
- [ ] While paging works, going "up" (e.g. with `C-p`) does not recenter
  suggestions that are "under" the input line and the prompt.
  
Stay tuned!

Ambrevar:
To do:

- [ ] The prompt buffer key bindings should be discoverable (e.g. with `C-h b`).

Ambrevar:
Closing in favour of #1157.

PR #1055, opened 2020-11-27
Asd refactor

Ambrevar:
This does mostly 2 things:

- Stick to ASDF conventions (mostly), use consistent style.
- Move most Makefile operations to `nyxt.asd`.

I've kept most Makefile rules for backward compatibility.  They just call to their .asd counterpart.

Result: The Makefile lost about 100 lines of difficult syntax, thus shrinking to about half its size (barely 90 lines).  We can now control every build rule in Lisp!

After all, why would we use `make` while we have ASDF? :)

Tip: Look at the individual commits, it will make this review easier.

### DIFF
--- a/.github/workflows/package-ubuntu.yml
+++ b/.github/workflows/package-ubuntu.yml
@@ -32,10 +32,9 @@ jobs:
         sudo apt-get install -y ruby ruby-dev rubygems dpkg-dev sbcl curl git-core zlib1g-dev
         sudo apt-get install -y libfixposix0 libfixposix-dev libwebkit2gtk-4.0-dev glib-networking gsettings-desktop-schemas xclip notify-osd enchant
 
-    - name: Fetch Common Lisp third-party dependencies
+    - name: Register Common Lisp third-party dependencies location
       shell: bash
       run: |
-        make quicklisp-extra-libs
         mkdir -p ~/.config/common-lisp/source-registry.conf.d/
         echo "(:tree \"$PWD/quicklisp-libraries\")" >> ~/.config/common-lisp/source-registry.conf.d/asdf.conf
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,10 +56,10 @@ jobs:
       shell: bash
       run: ros -e "(ql:update-all-dists :prompt nil)"
 
-    - name: Fetch Common Lisp third-party dependencies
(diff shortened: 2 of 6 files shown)


Ambrevar:
Question: Should I reverse the logic of the `+nyxt-internal-quicklisp` feature, and instead enable the internal Quicklisp by default, only to disable it if `nyxt-external-quicklisp` is present in the features.

Ambrevar:
Don't merge, this breaks the CI, it seems that it does not fetch the Cluffer dependency there.  I'll investigate.

Ambrevar:
Another benefit of this is that it turns `make` into an optional dependency.
Now we can build a binary from scratch with just one ASDF invokation:

``` shell
$ sbcl --eval '(asdf:make :nyxt/gtk-application)'
```

which much more Lispy, nah? :)

jmercouris:
that is indeed so much better! I have strongly preferred just running `(asdf:make :nyxt/gtk-application)`. When will we be able to merge this in?

jmercouris:
As per changing the internal quicklisp logic, that is up to you, either approach seems OK to me.

Ambrevar:
We need two things:

- Fix the CI: this is just about registering the modules properly.

- Make sure we don't rebuild dependencies when loading a system.  In
  particular, the above ASDF invokation should return very promptly when
  no source file was edited.  This was supported by the Makefile, so we
  want to make sure this time-saver is still there.

Ambrevar:
I decided against it, since it would not help with backward compatibility.

Ambrevar:
I've pushed most of this changes to master, until f5c2729fc2741a86e8ea3667f61559ad49512bb1.

What's missing:
- https://github.com/atlas-engineer/nyxt/pull/1055/commits/3c06ed2f7ff330457bfc78a0d7532ebb9be86dbf
- https://github.com/atlas-engineer/nyxt/pull/1055/commits/81c4c8130260b7921dce9da0c6711c601c34f561
- https://github.com/atlas-engineer/nyxt/pull/1055/commits/62fb5cfaeb6ce3cc4f7898da008382f930003158
- https://github.com/atlas-engineer/nyxt/pull/1055/commits/d714e29943d7b9937572c34dbe30779726f23ee9
- https://github.com/atlas-engineer/nyxt/pull/1055/commits/0f9682db088855434baa5a7b53b5dde082656d86

Ambrevar:
Continued in #1275.

PR #1058, opened 2020-12-01
Dbscan

jmercouris:
The results are promising!
Screenshot_20201201_160337

The feature extraction needs to be greatly optimized. This is orthogonal to the DBScan implementation, which is functional.


jmercouris:
Feature extraction work will come down the line, it will likely be an ongoing project... ready to merge when you've had a looksy.

jmercouris:
I would really like to add tests as well to validate every single component. That will also come later.

jmercouris:
I think I will use a different feature extraction instead of tf-idf vectorization, word counts may suffice.

libraries/analysis/README.org
```diff
@@ -0,0 +1,25 @@
...
+Analysis is a library that provides facilities to help analyze and
+understand data. Listed below are the classes:
+
+** Document
+The document class represents a document. After creating a document,
+you can perform several operations on it, some examples:
+
++ term count: how many times does a term appear in a document?
++ term frequency: how many times does a term appear / the total number
+  of words in the document?
```
Ambrevar, line 11: Is `/` supposed to mean "over"? If so, I suggest to write it in plain letters because `/` also means "or" and it can be ambiguous.

libraries/analysis/README.org
```diff
@@ -0,0 +1,25 @@
...
+
++ dictionary: which words appear in the document collection?
++ keywords: what are the important keywords in this document
+  collection?
+
+** Document Vertex
+The document vertex class represents a document that is part of a
+graph. The edges slot of the document vertex class is used to store
+edges of that particular vertex. The keys in the edges slot hash table
+are the actual vertexes, and the values are the edge weights.
```
Ambrevar, line 25: What about moving the class description to the `:documentation` class option of the respective classes?  This way the doc would be accessible from Nyxt!

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+               if label(Q) ≠ undefined then continue           /* Previously processed (e.g., border point) */
+               label(Q) := C                                   /* Label neighbor */
+               Neighbors N := RangeQuery(DB, distFunc, Q, eps) /* Find neighbors */
+               if |N| ≥ minPts then {                          /* Density check (if Q is a core point) */
+                   S := S ∪ N                                  /* Add new neighbors to seed set */
+               }
+           }
+       }
+   }
+"
```
Ambrevar, line 66: Maybe replace this with a permanent link to the corresponding Wikipedia article?

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+(defmethod generate-document-distance-vectors ((collection document-collection))
+  "Set the edge weights for all document neighbors (graph is fully connected)."
+  (with-accessors ((documents documents)) collection
+    (loop for document-a in documents
+          do (loop for document-b in documents
+                   do (setf (gethash document-b (edges document-a))
+                            (distance document-a document-b))))))
+
+(defmethod dbscan ((collection document-collection) &key (minimum-points 3)
+                                                         (epsilon 0.5))
```
Ambrevar, line 42: While I know that `epsilon` is a customary name, I wonder if `threshold` or `error-margin` wouldn't be more explicit.

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+   Please note: this function is not responsible for computing the
+   clusters, only for returning the list of pre-tagged documents in
+   cluster lists."
+  (let ((result (make-hash-table)))
+    (loop for document in (documents collection)
+          do (push document (gethash (cluster document) result (list))))
+    result))
+
+(defmethod distance ((vector-1 t) (vector-2 t))
+  "Calculate the euclidean distance between two vectors."
```
Ambrevar, line 25: "Euclidean" (capitalize)

Ambrevar: Maybe more conventionally "Return the Euclidian distance between VECTOR-1 and VECTOR-2."

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+(defclass document-cluster (document-vertex)
+  ((cluster :accessor cluster)
+   (neighbors :accessor neighbors)))
+
+(defmethod clusters ((collection document-collection))
+  "Return a list of clusters.
+
+   Please note: this function is not responsible for computing the
+   clusters, only for returning the list of pre-tagged documents in
+   cluster lists."
```
Ambrevar, line 18: So how are the clusters computed?  I suggest we add a reference to the functions that does this job.

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+
+;;; dbscan.lisp -- implementation of Density-based spatial clustering
+;;; of applications with noise (DBSCAN) algorithm
+
+(defclass document-cluster (document-vertex)
+  ((cluster :accessor cluster)
+   (neighbors :accessor neighbors)))
+
+(defmethod clusters ((collection document-collection))
+  "Return a list of clusters.
```
Ambrevar, line 14: It actually returns a hash-table.  Mistake?

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+               }
+           }
+       }
+   }
+"
+  (labels ((range-query (document)
+             "Return all points that have a distance less than epsilon."
+             (loop for vertex being the hash-keys of (edges document)
+                   when (and (<= (gethash vertex (edges document)) epsilon)
+                             (not (eq vertex document)))
```
Ambrevar, line 71: I don't understand this `eq` check, can you explain?

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+                   collect vertex))
+           (noisep (document)
```
Ambrevar, line 73: Took me some seconds to understand the name of this function, first I read "no-is-ep" then "no-i-sep" :p I think `noise-p` would be better here (although I'm still preferring `noise?`... :p )

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+             "A document must have a minimum number of neighbors to
+              not qualify as noise."
+             (setf (neighbors document) (range-query document))
```
Ambrevar, line 76: Should this be wrpped in `(unless (neighbors document) ...)` to avoid computing it if not needed?

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+             (if (< (+ 1 (length (neighbors document))) minimum-points)
+                 (progn (setf (cluster document) :noise) t)
```
Ambrevar, line 78: Why return `T` here?  `:noise` is already "true", so you could remove the `progn`.

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+             "Return all points that have a distance less than epsilon."
+             (loop for vertex being the hash-keys of (edges document)
+                   when (and (<= (gethash vertex (edges document)) epsilon)
+                             (not (eq vertex document)))
+                   collect vertex))
+           (noisep (document)
+             "A document must have a minimum number of neighbors to
+              not qualify as noise."
+             (setf (neighbors document) (range-query document))
+             (if (< (+ 1 (length (neighbors document))) minimum-points)
```
Ambrevar, line 77: Maybe replace `+ 1` by `1+`?
Also you can replace the if by an `and` and remove the else clause.

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+             (setf (cluster document) cluster)
+             (setf stack (neighbors document))
+             (loop while stack for item = (pop stack)
+                   when (and (slot-boundp item 'cluster)
+                             (eq (cluster item) :noise))
+                   do (setf (cluster item) cluster)
+                   unless (slot-boundp item 'cluster)
+                   do (setf (cluster item) cluster)
+                      (unless (noisep item)
+                        (append (neighbors item) stack))))))
```
Ambrevar, line 95: Looks like this `append` result ignored.  I think `stack` should be setf to the result.

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+                   collect vertex))
+           (noisep (document)
+             "A document must have a minimum number of neighbors to
+              not qualify as noise."
+             (setf (neighbors document) (range-query document))
+             (if (< (+ 1 (length (neighbors document))) minimum-points)
+                 (progn (setf (cluster document) :noise) t)
+                 nil)))
+    (loop for document in (documents collection)
+          with cluster = 0
```
Ambrevar, line 81: Previously the term `cluster` was used to refer to a cluster object, but here it's an index, which can make the following code a bit confusing.  What do you think about renaming `cluster` to `cluster-index` or similar?

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+          with stack = (list)
+          unless (or (slot-boundp document 'cluster)
+                     (noisep document))
+          do (incf cluster)
+             (setf (cluster document) cluster)
+             (setf stack (neighbors document))
+             (loop while stack for item = (pop stack)
```
Ambrevar, line 88: Maybe a newline between the `while` and the `for`?

libraries/analysis/dbscan.lisp
```diff
@@ -0,0 +1,95 @@
...
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :analysis)
+
+;;; dbscan.lisp -- implementation of Density-based spatial clustering
+;;; of applications with noise (DBSCAN) algorithm
+
+(defclass document-cluster (document-vertex)
+  ((cluster :accessor cluster)
+   (neighbors :accessor neighbors)))
```
Ambrevar, line 11: Can you explain what cluster and neighbours are?  Type and documentation?

libraries/analysis/dbscan.lisp
Ambrevar: So an item is a neighbour, right?  Shouldn't we rename it to `neighbour` then?

libraries/analysis/document-vector.lisp
```diff
@@ -0,0 +1,52 @@
...
+(defmethod word-count-vectorize ((document document) dictionary)
+  "Transform a document into a vector using word counts."
+  (let ((vector-form (make-array (length dictionary) :initial-element 0)))
+    (loop for word in dictionary
+          for index from 0 below (length vector-form)
+          do (setf (aref vector-form index) (term-count document word)))
+    (setf (vector-form document) vector-form)))
+
+(defmethod tf-idf-vectorize ((document document) (collection document-collection) dictionary)
+  "Transform a document into a vector using tf-idf."
```
Ambrevar, line 17: I'm not a big fan of "tf-idf", I find that it makes for cryptic names :)  
Can we come up with a more explicit name?
Should we stick to `tf-idf`, I suggest we define it in the docstrings everywhere it's used.

libraries/analysis/document-vector.lisp
```diff
@@ -0,0 +1,52 @@
...
+  (let ((dictionary (dictionary document-collection)))
+    (loop for document in (documents document-collection)
+          do (tf-idf-vectorize document document-collection dictionary))))
+
+(defmethod tf-vectorize-documents ((document-collection document-collection))
+  "Set the vector-form for a collection of documents."
+  (let ((dictionary (dictionary document-collection)))
+    (loop for document in (documents document-collection)
+          do (tf-vectorize document dictionary))))
+
```
Ambrevar, line 51: I think you can remove the empty lines here :)

libraries/analysis/document-vector.lisp
```diff
@@ -0,0 +1,52 @@
...
+  "Transform a document into a vector using tf."
+  (let ((vector-form (make-array (length dictionary) :initial-element 0)))
+    (loop for word in dictionary
+          for index from 0 below (length vector-form)
+          do (setf (aref vector-form index)
+                   (term-frequency document word)))
+    (setf (vector-form document) vector-form)))
+
+(defmethod word-count-vectorize-documents ((document-collection document-collection))
+  "Set the vector-form for a collection of documents."
```
Ambrevar, line 35: What is the `vector-form`?

libraries/analysis/document-vector.lisp
```diff
@@ -0,0 +1,52 @@
...
+  (let ((dictionary (dictionary document-collection)))
+    (loop for document in (documents document-collection)
+          do (word-count-vectorize document dictionary))))
+
+(defmethod tf-idf-vectorize-documents ((document-collection document-collection))
+  "Set the vector-form for a collection of documents."
```
Ambrevar, line 41: This docstring looks copy-pasted from above.  Mistake?

libraries/analysis/document-vector.lisp
```diff
@@ -0,0 +1,52 @@
...
+  (let ((dictionary (dictionary document-collection)))
+    (loop for document in (documents document-collection)
+          do (tf-idf-vectorize document document-collection dictionary))))
+
+(defmethod tf-vectorize-documents ((document-collection document-collection))
+  "Set the vector-form for a collection of documents."
```
Ambrevar, line 47: Same.

source/base-mode.lisp
```diff
@@ -150,20 +150,55 @@
...
+                     do (with-current-buffer buffer
+                          (analysis::add-document 
+                           collection
+                           (make-instance 'analysis::document-cluster
+                                          :source buffer
+                                          :string-contents (document-get-paragraph-contents)))))
+               (analysis::tf-idf-vectorize-documents collection)
+               (analysis::generate-document-distance-vectors collection)
+               (analysis::dbscan collection :minimum-points 3 :epsilon 0.1)
+               (analysis::clusters collection)))
```
Ambrevar, line 169: There are many unexported symbols here.  What about exporting them?

Ambrevar: What about having a "master" function in `analysis` that takes a list of document-clusters, runs tf-idf-vectorize-documents,  generate-document-distance-vectors, dbscan, then returns the clusters.  This way the library users would have only 2 entry points: the creation of the document-cluster objects and the master processing function.

libraries/analysis/README.org
jmercouris: it means 'divided by'

libraries/analysis/README.org
jmercouris: OK! will do!

libraries/analysis/dbscan.lisp
jmercouris: link may disappear/change with time, I will put a reference to the wiki article too though

libraries/analysis/dbscan.lisp
jmercouris: epsilon is the name used in the literature, and so I believe it is best this way. I can put a documentation about what it means though, would that suffice?

libraries/analysis/document-vector.lisp
jmercouris: it depends, just a vector that represents a document.

libraries/analysis/document-vector.lisp
jmercouris: Not a mistake, all the functions do the same thing, and could probably be a macro or function that accepts a function.

source/base-mode.lisp
jmercouris: I'm not comfortable with exporting symbols yet, as I have not decided upon an API. For example in Text rank/summarization I have an API decided, hence why I export the symbol.

jmercouris: In other words, there are lots of things I would like to yet change :-D

libraries/analysis/README.org
Ambrevar: OK, so I suggest we write it in full letters to avoid any ambiguity.

libraries/analysis/dbscan.lisp
Ambrevar: That's what I meant by "permanent link": use the history of the Wikipedia article to link to a content that won't change over time.

libraries/analysis/dbscan.lisp
Ambrevar: Expliciting what it is in the docstring would suffice I think.

libraries/analysis/document-vector.lisp
Ambrevar: `verctor-form` (with a dash) sounds like a well-defined term.  What about "vectorized document" instead?

source/base-mode.lisp
Ambrevar: Fair enough! :)  My second comment would help with this: if there is only one entry point, that's just one function for which to decide (now or later) whether to export it or not.

libraries/analysis/document-vector.lisp
jmercouris: `(vector-form document)` reads better than `(vectorized-document document)` I think

libraries/analysis/document-vector.lisp
Ambrevar: I see.  Then I suggest just extract  the 3 lines to a helper function which takes the function as argument.  Simple, no macros :)

libraries/analysis/dbscan.lisp
jmercouris: Who says wikipedia will always exist?

libraries/analysis/document-vector.lisp
Ambrevar: I find `vector-form` ambiguous because form is used both in mathematics and in programming, with a diffferent meanint.  It's not obvious here, I think, that "form" is the representation of the document.

libraries/analysis/dbscan.lisp
Ambrevar: It's backed up on the Internet archive, and if the latter ceases to exist, then working on a web browser will probably make little sense at this point :D 

That said, including the snippet in a separate document, or in the readme, would make sense to me.

libraries/analysis/document-vector.lisp
jmercouris: I would have just used the word vector, but that would cause problems with CL, that is why I settled on vector-form, what about vector-data?

Ambrevar: I like vector-data better.  It's not very specific but at least it's less ambiguous :)

libraries/analysis/document-vector.lisp
```diff
@@ -31,20 +31,22 @@
...
+                                :dictionary dictionary))))
 
-(defmethod tf-idf-vectorize-documents ((document-collection document-collection))
-  "Set the vector-form for a collection of documents."
-  (let ((dictionary (dictionary document-collection)))
-    (loop for document in (documents document-collection)
-          do (tf-idf-vectorize document document-collection dictionary))))
+(defmethod word-count-vectorize-documents ((document-collection document-collection))
+  (vectorize-documents document-collection (lambda (&key document dictionary)
+                                             (word-count-vectorize document dictionary))))
```
Ambrevar, line 42: You are double-wrapping the functions it seems :)

If you simply call `(vectorize-documents document-collection #'word-count-size)` and edit `vectorize-documents to remove the keyword args, it should work!

(You can also add the keyword args to word-count-vectorize instead.)

Same with the 2 functions below.

jmercouris: No it won't work, the functions accept a different amount of args, otherwise I would have done that :-D

jmercouris: I see now what you meant, I had the keyword args because initially I was going to change the source functions to accept keywords args, but then I changed my mind halfway through and forgot to remove them... :-P

Ambrevar: So what I proposed would work, right?

libraries/analysis/dbscan.lisp
Ambrevar: @jmercouris Wait!  What about this one?  Maybe you've merged a bit early :p

Ambrevar:
@jmercouris Looks like most review points about dbscan.lisp haven't been addressed.  Can we still review them?

jmercouris:
What do you mean? I fixed all of the things that you commented on

jmercouris:
Oh damn, I just realized they are only visible when I click on files-changed and then go to dbscan.lisp, I will take a look at them now, sorry about that!

libraries/analysis/dbscan.lisp
jmercouris: because I think it is better to be explicit and return a boolean value when possible

libraries/analysis/dbscan.lisp
jmercouris: We don't want to add a document to itself as a list of its neighbors, so we check that they are distinct entities

libraries/analysis/dbscan.lisp
jmercouris: Not a mistake, the hash table is effectively a list of clusters

libraries/analysis/dbscan.lisp
jmercouris: There are many ways for computing clusters, dbscan is just one of the possibilities.

libraries/analysis/dbscan.lisp
jmercouris: I will add a dash. Then let's decide that we'll eschew the convention and just always use a dash even for single words?

libraries/analysis/dbscan.lisp
jmercouris: Hm, yes, that should have been appendf!

libraries/analysis/document-vector.lisp
jmercouris: tf-idf is the term used in the literature, I am not inventing something new here :-D

libraries/analysis/dbscan.lisp
jmercouris: I do not believe it is possible to be computed accidentally

libraries/analysis/dbscan.lisp
jmercouris: It is a neighbor indeed, but to WHOM? Sometimes they are neighbors of the core cluster, sometimes they are neighbors of the core clusters neighbors. It is better left as item for this purpose

libraries/analysis/dbscan.lisp
jmercouris: Documentation updated with clarification.

libraries/analysis/dbscan.lisp
jmercouris: Done!

libraries/analysis/dbscan.lisp
jmercouris: I disagree with that change, I want to be very clear here that noise-p is a boolean return

libraries/analysis/dbscan.lisp
jmercouris: that is a good idea!

jmercouris: Done!

libraries/analysis/dbscan.lisp
jmercouris: it is just a processing queue, hence the term item

libraries/analysis/dbscan.lisp
Ambrevar: Why is `document` in its own list of `edges` then?  Couldn't it be removed?

Note: You can bind the values with `using hash-values VALUES` instead of calling `(gethash vertex (edges document))`, plus it's faster :)

libraries/analysis/dbscan.lisp
Ambrevar: Actually I was using this convention all along, thinking we had agreed on that ;)  So yes, let's keep using a dash!

libraries/analysis/dbscan.lisp
Ambrevar: `and` returns a boolean, so it's clear.  It's idiomatic Lisp to use `and` or `or` for this ;)

libraries/analysis/dbscan.lisp
Ambrevar: OK.

jmercouris:
Sorry about that, I don't know how I didn't see those! I think I have addressed everything now!

libraries/analysis/document-vector.lisp
Ambrevar: Sure, but it's specialized literature, not something that the average reader is expected to know, so it would be good to define (or url-link to definitions) all the terms we are using in the code.  (Preferably in docstrings.)

When I read the code I had to look up the terms on Wikipedia which is a bit frustrating since the expanded acronym is actually much more self-explanatory.

libraries/analysis/dbscan.lisp
Ambrevar: OK, but hen `neighbors` should only have a reader and no accessor.

jmercouris: I added a guard in place anyways (unless (slot-boundp document 'neighbors) ...)

libraries/analysis/document-vector.lisp
jmercouris: How about including the terms in the readme?

libraries/analysis/dbscan.lisp
jmercouris: Well, a document is in fact a neighbor to itself, that is how it is typically done

jmercouris: In this case, we are only interested in ones we need to calculate and process though, which is why we remove the document itself from its list of neighbors

libraries/analysis/document-vector.lisp
jmercouris: Well, I added them into the docstrings anyways :-D

jmercouris:
Now I believe I have addressed everything!

Ambrevar:
> Now I believe I have addressed everything!

Thanks!

PR #1061, opened 2020-12-03
Ospama multiversion

Ambrevar:
This fixes an important issue with Ospama's previous design: it adds multi-version packages support.

Bonuses:

- It fixes a bug in which it would fail when listing an installed package with a version that was not found in the current Guix package database.

- It's now possible to install any variation of a package `(version output)` pair.

### DIFF
--- a/libraries/ospama/ospama-guix.lisp
+++ b/libraries/ospama/ospama-guix.lisp
@@ -317,10 +317,17 @@ for non-standard profiles."
   (declare (ignore manager))
   (setf *guix-database* nil))
 
-(defmethod install-command ((manager guix-manager) profile)
-  (append (list (path manager) "install")
-          (when profile
-            (list (str:concat "--profile=" (namestring profile))))))
+(defmethod manager-install ((manager guix-manager) output-list &optional profile)
+  (run (append (list (path manager) "install")
+               (mapcar (lambda (output)
+                         (let ((package (parent-package output)))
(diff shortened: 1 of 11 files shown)


libraries/ospama/ospama-guix.lisp
```diff
@@ -156,6 +159,9 @@ value.
   (:export-accessor-names-p t)
   (:accessor-name-transformer #'class*:name-identity))
 
+(defun guix-package-p (object)
+  (mopu:subclassp (class-of object) 'guix-package))
```
jmercouris, line 163: while this does work, you could also implement a method that specializes on guix-package as a predicate

Ambrevar:
I'm actually going to do better: Add a class option to class-star to
automatically generate the predicate, just like defstruct does :)

jmercouris:
Sounds great!

Ambrevar:
Done.

What do you think of this predicate auto-definition?

jmercouris:
Very cool! Is there a reason for the usage of a defun rather than the method? Just preference?

Ambrevar:
2 lines less :p

A possible benefit of having a defun is that it catches accidental
overrides, like it did here.

Had we defined a defmethod, another defmethod in the file would be
redundant and uncaught.

We actually had this issue before the last commit on this branch :)

jmercouris:
Well then, why not :-D merge when ready!

PR #1071, opened 2020-12-12
Visual mode

### LINKED ISSUE #653: A "visual mode" for copying and pasting?
Hi there! I am not sure if I can open an issue with this, so if I am wrongly doing let me know to close or delete this.

Question: Is there a command so I can copy from the browser without having to use the mouse? I am thinking something like a visual mode of sort in vim or keyboard-select in urxvt.
If there is not a command, is it possible to have one?

Thanks!

kssytsrk:
Closes #653

### DIFF
--- a/source/visual-mode.lisp
+++ b/source/visual-mode.lisp
@@ -11,6 +11,13 @@
   "Visual mode."
   ((keymap-scheme
     (define-scheme "visual"
+      scheme:cua
+      (list
+       "up" 'backward-line
+       "down" 'forward-line
+       "left" 'forward-char
+       "right" 'forward-char
+       "escape" 'visual-mode)
       scheme:emacs
       (list
        "C-h" 'select-paragraph
@@ -26,7 +33,7 @@
        "C-e" 'end-line
        "M-a" 'backward-sentence
        "M-e" 'forward-sentence)
-      ;; vi keybindings only enable use of plain "visual" mode for now
+      ;; vi keybindings only enable use of vim's plain "visual" mode for now
       scheme:vi-normal
       (list
        "h" 'backward-char
@@ -35,19 +42,18 @@
        "l" 'forward-char
        "w" 'forward-word
        "$" 'end-line
-       "<end>" 'end-line
-       "<space>" 'forward-char
-       "<backspace>" 'backward-char
+       "keypadend" 'end-line
+       "space" 'forward-char
+       "backspace" 'backward-char
        "0" 'beginning-line
-       "<home>" 'beginning-line
+       "keypadhome" 'beginning-line
        "v" 'visual-mode
-       "<escape>" 'visual-mode
        "C-c" 'visual-mode)))
    (destructor
     (lambda (mode)
       (make-page-uneditable)
       (unlock-page-keypresses)
-      (setf *mark-set* nil)))
+      (defparameter *mark-set* nil)))
    (constructor
     (lambda (mode)
       (make-page-editable)
@@ -55,7 +61,7 @@
       (select-paragraph)
       ;; imitating visual mode in vim
       (if (equal (keymap-scheme-name (buffer mode)) scheme:vi-normal)
-        (setf *mark-set* t))))))
+        (defvar *mark-set* nil))))))
 
 (define-parenscript %add-paragraph-hints (&key annotate-visible-only-p)
   (defun qs (context selector)
@@ -269,9 +275,10 @@ identifier for every hinted element."
   (setf (ps:@ document body content-editable) "false"))
 
 (define-command select-paragraph ()
+  "Add hints to text elements on the page and open a minibuffer for selecting them."
   (query-paragraph-hints "Set caret on element" #'%follow-hint :annotate-visible-only-p t))
 
-(defparameter *mark-set* nil)
+(defvar *mark-set* nil)
 
 (define-parenscript is-collapsed ()
   ;; returns "true" if mark's start and end are the same value
--- a/source/visual-mode.lisp
+++ b/source/visual-mode.lisp
@@ -2,7 +2,7 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (uiop:define-package :nyxt/visual-mode
-    (:use :common-lisp :nyxt)
+  (:use :common-lisp :nyxt)
   (:import-from #:keymap #:define-key #:define-scheme)
   (:documentation "Visual mode."))
 (in-package :nyxt/visual-mode)
@@ -132,7 +132,7 @@ identifier for every hinted element."
     (if (or (ps:chain element offset-width)
             (ps:chain element offset-height)
             (ps:chain element (get-client-rects) length))
-      t nil))
+        t nil))
 
   (defun element-in-view-port-p (element)
     (ps:let* ((rect (ps:chain element (get-bounding-client-rect))))
@@ -140,7 +140,7 @@ identifier for every hinted element."
                (>= (ps:chain rect left) 0)
                (<= (ps:chain rect right) (ps:chain window inner-width))
                (<= (ps:chain rect bottom) (ps:chain window inner-height)))
-        t nil)))
+          t nil)))
 
   (defun object-create (element hint)
     (ps:create "type" "p" "hint" hint "identifier" hint "body" (ps:@ element |innerHTML|)))
@@ -178,9 +178,9 @@ identifier for every hinted element."
   (defun string-generate (n)
     "Generates a string from a number"
     (if (>= n 0)
-      (+ (string-generate (floor (- (/ n 26) 1)))
-         (code-char (+ 65
-                       (rem n 26)))) ""))
+        (+ (string-generate (floor (- (/ n 26) 1)))
+           (code-char (+ 65
+                         (rem n 26)))) ""))
 
   (add-stylesheet)
   (hints-add (qsa-text-nodes)))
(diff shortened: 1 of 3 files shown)


jmercouris:
Thank you very much! I will take a detailed look on Monday!

source/visual-mode.lisp
```diff
@@ -0,0 +1,391 @@
...
+       "$" 'end-line
+       "keypadend" 'end-line
+       "space" 'forward-char
+       "backspace" 'backward-char
+       "0" 'beginning-line
+       "keypadhome" 'beginning-line
+       "v" 'visual-mode
+       "C-c" 'visual-mode)))
+   (destructor
+    (lambda (mode)
```
jmercouris, line 53: `(declare (ignore (mode))` can be used to suppress a warning about an unused symbol

source/visual-mode.lisp
```diff
@@ -0,0 +1,391 @@
...
+      (make-page-uneditable)
+      (unlock-page-keypresses)
+      (defparameter *mark-set* nil)))
```
jmercouris, line 56: I believe that `*mark-set*` should be a slot in the `visual-mode` class. That way there would be no conflict if there were multiple visual modes active on different buffers at the same time :-)

source/visual-mode.lisp
```diff
@@ -0,0 +1,391 @@
...
+    (ps:chain sel (remove-all-ranges))
+    (ps:chain sel (add-range range))))
+
+(defmethod %follow-hint ((paragraph-hint paragraph-hint))
+  (set-caret-on-start :nyxt-identifier (nyxt/web-mode::identifier paragraph-hint)))
+
+(defun paragraph-elements-from-json (elements-json)
+  (loop for element in (cl-json:decode-json-from-string elements-json)
+        collect (let ((object-type (cdr (assoc :type element))))
+                  (log:info element)
```
jmercouris, line 218: this log slows down Nyxt quite a bit by outputting information that isn't relevant to the end user :-O

source/visual-mode.lisp
```diff
@@ -0,0 +1,391 @@
...
+      (let ((result (prompt-minibuffer
+                     :input-prompt prompt
+                     :default-modes '(element-hint-mode minibuffer-mode)
+                     :history nil
+                     :multi-selection-p multi-selection-p
+                     :suggestion-function
+                     (nyxt/web-mode::hint-suggestion-filter
+                      (paragraph-elements-from-json elements-json))
+                     :changed-callback
+                     (let ((subsequent-call nil))
```
jmercouris, line 240: We don't have to worry about highlighting in this context. This does bring up a good point, this could is a good candidate for an abstract selection function!

jmercouris:
Overall fantastic! I like it, no more mouse needed for selection! A few minor points/suggestions:

+ the code copied from the element hints part of the codebase could be abstracted and shared between the two
+ avoid the usage of globals as it causes problems when two or more visual hint modes exist at the same time

What do you think of a follow-up PR where you add the ability for the user to add/edit the text?

Would you like to make the changes I've suggested, or shall I handle them? Either way is OK with me!

thanks,

-John

jmercouris:
In case I wasn't clear enough above, thanks so much for this PR, it was very kind of you :-)

source/visual-mode.lisp
kssytsrk: I copied most of the code of this function from element-hint.lisp, it had the log:info call, should I still remove it?

Ambrevar:
Just tried it, it's awesome! :D
Thank you so much, this is an impressive contribution!

Some feedback:

- The first 2 times I called visual-mode and selected a paragraph, no
  caret appeared afterwards.  I can't reproduce right now.

- vi-normal and CUA do not have a "toggle-mark" binding, which makes
  them way less useful :p  Let's add one!  `v` seems to be a good binding for
  vi-normal.
  Traditionally in CUA the user would hold shift and move the caret around to
  select stuff.

- More generally, CUA keys are a bit lacking.  In particular keypadhome,
  pageend, home, etc.  Easy fix! :)

- Shouldn't we add a command to copy the text?  I suppose that's the
  main thing we want to do in visual-mode, right?

- Indentation is a bit off, e.g.

```lisp
(if *mark-set*
  "extend"
  "move")
```

should be


```lisp
(if *mark-set*
    "extend"
    "move")
```

Which editor are you using?

source/visual-mode.lisp
jmercouris: Yes, please remove it. It probably should not be in element-hint.lisp either...

kssytsrk:
>the code copied from the element hints part of the codebase could be abstracted and shared between the two

Nice idea, I think I might do that! Maybe the parts of element-hint.lisp specifically for follow-hint command should be moved to a separate file? 

>What do you think of a follow-up PR where you add the ability for the user to add/edit the text?

Oh, that could be done by just removing the 'make-page-uneditable function or calling 'make-page-editable. Maybe I should bind it to some key (although then what should be done with vi-like keybinds? they wouldn't work... adding some kind of insert-mode which you can quit with <esc>?).

>In case I wasn't clear enough above, thanks so much for this PR, it was very kind of you :-)

>Just tried it, it's awesome! :D
Thank you so much, this is an impressive contribution!

Thank you both a lot for the kind words :) 

>The first 2 times I called visual-mode and selected a paragraph, no
caret appeared afterwards. I can't reproduce right now.

I understand what you are talking about, that happens to me too sometimes but I can't quite figure out why - and for me, it occurs with inputs sometimes, too. Only when I click somewhere on the window with my mouse it goes away and doesn't return until I change the focus to another window and then return to Nyxt.

>vi-normal and CUA do not have a "toggle-mark" binding, which makes
them way less useful :p Let's add one! v seems to be a good binding for
vi-normal.

"v" quits visual mode as of now, because it seemed to me that this is like vim users actually use that key. I might be wrong though, I only read like, two tutorials on visual mode to learn about the keybindings, and they also never mentioned toggling the mark so I didn't add it.

>Traditionally in CUA the user would hold shift and move the caret around to
select stuff.

I'm not sure how this could be implemented, are there some kind of hooks/whatever other structures in Nyxt that would allow to run something on press and then on release of a key?

>Shouldn't we add a command to copy the text? I suppose that's the
main thing we want to do in visual-mode, right?

Don't we have M-w already? I thought that was enough, but if not, I can add another one.

>Which editor are you using?

Well, I use Emacs, and my slime common lisp indentation style is "classic"
```
(setq lisp-indent-function 'common-lisp-indent-function)
(setq common-lisp-style-default "classic")
```

It's set up this way because otherwise the indentation of 'loop is kind of messed up. I'm not sure what would be the way to set it up to not mess up the 'if.

source/visual-mode.lisp
kssytsrk: Oh wait, it wasn't in element-hint.lisp, I think I added it for my own debugging purposes. My bad. I'll remove it

Ambrevar:
> "v" quits visual mode as of now, because it seemed to me that this is like vim
> users actually use that key. I might be wrong though, I only read like, two
> tutorials on visual mode to learn about the keybindings, and they also never
> mentioned toggling the mark so I didn't add it.

I think `escape` should also quit visual mode.

> I'm not sure how this could be implemented, are there some kind of
> hooks/whatever other structures in Nyxt that would allow to run something on
> press and then on release of a key?

Why not just bind `shift-left`, `shift-up`, etc.?
If you need state (e.g. `text-marked-p`) you can add a slot to the mode.
Thus the command bound to shift+arrows would check the slot value:
- if it's nil, enable the mark.
- if it's t, leave the mark on.

Then you could bind other movement commands to disable the mark.

> Don't we have M-w already? I thought that was enough, but if not, I can add
> another one.

Right, that should do.

> Well, I use Emacs, and my slime common lisp indentation style is "classic"
> ```
> (setq lisp-indent-function 'common-lisp-indent-function)
> (setq common-lisp-style-default "classic")
> ```

For Nyxt you should leave these variables to their default values.
For `loop`, if the indentation is messed up it's probably because you are not
starting the line with the right keyword.  Example:

``` common-lisp
;; Good:
(loop for x in '(a b c)
      collect x)
	  
;; Not as good:
(loop for x in
      '(a b c)
	  collect
	    x)
```

If you prefer, you can also use syntactical keywords:


``` common-lisp
;; Good:
(loop :for x :in '(a b c)
      :collect x)
```

kssytsrk:
Hm, this is very weird, but I can't invoke anything with C-space anymore, 'execute-command always gets called instead, even if I try the version I've pushed to github (and which you both tested, I suppose?) without init and autoconfig.

Ambrevar:
I don't understand, are you saying that C-space on master does not mark the minibuffer suggestions?
Or is `execute-command` broken?

Can you provide a precise recipe exhibiting your problem?

kssytsrk:
I'm saying that in visual mode (from my fork) with emacs-mode turned on, when I press C-space I expect 'toggle-mark to be executed, but 'execute-command gets executed. If I change 'toggle-mark to anything else in the source of visual-mode.lisp, 'execute-command still gets executed instead of the other function bound to C-space. But I'm pretty sure it worked before, and I haven't yet made changes to the code (actually, I did, but I git stashed them and switched to the version of the branch I pushed to my fork)

Recipe: M-x visual-mode RET C-space
here i expect 'toggle-mark, but 'execute-command gets executed instead. It might be something wrong with my system (although I'm not sure how could only one keybinding be affected, and the others perfectly fine)

Ambrevar:
I believe you are mistaken: C-space is bound in the `buffer`'s
`override-map`, which is why it takes precedence over any mode.

In other words, you cannot bind anything to C-space unless you modify
the override-map (which only the user should do).

Hope that helps!

kssytsrk:
Oh. I remember we talked about this in #653 and everything was fine then:

> The C-space binding is free in Emacs mode since M-x is bound to
execute-command, just like in Emacs!

But alright. Then I'm accepting suggestions about the new keybinding for 'toggle-mark in emacs-mode, s-space is free and so is m-space...

jmercouris:
I like s-space, I think that is good, assuming you mean (shift).

Ambrevar:
kssytsrk <notifications@github.com> writes:

> Oh. I remember we talked about this in #653 and everything was fine then:
>
>> The C-space binding is free in Emacs mode since M-x is bound to
> execute-command, just like in Emacs!
>
> But alright. Then I'm accepting suggestions about the new keybinding for 'toggle-mark in emacs-mode, s-space is free and so is m-space...

Sorry, that was wrong, C-space is never touched, and M-x is not added to
the override map.

We have to go with some other binding by default, but I suggest we
document the example of how to install the C-space binding for Emacs users.

kssytsrk:
I think I fixed all things that were wrong in the previous version. 

@jmercouris I remember the part about abstracting shared code between element-hint.lisp and visual-mode.lisp but sadly I don't have the time to do it right now. Do you want to wait for me to get around to it before you merge this or could it be a followup PR later? :)

kssytsrk:
Oh wait, didn't add the documentation. Where should it go, by the way, just into the mode documentation string? Maybe I should put some notice about visual mode into manual? It seems to me that a lot of people would like to use that feature, although I may be biased because I wrote it, haha.

jmercouris:
You can put it in the manual if you like :-) It should definitely be at least in the mode documentation string.

With regards to abstracting the code, don't worry about it, either I will get to it later, or in another PR.

Thank you so much, I will have an in-depth look later!

kssytsrk:
Okay. One more question - I'm now writing the visual mode's documentation in the manual, and it looks like this:


Is (UNBOUND) instead of the commands' keybindings (as the visual-mode is not activated on the help buffer, the commands are really unbound at that point) really fine? Or should I add the keybindings by hand somehow?

jmercouris:
What you can do is pass a list of modes to `binding-keys` as a key argument, that way it will show the bindings in that mode.

jmercouris:
`(binding-keys 'my-function '(my-mode))`

jmercouris:
Actually, you should use `command-markup` if possible

`(command-markup 'my-function :modes '(my-mode))`

kssytsrk:
I see, thank you, that does make it better. But now I just get the CUA-like keybindigs, even though I have emacs-mode turned on in the help buffer. Calling something like this

```common-lisp
(binding-keys 'nyxt/visual-mode:forward-sentence 
              :modes (list (make-instance 'nyxt/visual-mode:visual-mode) 
                           (make-instance 'nyxt/emacs-mode:emacs-mode)))
```

doesn't help, either, even though the 'forward-sentence command is definitely bound in emacs-mode. 'binding-keys produces similar output.

jmercouris:
That is a bug in our implementation, I can handle that. If everything looks good otherwise, please push and I'll  be glad to merge :-)

kssytsrk:
> We have to go with some other binding by default, but I suggest we
document the example of how to install the C-space binding for Emacs users.

I'm not sure how to do this, by suggesting to add something into visual-mode's enable-hook? I tried something like this

```common-lisp
(in-package :nyxt-user)

(define-configuration nyxt/visual-mode:visual-mode
    ((enable-hook (hooks:add-hook
                   %slot-default
                   (nyxt::make-handler-mode
                    (lambda (mode)
                      (define-configuration buffer
                          ((override-map (let ((map (make-keymap "override-map")))
                                           (define-key map
                                             "C-space" 'nyxt/visual-mode::toggle-mark)
                                           map)))))
                    :name 'visual-mode-c-space-override)))))
```

and it didn't quite work. (obviously, there would have to be something similar that removes the keybinding/restores it to default in `disable-hook`)

jmercouris:
I think there is a slight misunderstanding here. It would be very hacky for a mode to try to override the override map, and it is not intended anyways (you can, but should not do this). The whole point of an override map should be that it should not be changed by any mode.

What would happen is the user would just change their init file and change the override map. So, for example in my config (as an emacs user, I have the following):

jmercouris:
```
(define-configuration web-buffer
    ((default-modes (append '(auto-mode emacs-mode) %slot-default))
     (override-map (let ((map (make-keymap "my-override-map")))
                     (define-key map
                       "C-q" 'quit
                       "C-1" 'make-window
                       "C-2" 'delete-current-window)
                     map))))
```

jmercouris:
Therefore, you would just put a line instructing emacs users to set their own override-map such that C-space is not bound, and therefore it could be used by visual mode :-)

kssytsrk:
Thank you. Now everything is done, I think.

jmercouris:
Thanks! I will try to take a final look later today, it will be a nice respite from my current work :-D

jmercouris:
Fantastic, work, thank you so much! We'll hopefully abstract the commonalities at some point with element-hint (if it makes sense to do so), but other than that, wonderful!

Issue #808, opened 2020-06-25
Random hangs when opening URL externally


Ambrevar:
Quite often when I open a URL externally while Nyxt is already running, it hangs the process.

After force-quitting it, restarting it results in a Nyxt window that shows up for a fraction of a second before crashing again.

The second restart finally works.

I suspect that the first restart crashes because of either a left-over process or a left-over socket.

As for why the external URL sometimes hangs Nyxt, I have no idea.  We'd first need a reliable recipe.

Ambrevar:
Tip to investigate this issue: Start Swank / Slynk in the init file to make it possible to connect to a hung Nyxt.

Ambrevar:
Tip: See if the issue does not come from the request-resource hook when current buffer has no URL.

Ambrevar:
@aartaka @jmercouris Have you been able to reproduce this recently?

jmercouris:
I don't ever open URLs externally, sorry :-\. I can test this if you like. Perhaps via a fuzzer of sorts?

Ambrevar:
What's a "fuzzer"?  A random URL opener?

aartaka:
I'm unable to reproduce -- it opens 10+ simulatenous links without problem. I've ran `open-external-urls` fro ~20 times with 0-13 urls, but seen to sign of hanging.

I'll try again tomorrow, but no luck in reproducing so far.

jmercouris:
When I say fuzzing in that case, yes, that is what I mean, randomizing some variable. We could randomize the URL and the timings to see if we can make a pattern emerge for crashing Nyxt.

Ambrevar:
@jmercouris: Out of curiosity, how to do open URLs from external
programs, e.g. from your email client?

jmercouris:
I open them in Safari, I use Nyxt for work

Ambrevar:
I just hit this error again today :(

jmercouris:
Since it is random, I have a strong feeling that it has to do with some thread state.

Ambrevar:
It is not necessarily random.  It may be because of some command that
was run for instance.  It's just that so far I've failed to see a pattern.

Ambrevar:
Actually, since it seems to happen way less than before, my guess is
that we've fixed one bug that would trigger it most often.  Now maybe
there is some other command that I only seldom run that triggers it as well.

jmercouris:
Well, the title is "random", just saying :-D

[...]

PR #1072, opened 2020-12-15
Fix startup race condition when invoking the minibuffer.

Ambrevar:
Major change: `finalize' is now in charge for spawning the first window.
This should make `startup-function's much easier to write on the user end.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -101,8 +101,8 @@ count since deleting windows may result in duplicate identifiers.")
                      :type (or function null)
                      :documentation "The function run on startup.  It takes a
 list of URLs (strings) as optional argument (the command line positional
-arguments).  It is run after the renderer has been initialized and after the
-`*after-init-hook*' has run.")
+arguments).  It is run after the renderer has been initialized, after the
+`*after-init-hook*' has run and after a first window has been spawned.")
    (startup-error-reporter-function nil
                                     :type (or function null)
                                     :export nil
@@ -212,10 +212,24 @@ editor executable."))
   ;; `messages-appender' requires `*browser*' to be initialized.
   (log4cl:add-appender log4cl:*root-logger* (make-instance 'messages-appender))
   (handler-case
-      (hooks:run-hook *after-init-hook*)
+      (hooks:run-hook *after-init-hook*) ; TODO: Run outside the main loop?
     (error (c)
       (log:error "In *after-init-hook*: ~a" c)))
-  (funcall-safely (startup-function browser) urls)
+  ;; Most of the code assumes the existence of a window.  Let's create it here
+  ;; so that the user does not have to bother with it in the `startup-function'.
+  (let ((window (window-make browser)))
+    (window-set-active-buffer window (help)))
+  ;; The `startup-function' must be run _after_ this function returns;
+  ;; `ffi-within-renderer-thread' runs its body on the renderer thread when it's
+  ;; idle, so it should do the job.  It's not enough since the
+  ;; `startup-function' may invoke the minibuffer, which cannot be invoked from
+  ;; the renderer thread: this is why we run the startup-function in a new
+  ;; thread from there.
+  (ffi-within-renderer-thread
+   browser
+   (lambda ()
+     (pexec ()
+       (funcall-safely (startup-function browser) urls))))
   ;; Set 'init-time at the end of finalize to take the complete startup time
   ;; into account.
   (setf (slot-value *browser* 'init-time)
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -575,7 +575,7 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
     ;; Modes might require that buffer exists, so we need to initialize them
     ;; after the view has been created.
     (initialize-modes buffer)
-    (when dead-buffer
+    (when dead-buffer                   ; TODO: URL should be already set.  Useless?
       (setf (url buffer) (url dead-buffer)))
     (buffers-set (id buffer) buffer)
     ;; Run hooks before `initialize-modes' to allow for last-minute modification
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -320,6 +320,11 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
 
 (define-user-class internal-buffer)
 
+(defun make-dummy-buffer ()
+  ;; Internal buffers are lighter than full-blown buffers which can have a
+  ;; WebKit context, etc.
+  (make-instance 'user-internal-buffer))
+
 (define-class status-buffer (user-internal-buffer)
   ((height 20
            :type integer
@@ -642,7 +647,7 @@ proceeding."
   (let ((window-with-same-buffer (find buffer (delete window (window-list))
                                        :key #'active-buffer)))
     (if window-with-same-buffer ;; if visible on screen perform swap, otherwise just show
-        (let ((temp-buffer (make-instance 'user-buffer))
+        (let ((temp-buffer (make-dummy-buffer))
               (old-buffer (active-buffer window)))
           (log:debug "Swapping old buffer ~a with other window ~a to switch to ~a"
                      (object-string (url old-buffer))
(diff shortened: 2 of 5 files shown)


Ambrevar:
@aartaka Can you confirm that you don't get the startup race condition you used to have?
@jmercouris Can you test on macOS?

Ambrevar:
Great news don't come alone, and this fix might also fix the long (non-breaking) hang that we see on startup when many buffers get restored.

See #808 and #1012.

This hang seems to be due to us trying to restore too many buffers at the same time when initializing.  Now that we delay the restoration, it does not seem to hang the startup anymore.

jmercouris:
And what if the user does not want to spawn a window?

jmercouris:
Will they still be able to start Nyxt without spawning a window right away?

jmercouris:
Actually you know what. I think you are right with your previous comment. The startup-function should be more limited in what it can do. If someone wants to dramatically change the startup of Nyxt it will take just a slight amount of effort more, but it will be a lot less effort for 99% of users who just want to change how their first buffer is formed.

source/browser.lisp
```diff
@@ -212,10 +212,24 @@ editor executable."))
...
   (log4cl:add-appender log4cl:*root-logger* (make-instance 'messages-appender))
   (handler-case
-      (hooks:run-hook *after-init-hook*)
+      (hooks:run-hook *after-init-hook*) ; TODO: Run outside the main loop?
     (error (c)
       (log:error "In *after-init-hook*: ~a" c)))
-  (funcall-safely (startup-function browser) urls)
+  ;; Most of the code assumes the existence of a window.  Let's create it here
+  ;; so that the user does not have to bother with it in the `startup-function'.
+  (let ((window (window-make browser)))
```
jmercouris, line 220: I disagree, I think the user should have to create the window in the startup-function.

jmercouris:
Other than my comment about window creation, it looks OK to me! Thank you Pierre!

Ambrevar:
I don't know if it makes sense to start Nyxt without starting a window right away.
If we want to support this use case nonetheless, then we need to redesign quite a few things in our code because too much of it assumes that a window exists.
So exposing an entry point to the user where a window does not exist is a recipe for disaster :p 

Anyways, in the context of this fix spawning the window beforehand matter: without it, there may be a race condition and the minibuffer would fail to spawn.

Ambrevar:
Do not merge, I want to fix something else with the socket logic.

Ambrevar:
I've just pushed a second commit which should help with the slow start when opening a URL externally.

The drawback is that `*socket-path*` can no longer be set from the init file.
It can be set from command line options, but obviously if the default browser is called by just `nyxt` it won't work.  So I guess we must introduce an environment variable here.  Thoughts?

jmercouris:
What about checking for an environment variable, otherwise load the init file and try to contact Nyxt that way? That way if the environment variable is set, it will be faster, otherwise it will be the same speed as before

jmercouris:
Scratch my above comment, we never did load the init file anyways, I e looked at your current implementation and it looks OK to me.

Ambrevar:
Indeed, this fix does not change the behaviour, we've never allowed to set the
socket-path from the init file.

Even if we did, I think the strategy of loading the init file if the env
variable is not set would not solve the problem: if the init file takes,
say, 2s to load, then each URL opened externally will also take 2s to load.

jmercouris:
Feel free to merge whenever :-)

Ambrevar:
To be sure, have you tested on macOS?

jmercouris:
I've not, I can do so

jmercouris:
I can confirm that it works fine on macOS

Ambrevar:
Thanks!  Will merge soon.

aartaka:
> @aartaka Can you confirm that you don't get the startup race condition you used to have?

I don't get the race condition. Works as a charm!

Ambrevar:
Thanks for reporting!


There are news however: I'm trying to fix the libsoup collision bug once and for
all.
The easiest way to fix it is to avoid creating web-buffer web view at the same
time.

9094ff2f489ef514e7d587f1abcde4309f4aa5b5 fixes the startup collisions by not
making contexts for internal buffers.  This is a great improvement already!

(Feedback welcome!)

Now a little bit harder is to fix the collisions that happen when restoring a
session.

Commit 168e5bd587ce434214b623a28d7c1e70f59f9537 adds a 0.05s sleep between each
buffer, which is enough on my machine but might not work on slower systems.
This is obviously very brittle.

A cleaner fix would be to separate buffer create from web view creation, and
only create the latter on demand, e.g. when we call window-set-current-buffer.

This has some significant implications though.  For instance `buffer-make`
would not initialize the modes anymore, the web-view creation function should do
it instead.

Before going down this rabbit hole, I'd like to confirm with you that this is a
sound decision.

Thoughts? @jmercouris @aartaka?

jmercouris:
It is most probably not a sound decision :-D. I would suggest to remove this fix from this PR so that we can merge it and master works again.

I would hold off on trying to patch the behavior of GTK bugs in Lisp. I suggest that we implement gobject backend before spending any time trying to fix this bug. If we are to fix this bug, we should make it transparent to our processes, the API implementation (aka renderer-gtk.lisp) should handle all details. There should be no special cases, if there are we need to improve the API or the API implementation.

Ambrevar:

> I would suggest to remove this fix from this PR so that we can merge it and master works again.

Why remove the fix?  At worst, the fix does not change anything on
slower hardware.  At best, it makes for a smoother experience on fast hardware.

> I would hold off on trying to patch the behavior of GTK bugs in
> Lisp. I suggest that we implement gobject backend before spending any
> time trying to fix this bug.

This bug is in Libsoup, see

https://gitlab.gnome.org/GNOME/libsoup/-/issues/192

It'll be there with the gobject backend as well.

> If we are to fix this bug, we should make it transparent to our processes, the API implementation (aka renderer-gtk.lisp) should handle all details. There should be no special cases, if there are we need to improve the API or the API implementation.

This is a fair point.

My suggestion of having just-in-time webview instantiation could be
restricted to renderer-gtk I believe.  I'll work out the details later.

Ambrevar:
I've tried 2 different approaches:

- Instantiate the buffer gtk-object in the accessor.
  This alone does not work change anything because when we initialize a buffer,
  we initialize the modes, which in turn may need access to the gtk-object
  (proxy-mode does).

  The way forward would be to delay mode initialization to when gtk-object is
  initialized.

  This does not seem well abstracted because now we have to deal with
  half-initialized buffers and it's unclear to me whether all the code base can
  properly deal with that.  We lack a single entry point for buffer initialization.

- Add the history entries directly to the buffer list.
  Then in `window-set-active-buffer` check if we are given a buffer or a
  history.  If a history, instantiate the corresponding buffer, when set this
  buffer.

  The problem is that now `buffer-suggestion-filter` needs to deal with both
  buffers and histories.  In general, any piece of code that relies on
  `buffer-list` would need to be updated.
  
Another approach which I haven't tried, but probably cleaner than the above.

- Define an unrestored-gtk-buffer class between `buffer` and `gtk-buffer`.
  This class would not have an `initialize-instance` after method.

  Then the session `restore` method would fill the buffer list with this
  unrestored-gtk-buffers which can be instantiated to full-blown `gtk-buffer`s
  just in time.

  We would still need a entry point to buffer initialization and the above still
  has the problem that our code base now would need to differentiate between
  gtk-buffers and unrestored-gtk-buffers.

- Better: instead of adding the unrestored buffers to `buffers`, add the session
  histories to a new slot, "unrestored buffers".  Then with the new minibuffer
  multi-source support, we can list these entries alongside the other buffers.
  This would give us clean data separation and not break any of the existing code.

This raises a question about the similarity between dead buffers (killed, but
can brought back to life with `reopen-buffer` and unrestored buffers.  Can both
be conflated?  I'm not sure because unrestored buffer only come with a history
and that's it, while dead buffers still hold the mode instances in memory.

Ambrevar:
Except for the sleep in the last commit, I've merged this as 4a1beacbe37014b542f5783811d9e681e36d9443.

Ambrevar:
We can keep discussing lazy buffer restoration in #1076.

PR #1073, opened 2020-12-16
waleee-cl on #nyxt at freenode pointed out there was a typo

fade:
... in the makefile. change is self explanatory.


Ambrevar:
Good catch, thanks!

PR #1075, opened 2020-12-17
Fix buffer search highlighting

### LINKED ISSUE #666: Search sometimes highlights the wrong text
Some highlighting is at the wrong place.

Recipe: on the help start page, search "the", press RETURN and see that
"y o" is highlighted in the following line:

> Customization is possible through the creation of a ~/.config/next/init.lisp file. From here you can override and redefine any of they of they of the functions by defining your init file
> as part of the :next package. For more information please see: customizing Next.

acwilton:
Closes #666 

The bug stems from the function get-substring-indices inside of the query-buffer parenscript found in search-buffer.lisp.

Given a string and a query, the function is supposed to calculate the collection of indices (if any) of the the beginning of each match it finds. In doing this, each time it finds a match, it creates a substring consisting of the text just past the last found match so that it can run a regex to find the next match. The indices collected by this function are those calculated by the consecutive calls to the regex, however these indices are relative to the substring the regex is working on when they should be relative to the original string passed to the function.

This explains why the bug seemed to only appear sometimes. It only manifested on a match if it was at least the second match found within a text node of the buffer, because then its starting index was wrong which messed everything up.

Also, not a part of the bug fix, but I did condense a let statement found lower in the file.

Let me know what you think and if there are any other changes that you would like for me to make to get this pull request accepted! This is my first real experience with common lisp and the web browser in general so I apologize for any newb mistakes. Programming in lisp is really fun and I really like this project!


jmercouris:
Hi Alex,

thanks so much for looking into this. This bug has been plaguing us for a long time, and I could not possibly figure out why it was not working! Search is an incredibly important feature, and it is nice to see another bug gone!

I'm glad you like Lisp and Nyxt. If you would be interested in participating more, I can definitely make suggestions on more stuff we could work on together.

Anyways, a review follows shortly, just wanted to acknowledge receipt of this PR.

source/search-buffer.lisp
```diff
@@ -49,13 +49,15 @@
...
-              collect index)))
+    (let ((rgx (ps:new (|RegExp| query (if (ps:lisp case-sensitive-p) "" "i"))))
+	  (index (- (length query))))
+      (loop with subindex = 0
+            until (= subindex -1)
+            do (setf subindex (ps:chain string (search rgx)))
+               (setf string (ps:chain string (substring (+ subindex (length query)))))
+	       (setf index (+ index subindex (length query)))
+            when (not (= subindex -1))
+              Collect index)))
```
jmercouris, line 60: capitalization/indentation fix :-)

jmercouris:
I've cherry picked and merged your commits, thanks once again :-)

Ambrevar:
Hi Alex,

Can't stress it enough, thank you so much for this fix!
It's been annoying me for a while as well and I never found the time to
get down to it.  (Our "to do" list is rather long... :p)

As feedback, I have only one nit: prefer spaces to tabs when indenting
Lisp code ;)

Cheers!

acwilton:
Thanks for the feedback. My mistake on using tabs! I certainly agree with using spaces (even outside of lisp), I must have just forgotten to change that setting in emacs. I'll also try to be a bit more careful about indentation and capitalization.

Definitely leave some recommendations for issues that could be worked on next and Ill see if I can find time to work on them. I can't promise anything as I am a busy grad student. In fact this contribution was actually for a software engineering class (had to pick an open source project, contribute to it, and write about it). But if I find some free time over winter break or after I'd love to spend it honing my lisp skills and progressing Nyxt!

Ambrevar:
Thanks for the offer! :)

The "good-first-issue" tag should give you an good list of issues you
can get started with:

https://github.com/atlas-engineer/nyxt/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-issue

Issue #930, opened 2020-09-05
Feature Request: Invidious Public Instances Redirect From YouTube


ccao001:
How could I add a function that with a single keyboard shortcut opens the modeline and redirects me from `youtube.com` to choose amongst one of several [online invidious instances](https://instances.invidio.us/)?

I would like the modeline to sort the instances by current health of the site. [See here](https://invidio.us/) or the `dailyRatios` and `weeklyRatio` json fields below for an example of what I mean about site health.

Any help or suggestions are greatly appreciated! A version of this program for [online peertube instances](https://instances.joinpeertube.org/instances) would be cool too.

# JSON code showing Invidious Instances as of Today:

```json
[
  [
    "invidious.snopyta.org",
    {
      "flag": "🇫🇮",
      "region": "FI",
      "stats": {
        "version": "2.0",
        "software": {
          "name": "invidious",
          "version": "0.20.1-13f58d6",
          "branch": "master"
        },
        "openRegistrations": true,
        "usage": {
          "users": {
            "total": 3056,
            "activeHalfyear": 2546,
            "activeMonth": 1729
          }
        },
        "metadata": {
          "updatedAt": 1599289689,
          "lastChannelRefreshedAt": 1599289686
        }
      },
      "type": "https",
      "uri": "https://invidious.snopyta.org/",
      "monitor": {
        "monitorId": 782758961,
        "statusClass": "success",
        "weeklyRatio": {
          "ratio": "99.367",
          "label": "success"
        },
        "name": "invidious.snopyta.org",
        "type": "http(s)",
        "dailyRatios": [
          {
            "ratio": "100.00",
            "label": "success"
          },
          {
            "ratio": "100.00",
            "label": "success"
          },
          {
            "ratio": "100.00",
            "label": "success"
          },
          {
            "ratio": "100.00",
            "label": "success"
          },
          {
            "ratio": "100.00",
            "label": "success"
          },
          {
            "ratio": "95.569",
            "label": "warning"
          },
          {
            "ratio": "100.00",
            "label": "success"
          }
        ]
      }
    }
  ],
  [
    "yewtu.be",
    {
      "flag": "🇫🇷",
      "region": "FR",
      "stats": {
        "version": "2.0",
        "software": {
          "name": "invidious",
          "version": "0.20.1-4a6e920",
          "branch": "master"
        },
        "openRegistrations": true,
        "usage": {
          "users": {
            "total": 1548,
            "activeHalfyear": 1548,
[...]

PR #1077, opened 2020-12-19
Update information in the documents/EXTENSIONS.org file

kssytsrk:
As I have renamed and generalized (see #930) the invidious-handler extension, I think the information in documents/EXTENSIONS.org file should be updated too. So, this PR is just about that, hope it's ok :)


jmercouris:
Thank you!

PR #1079, opened 2020-12-22
minor documentation fix: only web-buffer class has the 'load-status slot

kssytsrk:
As the `load-status` slot is in `web-buffer` class since 2-pre-release-2 nyxt version, the example of a `status-formatter` function in the documentation of the window class (in `source/window.lisp`) is currently buggy and will error out every time a non-web buffer is loaded (it will say that there is no slot like that in the `buffer` class). I think that the example should be corrected, adding a check if the buffer is a web buffer (it works perfectly fine with this check), and that's what this pull request is for.


Ambrevar:
Good catch, thank you very much!

PR #1093, opened 2021-01-06
WIP: Nosave (i.e. private) buffers

### LINKED ISSUE #247: incognito mode

aartaka:
This adds `nosave` buffers that other browsers call "private". Our interpretation is that `nosave` buffers should actually save _nothing_ to user machine. At all. Even if this breaks most websites.

 Most of the work was done in  #1007, https://github.com/joachifm/cl-webkit/pull/43, https://github.com/joachifm/cl-webkit/pull/40, https://github.com/joachifm/cl-webkit/pull/39, and, surprisingly, 88d0c5e3757f25566d0a2fe52db006402ae33138, so this is quite a lightweight patchset.

# What's new:

- `private-data-profile` renamed to `nosave-data-profile`.
- `nosave-buffer` class that has `nosave-data-profile` set by default but behaves like `web-buffer` otherwise.
- `set-url-nosave-buffer`, `follow-hint-nosave-buffer`, `follow-hint-nosave-buffer-focus`.
-  Intelligent Tracking Prevention is enabled by default, because it make one safer overall and doesn't cost anything. [Safari enabled it some year ago](https://webkit.org/blog/8613/intelligent-tracking-prevention-2-1/), so why shouldn't we so the same :)

# Caveats

- This needs to be merged strictly after GHT (#1007).
- `on-signal-notify-uri` behaves quirky with `nosave-buffers`, so I'll need to polish it.
- There were several unpleasant changes to `set-url` that made it somewhat incomprehensible. It needs some more thinking. @Ambrevar, @jmercouris, if you come up with a better design that what's in the patch -- don't hesitate to suggest it!
- Given that `cookie-manager` does not exist, no cookies are accepted, even temporary ones. This makes it impossible to use any cookie-relying website. Is that what we want? It's honest about not storing anything, though :)

# How Has This Been Tested?

- I ran Nyxt with this patchset and tried to log in to some websites and log out of them. Most of the time I was rejected because they needed cookie storage permissions :) In other words, it's quite untested. The fact of login rejection tells something about what is not stored, though!

Closes #247


source/renderer-gtk.lisp
```diff
@@ -495,10 +495,11 @@ Warning: This behaviour may change in the future."
...
 
 (defun make-data-manager (buffer)
-  (let ((path (expand-path (data-manager-path buffer))))
-    (apply #'make-instance `(webkit:webkit-website-data-manager
-                             ,@(when path `(:base-data-directory ,path))
-                             :is-ephemeral ,(not path)))))
+  (let* ((path (expand-path (data-manager-path buffer)))
+         (manager (apply #'make-instance `(webkit:webkit-website-data-manager
+                                           ,@(when path `(:base-data-directory ,path))
+                                           :is-ephemeral ,(not path)))))
```
Ambrevar, line 501: We should probably prefix this with a feature macro `#+WEBKIT2-TRACKING` since it may not be available for people with an old webkit.

source/buffer.lisp
```diff
@@ -277,6 +277,17 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
   (when (expand-path (cookies-path buffer))
     (ensure-parent-exists (expand-path (cookies-path buffer)))))
 
+(define-class nosave-buffer (user-web-buffer)
+  ((data-profile (make-instance 'nosave-data-profile))
+   (default-cookie-policy :never))
+  ;; TODO: Add reduce-tracking-mode etc. to default modes?
```
Ambrevar, line 283: I'd say yes :)  It'd be a sane default, which users can always tweak to their liking.

source/buffer.lisp
```diff
@@ -538,7 +538,8 @@ LOAD-URL-P controls whether to load URL right at buffer creation."
     (unless (url-empty-p url)
       (if load-url-p
           (buffer-load url :buffer buffer)
-          (setf (url buffer) (quri:uri url))))))
+          (setf (url buffer) (quri:uri url))))
+    buffer))
```
Ambrevar, line 542: Squash this commit in the one where you introduce the command?

Ambrevar:
Ha, this is great!  I didn't know we were that close to having nosave (also
known as "private" or "incognito") buffers!  Fantastic, thanks a lot!

> - There were several unpleasant changes to `set-url` that made it somewhat
>   incomprehensible. It needs some more thinking. @Ambrevar, @jmercouris, if you
>   come up with a better design that what's in the patch -- don't hesitate to
>   suggest it!

Can you detail what you don't like about `set-url`?

Personally I see that we have pass buffer creation around in too many places.
Why not make the options a (pseudo-)nomad then?
For instance, in a high-level buffer-making command, we would wrap the code with

``` common-lisp
(with-buffer-making-option (nosave ...)
  ; some intermediate callers)
```

and finally in `buffer-make`

``` common-lisp
(defun buffer-make (browser &key data-profile title default-modes dead-buffer)
  (multiple-value-bind (internal-buffer-p ...)
      (buffer-making-options)
    (apply #'make-instance (if internal-buffer-p
                               'user-internal-buffer
                               'user-web-buffer)
           :id (get-unique-buffer-identifier *browser*)
           (append (when title `(:title ,title))
                   (when default-modes `(:default-modes ,default-modes))
                   (when data-profile `(:data-profile ,data-profile)))))
  ...)
```

Thoughts?

source/buffer.lisp
aartaka: Now that I think about it, maybe we shouldn't set it that rigidly...

Given that ephemeral `WebKitWebsiteDataManager` manages cookies, it's shouldn't be necessary and will probably break less websites.

aartaka:
> Personally I see that we have pass buffer creation around in too many places.

Handling of that is exactly what I meant by "unpleasant changes", because adding `nosave` buffers into equation made `set-url` somewhat leaky and not obvious. E.g. What happens if we pass both `new-buffer-p` and `nosave-buffer-p`? What if we pass only `nosave-buffer-p`? Signature doesn't help in decyphering that.

> Why not make the options a (pseudo-)nomad then?
> For instance, in a high-level buffer-making command, we would wrap the code with
> 
> ```lisp
> (with-buffer-making-option (nosave ...)
>   ; some intermediate callers)
> ```
> 
> and finally in `buffer-make`
> 
> ```lisp
> (defun buffer-make (browser &key data-profile title default-modes dead-buffer)
>   (multiple-value-bind (internal-buffer-p ...)
>       (buffer-making-options)
>     (apply #'make-instance (if internal-buffer-p
>                                'user-internal-buffer
>                                'user-web-buffer)
>            :id (get-unique-buffer-identifier *browser*)
>            (append (when title `(:title ,title))
>                    (when default-modes `(:default-modes ,default-modes))
>                    (when data-profile `(:data-profile ,data-profile)))))
>   ...)
> ```


Yes, that sounds sane, especially given that we already use this kind of operations (e.g. `with-current-buffer`) :)

Ambrevar:
Now that #1110 is in, this draft can be resumed :)

aartaka:
Sure. Two things:
- Should the base branch be `master` now?
- Does `with-buffer-options` make sense as part of this patchset?

Ambrevar:
> - Should the base branch be `master` now?

Sure.  Did you have another branch in mind?

> - Does `with-buffer-options` make sense as part of this patchset?

Not sure.  If you think this patchset works without it, let's keep
with-buffer-options for later.

aartaka:
> Sure. Did you have another branch in mind?

No -- I'm just not sure what will happen to all the `global-history-tree` commits if we change the base branch they are already merged into -- will they display as the ones to merge, or not?

> - Does `with-buffer-options` make sense as part of this patchset?
> Not sure. If you think this patchset works without it, let's keep with-buffer-options for later.

Patchset will work without `with-buffer-options`. Let's leave it for later then.

jmercouris:
If you are worried, make a new branch and cherry pick your commits, that is always a possibility!

aartaka:
Closing this in favor of #1153.

PR #1097, opened 2021-01-10
Make javascript URLs follow-hint-able.

aartaka:
This enables `javascript:`-schemed URL to run properly when `follow-hint`-ed. 

# Caveats
- This complicates a frequently used `buffer-load` with yet another conditional branch. While [Epiphany does it the same way](https://github.com/GNOME/epiphany/blob/36ade6b01cf2084a26b453a1dc3f8ce0a77daf56/embed/ephy-web-view.c#L2771), I'm still not sure whether we should do it this way. Maybe a `request-resource` clause?

# How Has This Been Tested

- Nyxt runs alright with this patchset, `javascript:` URLs load after being `follow-hint`-ed.

How does it look to you?

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -29,8 +29,8 @@ On errors, return URL."
   ;; List of URI schemes: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
   ;; Last updated 2020-08-26.
   (let ((valid-schemes
-          (cons
-           "lisp"                       ; For Nyxt's internal URLs.
+          (append
+           '("lisp" "javascript")              ; For Nyxt's internal URLs and not-standardised ones.
            '("aaa" "aaas" "about" "acap" "acct" "cap" "cid" "coap" "coap+tcp" "coap+ws"
              "coaps" "coaps+tcp" "coaps+ws" "crid" "data" "dav" "dict" "dns" "example" "file"
              "ftp" "geo" "go" "gopher" "h323" "http" "https" "iax" "icap" "im" "imap" "info"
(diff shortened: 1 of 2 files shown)


source/urls.lisp
```diff
@@ -34,7 +34,7 @@ On errors, return URL."
            '("aaa" "aaas" "about" "acap" "acct" "cap" "cid" "coap" "coap+tcp" "coap+ws"
              "coaps" "coaps+tcp" "coaps+ws" "crid" "data" "dav" "dict" "dns" "example" "file"
              "ftp" "geo" "go" "gopher" "h323" "http" "https" "iax" "icap" "im" "imap" "info"
-             "ipp" "ipps" "iris" "iris.beep" "iris.lwz" "iris.xpc" "iris.xpcs" "jabber"
+             "ipp" "ipps" "iris" "iris.beep" "iris.lwz" "iris.xpc" "iris.xpcs" "jabber" "javascript"
```
Ambrevar, line 37: Can you leave this out of the list and move it next to "lisp"?  (E.g. with `(append '("lisp" "javascript") ...)`.)
This way we can easily update the list when https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml gets updated.

Ambrevar:
Looks good to me.  The caveat is legit, I'm not sure what's best for us yet.
`request-resource` would probably an update at some point.
For now, I suggest you add a `TODO:` comment explaining the caveat and what the alternative is.

source/urls.lisp
aartaka: Done!

aartaka:
`TODO:` is there, non-standardized schemes are separated from standardized ones.

Ambrevar:
Thanks!

PR #1098, opened 2021-01-10
Add support for images in follow-hint.

aartaka:
This make images openable with `follow-hint`.

# What's New
- `image-hint` class that behaves mostly like `link-hint`.
- `elements-from-json` relies on `str:string-case`. It does look clearer, doesn't it? 
- `(cdr (assoc ...) ...)` in `elements-from-json` uses `alex:assoc-value` now.

# Caveats
- Naming may be inconsistent. Examples: `img` tag vs. `image-hint` class, `src` property vs. `url` slot. Is that okay?
- `image-hint` inherits from `link-hint`. This seems to be okay as long as `body` is not moved to `link-hint`. 
- By the way, why `body` slot is in `hint` not in `link-hint`?

# How Has This Been Tested
- Nyxt compiles without warnings.
- Images on protonmail.com become openable. Should work for all the other images on the Internet that are properly coded.

What do you think of it?

EDIT: `alex:assoc-value`.

### DIFF
--- a/source/element-hint.lisp
+++ b/source/element-hint.lisp
@@ -75,7 +75,9 @@ identifier for every hinted element."
           ((equal "INPUT" (ps:@ element tag-name))
            (ps:create "type" "input" "hint" hint "identifier" hint))
           ((equal "TEXTAREA" (ps:@ element tag-name))
-           (ps:create "type" "textarea" "hint" hint "identifier" hint))))
+           (ps:create "type" "textarea" "hint" hint "identifier" hint))
+          ((equal "IMG" (ps:@ element tag-name))
+           (ps:create "type" "img" "hint" hint "identifier" hint "src" (ps:@ element src) "alt" (ps:@ element alt)))))
 
   (defun hints-add (elements)
     "Adds hints on elements"
@@ -115,7 +117,7 @@ identifier for every hinted element."
                          (rem n 26)))) ""))
 
   (add-stylesheet)
-  (hints-add (qsa document (list "a" "button" "input" "textarea"))))
+  (hints-add (qsa document (list "a" "button" "input" "textarea" "img"))))
 
 (define-parenscript remove-element-hints ()
   (defun hints-remove-all ()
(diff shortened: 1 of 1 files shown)


source/element-hint.lisp
```diff
@@ -199,29 +201,37 @@ identifier for every hinted element."
...
-                                        :body (plump:text (plump:parse (cdr (assoc :body element))))))
-                        ((equal "input" object-type)
-                         (make-instance 'input-hint
-                                        :identifier (cdr (assoc :identifier element))
-                                        :hint (cdr (assoc :hint element))))
-                        ((equal "textarea" object-type)
-                         (make-instance 'textarea-hint
-                                        :identifier (cdr (assoc :identifier element))
-                                        :hint (cdr (assoc :hint element))))))))
+                  (str:string-case object-type
```
Ambrevar, line 204: Note that `match` would have worked here too ;)

source/element-hint.lisp
```diff
@@ -199,29 +201,37 @@ identifier for every hinted element."
...
+                    ("textarea"
+                     (make-instance 'textarea-hint
+                                    :identifier (cdr (assoc :identifier element))
+                                    :hint (cdr (assoc :hint element))))
+                    ("img"
+                     (make-instance 'image-hint
+                                    :identifier (cdr (assoc :identifier element))
+                                    :url (cdr (assoc :src element))
+                                    :alt (cdr (assoc :alt element))
+                                    :hint (cdr (assoc :hint element))))))))
```
Ambrevar, line 229: While you are at it, you can replace all the `(cdr (assoc ...))` with `(alex:assoc-value ...)` :)

Ambrevar:
Thank you so much for this, I've been wanting this for so long! :)

> - Naming may be inconsistent. Examples: `img` tag vs. `image-hint` class, `src` property vs. `url` slot. Is that okay?

I prefer to keep Common Lisp code free from the harder-to-read HTML
nomenclature.
Maybe `url` could be renamed to `source` and `image-hint` to `image`?

> - `image-hint` inherits from `link-hint`. This seems to be okay as long as `body` is not moved to `link-hint`. 
> - By the way, why `body` slot is in `hint` not in `link-hint`?

I don't know.  @jmercouris?

source/element-hint.lisp
aartaka: I'm not fan of `match` -- it doesn't looks enough CL-ish to me. You're right, though -- it's a worthy alternative!

In this particular case, `str:string-case` both conveys the type information and reminds of standard `case`, so it's more informative :)

aartaka:
OK, there're more applicatios for `body`, actually -- `button-hint` uses it too.

source/element-hint.lisp
Ambrevar: What not CL-ish enough?  It's all s-exps, isn't it?

Ambrevar:
So no more TODO: then?  If so, remove (or clarify) the TODO: comment you've added and we can merge :)

aartaka:
> Maybe `url` could be renamed to `source` 

This I can agree with, but then inheriting from `link-hint` won't be possible. `url` is informative enough, so let's keep it there.

> and `image-hint` to `image`?

This I can't agree with -- `image-hint` is consistent with other `*-hint` classes, and it should be -- after all, it's not actual image we're handling, but a hint pointing at it :upside_down_face:

aartaka:
> So no more TODO: then? If so, remove (or clarify) the TODO: comment you've added and we can merge :)

Done! :)

source/element-hint.lisp
aartaka: It indeed is. It's just that `match` have originated in Scheme and it looks foreign in CL to me. 

But these are just opinions -- John prefers the grotesque and mighty `loop` macro over any other iteration construct, for example :)

Ambrevar:
Thanks!

PR #1101, opened 2021-01-17
Add event-based hints

aartaka:
This adds hints for HTML elements based on most frequent JS events -- `onclick` and `onfocus`.

How does it look to you?

### DIFF
--- a/source/element-hint.lisp
+++ b/source/element-hint.lisp
@@ -79,7 +79,9 @@ identifier for every hinted element."
           ((equal "IMG" (ps:@ element tag-name))
            (ps:create "type" "img" "hint" hint "identifier" hint "src" (ps:@ element src) "alt" (ps:@ element alt)))
           ((ps:@ element onclick)
-           (ps:create "type" "clickable" "hint" hint "identifier" hint "body" (ps:@ element |innerHTML|)))))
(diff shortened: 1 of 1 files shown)


source/element-hint.lisp
```diff
@@ -239,20 +251,27 @@ identifier for every hinted element."
          :documentation "The body of the anchor tag."))
   (:accessor-name-transformer #'class*:name-identity))
 
-(defclass button-hint (hint) ())
+(define-class clickable-hint (hint) ())
```
jmercouris, line 254: Why use define-class?

jmercouris:
It looks great, thank you :-). Only my question about define-class vs defclass, otherwise OK to merge!

source/element-hint.lisp
aartaka: It's purely for consistency reasons, as I've said in commit message. Having `define-class` in all the hint definitions is more consistent and creates less cognitive load than having two different macros used for the same thing.

jmercouris: Typically people attempt to use the most specific construct for their needs. For example one could use `(if xyz xy)` for everything, but instead it is preferred to use `(when xyz xy)`. In this case, we might as well use our own macro I guess.

PR #1104, opened 2021-01-21
Gobject

jmercouris:
As said in my email, in the end it took only a few lines, but the commits should stay for posterity. We will reference them as we reintroduce GO constructs into our codebase (thereby truly differentiating the renderers and their dependencies).


nyxt.asd
```diff
@@ -159,6 +159,13 @@
   :pathname "source/"
   :components ((:file "renderer-gtk")))
 
+(asdf:defsystem :nyxt/gobject/gtk
```
Ambrevar, line 162: Which one do you prefer:
- gobject/gtk
- gi/gtk
- gi-gtk
?

We also need to add an application to build the final thing.  The name of the application would be one of the above, with "-application" appended.

Ambrevar:
Have you tried on Darwin?

I'll test just now on Guix.

Ambrevar:
About keeping the commits: no problem, but what about:

- Squashing the first 7-8 commits which are not very useful.
- Merging instead of rebasing?

jmercouris:
> Have you tried on Darwin?
> 
> I'll test just now on Guix.

GObject does not work on Darwin, I had to manually generate the bindings/fix them myself some time ago. Therefore, it is not reasonable that it would be runnable on Darwin. That said, it should work on Darwin (probably the threading may have to be tweaked slightly).

jmercouris:
> About keeping the commits: no problem, but what about:
> 
> * Squashing the first 7-8 commits which are not very useful.
> * Merging instead of rebasing?

Merge commit is fine, I may squash commits that are not terribly insightful.

nyxt.asd
jmercouris: I prefer gobject/gtk despite how wordy it is :-D

Ambrevar:
> I had to manually generate the bindings/fix them myself some time ago.

If I get it right, we just need to fix the webkitgtk package on Darwin to include the GI files, right?

> Therefore, it is not reasonable that it would be runnable on Darwin. 

What do you mean, "not reasonable"?

jmercouris:
> > I had to manually generate the bindings/fix them myself some time ago.
> 
> If I get it right, we just need to fix the webkitgtk package on Darwin to include the GI files, right?
That sounds simple, but it is actually quite complex, I have an open ticket about this for about a year now :-D

> > Therefore, it is not reasonable that it would be runnable on Darwin.
> 
> What do you mean, "not reasonable"?

I mean very unlikely, I don't think anyone would be capable of running it on Darwin unless they are heavily involved in MacPorts as I was :-D

jmercouris:
> About keeping the commits: no problem, but what about:
> 
> * Squashing the first 7-8 commits which are not very useful.
> * Merging instead of rebasing?

Done!

jmercouris:
Well, it seems the problem has actually been solved, should be runnable on Darwin!

https://trac.macports.org/ticket/60024

Assuming this trac ticket is correct and the problem IS fixed.

Ambrevar:
Oops! Sorry about the mess :p

jmercouris:
Ay yai yai :-D

jmercouris:
I better merge while our repository still exists :-P

jmercouris:
In any case, I can't really test on macOS anymore... for now..

Ambrevar:
Hahaha! :D  Yup, I thought I was deleting a local branch, and accidentally deleted origin/gobject :(  Oopsy!

Ambrevar:
OK, it's working for me!  Writing from Nyxt @ GI now :)
I had to add the following to the Guix recipe:

```scheme
       ("gobject-introspection" ,gobject-introspection)
       ("cl-gobject-introspection" ,sbcl-cl-gobject-introspection)
       ("gtk" ,gtk+)
       ("webkitgtk" ,webkitgtk)
```

jmercouris:
Nice!

Ambrevar:
Actually it seems that the `webkitgtk` input is not needed for now. I assume this is because we never
```
(gir:require-namespace "WebKit2")
```
for now, but as soon as we do it will be required.

jmercouris:
Yes, I would advise to leave it in, because as soon as we port some of the previous GO code back in, it will be necessary

jmercouris:
Were you able to print still? Did you test copy/paste as well on your machine? I assume they should work, but I just want to verify before I merge this in.

Ambrevar:
I could call the print box and cancel it twice.
I'll try printing to PDF later.

I can copy-paste, but still with the work around installed, so that's
cheating.  Need to disable the workaround first.

Shall we define a feature macro for each renderer, so that we can have
code conditional to those renderers?
Otherwise, let's move the copy-paste workaround to renderer-gtk.lisp.

jmercouris:
I suggest we just move the copy paste workaround to renderer-gtk.lisp for now,

IF the user loads multiple renderers, then we will want the conditional feature, until we figure out those details though, that will be the easiest thing to do

jmercouris:
That was my experience too, printing worked :-)

jmercouris:
I've re-enabled copy/paste context menu in this port. We'll remove the workaround disabling normal copy/paste via keyboard on the main branch then!

Ambrevar:
I confirm that printing to PDF works!
Hooray!

Thank you very much for finally nailing this down! :D

jmercouris:
No problem monsieur!

PR #1106, opened 2021-01-22
Copy paste

jmercouris:
WIP

### DIFF
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -48,11 +48,6 @@ search.")
        "M-g" 'follow-hint-new-buffer-focus
        "C-u M-g" 'follow-hint-new-buffer
        "C-x C-w" 'copy-hint-url
-       ;; We bind C-y to paste to override the built-in GTK binding
-       "C-y" 'paste
-       "C-v" 'paste
-       "shift-insert" 'paste
-       "button2" 'paste-or-set-url
        "C-c" 'copy
        "button9" 'history-forwards
        "button8" 'history-backwards
@@ -90,7 +85,6 @@ search.")
        "keypadnext" 'scroll-page-down
        "keypadpageup" 'scroll-page-up
        "keypadprior" 'scroll-page-up)
-
       scheme:emacs
       (list
        "C-M-f" 'history-forwards-all-query
@@ -139,15 +133,13 @@ search.")
        "; f" 'follow-hint-new-buffer
        "button9" 'history-forwards
        "button8" 'history-backwards
-       "C-v" 'paste
-       "shift-insert" 'paste
        "+" 'zoom-in-page
        "hyphen" 'zoom-out-page
        "0" 'unzoom-page
        "z i" 'zoom-in-page
        "z o" 'zoom-out-page
        "z z" 'unzoom-page
-       "g h" 'jump-to-heading ; REVIEW: VI binding?  "gh" is probably good enough.
+       "g h" 'jump-to-heading
        "/" 'search-buffer
        "?" 'remove-search-hints
        "m f" 'bookmark-hint
@@ -162,12 +154,7 @@ search.")
        "space" 'scroll-page-down
        "s-space" 'scroll-page-up
        "pageup" 'scroll-page-up
-       "pagedown" 'scroll-page-down))))
-  ;; Init.
-  ;; TODO: Do we need to set the default URL?  Maybe not.
-  ;; (buffer-load (default-new-buffer-url (buffer %mode))
-  ;;                 (buffer %mode))
-  )
+       "pagedown" 'scroll-page-down)))))
 
 (sera:export-always '%clicked-in-input?)
 (define-parenscript %clicked-in-input? ()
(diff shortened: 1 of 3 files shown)


jmercouris:
Appears to be ready for review, I don't believe there is anything we should change in %paste.

I've left the bindings for C-y for %paste, because I am not sure how we would invoke the GTK native paste, I think this is OK for now.

source/renderer-gtk.lisp
```diff
@@ -3,6 +3,8 @@
 
 (in-package :nyxt)
 
+(push :renderer-gtk *features*)
```
jmercouris, line 6: This is one of those things that again doesn't exactly make sense. For example, if we have multiple renderers loaded, how will an exec'ing piece of code know whether to respect a feature or not?

Consider that paste works in renderer-gobject-gtk, but not in renderer-gtk. When we invoke C-v, what is our system to do? This is perhaps something that should be set with set-renderer as well, and ideally the user should only run a single renderer at a time. Actually, now that I am thinking about it, perhaps we don't need to use set-renderer, and we can use a feature to selectively define user classes?

Ambrevar:

> Appears to be ready for review, I don't believe there is anything we should change in %paste.
>
> I've left the bindings for C-y for %paste, because I am not sure how we would invoke the GTK native paste, I think this is OK for now.

If I'm not mistaken, GTK binds C-y to paste by default :)

That said, it would be nice to have our own command for pasting to let
the user customize the binding.

The current implementation works, but depends on trivial-clipboard
(which depends on xclip/xsel and has already cause troubles in the
past).
Maybe some FFI?

Ambrevar:
Good point!

Hmmm... I need to sleep on this.

jmercouris:
> Appears to be ready for review, I don't believe there is anything we should change in %paste. I've left the bindings for C-y for %paste, because I am not sure how we would invoke the GTK native paste, I think this is OK for now.
> If I'm not mistaken, GTK binds C-y to paste by default :) That said, it would be nice to have our own command for pasting to let the user customize the binding. The current implementation works, but depends on trivial-clipboard (which depends on xclip/xsel and has already cause troubles in the past). Maybe some FFI?

It does not appear that C-y is bound to paste by default, at least not on my system. Perhaps we can use some FFI in the future. Then it would be renderer specific though... Hm...

Ambrevar:
I swear someone had previously reported that C-y (or was it M-y?) was
hanging the browsing in CUA bindings, but I can't find the report again
on the issue tracker.

jmercouris:
Shall I merge this in?

Ambrevar:
Wait, I'll give it a try.

source/web-mode.lisp
```diff
@@ -162,12 +159,7 @@ search.")
        "space" 'scroll-page-down
        "s-space" 'scroll-page-up
        "pageup" 'scroll-page-up
-       "pagedown" 'scroll-page-down))))
-  ;; Init.
-  ;; TODO: Do we need to set the default URL?  Maybe not.
-  ;; (buffer-load (default-new-buffer-url (buffer %mode))
-  ;;                 (buffer %mode))
-  )
```
Ambrevar, line 170: Suggestion: when you remove old TODOs, add a comment in the commit message explaining why you think this TODO can be removed.
What do you think?

Ambrevar:
Just tested.

The feature-macro approach actually doesn't work: it's set _after_ the
rest of Nyxt is loaded, so the code enabled for renderer-gtk will never
be enabled, even for the original GTK renderer.

Related: The #-darwin should not be removed, because pasting works on Darwin.

So I suggest we just leave the `paste` bindings as they are.  They don't
do any hard and give more control to the user.

I couldn't find a way to rebind pasting in WebKitGTK.
@aartaka Any idea?

And while we are at it, the weirdest thing: Pasting works for me now
with cl-cffi-gtk! :p
What about you?

jmercouris:
Copy /paste now works for me. Printing buffer still crashes. I suggest we remove the overrides for copy/paste then, and leave the functions in case users want to bind them.

source/web-mode.lisp
jmercouris: OK! I think that is a good idea!

Ambrevar:
OK to merge on my end :)

jmercouris:
done!

aartaka:
> I couldn't find a way to rebind pasting in WebKitGTK.
> @aartaka Any idea?

That seems to require wrestling with [`GAction`](https://developer.gnome.org/gio/stable/GAction.html) (used for e.g., [`WebKitContextMenuItem`](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitContextMenuItem.html#webkit-context-menu-item-new-from-gaction)), and/or [GTK+ clipboards](https://developer.gnome.org/gtk3/stable/gtk3-Clipboards.html). Neither is yet bound in `cl-cffi-gtk`.

Ambrevar:
Thanks for investigating this, @aartaka!
I've opened issue #1108 about this.

PR #1110, opened 2021-01-27
WIP: Global history tree library

Ambrevar:
This is meant to supersede #1007.
The difference with #1007 is that now all the "global history tree" logic is moved to `libraries/history-tree`.
Besides, there are some important design differences:

- Unrelated buffers start off from a new root, they don't share any node.
- Each node can have multiple owners.  There is no such thing as a current node owner anymore.
- However there is a "current owner", which is just a shorthand to avoid having to specify the owner each time.
- A history has a list of `entries` (for now as a hash-table), which is all the raw data held by the history.
  So everything pointed to by nodes is in there, but there are also entries that don't match any node.  This is useful to remember the visited pages when you kill all buffers.  This allows us to import the former global history.

Status:
- Compiles.
- Starts.

Rest is broken, see below :p 

What's left to be done:

- [x] Serialization: This is what prevents us from loading any page right now.
- [x] Deserialization: To restore the history.
- [x] Need to use high-level functions such as `back` and `forward` instead of `visit` in `set-url-from-history`.
- [x] Fix `history-tree` and `buffer-history-tree`.
- [X] Fix tests :) 

Feedback welcome!  The `libraries/history-tree` files should be quite stable at this point, all that's left to be done will happen in `source` I think.  
@aartaka @jmercouris If you can have a look at it, that'd be great!  In particular:

- I have `value` method to retrieve the raw data from a node, but now I realize that this changes many calls in `source` for no good rease.  Rename it to `data` as it was originally?

- [x] @aartaka had renamed many `all-*` functions to `*-nodes`.  While writing the library, at some point it felt more natural to use the `all-*` format.  But now I'm wondering if this was wise.  Revert to `*-nodes` instead?
- [x] Rename `owner`s slot `creator` to `creator-identifier`?
- [x] Rename `history-tree` to `history`?
  No.
- [x] Do we need the `all-*-data` functions?  Seems rather trivial to me, and we are not calling them in `source`.
- [x] We have a bunch of functions related to the group of nodes that are contiguously owned by a given owner: `connected-size`, `all-contiguous-owned-nodes`, `node-contiguous-owned-parents`, etc.  Which term is better: `connected` or `contiguous`?  Maybe some other term altogether, like `cluster` or `group`? (Too vague?)


Ambrevar:
About buffer deletion: see the package documentation, I've explained it towards the end.

Ambrevar:
Important design question: what shall we do when "jumping" nodes, such as visiting a distant unowned parent or child? 
- Mark all the intermediate nodes as owned?
- Or not?

Which one is more natural?  If we mark them all as owned, the benefit is that it guarantees connectedness for all owners.  The drawback is that some nodes will be marked as owned (in other words, "visited") even though they weren't.

Consider this example:

```lisp
url1 (A) -> url2 (A) -> url3 (B)
```

Current buffer is B.
We go "all-backwards" to url1.

- If we own all nodes, then `url2` is marked as visited, even though it wasn't.
  But if we go `forward`, then we move to `url2`.
- If we jump nodes instead, then B history is made of the disconnected nodes pointed to `url1` and `url3`.
  If we go `forward`, Nyxt will message the user that no (owned) forward node exists.

Thoughts? 

P.S.: I think I've figured out serialization.

aartaka:
Wow, that's massive! 

>     * Unrelated buffers start off from a new root, they don't share any node.

As we discussed, we need to add at least a bit of customizability to it (e.g. `always-branch`  browser-slot).

>     * Each node can have multiple owners.  There is no such thing as a current node owner anymore.

How do we resolve the issue of defining the owner to draw glyphs in the `history-tree`? Are these even relevant anymore?

> Feedback welcome! The `libraries/history-tree` files should be quite stable at this point, all that's left to be done will happen in `source` I think.

Give me some time to process it :)

>     * I have `value` method to retrieve the raw data from a node, but now I realize that this changes many calls in `source` for no good rease.  Rename it to `data` as it was originally?

Hmmm, yes, that makes sense. `value` is also somewhat vague in this particular domain. At least it's more vague than `data` :)

>     * @aartaka had renamed many `all-*` functions to `*-nodes`.  While writing the library, at some point it felt more natural to use the `all-*` format.  But now I'm wondering if this was wise.  Revert to `*-nodes` instead?

`*-nodes` was my choice because `all-*` didn't suit some of the functions there. As far as I can see, this is not an issue anymore -- `all-*` suits all the functions of this category well enough :)

>     * Rename `owner`s slot `creator` to `creator-identifier`?

No opinion about that. I'd leave that for the time being and change it if it becomes unclear enough or shoots someone in the foot :)

>     * Rename `history-tree` to `history`?

No. Although this is no longer a straightforward tree, tree-like relations are still of importance.
 
>     * Do we need the `all-*-data` functions?  Seems rather trivial to me, and we are not calling them in `source`.

Really? 0_o

Then we can remove these.

>     * We have a bunch of functions related to the group of nodes that are contiguously owned by a given owner: `connected-size`, `all-contiguous-owned-nodes`, `node-contiguous-owned-parents`, etc.  Which term is better: `connected` or `contiguous`?  Maybe some other term altogether, like `cluster` or `group`? (Too vague?)

`contiguous` fits better there. I can't come up with a better alternative ¯\\\_(ツ)_/¯

aartaka:
> Important design question: what shall we do when "jumping" nodes, such as visiting a distant unowned parent or child?
> 
>     * Mark all the intermediate nodes as owned?
> 
>     * Or not?
> 
> 
> Which one is more natural? If we mark them all as owned, the benefit is that it guarantees connectedness for all owners. The drawback is that some nodes will be marked as owned (in other words, "visited") even though they weren't.
> 
> Consider this example:
> 
> ```lisp
> url1 (A) -> url2 (A) -> url3 (B)
> ```
> 
> Current buffer is B.
> We go "all-backwards" to url1.
> 
>     * If we own all nodes, then `url2` is marked as visited, even though it wasn't.
>       But if we go `forward`, then we move to `url2`.
> 
>     * If we jump nodes instead, then B history is made of the disconnected nodes pointed to `url1` and `url3`.
>       If we go `forward`, Nyxt will message the user that no (owned) forward node exists.

Hmmmm. It seems most suitable to me to utilize `binding`s there and just add `url3` as a `forward-child` of `url1` for B. This, however, breaks the order of tree traversal (e.g. `all-contiguous-owned-(parents|children)`) and there's no way to trace parent nodes (`binding` does only have `forward-child` slot, so it's singly-linked). Maybe we should extend `binding` to incorporate both `forward-child` and `parent`? This way owned node interrelations can be traced through the bindings. But then `binding` becomes a kind of `node` twin...

To reiterate: I'm in favor of visiting only what's actually visited and storing the owner-specific interrelations. This is complex, though...

Ambrevar:
> >     * Unrelated buffers start off from a new root, they don't share any node.
> 
> As we discussed, we need to add at least a bit of customizability to it
> (e.g. `always-branch` browser-slot).

Sorry, misunderstanding.  I mean on the library level, unrelated owners have
unrelated roots.

The question of branching off or not belongs to Nyxt.  And yes, I still need to
add a configuration slot :)

> How do we resolve the issue of defining the owner to draw glyphs in the
> `history-tree`? Are these even relevant anymore?

Indeed, this is no longer relevant.  Even before the glyphs drawing was a
bit dubious since it would change depending on which buffer was last focused :p.

For now I suggest to simply draw the tree of contiguous nodes for the current
buffer.

For the entire tree, we could add the list of owners as tooltips for instance.
Still need to think this through, but it's not urgent.

Ambrevar:
> To reiterate: I'm in favor of visiting only what's actually visited and storing
> the owner-specific interrelations. This is complex, though...

But is it what the user expects though?

When you run `all-backwards`, you are prompted a list of parents, here "url2"
and "url1".
So if you go to "url1", maybe you'd expect that "url2" would come up as a
forward-child.

Your suggestion makes sense, but as you said, it induces a significant cost in
terms of complexity (it turns the tree into a DAG).  Is it worth it, if it's all
just a subjective mental model?

In the end, the user has the minibuffer to fuzzy-search parents and children, so
adding intermediate nodes does not burden the user much since it's so easy to
skip them.

Thoughts?

jmercouris:
> * Unrelated buffers start off from a new root, they don't share any node.
> * Each node can have multiple owners.  There is no such thing as a current node owner anymore.
> * However there is a "current owner", which is just a shorthand to avoid having to specify the owner each time.
> * A history has a list of `entries` (for now as a hash-table), which is all the raw data held by the history.
>   So everything pointed to by nodes is in there, but there are also entries that don't match any node.  This is useful to remember the visited pages when you kill all buffers.  This allows us to import the former global history.
> 
> Status:
> 
> * Compiles.
> * Starts.
> 
> Rest is broken, see below :p
> 
> What's left to be done:
> 
> * Serialization: This is what prevents us from loading any page right now.
> * Deserialization: To restore the history.
> * Need to use high-level functions such as `back` and `forward` instead of `visit` in `set-url-from-history`.
> * Fix `history-tree` and `buffer-history-tree`.
> * Fix tests :)
> 
> Feedback welcome! The `libraries/history-tree` files should be quite stable at this point, all that's left to be done will happen in `source` I think.
> @aartaka @jmercouris If you can have a look at it, that'd be great! In particular:
> 
> * I have `value` method to retrieve the raw data from a node, but now I realize that this changes many calls in `source` for no good rease.  Rename it to `data` as it was originally?

Yes.

> * @aartaka had renamed many `all-*` functions to `*-nodes`.  While writing the library, at some point it felt more natural to use the `all-*` format.  But now I'm wondering if this was wise.  Revert to `*-nodes` instead?

I think either way is OK.

> * Rename `owner`s slot `creator` to `creator-identifier`?

`creator-id`

> * Rename `history-tree` to `history`?

No.

> * Do we need the `all-*-data` functions?  Seems rather trivial to me, and we are not calling them in `source`.

If not using, do not need. Could be useful for extension writers though.

> * We have a bunch of functions related to the group of nodes that are contiguously owned by a given owner: `connected-size`, `all-contiguous-owned-nodes`, `node-contiguous-owned-parents`, etc.  Which term is better: `connected` or `contiguous`?  Maybe some other term altogether, like `cluster` or `group`? (Too vague?)

The term used most often in the literature is connected. I would stick with that.

This is a big changeset, lots to see here :-)

I'm going to need some time to digest it all.

jmercouris:
Actually, I would suggest working on the new prompt-buffer while we think about these changes, and here is why:

1. it will unblock me and I will be able to help you in porting over all of the invokations to use the new prompt buffer
2. it will give us some time to digest all of these changes
3. it may provide you with a welcome change of pace

what do you think ?

Ambrevar:

> Actually, I would suggest working on the new prompt-buffer while we think about these changes, and here is why:
>
> 1. it will unblock me and I will be able to help you in porting over all of the invokations to use the new prompt buffer
> 2. it will give us some time to digest all of these changes
> 3. it may provide you with a welcome change of pace
>
> what do you think ?

I disagree: this change set is a blocker for Artyom.  Besides, my head
is still fresh on this issue.  Juggling back and forth between issues
induces a cognitive load which costs me extra time.

Last, I'm done with it in just a few hours :)

jmercouris:
OK :-D

Ambrevar:
Lo and behold!  The Great History Tree works! :D

I've just fixed critical issues, and overall it seems to be working as expected.
Rest assured that some bugs are still lurking around...

What's new:

- (De)Serialization is implemented.
- High-level API of `libraries/history-tree` is now properly used.
  The separation is neat, almost all the logic is contained within the library.
  
  Exception: The buffer restoration is done in Nyxt since it involves the buffer
  IDs Nyxt generates.  Maybe we could abstract this better.  Any idea?
  
- Old flat buffer is automatically restored!

- Branching now happens when middle-clicking or on C-button1 on a link.  It
  should also happen in other places, like link-hinting (not well tested, please do!)

- `history-forwards-to-dead-history-p` is commented out, since I don't think it
  makes much sense anymore.  Thoughts, @aartaka?
  
- Improved logging (debug), remove some handler-case when running from the REPL
  (so that we get the stack-traces).

What's left:

- Some renaming mentioned above (e.g. "value -> data", "connected", etc.).

- Rename `back` to `backward`, to be more consistent with `forward`?

- Set the default buffer (easy).

- Replace `entry` function slots by a single slot pointing to the
  `history-tree`.  This will make the serialized data much lighter.

- Store `data-last-access` in `entry`, since otherwise node-less entries would
  lose this precious piece of information.

- Re-enable the use of `history-forwards-prompting-p`.

- The comment in `history-add` says we can't use `with-data-access`.  @aartaka,
  is this still true?
  
- Now sessions are gone, `make-startup-function` should ask the use which
  history file to restore.
  

Overall, it's just easy stuff, beside testing of course.

@aartaka @jmercouris Please test and report if anything goes wrong for you.
With a bit of luck, we can merge tomorrow :)

jmercouris:
> Lo and behold! The Great History Tree works! :D

This is... amazing! A monumentally huge challenge. Kudos to you and Artyom both! Great perseverance on seeing it through to something beautiful. I don't think I've seen this concept executed (to this degree) in any other program- ever.

Well anyways, what more is there to say... I cannot exaggerate how significant this is!

> I've just fixed critical issues, and overall it seems to be working as expected.
> Rest assured that some bugs are still lurking around...

No doubt about it :-D. It is OK though, that's what pre-releases are for anyways, I'm sure we'll find more :-O

>   The separation is neat, almost all the logic is contained within the library.

Very nice! This is incredibly important as Nyxt grows.

>   Exception: The buffer restoration is done in Nyxt since it involves the buffer
>   IDs Nyxt generates.  Maybe we could abstract this better.  Any idea?

Not sure what I would move where in this case. I think this is OK.

> * Old flat buffer is automatically restored!
> * Branching now happens when middle-clicking or on C-button1 on a link.  It
>   should also happen in other places, like link-hinting (not well tested, please do!)
> 
> What's left:
> 
> * Some renaming mentioned above (e.g. "value -> data", "connected", etc.).
> * Rename `back` to `backward`, to be more consistent with `forward`?

Agreed.

> * Set the default buffer (easy).
> * Replace `entry` function slots by a single slot pointing to the
>   `history-tree`.  This will make the serialized data much lighter.
> * Store `data-last-access` in `entry`, since otherwise node-less entries would
>   lose this precious piece of information.
> * Re-enable the use of `history-forwards-prompting-p`.
> * The comment in `history-add` says we can't use `with-data-access`.  @aartaka,
>   is this still true?
> * Now sessions are gone, `make-startup-function` should ask the use which
>   history file to restore.
> 
> Overall, it's just easy stuff, beside testing of course.
> 
> @aartaka @jmercouris Please test and report if anything goes wrong for you.

I will be sure to report any failures on my system!

> With a bit of luck, we can merge tomorrow :)

Yay!

Once again, a lot of thanks to you and Artyom. Great work.

aartaka:
>     * High-level API of `libraries/history-tree` is now properly used.
>       The separation is neat, almost all the logic is contained within the library.
>       Exception: The buffer restoration is done in Nyxt since it involves the buffer
>       IDs Nyxt generates.  Maybe we could abstract this better.  Any idea?

I agree with John here -- buffer restoration belongs to the Nyxt side.

>     * Branching now happens when middle-clicking or on C-button1 on a link.  It
>       should also happen in other places, like link-hinting (not well tested, please do!)

Sure will test!

>     * `history-forwards-to-dead-history-p` is commented out, since I don't think it
>       makes much sense anymore.  Thoughts, @aartaka?

It doesn't make much sense, that's right. Feel free to delete it :)

>     * Some renaming mentioned above (e.g. "value -> data", "connected", etc.).
> 
>     * Rename `back` to `backward`, to be more consistent with `forward`?

Sure!

>     * Store `data-last-access` in `entry`, since otherwise node-less entries would
>       lose this precious piece of information.

This doesn't feel right. Isn't access time a meta-information (that should be stored in `node`s) there? 

However, I can't come up with a good way to both preserve access time and separate things neatly. This (storing access time in entries) seems to be the way to go.

>     * The comment in `history-add` says we can't use `with-data-access`.  @aartaka,
>       is this still true?

Yes, this is still true. The reason is, history is locked by the e.g. `set-url` right at the moment we try to access it in `history-add`. Two mutually locking accesses hang Nyxt permanently.

I've thought about it overnight. Maybe we need two kinds of data access -- one to just check the data, and one to modify it? This way we can check data in `set-url` (it's only for history completion) and modify it in `history-add`, all with properly syncronised history. So what I propose is to have two data-access macros:
- `with-data-access` (`with-data-modification`?) -- to modify data and sync it with the storage:
   - `restore`s data.
   - Locks data.
   - `store`s data back to where it came from.
- `with-data-refresh` (maybe `with-data-access` suits it better) -- to just compute something based on the data without modifying it:
   - `restore`s data.
   - **Does not** lock or store data.
   - Maybe we should somehow prohibit `setf` usage in the scope of this macro?

This can wait, though.

aartaka:
Just tested, feels dizzy yet powerful! `history-tree` would come in handy is checking the tree integrity visually, but it looks good as far as I got with SLY inspector :)

jmercouris:
```
The symbol "ROOT" is not external in the HISTORY-TREE package.

  Line: 4, Column: 22, File-Position: 150

  Stream: #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}>
   [Condition of type SB-INT:SIMPLE-READER-PACKAGE-ERROR]

Restarts:
 0: [CONTINUE] Use symbol anyway.
 1: [ABORT] abort thread (#<THREAD "Anonymous thread" RUNNING {100B8BAAA3}>)

Backtrace:
  0: (SB-IMPL::READ-TOKEN #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #\H)
  1: (SB-IMPL::READ-MAYBE-NOTHING #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #\H)
  2: (SB-IMPL::READ-LIST #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #<unused argument>)
  3: (SB-IMPL::READ-MAYBE-NOTHING #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #\()
  4: (SB-IMPL::READ-LIST #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #<unused argument>)
  5: (SB-IMPL::READ-MAYBE-NOTHING #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #\()
  6: (SB-IMPL::READ-LIST #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #<unused argument>)
  7: (SB-IMPL::READ-MAYBE-NOTHING #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #\()
  8: (SB-IMPL::READ-LIST #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #<unused argument>)
  9: (SB-IMPL::READ-MAYBE-NOTHING #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #\()
 10: (SB-IMPL::READ-LIST #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #<unused argument>)
 11: (SB-IMPL::READ-MAYBE-NOTHING #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #\()
 12: (SB-IMPL::%READ-PRESERVING-WHITESPACE #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> NIL (NIL) T)
 13: (SB-IMPL::%READ-PRESERVING-WHITESPACE #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> NIL (NIL) NIL)
 14: (READ #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> NIL :EOF NIL)
 15: (HISTORY-DESERIALIZE-SEXP #<SB-SYS:FD-STREAM for "file /home/jmercouris/.local/share/nyxt/history/default.lisp" {100C8855E3}> #<S-SERIALIZATION::SERIALIZATION-STATE {100C934DD3}>)
 16: ((LABELS %RESTORE :IN RESTORE))
 17: ((FLET "WITHOUT-INTERRUPTS-BODY-29" :IN SB-THREAD::CALL-WITH-RECURSIVE-LOCK))
 18: (SB-THREAD::CALL-WITH-RECURSIVE-LOCK #<CLOSURE (FLET SB-THREAD::WITH-RECURSIVE-LOCK-THUNK :IN %GET-USER-DATA) {7F01082FE5BB}> #<SB-THREAD:MUTEX "Anonymous recursive lock" owner: #<SB-THREAD:THREAD "An..
 19: (%GET-USER-DATA #<DEFAULT-DATA-PROFILE {100A8FD453}> #<HISTORY-DATA-PATH {100A931353}> #<HASH-TABLE :TEST EQUAL :COUNT 1 {10019887A3}>)
 --more--
```

jmercouris:
```
  #<STANDARD-GENERIC-FUNCTION HISTORY-TREE::ROOT (1)>
when called with arguments
  (#<HISTORY-TREE:HISTORY-TREE {100A353323}>).
   [Condition of type SB-PCL::NO-APPLICABLE-METHOD-ERROR]
See also:
  Common Lisp Hyperspec, 7.6.6 [:section]

Restarts:
 0: [RETRY] Retry calling the generic function.
 1: [ABORT] abort thread (#<THREAD "Anonymous thread" RUNNING {100264BF33}>)

Backtrace:
  0: ((:METHOD NO-APPLICABLE-METHOD (T)) #<STANDARD-GENERIC-FUNCTION HISTORY-TREE::ROOT (1)> #<HISTORY-TREE:HISTORY-TREE {100A353323}>) [fast-method]
  1: (SB-PCL::CALL-NO-APPLICABLE-METHOD #<STANDARD-GENERIC-FUNCTION HISTORY-TREE::ROOT (1)> (#<HISTORY-TREE:HISTORY-TREE {100A353323}>))
  2: (HISTORY-TREE:MAP-TREE #<CLOSURE (LAMBDA (NYXT/WEB-MODE::NODE) :IN NYXT/WEB-MODE:BUFFER-HISTORY-TREE) {1008EFB83B}> #<HISTORY-TREE:HISTORY-TREE {100A353323}> :FLATTEN NIL :INCLUDE-ROOT T :COLLECT-FUNC..
  3: (NYXT/WEB-MODE:BUFFER-HISTORY-TREE #<USER-WEB-BUFFER {10098B1D63}>)
  4: (FUNCALL-SAFELY #<FUNCTION NYXT/WEB-MODE:BUFFER-HISTORY-TREE>)
  5: ((LAMBDA NIL :IN NYXT::RUN-ASYNC))
  6: ((LAMBDA NIL :IN BORDEAUX-THREADS::BINDING-DEFAULT-SPECIALS))
  7: ((FLET SB-UNIX::BODY :IN SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE))
  8: ((FLET "WITHOUT-INTERRUPTS-BODY-4" :IN SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE))
  9: ((FLET SB-THREAD::WITH-MUTEX-THUNK :IN SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE))
 10: ((FLET "WITHOUT-INTERRUPTS-BODY-1" :IN SB-THREAD::CALL-WITH-MUTEX))
 11: (SB-THREAD::CALL-WITH-MUTEX #<CLOSURE (FLET SB-THREAD::WITH-MUTEX-THUNK :IN SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE) {7F5CC0BAED8B}> #<SB-THREAD:MUTEX "thread result lock" owner: #<SB-THREAD:THREAD "Ano..
 12: (SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE #<SB-THREAD:THREAD "Anonymous thread" RUNNING {100264BF33}> NIL #<CLOSURE (LAMBDA NIL :IN BORDEAUX-THREADS::BINDING-DEFAULT-SPECIALS) {100264BEDB}> NIL)
 13: ("foreign function: call_into_lisp")
 14: ("foreign function: new_thread_trampoline")
 --more--

```

jmercouris:
hm, having some problems on my machine... `history/history.lisp` appears to be malformed. Content missing. Not sure why, perhaps an error in how my system exists, let me try clearing the cache.

jmercouris:
I find the error message `"History entry is already the current URL` misleading when trying to navigate forward and there are no more forward entries.

jmercouris:
I deleted all of my history, and most things started working, `buffer-history-tree` does not work still, nor `history-tree`. However, it does work as expected otherwise! With regards to navigation!

Ambrevar:
> The symbol "ROOT" is not external in the HISTORY-TREE package.

@jmercouris: Artyom's history is not compatible with this patch, not should you copy the old master history to the `history/` folder.  It's all done automatically now.

Ambrevar:
> I find the error message "History entry is already the current URL misleading when trying to navigate forward and there are no more forward entries.

Agreed, will change it.

Ambrevar:
I'll fix the history-tree command as soon as possible, it's the most urgent thing right now.  Rest will come later today :)

Ambrevar:
Please keep reporting, also have a look at the commits from 2 days ago onward for code review.  Thanks!

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+;; TODO: Use fast sets for unique lookups?  Turns out hash-tables are overkill
+;; with SBCL on modern hardware, for less than 10.000.000 entries.
+;; Always use lists then?
+;; See TODO notes below mentioning "fast sets".
+;; See https://old.reddit.com/r/Common_Lisp/comments/l1z7ei/fast_set_library_or_gethash_vs_findmember/.
+
+;; TODO: Should we have different functions for finding nodes vs. "owned nodes",
+;; or pass an option as key argument?
+
+;; TODO: Is "Shared history tree" a better name than "Global history tree"?
```
jmercouris, line 19: Yes, it is a better name, but we are already here :-D

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+;; TODO: Turn unique defmethod to defuns.
```
jmercouris, line 20: Why? Performance? If performance, the gain is not worth considering.

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+
+(defmacro export-always (symbols &optional (package nil package-supplied?)) ; From serapeum.
```
jmercouris, line 22: I'm not understanding, why not just use Serapeum? Didn't want to bring in the whole package just for this snippet? If so, I agree.

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+(define-class node ()
+  ((parent nil
+           :type (or null node)
+           :documentation "If nil, it means the node is a root node.
+(The first of the parents.)")
+   (children '()
+             :type (list node)
+             :documentation "Order does not matter.")
+   (bindings (make-hash-table)
+             :documentation "The key is an `owner', the value is a
```
jmercouris, line 70: what is a `binding`? A reference? If so, I would suggest to name this to the more commonly understood term of `references`.

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+  (:export-class-name-p t)              ; TODO: Unexport?
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer #'class*:name-identity)
+  (:documentation "Node structure of the history tree."))
+
+(export 'value)
+(defmethod value ((node node))
+  (value (entry node)))
+
+(defmethod root ((node node))
```
jmercouris, line 87: very clever

libraries/history-tree/history-tree.lisp
Ambrevar: Nope, it's for the idiom / good practice.  A defmethod that is both internal and never dispatches on anything is confusing.  It feels like code is missing.

libraries/history-tree/history-tree.lisp
Ambrevar: Indeed!  This library has only 2 dependencies and this is the only reference (a trivial one) to one of Serapeum's helpers.

libraries/history-tree/history-tree.lisp
jmercouris: I see the class definition below, I just want to clarify that it is in fact a reference.

Ambrevar: A binding is, well, a binding between an owner and the ownee, i.e. the node.
I like the term "binding" because it's lexically connected with "owner".

In particular, it's not a reference because it goes both ways.

Another term I considered was "ownership".

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+
+The higher-level functions take care of adding entries for you, so you normally
+need not call this function.  See `add-child' instead.
+One case in which this function might be useful is when you want to import flat
+history data, e.g. a list of visited URLs that's not bound to any owner."
+  (cl-custom-hash-table:with-custom-hash-table
+    (let ((new-entry (make-entry history data)))
+      (multiple-value-bind (existing-entry-nodes found?)
+          (gethash new-entry (entries history))
+        (if found?
```
jmercouris, line 261: I'm not sure if this should be a question mark or not, but, just pointing it out in case :-)

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+            (progn
+              (setf (value (entry (first existing-entry-nodes))) data)
+              (entry (first existing-entry-nodes)))
+            (progn
+              (setf (gethash new-entry (entries history)) '())
+              new-entry))))))
+
+(define-class history-tree ()           ; TODO: Rename `history'?
```
jmercouris, line 269: I don't think so, we may have a history-list in some future... history-tree is nice, specific, and clear

libraries/history-tree/history-tree.lisp
jmercouris: OK, understood, then binding it shall stay!

libraries/history-tree/history-tree.lisp
jmercouris: Not to me it doesn't, a method operates on objects. I prefer `defun` only when I am working with primitives/coding with a bag-of-functions style.

source/browser.lisp
```diff
@@ -322,12 +322,13 @@ This is useful to tell REPL instances from binary ones."
                                        url))))
 
 (declaim (ftype (function (list-of-strings &key (:no-focus boolean)))))
-(defun open-urls (urls &key no-focus)
+(defun open-urls (urls &key no-focus parent-buffer)
```
jmercouris, line 325: I'm not sure this makes a lot of sense to me. `open-urls` is a discrete function that should probably always generate a new parent.

source/history.lisp
```diff
@@ -257,9 +257,13 @@ instance of Nyxt."
   (let ((history (htree:make)))
     history))
 
+(defvar flat-history-path (make-instance 'history-data-path :basename "history")
```
jmercouris, line 260: can you please move this into the globals file?

jmercouris: actually, let's keep it here

jmercouris:
This is just an absolutely gigantic changeset. Would it be OK if I just read it, and then we go over it together on Monday? I think it will be far more efficient.

jmercouris:
I will still write comments here, I just would like to do the final review over telephone.

source/browser.lisp
Ambrevar: You can't generate a parent.  I guess you meant "open a parent-less / independent buffer".

I agree with you, but then we need to revise our calls to `open-urls`.  We use it in a few spots where we should probably call `make-buffer` instead.

libraries/history-tree/history-tree.lisp
Ambrevar: I know... :p  
But I like this syntax more, and it's so local, _and_ in an "external" library, that really it does not matter much :)

source/history.lisp
Ambrevar: Good catch!

Ambrevar:
Sure, no problem!
This change set took 2 months for 2 people to complete, so it will take some time to digest! :p

source/browser.lisp
jmercouris: I did mean open a parent less buffer (root node)

Ambrevar: To be picky (:p), and because the history-tree is so complex I'd like avoid any confusion: the buffer is never a "node" in history-tree parlance.  Buffers are "owners", which have an origin node.  If buffers are parent-less, then the origin node is a root node (no parent).

libraries/history-tree/history-tree.lisp
Ambrevar: > a method operates on objects

It's one way to look at things, but this approach is not very idiomatic in Common Lisp, because dispatching can be done on anything, be it objects or primitive types.  The concepts of class and methods are quite separate.  Here I'm sometimes using methods on whatever data (may be primitives) and it's never dispatched.

aartaka: Another advantage of `defmethod`s: these can be "advised" (in a sense, with :before, :after, and :around methods), while `defun`s cannot. So, `defmethod`s are better for library customizability. Let's go with methods for this reason and for the generality of it :)

Ambrevar: But if the function is unexported, then customizability is out of the equation anyways.
For exported functions, I agree! :)

source/web-mode.lisp
```diff
@@ -192,192 +202,251 @@ search.")
...
   (let ((input (prompt-minibuffer
                 :input-prompt "Navigate to"
                 :suggestion-function (history-all-suggestion-filter))))
     (when input
-      (set-url-from-history input))))
+      (with-data-access (history (history-path buffer))
+        (htree:visit-all history input))
+      (load-url-if-not-current input))))
+
+(defun integer->unicode-geometric (string-integer)
```
aartaka, line 345: Not needed if don't use glyphs. We don't.

aartaka (commented):
Oh God, so much things there. Lots of good code too, Pierre! Thanks for taking care of this <3

Overall my review is mostly style and typos :)

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,895 @@
...
+   (nodes '()
+          :type (or null (cons node))
+          :documentation "The list of all owned nodes."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer #'class*:name-identity)
+  (:documentation "The high-level information about an owner."))
+
+(defmethod (setf current) (value (owner owner))
+  "This setter protects against setting OWNER's `current' slot to a invalid object."
```
jmercouris, line 197: "a" -> "an"

source/web-mode.lisp
Ambrevar: Removed.

source/web-mode.lisp
```diff
@@ -192,192 +202,251 @@ search.")
...
-          (htree:forward (history mode))
-          (match (htree:current (history mode))
-            ((guard n n) (buffer-load (url (htree:data n))))))
-        (echo "No forward history."))))
-
-(defun history-backwards-suggestion-filter (&optional (mode (find-submode
-                                                             (current-buffer)
-                                                             'web-mode)))
+  (let ((new-node
+          (with-data-access (history (history-path buffer))
```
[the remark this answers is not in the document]
Ambrevar: No we can't in `history-forwards.`  This is only available with `history-forwards-all-query`.

source/start.lisp
```diff
@@ -161,9 +161,9 @@ Example: --with-path bookmarks=/path/to/bookmarks
   (unless *keep-alive*
     (uiop:quit 0 nil)))
 
-(define-command quit-after-clearing-session ()
-  "Clear session then quit Nyxt."
-  (uiop:delete-file-if-exists (expand-path (session-path (current-buffer))))
+(define-command quit-after-clearing-session () ; TODO: Rename?
```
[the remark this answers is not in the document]
Ambrevar: Actually, we could define "session" as list of buffers, in which case this command needs no renaming.

If you disagree, then which name do you suggest?
What about `quit-after-deleting-buffers`?

source/mode.lisp
```diff
@@ -195,7 +195,7 @@ toggle command, return the toggle command of the parent."
         (m (mode-command (class-name (original-class m)))))))
 
 (defun make-mode (mode-symbol buffer)
-  (log:debug mode-symbol buffer (mode-command mode-symbol))
+  ;; (log:debug mode-symbol buffer (mode-command mode-symbol))
```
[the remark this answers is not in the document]
Ambrevar: No, but I needed this to help with debugging.  I was in haste and forgot to apply this to master instead.  My bad.

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+  "Go COUNT parent up from the current owner node, if possible.
+Return (VALUES HISTORY CURRENT-NODE) so that `back' and `forward' calls can
+be chained."
+  (let ((owner (current-owner history)))
+    (check-type count positive-integer)
+    (when (parent (current owner))
+      (let ((former-current (current owner)))
+        (visit history (parent (current owner)))
+        ;; Put former current node back as forward-child if it is not already
+        ;; the case, e.g. if current node was set manually.
```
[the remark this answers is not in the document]
Ambrevar: It's possible if `visit` was called.  It's not exported, but who it might be some day if we decide we want to allow the user to jump to other nodes without visiting intermediary nodes.

Anyways, this extra check does not cost much.

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,862 @@
...
+(defun current-owner-node (history)
+  (current (current-owner history)))
+
+(export-always 'with-current-owner)
+(defmacro with-current-owner ((history owner-identifier) &body body)
+  "Locally switch owner for HISTORY.
+See `set-current-owner' to set the owner persistently.
+Owner is created if OWNER-IDENTIFIER does not match any owner.
+OWNER-IDENTIFIER can be any value, even NIL."
+  (let  ((old-owner-identifier (gensym)))
```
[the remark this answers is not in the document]
Ambrevar: Good catch!

Ambrevar:
Update:
- `history-tree` and `buffer-history-tree` are fixed.
- I'm now storing and serializaing "data last access".  @aartaka I wanted to do it in the library itself because it fits better there.  Then I could not store the timestamp in the Nyxt `history-entry` because that would mean the htree library would have to know that the data had such a slot.

Another thing to do:

- [ ] There is a general slow down of many actions (e.g. buffer switching) when the history grows big.  My guess is that this is because we serialize the history much more often now, and serialization is also more costly because of the many nested structures.

libraries/history-tree/history-tree.lisp
```diff
@@ -2,255 +2,895 @@
...
+    (remove-if (complement (alex:curry #'owned-p owner))
+               (children node))))
+
+(export-always 'owned-children)
+(defun owned-children (owner)
+  "Return the OWNER's owned children for the current node."
+  (funcall (owned-children-lister owner) (current owner) ))
+
+(export-always 'owned-parent)
+(defun owned-parent (owner node)
```
jmercouris, line 221: Is there no '-p' suffix becasue this is kind of a predicate, but with more features?

tests/test-global-history.lisp
```diff
@@ -16,32 +16,28 @@
...
               (quri:uri "http://example.org")
               :test #'quri:uri=
               "value has quri:uri")
-          (is (title (rest entry))
+          (is (title entry)
               ""
               "value has no title"))
-        (history-add (quri:uri "http://example.org") :title "foo")
-        (is (hash-table-count (nyxt:get-data path))
+        (nyxt::history-add (quri:uri "http://example.org") :title "foo")
```
jmercouris, line 31: as I understand it, this demonstrates that we do not add URIs several times, even if the title of the page changes?

source/web-mode.lisp
```diff
@@ -19,9 +19,7 @@
 
 (define-mode web-mode ()
   "Base mode for interacting with documents."
-  ((history (htree:make)
-            :type htree:history-tree)
-   (history-blocklist '("https://duckduckgo.com/l/")
+  ((history-blocklist '("https://duckduckgo.com/l/")
                       ;; TODO: Find a more automated way to do it.  WebKitGTK
```
jmercouris, line 23: They probably examine the request resource, in there, there is probably some property about the HTTP status. Perhaps if it is a redirect, then they just ignore it.

source/web-mode.lisp
```diff
@@ -192,192 +202,236 @@ search.")
   "Scroll to top if no input element is active, forward event otherwise."
   (call-non-input-command-or-forward #'scroll-to-top :buffer buffer))
 
-(declaim (ftype (function (htree:node &optional buffer)) set-url-from-history))
-(defun set-url-from-history (history-node &optional (buffer (current-buffer)))
+(defun load-url-if-not-current (url-or-node &optional (buffer (current-buffer)))
```
jmercouris, line 205: the name `load-url` I think should suffice. Describing the implementation details of the function is probably not best suited to the function name. We don't care how a URL is loaded or whether it is loaded or not, only that the function loads a URL. what do you think?

Ambrevar:
- [ ] Data access lock: @aartaka I agree with your assessment in https://github.com/atlas-engineer/nyxt/pull/1110#issuecomment-769654395.  Let's work it out later.

source/browser.lisp
jmercouris: I don't understand why we both need a history entry and a node class, can you please explain that?

libraries/history-tree/history-tree.lisp
Ambrevar: Yup, it returns the parent :)

tests/test-global-history.lisp
Ambrevar: Correct, but note that the title gets updated in the global history!

source/start.lisp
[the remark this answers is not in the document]
jmercouris: `quit-after-deleting-buffers` is good

source/web-mode.lisp
Ambrevar: Sadly it's not an HTTP redirect :(

jmercouris:
I think this changeset is too large to be easily reviewed. I keep reading and forgetting what I read :-O I think the in the future it would be a good idea to break apart the development process. For example, the following cycles:

1. development of the library
2. integration of the library

as two separate write/merge steps. Thoughts?

source/web-mode.lisp
Ambrevar: I think it's important when reading the callers code to not confuse this function with something more general purpose.

It's internal to `web-mode` and it's only used to load history URLs.
Call it `load-history-url` then?

Ambrevar:
The problem is that it's hard to develop the library without validating it by an integration process.

Having all in one PR is not a big deal: with Magit you can restrict the diff to the files in `source/` if you just want to review the integration, or the files in `libraries` if you just want to review the library changes.

On GitHub you can also filter by files or by commit.

source/browser.lisp
Ambrevar: Because multiple nodes may point to the same entry.
An entry (for Nyxt) is a URL+Title+...+last access date.
When the data other than the URL gets updated, we want to  be able to update it so that it reflects on all nodes pointing to the entry.

If we had the entry data directly in the nodes, we would have to loop across all nodes which would be problematic.

Another benefit of having a separate `entry` storage is that it allows us to remember the visited pages even when no node is pointing to them. (This is the case, for instance, when you delete all buffers.)

jmercouris:
How do I do that in Magit?

Ambrevar:
With `--`.

jmercouris:
Ah, I didn't realize that could accept a glob!

aartaka:
>     * I'm now storing and serializaing "data last access".  @aartaka I wanted to do it in the library itself because it fits better there.  Then I could not store the timestamp in the Nyxt `history-entry` because that would mean the htree library would have to know that the data had such a slot.

Does library need to know about the entries' last access if they are disowned? I believe it's out of scope of library there. It's the code that uses library that needs to know details of `data` stored in the tree. Is it used only once in `history.lisp` and a couple of (not mandatory) times in `libraries/history-tree.lisp`.

>     * [ ]  There is a general slow down of many actions (e.g. buffer switching) when the history grows big.  My guess is that this is because we serialize the history much more often now, and serialization is also more costly because of the many nested structures.

Is addressing separation of data access and data modification enough to fix that? If so, I can push the `with-data-(access|lookup)` (still looking for better names) distinguishing code to `master` in a couple hours and patch this branch accordingly afterwards.

source/web-mode.lisp
jmercouris: OK!

source/web-mode.lisp
jmercouris: How is it done JS?

tests/test-global-history.lisp
jmercouris: :-)!

source/browser.lisp
jmercouris: I still don't see exactly the benefit, I guess we can discuss this during our call today :-)

I would think a cache structure would be enough to mitigate your concerns. That cache structure could be embedded within the history tree data structure and not exposed.

Ambrevar:

> Ah, I didn't realize that could accept a glob!

You can also give it a directory, or even multiple files.

source/web-mode.lisp
Ambrevar: I don't know, could very well be a server side redirection.  Epiphany saves the intermediate URL too...

Ambrevar:

>>     * I'm now storing and serializaing "data last access".  @aartaka I wanted to do it in the library itself because it fits better there.  Then I could not store the timestamp in the Nyxt `history-entry` because that would mean the htree library would have to know that the data had such a slot.
>
> Does library need to know about the entries' last access if they are disowned? I believe it's out of scope of library there. It's the code that uses library that needs to know details of `data` stored in the tree. Is it used only once in `history.lisp` and a couple of (not mandatory) times in `libraries/history-tree.lisp`.

You are right.
That said, storing the data last access in the entry data or in the
htree is more or less the same code.

What are the pros and cons then?

- If we store it in the data, this remove the need for the
  entry-accessors class, and we don't need this empty last-access slot
  when importing the old flat history.

- If we store it in the htree, then we don't need to deal with any of
  the last-access code in Nyxt's source.

Hmm...

>>     * [ ]  There is a general slow down of many actions (e.g. buffer switching) when the history grows big.  My guess is that this is because we serialize the history much more often now, and serialization is also more costly because of the many nested structures.
>
> Is addressing separation of data access and data modification enough to fix that? If so, I can push the `with-data-(access|lookup)` (still looking for better names) distinguishing code to `master` in a couple hours and patch this branch accordingly afterwards.

It could indeed, because it seems that we are re-serializing even when
there is no change.

I just need to make sure that this is the bottleneck :)

Anyways, this is a welcome change, so if you have time to send a pull
request that'd be awesome!

aartaka:
> > > * [ ] There is a general slow down of many actions (e.g. buffer switching) when the history grows big. My guess is that this is because we serialize the history much more often now, and serialization is also more costly because of the many nested structures. 

> > Is addressing separation of data access and data modification enough to fix that? If so, I can push the `with-data-(access|lookup)` (still looking for better names) distinguishing code to `master` in a couple hours and patch this branch accordingly afterwards.

> It could indeed, because it seems that we are re-serializing even when there is no change. I just need to make sure that this is the bottleneck :) Anyways, this is a welcome change, so if you have time to send a pull request that'd be awesome!

Ooooops, I've already pushed the new macro to `master` and `history-tree->ght` (see 4378d3c3f0cfca138cbad3ae07c4d40c52047685) and tested the race conditions. Seems to compile/work fine.

Another shower thought: we can make an `:around` method for `(re)store` that will check the last modification time of the data file. This will help us to avoid unnecessary (de)serialization of the same data.

aartaka:
CI is passing :)

Ambrevar:

> Another shower thought: we can make an `:around` method for `(re)store` that will check the last modification time of the data file. This will help us to avoid unnecessary (de)serialization of the same data.

If I'm not mistaken, we only deserialize once, when the data is first
accessed.

Serialization should happen every time `with-data-access` is called.
Can we do better than that?

Ambrevar:
@aartaka Just realized that you rebased this branch against master.
Please _never_ do this on a public branch (at least not without notifying everyone beforehand).

Rebasing and force-pushing can result in lost commits in history if I had pushed right before you did.
(It also gives me more work since I need to rebase my branch against this new branch.)

Thanks!

Ambrevar:
I've just implemented 2 things that @jmercouris suggested:

- Merge entry-accessors into entry.  This makes everything much simpler!   Great success!
- Merge Nyxt's `history-entry` into `htree:entry`.  This one has the benefit of reducing the size of the serialized data by about 50%!.  The drawback is that it is a significant change and there are still a few bugs. 

I see 2 options:

- Either I revert commits 494ca162e1f0b870deaeb83ab2df4855369f7f96 and b3e6419dbf19980c7ac19d542fc7a4175c1b5d3d and stick to the old model where `history-tree` is a value of `entry`.

- Or I change the model of `htree` by removing `value` in `entry`, thus forcing the library users to subclass `entry`.

Thoughts?

jmercouris:
I suggest forcing users to subclass entry, that is actually what i had in mind :-D. To enforce the API, for the slot you can have an initform of an error, so if the user doesn't subclass it, they can't use it.

jmercouris:
Perhaps this could be used: https://github.com/phoe/protest/blob/master/src/base/base.lisp

aartaka:
> @aartaka Just realized that you rebased this branch against master.
> Please _never_ do this on a public branch (at least not without notifying everyone beforehand).
> 
> Rebasing and force-pushing can result in commit loss if I had pushed right before you did.
> (It also gives me more work since I need to rebase my branch against this new branch.)
> 
> Thanks!

Oh, that was careless of me :cold_sweat:

Ambrevar:
> Oh, that was careless of me :cold_sweat: 

No worries! :)

Ambrevar:

> I suggest forcing users to subclass entry, that is actually what i had in mind :-D. To enforce the API, for the slot you can have an initform of an error, so if the user doesn't subclass it, they can't use it.

Agreed this could be a nice change, but this is too much a change and
this would be dragging this change further to another week or so.

Another problem is that the history-entry class would have both
nyxt::last-acess and htre::last-access, which is confusing :)

I suggest we postpone this idea.

I think it's even better if we implement it after the next pre-release.
Then we can make the history-entry class without a last-access slot.
Users from the next pre-release will see their history.lisp file updated
seamlessly.

jmercouris:
OK! Sounds like a plan to me! :-)

Ambrevar:
I confirm that the performance bottleneck is in the serialization calls for the
history.

I've tried speeding it up by removing the beautification of quri:uri and
timestamps.  Turns out it twice as slow without beautification! :p
Well, good for us, the file will be more readable.

I've just pushed a better fix: this serializes the history in a separate thread
and skips many unnecessary serialization ops.

cf62bd6eda079212765749007d13d3e417356049

The implementation surprised me: it's very short and actually did not involve
any change to `history.lisp`!  I just leverages CLOS and our data-storage
mini-library.  Another tell-tale sign that this library is well designed :)

--------------------------------------------------------------------------------

Pitfall: we are not locking access to the history data when serializing.  Should
we do it, it would dead-lock the whole browser for the same reason @aartaka
explained previously.

This means that we risk to save a potentially corrupt history.  In particular,
I'm not sure how s-serialization:sexp behaves when data structures like hash
tables are modifier while its serializing.  What I can gather from the code is
that it will serialize data that should not be there, and hopefully it won't go
out of bounds, but I'm not sure about this last point.

The main offender for the dead-lock here seems to be `nyxt:on-signal-notify-uri`
in web-mode.
One suggestion: run `history-add` and `store` in a separate thread.  Thoughts?

Ambrevar:
I've just pushed commands to delete history entries.

We should now be feature complete.
(For the remaining features, like prompting which history to restore or
restoring the current buffer, we will do it in another pull request.)

Please test as much as you can.

In the mean time, I'll go ahead with the renamings and other refactoring we
talked about.

aartaka:
> Pitfall: we are not locking access to the history data when serializing. Should
> we do it, it would dead-lock the whole browser for the same reason @aartaka
> explained previously.
> 
> This means that we risk to save a potentially corrupt history. In particular,
> I'm not sure how s-serialization:sexp behaves when data structures like hash
> tables are modifier while its serializing. What I can gather from the code is
> that it will serialize data that should not be there, and hopefully it won't go
> out of bounds, but I'm not sure about this last point.
 
I'm not sure that I totally get how `:around` methods work, but `store` called inside `with-data-access` is already covered by `bt:with-recursive-lock-held`. Is that not enough?

Ambrevar:
`store` is sometimes called outside `with-data-access`, like in `nyxt:on-signal-notify-uri`.

aartaka:
> `store` is sometimes called outside `with-data-access`, like in `nyxt:on-signal-notify-uri`.

Oh, that's my bad -- a relic from the times when `history-add` didn't use `with-data-access`. I forgot to remove it :(

In any way, I still think that `store` should only do storage without locking -- to separate concerns. `with-data-access` is to ensure locking there. This gives us a better control of effects (i.e. one can intentionally `store` without locking) and makes lock management simpler -- it's one place (`with-data-access`) to control locks in, instead of two.

This also maps well on the structure of the `data-storage` library:
- Low-level: files, `with-data-file`, `with(-maybe)-gpg-file`.
- Intermediate level: raw data/(de)serialization, `(re)store`, `get-data`.
- High-level: locking and convenience, `with-data-access`, `with-data-(lookup|unsafe)`.

Ambrevar:
Agreed!

> Oh, that's my bad -- a relic from the times when `history-add` didn't use `with-data-access`. I forgot to remove it :(

So is it safe to remove this line then?

Ambrevar:
I've just fixed one important bug.

Rest seems to work pretty smoothly for me so far.
Any bug to report @aartaka @jmercouris?

Ambrevar:
Bug fix is in 4869e273b04026f144e166ef7229636911ce7965.

aartaka:
> @aartaka Did you think about this issue when you worked on the GHT?

No, I didn't. I've made a choice of storing all the history back then, so it wasn't much of a problem. It's puzzling to solve with current version :(

Ambrevar:
For now I suggest to consider that dead buffers are equivalent to
history-reset buffers...

We can reconstruct the tree some other time.

Ambrevar:
Buffer reopening is fixed with 7a2763d123d48939204320d6e3a51f8b689d4fb2.

Ambrevar:
@aartaka: About `history-forwards-prompting-p`, do we still have a use case for it?
In my opinion:

- `history-backward` and `history-forward` should never prompt the user, they are meant to be "fast" commands.
- `history-forwards-direct-children` asks which of the direct children to go forward to. `conservative-history-movement-p` is used to know whether to list only the owned children or not.
- `history-forwards-maybe-query` is a combination of the above 2: if current node has multiple children, query which one to go to.  `conservative-history-movement-p` is also used here.

Doesn't it cover all use cases?  (I'm not listing the commands the go beyond the direct children.)
- `history-forwards-maybe-query` is the

aartaka:
Yes, seems okay to me. If one (e.g., me) wants to have prompting by default, they can just rebind the keys to `history-forwards-maybe query` :)

Ambrevar:
OK, I've just removed it.

I've fixed the last URL duplication issue I could experience.

It's ready to merge for me :D

jmercouris:
Works great for me :-)

Only one error, on restart:

```
  #<STANDARD-GENERIC-FUNCTION S-SERIALIZATION::DESERIALIZE-CLASS (1)>
when called with arguments
  (HISTORY-TREE:OWNER
   ((HISTORY-TREE:ORIGIN) (HISTORY-TREE:CREATOR-ID)
    (HISTORY-TREE::CREATOR-NODE) (HISTORY-TREE:CURRENT)
    (HISTORY-TREE:NODES))
   #<HASH-TABLE :TEST EQ :COUNT 3 {1001A15073}>).
   [Condition of type SB-PCL::NO-APPLICABLE-METHOD-ERROR]
See also:
  Common Lisp Hyperspec, 7.6.6 [:section]

Restarts:
 0: [RETRY] Retry calling the generic function.
 1: [RETURN-FROM-CALLBACK] GTK::RETURN-FROM-CALLBACK
 2: [ABORT] abort thread (#<THREAD "cl-cffi-gtk main thread" RUNNING {1006D14E23}>)

Backtrace:
  0: ((:METHOD NO-APPLICABLE-METHOD (T)) #<STANDARD-GENERIC-FUNCTION S-SERIALIZATION::DESERIALIZE-CLASS (1)> HISTORY-TREE:OWNER ((HISTORY-TREE:ORIGIN) (HISTORY-TREE:CREATOR-ID) (HISTORY-TREE::CREATOR-NODE)..
  1: (SB-PCL::CALL-NO-APPLICABLE-METHOD #<STANDARD-GENERIC-FUNCTION S-SERIALIZATION::DESERIALIZE-CLASS (1)> (HISTORY-TREE:OWNER ((HISTORY-TREE:ORIGIN) (HISTORY-TREE:CREATOR-ID) (HISTORY-TREE::CREATOR-NODE)..
  2: (HISTORY-DESERIALIZE-SEXP-INTERNAL (:OBJECT 4 :CLASS HISTORY-TREE:OWNER :SLOTS ((HISTORY-TREE:ORIGIN) (HISTORY-TREE:CREATOR-ID) (HISTORY-TREE::CREATOR-NODE) (HISTORY-TREE:CURRENT) (HISTORY-TREE:NODES)..
  3: (HISTORY-DESERIALIZE-SEXP-INTERNAL (:HASH-TABLE 3 :TEST EQUALP :SIZE 14 ...) #<HASH-TABLE :TEST EQ :COUNT 3 {1001A15073}>)
  4: (HISTORY-DESERIALIZE-SEXP-INTERNAL (:OBJECT 2 :CLASS HISTORY-TREE:HISTORY-TREE :SLOTS ((HISTORY-TREE:OWNERS :HASH-TABLE 3 :TEST EQUALP :SIZE ...) (HISTORY-TREE:CURRENT-OWNER-ID . "3") (HISTORY-TREE:EN..
  5: ((FLET SB-IMPL::F :IN MAP-INTO) (:OBJECT 2 :CLASS HISTORY-TREE:HISTORY-TREE :SLOTS ((HISTORY-TREE:OWNERS :HASH-TABLE 3 :TEST EQUALP :SIZE ...) (HISTORY-TREE:CURRENT-OWNER-ID . "3") (HISTORY-TREE:ENTRI..
  6: (SB-KERNEL:%MAP-FOR-EFFECT-ARITY-1 #<CLOSURE (FLET SB-IMPL::F :IN MAP-INTO) {7FA432B4DBEB}> ("2-pre-release-5-368-gaf546de0" (:OBJECT 2 :CLASS HISTORY-TREE:HISTORY-TREE :SLOTS (# # # # # #))))
  7: (SB-KERNEL:%MAP NIL #<CLOSURE (FLET SB-IMPL::F :IN MAP-INTO) {7FA432B4DBEB}> ("2-pre-release-5-368-gaf546de0" (:OBJECT 2 :CLASS HISTORY-TREE:HISTORY-TREE :SLOTS (# # # # # #)))) [more]
  8: (MAP-INTO ("2-pre-release-5-368-gaf546de0" NIL) #<CLOSURE (LAMBDA (X) :IN HISTORY-DESERIALIZE-SEXP-INTERNAL) {1001A1A8BB}> ("2-pre-release-5-368-gaf546de0" (:OBJECT 2 :CLASS HISTORY-TREE:HISTORY-TREE ..
  9: ((LABELS %RESTORE :IN RESTORE))
 10: ((FLET "WITHOUT-INTERRUPTS-BODY-29" :IN SB-THREAD::CALL-WITH-RECURSIVE-LOCK))
 11: (SB-THREAD::CALL-WITH-RECURSIVE-LOCK #<CLOSURE (FLET SB-THREAD::WITH-RECURSIVE-LOCK-THUNK :IN %GET-USER-DATA) {7FA432B4DF6B}> #<SB-THREAD:MUTEX "Anonymous recursive lock" owner: #<SB-THREAD:THREAD "cl..
 12: (%GET-USER-DATA #<DEFAULT-DATA-PROFILE {1006B9ED73}> #<HISTORY-DATA-PATH {10065FFA43}> #<HASH-TABLE :TEST EQUAL :COUNT 1 {10065FDB93}>)
 13: (BUFFER-MAKE #<USER-BROWSER {10054844F3}> :DATA-PROFILE NIL :TITLE "*Help*" :DEFAULT-MODES NIL :DEAD-BUFFER NIL :INTERNAL-BUFFER-P T :PARENT-BUFFER NIL)
 14: (MAKE-INTERNAL-BUFFER :TITLE "*Help*" :MODES NIL)
 15: (HELP)
 16: ((:METHOD FINALIZE (BROWSER T T)) #<USER-BROWSER {10054844F3}> NIL @2021-02-02T23:29:43.705711+01:00) [fast-method]
 17: ((LAMBDA (GTK::DATA) :IN "/home/jmercouris/Source/Lisp/cl-cffi-gtk/gtk/gtk.init.lisp") #.(SB-SYS:INT-SAP #X00000002))
 18: ((LAMBDA (SB-ALIEN::ARGS-POINTER SB-ALIEN::RESULT-POINTER FUNCTION) :IN "/home/jmercouris/Work/Atlas/Nyxt/quicklisp-client/dists/quicklisp/software/cl+ssl-20200610-git/src/bio.lisp") #<unavailable arg..
 19: ("foreign function: funcall_alien_callback")
 --more--
```

Ambrevar:
Have you deleted the old history/default.lisp?
If no, do it, then try again.

Otherwise, can you provide a precise recipe?

jmercouris:
I delete the old history, start Nyxt, visit any sites, then close Nyxt. I then restart Nyxt and have this error.

Ambrevar:
Can you provide a more precise recipe?

- Which site?
- The exact command that you run.
- The content of your history/default.lisp.

Ambrevar:
Actually, maybe the problem is with your version of
cl-prevalence.

I'm on https://github.com/40ants/cl-prevalence/commit/5a76be036092ed6c18cb695a9e03bce87e21b840.

jmercouris:
Very possibly, I'll update my QL dist and try again!

jmercouris:
That was it! Sorry for the false alarm!!!

jmercouris:
Also take your break :-D!

aartaka:
Yay! Thanks for this great design, Pierre! :tada:

jmercouris:
Indeed, Merci Beaucoup Monsieur :-)

Ambrevar:
Hooray! :)

Issue #1111, opened 2021-01-28
FreeBSD status report


nayuki-d:
I've successfully compiled Nyxt/GTK (master/main branch) with SBCL 2.1.0 on FreeBSD 13-Stable natively and on Linux emulation. All dependencies (except from Cluffer) grabbed from Quicklisp.
You can reproduce the lisp compilation process of the docs just fine.
For make compile instructions, you'll need to use "gmake" instead of "make" and edit Makefile (I've replaced Linux for FreeBSD in line 101).
Sly integration also works.
![screenshot](https://user-images.githubusercontent.com/70187554/106088366-bd220000-611d-11eb-88eb-def020c22abf.png)

Ambrevar:
Excellent!
Thank you so much for your report!
I've updated the Makefile and the documentation to reflect this:
e0846f7900ff4c9a17f7244cad70352122e9ca01

@wahjava Shall we close #800 then?

jmercouris:
This is great! Thank you for figuring this out! Maybe now I will uninstall NixOS and go back to BSD :-D

wahjava:
Thanks for working on it. This works great now for me too on `12.2-RELEASE`

PR #1113, opened 2021-01-28
FreeBSD related changes

nayuki-d:
#1111


Ambrevar:
Thanks, very useful!

Don't you also need gsettings-desktop-schemas?

nayuki-d:
gsettings-desktop-schemas is a build and runtime dependency of net/glib-networking. Both pkg(8) and ports installation will install it automatically.

Ambrevar:
Thanks a lot!

jmercouris:
+1 thank you!

PR #1116, opened 2021-02-02
renderer-gtk*.lisp: re-enable native renderer download

jmercouris:
It seems to be enough to just re-enable it in the context menu. Then it bypasses the download manager and allows the engine to handle it.

That said, perhaps I should write a interface so that you can at least monitor a download initiated in this manner, or perhaps specify a download path... here is the reference: https://webkitgtk.org/reference/webkit2gtk/stable/WebKitDownload.html


Ambrevar:
Indeed, we need to capture the download progress.  It's not hard, just a few more bindings.

That said, supporting cookies in our download manager might be even easier to do, and it will be more useful in the future!
(Dexador accepts a cookie argument.)

jmercouris:
I see as adding cookie support to dexador not mutually exclusive :-)

jmercouris:
I will investigate both!

Ambrevar:
Why did you close this?

Also I see it's merged, I would have like to review and test this.
Any reason you merged it so soon?

I don't understand the new feature: now we have both a CL download manager _and_ downloads via right-click?  The right click download, if I'm not mistaken, won't download to the directory configured by the user, nor will it report any progress (which is very confusing).  Am I missing something?

jmercouris:
I made a new branch where I've created download displays, will add a new pull request soon!

jmercouris:
More details following soon :-)

jmercouris:
To clarify, this branch was only about enabling the context menu

Ambrevar:
Ah OK, thanks!

PR #1120, opened 2021-02-03
Add default methods to define-ffi-generic

aartaka:
This adds a new piece of syntax to the `define-ffi-generic` -- it now accepts all the `defgeneric` options and method definitions. It's applied to the `ffi-buffer-set-zoom-level` there, both to showcase it and close the issue with the absent default implementation of that.

# Motivation

This came to me when we were discussing the default implementation of zooming with JavaScript for renderers that might have no zooming support. We need to have a default method, and the most obvious place to define it in is the generic declaration. It does directly correlate with `defgeneric` that we depend on and take inspiration from :)

# How Has This Been Tested

I've macroexpanded it several times to make sure everything's right :)

EDIT: Update with the support of options and all the `defgeneric` things.


jmercouris:
I like it. I like your proposed syntax even more :-) I may take a crack at it later if you can't figure it out!

aartaka:
One way is to use `&body options` and parse every one of these, like `defgeneric` does. This will also allow us to pass options to `defgeneric` through `define-ffi-generic`. Isn't it overpowered for the purpose of FFI generics, though?

aartaka:
Done: it's definitely owerpowered, but now `define-ffi-generic` accepts all the options of `defgeneric`, so we can do `(:method (arg1 arg2) body ...))` in it, and all the magical things `defgeneric` supports :)

jmercouris:
Ah, it is just a thin wrapper at this point :-D

jmercouris:
Thank you Artyom!

PR #1121, opened 2021-02-03
Rename with-data-lookup to with-data-unsafe.

aartaka:
This is mainly to discuss the renaming of `with-data-lookup` to `with-data-unsafe` (otherwise I'd probably push this right to `master`):
- Is the new name clear enough? Maybe `with-data-unsafe-lookup`?
- Should it have a `default` value? Is there any use to it, actually?
- I could use some documentation proofreading :)

Thank you!


source/data-storage.lisp
```diff
@@ -317,16 +317,17 @@ For a faster and modification-unsafe version, see `with-data-lookup'."
...
+(defmacro with-data-unsafe ((data-var data-path &key default) &body body)
   "Bind the data to DATA-VAR for a fast non-modifying lookup.
 Bind the DATA-VAR to the value of the data from DATA-PATH to reuse it.
 In case there's no data, bind DATA-VAR to DEFAULT.
 
 \"non-modifying\" means that you data you set DATA-VAR to will be
 neither persisted to disk nor saved in memory.
 
-For modification-safe macro, see `with-data-access'."
+BEVARE: this is not a thread-safe macro, so data may be altered while
```
Ambrevar, line 329: Typo: Beware.
Don't spell it in uppercase, this is reserved for the parameters.

Ambrevar:
I'm find with the `with-data-unsafe` name.  @jmercouris Thoughts?

I don't know if the `default` makes sense.  You'd essentially be looking up data in an (usually) empty data structure.  But if we remove the `default` option, then this macro is an empty wrapper around `get-data`.
So maybe it's better to rename `get-data` instead.  Thoughts?

jmercouris:
I think that having two different symbols that do the same thing is a little bit silly. I would rename `get-data`. If I later wanted to make a (by default) safe version of get data, at that point I would introduce a new symbol.

aartaka:
What I recently came up with is that we can have additional keyword arguments in `with-data-(access|unsafe)`. For example:
- `key` to apply some function to the data right away.
- `if-null` to decide what to do with empty data.

I've added `key` argument to `with-data-unsafe here, to show that it saves some lines and can be a good way to show one's intent. E.g., in auto-mode we want to get previous URL user visited, and not the whole history. Having it in `with-data-unsafe` instead of additional `let` clause is more straightforward, I think.

However, I may be mistaken about readability of that. How does it look to you?

Ambrevar:
Looks good to me!  Indeed, the `:key` argument justifies the `with-data-unsafe` macro now :)

I don't see `if-null` in the definition, am I missing something?
Anyways, I'm not sure `if-null` would be very useful in practice.

@jmercouris ?

aartaka:
> I don't see `if-null` in the definition, am I missing something?
> Anyways, I'm not sure `if-null` would be very useful in practice.

It's not there yet, I want to add it to both `with-data-access` and `with-data-lookup` once we merge this. One use-case would be to do nothing if there's no data, like we already do with `when` in `history-suggestion-filter`. This can be abstracted into the keyword (in part to refer to `with-open-file` behavior).

jmercouris:
I don't have a valuable insight to offer I'm afraid!

aartaka:
Pierre, you were right, `if-null` is not useful at all. Same for `key` in `with-data-access` -- we use `with-data-access` to modify data, and not to retrieve it, as in `with-data-unsafe`.

I've rebased this branch over master and resolved the conflicts. Safe to merge, I believe.

Ambrevar:
Thank you!

PR #1123, opened 2021-02-03
Download native

jmercouris:
Support for native download!

Uses new user-interface library to download percentage complete!

Unfortunately creates a new buffer per download I think... not sure how to get around that without significant work. I would also need to make a list view for my user-interface.

### DIFF
--- a/libraries/user-interface/user-interface.lisp
+++ b/libraries/user-interface/user-interface.lisp
@@ -46,3 +46,21 @@
 
 (defmethod text ((paragraph paragraph))
   (slot-value paragraph 'text))
+
+(defclass progress-bar (ui-element)
+  ((percentage :initform 0 :initarg :percentage
+               :documentation "The percentage the progress bar is
+filled up, use a number between 0 and 100.")))
+
+(defmethod percentage ((progress-bar progress-bar))
+  (slot-value progress-bar 'percentage))
+
+(defmethod object-expression ((progress-bar progress-bar))
+  `(:div :class "progress-bar-base"
+         (:div :class "progress-bar-fill"
+               :id ,(id progress-bar))))
+
+(defmethod (setf percentage) (percentage (progress-bar progress-bar))
+  (setf (slot-value progress-bar 'percentage) (format nil "~D%" percentage))
+  (when (slot-boundp progress-bar 'buffer)
+    (update progress-bar)))
--- a/libraries/user-interface/user-interface.lisp
+++ b/libraries/user-interface/user-interface.lisp
@@ -27,11 +27,27 @@
   (:documentation "Propagate changes to the buffer."))
 
 (defclass button (ui-element)
-  ((text :initform "" :accessor text :initarg :text)
-   (url :initform "" :accessor url :initarg :url)))
+  ((text :initform "" :initarg :text)
+   (url :initform ""  :initarg :url)))
+
+(defmethod (setf text) (text (button button))
+  (setf (slot-value button 'text) text)
+  (when (slot-boundp button 'buffer)
+    (update button)))
+
+(defmethod (setf url) (url (button button))
+  (setf (slot-value button 'url) url)
+  (when (slot-boundp button 'buffer)
+    (update button)))
+
+(defmethod text ((button button))
+  (slot-value button 'text))
+
+(defmethod url ((button button))
+  (slot-value button 'url))
 
 (defmethod object-expression ((button button))
-  `(:a :class "button" :href ,(url button) ,(text button)))
+  `(:a :id ,(id button) :class "button" :href ,(url button) ,(text button)))
 
 (defclass paragraph (ui-element)
   ((text :initform "" :initarg :text)))
--- a/libraries/user-interface/user-interface.lisp
+++ b/libraries/user-interface/user-interface.lisp
@@ -61,6 +61,6 @@ filled up, use a number between 0 and 100.")))
                :id ,(id progress-bar))))
 
 (defmethod (setf percentage) (percentage (progress-bar progress-bar))
-  (setf (slot-value progress-bar 'percentage) (format nil "~D%" percentage))
+  (setf (slot-value progress-bar 'percentage) (format nil "~D%" (round percentage)))
   (when (slot-boundp progress-bar 'buffer)
     (update progress-bar)))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -267,36 +267,33 @@ current buffer."
   (hooks:run-hook (before-download-hook *browser*) url) ; TODO: Set URL to download-hook result?
   (match (download-engine buffer)
     (:lisp
-     (when (eq proxy-address :auto)
-       (setf proxy-address (proxy-address buffer :downloads-only t)))
-     (let* ((path (download-path buffer))
-            (download-dir (expand-path path)))
-       (declare (type (or quri:uri null) proxy-address))
-       (when download-dir
-         (let* ((download nil))
-           (flet ((unsafe-download ()
-                    (with-data-access (downloads path)
-                      (setf download (download-manager:resolve
-                                      url
-                                      :directory download-dir
-                                      :cookies cookies
-                                      :proxy proxy-address))
-                      (push download downloads)
-                      ;; Add a watcher / renderer for monitoring download
-                      (let ((download-render (make-instance 'download)))
-                        (setf (uri download-render) (object-string url))
-                        (push download-render (downloads *browser*))
-                        (bt:make-thread
-                         (lambda ()
-                           (download-watch download-render download))))
-                      download)))
-             (if *keep-alive*
-                 (unsafe-download)
-                 (handler-case
-                     (unsafe-download)
-                   (error (c)
-                     (echo-warning "Download error: ~a" c)
-                     nil))))))))
+     (alex:when-let* ((path (download-path buffer))
+                      (download-dir (expand-path path)))
+       (when (eq proxy-address :auto)
+         (setf proxy-address (proxy-address buffer :downloads-only t)))
+       (let* ((download nil))
+         (flet ((unsafe-download ()
+                  (with-data-access (downloads path)
+                    (setf download 
+                          (download-manager:resolve url
+                                                    :directory download-dir
+                                                    :cookies cookies
+                                                    :proxy proxy-address))
+                    (push download downloads)
+                    ;; Add a watcher / renderer for monitoring download
+                    (let ((download-render (make-instance 'download :uri (object-string url))))
+                      (push download-render (downloads *browser*))
+                      (bt:make-thread
+                       (lambda ()
+                         (download-watch download-render download))))
+                    download)))
+           (if *keep-alive*
+               (unsafe-download)
+               (handler-case
+                   (unsafe-download)
+                 (error (c)
+                   (echo-warning "Download error: ~a" c)
+                   nil)))))))
     (:renderer
      (ffi-buffer-download buffer (object-string url))))
   (list-downloads))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -241,20 +241,20 @@ editor executable."))
 
 (defun download-watch (download-render download-object)
   "Update the *Downloads* buffer.
-This function is meant to be run in the background."
+This function is meant to be run in the background. There is a
+potential thread starvation issue if one thread consumes all
+messages. If in practice this becomes a problem, we should poll on
+each thread until the completion percentage is 100 OR a timeout is
+reached (during which no new progress has been made)."
   (when download-manager:*notifications*
     (loop for d = (calispel:? download-manager:*notifications*)
           while d
           when (download-manager:finished-p d)
             do (hooks:run-hook (after-download-hook *browser*))
           do (sleep 0.1) ; avoid excessive polling
-             (let ((completion-percentage
-                     (* 100 (/ (download-manager:bytes-fetched download-object)
-                               (max 1 (download-manager:bytes-total download-object))))))
-               (setf (user-interface:percentage (progress download-render))
-                     completion-percentage)
-               (setf (user-interface:text (paragraph download-render))
-                     (format nil "Completion: ~,2f%" completion-percentage))))))
+             (setf (completion-percentage download-render)
+                   (* 100 (/ (download-manager:bytes-fetched download-object)
+                             (max 1 (download-manager:bytes-total download-object))))))))
 
 ;; TODO: To download any URL at any moment and not just in resource-query, we
 ;; need to query the cookies for URL.  Thus we need to add an IPC endpoint to
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -282,6 +282,8 @@ current buffer."
                     (push download downloads)
                     ;; Add a watcher / renderer for monitoring download
                     (let ((download-render (make-instance 'download :uri (object-string url))))
+                      (setf (destination-path download-render)
+                            (download-manager:filename download))
                       (push download-render (downloads *browser*))
                       (bt:make-thread
                        (lambda ()
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -264,40 +264,39 @@ This function is meant to be run in the background."
 ;; need to query the cookies for URL.  Thus we need to add an IPC endpoint to
 ;; query cookies.
 (export-always 'download)
-(defun download (url &key buffer cookies (proxy-address :auto))
+(defmethod download ((buffer buffer) url &key cookies (proxy-address :auto))
   "Download URL.
 When PROXY-ADDRESS is :AUTO (the default), the proxy address is guessed from the
 current buffer."
-  (let ((buffer (or buffer (current-buffer))))
-    (hooks:run-hook (before-download-hook *browser*) url) ; TODO: Set URL to download-hook result?
-    (if (download-with-lisp-engine-p buffer)
-        (progn
-          (when (eq proxy-address :auto)
-            (setf proxy-address (proxy-address buffer :downloads-only t)))
-          (let* ((path (download-path buffer))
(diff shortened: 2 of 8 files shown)


jmercouris:
Well, now we have a single buffer, all is well again :-)

jmercouris:
I found a bit of a shortcut, I would still like to use my user-interface library with a list and dynamically append. That will however wait, I am still deciding on how I want to do some things.

jmercouris:
Ready for review!

jmercouris:
To test the native downloader, you can use the following:

```
(define-configuration web-buffer
  ((download-with-lisp-engine-p nil)))
```

jmercouris:
Tomorrow I plan on working on the following enhancements:

1. respect user download directory
2. show progress bar
3. and hopefully, a cancel button

with those done, i believe it will be ready for 2.0.0!

source/browser.lisp
```diff
@@ -263,43 +263,40 @@ This function is meant to be run in the background."
 ;; TODO: To download any URL at any moment and not just in resource-query, we
 ;; need to query the cookies for URL.  Thus we need to add an IPC endpoint to
 ;; query cookies.
-(declaim (ftype (function (quri:uri &key
-                                    (:cookies (or string null))
-                                    (:proxy-address t))
-                          (values (or null download-manager:download) &rest t))
-                download))
```
Ambrevar, line 270: Any reason for removing the type?  It seems to be valuable here.

source/browser.lisp
```diff
@@ -263,43 +263,40 @@ This function is meant to be run in the background."
...
+                         download)))
+                (if *keep-alive*
+                    (unsafe-download)
+                    (handler-case
+                        (unsafe-download)
+                      (error (c)
+                        (echo-warning "Download error: ~a" c)
+                        nil)))))))
+        (unless (find-buffer 'download-mode)
+          (list-downloads)))
```
Ambrevar, line 298: Why moving this from `request-resource` to here?
This means that when using the FFI downloads it won't display the download buffer?

source/buffer.lisp
```diff
@@ -197,6 +197,9 @@ The handlers take the buffer as argument.")
                   :documentation "Path of directory where downloads will be
 stored.  Nil means use system default.
 Downloads are kept in browser's `user-data', keyed by the expanded `download-path'.")
+   (download-with-lisp-engine-p :initform t
+                                :documentation "Use the Lisp download
+engine or the renderer based download engine.")
```
Ambrevar, line 202: Maybe more idiomatic: "When true, use the Lisp download engine instead of the renderer-based download engine."

When possible, I prefer to use `nil` as default value, since non-zero default values are prone to trip programmers.  So here I would call the variable `download-with-renderer-p` and default it to `nil`.  Thoughts?

Or... Actually, we could be a bit more future proof and instead call it `download-engine` and specify which engine we want.  We might very well have, say, a cURL-based engine at some point! :)

source/browser.lisp
```diff
@@ -263,43 +263,40 @@ This function is meant to be run in the background."
...
-(declaim (ftype (function (quri:uri &key
-                                    (:cookies (or string null))
-                                    (:proxy-address t))
-                          (values (or null download-manager:download) &rest t))
-                download))
 (export-always 'download)
-(defun download (url &key
-                       cookies
-                       (proxy-address :auto))
+(defmethod download ((buffer buffer) url &key cookies (proxy-address :auto))
```
Ambrevar, line 267: Why turn it into a method?

source/renderer-gtk.lisp
```diff
@@ -32,7 +32,17 @@ want to change the behaviour of modifiers, for instance swap 'control' and
...
+   (downloads
+    :documentation "List of downloads. Used for rendering by download manager.")
+   (download-buffer-style
+    (cl-css:css
+     '((".download-url"
+        :overflow "auto"
+        :white-space "nowrap")
+       (".download-url a"
+        :color "black")))
+    :documentation "Style applied to the download buffer."))
```
Ambrevar, line 45: Could we move this to `download-mode` and call it `style` instead?  This would pave the way for our theming framwork! :)

source/renderer-gtk.lisp
```diff
@@ -853,6 +863,34 @@ requested a reload."
...
+              for uri = (first download)
+              for paragraph = (second download)
+              do (user-interface:connect paragraph buffer)
+              collect
+                 (markup:markup
+                  (:div
+                   (:hr)
+                   (:p :class "download-url" (:b "URL: ") (:a :href uri uri))
+                   (:span (markup:raw (user-interface::object-string paragraph)))))))))))
+
```
Ambrevar, line 893: What about using the same download buffer that we are using in Lisp?  Would be nice to factor both the code and the user experience.
Maybe you just wanted to first get something to work and do the interface factoring in the next iteration :)

Ambrevar:
Looking great, thanks for working on this!
I'll test later, let me know when you are done with your to-do list! :)

source/browser.lisp
jmercouris: The download buffer for FFI is different :-)!

source/buffer.lisp
jmercouris: A good point, I will use keyword arguments!

source/browser.lisp
jmercouris: I want to require a buffer, maybe we'll specialize it for different kind of buffers. That is actually what I had in mind.

source/renderer-gtk.lisp
jmercouris: I don't believe so, the native download buffer is different from the engine based one.

source/browser.lisp
jmercouris: No real reason, just the compiler errors were frustrating me. I've never enjoyed writing type declarations :-D

source/renderer-gtk.lisp
Ambrevar: Ideally we would use the same buffers though, so that would do it, no?

source/browser.lisp
Ambrevar: Hmm, I can't imagine how different kind of downloads could be mapped to different buffer classes (i.e. web-buffer, internal-buffer, etc.).  Can you expand on the idea you had in mind?

source/browser.lisp
Ambrevar: Well, now that it's a defmethod, you must remove it.
With a function, the type declaration would just be `(:buffer buffer)` for a keyword argument.

source/browser.lisp
Ambrevar: Same as below, if it were the same buffer then we could factor this code, right?

jmercouris: IF, but they are not!

source/browser.lisp
jmercouris: For example, we have a Qt buffer and GTK buffer, and they are both set to use the native renderer download, they will use two different download functions.

source/renderer-gtk.lisp
jmercouris: I mostly did not want to touch the lisp download engine :-D

jmercouris:
Progress bars : done!

source/browser.lisp
Ambrevar: But `ffi-buffer-download` already dispatches over the type of buffer.  So isn't this a double layer of dispatching for the same thing?

source/renderer-gtk.lisp
Ambrevar: The engine is in libraries/download-manager, no need to touch that monster indeed.

That said, I think we should not have 2 interfaces for the same thing on the Nyxt side.  The download buffer needs a good revamping and now seems to be the perfect time to do it! :)

Thoughts?

source/browser.lisp
jmercouris: OK, then  I am fine with that. I prefer methods anyways.

source/browser.lisp
jmercouris: It does, only because I was indecisive. I really would have liked to make a download that specializes on `gtk-buffer`, but then I was thinking I would have to check the 'use-native-renderer' slot within `gtk-buffer` as well to decide what to do.

source/renderer-gtk.lisp
jmercouris: IF we convert the old engine to use the widgets, it should be possible, I will look into it.

source/renderer-gtk.lisp
jmercouris: You are right, and it kills me deep down inside :'(

Ay yai yai!

jmercouris:
I'm afraid that setting the destination is out of the cards:

I tried to first set the property of the Gobject the WebkitDownload, and it complained that property wasn't writable. So I then used the following CFFI definition:

```
(defcfun "webkit_download_set_destination" :void
  (webkit-download (g-object webkit-download))
  (uri :string))
(export 'webkit-download-set-destination)
```

and that worked, but then the download wouldn't start. Note that the documentation says the following:

> Sets the URI to which the downloaded file will be written. This method should be called before the download transfer starts or it will not have any effect on the ongoing download operation

Not sure how to ENSURE that it gets called before the download starts. I'm not sure because webkit:webkit-web-view-download-uri returns a WebkitDownload object that immediately starts downloading.

jmercouris:
If there is another way to make WebKitDownload objects pointing to a given resource, I don't know it.

jmercouris:
Reference: https://webkitgtk.org/reference/webkit2gtk/2.4.2/WebKitWebView.html#webkit-web-view-download-uri
https://webkitgtk.org/reference/webkit2gtk/2.4.2/WebKitDownload.html#

jmercouris:
Luckily, the path can at least still be customized: https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#G-USER-DIRECTORY-DOWNLOAD:CAPS

source/renderer-gtk.lisp
Ambrevar: That'd be great, thanks!

jmercouris:
I've also connected to the decide-destination signal, again, with no luck.

jmercouris:
Ask and ye shall receive. A unified interface! The only thing I'm not happy with is the download method... I think it is too convoluted. Too hard to understand what is occurring.

libraries/user-interface/user-interface.lisp
```diff
@@ -46,3 +46,21 @@
 
 (defmethod text ((paragraph paragraph))
   (slot-value paragraph 'text))
+
```
jmercouris, line 49: Yes, I could have used a native HTML progress, but this allows us to make a nicer looking one :-)

source/browser.lisp
```diff
@@ -241,65 +239,67 @@ editor executable."))
     (error "There is no current *browser*. Is Nyxt started?")))
 
 
-(defun download-watch ()
+(defun download-watch (download-render download-object)
   "Update the *Downloads* buffer.
 This function is meant to be run in the background."
-  ;; TODO: Add a (sleep ...)?  If we have many downloads, this loop could result
```
jmercouris, line 247: I've added the sleep below!

Ambrevar:
To be clear, is this ready for testing then?

jmercouris:
This is ready for testing, and merging.

jmercouris:
Note: I've removed some of the information from the old download manager. I believe we should make the information show a little bit configurable. I found the old download manager too busy. At any rate, I think that belongs in a subsequent pull request, and possibly (in my opinion), post 2.0.0.

jmercouris:
I also would still like to add the cancel button. I know how to do it in the native download, but not in the lisp engine. To avoid some of the complications with asymmetry in the interface, I decided to omit it for now.

jmercouris:
> Ask and ye shall receive. A unified interface! The only thing I'm not happy with is the download method... I think it is too convoluted. Too hard to understand what is occurring.

I've cleaned it a little bit more, I guess it is OK for now.

source/browser.lisp
```diff
@@ -241,65 +239,64 @@ editor executable."))
...
-  ;; in too high a frequency of refreshes.
   (when download-manager:*notifications*
```
jmercouris, line 245: I do see that there is a potential starvation issue if things are timed just right such that a particular download will never get updated (i.e. one thread will consume all the notifications). That said, I think this should not use calispel to update, it should just be on a fixed poll. Something like, each thread updating every .025 seconds, and then ceasing to update when the bytes-fetched is equal to the bytes-total (or after some timeout).

Ambrevar:
I've just tested it, it works really well!
I can switch download engine from the REPL and it just works, fantastic! :)

Great job!

Feedback:

- I tried downloading
  https://lispcookbook.github.io/cl-cookbook/common-lisp-cookbook.pdf
  with the `:lisp` engine.

  It downloaded properly, but the progress bar got stuck at 83%.
  This did not happen when I downloaded the `.epub`, may be random.
  The bug is probably that we are not notifying / receiving notification for the
  last chunk.
  
- I would not display "Completion", and just the percentage, on the same line as
  the bar.
  Even better: in the bar!  Is that possible?

- You've removed the download directory and the file name, but they are useful.
  The filename is useful when the URL does not end with the filename (happens sometimes).
  Could you restore them?

- The `:renderer` engine downloads to `$HOME`, as per John' comments.  @aartaka:
  Any idea how to fix that?

Ambrevar:
To recap what remains to be done (sorry, I'm a bit lost :p), before 2.0:

- Add a cancel button.
  For the `:lisp` engine, I think what we got to do is kill the download
  thread.  I don't know how to stop a dexador `get` call otherwise.
  
- Customize the download directory for the `:renderer` engine.

Ambrevar:
> Note: I've removed some of the information from the old download manager. I
> believe we should make the information show a little bit configurable. I found
> the old download manager too busy. At any rate, I think that belongs in a
> subsequent pull request, and possibly (in my opinion), post 2.0.0.

I think we need at least to tell _where_ files are downloaded, i.e. the download
directory.  Otherwise, it's confusing to newcomers.

About the filename: what about making it a tooltip of the URL?  The drawback is
that the filename won't be searchable in the buffer.

source/download-mode.lisp
```diff
@@ -81,7 +76,8 @@
       (fuzzy-match (input-buffer minibuffer) filenames))))
 
 (define-command download-open-file ()
-  "Open a downloaded file.
+  "Open a downloaded file. This command only works for downloads
+started by the :lisp download engine.
```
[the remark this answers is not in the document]
jmercouris: Doesn't really work if the downloads are from a heterogeneous set of engines.

Ambrevar:

>  (define-command download-open-file ()
> -  "Open a downloaded file.
> +  "Open a downloaded file. This command only works for downloads
> +started by the :lisp download engine.
>
> Doesn't really work if the downloads are from a heterogeneous set of engines.

Right.
Can't we just fix it to work with the :renderer engine?  Looks easy:
instead of storing the filename, both the :lisp and the :renderer
engines should store the full path to the downloaded path.

The `get-downloaded-filename` can be removed and the
`downloaded-files-suggestion-filter` becomes trivial.

Thoughts?

jmercouris:
> (define-command download-open-file () - "Open a downloaded file. + "Open a downloaded file. This command only works for downloads +started by the :lisp download engine. Doesn't really work if the downloads are from a heterogeneous set of engines.
> Right. Can't we just fix it to work with the :renderer engine? Looks easy: instead of storing the filename, both the :lisp and the :renderer engines should store the full path to the downloaded path. The `get-downloaded-filename` can be removed and the `downloaded-files-suggestion-filter` becomes trivial. Thoughts?

May work, I will think about it!

jmercouris:
>     * You've removed the download directory and the file name, but they are useful.
>       The filename is useful when the URL does not end with the filename (happens sometimes).
>       Could you restore them?

jmercouris:
> Great job!

Thank you!

>     * I would not display "Completion", and just the percentage, on the same line as
>       the bar.
>       Even better: in the bar!  Is that possible?

I don't like text within bars. The problem is that it always looks strange when the bar is going through the text. That is why it has been omitted from the progress bar implementation.

>     * You've removed the download directory and the file name, but they are useful.
>       The filename is useful when the URL does not end with the filename (happens sometimes).
>       Could you restore them?

I will try to restore them on a per-download basis, but rather with a button to open the file :-)

aartaka:
> * The `:renderer` engine downloads to `$HOME`, as per John' comments.  @aartaka:
>   Any idea how to fix that?

Is [`"decide-destination"` signal](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitDownload.html#WebKitDownload-decide-destination) what we're looking for? It is called before the download starts, if I've got it right.

jmercouris:
Yes, I've already tried to set the destination during decide destination, though maybe my CFFI signature or URI I passed was malformed. I can put the changes on a branch and you can see if you like.

aartaka:
Then I'm out of ideas here :dizzy_face:

jmercouris:
AAHHH! I just realized, URL WAS malformed! I forgot the file:/// prefix!!!

jmercouris:
The bounty of our editor hath delivered. We can now set the path, open the file and more. We are now ready to merge! I can add the cancel buttons later.

jmercouris:
What's that you say? You wish to be able to cancel a download? ASK NO MORE! It is now possible!

jmercouris:
Download open file is now functional, and with that, everything is addressed :-)

aartaka:
`cl-webkit` submodule commit should probably be bumped up—Nyxt doesn't compile with `cl-webkit` commit aecd76a2bcc7bfee049c91d94cc75f893800bf37.

Ambrevar:
Done in cc16e7e9e99e576b2c1dbb2228fadc25691ae394.

jmercouris:
Ah, thank you! I forgot it is a submodule nowadays!

Issue #1102, opened 2021-01-20
Move class-star to a separate repository


Ambrevar:
I've asked defclass* upstream if they are still maintaing their package and would accept to merge the class* changes.

If not, then I suggest we publish this library, it's too useful to be left inside Nyxt :)
@jmercouris: Shall I move it to @atlas-engineer or to @ambrevar?

jmercouris:
I would suggest all libraries we build in the context of Nyxt go to the atlas engineer repository, so that our community contributions are evident

Ambrevar:
Good news, upstream has accepted most of my patches.
Only one remains dangling: https://github.com/hu-dwim/hu.dwim.defclass-star/pull/3

If it gets merged, then we can switch to the latest version of `hu.dwim.deefclass-star`.

Ambrevar:
For now Quicklisp only publishes the `stable` tag of https://github.com/hu-dwim/hu.dwim.defclass-star, so we need to wait until this gets updated before we can get rid of our class-star library.

Ambrevar:
Maybe Quicklisp won't update the package anytime soon, so instead I suggest using a submodule for defclass-star.

Then we can remove `libraries/class-star` and define our own `define-class` in Nyxt that automatically sets the user class when the class name is exported.

Ambrevar:
When we do, rename `defcondition*` to `define-condition*`.

PR #1128, opened 2021-02-05
Upstream class star

Ambrevar:
Following up to #1102, I've upstreamed most of the changes brought by our `class-star`.
In particular, upstream has adopted the `class*` nickname which means we can no longer use it.

This change set it meant to ease the transition before the library gets updated on Quicklisp.

1. When Quicklisp updates defclass-star, then we will be able to finalize the package nickname renaming by renaming all the 

```lisp
(hu.dwim.defclass-star:make-name-transformer name)
```

to
```lisp
(class*:make-name-transformer name)
```

2. One change is still pending upstream: https://github.com/hu-dwim/hu.dwim.defclass-star/pull/3.  If it gets accepted, then we can safely remove our  `class-star`.  If not, then we need to update `class-star` to apply this one change.

Please check if this works for you.


aartaka:
Works for me!

Ambrevar:
Great!
@jmercouris OK to merge?

Ambrevar:
Thanks for testing Artyom!

jmercouris:
OK to merge, I guess we'll make it less wordy when upstream is done.

Issue #1095, opened 2021-01-07
Auto-mode does not handle restored buffers


Ambrevar:
Recipe:

- Open a URL for which you have an auto-mode rule.
- Restart the browser.

Result: The buffer has all default modes, the rule is not applied.

The same problem happens with `reopen-last-buffer` and `reopen-buffer`.

Any idea, @aartaka?

aartaka:
Hmmmm. Just checked. In my case it stored a `#<nyxt/reading-line-mode:user-reading-line-mode {1004F57EA3}>` instead of `reading-line-mode` in the rule. Was there a deserialization error in logs in your case? If so, then something's getting wrong with serialization of mode names.

aartaka:
Does b75c6bc solve that for you? I've tested and it seems problem indeed was in serialization.

Ambrevar:
I'll test later (possibly this weekend), working on the FOSDEM
presentation at the moment :)

Ambrevar:
Thanks for looking into it ;)

aartaka:
As discussed with @jmercouris, b75c6bc does not solve the problem -- while everything is serialized fine, there are still mode objects in memory, and auto-mode cannot process these.

Ambrevar:
I can still reproduce... :(

Ambrevar:
And... I can still reproduce! xD

Ambrevar:
Maybe related to #1875.

aartaka:
Done in 0d5ceadf9887b3bb805bb6106d8b44dec0033245.

Ambrevar:
Thanks!

PR #1135, opened 2021-02-09
Add better type checking to auto-mode with mode-invocation class.

aartaka:
This adds `mode-invocation` class to finally free us from checking whether `auto-mode`-rule has symbol, mode, or list as mode representation, and bugs stemming from this. This also includes some type declarations/checks added for type safety around the code of auto-mode.

# Things To Discuss:
- Is `mode-invocation` a good name? Alternatives:
  - `mode-invokation` (what's the right spelling?)
  - `mode-activation`
  - `mode-symbol` (see #868), although it collides with the function of the same name in mode.lisp.
- Maybe integrate this into mode.lisp too, so that we can always get mode arguments around when calling modes.
- Is there a better way to get mode symbol, other than `(nyxt::sym (nyxt::mode-command mode))`?
- Abstracts the `(or (cons mode-invocation) null)` type to `list-of-mode-invocations`?
- Initially, we persisted modes without package prefixes. We did switch to prefixes some time ago. Maybe switch back to unprefixed symbols for readability? User-edited unprefixed symbols work alright, though.

# Future work:
- #868 is covered (except for prompting what modes to enable, but is it really necessary?), close it?
- Mark some mode arguments as saveable by `auto-mode`. Example: we can move `current-zoom-ratio` to `web-mode` and persist it if it's not default. This way one will have their most comfortable zoom ratio depending on the website they visit. Cool, eh?

# How Has This Been Tested:
- Compiles. Runs.
- #1095 is not reproduced, so it should be fixed.

Closes: #1095

EDIT: Add note on testing.

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -12,40 +12,56 @@
   (trivial-package-local-nicknames:add-package-local-nickname :sera :serapeum)
   (trivial-package-local-nicknames:add-package-local-nickname :hooks :serapeum/contrib/hooks))
 
-(declaim (ftype (function ((or symbol root-mode list)) (values symbol &optional)) maybe-mode-name))
-(defun maybe-mode-name (mode)
-  (let ((mode (if (listp mode)
-                  (first mode)
-                  mode)))
-    (if (symbolp mode) mode (mode-name mode))))
+(define-class mode-invocation ()
+  ((name (error "Mode invocation should have a name to call mode through.")
+         :type symbol
+         :documentation "Mode symbol to call the mode with.
+Package prefix is optional.")
+   (arguments nil ;; TODO: "args" instead of "arguments"?
+              :type list
+              :documentation "Arguments to activate the mode with."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
 
-(declaim (ftype (function ((or symbol root-mode list) (or symbol root-mode list)) boolean)
-                mode-equal))
-(defun mode-equal (mode1 mode2)
-  (string-equal (symbol-name (maybe-mode-name mode1))
-                (symbol-name (maybe-mode-name mode2))))
+(sera:export-always 'equals)
+(defmethod equals ((mode1 mode-invocation) (mode2 mode-invocation))
+  (string-equal (symbol-name (name mode1))
+                (symbol-name (name mode2))))
 
-(declaim (ftype (function ((or symbol root-mode list) (or root-mode null)) list)
+(defgeneric mode-invocation (mode)
+  (:method ((mode mode-invocation))
+    mode)
+  (:method ((mode nyxt:root-mode))
+    (make-instance 'mode-invocation :name (mode-name mode)))
+  (:method ((mode symbol))
+    (make-instance 'mode-invocation :name (nyxt::sym (nyxt::mode-command mode))))
+  (:method ((mode list))
+    (check-type mode (cons symbol *))
+    (make-instance 'mode-invocation
+                   :name (first mode)
+                   :args (rest mode))))
+
+(declaim (ftype (function (list (or auto-mode null))
+                          (or (cons mode-invocation *) null))
                 rememberable-of))
 (defun rememberable-of (modes auto-mode)
   "Filter MODES based on rememberability by AUTO-MODE."
   (if auto-mode
-      (set-difference modes (non-rememberable-modes auto-mode)
-                      :test #'mode-equal)
+      (set-difference (mapcar #'mode-invocation modes)
+                      (non-rememberable-modes auto-mode)
+                      :test #'equals)
       modes))
 
 (define-class auto-mode-rule ()
   ((test (error "Slot `test' should be set.")
          :type list)
    (included '()
-             :type list
-             :documentation "The list of modes to enable on rule activation.
-Here 'modes' can be either mode symbols or a list when the first element is the
-mode symbols and the rest are the arguments to pass to the mode initiatiation
-function.")
+             :type (or (cons mode-invocation *) null)
+             :documentation "The list of `mode-invocation's to enable on rule activation.")
    (excluded '()
-             :type list-of-symbols
-             :documentation "The list of mode symbols to disable on rule activation.")
+             :type (or (cons mode-invocation *) null)
+             :documentation "The list of `mode-invocation's to disable on rule activation.")
    (exact-p nil
             :documentation "If non-nil, enable the INCLUDED modes exclusively.
 Enable INCLUDED modes plus the already present ones, and disable EXCLUDED modes, if nil."))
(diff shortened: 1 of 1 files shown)


jmercouris:
`mode-invocation` is the right spelling. I think I have been writing `invokation` because of German...

Ambrevar:
I think the pitfall comes for that in English, you would write 
"invoke" but "invocate".

source/auto-mode.lisp
```diff
@@ -12,40 +12,57 @@
   (trivial-package-local-nicknames:add-package-local-nickname :sera :serapeum)
   (trivial-package-local-nicknames:add-package-local-nickname :hooks :serapeum/contrib/hooks))
 
-(declaim (ftype (function ((or symbol root-mode list)) (values symbol &optional)) maybe-mode-name))
-(defun maybe-mode-name (mode)
-  (let ((mode (if (listp mode)
-                  (first mode)
-                  mode)))
-    (if (symbolp mode) mode (mode-name mode))))
+(define-class mode-invocation () ; TODO: Better name? mode-symbol collides with the mode.lisp function.
```
Ambrevar, line 15: I like the name :)

source/auto-mode.lisp
```diff
@@ -12,40 +12,57 @@
...
+  ((name (error "Mode invocation should have a name to call mode through.")
+         :type symbol
+         :documentation "Mode symbol to call the mode trough.
+Can be un-prefixed (i.e. interned to the current-package).")
```
Ambrevar, line 19: Typo: "trough" -> "through".  That said, I think it should be "with".

Maybe second line can be simplified:
"Package prefix is optional."

source/auto-mode.lisp
```diff
@@ -12,40 +12,57 @@
...
 
-(declaim (ftype (function ((or symbol root-mode list) (or symbol root-mode list)) boolean)
-                mode-equal))
-(defun mode-equal (mode1 mode2)
-  (string-equal (symbol-name (maybe-mode-name mode1))
-                (symbol-name (maybe-mode-name mode2))))
+(sera:export-always 'equals)
+(defmethod equals ((mode1 mode-invocation) (mode2 mode-invocation))
+  (string-equal (symbol-name (name mode1))
+                (symbol-name (name mode2))))
```
Ambrevar, line 30: Good catch for naming this `equals`!

source/auto-mode.lisp
```diff
@@ -12,40 +12,57 @@
...
 
-(declaim (ftype (function ((or symbol root-mode list) (or root-mode null)) list)
+(defgeneric mode-invocation (mode)
+  (:method ((mode mode-invocation))
+    mode)
+  (:method ((mode nyxt:root-mode))
+    (make-instance 'mode-invocation :name (mode-name mode)))
+  (:method ((mode symbol))
+    ;; TODO: Is there other way to know the right symbol for sure?
```
Ambrevar, line 38: If it's a symbol?  What you've used seems correct.  Why do you ask?

source/auto-mode.lisp
```diff
@@ -12,40 +12,57 @@
...
-             :documentation "The list of modes to enable on rule activation.
-Here 'modes' can be either mode symbols or a list when the first element is the
-mode symbols and the rest are the arguments to pass to the mode initiatiation
-function.")
+             :type (or (cons mode-invocation) null)
+             :documentation "The list of `mode-invocation's to enable on rule activation.")
    (excluded '()
-             :type list-of-symbols
-             :documentation "The list of mode symbols to disable on rule activation.")
+             :type (or (cons mode-invocation) null)
```
Ambrevar, line 64: Above you've used the syntax `(cons TYPE *)` which reads nicer in my opinion.  But I'm not completely sure it's equivalent.  If it works with the `*`, I'd rather add it.

source/auto-mode.lisp
```diff
@@ -92,23 +110,21 @@ Enable INCLUDED modes plus the already present ones, and disable EXCLUDED modes,
 
 (defun save-last-active-modes (auto-mode url)
   (when (can-save-last-active-modes auto-mode url)
-    ;; `last-active-modes' must be a list separate from the mode list, otherwise
-    ;; when modes get modified, last-active-modes would also be altered.
```
Ambrevar, line 96: Why remove this comment?

source/auto-mode.lisp
```diff
@@ -137,19 +154,23 @@ non-new-page requests, buffer URL is not altered."
...
-                            (buffer auto-mode))))
-        (member mode (or (and matching-rule (union (included matching-rule)
-                                                   (excluded matching-rule)))
-                         ;; Mode is covered by auto-mode only if it is
-                         ;; in last-active-modes and gets enabled.
-                         ;; If it gets disabled, user should be prompted,
-                         ;; because they may want to persist it.
-                         (and enable-p (last-active-modes auto-mode)))
-                :test #'mode-equal))))
+  (let ((mode (mode-invocation mode)))
```
Ambrevar, line 160: What about renaming this local `mode` to `invocation` to be consistent with `make-mode-toggle-prompting-handler` below?

source/auto-mode.lisp
```diff
@@ -233,11 +255,15 @@ Be careful with deleting the defaults -- it can be harmful for your browsing.")
 We need to store this to not overwrite the `last-active-modes' for a given URL,
 if `auto-mode-handler' will fire more than once.")
    (last-active-modes '()
-                      :documentation "The list of modes that were enabled
+                      :type (or (cons mode-invocation) null)
```
Ambrevar, line 258: Same comment as above for the `cons` type syntax.

Ambrevar:
Excellent, thank you so much for addressing this, it has been annoying me for too long ;)

Also very good improvement quality-wise.  A good data model is everything!

I haven't had the time to test it on my machine, will do later.

source/auto-mode.lisp
aartaka: Using two un-`export`-ed functions didn't seem right, so I thought me could've had some `export`-ed way to do that.

source/auto-mode.lisp
aartaka: `(cons TYPE)` is equivalent to `(cons TYPE *)` (source: [standard entry on `cons`](http://www.lispworks.com/documentation/HyperSpec/Body/t_cons.htm)). I'll add the star, it indeed does look better!

source/auto-mode.lisp
aartaka: There's now a layer of `mode-invocation`s between `default-modes` and `last-active-modes` so we don't need to care about copying `default-modes` by reference. It's irrelevant.

source/auto-mode.lisp
```diff
@@ -12,40 +12,56 @@
...
-  (let ((mode (if (listp mode)
-                  (first mode)
-                  mode)))
-    (if (symbolp mode) mode (mode-name mode))))
+(define-class mode-invocation ()
+  ((name (error "Mode invocation should have a name to call mode through.")
+         :type symbol
+         :documentation "Mode symbol to call the mode with.
+Package prefix is optional.")
+   (arguments nil ;; TODO: "args" instead of "arguments"?
```
jmercouris, line 20: arguments is better

source/auto-mode.lisp
jmercouris: :-)!

jmercouris:
Merci beaucoup Artyom!

Issue #1107, opened 2021-01-24
Nyxt icon missing in Archlinux gnome-session


utkarsh181:
Thanks for nyxt browser,

I have installled nyxt from AUR and even though Nyxt shows its
icon on startup, its icon disappears and Nyxt get replaced by
"sbcl".  In the given screenshot I am trying to highlight the missing
icon in gnome-app-preview when super key is pressed.

![Screenshot from 2021-01-24 19-00-43](https://user-images.githubusercontent.com/56307408/105632433-10bfef80-5e79-11eb-9f04-73ee689cd5b2.png)

Ambrevar:
Thanks for the bug report :)

What do you mean with "Nyxt get replaced by sbcl"?  Where do you see this?

I know why the icon changes, maybe because the window name is set at runtime?

utkarsh181:
Pierre Neidhardt <notifications@github.com> writes:

> Thanks for the bug report :)
>
> What do you mean with "Nyxt get replaced by sbcl"?  Where do you see this?

I am making an assumption that you are familiar with recent change
in gnome3 layout!

In the top left corner gnome has "Activities" area the same status bar
also shows focused application name which shows sbcl.
-- 
Utkarsh Singh

Ambrevar:
OK, makes sense.
We would need to try _not_ renaming the window, see if this helps with this issue.

aadcg:
@utkarsh181 Are you running GNOME on Wayland or X?  I can reproduce this issue on Wayland.  On X there's no issue with the Nyxt icon.

utkarsh181:
André Alexandre Gomes <notifications@github.com> writes:

> @utkarsh181 Are you running GNOME on Wayland or X?  I can reproduce this issue on Wayland.  On X there's no issue with the Nyxt icon.
>
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/atlas-engineer/nyxt/issues/1107#issuecomment-774432337

Yes I am running Gnome on Wayland.
-- 
Utkarsh Singh

aadcg:
Hm, it seems that on Wayland the `.desktop` file isn't being read, but I don't understand why.

[I played a bit with this thing called "looking glass" that you can trigger by hitting `Alt+F2` and typing `lg`](https://blogs.gnome.org/mclasen/2016/11/15/help-my-app-icon-is-missing/). 

[...]

PR #1141, opened 2021-02-10
Fix issue #1107

aadcg:
Addresses the fact that the Nyxt icon is missing on GNOME Wayland.

The key observation is that (gdk:gdk-set-program-class "nyxt") sets the
window manager class (WM class) to "nyxt" under the X window system,
whereas on Wayland the WM class is "sbcl" (which is the default for an
SBCL application).

(gdk:gdk-set-program-class "nyxt") is deprecated and not recommended.

The only shortcoming that I can think of is that if a user runs several SBCL applications, they will all be grouped together (under the same "frame").

Let me know what you think.

### DIFF
--- a/assets/nyxt.desktop
+++ b/assets/nyxt.desktop
@@ -13,3 +13,4 @@ Categories=GTK;Network;WebBrowser;
 MimeType=text/html;text/xml;application/xhtml+xml;application/xml;application/rss+xml;application/rdf+xml;image/gif;image/jpeg;image/png;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;x-scheme-handler/chrome;video/webm;application/x-xpinstall;
 StartupNotify=true
 Actions=NewWindow;
+StartupWMClass=sbcl


Ambrevar:
Thanks for working on this!

OK for the .desktop.  

For `gdk-set-program-class`, while it's true that it's deprecated I'd like an alternative so that the class is "nyxt" and not `sbcl`.  It has the shortcoming you mention, plus it would break existing workflows (like mine :p) which rely on the WM class name to find the windows ;)

Nit: Please name pull requests after what they do, not after the (cryptic) issue number they fix.  However, you are right to mention the issue number (either in the title or in the content, does not matter).

aadcg:
> it would break existing workflows (like mine :p) which rely on the WM class name to find the windows ;)

Right, makes sense. I force pushed the necessary changes.

Shall I create an issue to remind us that we need a better way to set the wmclass to "nyxt"? Given that it misbehaves on wayland and the method that is working under X is deprecated.

Ambrevar:
If we can't figure it out at the moment, then yes, we can merge this and
save the WM class reminder in a new issue.

Thanks!

aadcg:

> If we can't figure it out at the moment, then yes, we can merge this and
> save the WM class reminder in a new issue.

Even if such a solution exists, I wonder if we should ship it.  Up until
recently I was using sway, and got used to the fact that the wmclass was
"sbcl".  Eventually an angel descended from the skies and I gave in to
EXWM.  By the same token you mentioned regarding workflows like yours,
perhaps we should preserve things the way they are.  What do you think?

Regarding other solutions, I'm basically on a dark room trying to find a
pair of socks.

I found this

https://specifications.freedesktop.org/startup-notification-spec/startup-notification-0.1.txt

but it seems useless.

There's also this

https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-wmclass

Emacs sets the correct wmclass on both X and Wayland this way, on its C
core.  I have no idea if the same idea could be helpful for Nyxt...

Is any of this actually valuable?

Ambrevar:
Thanks for investigating.
Let's merge this for now, we can discuss the problem further in #1107.

Ambrevar:
Sadly gtk-window-set-wmclass is the same as what we are using, it's what
triggers the issue.

Adter some investigation, I found this:

https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-set-application-name

and

https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-set-prgname

I think we need `g-set-prgname`.  cl-cffi-gtk has it, so we are good.

@aadcg Wanna test and send a patch to use this instead?

aadcg:
Thanks for taking a look @Ambrevar 

> @aadcg Wanna test and send a patch to use this instead?

Sure, I will take a look.

Ambrevar:
> Sure, I will take a look.

Looks like someone else already did! :)

https://github.com/atlas-engineer/nyxt/pull/1174

PR #1144, opened 2021-02-11
Password improvements

jmercouris:
Several improvements to the interface:

+ Use slots within class to denote the location of the CLI (executable) program
+ Turn the sleep-timer into a slot so that it may be configured per password manager
+ Simplify the `make-xyz` methods for each password-interface
+ Enable the use of define-configuration macro for consistent configuration of password interfaces

### DIFF
--- a/libraries/password-manager/password-keepassxc.lisp
+++ b/libraries/password-manager/password-keepassxc.lisp
@@ -11,6 +11,8 @@
   (:export-accessor-names-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
 
+(push 'keepassxc-interface *interfaces*)
+
 (defmethod list-passwords ((password-interface keepassxc-interface))
   (let* ((st (make-string-input-stream (master-password password-interface)))
          (output (execute password-interface (list "ls" (password-file password-interface))
--- a/libraries/password-manager/password-keepassxc.lisp
+++ b/libraries/password-manager/password-keepassxc.lisp
@@ -22,41 +22,38 @@
 
 (defmethod list-passwords ((password-interface keepassxc-interface))
   (let* ((st (make-string-input-stream (master-password password-interface)))
-         (output (uiop:run-program (list (executable password-interface)
-                                         "ls" (password-file password-interface))
-                                   :input st :output '(:string :stripped t))))
+         (output (execute password-interface (list "ls" (password-file password-interface))
+                          :input st :output '(:string :stripped t))))
     (remove "Recycle Bin/" (rest (cl-ppcre:split "\\n" output)) :test #'equal)))
 
 (defmethod clip-password ((password-interface keepassxc-interface) &key password-name service)
   (declare (ignore service))
   (with-input-from-string (st (master-password password-interface))
-    (uiop:run-program (list (executable password-interface)
-                            "clip"
-                            (password-file password-interface)
-                            password-name
-                            ;; Timeout for password
-                            (format nil "~a" (sleep-timer password-interface)))
-                      :input st :output '(:string :stripped t))))
+    (execute password-interface (list "clip"
+                                      (password-file password-interface)
+                                      password-name
+                                      ;; Timeout for password
+                                      (format nil "~a" (sleep-timer password-interface)))
+             :input st :output '(:string :stripped t))))
 
 (defmethod save-password ((password-interface keepassxc-interface) &key password-name password service)
   (declare (ignore service))
   (with-input-from-string (st (format nil "~a~C~a"
                                       (master-password password-interface)
                                       #\newline password))
-    (uiop:run-program (list (executable password-interface)
-                            "add" "--password-prompt" (password-file password-interface)
-                            (if (str:emptyp password-name)
-                                "--generate"
-                                password-name))
-                      :input st)))
+    (execute password-interface (list "add" "--password-prompt" (password-file password-interface)
+                                      (if (str:emptyp password-name)
+                                          "--generate"
+                                          password-name))
+             :input st)))
 
 (defmethod password-correct-p ((password-interface keepassxc-interface))
   (when (master-password password-interface)
     (handler-case
         (let* ((st (make-string-input-stream (master-password password-interface)))
-               (output (uiop:run-program (list (executable password-interface)
-                                               "ls" (password-file password-interface))
-                                         :input st :output '(:string :stripped t))))
+               (output (execute password-interface (list (executable password-interface)
+                                                         "ls" (password-file password-interface))
+                                :input st :output '(:string :stripped t))))
           (remove "Recycle Bin/" (rest (cl-ppcre:split "\\n" output)) :test #'equal))
       (uiop/run-program:subprocess-error ()
         nil))))
--- a/libraries/password-manager/password-keepassxc.lisp
+++ b/libraries/password-manager/password-keepassxc.lisp
@@ -4,12 +4,9 @@
 (in-package :password)
 
 (define-class keepassxc-interface (password-interface)
-  ((executable :initform (executable-find "keepassxc-cli"))
-   (password-file :accessor password-file
-                  :initarg :file)
-   (master-password :accessor master-password
-                    :initarg :master-password
-                    :initform nil))
+  ((executable (executable-find "keepassxc-cli"))
+   (password-file)
+   (master-password :initform nil))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
--- a/libraries/password-manager/password-keepassxc.lisp
+++ b/libraries/password-manager/password-keepassxc.lisp
@@ -11,12 +11,6 @@
   (:export-accessor-names-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
 
-(eval-when (:compile-toplevel :load-toplevel :execute)
-  (export 'make-keepassxc-interface))
-(defun make-keepassxc-interface ()
-  (make-instance 'keepassxc-interface))
-(push #'make-keepassxc-interface interface-list)
-
 (defmethod list-passwords ((password-interface keepassxc-interface))
   (let* ((st (make-string-input-stream (master-password password-interface)))
          (output (execute password-interface (list "ls" (password-file password-interface))
--- a/libraries/password-manager/password-keepassxc.lisp
+++ b/libraries/password-manager/password-keepassxc.lisp
@@ -3,11 +3,6 @@
 
 (in-package :password)
 
-(eval-when (:compile-toplevel :load-toplevel :execute)
-  (export '*keepassxc-cli-program*))
-(defvar *keepassxc-cli-program* nil
-  "The path to the executable.")
-
 (defclass keepassxc-interface (password-interface)
   ((password-file :accessor password-file
                   :initarg :file)
@@ -18,15 +13,16 @@
 (eval-when (:compile-toplevel :load-toplevel :execute)
   (export 'make-keepassxc-interface))
 (defun make-keepassxc-interface ()
-  (unless *keepassxc-cli-program*
-    (setf *keepassxc-cli-program* (executable-find "keepassxc-cli")))
-  (when *keepassxc-cli-program*
-    (make-instance 'keepassxc-interface)))
+  (make-instance 'keepassxc-interface))
 (push #'make-keepassxc-interface interface-list)
+
+(defmethod initialize-instance :after ((password-interface keepassxc-interface))
+  (unless (slot-boundp password-interface 'executable)
+    (setf (executable password-interface) (executable-find "keepassxc-cli"))))
 
 (defmethod list-passwords ((password-interface keepassxc-interface))
   (let* ((st (make-string-input-stream (master-password password-interface)))
-         (output (uiop:run-program (list *keepassxc-cli-program*
+         (output (uiop:run-program (list (executable password-interface)
                                          "ls" (password-file password-interface))
                                    :input st :output '(:string :stripped t))))
     (remove "Recycle Bin/" (rest (cl-ppcre:split "\\n" output)) :test #'equal)))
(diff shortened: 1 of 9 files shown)


libraries/password-manager/password-keepassxc.lisp
```diff
@@ -3,51 +3,47 @@
...
-  (export '*keepassxc-cli-program*))
-(defvar *keepassxc-cli-program* nil
-  "The path to the executable.")
-
-(defclass keepassxc-interface (password-interface)
-  ((password-file :accessor password-file
+(define-class keepassxc-interface (password-interface)
+  ((executable :initform (executable-find "keepassxc-cli"))
+   (password-file :accessor password-file
                   :initarg :file)
```
aartaka, line 9: Not really necessary in `define-class`. 

A different initarg is actually worse than the autogenerated one in most cases. You need to remember it, instead of using slot name. Maybe remove this and settle with the one generated by `define-class` -- `:password-file`?

libraries/password-manager/password-pass.lisp
```diff
@@ -3,26 +3,22 @@
...
-(defvar *password-store-program* nil
-  "The path to the executable.")
-
-(defclass password-store-interface (password-interface)
-  ((password-directory :reader password-directory
+(define-class password-store-interface (password-interface)
+  ((executable (executable-find "pass"))
+   (sleep-timer (or (uiop:getenv "PASSWORD_STORE_CLIP_TIME") 45))
+   (password-directory :reader password-directory
                        :initarg :directory
```
aartaka, line 10: Again, unnecessary confusing initarg?

libraries/password-manager/password-pass.lisp
```diff
@@ -3,26 +3,22 @@
...
                        :initform (or (uiop:getenv "PASSWORD_STORE_DIR")
```
aartaka, line 11: `define-class` allows setting initform as the first form after slot name :)

libraries/password-manager/password.lisp
```diff
@@ -3,12 +3,11 @@
 
 (in-package :password)
 
-(eval-when (:compile-toplevel :load-toplevel :execute)
-  (export '*sleep-timer*))
-(defvar *sleep-timer* 15)
-
 (defclass password-interface ()
-  ())
+  ((executable :accessor executable :initarg :executable
```
aartaka, line 7: `:accessor` and `:initarg` are generated by `define-class` aren't they?

libraries/password-manager/password.lisp
```diff
@@ -3,12 +3,11 @@
...
+               :documentation "The program to query for password information.")
+   (sleep-timer :accessor sleep-timer :initarg :sleep-timer :initform 15
```
aartaka, line 9: Same as above -- initform first, everything else is autogenerated.

aartaka (commented):
Looks all good to me, except for some places where `define-class` is not used to its full power (i.e. autogenerated accessors and initargs, initforms after slot names). Fixing that will result in clearer and shorter code.

Do we need `interface-list` and `make-*-interface` helpers, after all? Maybe it's better to just expose the interfaces to user? Then they just set `password-interface` of `*browser*` to `(make-instance 'password:*-interface)` and we don't need unnecessary global state. Of course, this should be explained in the documentation.

libraries/password-manager/password-keepassxc.lisp
```diff
@@ -3,51 +3,47 @@
...
    (master-password :accessor master-password
                     :initarg :master-password
```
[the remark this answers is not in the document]
jmercouris: what is not necessary?

libraries/password-manager/password-keepassxc.lisp
jmercouris: oh the initarg I see!

libraries/password-manager/password-pass.lisp
jmercouris: I translated these defclass to define-class without looking carefully at the slots!

libraries/password-manager/password.lisp
jmercouris: this is not a define-class form

libraries/password-manager/password-keepassxc.lisp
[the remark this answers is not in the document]
aartaka: `:initarg` keyword and it's value. Initargs are generated automagically.

libraries/password-manager/password.lisp
aartaka: Oh, didn't see that. Why not, though?

jmercouris:
> Looks all good to me, except for some places where `define-class` is not used to its full power (i.e. autogenerated accessors and initargs, initforms after slot names). Fixing that will result in clearer and shorter code.
> 
> Do we need `interface-list` and `make-*-interface` helpers, after all? Maybe it's better to just expose the interfaces to user? Then they just set `password-interface` of `*browser*` to `(make-instance 'password:*-interface)` and we don't need unnecessary global state. Of course, this should be explained in the documentation.

Absolutely, it seems silly to me when they can just initialize the object themselves!

jmercouris:
Now, here is the one tricky part: if the user uses the define-configuration macro to alter these classes, how will they know the name of the class to initialize?

jmercouris:
We could do a (define-user-class as we do for renderer-gtk type things, at which point we'd be back at a (make-password-interface) type function :-D

jmercouris:
I have fixed it, please @aartaka can you test this on your system? I don't use a password manager and cannot test...

jmercouris:
From my testing, without running the executables, it *should* work.

jmercouris:
Here is a sample configuration:

```
(define-configuration password:keepassxc-interface
    ((password:executable "path/to/my/interface/executable")))
```

aartaka:
Hmmmmm... It does not display the minibuffer when I execute `copy-password`.

aartaka:
Twice as weird:
```
DEBUG - Password interface USER-KEEPASSXC-INTERFACE uses executable "/home/aartaka/.guix-profile/bin/keepassxc-cli".
WARN - Warning: Password manager "user-keepassxc-interface" does not support saving passwords.
```
Some inheritance issue here with `has-method-p` and user classes?

jmercouris:
Strange, that may be what is going on here, yes.

jmercouris:
I'll try to take a look tomorrow.

aartaka:
I've just pushed the fix to `has-method-p`.

The problem with `copy-password` still persists, though.

jmercouris:
Hm, interesting. if you do figure it out, feel free to push to this branch directly

aartaka:
Holy God, that was a tough typo to notice. See ff26769e for details about what broke it for me.

aartaka:
KeePassXC works like a charm and is a pleasure to configure! While I cannot test Pass and Security, they look okay.

jmercouris:
Oh wow! What a catch! Thank you Artyom :-)! I'll merge this in tomorrow.

nyxt.asd
```diff
@@ -270,7 +270,8 @@
                cl-ppcre
                str
                trivial-clipboard
-               uiop)
+               uiop
```
Ambrevar, line 273: I don't know why UIOP was here, but it should not be needed since it's a dependency of ASDF.

source/password.lisp
```diff
@@ -3,6 +3,25 @@
 
 (in-package :nyxt)
 
+;;; define user classes so that users may apply define-configuration
+;;; macro to change slot values
+(loop for interface in password:*interfaces* do
+         (eval `(define-user-class ,(intern (symbol-name interface)
+                                            (package-name (symbol-package interface))))))
```
Ambrevar, line 10: `eval` is brittle here.

One problem is that this code ie executed only once, which means that updates to password/libraries won't be carried out here.

Another problem is with third-party libraries: if someone defines a new password interface (say, LastPass), then the user class won't be defined.

Instead of this loop, I suggest making the user class at run time, e.g. in `make-password-interface`.  Thoughts?

source/password.lisp
```diff
@@ -3,6 +3,25 @@
...
+
+(defun make-password-interface ()
+  "Make a password interface for all user classes, the first one with
+a non-nil value of executable is considered to be the interface to
+make/return."
+  (labels ((user-password-interface-symbols ()
+             (loop for interface in password:*interfaces*
+                   collect (intern (str:concat "USER-" (symbol-name interface))
+                                   (package-name (symbol-package interface)))))
```
Ambrevar, line 19: Use `user-class-name` here.

libraries/password-manager/password.lisp
Ambrevar: Ping! :)

source/password.lisp
```diff
@@ -3,6 +3,25 @@
...
+           (user-password-interfaces ()
+             (loop for interface-symbol in (user-password-interface-symbols)
+                   collect (make-instance interface-symbol))))
+    (find-if #'password:executable (user-password-interfaces))))
```
Ambrevar, line 23: Better: this whole function can be replaced with

```lisp
    (find-if #'password:executable
             (mapcar (alex:compose #'make-instance #'user-class-name)
                     password:*interfaces*))))
```

Ambrevar:
This is a welcome change, great work!
I like the `execute` macro and how the `executable` is simply embedded in the slot!

I have one major concern though: the user classes are defined at compile time, which seems wrong.
Instead, I suggest making it part of the  `make-password-interface`.  I don't think `define-user-class` can be used (at least not without `eval`).  Instead, have a look at the body of `define-user-class`, it leverages `closer-mop:ensure-class`: since it's a function, then there is a no need for `eval`.

source/password.lisp
jmercouris: A good point, I wil do that.

nyxt.asd
jmercouris: It is not a guarantee that it will always be a dependency of ASDF.

Ambrevar:
Yes, it's in the doc:

>   * More generally, you can always rely on symbols from packages
>     'asdf', 'common-lisp' and 'uiop' being available in '.asd' files --

jmercouris:
Aha! today I learned! It says though they are available in the context of asd files, not that they will be within the compiled system, no?

Ambrevar:
> Aha! today I learned! It says though they are available in the context of asd files, not that they will be within the compiled system, no?

Since ASDF is the one responsible for loading the dependencies, it thus
guarantees that UIOP is loaded for the target system.

jmercouris:
It does not gaurantee it is in the compiled system! You can load and compile a system without asdf!

source/password.lisp
jmercouris: After relooking, we need it to happen before the user uses a define-configuration, therefore it must happen as a top level form. How else might we do it?

jmercouris: If they need a user class for their own configuration, that doesn't really make much sense, does it?

jmercouris: That is, if they make a new password interface class for some new interface, they wouldn't need a (define-user-class 'my-new-password-interface-class ...)

jmercouris:
I've made the improvements on master, I'm still not sure about `make-password-interface-user-classes` but it is the only way I could think of satisfying the potential problem. I think for now it is OK.

Ambrevar:

> It does not gaurantee it is in the compiled system! You can load and compile a system without asdf!

If a system depends on it, then UIOP is automatically dragged in the
compiled system.

Ambrevar:

> After relooking, we need it to happen before the user uses a define-configuration, therefore it must happen as a top level form. How else might we do it?

Hmmm... Maybe in `start` or somewhere similar, before the user init file
is loaded.

As a rule of thumb, we should never run code at the top-level.  This is
true for most programs and library, and it's especially true for Nyxt
since it's extensible: top-level code can't see the user extensions,
thus it won't behave as expected for the user :/

jmercouris:
The extensions must write their own `define-user-class` forms, otherwise the user will not be able to use a `define-configuration` macro within their init file for said class. This cannot be helped- UNLESS we provide a macro like `(define-configurable-class)` which operates like `(define-class)` BUT also spits out a `(define-user-class)`.

source/password.lisp
Ambrevar: I didn't mean (just) user-defined password interfaces, but most importantly third-party libraries.  For these, Nyxt need to define the user interface after it's started.

Forget what I said: it should be done _after_ the user init file is loaded, since third-party libraries are loaded only then.

Ambrevar:
> The extensions must write their own `define-user-class` forms, otherwise the user will not be able to use a `define-configuration` macro within their init file for said class. This cannot be helped- UNLESS we provide a macro like `(define-configurable-class)` which operates like `(define-class)` BUT also spits out a `(define-user-class)`.

Good point.
Forget what I just wrote then :p

Would be nice if we had some way to express "as soon as a subclass of
'password' is created, register its user class".

Otherwise the third-party package would have to do it.  They would have
to provide:

- the specialized password implementation;
- a tiny Nyxt-specialized file that contains the user class.

jmercouris:
> The extensions must write their own `define-user-class` forms, otherwise the user will not be able to use a `define-configuration` macro within their init file for said class. This cannot be helped- UNLESS we provide a macro like `(define-configurable-class)` which operates like `(define-class)` BUT also spits out a `(define-user-class)`.
> Good point. Forget what I just wrote then :p Would be nice if we had some way to express "as soon as a subclass of 'password' is created, register its user class". Otherwise the third-party package would have to do it. They would have to provide: - the specialized password implementation; - a tiny Nyxt-specialized file that contains the user class.

Right, that is why I am suggesting that we provide them a define-class type macro that automatically creates a user-class* variant for it. What do you think?

Ambrevar:

> Right, that is why I am suggesting that we provide them a define-class type macro that automatically creates a user-class* variant for it. What do you think?

Seems to be right, but I remember thinking about it in the past and
decided to go against it.  Not sure why, can't remember.  Maybe simply
because I was too lazy or just unsure it was going to be useful?

Anyways, I suggest sending a pull request, to see how it would look like.

We could have our `define-class' do it, since in the future we are going
to leverage defclass* for the rest.

Issue #1140, opened 2021-02-09
With `open-external-link-in-new-window-p` set to `t` and internal buffer as current buffer, `set-url` opens a new window


edgar-vincent:
**Describe the bug**

**Precise recipe to reproduce the issue**

With `open-external-link-in-new-window-p` set to `t`, opening a URL with `set-url` with an internal buffer as current buffer will open a new window, although it is not an external link.

**Information**
- OS name+version: Arch Linux
- Graphics card and driver: Intel Inegrated Graphics
- Desktop environment / Window manager name+version: Gnome 3.38.3
- How you installed Nyxt (Guix pack, package manager, build from source): nyxt-browser-git from the AUR
- Information from command copy-system-information:
```
Nyxt version: 2-pre-release-5-438-g5355f34e 
Operating system kernel: Linux 5.10.13-arch1-2
Lisp implementation: SBCL 2.1.1
Features: (WEBKIT2 WEBKIT2-2.30 WEBKIT2-TRACKING WEBKIT2-MUTE WEBKIT2-EMOJI
           WEBKIT2-MEDIA WEBKIT2-SANDBOXING GTK-3-22 GTK-3-20 GTK-3-18 GTK-3-16
           GTK-3-14 GTK-3-12 GTK-3-10 GTK-3-8 GTK-3-6 GTK-3-4 GTK GDK-3-22
           GDK-3-20 GDK-3-18 GDK-3-16 GDK-3-14 GDK-3-12 GDK-3-10 GDK-3-8
           GDK-3-6 GDK-3-4 CAIRO-1-10 CAIRO-1-12 GDK-PIXBUF GLIB-2-30 GLIB-2-32
           GLIB-2-34 GLIB-2-36 GLIB-2-38 GLIB-2-40 GLIB-2-42 GLIB-2-44
           GLIB-2-46 GLIB-2-48 GLIB-2-50 GLIB-2-52 GLIB-2-54 GLIB-2-56
           GLIB-2-58 GLIB FSET-EXT-STRINGS CUSTOM-HASH-TABLE-NATIVE SWANK
           PLUMP-UTF-32 GLOBAL-VARS DECLARE-TYPES PARENSCRIPT NAMED-READTABLES
           OSICAT-FD-STREAMS 21BIT-CHARS CHUNGA FLEXI-STREAMS CLOSER-MOP
[...]

PR #1145, opened 2021-02-12
browser.lisp: fix spawning new windows from internal buffers

aadcg:
A dirty fix for issue #1140.

There was quite a bit for me to unpack while looking at this. I don't know much about browsers honestly.

I hate this solution but I'm afraid my understanding isn't still quite where it should be.

Namely, I wonder if it makes sense that there is a `new-window-p` slot in the `request-data` class, and the `open-external-link-in-new-window-p` slot in the `browser` class.

I feel that some rewriting is probably the right thing to do but I feel I'm not up to it yet.  I need to understand the architecture of Nyxt a bit better first.

If you have something on your mind that you would like to detail, I can probably follow it and take actions accordingly.


jmercouris:
A solution is a solution even if not ideal! I guess this does fix the problem :-D. I would like some time to think about what potential rearchitecting (as you've hinted at) may be necessary for a cleaner (generalized) solution.

Ambrevar:
I can't remember the exact flow of events of the top of my head, but since I wrote this I have a feeling that this was done poorly :)  It's also very old code, so it's probably a good time to refactor it now that we've got a more stable code base.

I'll have a look later, I'm sure we can come up with a clean design here.
It's also probably not much code and André's intuition may be right.

aadcg:
Feel free to take a deeper look and to have a short discussing over a (possible) redesign.

While I feel I'm of little valuable to the discussion, I'd be happy to make any possible changes you find necessary :)

Ambrevar:
I just had a look: at first glance, I thought your fix was actually the right thing to do here... Until I realized that the mistake lied one level up, and actually the log message gives us a hint here:

```lisp
          (log:debug "Load URL from internal buffer in new buffer: ~a" (object-display url))
          (open-urls (list (object-string url)))
```

The problem is that we are not supposed to call `open-urls` here, a simple `make-buffer-focus` suffices!

Fixed with 1212f4722a6c2936f603de5133f58cc3497db7bd.
Thanks for investigating @aadcg !

Ambrevar:
gg

aadcg:

> The problem is that we are not supposed to call open-urls here, a simple
> make-buffer-focus suffices!

I'm trying to follow you, but I'm wondering if this solution works what
if all buffers are internal-buffers.  So, perhaps, make-buffer-focus can
move the focus to the new created buffer, before the url is loaded?

Thank you @ambrevar!

Ambrevar:
Sorry, I don't understand the problem.  Can you provide a recipe?

aadcg:

> Sorry, I don't understand the problem. Can you provide a recipe?

Sorry, I noticed that there were misprints in my comment!  I edited.

I thought about it again, and it seems to me that your solution will
work for sure in the case I wondered about.  But I'll check it myself
when I overcome my building issues.

Ambrevar:
Sorry, still don't understand! :)

The context here is the current buffer.  The other buffers don't matter here.  So having only internal-buffers in our buffer list should not change anything.

`make-buffer-focus` creates a buffer, loads the URL in the background in focuses to it.  I don't think that loading the URL before or after the focus would change anything either.

Does that make sense?

aadcg:

> I don't think that loading the URL before or after the focus would
> change anything either.

Exactly, that was what made me wonder.  All clear now.

PR #1146, opened 2021-02-12
Fix and review some docstrings

aadcg:


source/browser.lisp
```diff
@@ -370,13 +370,14 @@ If none is found, fall back to `scheme:cua'."
...
    (known-type-p nil
-                 :documentation "Whether the request is for a contented with
-supported MIME-type (e.g. a picture that can be displayed in
-the web view.")
+                 :documentation "Whether the request is for content with
+supported MIME-type, e.g. a picture that can be displayed in the web
+view.")
    (keys '()
-         :documentation "The key sequence that was pressed to generate the request."))
+         :documentation "A list of input key sequences that generate the request."))
```
Ambrevar, line 380: It's not a list of input key sequences, that would mean it's a list of list! :)
If you don't like the word "sequence", then we can write "A list of key chords".

source/browser.lisp
```diff
@@ -370,13 +370,14 @@ If none is found, fall back to `scheme:cua'."
...
-                 :documentation "Whether the request wants to happen in a new window.")
+                 :documentation "Whether the request takes place in a
+new window.")
    (known-type-p nil
-                 :documentation "Whether the request is for a contented with
-supported MIME-type (e.g. a picture that can be displayed in
-the web view.")
+                 :documentation "Whether the request is for content with
+supported MIME-type, e.g. a picture that can be displayed in the web
+view.")
```
Ambrevar, line 378: While we are at it, I suggest we replace the academic sounding "e.g." with "such as".

source/external-editor.lisp
```diff
@@ -13,10 +13,10 @@ is blocking, invoke on a separate thread when possible."
...
 
 (define-command fill-input-from-external-editor ()
-  "This command will open your editor specified by your VISUAL-EDITOR
-  of the BROWSER class, if unset, it will default to your VISUAL
-  environment variable. It will then capture whatever text you enter
-  and save in your editor."
+  "This command will open your editor specified by your VISUAL-EDITOR of
+the BROWSER class, if unset, it will default to your VISUAL environment
+variable. It will then capture whatever text you enter and save in your
+editor."
```
Ambrevar, line 19: Did you just refill this paragraph?
If so, then it means that the fill-column was not the same as your when this was written.

We currently don't enforce any fill-column.  @jmercouris: Shall we add an Emacs local file / dir property to set the fill-column to a well defined value to avoid this kind of issues?

In all cases, this belongs to a separate patch set.

Ambrevar:
Thanks for the patch, documentation improvements are always welcome!

Could you rebase against master, so that this pull request only contains the change commits and _no_ merge commit?

Also it's good practice to name the pull requests as precisely as possible (I know I know I don't always do it :p) so here maybe "Fix some docstring typos" would be more explicit ;)

Thanks!

aadcg:
> Could you rebase against master, so that this pull request only contains the change commits and _no_ merge commit?
 
Oops, still a bit of a git noob here.  I have briefly read Progit at some point but it wouldn't hurt to review some things again.

> Also it's good practice to name the pull requests as precisely as possible (I know I know I don't always do it :p) so here maybe "Fix some docstring typos" would be more explicit ;)

Sure!

source/external-editor.lisp
aadcg: What you're addresing is important but notice that there were extra whitespaces in the beginning of each line.  I refill it indeed, but I was just fixing the following visual issue.

Ambrevar: Ah, yes, this is another styling issue.  John likes it better this way, but right now it does not work well with how we display docstrings within Nyxt.

In the future, we might be able to parse and format the docstrings more cleanly.  This would be the proper fix.

In the mean time, I suggest we avoid the leading whitespace, as you did here.  Thanks!

source/browser.lisp
aadcg: Oh right, of course. 

Sequences doesn't brother me.  I wanted to avoid "pressed".  What about "The key sequence that generates the request"?

Ambrevar: Perfect!

source/external-editor.lisp
jmercouris: A fill column would be a good idea, yes! I can look at some CSS that might fix the leading breaking spaces, or maybe I can do parsing to remove excess spaces from docstrings. I believe that we can utilize: str:collapse-whitespaces

Ambrevar:

> A fill column would be a good idea, yes! I can look at some CSS that might fix the leading breaking spaces, or maybe I can do parsing to remove excess spaces from docstrings. I believe that we can utilize: str:collapse-whitespaces

Sounds good!

aadcg:
Is there something else I can do for now?

source/browser.lisp
```diff
@@ -370,13 +370,14 @@ If none is found, fall back to `scheme:cua'."
...
    (known-type-p nil
-                 :documentation "Whether the request is for a contented with
-supported MIME-type (e.g. a picture that can be displayed in
-the web view.")
+                 :documentation "Whether the request is for content with
+supported MIME-type, such as a picture that can be displayed in the web
+view.")
    (keys '()
-         :documentation "The key sequence that was pressed to generate the request."))
+         :documentation "The key sequence that generates the request."))
```
Ambrevar, line 380: Nit: generated

jmercouris:
No, this looks OK!

Ambrevar:
Looks good to me, just a small typo left :)

aadcg:
> Looks good to me, just a small typo left :)

Done!

jmercouris:
Thank you!

PR #1151, opened 2021-02-16
Protect auto mode

Ambrevar:
This is a follow up to #1135.

It fixes a bunch of things, see the commit messages.
Important fix: now it does not fail when the mode symbol is not known.  This can happen when the user uninstalls a third-party package, of if they remove their own mode from their `init.lisp`.

If you like `maybe-muffle-condition` (better name?) I'll use it in other places, like `open-urls` and history's `restore`.

@jmercouris @aartaka ?


aartaka:
`maybe-muffle-condition` is a reasonable name and it fits `restore`. Indeed, applying it to other places sounds great!

jmercouris:
Without having tested it, this changes appears reasonable enough to me.

source/auto-mode.lisp
```diff
@@ -37,7 +37,12 @@ Package prefix is optional.")
   (:method ((mode nyxt:root-mode))
     (make-instance 'mode-invocation :name (mode-name mode)))
   (:method ((mode symbol))
-    (make-instance 'mode-invocation :name (nyxt::sym (nyxt::mode-command mode))))
+    (let* ((command (nyxt::mode-command mode))
+           (command-sym (and command
```
jmercouris, line 41: A bit of a strange way of setting the value via the use of a predicate. Maybe this would be a good candidate for when let.

Ambrevar: I don't think we have `if-let*` or `when-let*`.  Any better suggestion?

jmercouris: what do you mean? it is in Alexandria.

Ambrevar:
`maybe-muffle-condition` is really the macro counterpart of `funcall-safely`.  I think the names should highlight their relationship.

What about `with-safe-body`?

source/auto-mode.lisp
Ambrevar: O.o !?!?!  Wow, somehow I missed these functions and more than once I wrote code with clumsy `let`s because I thought they were missing!!!

Today I learnt something! :p

Ambrevar: Done (force-pushed).

aartaka:
> `maybe-muffle-condition` is really the macro counterpart of `funcall-safely`. I think the names should highlight their relationship.
> 
> What about `with-safe-body`?

`safe` is too generic, so it won't help there. In the same vein, it does not fully fit `funcall-safely`. `quiet` maybe? `funcall-quiet` and `with-quiet-body`:
- sounds like something suppressing conditions,
- reminds of `-q`/`--quiet` args to UN*X commands that mean no reporting/output,
- gives you no guarantees of safety.

But I'm okay with `with-safe-body` if we stick to `funcall-safely` -- relationship is important to emphasize.

Ambrevar:
"Muffled" is more common than "quiet" in Common Lisp, so what about "with-muffled-body"?

aartaka:
> "Muffled" is more common than "quiet" in Common Lisp, so what about "with-muffled-body"?

I'm totally OK with that :)

PR #1152, opened 2021-02-16
buffer.lisp: move box-style, highlighted-box-style, hints-alphabet

jmercouris:
buffer.lisp: move box-style, highlighted-box-style, hints-alphabet

makes it more logical. The buffer should not have slots relating to the int alphabet, etc. Ideally we would make a hint-mode as well. I tried doing this, but it ended up being a gigantic rearchitecture that involved tweaking search-buffer, visual-mode, and all element-hints. I will save that for another time (after I think more about how to generalize the hinting facilities).

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -160,25 +160,6 @@ distance scroll-left or scroll-right will scroll.")
                       :documentation "The ratio of the page to scroll.
 A value of 0.95 means that the bottom 5% will be the top 5% when scrolling
 down.")
-   (box-style (cl-css:css
-               '((".nyxt-hint"
-                  :background "linear-gradient(#fcff9e, #efcc00)"
-                  :color "black"
-                  :border "1px black solid"
-                  :padding "1px 3px 1px 3px"
-                  :border-radius "2px"
-                  :z-index #.(1- (expt 2 31)))))
-              :documentation "The style of the boxes, e.g. link hints.")
-   (highlighted-box-style (cl-css:css
-                           '((".nyxt-hint.nyxt-highlight-hint"
-                              :font-weight "500"
-                              :background "#fcff9e")))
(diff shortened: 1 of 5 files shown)


jmercouris:
I also think that the following should somehow be made legal:

```
(DEFINE-CONFIGURATION nyxt/web-mode:web-mode
  ((HINTS-ALPHABET "asdfghjkl")))
```

instead of having to do:

```
(DEFINE-CONFIGURATION nyxt/web-mode:web-mode
  ((nyxt/web-mode::HINTS-ALPHABET "asdfghjkl")))
```

jmercouris:
In other words, within the define-configuration form, the body should prefix all symbols automatically. How to do this? I don't know. There are a number of challenges...

Ambrevar:
Thanks, looks good to me.

Out of curiosity, GitHub (and the email I received) mentions that @aartaka and my review were requested, but then it got merged immediately afterwards.  GitHub bug or mistake?

About the slot package prefix question: there is a slight decision problem here: some slots may indeed be in nyxt-user while others be in the symbol mode package.  This can happen when a mode inherits from another one, in a different package.  So I think it's better to be explicit here, lest we trip the user with hard-to-spot bugs.

jmercouris:
What happened was that I was going to request a review, then I realized that we had already talked about it on Monday and said we should do it.

OK, let us not trip up the user then indeed! I would like to make the auto-config.lisp smarter then, as it currently emits unprefixed symbols.

jmercouris:
Unprefixed both for class names and slot designators.

Ambrevar:
Can you please open an issue about this?

jmercouris:
OK!

jmercouris:
https://github.com/atlas-engineer/nyxt/issues/1156

Ambrevar:
Thanks!

jmercouris:
C'est pas no problem!

PR #1153, opened 2021-02-17
Nosave buffers

### LINKED ISSUE #247: incognito mode

aartaka:
This is a second attempt at nosave buffers (supersedes #1093).

# What's new:

- `private-data-profile` renamed to `nosave-data-profile` for clarity and obvious relation to `nosave-buffer`.
- `nosave-buffer` class that has `nosave-data-profile` set by default but behaves like `web-buffer` otherwise.
  - All `nosave-buffer`s have separate ephemeral WebKit contexts and data managers.
  - `set-url` minibuffer history and all the data is isolated, so no data is leaking.
  - `nosave-buffer` cannot be parent to any other buffer.
  - History of `nosave-buffer` is not saved (although I checked restored buffers only and haven't looked an owners' IDs in the history file).
- Lots of places where we access data while being in between two buffers now use `with-current-buffer` to explicitly state the buffer to `expand-path` in.
- `set-url` and `follow-hint` versions for `nosave-buffer`

# Things To Discuss:

- VI bindings for `nosave-buffer` commands? I need help here, as I'm not familliar with VI keybinding mnemonics principles.
- `follow-hint-new-buffer` from the `nosave-buffer` is contagious: instead of new `web-buffer`, a new `nosave-buffer` is created. Should it be this way?
- There's quite a complex format-string used in `set-url` that I introduced. Maybe it's better to stick to older form (i.e., explicit `if`s instead of `format` DSL conditionals)?
- Make `nosave-buffer` visually noticeable? We discussed painting mode-line in a darker palette to emphasize `nosave-buffer` before.

# How Has This Been Tested
- I've ran Nyxt and used `nosave-buffer`s on several social media websites (logging in, out, browsing, and closing the buffer). No cookies, user-data, or any other trail of `nosave-buffer`s was noticed after their deletion.
- I believe it's merge-ready, but could've not seen some bugs, so review and testing is welcome :)

Closes #247

EDIT: Add question about noticeability of `nosave-buffer`s.

### DIFF
--- a/source/base-mode.lisp
+++ b/source/base-mode.lisp
@@ -22,6 +22,7 @@
                      "C-l" 'set-url-from-current-url
                      "C-u C-l" 'set-url ; `set-url' is listed in the tutorial, so it should be bound.
                      "M-l" 'set-url-new-buffer
+                     "C-u M-l" 'set-url-nosave-buffer
                      "f5" 'reload-current-buffer
                      "C-r" 'reload-current-buffer
                      "C-R" 'reload-buffer
@@ -75,6 +76,7 @@
                      "C-pagedown" 'switch-buffer-next
                      "C-l" 'set-url
                      "M-l" 'set-url-new-buffer
+                     "C-u M-l" 'set-url-nosave-buffer
                      "C-t" 'make-buffer-focus
                      "C-r" 'reload-current-buffer
                      "C-R" 'reload-buffer
@@ -118,6 +120,7 @@
                      "B" 'make-buffer-focus
                      "o" 'set-url
                      "O" 'set-url-new-buffer
+                     ;; TODO: vim-ish set-url-nosave-buffer keybinding?
                      "m u" 'bookmark-url
                      "m d" 'bookmark-delete
                      "R" 'reload-current-buffer
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -874,10 +874,8 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
     (when history
       (containers:insert-item history (url (current-buffer))))
     (let ((url (prompt-minibuffer
-                :input-prompt (format nil "Open URL in ~A buffer"
-                                      (if new-buffer-p
-                                          "new"
-                                          "current"))
+                :input-prompt (format nil "Open URL in ~:[current~;new~]~:[~; nosave~] buffer"
+                                      new-buffer-p nosave-buffer-p)
                 :input-buffer (if prefill-current-url-p
                                   (object-string (url (current-buffer))) "")
                 :default-modes '(set-url-mode minibuffer-mode)
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -567,7 +567,8 @@ If URL is `:default', use `default-new-buffer-url'."
   "Make buffer with title TITLE and modes DEFAULT-MODES.
 Run `*browser*'s `buffer-make-hook' over the created buffer before returning it.
 If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
-  (let* ((buffer (if dead-buffer
+  (let* ((nosave-buffer-p (or nosave-buffer-p (nosave-buffer-p parent-buffer)))
+         (buffer (if dead-buffer
                      (progn
                        ;; Dead buffer ID must be renewed before calling `ffi-buffer-make'.
                        (setf (id dead-buffer) (get-unique-buffer-identifier *browser*))
@@ -599,7 +600,9 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
           ;; default value.
           (unless (htree:owner history (id buffer))
             (htree:add-owner history (id buffer)
-                             :creator-id (when parent-buffer
+                             :creator-id (when (and parent-buffer
+                                                    (not nosave-buffer-p)
+                                                    (not (nosave-buffer-p parent-buffer)))
                                            (id parent-buffer)))))))
     ;; Run hooks before `initialize-modes' to allow for last-minute modification
     ;; of the default modes.
@@ -865,7 +868,9 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
 
 (define-command set-url (&key new-buffer-p prefill-current-url-p nosave-buffer-p)
   "Set the URL for the current buffer, completing with history."
-  (let ((history (unless nosave-buffer-p (minibuffer-set-url-history *browser*))))
+  (let ((nosave-buffer-p (or nosave-buffer-p
+                             (nosave-buffer-p (current-buffer))))
+        (history (unless nosave-buffer-p (minibuffer-set-url-history *browser*))))
     (when history
       (containers:insert-item history (url (current-buffer))))
     (let ((url (prompt-minibuffer
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -274,6 +274,17 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
   (when (expand-path (cookies-path buffer))
     (ensure-parent-exists (expand-path (cookies-path buffer)))))
 
+(define-class nosave-buffer (user-web-buffer)
+  ((data-profile (make-instance 'nosave-data-profile))
+   (default-modes '(reduce-tracking-mode certificate-exception-mode
+                    web-mode base-mode)))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:export-predicate-name-p t)
+  (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
+
+(define-user-class nosave-buffer)
+
 (define-class internal-buffer (user-buffer)
   ((style #.(cl-css:css
              '((body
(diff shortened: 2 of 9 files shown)


jmercouris:
I see this, and will go through it shortly! Just acknowledging receipt!

source/buffer.lisp
```diff
@@ -861,6 +902,10 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
   "Prompt for a URL and set it in a new focused buffer."
   (set-url :new-buffer-p t))
 
+(define-command set-url-nosave-buffer ()
```
jmercouris, line 905: It could be argued that this would be better implemented as a regular argument to set-url. I'm personally not sure what is better.

Here is the question to answer: will we ever need to change some thing in both set-url-nosave-buffer and set-url? If so, then it should probably be an argument. If they will not evolve in lockstep, then this is probably OK.

aartaka: But it _is_ just a wrapper around the `set-url` that has an argument indicating whether it creates `nosave-buffer` :)

We do need `set-url-nosave-buffer` to bind it to keys, in the same way as we need `set-url-new-buffer` and `set-url-from-current-url`.

jmercouris:
Firstly; thank you Artyom for this long awaited feature! It will be a pleasure to have!

My thoughts on the high level concepts/questions within this pull request:

>     * `private-data-profile` renamed to `nosave-data-profile` for clarity and obvious relation to `nosave-buffer`.
OK

>     * `nosave-buffer` class that has `nosave-data-profile` set by default but behaves like `web-buffer` otherwise.
OK

>       * All `nosave-buffer`s have separate ephemeral WebKit contexts and data managers.
Good catch!

>       * `set-url` minibuffer history and all the data is isolated, so no data is leaking.
OK

>       * `nosave-buffer` cannot be parent to any other buffer.
OK

>       * History of `nosave-buffer` is not saved (although I checked restored buffers only and haven't looked an owners' IDs in the history file).
OK, will verify.

>     * Lots of places where we access data while being in between two buffers now use `with-current-buffer` to explicitly state the buffer to `expand-path` in.
What do you mean about accessing data "between two buffers"? Can you please give an example?

>     * `set-url` and `follow-hint` versions for `nosave-buffer`
As I've commented in the source, if set-url and set-url-nosave-buffer can evolve independently, this is OK. Otherwise, I would suggest an argument to set-url.

> # Things To Discuss:
> 
>     * `follow-hint-new-buffer` from the `nosave-buffer` is contagious: instead of new `web-buffer`, a new `nosave-buffer` is created. Should it be this way?

It should be this way, otherwise the user may be surprised by unintended behavior. For example, in every browser, when you make a 'private window' and you make new tabs, all of those tabs are private. What may be unclear to a user is how to now make a new non nosave-buffer. What will they have to do? Shall we prompt them?

>     * There's quite a complex format-string used in `set-url` that I introduced. Maybe it's better to stick to older form (i.e., explicit `if`s instead of `format` DSL conditionals)?

This is good. Sure, nobody can read it, but that's fine :-D. There are lots of people who criticize format, myself often included. I would still like to learn some of its more advanced syntax.

>     * Make `nosave-buffer` visually noticeable? We discussed painting mode-line in a darker palette to emphasize `nosave-buffer` before.

Yes, I will think about how to do this. It should probably be something with the status area. I may push something on this branch if you don't mind.

jmercouris:
Potential bug: middle clicking on a link within a nosave-buffer appears to result in the creation of a normal buffer.

aartaka:
> > ```
> > * Lots of places where we access data while being in between two buffers now use `with-current-buffer` to explicitly state the buffer to `expand-path` in.
> > ```
> 
> What do you mean about accessing data "between two buffers"? Can you please give an example?

When we switch buffer, we set the history current owner to the new buffer. Previously we were getting the history from the new buffer path, governed by the `data-profile` of the _old_ buffer (see `current-data-profile` and its usage in `get-data` -- it looks at the `%buffer` and gets profile there). It's not right, because _both_ path and profile should come from new buffer to make sure nosave buffers don't leak the information to the web buffers that they switch to, and vice versa.

> > ```
> > * `set-url` and `follow-hint` versions for `nosave-buffer`
> > ```
> 
> As I've commented in the source, if set-url and set-url-nosave-buffer can evolve independently, this is OK. Otherwise, I would suggest an argument to set-url.

There is an argument :upside_down_face: 

The  separate command is for the user to bind, as you can see in `base-mode`.

> > ```
> > * `follow-hint-new-buffer` from the `nosave-buffer` is contagious: instead of new `web-buffer`, a new `nosave-buffer` is created. Should it be this way?
> > ```
> 
> It should be this way, otherwise the user may be surprised by unintended behavior. For example, in every browser, when you make a 'private window' and you make new tabs, all of those tabs are private. 

I do agree with this logic, but we still need to re-iterate it in the manual or elsewhere.

> What may be unclear to a user is how to now make a new non nosave-buffer. What will they have to do? Shall we prompt them?

`make-buffer` :) That's not always obvious, though.

> > ```
> > * There's quite a complex format-string used in `set-url` that I introduced. Maybe it's better to stick to older form (i.e., explicit `if`s instead of `format` DSL conditionals)?
> > ```
> 
> This is good. Sure, nobody can read it, but that's fine :-D. There are lots of people who criticize format, myself often included. I would still like to learn some of its more advanced syntax.

I've learned it and now it's dragging me to the dark side of convoluted Turing-complete control strings that replace functions ;)

> > ```
> > * Make `nosave-buffer` visually noticeable? We discussed painting mode-line in a darker palette to emphasize `nosave-buffer` before.
> > ```
> 
> Yes, I will think about how to do this. It should probably be something with the status area. I may push something on this branch if you don't mind.

I don't mind! :)

aartaka:
> Potential bug: middle clicking on a link within a nosave-buffer appears to result in the creation of a normal buffer.

Lots of things do that, actually, including some of the `set-url` commands. It was unclear to me in general -- should we always make buffers opened from the `nosave` one to be `nosave` too? 

Now I'm sure we should. I'll change the behavior to be even more contagious :)

aartaka:
Done! Now `nosave-buffer`s are almost unescapable -- the only way to create a regular buffer from the `nosave` one is to call `make-buffer`.

jmercouris:
> Done! Now `nosave-buffer`s are almost unescapable -- the only way to create a regular buffer from the `nosave` one is to call `make-buffer`.

Perfekt!

jmercouris:
I've just added a glyph for now. I think that should suffice. I went back and forth many times in my head about what the right thing to do is... :-D

jmercouris:
About the last point: separate argument to bind. At one point we used to be able to bind lambdas, so you could just bind a lambda with the argument. I'm not sure if this possible any more. Something tells me it is not (but only because of an fdeclaration).

jmercouris:
OK, apparently, you CAN bind a lambda, we did change that declaration :-D

source/base-mode.lisp
```diff
@@ -118,6 +120,7 @@
                      "B" 'make-buffer-focus
                      "o" 'set-url
                      "O" 'set-url-new-buffer
+                     ;; TODO: vim-ish set-url-nosave-buffer keybinding?
```
Ambrevar, line 123: I don't think there is one.  If I recall correctly, Qutebrowser takes an argument in the command prompt.

A few "vim-ish" options:
- `g o`: Indeed, `g` is sometimes the equivalent of a prefix argument in Emacs, for instance `gj` goes one visual line down.
- `z o`: same.

source/buffer.lisp
```diff
@@ -528,24 +556,27 @@ If URL is `:default', use `default-new-buffer-url'."
...
 (defun buffer-make (browser &key data-profile title default-modes
-                              dead-buffer internal-buffer-p parent-buffer
-                              no-history-p)
+                              dead-buffer nosave-buffer-p internal-buffer-p
+                              parent-buffer no-history-p)
   "Make buffer with title TITLE and modes DEFAULT-MODES.
 Run `*browser*'s `buffer-make-hook' over the created buffer before returning it.
 If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
-  (let* ((buffer (if dead-buffer
+  (let* ((nosave-buffer-p (or nosave-buffer-p (nosave-buffer-p parent-buffer)))
```
Ambrevar, line 570: Set default value of the keyword arg to `(nosave-buffer-p parent-buffer)`, then you can remove this line.

source/buffer.lisp
```diff
@@ -823,16 +866,16 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
             (ffi-buffer-evaluate-javascript buffer (quri:url-decode (quri:uri-path url)))
             (ffi-buffer-load buffer url))))))
 
-(define-command set-url (&key new-buffer-p prefill-current-url-p)
+(define-command set-url (&key new-buffer-p prefill-current-url-p nosave-buffer-p)
   "Set the URL for the current buffer, completing with history."
-  (let ((history (minibuffer-set-url-history *browser*)))
+  (let ((nosave-buffer-p (or nosave-buffer-p
+                             (nosave-buffer-p (current-buffer))))
```
Ambrevar, line 872: Same here, you can set the default value of the keyword arg.

source/manual.lisp
```diff
@@ -274,7 +274,7 @@ existing instance instead of a separate instance that exits immediately.")
 instance must be non-nil.")
     (:p "To let know a private instance of Nyxt to load a foo.lisp script and run it's
```
Ambrevar, line 275: Ha, there is an old typo here, it should read "run its `foo'" (no apostrophe).

source/web-mode.lisp
```diff
@@ -133,6 +137,7 @@ search.")
        "f" 'follow-hint
        "F" 'follow-hint-new-buffer-focus
        "; f" 'follow-hint-new-buffer
+       ;; TODO: vim-ish keybindings for follow-hint-nosave-buffer(-focus)?
```
Ambrevar, line 140: Same as above :)

Ambrevar:
Looking great, thanks a lot @aartaka!
Could you address the nits I've mentioned here though?  Thanks!

source/web-mode.lisp
jmercouris: Just to clarify "g f"?

jmercouris:
I've addressed the keybindings.

source/buffer.lisp
jmercouris: Wow, I didn't even know it was possible to use previous values in the lambda list to set subsequent values!

jmercouris:
```
(defun x (&key y (z (+ 10 y))) (format nil "y: ~a z: ~a" y z))
```

```
CL-USER> (x :y 5)
"y: 5 z: 15"
```

seems today is my day to learn!

jmercouris:
All things addressed.

PR #1155, opened 2021-02-17
ffi-buffer-enable-sound gets a sane definition

aadcg:
Minor fix so that the behaviour matches the method's name.

--

Already discussed with @jmercouris.


Ambrevar:
Good catch, thanks!

PR #1157, opened 2021-02-17
Prompt buffer (minibuffer overhaul)

Ambrevar:
This is a rebase of #1054.

Left to be done:

- [x] Backport `test-fuzzy.lisp` to make sure we have at least the same matching
      quality as master.
	  
- [ ] Tune paging.  Have C-p recenter on suggestion?

  See if we can keep
  2eeafbac2bea481b04fce0e7e5f8dbfbdcb70acc ("prompt-buffer.lisp: remove
  element-in-view-port code"), or else revert it.
  
- [x] (Optional) make-source should return a lambda that creates
  minibuffer-source objects.  Indeed, we don't want global, non-thread-safe
  minibuffer-source object.  The source object should be instantiated with the
  minibuffer.
  
  On second though, do we really need a macro to create top-level sources?  What
  about just using define-class?  For instance

  ``` common-lisp
  (define-class minibuffer-foo-source (minibuffer-source)
    ((initial-suggestions '("foo" "bar"))
     (filter #'slow-identity-match)))
  
  (minibuffer:make
   :sources (list (make-instance 'minibuffer-foo-source)))
  ```

- [ ] I'm thinking of renaming `initializer` and `cleanup` to `constructor` and
  `destructor` respectively.  This would be consistent with modes.
  (It's done now.)

  While we are at it, we can make methods for them as suggested in #1005.  We
  could still keep the slots which would allow for instance-specific overrides.

- [x] Have `prompt-buffer` inherit from `prompter` instead of having a
  `prompter` slot.  This would make the Nyxt side of things way less verbose, it
  will also be easier for the end user to invoke the prompt buffer.

- [x] Rename the `prompter:prompter-source` class to `prompter:source`.

- [x] Contextual help display and discoverable bindings.
  `C-h b` should display the bindings in a help buffer.
	  
  Add an action which opens a nested prompt buffer with all the key
  bindings.  Pressing `return` executes the bound command in the parent
  prompt buffer.
	
- [ ] Should we have `C-h m` to explain how the prompt buffer works?
  
- [ ] VI bindings

- [x] Resume previous prompt-buffers.

- [ ] Extend support for input history.

  - `set-url2` has it, add support for all commands.
  - Make sure duplicates are not inserted.
  
- [ ] Highlight portion of matching text.

- [x] Implement `search-buffer2` command using `prompt-buffer`.

- [-] Restore multi-buffer search leveraging the new multiple source feature.
      
	  Still need to fix following across buffers.
	  
- [ ] Make meta command which allows for selecting sources, then drop into new
  prompt buffer with given sources.

- [x] Performance: `set-url2` takes a while to show up when the history is big.

- [ ] `switch-buffer2` loses the prompt focus when "following" the selection.

- [x] Cosmetic: The top of the prompt input is swallowed.

- [ ] Cosmetic: Option to not display the column names / .  For instance in "New
      URL" source.
	  
- [ ] Add `with-prompt` helper to avoid having to `(when result ...)` after each
      prompt.
	  We could even support chained prompts, like `sera:and-let*`.
	  
- [ ] Echo message when no source has any suggestion (nor do they allow for raw
      user input) instead of showing a useless prompt-buffer.
	  
	  Actually, it may be useful to show an empty prompt buffer to emphasize
      what's happening (echo messages are not always seen), and to display the
      list of sources.
	  
- [ ] Add action to sort by column.
      Would be great if we had mouse support, i.e. click on the column name to
      sort by this column in ascending / descending order.

- [ ] Add column filters.

  This one is tricky, because we want to make it both flexible and
  explorable.
  We could add actions, but it can be a bit tedious to go through multiple
  menus just to filter, say, by today's date or the +foo tag.
  
  One option would be that the action actually inserts a special syntax in
  the prompt which does the filtering.
  This special syntax could be started from `(`.
  
  Example:
  
  ```lisp
  ;; first arg is column name
  (filter "URL" (match-host "example.org") (match-port "80"))
  (filter "tags" (match "lisp" "games") (match-not "racket"))
  ```

  Could also be cool to enable column and predicate completion.
  
  When implemented, remove `tags.lisp`.
	  
- [x] Shouldn't we avoid to play with `suggestion` object from the Nyxt side?
      Then we should not have `current-suggestion`, `marked-suggestions`, etc.
      Instead we should have functions that return the values.
	  
	  Nyxt uses suggestion objects mostly to get the properties.  Can we do
      differently?
	  Actually no since each object properties can be used by filters, the user
      must be able to access them.
	  
- [x] Make sure actions support both a single selection and
      marked-suggestions when `multi-selection-p` is T.
	  
- [ ] Remove `object-string`, `object-display` if we don't need it anymore?
      Then remove `object-display.lisp`.
      What about tab-insert?
	  
- [x] Remove `fuzzy.lisp` when done.

- [ ] Remove `must-match-p`.

- [x] Use `yes-no-source` in `if-confirm`.

- [ ] Suggestion properties should be an _alist_, so that properties can have
      arbitrary names and casing.
	  I suggest un-dotted alists since they are shorter, more readable, more
      beginner-friendly.

	  So replace `(:name "foo" :desc "bar")` by 
	  `(("Name" "foo") ("My description" "bar"))`.

- [ ] Rename `properties` to `attributes` to avoid confusion with symbol
      properties and "property lists" (`plist`).

- [ ] Don't expose `suggestion-maker`, revert to `suggestion-filter-funtion` and
      rename it `attribute-function`.

- [ ] Support multiple persistent actions, and add action to choose which one to
      enable.
      See also https://github.com/emacs-helm/helm/issues/1926.

- [ ] 3 types of functions can be run from the prompt buffer:
  - actions
  - persistent actions
  - commands (bound to keys)
  
  Can we conflate these into just 2 types, e.g. should all commands be actions?
  If so, we need to away to set a bunch of default actions with their keymap.
  Inheritance?  Store actions in `prompter`?

  But isn't the job of the `prompt-buffer-mode` to deal with keymaps?  It would
  be more consistent for the user.

  We can list actions and the previous point will allow us to list persistent
  commands.
  We need a way to list commands.  `C-h b` mentioned above?
  
  Add command to list all functions?  Multi-source?
  
- [ ] Style: Refactor `update-suggestion-html` and a few other functions in `prompt-buffer.lisp`.


source/renderer-gtk.lisp
```diff
@@ -167,30 +167,6 @@ not return."
   (:export-class-name-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
 
-(define-class gtk-window ()
```
jmercouris: I'm not sure I understand, accidental deletion?

Ambrevar: The definition got accidentally duplicated during the rebase.
Indeed, on master you had moved the class defs to the top of the file.  Either Git got confused, or that was me :)

Ambrevar:
I've just pushed support for async initial suggestions.  This is great since this means that the prompt buffer will always pop up instantaneously even when sources take a while to display their initial suggestions.  No more hangs!

Ambrevar:
You can now resume prompts!  Try it out with the `resume-prompt` command!  It's awesome!

Ambrevar:
Prompt buffer history is in, try it out with `M-h`.

Search is also tested and working: see `search-buffer2` and `search-buffers2`.
Still needs some fine-tuning, but overall the prompter architecture seems
flexible enough to accommodate this! :)

Ambrevar:
@jmercouris: Last commit should be good to do!  There are a few nits to fix here
and there, but overall the API should be relatively stable from now on.

You can start converting all the minibuffers to prompt-buffer! :)
One exception: bookmark filters.  We still need to work on a generalized column
filtering syntax as described above.

Before getting started, have a look at the commits from last week until today
and let me know if you have any feedback.

libraries/prompter/prompter-source.lisp
```diff
@@ -46,7 +46,7 @@ Suitable as a `source' `suggestion-property-function'."
     (t (list :default (write-to-string object)))))
 
 (define-class suggestion ()
-  ((value nil
+  ((value nil ; TODO: Rename `data' as with the GHT?  Maybe confusing since we have `match-data'.
```
jmercouris, line 49: Yes, I would also prefer data

Ambrevar: So any better naming suggestion for `match-data` then?

jmercouris: matching-data?

Ambrevar:
I mean it's the "data" part that I don't like.

Plus the name is misleading, since we would expect that "match-data"
would contain what's matched, while really it's arbitrary data.

Ambrevar:
Shall we have the `:sources` initargs accept both a list and a source object?
I'd say yes, since-source prompts are only too common...

jmercouris:
Yes, it should accept either a list, or a single source, that would simplify a lot of code.

Ambrevar:
Done.

source/window.lisp
```diff
@@ -130,13 +128,22 @@ The handlers take the window as argument."))
...
+(defmethod prompter:object-properties ((window window))
+  (list :id (id window)
+        :active-buffer (title (active-buffer window))))
+
+(define-class window-source (prompter:source)
+  ((prompter:name "Windows")
+   (prompter:must-match-p t)
+   (prompter:multi-selection-p t)
+   (prompter:initial-suggestions (window-list))
+   (prompter:actions '(window-delete))))
```
jmercouris, line 140: @Ambrevar 

When I wrote this I realized it did not work. Then I tried to do `(lambda (i) (mapcar #'window-delete i))` and that also did not work. How should I go about mapping the action to each object?

Ambrevar: Ah, yes, this is the part of the API I forget to fix :p 
(I wrote it down but then I forgot...)

I think the fix would be in prompter.lisp: `resolve-selection`.  I suggest the following: 

- If the current source accepts multiple selections, the action is `mapcar`ed on the selections and we return the list of results.

- Otherwise the action is funcalled and we return the single result.

Is this wise though?  This means that the call site `prompt` can return either lists or non-list values depending on the current source.  This would force the user to always check if the return value is a list or not.

Another option is to always return a list.  This could be a bit counter-intuitive for prompts that return only single values.

Last option I can think of is to return two values:
- a single value (in case of multiple selections, this would be the first mark)
- a list of values (in case of single selection, this would just repeat the first value, wrapped in a list).

Thoughts?

jmercouris: I like your first suggestion the most. I think it could be enhanced by always returning two `(values x y)`, where Y is a complete list of objects matched.

jmercouris: Scratch what I just said, it makes no sense. I need to think about this a little bit.

Ambrevar: Your preference is actually my 3rd suggestion, no?

jmercouris: I was trying to combine the first and third suggestion somehow. The third suggestion is the most flexible, the first suggestion is the most simple.

jmercouris:
Do we want the :prompt argument to prompt to automatically append a colon if it is not present at?

E.g.

```
(prompt
                  :prompt "Ropen buffer(s):"
                  :sources (list (make-instance 'recent-buffer-source)))
```

AND

```
(prompt
                  :prompt "Ropen buffer(s)"
                  :sources (list (make-instance 'recent-buffer-source)))
```

would be both acceptable?

Ambrevar:

> ```
> (prompt
>                   :prompt "Ropen buffer(s):"
>                   :sources (list (make-instance 'recent-buffer-source)))
> ```
>
> AND
>
> ```
> (prompt
>                   :prompt "Ropen buffer(s)"
>                   :sources (list (make-instance 'recent-buffer-source)))
> ```

Why not, the only pitfall is if a prompt specifically needs `::`, then
we would not be able to write it... Unless we suffix the string with
`::`.  So yeah, I don't see any problem with that!

source/window.lisp
Ambrevar: I thought a little bit more about it.  I think having different return types is opening a window to let in a big flow of bugs (especially if we consider third-party extensions, user configs).

So I suggest one of the two:

- Always return a list.
  Benefit: it's simple and `prompt` calls can be `mapcar`ed.
  Drawback: For all non-multiselection prompts, the caller will have to call `(first ...)`.

- Return two values `(FIRST-SELECTION ALL-SELECTION-LIST)`.
  Benefit: Direct access to the return value for non-multiselection prompts.
  Drawback: To access ALL-SELECTION-LIST directory (such as when `mapcar`ing a `prompt` call) the caller must first wrap it with `(nth-value 1...)`.

I'm leaning towards the first option.

-

jmercouris: I concur, let's go with the first option, I think it is the most logical compromise for now.

jmercouris: Having the API attempt to DWIM may have disastrous results, therefore it is best indeed.

Ambrevar:
Any particular reason why you merged master?
I find that it makes it a bit difficult to follow the history now, I'd rather keep the history lineear on this branch.  What do you think?

Can we revert this change and, if need be, rebase against master instead?

jmercouris:
I merged master because I didn't want to rebase on you in case you were doing something. Sorry if that made things confusing! Why I merged master: I wanted to update password.lisp and didn't want to work with the old-code!

jmercouris:
How can I `(prompt ...)` without any sources just to collect raw input from the user? Is this possible?

I'm trying to replace calls like this:
```
(prompt-minibuffer
         :input-prompt argument)
```

jmercouris:
Not sure how to undo my merge. I tried to revert the merge commit, and then it gives me a prompt "Replay merges relative to parent?". Apparently this is expecting some sort of integer...

jmercouris:
I think I've got it figured out.

jmercouris:
Here is what I ended up doing:

1. make a new branch from `prompt-buffer` (titled: `prompt-buffer-changes`) just in case
2. revert the branch `prompt-buffer` to before the merge-commit
3. cherry pick my changes from `prompt-buffer-changes` to `prompt-buffer`
4. rebase master onto `prompt-buffer`
5. force push

please let me know if you see anything problematic with this approach.

Ambrevar:
To undo the merge, it's easier to 
- backup the branch
- reset to the commit before the merge.
- rebase against master.
- cherry-pick the commits you added after the merge.

You are right that you should not rebase a public branch.  But if you let me know in advance and I confirm, then you can go ahead.
Otherwise, just use the "old code" in password.lisp: the change is trivial and rebasing will be trivial as well when we are done with this branch.

jmercouris:
Aha you pretty much just wrote exactly what I did :-D

Ambrevar:

> How can I `(prompt ...)` without any sources just to collect raw input from the user? Is this possible?
>
> I'm trying to replace calls like this:
> ```
> (prompt-minibuffer
>          :input-prompt argument)
> ```

Have a look at set-url2, it has a New URL source.

I think we should have a prompter:query-input-source just like we have a
prompter:yes-or-no-source, since this is probably going to be a common
use case.

Thoughts?

jmercouris:
> How can I `(prompt ...)` without any sources just to collect raw input from the user? Is this possible? I'm trying to replace calls like this: ``` (prompt-minibuffer :input-prompt argument) ```
> Have a look at set-url2, it has a New URL source. I think we should have a prompter:query-input-source just like we have a prompter:yes-or-no-source, since this is probably going to be a common use case. Thoughts?

That's more or less what I ended up doing for now. I guess we could call it "raw-input-source" or something.

jmercouris:
I don't understand why using prompt in `command-commands.lisp` fails within `execute-extended-command`. Can you please try it and see if you can see what is happening?

Ambrevar:
I'll have a look a bit later.

Ambrevar:
> I don't understand why using prompt in `command-commands.lisp` fails within `execute-extended-command`. Can you please try it and see if you can see what is happening?

Works for me :p

Have you reloaded?
Can you share your stacktrace?

source/window.lisp
jmercouris: I've changed the API now, and I'm playing around with always returning a list. I'm not sure it is the right thing to do.

jmercouris: What I don't like is the action in jump-heading.lisp.

here is an example:

```
(define-parenscript scroll-page-to-heading (heading)
               (ps:chain heading (scroll-into-view t))))))


(define-class heading-source (prompter:source)
  ((prompter:name "Headings")
  ...
   (prompter:actions '(scroll-page-to-heading))))
```

What was so clean will now have me writing a new action/defun JUST for this specific use case (since I cannot use a lambda or wrap scroll-page-to-heading).

jmercouris: We could add a slot to prompter to determine whether to return a list or a single value, I think that would solve our problem. It would make it VERY explicit. By default it could return a list UNLESS, the user specifies to return a single value.

jmercouris: You can see the implementation of always returning a list starting with commit: 7180a95c

In commit 31231818 you can see the effects that it has on the code.

I'm not sure it is too bad. I can accept it. That said, I would really like your feedback on allowing the user to specify whether they want a single return value, or a list. I think that may solve our problem!

jmercouris: Please note, most of the prompt-buffer converted code is broken with commit 7180a95 (since the original implementation did not always return a list).

Ambrevar:
What do you think of binding `shift-return` to `return-selection-over-action` and `M-return` / `C-return` to `return-input`?

Do we still need `return-input` by the way?  Feels pretty useful now.

jmercouris:
> What do you think of binding `shift-return` to `return-selection-over-action` and `M-return` / `C-return` to `return-input`?
> 
> Do we still need `return-input` by the way? Feels pretty useful now.

What do you mean by `return-input` being useful, did you mean useless?

I personally think the bindings are OK as they are.

Ambrevar:

>> What do you think of binding `shift-return` to `return-selection-over-action` and `M-return` / `C-return` to `return-input`?
>> 
>> Do we still need `return-input` by the way? Feels pretty useful now.
>
> What do you mean by `return-input` being useful, did you mean useless?

Oops!  Indeed, I meant "useless".

> I personally think the bindings are OK as they are.

My point is that the "shift" prefix is the most "familiar" one.
For instance many GUI and web apps use "shift+return" to send text in a
chat box.

In terms of consistency, it mirrors scrolling: "space" scrolls down,
while "shift+space" reverses the behaviour and scrolls up.

If it's the same for you, then I'd rather use shift+return.

Thoughts?

source/buffer.lisp
```diff
@@ -856,60 +876,93 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
...
+  "Load a URL in a new nosave buffer."
+  (make-buffer-focus :url url :nosave-buffer-p t))
+
+(defun new-buffer-load-from-history (history-suggestion)
+  "Load a URL in a new buffer (from history)."
+  (make-buffer-focus :url (url history-suggestion)))
+
+(defun new-nosave-buffer-load-from-history (history-suggestion)
+  "Load a URL in a new buffer (from history)."
+  (make-buffer-focus :url (url history-suggestion)))
```
Ambrevar, line 895: I believe we can remove these `defuns` and use `make-command` instead now :)

source/web-mode.lisp
```diff
@@ -253,45 +253,54 @@ and to index the top of the page.")
...
+  ((prompter:name "Parent URLs")
+   (buffer :initarg :buffer :accessor buffer :initform nil)
+   (prompter:must-match-p t)
+   (prompter:constructor
+    (lambda (source)
+      (funcall (if (conservative-history-movement-p (find-mode (buffer source) 'web-mode))
+                   #'htree:all-contiguous-owned-parents
+                   #'htree:all-parents)
+               (get-data (history-path (buffer source)))))))
+  (:export-class-name-p t))
```
Ambrevar, line 266: You've remove `with-data-unsafe` here, which does not seem right.  Any specific reason?

source/buffer.lisp
```diff
@@ -941,83 +994,97 @@ this command it cycles through all buffers."
...
+                            (alex:mappend 
+                             #'modes 
+                             (uiop:ensure-list (buffers source)))
+                            :test (lambda (i y) (equal (mode-name i)
+                                                       (mode-name y)))))))
+  (:export-class-name-p t))
+
+(define-class inactive-mode-source (prompter:source)
+  ((prompter:name "Inactive modes")
+   (buffers :initarg :buffers :accessor buffers :initform nil)
```
Ambrevar, line 1033: We can avoid the verbosity with `define-class`:

```lisp
(buffers nil)
...
```

Ambrevar:
@jmercouris I think `if-confirm` belongs to `prompt-buffer.lisp`, no need to create a separate file.
Besides, we also need to convert it to use the prompter.  Note that `prompter` has a `yes-no-source`.

source/web-mode.lisp
jmercouris: I've replicated the macro manually, we don't need to access the data within a body, we just need the data.

jmercouris:
> @jmercouris I think `if-confirm` belongs to `prompt-buffer.lisp`, no need to create a separate file.
> Besides, we also need to convert it to use the prompter. Note that `prompter` has a `yes-no-source`.

It must be in a separate file. The macro is used before the prompt buffer appears. That is why the 'minibuffer-helper' file existed. I did not want to do it, I was forced into it by the compiler :-D

source/buffer.lisp
jmercouris: I don't think this is an important change, if we are inconsistent about it, it won't matter. In any case, I like being verbose sometimes. This is a feature of class-star I do not particularly agree with!

jmercouris:
> John Mercouris ***@***.***> writes:
> > What do you think of binding `shift-return` to `return-selection-over-action` and `M-return` / `C-return` to `return-input`? > > Do we still need `return-input` by the way? Feels pretty useful now. What do you mean by `return-input` being useful, did you mean useless?
> Oops! Indeed, I meant "useless".
> I personally think the bindings are OK as they are.
> My point is that the "shift" prefix is the most "familiar" one. For instance many GUI and web apps use "shift+return" to send text in a chat box. In terms of consistency, it mirrors scrolling: "space" scrolls down, while "shift+space" reverses the behaviour and scrolls up. If it's the same for you, then I'd rather use shift+return. Thoughts?

I find RET to be the most consistent. Since the 90s people use RET to input text. If we had a multiline form, I could understand the rationale of using shift+RET to input, but we have a single line form.

source/web-mode.lisp
Ambrevar: I think there is a misunderstanding: the point of `with-data-unsafe` is that we convey in English that the following code snippet cannot rely on the integrity of the data.  As a good practice, we should always use `with-data-unsafe` when we are not locking the data.

source/buffer.lisp
Ambrevar: Inconsitency is not a good thing I believe.  There is no benefit in being verbose here.

The main purpose of class* is reduced verbosity after all.

Ambrevar:
> I find RET to be the most consistent. Since the 90s people use RET to input text. If we had a multiline form, I could understand the rationale of using shift+RET to input, but we have a single line form.

I'm not saying we change RET.
I'm saying we rebind C-RET to shift-RET.

Ambrevar:
> It must be in a separate file. The macro is used before the prompt buffer appears. That is why the 'minibuffer-helper' file existed. I did not want to do it, I was forced into it by the compiler :-D

Misunderstanding?  `resume-prompt` uses the prompt and it's defined in `prompt-buffer.lisp`.

source/web-mode.lisp
jmercouris: OK! I will restore it then :-)

jmercouris:
> It must be in a separate file. The macro is used before the prompt buffer appears. That is why the 'minibuffer-helper' file existed. I did not want to do it, I was forced into it by the compiler :-D
> Misunderstanding? `resume-prompt` uses the prompt and it's defined in `prompt-buffer.lisp`.

Give it a try, if you can get it to compile, I am all in favor! :-)

jmercouris:
> I find RET to be the most consistent. Since the 90s people use RET to input text. If we had a multiline form, I could understand the rationale of using shift+RET to input, but we have a single line form.
> I'm not saying we change RET. I'm saying we rebind C-RET to shift-RET.

Aha! OK, yes that is fine :-)!

source/buffer.lisp
jmercouris: In this case whether I include these extra forms or not within a class-star defclass form, the result is the same. Therefore I do not think inconsistency here is a problem. For example, inconsistently adding docstrings to the top of a defun is not a problem, the meaning is the same.

Ambrevar: Indeed it is equivalent, but the consistency I'm talking about is in terms of style.  Consistent style helps with readability, especially for newcomers.

More specifically on this issue, adding verbosity to the code is only beneficial if there is, well, a benefit behind it.  Here I don't understand what we benefit from explicitly specifying the default properties of the slot. What do you think?

jmercouris: There is no benefit, I am just saying it doesn't matter :-D

Will it be a readability issue for newcomers? I'm not convinced! It is slightly different, but no different than CLOS. If someone is working on our codebase, I think we can assume some proficiency with CLOS!

That said, it does convey some intention. One reading the codebase might wonder why I have specifically used these slot designators. Is there some meaning here? I know I would pause and look over it and try to understand why. So, the conclusion is: it should be changed, but for different reasons than you proposed :-O :-D

Ambrevar: I'm confused :p So which reason is it then?

jmercouris: Not because it is inconsistent, but because it conveys some intention. For example, the usage of IF vs WHEN can be equivalent, but IF conveys a completely different meaning.

Ambrevar:
>> Misunderstanding? `resume-prompt` uses the prompt and it's defined in `prompt-buffer.lisp`.
>
> Give it a try, if you can get it to compile, I am all in favor! :-)

Oh, I see, it's a macro.
What about moving it to configuration.lisp then?

source/buffer.lisp
Ambrevar: So what's the intention here?

jmercouris:
> > Misunderstanding? `resume-prompt` uses the prompt and it's defined in `prompt-buffer.lisp`. Give it a try, if you can get it to compile, I am all in favor! :-)
> Oh, I see, it's a macro. What about moving it to configuration.lisp then?

OK!

Ambrevar:
What about making a higher-order-function that converts a function that takes a
single argument into a function that takes a list and calls the former function
over the first element?

Example:

``` common-lisp
; buffer-load takes 1 arg, the URL:
(funcall (unwrap-list-arg 'buffer-load) '("url1" "url2"))
; -> Calls `buffer-load' over "url1".
```

From there, we can make a helper that builds a command in the same way.
Then it would be enough to do this:

``` common-lisp
(define-class global-history-source (prompter:source)
  ((prompter:name "Global history")
    ;...
   (prompter:actions (list (make-unwrapped-command 'buffer-load)
                           (make-unwrapped-command 'make-buffer-focus)))))
```

instead of

``` common-lisp
(define-class global-history-source (prompter:source)
  ((prompter:name "Global history")
    ;...
   (prompter:actions (list (make-command buffer-load* (history-entries)
                             (buffer-load (first headings)))
                           (make-command make-buffer-focus* (history-entries)
                             (make-buffer-focus :url (first headings)))))))
```

Thoughts?

jmercouris:
That is a solution indeed. Can you show me how MAKE-COMMAND-UNWRAPPED would expand? Also what about calling it ENSURE-CAR-COMMAND (to mirror alexandria:ensure-car)?

jmercouris:
My proposed expansion:

Source:
```
(ensure-car-command 'buffer-load)
```

Expansion:
```
(define-command ensure-car-buffer-load (list)
   (buffer-load (alex:ensure-car list)))
```

Ambrevar:
And the reverse behaviour.  Currently to generate a command that works on a list
instead of just a single element, we need to use `make-command` and `mapcar` the
original function.

So we could have

``` common-lisp
(make-wrapped-command 'buffer-load)
```

instead of 

``` common-lisp
(make-command 'buffer-load** (history-entries)
  (mapcar 'buffer-load history-entries))
```

jmercouris:
I like both ideas, I just want to see the expansion :-)

jmercouris:
Final name suggestion: make-unmapped-command
                                      make-mapped-command

Ambrevar:
> Final name suggestion: make-unmapped-command
>                                       make-mapped-command

OK for `make-mapped-command`.  Not sure make-unmapped-command means much
though, but I'm OK with it.

Ambrevar:
Should we force actions to be commands?

In many ways, it makes sense: actions are interactive (in the Emacs
sense), just like commands.

Other benefits:

- We could define object-properties on actions (right now we need to use
  `prompter:suggestion-property-function` instead.).

- Commands have a `last-access` slot, so we could filter the actions by
  last access as well.
  It's only mildly useful.

- There are less commands than functions, so it's easier to search them.

jmercouris:
I don't think we should force them to be commands. I see us potentially having to wrap many defuns as commands to make them actions.

Ambrevar:
> I don't think we should force them to be commands. I see us potentially having to wrap many defuns as commands to make them actions.

Good point.  It would be a one-liner though.  Hmm... Well, we will see
when we get more experience with it.

Ambrevar:
I've been thinking about `must-match-p`, which is one of the few API points
that's not well defined.

In my understanding, we never need it.  Here are a few cases when we may need
arbitrary user inputs:

- Specify a new URL.
- Specify a new file.
- Query strings.

For the first two, using `must-match-p` on the URL history / file-listing source
means the 

For the last one, we have `pompter:raw-input`.

Currently we don't have a way to display user input among the suggestions
of the URL history / file-listing.

If we always use a dedicated source for user input, we don't need the
`return-input` (C-ret) command either.

Finally, the user input often does not support the same actions as URL history
and file-listing.

So I'm in favour of removing it.  What do you think?

jmercouris:
I see your reasoning, but let us leave it in until we are done, then we will know for sure. Can you please implement the map-/unmap command macro so that I may utilize it when updating all sources?

jmercouris:
I find myself repeating this pattern quite often:

```
(let ((url (first (prompt
                     :prompt "New distribution URL"
                     :sources (make-instance 'prompter:raw-source)))))
    (ql-dist:install-dist url :prompt nil))
```

where I do `(first (prompt ...)`, I think this means we would benefit from something like:

```
(promt
  :prompt "xyz"
  :return-single-value t ...)
```

In the end therefore we would have the following:

actions are always called with a list, we have map/unmap command.

Prompts return a list or a single value depending on what the user specifies.

What do you think?

Ambrevar:
> I see your reasoning, but let us leave it in until we are done, then we will know for sure. Can you please implement the map-/unmap command macro so that I may utilize it when updating all sources?

I'd like to update the API on the library side now, because it's a
blocker for the rest.
Note that I won't break the Nyxt source and I'll leave a useless
:must-match-p for now.  Is that OK?

Ambrevar:
> In the end therefore we would have the following:
>
> actions are always called with a list, we have map/unmap command.
>
> Prompts return a list or a single value depending on what the user specifies.
>
> What do you think?

Actually `first` is shorter than `:return-single-value` :p  There does not
seem to be much to gain here.  Do you see some other benefit?

Ambrevar:
I'll work on the map/unmap macros after I'm done with the must-match-p
and empty source testing.

jmercouris:
> I see your reasoning, but let us leave it in until we are done, then we will know for sure. Can you please implement the map-/unmap command macro so that I may utilize it when updating all sources?
> I'd like to update the API on the library side now, because it's a blocker for the rest. Note that I won't break the Nyxt source and I'll leave a useless :must-match-p for now. Is that OK?

OK!

Ambrevar:
Actually it's OK, I found a way to fix the must-match-p code in the libraries while keeping the logic.

jmercouris:
> John Mercouris ***@***.***> writes:
> In the end therefore we would have the following: actions are always called with a list, we have map/unmap command. Prompts return a list or a single value depending on what the user specifies. What do you think?
> Actually `first` is shorter than `:return-single-value` :p There does not seem to be much to gain here. Do you see some other benefit?

No real benefit, I just like how it looks, that we are returning a single value.

Ambrevar:
make-mapped-command and make-unmapped-command just pushed!
I've included 2 examples.

Now you can fix switch-buffer, set-url, etc.! :)

jmercouris:
> make-mapped-command and make-unmapped-command just pushed!
> I've included 2 examples.
> 
> Now you can fix switch-buffer, set-url, etc.! :)

Yay! I will use it on any conversions left, and then slowly backport it, but it should not be blocking. We can do it on master too.

Ambrevar:
I mean right now many many commands are broken because the actions don't
expect a list.  We need to fix them before merging to master I believe.

jmercouris:
I have been doing that :-)

Ambrevar:
Alrighty! :)

libraries/prompter/prompter.lisp
```diff
@@ -325,8 +325,8 @@ If STEPS is negative, go forward and selection first suggestion."
 If there is no marks, the current selection value is returned as a list of one element.
 For instance, if the selected element value is NIL, this returns '(NIL).
 If there is no element, NIL is returned."
-  (or (mapcar #'value (or (all-marks prompter)
-                          (uiop:ensure-list (selected-suggestion prompter))))
+  (or (or (all-marks prompter)
```
jmercouris, line 328: nested OR means that you can pop up the two subclauses to the parent clause.

Ambrevar:
Good catch, and fixed! :)

libraries/prompter/prompter.lisp
Ambrevar: Good catch, and fixed! :)

jmercouris:
All files converted. Warnings still remain where there are unused elements such as 

```
(hooks:define-hook-type minibuffer (function (minibuffer)))
```

in browser.lisp. I will try to take care of some of these tomorrow, though I also invite you to do it too :-D

Last note: I still think it would be good to offer `(prompt :return-single t)` to avoid another level of indentation by doing `(first (prompt  ... ))`.

Ambrevar:
Good early morning John!

> All files converted. Warnings still remain where there are unused elements such as 
>
> ```
> (hooks:define-hook-type minibuffer (function (minibuffer)))
> ```

I'll take care of it.

> in browser.lisp. I will try to take care of some of these tomorrow, though I also invite you to do it too :-D

Will do!

> Last note: I still think it would be good to offer `(prompt :return-single t)` to avoid another level of indentation by doing `(first (prompt  ... ))`.

I'm still not in favour.  I don't see the benefits:

- It's longer than (first (prompt ...)).
- It's something more do learn.

However, I may have a solution that will get us both happy.  If you look
at the original post in the issue (check it out again, I've updated lots
of items) you'll find this:

- [ ] Add `with-prompt` helper to avoid having to `(when result ...)` after each
      prompt.  We could even support chained prompts, like `sera:and-let*`.--8<---------------cut here---------------end--------------->8---

Indeed, we are not only repeating many `first`, but also the

(let ((result (prompt ...)))
  (when result
     ...))

pattern.

What about we restore our good old `with-result`? :)
Then we could have

;; Single result, `first` is called.
(with-result (result PROMPT-ARGS)
  ...)

;; Multiple results, `first' is NOT called.
(with-results (results PROMPT-ARGS)
  ...)

Thoughts?

Ambrevar:
Could even chain them, although that's not something we'd use very often:

```lisp
(with-results* ((results1 PROMPT-ARGS1)
                (results2 PROMPT-ARGS2-THAT-CAN-REUSE-RESULTS1)
  ...)
```

source/certificate-exception-mode.lisp
```diff
@@ -50,9 +37,10 @@ To make this change permanent, you can customize
 \(setf nyxt/certificate-exception-mode:*default-certificate-exceptions*
       '(\"nyxt.atlas.engineer\" \"example.org\"))"
   (if (find-submode buffer 'certificate-exception-mode)
-      (let ((input (prompt-minibuffer
-                    :input-prompt "URL host to exception list:"
-                    :suggestion-function (previous-history-urls-suggestion-filter))))
+      (let ((input (first (prompt
+                           :prompt "URL host to exception list:"
+                           :sources (list (make-instance 'nyxt/web-mode::history-all-source
+                                                         :buffer buffer))))))
```
Ambrevar, line 43: I think it should be `history-backwards-source`, no?

Ambrevar:
In ef5a0b5b5c6fffa9ea150620b951695f80d47d55 I've just added the option to instantiate the sources from symbols instead of objects.  `make-instance` is then automatically called.
Do you like it?

I've added more run-time validation.

I've done this because there were a few such issues were symbols were used instead of objects, like in `if-confirm`.

source/certificate-exception-mode.lisp
jmercouris: In the original implementation it just said "history" didn't specify a direction. I think better to include more than less.

jmercouris:
I'm not sold on the with-result. Let me think about it for a little bit of time. In any case, that does not block the merge of this PR. If you are OK with it, I will not touch this anymore, and leave the final merging up to you! So, feel free to rebase master etc.

source/certificate-exception-mode.lisp
Ambrevar: Maybe the original documentation wasn't clear: the reason why we prompted the URL instead of just showing the last URL is because in case of redirections, the last URL would be the wrong one for which to add the certificate.
We only need to show the URLs that were traverse for the last queries, so in other words, the parent URLs.

Parent URLs themselves include more than necessary, but that's OK since they are ordered (most recent at the top).

Showing all history URLs is both overwhelming and breaks this "most recent at the top" ordering.

Does that make sense?

jmercouris:
Ready to merge :-)

Ambrevar:
Just pushed a "fix fix", I'll have a final review tomorrow and merge!

Ambrevar:
Merged!

There are important issues before we can release a new pre-release though.  I'll open a new issue about it.
I'll also open another issue about what we've discussed here that hasn't been implemented yet.

Ambrevar:
See https://github.com/atlas-engineer/nyxt/issues/1212.

PR #1158, opened 2021-02-17
Move commands to appropriate mode files

jmercouris:
Relatively simple change, moves commands that rely on a mode, to the file where the mode itself is defined.


jmercouris:
I forgot, we had already discussed this as well on Monday...

jmercouris:
Specifically in the context of 'why is list-buffers defined in base-mode???'

source/browser.lisp
```diff
@@ -58,7 +58,6 @@ This slot is mostly meant to clean up the thread if necessary.")
    (password-interface (make-password-interface)
                        :export nil)
    (messages-content '()
-                     :export nil
```
Ambrevar, line 61: I think this should remain unexported: this is an implementation detail, we don't want the user to access it directly.

Or, at least, just define a `reader`.

Ambrevar:
Thanks!

source/browser.lisp
jmercouris: A good call, a reader it should be!

Ambrevar: Done with 12776890f0ad6312126e3c3b46bc28c0a2ddb605.

jmercouris: Thank you!

PR #1159, opened 2021-02-17
remove file-manager-mode

jmercouris:
it requires a complete rewrite and is currently non-functional.


jmercouris:
There are still errors, this pull request is just to see if you are in favor of the idea!

Ambrevar:
On the one hand, if we were to leave the file manager as is, I agree it's not a good fit for 2.0 because it's too "alpha".

On the other hand... The file manager is the perfect test bed for the new prompt buffer! :)
Indeed, with Helm the most advanced "helm" is `helm-find-files`.  

Once prompt buffer is in, I'd like to reimplement the file manager using the prompt buffer API and see if I can manage to get something as good, if not better than helm-find-files.

My suggestion: let's leave this pull request open for the time being and wait until I play with the prompt buffer.

Thoughts?

jmercouris:
I think you make a good point, it is absolutely a good candidate for the new prompt buffer. I think it would be best done in 2.1 though, because it will most likely be a far bigger project than we think :-).

I have a lot of things in mind...

Ambrevar:
OK for 2.1, but I'd like to have a proof of concept before 2.0, this
would reassure me the API is not a blocker.

jmercouris:
OK, perhaps we can develop a more minimal version?

Ambrevar:
More minimal than the current one or than helm-find-files?

I think a minimal version would entail:
- listing the files;
- collecting the file size (to display as a second column);
- adding bindings to go up/down directories;
- adding a binding/action to "open" a file externally or in the browser.

With the new prompt buffer, that would be... completely trivial! :)

By the way, here Osicat might prove critical for performance.
Well, if we remove it and need it later, we can always reintroduce it.

jmercouris:
Closing this pull request. Will reopen with a new version post 2.0.0.

PR #1160, opened 2021-02-17
watch-mode.lisp: first implementation attempt

aadcg:
This is a no-frills implementation of watch-mode, a mode that reloads buffers at regular intervals.

This can be useful in a number of ways (reload a weather web page, news, etc).

@jmercouris does this qualify as a good basis for what you had in mind?

Ideas and feedback are welcome :)

### DIFF
--- a/source/watch-mode.lisp
+++ b/source/watch-mode.lisp
@@ -1,0 +1,52 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/watch-mode
+    (:use :common-lisp :trivia :nyxt)
+  (:documentation "Mode for reloading buffers at regular time
+  intervals."))
+(in-package :nyxt/watch-mode)
+
+(define-mode watch-mode ()
+  "Reload the current buffer at regular time intervals."
+  ((thread-label (str:concat "watch-mode-" (id (current-buffer)))
+                 :export nil
+                 :documentation "Label that identifies a thread.")
+   (time-interval
+    (let* ((time-units '("days" "hours" "minutes" "seconds"))
+           (to-seconds-alist (pairlis time-units '(86400 3600 60 1)))
+           (active-time-units (prompt-minibuffer
+                               :input-prompt "Time unit(s)"
+                               :multi-selection-p t
+                               :suggestion-function (lambda (minibuffer) time-units)))
+           (times (mapcar (lambda (unit)
+                            (parse-integer
+                             (prompt-minibuffer
+                              :input-prompt (format nil "Time interval (~a)" unit)
+                              :hide-suggestion-count-p t)
+                             :junk-allowed t))
+                          active-time-units))
+           (to-seconds-multipliers
+             (mapcar
+              (lambda (elem) (cdr (assoc elem to-seconds-alist :test 'string-equal)))
+              active-time-units)))
+      (prompt-minibuffer
+       ;; TODO better way to format the string?
+       :input-prompt (format nil "Confirm refreshing every ~:@{~{~d ~}~a~:@}?"
+                             (list times active-time-units)))
+      (apply '+ (mapcar (lambda (time multiplier) (* time multiplier))
+                        times to-seconds-multipliers)))
+    :type number
+    :documentation "Time interval in seconds.")
+   (destructor
+    (lambda (mode)
+      (let ((threads (bt:all-threads)))
+        (bt:destroy-thread
+         (nth (position (thread-label mode)
+                        (mapcar 'bt:thread-name threads) :test 'string-equal)
+              threads)))))
+   (constructor
+    (lambda (mode) (bt:make-thread
+               (lambda () (loop (buffer-load (url (buffer mode)))
+                           (sleep (time-interval mode))))
+               :name (thread-label mode))))))
--- a/source/watch-mode.lisp
+++ b/source/watch-mode.lisp
@@ -7,46 +7,40 @@
   intervals."))
 (in-package :nyxt/watch-mode)
 
+(defun seconds-from-user-input ()
+  "Read from the minibuffer numerical time inputs and collate them into
+seconds."
+  (let* ((time-units '("days" "hours" "minutes" "seconds"))
+         (to-seconds-alist (pairlis time-units '(86400 3600 60 1)))
+         (active-time-units (prompt-minibuffer
+                             :input-prompt "Time unit(s)"
+                             :multi-selection-p t
+                             :suggestion-function (lambda (minibuffer)
+                                                    (fuzzy-match
+                                                     (input-buffer minibuffer)
+                                                     time-units))))
+         (times (mapcar (lambda (unit)
+                          (parse-integer
+                           (prompt-minibuffer
+                            :input-prompt (format nil "Time interval (~a)" unit)
+                            :hide-suggestion-count-p t)
+                           :junk-allowed t))
+                        active-time-units))
+         (to-seconds-multipliers
+           (mapcar
+            (lambda (elem) (cdr (assoc elem to-seconds-alist :test 'string-equal)))
+            active-time-units)))
+    (prompt-minibuffer
+     ;; TODO better way to format the string?
+     :input-prompt (format nil "Confirm refreshing every ~:@{~{~d ~}~a~:@}?"
+                           (list times active-time-units)))
+    (apply '+ (mapcar (lambda (time multiplier) (* time multiplier))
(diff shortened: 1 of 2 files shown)


source/watch-mode.lisp
```diff
@@ -0,0 +1,29 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/watch-mode
+    (:use :common-lisp :trivia :nyxt)
+  (:documentation "Mode for reloading buffers at regular time
+  intervals."))
+(in-package :nyxt/watch-mode)
+
+(defun watch-mode-thread-name (buffer) (str:concat "watch-mode-" (id buffer)))
```
jmercouris, line 10: I would to create a slot for the thread name within the watch-mode (you can still use this function to generate the thread name), and then use the slot within the destructor.

You can assign the value of the slot within your constructor.

source/watch-mode.lisp
```diff
@@ -0,0 +1,29 @@
...
+
+(define-mode watch-mode ()
+  "Reloads the current buffer at regular time intervals."
+  ((time-interval (nth-value 0 (parse-integer (prompt-minibuffer
```
jmercouris, line 14: A good first prompt, how about expanding it to ask for multiple units and then collating those into a time? Then you can display that time to the user as a final confirmation.

Example:
1. How many days?
2. How many hours?

User inputs 1, and 6. The prompt then says "Confirm refreshing every 1 day and 6 hours?"

jmercouris:
Thank you André. This is indeed a good first implementation! I left some comments with some suggestions. Please do not hesitate if something is unclear or you disagree!

source/watch-mode.lisp
aadcg: Ah! That was my first thought but overthinking took over. I was worried that slot values are mutable while the object is alive. Is there a way to prevent it or am I worrying too much?

source/watch-mode.lisp
aadcg: I thought about this. Another idea was the following: say that a user sets the timer to 1 hour. Perhaps the minibuffer could prompt the user "would you like to change to that buffer?". This would be optional of course.  Another thing, but this is a btw, is that it would be nice to have a `run-at`time` command. I would like to have that functionality so that I can toggle to `dark-mode` at sunset, etc. What do you think?

source/watch-mode.lisp
Ambrevar: The conventional way to say "don't touch this" in CLOS is:
- don't export the slot / accessors;
- (internally) don't set an accessor / writer.

As you can see, there is a granularity which depends on what you need.

source/watch-mode.lisp
```diff
@@ -0,0 +1,29 @@
...
+(uiop:define-package :nyxt/watch-mode
+    (:use :common-lisp :trivia :nyxt)
+  (:documentation "Mode for reloading buffers at regular time
+  intervals."))
+(in-package :nyxt/watch-mode)
+
+(defun watch-mode-thread-name (buffer) (str:concat "watch-mode-" (id buffer)))
+
+(define-mode watch-mode ()
+  "Reloads the current buffer at regular time intervals."
```
Ambrevar, line 13: Write the docstrings in the imperative tense, i.e. "Reload".

source/watch-mode.lisp
Ambrevar: `nth-value 0` is a no-op.  Any specific reason for using it?
Also pass `:junk-allowed t` to `parse-integer so that it does not raise a condition if the user inputs junk.

Ambrevar: Does it actually work to call the minibuffer / prompt buffer from a mode initform? O.o 
If so, then... why not!  Cool use of CLOS with the minibuffer, I had never thought of that! :)

source/watch-mode.lisp
```diff
@@ -0,0 +1,29 @@
...
+  ((time-interval (nth-value 0 (parse-integer (prompt-minibuffer
+                           :input-prompt "Time interval (seconds)")))
+            :type number
+            :documentation "Defines the regular time interval.")
```
Ambrevar, line 17: What's the unit here?  Seconds?

source/watch-mode.lisp
```diff
@@ -0,0 +1,29 @@
...
+   (destructor
+    (lambda (mode)
+      (let ((threads (bt:all-threads)))
+        (bt:destroy-thread
+         (nth (position (watch-mode-thread-name (buffer mode))
+                        (mapcar 'bt:thread-name threads) :test 'string-equal)
+                            threads)))))
```
Ambrevar, line 24: As John mentioned, it's better (and less brittle -- relying on the thread name could get you to kill an unrelated thread that happens to be named the same) to store the thread in the mode slot.  Then you have direct access to the thread you want to kill.

source/watch-mode.lisp
```diff
@@ -0,0 +1,29 @@
...
+   (constructor
+    (lambda (mode) (bt:make-thread
+               (lambda () (loop (buffer-load (url (buffer mode)))
+                           (sleep (time-interval mode))))
```
Ambrevar, line 28: Indentation seems off.

Ambrevar:
Cool stuff!  Congrats on your first mode, André! :)

I was wondering about the user experience: what if the user forgets about watch mode, then having a page automatically reload could be jarring, even look like a bug! :p 

A few thoughts:

- Maybe echo "Reload in 3...2...1..." 3 seconds before the reload.
- Alternatively, we could display the mode name as "watch-mode:3s" when only 3s are left.  Display it in bold, red, etc.
- What if the user decides last second to not reload the page?  We should give a way to easily disable watch mode.  I guess a keybinding could do.

source/watch-mode.lisp
Ambrevar: For `run-at-time`: I think there is already a CL library for that kind of things.  Maybe cl-cron or similar.  See awesome-cl for suggestions.

source/watch-mode.lisp
aadcg: I think I understood the ideas. 

In this case the `define-mode` macro exports the accessors, if I understand correctly. So, can I override that behaviour when I define the mode?

jmercouris: You may override that behavior yes, you can set `:export` to nil, see elsewhere in the codebase for examples.

source/watch-mode.lisp
aadcg: > `nth-value 0` is a no-op. Any specific reason for using it?

It only shows my ignorance ahah! All clear now :)
 
> Also pass `:junk-allowed t` to `parse-integer so that it does not raise a condition if the user inputs junk.

Right!

jmercouris: Run at time is probably outside of the scope of this mode, I suggest we keep it light and simple for now!

source/watch-mode.lisp
aadcg: Right! Makes me think of my teacher at school, who used to say "2 what? Potatoes!??"

aadcg:
> I was wondering about the user experience: what if the user forgets about watch mode, then having a page automatically reload could be jarring, even look like a bug! :p
> 
> A few thoughts:
> 
> * Maybe echo "Reload in 3...2...1..." 3 seconds before the reload.
> * Alternatively, we could display the mode name as "watch-mode:3s" when only 3s are left.  Display it in bold, red, etc.
> * What if the user decides last second to not reload the page?  We should give a way to easily disable watch mode.  I guess a keybinding could do.

Justified concerns.  I will take into account your ideas.

My plan for now is to wrap this up and write an article about it. Notice that I forced pushed changes so please take a look :)

jmercouris:
Just keep the commits, they show the iteration and development

Ambrevar:
It's good practice to squash / remove useless commits, like typos.
Commits showing different attempts can be useful to keep in history.

Ambrevar:
Is there a particular reason for storing the thread label instead of the thread itself?
Identifying threads by label is brittle in general, it's often safer to refer to objects directly.

aadcg:
> Is there a particular reason for storing the thread label instead of the thread itself?
> Identifying threads by label is brittle in general, it's often safer to refer to objects directly.

I finally understood (hopefully).

Could you take a look? 

You're both very patience. Sorry for my misunderstandings.

jmercouris:
While it is perfectly valid to initialize the thread within the slot itself, I would initialize it within the constructor personally, it is up to you of course!

jmercouris (commented):
Merged on master.

jmercouris:
Thanks again André!

PR #1165, opened 2021-02-18
remove dependency on usocket

jmercouris:
it is enough to test a hostname with iolib. that said, I do not believe we will be able to easily get rid of iolib...!

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -57,7 +57,7 @@ On errors, return URL."
               ;; support "localhost" and the current system hostname.
               ;; get-host-by-name may signal a ns-try-again-condition which is
               ;; not an error, so we can't use `ignore-errors' here.
-              (handler-case (usocket:get-host-by-name (quri:uri-host uri))
+              (handler-case (iolib/sockets:lookup-hostname (quri:uri-host uri) :ipv6 iolib/sockets:*ipv6*)
                 (t () nil)))))))
 
 (declaim (ftype (function (t) quri:uri) ensure-url))
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -28,9 +28,8 @@ On errors, return URL."
 (defun valid-url-p (url)
   ;; List of URI schemes: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
   ;; Last updated 2020-08-26.
-  (let ((valid-schemes
-          (append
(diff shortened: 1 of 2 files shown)


Ambrevar:
Good catch! I didn't know iolib could do this!

Ambrevar:
Tests are failing though, so this might not be completely equivalent.  Can you investigate?

jmercouris:
Tests are now passing. It seems that usocket was also "allowing" IP addresses. I've since added a OR clause to check if it is an IP address.

source/urls.lisp
```diff
@@ -42,23 +42,32 @@ On errors, return URL."
...
-              ;; get-host-by-name may signal a ns-try-again-condition which is
-              ;; not an error, so we can't use `ignore-errors' here.
-              (handler-case (usocket:get-host-by-name (quri:uri-host uri))
-                (t () nil)))))))
+    (flet ((hostname-found-p (name)
+             (handler-case (iolib/sockets:lookup-hostname name)
+               (t () nil)))
+           (valid-ip-address-p (address)
+             (cl-ppcre:scan
+              (cl-ppcre:create-scanner "\\d{1,3}\.\\d{1,3}\.\\d{1,3}\.\\d{1,3}")
```
Ambrevar, line 50: Note: You could have removed `create-scanner`, just keep the string.
This is wrong though: it accepts characters before and after the IP address, so you need to put the regexp between `^` and `$`.

Actually, scratch all that: there is `quri:ip-addr-p` :)

source/urls.lisp
```diff
@@ -42,23 +42,32 @@ On errors, return URL."
              "thismessage" "tip" "tn3270" "turn" "turns" "tv" "urn" "vemmi" "vnc" "ws" "wss"
              "xcon" "xcon-userid" "xmlrpc.beep" "xmlrpc.beeps" "xmpp" "z39.50r" "z39.50s")))
```
Ambrevar, line 43: While we are at it, could you let-bind the Nyxt internal scheme to an explicit name (say "nyxt-internal-schemes"), same with the IANA schemes, then remove the comment "For Nyxt's internal URLs and not-standardised ones".

source/urls.lisp
jmercouris: Thank you!

jmercouris:
Thank you for your feedback!

PR #1166, opened 2021-02-19
move quicklisp-libraries -> libraries

jmercouris:
there is no semantic difference between regular libraries and
'quicklisp libraries'

our regular libraries, when moved into separate repostories, will also
likely live as git submodules within the libraries folder

### DIFF
--- a/.github/workflows/package-ubuntu.yml
+++ b/.github/workflows/package-ubuntu.yml
@@ -37,7 +37,7 @@ jobs:
       run: |
         make quicklisp-extra-libs
         mkdir -p ~/.config/common-lisp/source-registry.conf.d/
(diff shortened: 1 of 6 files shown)


Ambrevar:
I believe this didn't work when I last tried.

I'll try again.

Ambrevar:
I haven't forgotten about this, I'll come back to it later!

Ambrevar:
This patch deletes the submodule, which breaks the build.

Could you amend the patch by moving the submodules using `git mv old/submod new/submod`?

jmercouris:
Yes, I can do so :-)

jmercouris:
Will reopen a new pull request with changes. Closing for now.

PR #1174, opened 2021-02-24
Fix for 1173, uses glib:g-set-prgname

jeffbowman:
As recommended in the issue, changed to use the
glib:g-set-prgname. The WM_CLASS after the change is reported as:

WM_CLASS(STRING) = "nyxt", "Sbcl"

### DIFF
--- a/source/renderer-gobject-gtk.lisp
+++ b/source/renderer-gobject-gtk.lisp
@@ -19,7 +19,7 @@
   (defmethod ffi-initialize ((browser gtk-browser) urls startup-timestamp)
     (log:info "Initializing Gobject-GTK Interface")
     (bt:make-thread (lambda ()
-                      (gdk:gdk-set-program-class "nyxt")
+                      (glib:g-set-prgname "nyxt")
                       (gir:invoke ((gir:ffi "Gtk") 'main)))
                     :name "main thread")
     (finalize browser urls startup-timestamp)))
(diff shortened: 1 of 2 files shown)


Ambrevar:
Thanks a lot for testing this!

According to the following example

https://gitlab.gnome.org/GNOME/gtk/-/blob/master/examples/bp/bloatpad.c

it seems that the function should be called before entering the main loop.  Can you try moving the call to before `gtk:within-main-loop`?

Also have you tried it on Wayland?

jeffbowman:
Thanks for the feedback, I'll try moving the code and update the PR.

I have not tried on Wayland, I don't typically use it, but I'll try setting up an environment and testing there as well.

Ambrevar:
Thanks a lot! :)

jeffbowman:
PR updated, also tested on Wayland, the "app_id" (which seems to be the equivalent of WM_CLASS (??)) in Wayland is set as "nyxt".  Used SwayWM for testing in Wayland, ran the command `swaymsg -t get_tree | grep app_id` to verify the value.

Ambrevar:
Excellent, thanks a lot!

Ambrevar:
Special thanks for the thorough testing on Wayland ;)

Ambrevar:
@jeffbowman A few nits:
- No need to mention the issue in the commit message.
- Make sure to use spaces for indentation.

Thanks!

jeffbowman:
oh, shoot... I thought I had Emacs configured to only use spaces... let me fix that and update the PR... my bad, I should have checked.

Ambrevar:
Don't worry, I've already merged and fixed the indentation foryou.
For next time ;)

jeffbowman:
Thanks! For next time for sure! :-)

PR #1177, opened 2021-02-26
Per buffer password interface

### LINKED ISSUE #1147: Re-add manual example to force password manager
In #1144 the example for force-setting the password manager to, say, KeepassXC when `pass` was also installed is gone.

We need to add an example showing how to set a specific password manager.

Ambrevar:
Fixes #1147 for good.

### DIFF
--- a/source/password.lisp
+++ b/source/password.lisp
@@ -28,10 +28,10 @@ for which the `executable' slot is non-nil."
       (fuzzy-match (input-buffer minibuffer) password-list))))
 
 (defun password-debug-info ()
-  (when (password-interface *browser*)
+  (alex:when-let ((interface (password-interface (current-buffer))))
     (log:debug "Password interface ~a uses executable ~s."
-               (class-name (class-of (password-interface *browser*)))
-               (password:executable (password-interface *browser*)))))
+               (class-name (class-of interface))
+               (password:executable interface))))
 
 (defun has-method-p (object generic-function)
   "Return non-nil if OBJECT is a specializer of a method of GENERIC-FUNCTION."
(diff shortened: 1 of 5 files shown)


jmercouris:
I think the changes are good. One point of uncertainty: should it be per buffer, or per browser? I think it should be per browser instance.

Ambrevar:
The rule of thumb here is that whatever we can make per-buffer, we
should make per-buffer: the more "local" the better because then we can
easily make modes to tweak the password manager settings per buffer!

Does that make sense?

jmercouris:
That makes sense, in that case, I am OK with this change! :-)

thank you monsieur!

source/password.lisp
```diff
@@ -53,44 +53,44 @@ for which the `executable' slot is non-nil."
...
-            (password:clip-password (password-interface *browser*)
+            (password:clip-password (password-interface buffer)
                                     :password-name password-name
                                     :service service)
           (error (c)
             (echo-warning "Error retrieving password: ~a" c))))
       (echo-warning "No password manager found.")))
 
-(define-command copy-password ()
+(define-command copy-password (&optional (buffer (current-buffer)))
```
jmercouris, line 85: probably should be &key, I know we use a lot of &optional, especially if we are going to use the dispatch that we talked about over email!

Ambrevar: `&optional` is not incompatible with `&key` so it would work with the mode dispatch.  No strong opinion here, plus it's super easy to switch from optional to key, it does not break anything.

Should we change all command `&optional` to `&key` in a later pull request?

jmercouris: Yes, it is not important for now, just an observation :-)

Ambrevar:
Feel free to merge then!

PR #1178, opened 2021-02-26
nyxt.desktop: group nyxt windows under the same wm-class

aadcg:
Now that the WM class is properly set for both X and Wayland we can set
it to "nyxt" so that all windows are grouped together under the same
class.

See github issue #1107 and pull requests #1141 and #1174.

--

Just is just a minor follow-up to #1174.

EDIT: I tested on GNOME wayland, and it's working fine.


Ambrevar:
Perfect, thanks!

PR #1181, opened 2021-02-27
WIP: diffs in nyxt

aadcg:
This is a first draft on implementing diff-like interfaces in Nyxt.

What's a diff?  We inevitably think of GNU diff &#x2013; a tool that compares
plain text files line by line.  Roughly, it computes the longest common
substring between the lines of each of the 2 files.

There are two lessons to take from here.  A diff is a procedure that
operates on two entities and returns nil when, within a certain meaning
of sameness, they're the same.  Also, it operates in a certain level of
granularity.  In the case of GNU diff, that level is defined to be the
line.  Newlines seem to be a sensible way to partition plain text files.

Let's think about some examples.  Programmers use diffs to compare
changes in source code.  In this case the meaning of sameness is "two
plain text files are the same when they don't differ by any characters".
The level of granularity is the line.  Notice, however, that one could
think of the following meaning of sameness "two source code files are
the same if they produce the same result".  In the case of LISP, the
following two entities would be the same.

```lisp
    (+ 1 1)
```
```lisp
    (+
     1
     1)
```

In such a case, the unit of granularity couldn't be the line anymore,
but the s-exps.  And probably newlines would need to be transformed in
spaces for obvious reasons.  Disregard some abuse of language and/or
imprecisions.

The above thought experiment suffices to understand that our idea of GNU
diff is hopeless for the purpose in Nyxt, in the context of web buffers.
In this case, the idea of sameness is "those things that leave the
visual aspect of the page invariant" and the level of granularity is the
html tag, i.e. things bounded by `<foo>` and `</foo>`.

For now, we focus on textual differences between web pages.  We
disregard differences in terms of fonts, images or other visual aspects.

The above I understood relatively fast.  I had an idea on how to
implement such a diff, but luckily I came across cl-html-diff by John
Wiseman that already does most of the job!

From now on, by diff we mean html diff.  Such a diff takes 2 strings
(i.e. tags, including nested tags) and outputs a new modified tag.  This
modified tag includes substrings enclosed by `ins` or `del` tags, which
serve the purpose of visually marking text as deleted or added.

Example:

Say our inputs are (notice that the order matters!):
`<p>Hello World!</p>`
`<p>Bye World!</p>`

then the output would like that in html:

<p><del class=\"nyxt-diff\">Hello </del><ins class=\"nyxt-diff\">Bye </ins>World!</p>

The style of the `ins` and `del` tags can be modified with the `style`
tag, and that can be included anywhere in an html file.  Such a style
can have classes so that the `ins` and `del` tags inserted by the diff
don't interfere with those already present in the original html file.

Let's talk about user interfaces.  Again, I emphasise that any tag-like
input is accepted by diff.  That means that a whole html page can be
provided, or any subset, say a paragraph.  John recently implemented a
method `ffi-buffer-get-document` that fetches the html of a given
buffer.  Perhaps the most common use case is that the user will be
prompted to choose two buffers (notice again that the order matters) and
this will spawn a new diff buffer.

Diff mode, to my mind, is the mode in which diff buffers by default are
instantiated.  Such a mode can include keybindings to easily close the
buffer (say by pressing `q`), or it could center the buffer while
navigating through diff blocks by pressing `n` and `p`.  Just as a side
note, perhaps sooner or later, we should think about narrowing buffers
(narrow to heading, paragraph, etc).

Since the order matters, I suggest using terminology such as "new" and
"old" instead of a neutral naming such as "A" and "B".

Though the html-diff as is doesn't export this method, it provides a
side-by-side view as an option.  Although I didn't find any
inconsistencies or strange behaviour, I didn't test it enough.

There should also be an interface to select html files from the user's
filesystem, and perhaps an easy way for the user to choose tags within
buffers (say we want to compare a given paragraph of a README from the
master branch against an older branch).

In general, the options are vast.  We should focus in what users most
expect and use.

I have a bunch of issues to address.  You can get a sense of them by
looking at commit e98f694a.

Feedback is welcome so that I can get this ("[damn deal](https://invidious.snopyta.org/watch?v=3H4FjCDYtcw)") done.

Here's a preview &#x2013; the diff between the master and 1.0.0 tag of nyxt's
github page.

Screenshot from 2021-02-27 19-21-11

-- 

I put myself into troubles while trying to set this up with quicklisp.
Maybe someone can help me with that.  I tried to package this into guix
but I'm having an issue.  For some reason the test asd file is being
loaded before the asd.  Pierre will help me sure!  You can take a look
at what I already did [here](https://git.sr.ht/~aadcg/aadcg-guix-channel/tree/master/item/packages/aadcg-lisp-xyz.scm).

### DIFF
--- a/source/diff-mode.lisp
+++ b/source/diff-mode.lisp
@@ -1,0 +1,77 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/diff-mode
+    (:use :common-lisp :nyxt)
+  (:documentation "Mode for displaying web-buffer diffs."))
+(in-package :nyxt/diff-mode)
+
+;; colours based on the modus-operandi theme by Protesilaos Stavrou, which
+;; follows the highest standard on accessibility
+(defvar html-diff-style
+  "<style>
+  ins
+  {background-color: #bbeabb;
+   text-decoration: none}
+  del
+  {background-color: #efcbcf;
+   text-decoration: none}
+</style>
+"
+  "TODO")
+
+;; this is the right way, but the I can't define a background-colour for the del
+;; and ins tag within the nyxt-diff-replace class
+;; (defvar html-diff-style
+;;   "<style>
+;;   .nyxt-diff-insert
+;;   {background-color: #bbeabb;
+;;    text-decoration: none}
+;;   .nyxt-diff-delete
+;;   {background-color: #efcbcf;
+;;    text-decoration: none}
+;;   .nyxt-diff-replace
+;;   {background-color: #ecdfbf;
+;;    text-decoration: none}
+;; </style>"
+;;   "TODO")
+
+;; this doesn't contemplate if buffer *diff* already exists
+;; add a proper title perhaps?  (*diff* + old buffer name + new buffer name)
+(defun make-diff-buffer (html-diff-string html-diff-style)
+  "TODO"
+  (with-current-html-buffer (buffer "*diff*" 'base-mode)
+    ;; FIXME title isn't overridden when the html contains the title tag
+    (str:concat html-diff-style
+                (if (str:contains? "<title>" html-diff-string)
+                    (ppcre:regex-replace "<title>.*</title>" html-diff-string "")
+                    html-diff-string))))
+
+(defun diff-buffers ()
+  "TODO"
+  ;; facilitates the following use case:
+  ;; the user wants to make a diff between the current-buffer and last inactive
+  ;; buffer
+  (let* ((old-html (ffi-buffer-get-document
+                    (prompt-minibuffer
+                     :input-prompt "Old buffer"
+                     :suggestion-function (buffer-suggestion-filter))))
+         (new-html (ffi-buffer-get-document
+                    (prompt-minibuffer
+                     :input-prompt "New buffer"
+                     :suggestion-function (buffer-suggestion-filter
+                                           :current-is-last-p t))))
+         (diff-html (html-diff:html-diff old-html
+                                         new-html
+                                         :insert-class "nyxt-diff-insert"
+                                         :delete-class "nyxt-diff-delete"
+                                         :replace-class "nyxt-diff-replace")))
+    (make-diff-buffer (princ diff-html) html-diff-style)))
+
+;; (define-mode diff-mode ()
+;;   "TODO"
+;;   ((new-html :documentation "TODO")
+;;    (old-html :documentation "TODO")
+;;    (default-display-diff-view :documentation "TODO")
+;;    (destructor (lambda (mode) TODO))
+;;    (constructor (lambda (mode) TODO))))
--- a/source/diff-mode.lisp
+++ b/source/diff-mode.lisp
@@ -3,75 +3,63 @@
 
 (uiop:define-package :nyxt/diff-mode
     (:use :common-lisp :nyxt)
+  (:import-from #:keymap #:define-key #:define-scheme)
   (:documentation "Mode for displaying web-buffer diffs."))
+
 (in-package :nyxt/diff-mode)
 
-;; colours based on the modus-operandi theme by Protesilaos Stavrou, which
-;; follows the highest standard on accessibility
-(defvar html-diff-style
-  "<style>
-  ins
-  {background-color: #bbeabb;
-   text-decoration: none}
-  del
-  {background-color: #efcbcf;
-   text-decoration: none}
-</style>
-"
-  "TODO")
-
-;; this is the right way, but the I can't define a background-colour for the del
-;; and ins tag within the nyxt-diff-replace class
-;; (defvar html-diff-style
-;;   "<style>
-;;   .nyxt-diff-insert
-;;   {background-color: #bbeabb;
-;;    text-decoration: none}
-;;   .nyxt-diff-delete
-;;   {background-color: #efcbcf;
-;;    text-decoration: none}
-;;   .nyxt-diff-replace
-;;   {background-color: #ecdfbf;
-;;    text-decoration: none}
-;; </style>"
-;;   "TODO")
-
-;; this doesn't contemplate if buffer *diff* already exists
-;; add a proper title perhaps?  (*diff* + old buffer name + new buffer name)
-(defun make-diff-buffer (html-diff-string html-diff-style)
-  "TODO"
-  (with-current-html-buffer (buffer "*diff*" 'base-mode)
-    ;; FIXME title isn't overridden when the html contains the title tag
-    (str:concat html-diff-style
-                (if (str:contains? "<title>" html-diff-string)
-                    (ppcre:regex-replace "<title>.*</title>" html-diff-string "")
-                    html-diff-string))))
-
-(defun diff-buffers ()
-  "TODO"
-  ;; facilitates the following use case:
-  ;; the user wants to make a diff between the current-buffer and last inactive
-  ;; buffer
-  (let* ((old-html (ffi-buffer-get-document
-                    (prompt-minibuffer
-                     :input-prompt "Old buffer"
-                     :suggestion-function (buffer-suggestion-filter))))
-         (new-html (ffi-buffer-get-document
-                    (prompt-minibuffer
-                     :input-prompt "New buffer"
-                     :suggestion-function (buffer-suggestion-filter
-                                           :current-is-last-p t))))
-         (diff-html (html-diff:html-diff old-html
-                                         new-html
-                                         :insert-class "nyxt-diff-insert"
-                                         :delete-class "nyxt-diff-delete"
-                                         :replace-class "nyxt-diff-replace")))
-    (make-diff-buffer (princ diff-html) html-diff-style)))
-
-;; (define-mode diff-mode ()
-;;   "TODO"
-;;   ((new-html :documentation "TODO")
-;;    (old-html :documentation "TODO")
-;;    (default-display-diff-view :documentation "TODO")
-;;    (destructor (lambda (mode) TODO))
-;;    (constructor (lambda (mode) TODO))))
+(define-mode diff-mode ()
+  "Diff mode is used to view the diffs between two buffers."
+  (;; (buffer (make-internal-buffer :title "*diff*"
+   ;;                               :modes '(base-mode))
+   ;;         :documentation "TODO")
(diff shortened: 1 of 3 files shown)


Ambrevar:
This looks fantastic!

I haven't looked at the details yet, but about the issue of "semantic
diff", have you seen

  https://fosdem.org/2021/schedule/event/sexpressiondiff/

?

Seems like exactly what you are trying to solve.
I haven't looked at cl-html-diff yet.  Note that it should be packaged
for Guix before we can merge this ;)

source/diff-mode.lisp
```diff
@@ -0,0 +1,77 @@
...
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/diff-mode
+    (:use :common-lisp :nyxt)
+  (:documentation "Mode for displaying web-buffer diffs."))
+(in-package :nyxt/diff-mode)
+
+;; colours based on the modus-operandi theme by Protesilaos Stavrou, which
+;; follows the highest standard on accessibility
+(defvar html-diff-style
```
jmercouris, line 11: For CSS, we always use cl-css (https://github.com/Inaimathi/cl-css). You can find some examples of it in the Nyxt codebase.

source/diff-mode.lisp
```diff
@@ -0,0 +1,77 @@
...
+;;    text-decoration: none}
+;; </style>"
+;;   "TODO")
+
+;; this doesn't contemplate if buffer *diff* already exists
+;; add a proper title perhaps?  (*diff* + old buffer name + new buffer name)
+(defun make-diff-buffer (html-diff-string html-diff-style)
+  "TODO"
+  (with-current-html-buffer (buffer "*diff*" 'base-mode)
+    ;; FIXME title isn't overridden when the html contains the title tag
```
jmercouris, line 44: You can use javascript to force the title to be something you want. Another (and probably simpler) option would be to set the title of the buffer slot, and not change it at all within the HTML.

source/diff-mode.lisp
```diff
@@ -0,0 +1,77 @@
...
+                     :suggestion-function (buffer-suggestion-filter
+                                           :current-is-last-p t))))
+         (diff-html (html-diff:html-diff old-html
+                                         new-html
+                                         :insert-class "nyxt-diff-insert"
+                                         :delete-class "nyxt-diff-delete"
+                                         :replace-class "nyxt-diff-replace")))
+    (make-diff-buffer (princ diff-html) html-diff-style)))
+
+;; (define-mode diff-mode ()
```
jmercouris, line 71: I would also move the html-diff-style here when you make this class

source/diff-mode.lisp
```diff
@@ -0,0 +1,77 @@
...
+  {background-color: #bbeabb;
+   text-decoration: none}
+  del
+  {background-color: #efcbcf;
+   text-decoration: none}
+</style>
+"
+  "TODO")
+
+;; this is the right way, but the I can't define a background-colour for the del
```
jmercouris, line 23: CSS works on the order of specificity. Most recent, most specific rules are applied. In this case you want to change the behavior of ins/del within the context of nyxt-diff-replace. I believe it would be something like this:

ins.nyxt-diff-replace

and

del.nyxt-diff-replace

jmercouris:
Very cool André. It has turned out way better than I imagined! I think that the library you've found handles the problem of diffs quite well. Should we wish to improve it, we can of course write our own.

I also agree with the enhancements of "n" and "p" for navigating diffs, but I suggest we leave those for a future pull request. We can quite easily bind "q" to close the diff though, in diff-mode, we'll just use a keymap in which "q" is bound to kill the current buffer.

I've left some rough feedback to hopefully answer some of your questions about the code!

jmercouris:
Once again, thanks so much!!

source/browser.lisp
```diff
@@ -630,17 +630,23 @@ sometimes yields the wrong result."
...
-              (ps:chain document document-element |innerHTML|)))
+    (pflet ((get-html (start end)
+                      (ps:chain document document-element |innerHTML| (slice (ps:lisp start)
+                                                                             (ps:lisp end))))
+            (get-html-length ()
+                             (ps:chain document document-element |innerHTML| length)))
       (with-current-buffer buffer
-        (get-html)))))
+        (let ((slice-size 10000))
+          (reduce #'str:concat
```
Ambrevar, line 647: Or `str:join`?

source/browser.lisp
```diff
@@ -630,17 +630,23 @@ sometimes yields the wrong result."
...
-    (pflet ((get-html ()
-              (ps:chain document document-element |innerHTML|)))
+    (pflet ((get-html (start end)
+                      (ps:chain document document-element |innerHTML| (slice (ps:lisp start)
+                                                                             (ps:lisp end))))
+            (get-html-length ()
+                             (ps:chain document document-element |innerHTML| length)))
       (with-current-buffer buffer
-        (get-html)))))
+        (let ((slice-size 10000))
```
Ambrevar, line 646: Why 10000?  Is it at random?  Maybe a small comment or a more precise variable name would clarify this.

source/diff-mode.lisp
```diff
@@ -0,0 +1,77 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/diff-mode
+    (:use :common-lisp :nyxt)
+  (:documentation "Mode for displaying web-buffer diffs."))
+(in-package :nyxt/diff-mode)
+
+;; colours based on the modus-operandi theme by Protesilaos Stavrou, which
+;; follows the highest standard on accessibility
```
Ambrevar, line 10: Style: I prefer sentences starting with a capital and ending with a period.

Ambrevar:
The screenshot is amazing! :)

source/browser.lisp
jmercouris: It is relatively random, it is also a relatively reliable value :-)

source/browser.lisp
jmercouris: 2 2 2 2 2 2 2 2 2 2

aadcg:
> [...] about the issue of "semantic diff", have you seen https://fosdem.org/2021/schedule/event/sexpressiondiff/ ? Seems like exactly what you are trying to solve. 

So people are indeed thinking about that!  It seemed strange that I was the only silly guy thinking about that.  And it seems to be less easy that it sounded. Will have a closer look later. Thanks for sharing!

aadcg:
Let me tell you how things stand at the moment.  I've been struggling to
make sense of the architecture of diff-mode, in the sense that I must
make it fit with the way `define-mode` was devised.

The first observation is that the aforementioned macro creates a command
available for the users in the minibuffer.  But what does it mean to
trigger diff-mode in a given buffer?  To my mind, diff-mode should be
activated in the newly created diff buffer, not in current-buffer.

As things stand at commit eb387af4, that can be achieved by running:

```
(diff-mode :buffer (make-internal-buffer :modes '(base-mode)))
```

Within the diff buffer, one can call diff-mode again to get rid of it.

The issue is that, as things stand today, there's no way to achieve it
interactively -- only pragmatically, as shown above.

It seems justified to me that modes should be able to control where
they're being activated interactively.  I'm not saying that users should
choose at the minibuffer where the mode will be activated.  It should be
easy, within the definition of modes, to override the behaviour of
current-buffer.  You can find at commit eb387af4 that I naively added
a buffer slot.

I know I could define a function that makes an instance of diff-mode and
pass it a buffer and so on.  But again, the diff-mode minibuffer command
must have some meaning...

I could also show the diff in the current-buffer and open the buffer
that had the focus somewhere else.  But it seems hacky.

Let me know what you think.

I still need to implement a smart way for the users to pass buffers,
files, or both.

aadcg:
@jmercouris, you can have a look when you're ready. Most of the things commented could use your feedback. Thanks.

jmercouris:
Only two minor things, not exactly even necessary, but they are my comments about how I would make things perfect :-)!

Well done André!

source/diff-mode.lisp
```diff
@@ -0,0 +1,83 @@
...
+                   :text-decoration "none"
+                   :background-color "#efcbcf")))
+                 :documentation "Diff colours for its visual representation.
+They're based on the modus-operandi theme by Protesilaos Stavrou, which follows
+the highest standard on accessibility.")
+   (keymap-scheme (define-scheme "diff"
+                    scheme:cua
+                    (list "q" 'delete-current-buffer))
+                  :type keymap:scheme)
+   (destructor (lambda (instance) (nyxt::buffer-delete (buffer instance))))
```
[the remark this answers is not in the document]
aadcg: There 2 separate things here.

- creating an instance of diff-mode;
- triggering the mode in a given buffer (with the "toggle").

In the future, the toggle will not be accessible interactively, only
programmatically.  Anyway, there should be a meaningful action
associated with running `diff-mode` (the "toggle") programmatically when
diff-mode is already active in a given buffer.  It could be argued that
there's no need to get rid of the buffer.  Indeed, the mode could simply
be turned on and off.  But let's explore the symmetry of the problem.
The mode is triggered by passing an (internal) buffer, so disabling it
should get rid of that buffer.  Simply composing calls of the toggle (by
passing an existing buffer where diff-mode is active) would turn the
mode on and off.  But that is completely useless and makes little sense.

In general I wanted, perhaps with little success, to distinguish between:

- the data associated a given diff (the slots of a given instance);
- its visual representation in a buffer.

source/diff-mode.lisp
```diff
@@ -0,0 +1,83 @@
...
+                   (buffer instance))
+                  ;; setting the title the following way doesn't work
+                  ;; (setf (title (buffer instance)) "*diff*")
+                  ;; (set-window-title)
+                  ;; a temporary fix below
+                  (ffi-buffer-evaluate-javascript
+                   (buffer instance)
+                   (ps:ps (setf (ps:chain document title) "*diff*")))))))
+
+(defun prompt-old-html ()
```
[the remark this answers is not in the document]
aadcg: Indeed. I was thinking ahead because I made some drafts on how the future interface would look like when `file-manager-mode` would be factored in. I will make this change.

aadcg:
Looking good from my side.

Ambrevar:
OK to merge then?

aadcg:
> OK to merge then?

From my side, yes :)

Ambrevar:
@jmercouris: Feel free to merge!

jmercouris:
thank you very much for your persistence in following through!

Issue #1117, opened 2021-02-02
Global history polishing


Ambrevar:
Once #1110 is merged, we will have the following points to address:

1. [X] Replace some uses of `open-urls` with `make-buffer`.  Then we can remove the `parent-buffer` argument from `open-urls`.
2. [X] Prompt the user for the history file to restore on startup.
3. [X] Delete `session.lisp` if there is nothing else to use in there.
4. [x] Make sure history data access is thread safe (use locks properly with `with-data-access` macro and friends.)
5. [X] Restore last current buffer on startup.
6. [ ] Restore history tree of dead buffers.
  This probably requires the storage of (new) nodes in a `buffer` slot.  We could then implement a helper in `htree` to add all the children given by the nodes passed in argument.

- [ ] Can we inherit history-entry from entry?
  Without it, imported file is 6.4M.
  3.3M with deduplication fix.

  2.3M with obj deduplication fix.

Ambrevar:
@aartaka Interested by some of the above tasks?

aartaka:
1, 2, 4, 5 — definitely! Other ones — can't yet wrap my head around these. Maybe you'll be better at completing 3 and 6 :)

Ambrevar:
6 I understand.  Out of curiosity, what puzzles you about 3?

jmercouris:
I would also like to add: replace binding class with struct.

Ambrevar:
John Mercouris <notifications@github.com> writes:

> I would also like to add: replace binding class with struct.

I tried to do it, but in the end it resulted in little benefits (for
quite a few changes).

Instead, I've unexported the binding class, which is what we want here I
believe: a small structure that's just an implementation detail.

At this point I find that the difference between structs and classes in
Common Lisp is mostly due to some weird (historical) idiosyncrasies in
the standard.

`define-class` fully supersedes structs, and the `export` options give
us the fine control we want.

Thoughts?

Ambrevar:
7\. (Optional, only after next pre-release) Remove `value` slot from
`entry` class and force user to subclass `entry` instead.

This involves quite a few changes to history-tree.
Benefits:

- One level of indirection less.
- Easier to navigate the inspector :)
- A serialized file that's 50% smaller.
- Possibly faster serialization (I haven't done any benchmark though).

aartaka:
> 6 I understand. Out of curiosity, what puzzles you about 3?

[...]

Issue #958, opened 2020-09-17
Default session pollution when using several sessions


aartaka:
Now that #937 is merged, we have a cool ability to store sessions to the user-named files and restore them at will. The downside is that the default session gets cluttered with histories of other sessions. This can be undesirable for people that rely on their session data and don't want other domains/contexts/sessions to pollute it.

Our options are:
- [Doing nothing](https://github.com/atlas-engineer/nyxt/pull/937#issuecomment-689510145)
- [Redefining default session](https://github.com/atlas-engineer/nyxt/pull/937#issuecomment-690132519)
- Something else?

Thoughts?

Ambrevar:
Let's focus on the use cases.

The way I see it, "sessions" represent separate navigation data.
For instance:

- When I start Nyxt with the default session, I have a bunch of random
  buffers.

- When I start Nyxt with my "work" session, I only want the buffers that
  are related to work.

I can't see why a user would want to start Nyxt and merge the "work"
buffers into "default".

When I start Nyxt with my work buffers and do some navigation, I expect
that this will be saved in the "work" session.

Same for the default session: it should not contain "work" buffers not
alter the "work" session.

Can you think of other use cases?

jmercouris:
Sessions become more akin to perspectives/projects in projectile then. You could imagine that someone saves a session file with every project they are working on. Perhaps even members of a team could store their session files in a repository folder for the project for better collaboration.

Ambrevar:
Related issue: By detault (and on first start) Nyxt will ask to restore
the session, even if there is none.

I suggest to turn restore-session-by-name to a no-op when
session-name-suggestion-filter is empty.

aadcg:
No longer makes sense after c3e5d897c3752774668660343bb8216d200e1dc8.

PR #1183, opened 2021-03-01
Restore history safely

aartaka:
This ticks a bunch of boxes in #1117, namely:
- Last opened buffer is restored, and
- Named sessions are restored at startup if user wishes them to.

# Things To Discuss:
- Do we need to prompt the user on `Restore session?`, actually? Maybe it's simpler and faster to just present a list of sessions to pick from, as we did before?

# How Has This Been Tested
- I've ran Nyxt for some time. You can see quite a complex `htree:set-current-owner` in `restore`. It's there because I've been getting errors like "owner with identifier X does not exist" because of outdated `htree:current-owner-id`. Not sure errors are gone, though ¯\\\_(ツ)\_/¯

How does it look to you?

### DIFF
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -313,7 +313,7 @@ This is deprecated.
 We keep this variable as a means to import the old format to the new one.")
 
 (defmethod restore ((profile data-profile) (path history-data-path)
-                    &key &allow-other-keys)
+                    &key restore-buffers-p &allow-other-keys)
   "Restore the global/buffer-local history and session from the PATH."
   (labels ((restore-buffers (history)
              "For each owner, make buffer, swap owner identifier for buffer id.
@@ -344,6 +344,10 @@ We keep this variable as a means to import the old format to the new one.")
                           (setf (htree:creator-id owner)
                                 (gethash (htree:creator-id owner) old-id->new-id)))
                         (htree:owners history))
+               ;; Current owner can be outdated.
+               (sera:and-let* ((new-id (gethash (htree:current-owner-id history) old-id->new-id)))
+                 (when (htree:owner history new-id)
+                   (htree:set-current-owner history new-id)))
                (setf (htree:owners history) new-owners))
              ;; TODO: Focus last buffer.
              ;; (let ((latest-id (id (htree:current
@@ -361,13 +365,8 @@ We keep this variable as a means to import the old format to the new one.")
                    (hash-table-count (htree:entries history))
                    (expand-path path))
              (setf (get-data path) history)
-             ;; REVIEW: Does it really belong to the data restoration?
-             ;; Maybe move back to the startup function?
-             (match (session-restore-prompt *browser*)
-               (:always-ask (if-confirm ("Restore session?")
-                                        (restore-buffers history)))
-               (:always-restore (restore-buffers history))
-               (:never-restore (log:info "Not restoring session."))))
+             (when restore-buffers-p
+               (restore-buffers history)))
 
            (restore-flat-history (old-history old-path)
              (echo "Importing deprecated global history of ~a URLs from ~s."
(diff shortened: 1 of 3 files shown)


source/history.lisp
```diff
@@ -446,9 +449,10 @@ Useful for session snapshots, as `restore-history-bu-name' will restore opened b
...
-    (restore (data-profile (current-buffer)) path)
-    ;; TODO: Maybe modify `history-path' of all the buffers instead of polluting history?
-    (setf (get-data (history-path (current-buffer))) (get-data path))))
+    (let ((old-buffers (buffer-list)))
+      (setf (get-data path) (make-history-tree))
+      (restore (data-profile (current-buffer)) path :restore-buffers-p t)
+      ;; TODO: Maybe modify `history-path' of all the buffers instead of polluting history?
+      (setf (get-data (history-path (current-buffer))) (get-data path))
+      (dolist (buffer old-buffers)
+        (buffer-delete buffer)))))
```
Ambrevar, line 452: Don't delete old buffers if restoration failed?

source/history.lisp
```diff
@@ -344,6 +344,14 @@ We keep this variable as a means to import the old format to the new one.")
                           (setf (htree:creator-id owner)
                                 (gethash (htree:creator-id owner) old-id->new-id)))
                         (htree:owners history))
+               (htree:set-current-owner
+                ;; Current owner can be outdated.
+                history (or (sera:and-let* ((new-id (gethash (htree:current-owner-id history)
+                                                             old-id->new-id))
+                                            (new-owner (htree:owner history new-id)))
+                              new-id)
```
Ambrevar, line 352: If you are not using `new-owner`, maybe use `when` instead.

source/history.lisp
```diff
@@ -344,6 +344,14 @@ We keep this variable as a means to import the old format to the new one.")
...
+                            ;; But new owner can be nil.
+                            (htree:current-owner-id history)))
```
Ambrevar, line 354: This is setting the owner to it's current value, so it's a no-op, no?

source/start.lisp
```diff
@@ -290,8 +290,14 @@ To change the default buffer, e.g. set it to a given URL:
       (let ((buffer (current-buffer)))
         ;; Restore session before opening command line URLs, otherwise it will
         ;; reset the session with the new URLs.
-        ;; TODO: Select which history file to load.
-        (get-user-data (data-profile buffer) (history-path buffer))
+        (match (session-restore-prompt *browser*)
+          (:always-ask (if-confirm ("Restore session?")
+                                   (restore-history-by-name)))
```
Ambrevar, line 295: Can we remove the `if-confirm` if we are already asking which session to restore?

source/history.lisp
aartaka: How do we check that? `restore` has a `with-muffled-body` in it, preventing us from catching the condition... 

Checking the buffer list is not an option either -- One can `restore-history-by-name` for history entries only.

Let's leave it as it is.

source/history.lisp
aartaka: It's no-op only in case there's no new owner to set. So what this says is "if there is a new owner to set, based on restored buffers, set it. Otherwise, do nothing. 

Thanks for asking -- you've made me think about the necessity of such an expression! I think that a simple `when` will actually suffice there.

source/start.lisp
aartaka: Yes! That sounds good to me :)

aartaka:
All addressed, feel free to merge :)

source/history.lisp
Ambrevar: What about simply removing `with-muffled-body` from `restore`?  Or return a meaningful value: the history or nil on failure.
This operation is inherently prone to errors, so it's important to handle this well.

aartaka: Yes, that's fair. See 60830aad for a fix.

aartaka:
I believe it's all adressed now. `restore-history-by-name` now looks at new history and deletes buffers only if new history is non-nil.

Ambrevar:
Just tested:

- I think we should only list files with the `lisp` extensions.  Currently, with `default.lisp.backup1` and `default.lisp.backup2` the prompt lists `default.lisp` twice, which is not very helpful.

- Just to confirm: the history is loaded from the designated path but _still_ save to the user-configured history path (`default.lisp` by default).  Is this the expected behaviour?  If so, I think we should underline this in the documentation and maybe in an echo message.

aartaka:
1. Yes, name listing is primitive... I'll try filtering these.
2. There's already an issue about that: #958. But yes, documentation should be clearer on that. I'll add the note to the `restore-history-by-name` docstring.

aartaka:
Done: 
- Names are filtered.
- Docstring warns about default session pollution.

Ambrevar:
I've editted the docstring in d92afd94b490f9c2e8feaa8f2263b751eef2d05e to highlight which file gets overwritten.  Let me know if you don't like it, we can keep tweaking it! :)

PR #1188, opened 2021-03-03
Remove osicat

Ambrevar:
This specializes some code to use SBCL extensions instead of Osicat.  It still falls back to Osicat when not using SBCL.

 This allows us to get rid of Osicat, which is a bit harder to package than the rest because the `.so` is generated within the Osicat package.

@jmercouris Can you test to make sure I didn't break it for you?

### DIFF
--- a/build-scripts/guix.scm
+++ b/build-scripts/guix.scm
@@ -179,7 +179,7 @@
        ("mk-string-metrics" ,cl-mk-string-metrics)
        ("moptilities" ,cl-moptilities)
        ("named-readtables" ,cl-named-readtables)
-       ("osicat" ,cl-osicat)
+       ;; ("osicat" ,cl-osicat) ; Not needed for SBCL.
        ("parenscript" ,cl-parenscript)
        ("plump" ,cl-plump)
        ("quri" ,cl-quri)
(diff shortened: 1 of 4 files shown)


jmercouris:
I can't really test on macOS anymore. I wonder if we should make a 'trivial-readlink' library... do you think it is justified?

Ambrevar:
Too trivial maybe.
What about adding it to file-attributes?  We can ask upstream.

Ambrevar:
Why can't you test on macOS?
If we can't, shall we drop official support then?

jmercouris:
I don't have my macbook anymore. We should consider dropping official support yes. Just getting Nyxt to run on there is a huge challenge in of itself..

Ambrevar:
Wanna send a patch to update the doc then? :p

jmercouris:
Maybe... I'm still not sold on dropping support. Nyxt /should/ continue to work on macOS.

Ambrevar:
I'm confused.  Are you saying you changed your mind after the previous
comment and want to keep macOS support?

If so, how?

jmercouris:
I'm saying we don't need to drop it if it ain't broken 🙃

Ambrevar:
I meant updating the documentation / home page to mention that macOS is
not tested or something like that.

jmercouris:
OK, I can do that.

Ambrevar:
Shall we merge this then?

jmercouris:
Yes

jmercouris:
thank you!

PR #1189, opened 2021-03-04
Enable automatic WebKit extensions loading.

aartaka:
This adds an ability to load arbitrary [WebKitWebExtension](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebExtension.html)s at `gtk-buffer` creation. Use-case: [bloKit](https://github.com/dudik/blockit).

# Things To Discuss:

- Maybe we should load extensions in all buffers, including `nosave` ones? For now, there's an overloaded `expand-path` method for `gtk-extensions-data-path` and `nosave-data-profile` that returns nil, so `nosave` buffers are extension-less.
- Is `gtk-extensions-data-path` a good name? Maybe `webkit-extensions-data-path`? Use of the `gtk-` prefix is consistent with all the renderer-gtk.lisp, though

# How Has This Been Tested
- Compiles.
- Runs.
- [test-extension from `cl-webkit`](https://github.com/joachifm/cl-webkit/tree/master/demo/test-extensions) loads and prints the "Page X created for (null)" message in the inferior lisp.

How does it look to you?

### DIFF
--- a/source/renderer-gtk.lisp
+++ b/source/renderer-gtk.lisp
@@ -59,7 +59,10 @@ want to change the behaviour of modifiers, for instance swap 'control' and
    (data-manager-path (make-instance 'data-manager-data-path
                                      :dirname (uiop:xdg-cache-home +data-root+))
                       :documentation "Directory in which the WebKitGTK
-data-manager will store the data separately for each buffer."))
+data-manager will store the data separately for each buffer.")
+   (gtk-extensions-path (make-instance 'gtk-extensions-data-path
+                                       :dirname (uiop:xdg-config-home +data-root+ "extensions"))
+                        :documentation "Directory to store the WebKit-specific extensions in."))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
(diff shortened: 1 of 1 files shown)


Ambrevar:
Looks good to me!

About the nosave data profile, I don't know.  Can the user control which extensions gets enabled, or are they necessarily loaded?

@jmercouris Thoughts on this and on the extension path?

aartaka:
> About the nosave data profile, I don't know. 

Well, the current behavior is consistent with the rest of the code, so let's stick to it.

> Can the user control which extensions gets enabled, or are they necessarily loaded?

All of the extensions in the directory are loaded. It's probably possible to make it configurable, but that's not trivial and will require some C code :)

jmercouris:
I am OK with it :-)

thank you Artyom!

Ambrevar:
If WebKitGTK would take multiple directories, then we could easily
choose which extension to load based on the directories passed to the
entry point.

jmercouris:
Or extensions could be symlinked in and out of the directory if it is not possible to pass in a list of directories

Ambrevar:
Indeed!

aartaka:
> If WebKitGTK would take multiple directories, then we could easily choose which extension to load based on the directories passed to the entry point.

There's not such feature in WebKitGTK, as far as I know.

PR #1193, opened 2021-03-06
Replace chmod call by file-attributes lib.

### LINKED ISSUE #1184: Replace chmod call  by file-attributes
I just discovered the https://shinmera.github.io/file-attributes/ library.
It allows us to set the permissions.  This would remove the need to call `chmod` in `set-socket-permissions`.

@jmercouris What do you think?  Does file-attributes look good for distribution?
It's working on Guix, I'm asking since you had issues with Osicat on macOS.

Maybe test the lib on macOS?

Ambrevar:
Fixes #1184.

This introduces a new CL dependency, https://shinmera.github.io/file-attributes/.

@jmercouris Let me know if you can build this.

### DIFF
--- a/build-scripts/guix.scm
+++ b/build-scripts/guix.scm
@@ -105,14 +105,6 @@
            (lambda _
              (setenv "NYXT_VERSION" ,version)
              #t))
-         (add-before 'build 'fix-paths
-           (lambda* (#:key inputs #:allow-other-keys)
-             (substitute* "source/global.lisp"
-               (("\\(defvar \\*chmod\\-command\\* \"chmod\")")
-                (string-append "(defvar *chmod-command* \""
-                               (assoc-ref inputs "coreutils")
-                               "/bin/chmod\")")))
-             #t))
          (add-before 'check 'configure-tests
            (lambda _
              (setenv "NYXT_TESTS_NO_NETWORK" "1")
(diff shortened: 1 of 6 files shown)


jmercouris:
Nice, thank you!

I tried it, and I always get:

```
(nyxt:start)
<ERROR> [20:14:06] Could not bind socket /home/jmercouris/.local/share/nyxt/nyxt.socket, non-socket file exists.
```

unless I manually delete the nyxt.socket file. On another note, setting:

```
(setf *socket-path* nil)
```
in my init file no longer has any effect. This makes it quite challenge to run Nyxt while developing Nyxt at the same time! :-O

jmercouris:
Hm, it seems I have the socket problem also on master...

jmercouris:
The problem with this loading is in start-browser, the socket is queried BEFORE the user init is loaded.

jmercouris:
Right, seems I must use the --no-socket configuration. I see why it is done, it improves the performance of URLs loaded in single instance mode. Not sure I like that trade, but alea iacta est! :-D

jmercouris:
That said, on latest master no socket does not work. I tried via a command line argument, and via the REPL (`(nyxt:start :no-socket t)`) . Nothing happens. Any idea why?

jmercouris:
I've fixed the --no-socket configuration issue on master.

PR #1194, opened 2021-03-08
Fix get user data

Ambrevar:
@aartaka I've defined a default method for `get-user-data`.  Can you  confirm this is OK?
Or did you really intend to not define the default?

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -200,7 +200,7 @@ editor executable."))
               (restore profile path))
             user-data)))))
 
-(defmethod get-user-data ((profile default-data-profile) (path data-path))
+(defmethod get-user-data ((profile data-profile) (path data-path))
   (%get-user-data profile path (user-data-cache *browser*)))
 
 (defmethod get-containing-window-for-buffer ((buffer buffer)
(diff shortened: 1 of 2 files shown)


Ambrevar:
@jmercouris @aadcg This is what we need to get the "dev" data profile to work.
To repeat what's in the manual, add this to your init.lisp:

```lisp
(define-class dev-data-profile (data-profile)
  ((name :initform "dev"))
  (:documentation "Development profile."))

(defmethod nyxt:expand-data-path ((profile dev-data-profile) (path data-path))
  "Persist data to /tmp/nyxt/."
  (expand-default-path (make-instance (class-name (class-of path))
                                      :basename (basename path)
                                      :dirname "/tmp/nyxt/")))
```

Then start Nyxt from the REPL with `(nyxt:start :socket "/tmp/nyxt.socket" :data-profile "dev")`.

aartaka:
> @aartaka I've defined a default method for `get-user-data`. Can you confirm this is OK?
> Or did you really intend to not define the default?

It is all okay :)

Ambrevar:
Glad I could help!
The "dev" workflow had been broken since September...  We really need better testing for our startup process.  I'll open an issue about it.

PR #1195, opened 2021-03-08
Properly set default package to :nyxt-user.

Ambrevar:
I've finally understood how to properly set the global `*package*`.

I'm sending this as a pull request because I'd like anyone who's doing some funny business with their init file / remote execution / Lisp REPL to test this and see if it behaves as expected.

Thoughts? @jmercouris @aadcg @aartaka


aadcg:
I don't follow everything that is going on here, but I'll let you know if smth behaves unexpectedly on my end.

Ambrevar:
If you run `lisp-repl` and type `*package*` there, it should return the
:nyxt-user package.

If you've never expected some default package from Nyxt, then this
change shouldn't bother you much.

jmercouris:
I see, thank you!

Issue #1197, opened 2021-03-09
Issue with libfixposix0


aadcg:
There seems to be an issue with this debian package.  Any ideas?

```
2021-03-09T08:33:03.9908129Z ##[group]Run sudo apt-get update -qq
2021-03-09T08:33:03.9908827Z [36;1msudo apt-get update -qq[0m
2021-03-09T08:33:03.9909984Z [36;1msudo apt-get install -y libfixposix0 libfixposix-dev libwebkit2gtk-4.0-dev glib-networking gsettings-desktop-schemas xclip notify-osd enchant[0m
2021-03-09T08:33:03.9959683Z shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
2021-03-09T08:33:03.9960227Z ##[endgroup]
2021-03-09T08:33:13.6994230Z Reading package lists...
2021-03-09T08:33:13.8675847Z Building dependency tree...
2021-03-09T08:33:13.8702233Z Reading state information...
2021-03-09T08:33:14.8992827Z E: Unable to locate package libfixposix0
```

Ambrevar:
Isn't it `libfixposix`, or `libfixposix-dev`?

aadcg:
@Ambrevar it seems you made a mistake in commit 5d408fc848c136a3f42af8709829b38b86dc2d23 then.

I'm raising this issue because I was pushing commits to a pull request and I was getting emails from the CI job saying that it failed. What I pasted above was a snippet from the log file.

But then I don't understand why this is a problem only now... Your commit is from November, so it has been like that for ages.

I am indeed using the guix system!

Ambrevar:
Looks like an error with the pipeline itself, try re-running the job see
if the error goes.

aadcg:
I tried and it didn't work. Will try again in the next days.

Ambrevar:
Can you try updating the libfixposix dependencies in `.github/workflows/tests.yml`?

PR #1200, opened 2021-03-09
tests.yml: update libfixposix version.

aadcg:
Solves #1197

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,7 +36,7 @@ jobs:
       shell: bash
       run: |
         sudo apt-get update -qq
-        sudo apt-get install -y libfixposix0 libfixposix-dev libwebkit2gtk-4.0-dev glib-networking gsettings-desktop-schemas xclip notify-osd enchant
+        sudo apt-get install -y libfixposix3 libfixposix-dev libwebkit2gtk-4.0-dev glib-networking gsettings-desktop-schemas xclip notify-osd enchant
 
     - name: Install Roswell
       shell: bash


Ambrevar:
The CI uses ubuntu-latest, not Debian.

GitHub is in the process of updating ubuntu-latest to 20.04, maybe they've just rolled out the update.

aadcg:
> The CI uses ubuntu-latest, not Debian.

I thought it doesn't matter since both use apt. Ubuntu-latest must use the stable apt "channel" or whatever it's called.

I don't have permissions to re-run this job, but could you please do it for me?

Ambrevar:
It matters: APT is only the program, while the distribution repository
are a whole different thing.

Packages that are in Ubuntu are not necessarily in Debian and vice
versa.

Also distribution version matters.

aadcg:
The only thing happening here is that the github runners moved to
a newer ubuntu version and the packages of apt changed.  When you
wrote `tests.yml` it was called `libfixposix0`, but now it's
called `libfixposix3`.  Why?  No idea.

Keep in mind that my understanding is very shallow but here's
what I think.

It doesn't seem to make much sense to test on ubuntu everytime a
commit lands on master.  That would make sense to happen on every
release (perhaps only the proper ones, not the pre-releases) tho.
It is when you do a release that you want to exhaust all of the
possible combinations to make sure stuff works.  This is what you
call a "matrix".

Now, it also makes sense to test when commits land on master.
But I'd stick to guix and nix for that.  Assuming that the cl
build system will be fixed in nix soon, we could even forget
about guix since it would ensure that we have a healthy nyxt on
gnu/linux and macOS.  Well, only if it weren't for our
fervant (but JUSTIFIED) beliefs that guix is orders of magniture
better than nix.  I see guix today as a way to have "consensus"
among nyxt devs.  I don't see why we should care to have
a "multi-platform" dev environment.  Guix even allows us to build
with and without quicklisp.

As a rule of thumb we should think about dev, test and production
envs.  For dev I'd make it as minimal and dumb as possible, but
highly reliable and dependable (in other words, guix).  For test
we need to care about more things like testing on ubuntu on the
current latest release, possibly try to see how it behaves with
other cl compilers, etc.  For production, well, unless we'll
start to have some kind of NaaS (Nyxt as a service) running on
the cloud, I don't see the point.

Hopefully I didn't say anything too stupid :)

aadcg:
> This is expected indeed, I just don't see why it previously failed for
you, then stopped failing.

Perhaps I wasn't clear. It seems that the github runners update took effect first on my fork, and later on this repository. I made this change and the job run succesfully on my fork, but then it failed when I started this PR. Then, I asked you to re-run and it passed!

> We test on Ubuntu because it's been the easiest route to deploy a test
pipeline since it's what GitHub actions offer.

I see. I'm not sure about pinning a version though. This should me merged ASAP so that we stop getting annoying e-mails :)

> A few months ago Guix didn't run well in Docker for pipelines.  There's
been some work done since then, I need to retry.

I see. Can I be of help?

Ambrevar:
> Perhaps I wasn't clear. It seems that the github runners update took effect first on my fork, and later on this repository. I made this change and the job run succesfully on my fork, but then it failed when I started this PR. Then, I asked you to re-run and it passed!

Oh, OK, yes that would make sense.

> I see. I'm not sure about pinning a version though. This should me merged ASAP so that we stop getting annoying e-mails :)

Done.

>> A few months ago Guix didn't run well in Docker for pipelines.  There's
> been some work done since then, I need to retry.
>
> I see. Can I be of help?

If you can set a GitHub action to run Guix and build Nyxt, that'd be grand!

PR #1201, opened 2021-03-09
Command overhaul

### LINKED ISSUE #1192: Allow on-the-fly command definition (like hook handlers)
Commands are very much like hook handlers (from the Serapeum library).

We don't want to allow users to bind keys / prompt buffer actions to lambdas or function objects, because:

- Lambdas are unnamed blackboxes and are hard to manipulated (e.g. can't really list them with meaningful names).
- When a function is updated, the bound function object still refers to the old function.

But hook handlers don't suffer from this problem since they can wrap around lambdas which can use function _symbols_ + they always have a name.

So I suggest we update our `command` API this way:

- Rename `sym` to `fn` for consistency with handlers.
  Allow both lambdas and symbols there.
- Add a `name` slot.
- Add a `make-command` macro which would be just like `define-command` except that it would not define a top-level `defun`.

Ambrevar:
Fixes #1192.

This is based on the #1157 prompt buffer pull request because we need this change there and we don't want to rebase excessively.

So look at the commits starting from

https://github.com/atlas-engineer/nyxt/commit/78d18afbf1c63345799e375ea477a338f022ac7b
"command: Rename access-time slot to last-access and use local-time."

This changes many core elements, so please review carefully.

Highlights:

- Cleaner, simpler, more consistent.
- No more useless `pkg` slot, all the `find-symbol` and `symbol-package` dance is gone.
- Commands are pretty-printed.
- Commands are funcallable!
- New `make-command` for local commands, as opposed to `define-command` which defines top-level commands.


Ambrevar:
Don't merge yet, I need to do some more testing myself.

source/command.lisp
```diff
@@ -6,25 +6,46 @@
...
+We need a `command' class for multiple reasons:
+- Identify commands uniquely.
+
+- Customize prompt buffer display value with properties.
+
+- Last access: This is useful to sort command by the time they were last
+  called.  The only way to do this is to persist the command instances."))
+
+(defmethod initialize-instance :after ((command command) &key)
+  (closer-mop:set-funcallable-instance-function command (fn command)))
```
jmercouris, line 44: Voila!

jmercouris:
beautiful work :-)

only thing I don't understand is why change "sym" to "name". I see the symbol as effectively being a way to name a function, to point to it. I didn't test the code yet, but everything appears very sound design wise!

jmercouris:
I will not merge this onto master, only onto the prompt-buffer branch so that I may use it :-)

Ambrevar:
The renaming from `sym` to `name` has some justifications:

- Consistency with Serapeum's hook handlers.
- `sym` is not a word while `name` is, which is more Lispy and reads better.
- A symbol is a characteristic, not an intention, so in that sense `name` is more precise.  (Indeed, almost everything in Lisp is a symbol.)

PR #1205, opened 2021-03-11
WIP: Extend `fill-input-from-external-editor`

aadcg:
There are many issues here.  

My js knowledge is awful so ideas and suggestions are welcome.

I don't understand what `ffi-within-renderer-thread` does, but when I use it, it doesn't do what I want.

### DIFF
--- a/source/external-editor.lisp
+++ b/source/external-editor.lisp
@@ -17,23 +17,30 @@ so invoke on a separate thread when possible."
                (external-editor-program *browser*) p)
     ;; launch-program runs asynchronously but it misbehaves
     (uiop:run-program (list (external-editor-program *browser*)
-                               (uiop:native-namestring p)) :ignore-error-status t)
+                            (uiop:native-namestring p))
+                      :ignore-error-status t)
     (uiop:read-file-string p)))
 
 (define-parenscript select-input-field ()
   (let ((active-tag (ps:chain document active-element tag-name)))
     (when (or (string= active-tag "INPUT")
               (string= active-tag "TEXTAREA"))
+      ;; (ps:chain document (exec-command "selectAll"))
       (ps:chain document active-element (select)))))
 
-(define-parenscript undo-selection ()
-  (ps:chain window (get-selection) (remove-all-ranges)))
+;; this could be an optional exiting behaviour
+;; (define-parenscript undo-selection ()
+;;   (ps:chain window (get-selection) (remove-all-ranges)))
+
+(define-parenscript set-caret-on-end ()
+  (ps:chain document (get-selection) (collapse-to-end)))
 
 ;; known issues:
 ;; fails on duckduckgo's search bar because it loses its focus
 (define-command fill-input-from-external-editor ()
   "Edit the current input field using `external-editor-program'."
-  (select-input-field)
   (bt:make-thread
-   (lambda () (%paste :input-text (open-with-external-editor (%copy)))
-         (undo-selection))))
+   (lambda ()
+     (select-input-field)
+     (%paste :input-text (open-with-external-editor (%copy)))
+     (set-caret-on-end))))
--- a/source/external-editor.lisp
+++ b/source/external-editor.lisp
@@ -3,13 +3,21 @@
 
 (in-package :nyxt)
 
-(defun get-text-from-external-editor ()
-  "Open an external editor to capture text. Please note, this function
-is blocking, invoke on a separate thread when possible."
-  (uiop:with-temporary-file (:directory (uiop:xdg-data-home +data-root+) :pathname p)
-    (log:debug "External Editor: ~a opening: ~a" (external-editor-program *browser*) p)
-    (uiop:run-program (list (external-editor-program *browser*) (uiop:native-namestring p))
-                      :ignore-error-status t)
+(defun open-with-external-editor (&optional input-text)
+  "Edit `input-text' using `external-editor-program'.
+Create a temporary file and return its content.  The editor runs synchronously
+so invoke on a separate thread when possible."
+  (uiop:with-temporary-file (:directory (uiop:xdg-data-home +data-root+)
+                             :pathname p)
+    (when (> (length input-text) 0)
+      (with-open-file (f p :direction :io
+                           :if-exists :append)
+        (write-sequence input-text f)))
+    (log:debug "External Editor: ~a opening: ~a"
+               (external-editor-program *browser*) p)
+    ;; launch-program runs asynchronously but it misbehaves
+    (uiop:run-program (list (external-editor-program *browser*)
+                               (uiop:native-namestring p)) :ignore-error-status t)
     (uiop:read-file-string p)))
 
 (define-command fill-input-from-external-editor ()
--- a/source/external-editor.lisp
+++ b/source/external-editor.lisp
@@ -13,9 +13,8 @@ so invoke on a separate thread when possible."
       (with-open-file (f p :direction :io
                            :if-exists :append)
         (write-sequence input-text f)))
-    (log:debug "External Editor: ~a opening: ~a"
+    (log:debug "External Editor: ~s opening: ~s"
                (external-editor-program *browser*) p)
-    ;; launch-program runs asynchronously but it misbehaves
     (uiop:run-program (list (external-editor-program *browser*)
                             (uiop:native-namestring p))
                       :ignore-error-status t)
@@ -25,18 +24,27 @@ so invoke on a separate thread when possible."
   (let ((active-tag (ps:chain document active-element tag-name)))
     (when (or (string= active-tag "INPUT")
               (string= active-tag "TEXTAREA"))
-      ;; (ps:chain document (exec-command "selectAll"))
       (ps:chain document active-element (select)))))
-
-;; this could be an optional exiting behaviour
-;; (define-parenscript undo-selection ()
-;;   (ps:chain window (get-selection) (remove-all-ranges)))
 
 (define-parenscript set-caret-on-end ()
   (ps:chain document (get-selection) (collapse-to-end)))
 
-;; known issues:
-;; fails on duckduckgo's search bar because it loses its focus
+;; TODO:
+
+;; BUG: Fails when the input field loses its focus, e.g the duckduckgo search
+;; bar.  Can probably be solved with JS.
+
+;; There could be an optional exiting behaviour -- set-caret-on-end or
+;; undo-selection.
+
+;; (define-parenscript undo-selection ()
+;;   (ps:chain window (get-selection) (remove-all-ranges)))
+
+;; It could be extended so that the coordinates of the cursor (line,column)
+;; could be shared between Nyxt and the external editor.  A general solution
+;; can't be achieved since not all editors, e.g. vi, accept the syntax
+;; `+line:column' as an option to start the editor.
+
 (define-command fill-input-from-external-editor ()
   "Edit the current input field using `external-editor-program'."
   (bt:make-thread
--- a/source/external-editor.lisp
+++ b/source/external-editor.lisp
@@ -20,14 +20,20 @@ so invoke on a separate thread when possible."
                                (uiop:native-namestring p)) :ignore-error-status t)
     (uiop:read-file-string p)))
 
+(define-parenscript select-input-field ()
+  (let ((active-tag (ps:chain document active-element tag-name)))
+    (when (or (string= active-tag "INPUT")
+              (string= active-tag "TEXTAREA"))
+      (ps:chain document active-element (select)))))
+
+(define-parenscript undo-selection ()
+  (ps:chain window (get-selection) (remove-all-ranges)))
+
+;; known issues:
+;; fails on duckduckgo's search bar because it loses its focus
 (define-command fill-input-from-external-editor ()
-  "This command will open your editor specified by your VISUAL-EDITOR of
-the BROWSER class, if unset, it will default to your VISUAL environment
-variable. It will then capture whatever text you enter and save in your
-editor."
+  "Edit the current input field using `external-editor-program'."
+  (select-input-field)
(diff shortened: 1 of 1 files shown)


source/external-editor.lisp
```diff
@@ -3,13 +3,21 @@
...
+                             :pathname p)
+    (when (> (length input-text) 0)
+      (with-open-file (f p :direction :io
+                           :if-exists :append)
+        (write-sequence input-text f)))
+    (log:debug "External Editor: ~a opening: ~a"
+               (external-editor-program *browser*) p)
+    ;; launch-program runs asynchronously but it misbehaves
+    (uiop:run-program (list (external-editor-program *browser*)
+                               (uiop:native-namestring p)) :ignore-error-status t)
```
Ambrevar, line 20: Style: Insert new line to group arguments by their calling function.

source/external-editor.lisp
```diff
@@ -20,14 +20,20 @@ so invoke on a separate thread when possible."
...
+  (let ((active-tag (ps:chain document active-element tag-name)))
+    (when (or (string= active-tag "INPUT")
+              (string= active-tag "TEXTAREA"))
+      (ps:chain document active-element (select)))))
+
+(define-parenscript undo-selection ()
+  (ps:chain window (get-selection) (remove-all-ranges)))
+
+;; known issues:
+;; fails on duckduckgo's search bar because it loses its focus
```
Ambrevar, line 33: What fails exactly?

Ambrevar:
Why copying and pasting instead of fetching the input tag value directly?

About `ffi-within-renderer-thread`: So GTK (and WebKitGTK) runs on a separate thread.  All the FFI calls must happen on this thread. 

Usually you need not bother with this and actually I believe we don't need it anymore for this.

source/external-editor.lisp
aadcg: Hard to explain by words so let me give you a recipe.

- go to [duckduckgo](https://duckduckgo.com/) and type smth;
- in the REPL try to do the following:

```
(ffi-buffer-evaluate-javascript 
            (current-buffer) 
            (ps:ps (ps:chain document active-element value)))
```

Notice that it doesn't work.

But the following works when you go back to the browser after issuing the command from the REPL

```
(progn (sleep 2)
                  (ffi-buffer-evaluate-javascript 
                   (current-buffer) 
                   (ps:ps (ps:chain document active-element value))))
```

Basically what's happening is that when the user changes windows that input field stops having the focus. I don't understand why it behaves like that in duckduckgo. All inputs text field I've tested work differently (i.e. they don't loose the focus).

Ambrevar: Sorry, don't know either :(

aadcg:

> Why copying and pasting instead of fetching the input tag value
> directly?

Indeed I started to write smth along the lines of:

```lisp
(define-parenscript set-input-field (text)
  (setf (ps:chain document active-element value) (ps:lisp text)))

(define-parenscript get-input-field ()
  (let ((active-tag (ps:chain document active-element tag-name)))
    (when (or (string= active-tag "INPUT")
              (string= active-tag "TEXTAREA"))))
  (ps:chain document active-element value))
```

But then I understood that I'm reinventing the wheel and that most of
this code is already written.  In the sense that `%paste` checks if the
current tag is an input field, and sets the text depending on the fact
that there's selected text or not.

It seemed elegant to leverage `%copy` and `%paste`.  I even went way
further.  I renamed `fill-input-from-external-editor` to
`invoke-external-editor` and set the former as an alias to the latter.
Why?  Basically it would provide a way to edit whatever selection on a
text editor (regardless of it being a some text on a web page or the
whole input field so that it can be edited).

In short: selected text goes to externals editors (via `%copy`) and smth
ONLY comes out when there's an active input field.  But `%paste` takes
cares of that!  So, what are the only missing pieces?  A way to select
the whole text of a given input field, and then placing the caret in the
end of it when the user is finished with editing.

There are two issues at hand to make that happen:

- setting the caret properly;
- deal with those inputs field that lose their focus (like duckduckgo).

You'll either hate it or love it.  Let me know :)

> About ffi-within-renderer-thread: So GTK (and WebKitGTK) runs on a
> separate thread. All the FFI calls must happen on this thread.
>
> Usually you need not bother with this and actually I believe we don't
> need it anymore for this.

I see.

source/external-editor.lisp
aadcg: No worries. It's js, not rocket science, I'll manage.

Ambrevar:
Ideally, the caret would not be set at the end but where it currently is.  For instance if the input field is

```txt
line1
li|ne2
```

and `|` is the HTML cursor, we need to be able to replicate that in the editor.  For most editors, it's as simple as editing a temporary file with `:$LINE:$COLUMN` appended.  For instance, if you dump the input field content into `/tmp/foo`, then call the editor over `/tmp/foo:1:2` should do the right thing (assuming first line index is 0).

aadcg:
> Ideally, the caret would not be set at the end but where it currently
> is. For instance if the input field is
>
> line1
> li|ne2
>
> and | is the HTML cursor, we need to be able to replicate that in the
> editor. For most editors, it's as simple as editing a temporary file
> with :$LINE:$COLUMN appended. For instance, if you dump the input field
> content into /tmp/foo, then call the editor over /tmp/foo:1:2 should do
> the right thing (assuming first line index is 0).

Indeed, I agree that seems like sensible behaviour!

But I have concerns that such a strategy isn't general enough, and it
makes little sense to particularise at the moment.

I was a bit sceptical that most editors accept the syntax you mention to
pass the line and column.  After taking a look at the vi manual, and the
following page, it seems justified:

https://unix.stackexchange.com/questions/174887/vi-commandline-goto-line-and-column?newreg=d78ae1f308354ce692ed08c7ea25feac

I would provide a spartan solution that works for any EDITOR, and
perhaps implement a subset of your suggestions for emacs (because
I might be interested in that functionally, personally).

Since we're touching this topic, I still find it very hard to understand
where is the line between "core" Nyxt and "extensions".  To my mind,
Nyxt draws (very wisely) a line between the renderer and the "chrome".
But, within the "chrome", a line must also be drawn!  If that line isn't
drawn anytime soon, I think we (developers/maintainers) won't be able to
handle so much functionality.  That burden can't be on us.  We should
provide spartan, but rigorously verified functionality that is general
enough so that the "bazaar" can then figure things out by itself.  Is
there a roadmap for this?  Isn't my concern justified?

Concretely, things such as testing, formal verification, security,
running on all platforms (including Windows!) and such concerns should
be our priority.  The community (when big enough) will figure out the
hacks, and it flourish if the foundation is solid enough.

Let me give a concrete example.  Think about the minibuffer.  Shouldn't
Nyxt only provide one or several completion mechanisms (say fuzzy,
regex, substring, whatever) and a no-frills interface.  Then, the
community could decide what kind of interface (helm-like, whatever-like)
they want.  Don't get me wrong.  The minibuffer is a work of art.  I'm
wondering if the task of figuring out such high-level and opinionated
things should land on "us".

Ambrevar:
This kind of editor support is widespread among keyboard-driven browsers
(Qutebrowser and many more).

So I believe this answers both questions:

- The `$LINE:$COLUMN` syntax is a reasonable default that we can expect
  would work for most editors.
  We can add a method to allow the user to fully customize the editor
  call for their editor, if need be.

- It fits the core of Nyxt.


About extensions:

- Any piece of website-specific code belongs to an extension.
- For the rest, if it's not general and well tested enough, I suggest we
  leave it to extensions for now (like the video downloaded and maybe
  soon the file manager since we might not have the time to revamp it
  for 2.0).

As you hinted, testing is the key here.
I don't think we should hold back the features we can add to Nyxt.  What
we need to work on, however, is our test suite to give us a guarantee
that the whole code base works.

We are far from this goal, and after the renderer polishing this might
be my next task.

As for the minibuffer, I invite you to try out the new `prompt-buffer`
branch, it should answer your concern! :)
It's both more powerful and more customizable than Helm and it comes
with a test suite (that I'm polishing today).

aadcg:

> This kind of editor support is widespread among keyboard-driven browsers
> (Qutebrowser and many more).
>
> So I believe this answers both questions:
>
> - The `$LINE:$COLUMN` syntax is a reasonable default that we can expect
> would work for most editors.
> We can add a method to allow the user to fully customize the editor
> call for their editor, if need be.
>
> - It fits the core of Nyxt.

I see.  What I provided in 21f12f27 is already an upgrade compared to
what is in place (take a moment to read the commit log since I added
some interesting nits about emacs).  But it doesn't implement this
line:col rationale yet, and I don't plan to work on that at the moment.

> About extensions:
>
> - Any piece of website-specific code belongs to an extension.
> - For the rest, if it's not general and well tested enough, I suggest we
> leave it to extensions for now (like the video downloaded and maybe
> soon the file manager since we might not have the time to revamp it
> for 2.0).

Fair.

> As you hinted, testing is the key here.
> I don't think we should hold back the features we can add to Nyxt. What
> we need to work on, however, is our test suite to give us a guarantee
> that the whole code base works.
>
> We are far from this goal, and after the renderer polishing this might
> be my next task.

Sounds good.

> As for the minibuffer, I invite you to try out the new `prompt-buffer`
> branch, it should answer your concern! :)
> It's both more powerful and more customizable than Helm and it comes
> with a test suite (that I'm polishing today).

Music to my ears :)

source/external-editor.lisp
```diff
@@ -3,23 +3,44 @@
...
+so invoke on a separate thread when possible."
+  (uiop:with-temporary-file (:directory (uiop:xdg-data-home +data-root+)
+                             :pathname p)
+    (when (> (length input-text) 0)
+      (with-open-file (f p :direction :io
+                           :if-exists :append)
+        (write-sequence input-text f)))
+    (log:debug "External Editor: ~a opening: ~a"
+               (external-editor-program *browser*) p)
+    ;; launch-program runs asynchronously but it misbehaves
```
Ambrevar, line 18: How does it misbehaves?

source/external-editor.lisp
```diff
@@ -3,23 +3,44 @@
...
+  "Edit `input-text' using `external-editor-program'.
+Create a temporary file and return its content.  The editor runs synchronously
+so invoke on a separate thread when possible."
+  (uiop:with-temporary-file (:directory (uiop:xdg-data-home +data-root+)
+                             :pathname p)
+    (when (> (length input-text) 0)
+      (with-open-file (f p :direction :io
+                           :if-exists :append)
+        (write-sequence input-text f)))
+    (log:debug "External Editor: ~a opening: ~a"
```
Ambrevar, line 16: Can you replace the format string as follows: `"External editor ~s opens ~s"`.  The `~s` ensure the strings are quoted, which makes it more readable when the argument contains spaces or double quotes for instance.

source/external-editor.lisp
```diff
@@ -3,23 +3,44 @@
...
+               (external-editor-program *browser*) p)
+    ;; launch-program runs asynchronously but it misbehaves
+    (uiop:run-program (list (external-editor-program *browser*)
+                            (uiop:native-namestring p))
```
Ambrevar, line 20: Important! Always use `uiop:parse-native-namestring` when you get a path externally.
Otherwise `foo[bar` will raise a condition.

source/external-editor.lisp
```diff
@@ -3,23 +3,44 @@
...
                       :ignore-error-status t)
     (uiop:read-file-string p)))
 
+(define-parenscript select-input-field ()
+  (let ((active-tag (ps:chain document active-element tag-name)))
+    (when (or (string= active-tag "INPUT")
+              (string= active-tag "TEXTAREA"))
+      ;; (ps:chain document (exec-command "selectAll"))
```
Ambrevar, line 28: What's the different between `selectAll` and `select`?
If you want to leave the comment, please explain why.

source/external-editor.lisp
```diff
@@ -3,23 +3,44 @@
...
+      (ps:chain document active-element (select)))))
+
+;; this could be an optional exiting behaviour
```
Ambrevar, line 31: Style: I like to write my comments by respecting English punctation, that is, start with a capital and end with a period.  This makes it easier to the eye to separate from code.

Ambrevar:
I've added some last few comments.  Beside these nits, looks good to me, I think we can merge.
Thanks for working on this, this is a feature I've been wanting for long! :)

Before merging, could you comment on what's left to be done?
Either add some `TODO:` comments, or a  comment at the top that recaps the situation.

Thanks!

source/external-editor.lisp
aadcg: Your suggestion raises an error: 
```
; Evaluation aborted on #<SIMPLE-TYPE-ERROR expected-type: (OR STRING NULL)
                    datum: #P "/path/"
```

source/external-editor.lisp
aadcg: I've made my mind.  I don't want to leave any comment :)

`selectAll` basically mimics C-a.  Since we're checking if the active tag is either `INPUT` or `TEXTAREA` it makes more sense to leave things as they are.

source/external-editor.lisp
aadcg: I dropped my irrelevant comment.  I tried to understand if there would be any benefit of implementing this using `lanch-program` and I don't think there is :)

aadcg:
Looks good to me. Let me know what you think!

source/external-editor.lisp
Ambrevar: My bad, I missed that `p` was coming from a CL-generated pathname (here `uiop:with-temporary-file`.  Since it's assured to be a non-wildcard pathname, then `native-namestring` is correct here indeed!

Ambrevar:
Merged with 25957747b2b1459572df32710401071ae6a03b47 where I've added error checking.

Thanks a lot for this great improvement!

PR #1206, opened 2021-03-11
Add CUA copy and paste bindings to the minibuffer

nnoodle:
No <kbd>s-delete</kbd> since I don't think there's a command analogous to cut.


Ambrevar:
Looks good to me, thanks for your contribution!

PR #1208, opened 2021-03-15
WIP: Add process-mode.

aartaka:
This adds a general conditional action-running mode -- `process-mode`. Some kind of Cron inside Nyxt :)

# What's New
- `process-mode` that basically executes its `action` when its `firing-condition` is true.
- Derived `preview-mode` that looks at whether the file was changed and reloads the buffer if necessary.
- Derived `repeat-mode` that repeats its own `repetitive-action` (usually a command, like `nyxt/web-mode:scroll-down`) every `repeat-interval` seconds.
- `watch-mode`, now derived from `repeat-mode`, with the only difference that it has a `repetitive-action` set to `buffer-load` by default.

# Things To Discuss
- Do we need e.g. `watch-mode` now that one can basically do `repeat-every` and choose repetition interval and command to execute?
- Currently, `preview-mode` relies on `sb-posix` instead of removed dependency on `file-attributes`. What's the right way to look at mtime of a file now?

# How Has This Been Tested?
- Compiles.
- Runs.
- `repeat-mode` scrolls page down, `watch-mode` reloads it.

EDIT: All questions resolved, what's left is testing.
EDIT: Pre-final PR message form.

### DIFF
--- a/source/action-mode.lisp
+++ b/source/action-mode.lisp
@@ -1,0 +1,209 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/action-mode
+  (:use :common-lisp :trivia :nyxt)
+  (:import-from #:class-star #:define-class)
+  (:documentation "Act on file/directory based on a certain condition."))
+(in-package :nyxt/action-mode)
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (trivial-package-local-nicknames:add-package-local-nickname :alex :alexandria)
+  (trivial-package-local-nicknames:add-package-local-nickname :sera :serapeum)
+  (trivial-package-local-nicknames:add-package-local-nickname :file-attributes :org.shirakumo.file-attributes))
+
+;; REVIEW: `action' and `cleanup' resemble `constructor' and `destructor'.
+;; TODO: Maybe subclass `action-mode' instead?
+;; This will sacrifice the flexibility of `action's, though...
+(define-class action ()
+  ((firing-condition t
+                         :type (or boolean function)
+                         :documentation "The condition for the action firing.
+Can be boolean (T to always fire, NIL to never fire), or function over URL and mode.")
+   (action nil
+           :type (or (function (quri:uri root-mode)) null)
+           :documentation "The action (function) to do with file URL and action-mode.")
+   (cleanup nil
+            :type (or (function (quri:uri root-mode)) null)
+            :documentation "Function to run when action ends.
+Accepts the path to the acted-on document and action-mode instance."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:export-predicate-name-p t)
+  (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
+
+(defmethod nyxt:object-string ((action action))
+  (str:downcase (symbol-name (sera:class-name-of action))))
+
+(defmethod nyxt:object-display ((action action))
+  (str:downcase (symbol-name (sera:class-name-of action))))
+
+(defun updated-file-p (path-url mode)
+  (and
+   (quri:uri-file-p path-url)
+   (or (null (last-access (action mode)))
+       (local-time:timestamp>
+        (local-time:universal-to-timestamp
+         (file-attributes:modification-time (quri:uri-path path-url)))
+        (last-access (action mode))))))
+
+(define-class preview-action (action)
+  ((firing-condition #'updated-file-p)
+   (action
+    #'(lambda (path-url mode)
+        (buffer-load path-url :buffer (buffer mode))
+        (setf (last-access (action mode)) (local-time:now))))
+   (last-access nil
+                :type (or local-time:timestamp null)
+                :documentation "The time `path-url' file was last accessed."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
+
+(define-class server-action (preview-action)
+  ((action #'(lambda (path-url mode)
+               (setf (acceptor (action mode))
+                     (if (acceptor (action mode))
+                         (hunchentoot:start
+                          (hunchentoot:stop (acceptor (action mode))))
+                         (hunchentoot:start
+                          (or (acceptor (action mode))
+                              (make-instance 'hunchentoot:easy-acceptor
+                                             :document-root (uiop:pathname-directory-pathname
+                                                             (pathname (quri:uri-path path-url)))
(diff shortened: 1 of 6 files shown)


aadcg:

> * @aadcg, is it okay to devour watch-mode? It fits too well with this
>   idea of regular actions based on conditions.

Yes bugatyr!  I appreciate that you ask, but you needn't!

jmercouris:
Oh boy oh boy :-D! Very exciting stuff!

What follows is my best attempt to answer your questions:
+ I think there should be an action-mode. Preview-mode, server-mode, and watch-mode can inherit from action-mode. Benefits of said architecture:
  + Server-mode can exist in a different extension/system. 
  + Action-mode can exist in the Nyxt core.
  + Server-mode can inherit all of the base behavior it needs from action-mode without pulling in a dependency for a server into the Nyxt core.
+ As implied by the previous bullet point, I don't think it is a good idea to bring in a web server dependency into the Nyxt core. Hence my reasoning for the inheritance structure I've described.
+ Maybe have several actions associated to the mode? Given the above inheritance structure, you would instead have several different instances active. I think this is the more tenable long-term solution. It will probably be easier to view each action/event pair as an individual object, rather than conflating them into a master object.

Hopefully that feedback is helpful! If you need any clarification, please do not hesitate to ask! Once again, cool stuff!

source/action-mode.lisp
```diff
@@ -0,0 +1,209 @@
...
+   (sleep-time :documentation "The amount of time to sleep between reloads."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
+
+(define-class line-count-action (nyxt/action-mode:preview-action)
+  ((action
+    #'(lambda (path-url mode)
+        (let ((lines (length (uiop:read-file-lines (quri:uri-path path-url)))))
+          (when (\= (lines (action mode)) lines)
```
Ambrevar, line 143: Typo in `\=`?

source/action-mode.lisp
```diff
@@ -0,0 +1,209 @@
...
+         :documentation "The port server starts on."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
+
+(defun seconds-from-user-input ()
+  "Read from the minibuffer numerical time inputs and collate them into
+seconds."
+  (let* ((time-units '("days" "hours" "minutes" "seconds"))
+         (to-seconds-alist (pairlis time-units '(86400 3600 60 1)))
```
Ambrevar, line 101: Maybe replace the hard-coded numbers with `(* 60 60)` and `(* 24 60 60)`.

Ambrevar:
I like the term "king of Cron" (second sentence in the original post) :)

This looks very cool!  Great job!
I agree with John, I believe that the server should be an extension (at least for now).

What about the naming?  I find `action` too generic.
If we are talking "cron", what about "planner" or "scheduler"?  In this context, we often speak of "tasks" rather than "actions".  Thoughts?

source/action-mode.lisp
aadcg: Fine for me.  I'm just so used to those numbers that I identify them immediately.

Ambrevar: I believe this is the kind of knowledge assumptions we should avoid: the less knowledge required, the easier to read, the easier for newcomers to join.

This is one of the benefits of the Lisp syntax and conventions: by focusing and readability and plain English, Lisp code can be very readable even to people who've never programmed before.

aartaka:
> What follows is my best attempt to answer your questions:
> 
>     * I think there should be an action-mode. Preview-mode, server-mode, and watch-mode can inherit from action-mode. Benefits of said architecture:
>       
>       * Server-mode can exist in a different extension/system.
>       * Action-mode can exist in the Nyxt core.
>       * Server-mode can inherit all of the base behavior it needs from action-mode without pulling in a dependency for a server into the Nyxt core.

Fair points!

>     * As implied by the previous bullet point, I don't think it is a good idea to bring in a web server dependency into the Nyxt core. Hence my reasoning for the inheritance structure I've described.

Sure! I'll refactor it as an inheritable mode then and move `server-mode` to a separate repo :)

>     * Maybe have several actions associated to the mode? Given the above inheritance structure, you would instead have several different instances active. I think this is the more tenable long-term solution. It will probably be easier to view each action/event pair as an individual object, rather than conflating them into a master object.

Yes, that's the same thing I came to while thinking about `mode` vs. `action`. Modes are clearer.

> Hopefully that feedback is helpful! If you need any clarification, please do not hesitate to ask! Once again, cool stuff!

It's all clear, thank you!

aartaka:
> I like the term "king of Cron" (second sentence in the original post) :)

Wonderful typo it is xD

> What about the naming? I find `action` too generic.
> If we are talking "cron", what about "planner" or "scheduler"? In this context, we often speak of "tasks" rather than "actions". Thoughts?

Cron-like behavior is only a subset of this mode applications, as `firing-condition` can be anything. Maybe `process-mode` to emphasize the fork-like thread creation? `task-mode` is a good one too!

source/action-mode.lisp
aadcg: No worries from my side. Already agreed above :)

aartaka:
UPD: It compiles, it's rebased over master, most comments are addressed. 

What's left is testing and deciding when to merge. This seems to be quite a new feature, and we agreed to not merge new features until 2.0. So I test and leave it until then?

Ambrevar:
I'd prefer to merge this after 2.0.
About the branch: well, this is already on a branch.

2.0 involves still quite some work.  If we encumber ourselves with both a 2.0 branch _and_ a master branch, it will be more additional work just to backport all the universal changes on the other branch.  I'd rather have this extra load after 2.0 is out.

Does that make sense?
What do you think?

aartaka:
Yes, that does sound reasonable!

jmercouris:
Taking a look at this now Artyom!

jmercouris (approved):

source/repeat-mode.lisp
```diff
@@ -38,7 +26,19 @@ Defaults to one second.")
                       :type (or null (function (repeat-mode)))
                       :documentation "The action to repeat.
 Function taking a `repeat-mode' instance.")
-   (constructor #'initialize-repeat-mode)))
+   (constructor #'initialize)))
```
aartaka, line 29: Isn't the generic function name simply inherited from `process-mode` there?

jmercouris: Not sure what you mean. Dispatch will happen correctly though.

jmercouris: I see what you are saying, in fact it is. this can be removed.

jmercouris:
Merci beaucoup monsieur Artyom!

Issue #1065, opened 2020-12-09
Replace Quicklisp with CLPM?


Ambrevar:
CLPM is a Common Lisp package manager: https://www.clpm.dev/

It seems that we could use it to fetch pinned dependency versions.  So instead of relying on an external distribution (Quicklisp) which we don't control (and thus may break Nyxt), we could use CLPM both for the setup and the runtime Lisp system management.

Ambrevar:
The issue https://github.com/atlas-engineer/nyxt/issues/1217 emphasizes the urgency to have our own distribution: Quicklisp is a moving target and it makes Nyxt unreproducible.

On the Guix front, everything is fine, but as soon as we leave Guix the build robustness blows up.

jmercouris:
Indeed, hence my original bringing up of CLPM :-)

crocket:
Does CLPM support offline build if dependencies are already present?

Can CLPM rely on dependencies installed by linux distribution package manager?

Ambrevar:
I don't enough about CLPM to give you good answers.

For now, we are going to rely on Git submodules to fix the urgent Quicklisp issues.
https://github.com/atlas-engineer/nyxt/pull/1802

If we get more comfortable with CLPM and it proves to be the right tool, we might switch to it.

jmercouris:
Well, we are no longer using QL :-)

Ambrevar:
But we are not using CLPM yet, so I suggest we reopen this!

jmercouris:
Why do we need CLPM though? I think pinning things the way we've done is probably the most robust, and doesn't require the user to install any other software.

Ambrevar:
In the future, when Nyxt becomes more of a full-featured CL development
environment, having a proper Common Lisp package manager would be good!

crocket:
Can CLPM build nyxt offline? Gentoo Linux builds programs offline. All dependencies should either be specified in gentoo package or exist as a separate gentoo package.

Ambrevar:
I don't know yet, but that's indeed a critical requirement.

ohmree:
The `--local` and maybe the `--no-resolve` flags to `clpm bundle install` might be of interest here.
Note that I'm not at all an experienced clpm user, I just did a super quick search using the cli because I remembered seeing a flag for this.
If it is a direction you're willing to investigate then maybe you could try contacting clpm's maintainer and/or opening issues [on its repo](https://gitlab.common-lisp.net/clpm/clpm).

```
[...]

PR #1211, opened 2021-03-16
Library move

jmercouris:
Move quicklisp-libraries -> libraries.

### DIFF
--- a/.github/workflows/package-ubuntu.yml
+++ b/.github/workflows/package-ubuntu.yml
@@ -37,7 +37,7 @@ jobs:
       run: |
         make quicklisp-extra-libs
         mkdir -p ~/.config/common-lisp/source-registry.conf.d/
-        echo "(:tree \"$PWD/quicklisp-libraries\")" >> ~/.config/common-lisp/source-registry.conf.d/asdf.conf
+        echo "(:tree \"$PWD/libraries\")" >> ~/.config/common-lisp/source-registry.conf.d/asdf.conf
 
     - name: Build special SBCL and make package
       shell: bash
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,7 +62,7 @@ jobs:
       shell: bash
       run: |
         make quicklisp-extra-libs
-        echo "(:tree \"$PWD/quicklisp-libraries\")" > ~/.config/common-lisp/source-registry.conf.d/asdf.conf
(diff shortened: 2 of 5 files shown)


jmercouris:
@Ambrevar :-D?

Ambrevar:
I'm hot on the pipeline today, no worries, I'll get to it! :)

jmercouris:
:-D

Ambrevar:
The patch seems correct (I haven't tested) but I'm having second thoughts: in the past we've had issues with people having stale libraries in quicklisp-libraries.
Having a separate folder makes it easier to just tell them "delete the content of the `quicklisp-libraries` folder".  Without a separate directory, we'd need to list the exact Nyxt-owned libraries to keep.

Does that make sense?
Can you think of a better way to do it?

Of course, should we adopt CLPM #1065, none of this would be an issue, we would not even need submodules in the first place.

jmercouris:
I think all git submodules should be updated, whether they are on quicklisp or not. We should be able to just give them a command to update all submodules. Therefore, I think it would still be OK. What do you think?

Ambrevar:
Misunderstanding.  What I'm saying is unrelated to the update of git
submodules.

The problem is when people try to build _without_ using the submodules,
for instance using Guix, Nix or ~/common-lisp.

In this case, having the library submodules checked out may create a
conflict (for instance if they forgot NYXT_INTERNAL_QUICKLISP=false).
Should we move submodules to libraries, ASDF would have no way to know
whether to load the Guix package or the library in `nyxt/libraries`.

Does that make sense?

jmercouris:
I'm not sure I understand. How could someone build without using the submodules? Using their own outdated quicklisp?

Can we not force the makefile to use the libraries located in the libraries folder when `NYXT_INTERNAL_QUICKLISP=true`?

jmercouris:
Oh, I see, you don't mean literally without the source of those submodules, you mean without the mechanism of the git submodules for retrieving said source!

jmercouris:
In that case, is what I said still not true? Can our makefile not impose certain behaviors on ASDF based on the value of 'NYXT_INTERNAL_QUICKLISP`?

Ambrevar:
Example with Guix: Nyxt is built using the cl-cluffer (or sbcl-cluffer)
package.  The submodules are never retrieved when using Guix.

Ambrevar:
Currently, when NYXT_INTERNAL_QUICKLISP is false, quicklisp-libraries is
not added to ADSF search path.

But `libraries` is always added.

So if we move the submodules to `libraries`, we have no way to say
"don't load the submodules" anymore.

Issue #1227, opened 2021-03-23
Nyxt has window class "Sbcl"


franburstall:
**Describe the bug**

According to `exwm`, nyxt windows have `WM_CLASS` set to "Sbcl".  They used to have class "nyxt".

**Precise recipe to reproduce the issue**
Fire nyxt under `exwm` and inspect the value of the emacs variable `exwm-class-name`.

**Information**
- OS name+version: Debian 10
- Graphics card and driver:
- Desktop environment / Window manager name+version: exwm
- How you installed Nyxt (Guix pack, package manager, build from source): built from source, master branch at f566fa89.

- Information from command copy-system-information:

Nyxt version: draft-do-not-use-me-11-gf566fa89 
Operating system kernel: Linux 4.19.0-14-amd64
Lisp implementation: SBCL 2.1.1
Features: (SLYNK WEBKIT2 WEBKIT2-2.30 WEBKIT2-TRACKING WEBKIT2-MUTE
           WEBKIT2-EMOJI WEBKIT2-MEDIA WEBKIT2-SANDBOXING GTK-3-22 GTK-3-20
           GTK-3-18 GTK-3-16 GTK-3-14 GTK-3-12 GTK-3-10 GTK-3-8 GTK-3-6 GTK-3-4
           GTK GDK-3-22 GDK-3-20 GDK-3-18 GDK-3-16 GDK-3-14 GDK-3-12 GDK-3-10
           GDK-3-8 GDK-3-6 GDK-3-4 CAIRO-1-10 CAIRO-1-12 GDK-PIXBUF GLIB-2-30
           GLIB-2-32 GLIB-2-34 GLIB-2-36 GLIB-2-38 GLIB-2-40 GLIB-2-42
           GLIB-2-44 GLIB-2-46 GLIB-2-48 GLIB-2-50 GLIB-2-52 GLIB-2-54
           GLIB-2-56 GLIB-2-58 GLIB FSET-EXT-STRINGS CUSTOM-HASH-TABLE-NATIVE
           SWANK PLUMP-UTF-32 GLOBAL-VARS DECLARE-TYPES PARENSCRIPT
           NAMED-READTABLES 21BIT-CHARS CHUNGA FLEXI-STREAMS CLOSER-MOP
           CL-PPCRE-UNICODE CL-UNICODE CL-PPCRE BORDEAUX-THREADS SPLIT-SEQUENCE
[...]

PR #1229, opened 2021-03-23
Fix for window class, use gdk-set-program-class

jeffbowman:
Fix for #1227

xprop reports: WM_CLASS(STRING) = "nyxt", "Nyxt"
app_id in Sway (Wayland): "nyxt"


Ambrevar:
Thanks!

PR #1230, opened 2021-03-23
visual-mode.lisp: add more keybindings

Lithrein:
Hi,

I added some keybindings to the visual mode: cua keybindings can now select word by word, to the beginning/end of line, and clear the selection (there is a breaking change concerning the keybinding used to clear the selection), and vim-keybinding was missing the select word backward.

### DIFF
--- a/source/visual-mode.lisp
+++ b/source/visual-mode.lisp
@@ -17,7 +17,7 @@
        "down" 'forward-line
        "left" 'backward-char
        "right" 'forward-char
-       "escape" 'visual-mode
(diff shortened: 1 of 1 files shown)


jmercouris:
Thank you! I will try to take a look tomorrow :-)

jmercouris:
I have not forgotten, just haven't had a moment yet :-)

jmercouris:
I will merge this, only one thing I will change I'll bring back the "Escape" binding for CUA. I think that is what the users would expect. Thanks so much for this improvement!

PR #1236, opened 2021-03-24
base-mode.lisp: Cleanup.

aadcg:
The cua keymap is a parent of emacs and vi so there's no need to repeat
the same information.

`list-buffers' moves to cua keymap for the sake of generality.

--

As discussed with @ambrevar.

### DIFF
--- a/source/base-mode.lisp
+++ b/source/base-mode.lisp
@@ -4,7 +4,8 @@
 (in-package :nyxt)
 
 (define-mode base-mode ()
-  "Mode that does nothing but bind the general-purpose key bindings."
+  "Bind general-purpose commands defined by `define-command'.
+This mode is a good candidate to be passed to `make-buffer'."
   ((keymap-scheme (define-scheme "base"
                     scheme:cua
                     (list
@@ -12,6 +13,7 @@
                      "C-[" 'switch-buffer-previous
                      "C-]" 'switch-buffer-next
                      "C-x b" 'switch-buffer
+                     "C-x C-b" 'list-buffers
                      "C-x k" 'delete-buffer
                      "C-w" 'delete-current-buffer
                      "C-x C-k" 'delete-current-buffer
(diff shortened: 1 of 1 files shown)


Ambrevar:
Looks good to me.  @jmercours ?

jmercouris:
Yes. We must be aware though that if we DO remove something or change something from the CUA keymap, we will want to preserve the original bindings in the Emacs keymap. For example:

We rebind "reload-current-buffer" to "F5" in CUA. The Emacs users would still expect "C-r" to work. Therefore, we would also have to add a binding for "C-r" in the Emacs keymap. In that sense I actually believe this change to be a slight regression. Yes, there is duplication of information, but there is no advantage to de-duplication here, only potential disadvantage.

The Emacs/CUA bindings do not need to evolve in lockstep, and likely will diverge in time. If you are OK with this compromise, and are aware of it (hence why I originally duplicated all values), then please merge.

Ambrevar:
I agree with John about _some_ keys.  For others, there are actual
duplicates that make little sense.

For instance, `"C-[" 'switch-buffer-previous` is not an Emacs binding.
Thus it only needs to be in CUA.

I propose the following strategy:

- CUA bindings should be self-standing.
- Emacs bindings don't need to duplicate CUA bindings when there already
  is an Emacs-specific binding.
- Duplicate the Emacs binding if it happens to be both in CUA and an
  official Emacs binding.
- Same for VI bindings.

Thoughts?

jmercouris:
That appears to be a sound strategy to me!

aadcg:
Nicely observed @jmercouris.  Does 4302e2f8 follow the rationale proposed by @Ambrevar?

Ambrevar:
Indeed!
I think `"C-x C-k" 'delete-current-buffer` is missing.

I guess we could seize the opportunity to fix the bookmarks bindings.
In Emacs, bookmarks are bound to

- `C-x r M		bookmark-set-no-overwrite`
- `C-x r b		bookmark-jump`
- `C-x r l		bookmark-bmenu-list`
- `C-x r m		bookmark-set`

I suggest the following then:

- `C-x r m`: bookmark-page
- `C-x r M`: bookmark-current-page
- `C-x r j`: set-url-from-bookmark
- `C-x r J` set-url-from-bookmark-new-buffer
- `C-x r u`: bookmark-url
- `C-x r k`: bookmark-delete
Thoughts?

aadcg:

> I think "C-x C-k" 'delete-current-buffer is missing.

Ah, I left it out since it's not an emacs binding.  I guess you're
including it because users might type C-x C-k instead of C-x k?  It's ok
from my side.

> I guess we could seize the opportunity to fix the bookmarks bindings.
> In Emacs, bookmarks are bound to
>
> * C-x r M bookmark-set-no-overwrite
> * C-x r b bookmark-jump
> * C-x r l bookmark-bmenu-list
> * C-x r m bookmark-set
>
> I suggest the following then:
>
> * C-x r m: bookmark-page
> * C-x r M: bookmark-current-page
> * C-x r j: set-url-from-bookmark
> * C-x r J set-url-from-bookmark-new-buffer
> * C-x r u: bookmark-url
> * C-x r k: bookmark-delete
>   Thoughts?

All good for me.  Even though, sooner of later I want to spend some time
really thinking about the emacs keymap.  Overall, it should be less
prefixed (a bit more like the vi normal mode keymap) and should resemble
eww as much as possible.  We have discussed this already and some users
raised this as well.

aadcg:
Should be good now :)

jmercouris:
Thank you!

PR #1240, opened 2021-03-25
Catch all errors! O.o

Ambrevar:
This is certainly one of my most important contributions: it catches hopefully all Lisp errors at runtime, which means no more Lisp-induced crashes for Nyxt!

2-pre-release-6 introduced many crash regressions for stupid mistakes that should not have crashed the browser anyways.

With this patchset, this kind of errors are reported, but won't crash the browser anymore!

This introduces a few changes:

- Removal of `funcall-safely`.  Hopefully OK, please test.
- Renamed `*keep-alive*` to the more meaningful `*run-from-repl-p*`.
- Renamed `with-muffled-body` to the more accurate `with-protect`.

Please test and let me know what you think! :)

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -65,6 +65,7 @@
                (:file "user-interface")
                ;; Core functionality
                (:file "global")
+               (:file "concurrency")
                (:file "data-storage")
                (:file "configuration")
                (:file "command")
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -88,7 +88,7 @@ Enable INCLUDED modes plus the already present ones, and disable EXCLUDED modes,
                   (or (position (first test2) priority-list) 4)))))
       (first (sort (remove-if-not
                     #'(lambda (rule)
-                        (funcall-safely
+                        (funcall
                          ;; REVIEW: Use `eval'?
                          (apply (first (test rule)) (rest (test rule))) url))
                     rules)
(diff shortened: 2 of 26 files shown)


Ambrevar:
Important note:
In the future, we should always use `run-thread` instead of `bt:make-thread` and `within-gtk-thread` instead of `gtk:within-gtk-thread`.

jmercouris:
Oh boy oh boy, let me give this bad boy a try :-D

jmercouris:
Appears to work... :-D

Ambrevar:
I just had a hang...
Let's keep using it for a day at least (from the binary, _not from the REPL_).

jmercouris:
Well, that will have to be you for now, I can ONLY run from a REPL :-/

Ambrevar:
Why?
Can I help you build the Nyxt binary?

jmercouris:
Indeed, as we discussed, we can try to figure it out on Monday :-)

Ambrevar:
Actually, you can test the new behaviour from the REPL too if you set `*run-from-repl-p*` to NIL before starting Nyxt.

jmercouris:
I'm not sure that would achieve what we want. I would still be attached to the slime debugger. I still believe only you can test for now.

Ambrevar:
Precisely: if you see the SLIME debugger, it means it didn't work!
So it would test what we want.

jmercouris:
Ah, I see! OK! I will do that then.

Ambrevar:
Thanks!

I've been running it for some time now, looks fine so far.

Ambrevar:
It's working for me all day.

OK to merge?

jmercouris:
Is it OK if I keep using Nyxt over the weekend first?

Ambrevar:
Sure!

jmercouris:
No problems, feel free to merge!

Ambrevar:
Merged!

PR #1242, opened 2021-03-25
Add editor.lisp

jmercouris:
buffer.lisp: add text-editor buffer class, contains a file slot, and a
mode that desginates the editor overlay

editor.lisp: create an editor-mode and a abstract implementation for
editors to implement

### DIFF
--- a/source/editor.lisp
+++ b/source/editor.lisp
@@ -9,10 +9,10 @@
   "Mode for editor modes to extend."
   ())
 
-(defgeneric write-file (editor &key if-exists)
+(defgeneric write-file (buffer-editor &key if-exists)
   (:documentation "Write the file to storage."))
 
-(defgeneric open-file (editor file)
+(defgeneric open-file (buffer-editor file)
   (:documentation "Open the file in the editor."))
 
 (defgeneric get-content (editor)
--- a/source/editor.lisp
+++ b/source/editor.lisp
@@ -26,7 +26,9 @@
                                      :if-exists if-exists))
 
 (defmethod open-file ((buffer editor-buffer) file)
-  (set-content (editor buffer) (uiop:read-file-string file)))
+  (if (uiop:file-exists-p file)
+      (set-content (editor buffer) (uiop:read-file-string file))
+      (set-content (editor buffer) "")))
 
 (define-command editor-open-file (&key (buffer (current-buffer)))
   "Open a file in the internal editor."
--- a/source/editor.lisp
+++ b/source/editor.lisp
@@ -57,3 +57,9 @@
   "Write the FILE of the BUFFER.."
   (write-file buffer :if-exists if-exists)
   (echo "File ~a written to storage." (file buffer)))
+
+(define-command open-new-editor-with-file ()
+  "Open a new editor and query a file."
+  (let ((buffer (make-editor-buffer)))
+    (set-current-buffer buffer)
+    (editor-open-file :buffer buffer)))
--- a/source/editor.lisp
+++ b/source/editor.lisp
@@ -1,13 +1,12 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-;;;; This class is used to define a protocol for editors to implement.
-
 (in-package :nyxt)
 
 (define-mode editor-mode ()
   "Mode for editor modes to extend."
-  ())
+  ()
+  (:documentation "This class is used to define a protocol for editors to implement."))
 
 (defgeneric write-file (buffer-editor &key if-exists)
   (:documentation "Write the file to storage."))
@@ -32,14 +31,14 @@
 (define-command editor-open-file (&key (buffer (current-buffer)))
   "Open a file in the internal editor."
   (let ((file (first (prompt
-                      :prompt "Open File"
+                      :prompt "Open file"
                       :sources (make-instance 'prompter:raw-source
-                                              :name "Absolute File Path")))))
+                                              :name "Absolute file path")))))
     (open-file buffer file)
     (setf (file buffer) file)
     (setf (title buffer) file)
     (setf (url buffer) (quri:uri file))))
 
 (define-command editor-write-file (&key (buffer (current-buffer)) (if-exists :error))
-  "Write a file in the internal editor."
+  "Write the FILE of the BUFFER.."
   (write-file buffer :if-exists if-exists))
--- a/source/editor.lisp
+++ b/source/editor.lisp
@@ -1,0 +1,45 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+;;;; This class is used to define a protocol for editors to implement.
+
+(in-package :nyxt)
+
+(define-mode editor-mode ()
+  "Mode for editor modes to extend."
+  ())
+
+(defgeneric write-file (editor &key if-exists)
+  (:documentation "Write the file to storage."))
+
+(defgeneric open-file (editor file)
+  (:documentation "Open the file in the editor."))
+
+(defgeneric get-content (editor)
+  (:documentation "Get the content of the editor."))
+
+(defgeneric set-content (editor content)
+  (:documentation "Set the content of the editor."))
+
+(defmethod write-file ((buffer editor-buffer) &key (if-exists :error))
+  (alexandria:write-string-into-file (get-content (editor buffer))
+                                     (file buffer)
+                                     :if-exists if-exists))
+
+(defmethod open-file ((buffer editor-buffer) file)
+  (set-content (editor buffer) (uiop:read-file-string file)))
+
+(define-command editor-open-file (&key (buffer (current-buffer)))
+  "Open a file in the internal editor."
+  (let ((file (first (prompt
+                      :prompt "Open File"
+                      :sources (make-instance 'prompter:raw-source
+                                              :name "Absolute File Path")))))
+    (open-file buffer file)
+    (setf (file buffer) file)
+    (setf (title buffer) file)
+    (setf (url buffer) (quri:uri file))))
+
+(define-command editor-write-file (&key (buffer (current-buffer)) (if-exists :error))
+  "Write a file in the internal editor."
+  (write-file buffer :if-exists if-exists))
--- a/source/editor.lisp
+++ b/source/editor.lisp
@@ -43,4 +43,5 @@
 
 (define-command editor-write-file (&key (buffer (current-buffer)) (if-exists :error))
   "Write the FILE of the BUFFER.."
-  (write-file buffer :if-exists if-exists))
+  (write-file buffer :if-exists if-exists)
+  (echo "File ~a written to storage." (file buffer)))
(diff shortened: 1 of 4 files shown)


jmercouris:
You may test these changes in conjunction with nx-ace mode. Make sure to set your default-modes of your editor-buffer to include ace-mode.

jmercouris:
Here is a screenshot for some motivation :-)

Screenshot_20210325_163018

jmercouris:
Sample configuration for your convenience:

```
(define-configuration editor-buffer
    ((default-modes (append '(ace-mode) %slot-default))))
```

source/editor.lisp
```diff
@@ -0,0 +1,45 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+;;;; This class is used to define a protocol for editors to implement.
+
+(in-package :nyxt)
+
+(define-mode editor-mode ()
+  "Mode for editor modes to extend."
+  ())
```
Ambrevar, line 10: The file should be named `editor-mode.lisp` and define the package of the same name, no?

source/editor.lisp
```diff
@@ -0,0 +1,45 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+;;;; This class is used to define a protocol for editors to implement.
+
+(in-package :nyxt)
+
+(define-mode editor-mode ()
+  "Mode for editor modes to extend."
```
Ambrevar, line 9: Why extend?  I don't understand the docstring.

source/editor.lisp
```diff
@@ -0,0 +1,45 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+;;;; This class is used to define a protocol for editors to implement.
```
Ambrevar, line 4: Add this to the package documentation.

source/editor.lisp
```diff
@@ -0,0 +1,45 @@
...
+
+(defmethod open-file ((buffer editor-buffer) file)
+  (set-content (editor buffer) (uiop:read-file-string file)))
+
+(define-command editor-open-file (&key (buffer (current-buffer)))
+  "Open a file in the internal editor."
+  (let ((file (first (prompt
+                      :prompt "Open File"
+                      :sources (make-instance 'prompter:raw-source
+                                              :name "Absolute File Path")))))
```
Ambrevar, line 37: Sentence case.

source/editor.lisp
```diff
@@ -0,0 +1,45 @@
...
+                                     (file buffer)
+                                     :if-exists if-exists))
+
+(defmethod open-file ((buffer editor-buffer) file)
+  (set-content (editor buffer) (uiop:read-file-string file)))
+
+(define-command editor-open-file (&key (buffer (current-buffer)))
+  "Open a file in the internal editor."
+  (let ((file (first (prompt
+                      :prompt "Open File"
```
Ambrevar, line 35: Sentence case.

source/editor.lisp
```diff
@@ -0,0 +1,45 @@
...
+                      :sources (make-instance 'prompter:raw-source
+                                              :name "Absolute File Path")))))
+    (open-file buffer file)
+    (setf (file buffer) file)
+    (setf (title buffer) file)
+    (setf (url buffer) (quri:uri file))))
+
+(define-command editor-write-file (&key (buffer (current-buffer)) (if-exists :error))
+  "Write a file in the internal editor."
```
Ambrevar, line 44: Maybe more specifically "Write the file in BUFFER's internal editor.".

Ambrevar:
Looking great!  I'll test tomorrow, this is exciting!

Nit: Can you split the first commit and first add a commit where you change the buffer-make arguments?  This would be clearer and help with bisecting, in case this change causes any issue.

source/editor.lisp
jmercouris: They inherit from it so that they can implement the protocol.

source/editor.lisp
jmercouris: maybe. It also contains implementations for things outside the scope of editor-mode.lisp

source/editor.lisp
jmercouris: SpOnGeBob CaSe? :-D

source/editor.lisp
jmercouris: There is no package, I'll add it to the class documentation.

jmercouris:
I cannot split the first commit. Why? I need to adapt buffer-make for the editor-buffer to be made. Hence why I left it as one commit.

source/editor.lisp
jmercouris: I've thought about it, and I've decided against it. I would like to leave the name editor.lisp.

jmercouris:
I've made the changes!

jmercouris:
There is quite a bit more room for extensibility within the nx-ace mode (and there is a lot more that I would eventually like to do it), but that does not necessitate any changes on the side of the `editor.lisp` API. I've kept it purposely VERY small and minimalistic for now.

source/editor.lisp
Ambrevar: All modes need a package though.
We can do like with `help`: have both `editor.lisp` and `editor-mode.lisp`.

source/editor.lisp
Ambrevar: I'm not sure I understand.  In any case, I think the docstring deserves an explanation, maybe an example.

source/editor.lisp
Ambrevar: There must be a package for the mode.

source/editor.lisp
Ambrevar: That'd be even better, maybe we can have a mode for that! :)

source/editor.lisp
```diff
@@ -0,0 +1,65 @@
...
+                      :prompt "Open file"
+                      :sources (make-instance 'prompter:raw-source
+                                              :name "Absolute file path")))))
+    (open-file buffer file)
+    (setf (file buffer) file)
+    (setf (title buffer) file)
+    (setf (url buffer) (quri:uri file))))
+
+(define-command editor-write-file (&key (buffer (current-buffer)) (if-exists :error))
+  "Write the FILE of the BUFFER.."
```
Ambrevar, line 57: FILE is not an argument, so it should not be uppercase.  Maybe specify what "file" is exactly?
Typo: double period.

Ambrevar:
> I cannot split the first commit. Why? I need to adapt buffer-make for the editor-buffer to be made. Hence why I left it as one commit.

I don't understand.  If the first commit is that adaptation of
buffer-make, then editor-buffer does not exist yet, so no problem.

source/editor.lisp
jmercouris: File can be multiple things. That is why it is not specified.

jmercouris:
I tried splitting up the packages and I really don't like it. I do not think it is needed that we use a separate package here.

source/editor.lisp
jmercouris: Why?

jmercouris:
If you want, you can show me how you would split it on a branch and we can talk about it. What do you think?

jmercouris:
Maybe if I just move the mode and the commands, maybe that would work. I don't know, hmm...

jmercouris:
OK, I came up with something I am happy with!

source/editor.lisp
Ambrevar: Because that's how mode commands work: their commands rely on the fact that they belong to the mode package.

In particular, this allows us to have commands like `write-file` only appear in `C-space` when the current buffer has `editor-mode`.

source/editor.lisp
Ambrevar: Misunderstanding?  What I meant is that is that uppercase is reserved to function arguments by convention.

To clarify the docstring, I'd write something like "Write to disk the file associated to the BUFFER."

Ambrevar:
OK for the package and mode file, thanks!

Ambrevar:
I'll test as soon as possible.

jmercouris:
Have you had a chance to experience the nirvana that is editor-mode? :-)

Ambrevar:
Finally tested!  It's super exciting!  I like the architecture, seems so trivial it's actually mindboggling.

Some notes:
- `make-editor-buffer` creates an unamed buffer, so it seems invisible in the buffer list.
- The buffer that's created with `make-editor-buffer` is expected to be empty, but actually contains "Loading..." which might be a bug :)
- The editor buffer does not inherit from base-mode, which makes essential bindings like switch-buffer inaccessible.  Same issue with lisp-repl I think.

Good job!

jmercouris:
> Finally tested! It's super exciting! I like the architecture, seems so trivial it's actually mindboggling.
> 
> Some notes:
> 
>     * `make-editor-buffer` creates an unamed buffer, so it seems invisible in the buffer list.

That's true. It is just like `make-buffer`. You are intended to use the command `open-new-editor-with-file` if you want to focus something. Maybe I should set the default URL to "scratch" or something like that?

>     * The buffer that's created with `make-editor-buffer` is expected to be empty, but actually contains "Loading..." which might be a bug :)

Hm, good point. I should remove that text!

>     * The editor buffer does not inherit from base-mode, which makes essential bindings like switch-buffer inaccessible.  Same issue with lisp-repl I think.

Hm. I figured that base-mode would be enabled because of the following definition: 

```
(define-class editor-buffer (internal-buffer)
  ((file :documentation "The file being edited.")
   (default-modes '(base-mode)))
  (:export-class-name-p t)
  (:export-accessor-names-p t)
  (:export-predicate-name-p t)
  (:documentation "Each editor buffer matches a file. Each editor buffer
  contains an editor mode instance."))
```
Of course the user can remove base-mode, but I imagine they will append if they still want these bindings.

> Good job!

Thank you!

jmercouris:
I have made some changes and I believe I have addressed all feedback! :-D

Ambrevar:
Another issue: when there is no editor configured, the editor commands raise an
error.

Should this be handled more gracefully?

jmercouris:
No, this should signal an error. I think instead of avoiding conditions, we should have the conditions shown to the user with the appropriate restarts.

jmercouris:
What do you think?

Ambrevar:
Sure, but we don't have restart support yet!

If we merge this, then we expose commands that are unusable by default
and trigger an unreadable error message.  This is not very accessible.

I suggest we just handle this until we implement restart support.

jmercouris:
OK! Fair enough, I will make a note in the commit.

jmercouris:
Done!

jmercouris:
It is now impossible for user facing commands to attempt to use a command that relies on `(editor buffer)` without it being configured.

Ambrevar:
Looks OK, but shouldn't we also catch this earlier to prevent the
creation of useless buffers?

jmercouris:
Doing so is significantly more complex. We would have to examine the slots of a editor buffer and see if there is a symbol that would point to the creation of a valid editor mode. There are many ways this can go wrong.

Ambrevar:
Let me think about this, I think we've already done this kind of things before.

jmercouris:
Let's not make things too complex. I think this design is more than good enough!

Issue #1233, opened 2021-03-24
Add command to jump to first input field


edgar-vincent:
Hi all,

It has been requested many times, including by myself, so I thought it should be mentioned here.
It would be nice to have a command which focuses the first input field in a given buffer, similar to the one bound to `gi` in Tridactyl and, I assume, other similar implementations. It would make it easier to quickly jump to a search field, for instance in a search engine.

Thanks a lot!

EV

Ambrevar:
Good idea!  This is relatively easy to implement, so I've tagged this issue appropriately.
Any taker? ;)

aadcg:
I can take a look.  Should there be a keybinding for it @Ambrevar?

Ambrevar:
I suppose `tab` should do?
Seems to be the default in most browsers.

aadcg:
@edgar-vincent your request has been fulfilled.  You will be able to do that by hitting `M-i`, or `Alt+i`.

edgar-vincent:
Thanks a lot! I'll do some testing.

edgar-vincent:
Unfortunately, it doesn't seem to be working on Duck Duck Go.

Ambrevar:
It's a known issue indeed.
The good news: TAB works :)

aadcg:
Indeed, this flaw was documented when it was developed.  The inputs fields in duckduckgo behave rather strange. They lose the focus after events such as switch tab/buffer. You can use `h` to get to duckduckgo's search bar. This functionality is embedded in it.

edgar-vincent:
@Ambrevar @aadcg Sorry about that. Now I remember you mentioning it at the time. At least it's out in the open now :)

PR #1243, opened 2021-03-25
buffer.lisp: add command to focus the first input field.

aadcg:
As request in #1233.

Notice that TAB can't be used as a keybinding since this key is already bounded in GTK.

As usually this misbehaves on duckduckgo. Users can hit the "h" key to get to the search bar though.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1099,3 +1099,12 @@ ARGS are passed to the mode command."
   "Print the current buffer."
   (pflet ((print-buffer () (print)))
          (print-buffer)))
+
+(define-command focus-first-input-field (&key (buffer (current-buffer)))
+  "Move the focus to the first input field of `buffer'."
+  (pflet ((focus () (ps:chain document
+                              (get-elements-by-tag-name "INPUT")
+                              (item 0)
+                              (focus))))
+    (with-current-buffer buffer
+      (focus))))
--- a/source/base-mode.lisp
+++ b/source/base-mode.lisp
@@ -58,7 +58,8 @@
                      "C-shift-t" 'reopen-buffer
                      "C-T" 'reopen-buffer
                      "C-p" 'print-buffer
-                     "C-x C-f" 'open-file)
+                     "C-x C-f" 'open-file
+                     "M-i" 'focus-first-input-field)
 
                     scheme:emacs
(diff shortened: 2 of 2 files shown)


source/buffer.lisp
```diff
@@ -1099,3 +1099,12 @@ ARGS are passed to the mode command."
   "Print the current buffer."
   (pflet ((print-buffer () (print)))
          (print-buffer)))
+
+(define-command focus-first-input-field (&key (buffer (current-buffer)))
+  "Move the focus to the first input field of `buffer'."
```
Ambrevar, line 1104: Convention: Surrounding with "`...'" is for defined symbols.  For arguments, don't surround with quotes and instead use ALL-CAPS.

source/buffer.lisp
```diff
@@ -1099,3 +1099,12 @@ ARGS are passed to the mode command."
...
+  (pflet ((focus () (ps:chain document
+                              (get-elements-by-tag-name "INPUT")
+                              (item 0)
+                              (focus))))
```
Ambrevar, line 1108: I think there is an HTML attribute that defines the default focus.
There is this: https://www.w3docs.com/snippets/html/how-to-set-default-html-form-focus-without-javascript.html
but somehow I remember something else.  @jmercouris?

@aadcg I suggest first checking for the default element and if there is none, then use your `focus` code.  Does that make sense?

aadcg: I'm confused about your suggestion.  Doesn't this `autofocus` set the focus to the first input field when a page is loaded?

Ambrevar: Duh, sorry, that was very dumb of me indeed!
All good then!

aadcg:
We didn't discuss the keybinding I chose, `M-i`, but I assume you're ok with it :)

Ambrevar:
I'm OK with it.

I want to review all bindings before 2.0, so there will be a time to
re-discuss it! :)

aadcg:
I also didn't apply the documentation style you suggested, though I did it on my local checkout. Feel free to do that change. Thanks.

aadcg:
> I want to review all bindings before 2.0, so there will be a time to
re-discuss it! :)

Oh, I like that.

PR #1244, opened 2021-03-26
Fix three instances of overide-map spelling error

hendursaga:
First pull request :)


aadcg:
Thank you for bringing this up and congrats on your first pull request :)

Ambrevar:
Ouch, copy-paste typos can be embarassing :)  Thanks for the fix!

Issue #869, opened 2020-07-27
Feature request: Crawling search?


aartaka:
I'd be nice to have a command to search the content of current page and of all the pages that links from the current page lead to.

Example: If I want to find the signature of `webkit_input_method_underline_copy` on [WebKitGTK2 documentation website](https://webkitgtk.org/reference/webkit2gtk/stable/index.html), I need to click links for class documentation in random order and hope I'll find the needed signature in less than `O(n)` worst-case time (or even exponential time if documentation pages are nested). If there was this crawling search command, then I'd be able to just search for the function name and discover it in [InputMethodContext](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitInputMethodContext.html) class documentation in strictly `O(1)` worst-case time. 

Such a feature sounds like a great productivity boost to me (especially in case of documentation websites that tend to be deeply nested and unobvious to navigate), while it seems to be relatively straightforward to implement on top of minibuffers, tree-like structures, `with-result` and suggestion functions, all of which are more or less implemented in Nyxt.

jmercouris:
This is a great idea, and this is on our todo list :-)

We call it recursive search!

aadcg:
Is anyone working on this? I'd like to give it a try.

Ambrevar:
I'm not! :)

jmercouris:
I am also not working on it, feel free to give it a shot :-)

aadcg:
I've thought about this plenty of time and I'm, arguably, in a better position to pursue this goal. I've drafted implementations too. While it's not a super high priority (when compared to #2989), it will probably be the next big feature I'll contribute.

Issue #55, opened 2018-01-10
Buffer splitting/tiling


dunkyp:
Is anyone else interested in splitting windows emacs style. I've implemented a super simple prototype in the cocoa version (splitting window in half horizonally and manually updating the window url from the repl) and it seems to work nicely.
Does anyone else like this idea. I've included a screenshot of the hack in action, just want to gather opinions on whether it's worth implementing this fully (horizonal and vertical splitting, resizing window switching etc.)

[screenshot](https://i.imgur.com/jx0Xpjy.png)

dunkyp:
Well, that's great news I have a weekend project sorted ;). The implementation is just the simplest possible but I'll push changes once I have them to a fork which I'll create now.

jmercouris:
Ok great, I'm so happy to hear this, you have no idea :)

jmercouris:
@dunkyp Could you please post the code you did to split the screen? Regardless of how incomplete it may be. I'm quite curious to see!

shanemhansen:
@dunkyp  Interested in the split screen as well.

jmercouris:
I'm afraid the implementation has changed so much since this RFC, that it is at the moment entirely impossible. We have to change a lot of things in the port architecture to support this. I would like window splitting myself as well. I don't like to say it, but this is probably something that will most likely have to wait for version 2.0.

Ambrevar:
Hmm... Maybe not.  I gave this some thoughts and I think we might even be able
to roll out a fully-flexible window management without breaking backward
compatibility.

There are more pressing issues at the moment, but I might work on this not to
far in the future.

jmercouris:
If you can think of a simple way to do this, I would be very happy as this is one feature I really would love to have!

I can't ever wrap my head around the tree structures and how to implement things though :\

Ambrevar:
We are very close to being able to have a vertical minibuffer.
A few FFI functions would be enough.

That said, it would be more interesting to implement full-fledged window-manager à-la Emacs.  Then the vertical minibuffer would come for free.

Maybe have a look at Lem / StumpWM to avoid reinventing the wheel.

shaunsingh:
Ran into this bug today, which i quite like
![20230302_13h34m15s_grim](https://user-images.githubusercontent.com/71196912/222521006-86e81cc3-3864-4201-ad50-05de8f6312cd.png)

[...]

PR #1246, opened 2021-03-26
WIP: Recursive search

aadcg:
This is boring and it pretends to be the laziest and dumbest possible
implementation of such a feature.  The idea was to probe what is place
and what might be lacking.  I have done some further exploratory work
beyond this, but it makes little sense to share at this point.

The ingredients of a recursive search are straightforward:
1) get all links of a web page;
2) filter them;
3) open them as buffers;
4) search them.

For 1) to be fully realised, a recursive method with a depth level
parameter is required.  I use the concept of a "nth-neighbour" to
designate the nth level of links on a web page.  The 1st-neighbours of a
web page are all of its links, while the 2nd-neighbours are the links of
the 1st-neighbours.  The 0th-neighbour is a web page, if you will.  You
can also think of this as a tree with a base point which corresponds to
a web page.  The depth level between two nodes is just the number of
edges that sets them apart.

The filtration, 2), is a myriad of things.  They can be divided into
user defined filtrations and others.  The former are easily covered by
the prompt buffer.  The simplest example of the other kind is that web
pages can have links to itself, and that must be excluded.  More
"advanced" filtrations would mean defining similarity metrics between
web pages and do the search on the ones that score higher.  Text rank is
already in place.  This is food for thought at a later stage.

Opening the buffers, 3), is costly.  John mentioned that opening buffers
without CSS would help.  I have no idea how can that be achieved.  I
didn't focus on this part at all.  I wonder what could be done in
another way.  Eww uses libxml2 to fetch the textual content of web pages
for example.  I don't think that opening 100 buffers with the renderer,
and then searching will ever be an option.  That's why I mention HTML
parsers.  Still, there are plenty of use cases where the set of buffers
to be opened is small enough.

Regarding the search, 4), I think there's a lot of work.  The current
search isn't satisfactory.  I'm not criticising the amazing work that is
already in place, but acknowledging something that is clear.  Nyxt isn't
the extensible browser before being an humble browser.  I would:

- write tests for the current search so that things can be tweaked with
  confidence;
- these tests must include human generated html that address plenty of
  corner/extreme cases;
- do a throughout analysis with lots of trial and error to see how it
  can be improved.

Overall I think the search conflates too many things.  A search can be
non-incremental or incremental.  The former isn't implemented, even
though it's arguably simpler.  It's also hard to decouple the
highlighting from the search.  This itches me brutally.

Another topic is the user interface.  The prompt buffer has all the
functionality needed.  The kind of interface I have in my mind is smth
along the lines of XREF or OCCUR in the emacs world.  It makes little
sense to write smth new when the prompt buffer is already a kitchen
sink.  The question is: is it easy to put the contents of the
prompt-buffer into a buffer where it can make use of all the screen
height available.

We can also think about the applications of such a feature.  I got
interested because I have smth particular in mind that would perhaps
attract the mathematical community.  Actually, the spirit and motivation
is EXACTLY what Artyem described
[here](https://github.com/atlas-engineer/nyxt/issues/869).  Another
interesting thought is that this would help us to have an info-like
interface to manuals.  Including the Nyxt manual, when being read from
Nyxt.  In this case, an "all" parameter of the recursive method would
make lots of sense.  This parameter would do the obvious, i.e. descend
into the depths of infinity and beyond.

You can test the current command from a page like this
[one](https://nyxt.atlas.engineer/article/diff-mode.org).  It sucks in
multiple ways, I know that.

If you're curious, the `list-linked-pages` is able to get the 5460 links
of the wikipedia USA page in 2.5 seconds on my machine.

For me the heart of the problem is to get 1) and 4) right.  On 1) I have
done some work, but it's not finished.  About 4), it's not up to me to
make such strategic decisions.

Let me know what you think.


aadcg:
Just in case anyone wants to see what I've been up to and wants to give advice.

Ambrevar:
@jmercouris Do you have time to review this?

jmercouris:
I can try to get to this tomorrow.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
                                        :case-sensitive-p case-sensitive-p
                                        :buffer buffer))
                       buffers))))
+
+;; No significant benefits noticed in using fixed-sized arrays vs lists.  To be
```
jmercouris, line 207: Lists are very well optimized in Lisp. I would not worry about the costs on the Lisp side. The most expensive operations by a very wide margin occur within GTK/C code.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+;;     (with-current-buffer buffer
+;;       (let* ((len (1- (parse-integer (nlinks))))
+;;              (array (make-array len :initial-element "")))
+;;         (dotimes (i len)
+;;           (setf (elt array i) (fetch-link i)))
+;;         array))))
+
+(defun 1th-neighbours-from-buffer (buffer)
+  "TODO"
+  ;; it can be assumed that url1 and url2 are of quri type
```
jmercouris, line 221: if you want to ensure a specifc type you can do a compiler declaration using `declaim`. There are great examples in the codebase to look at :-)

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+  (defun distinct-web-page-p (url1 url2)
```
jmercouris, line 222: We have a `equal-url` in bookmark.lisp which you may consider using. If you think `equal-url` does not belong in bookmark.lisp, please feel free to move it.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+    (not (equalp (quri:uri-path url1) (quri:uri-path url2))))
+
+  (defun shared-scheme-p (url1 url2)
+    "TODO"
+    ;; either share the scheme (think of lisp://) or belong to the http, https
+    ;; equivalence class.
+    (or (equalp (quri:uri-scheme url1) (quri:uri-scheme url2))
+        (and (quri:uri-http-p url1) (quri:uri-http-p url2))))
+
+  (defun shared-domain-p (url1 url2)
```
jmercouris, line 235: I would call this `equal-domain-p`, or something a long those lines. Shared makes me imagine shared state between two objects.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+(defun 1th-neighbours-from-buffer (buffer)
+  "TODO"
+  ;; it can be assumed that url1 and url2 are of quri type
+  (defun distinct-web-page-p (url1 url2)
+    "TODO"
+    ;; perhaps slower than a much simpler solution that just checks for "#" in
+    ;; both strings?
+    (not (equalp (quri:uri-path url1) (quri:uri-path url2))))
+
+  (defun shared-scheme-p (url1 url2)
```
jmercouris, line 228: I would call this `equal-scheme-p`. Same as written below, it makes me imagine shared state.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+  "Apply FUNCTION on ARG for N times."
+  (if (zerop n)
+      arg
+      (nth-power-function function (1- n) (funcall function arg))))
+
+  (nth-power-function '1th-neighbours-from-list depth url-list))
+
+;; Interfaces
+
+;; There should be 3 commands.  For all the them, the user passes the depth k.
```
jmercouris, line 289: This appears logically sound.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+;; he wished to open.  Then the search happens.
+
+;; recursive-search-from-buffers
+;; The user chooses the buffers and the hinting mechanism passes the list of
+;; links.
+
+;; recursive-search-from-buffer
+;; As above but the hinting mechanism happens in the current buffer.
+
+;; recursive-search-from-current-buffer
```
jmercouris, line 301: I'm not sure I understand the difference between `recursive-search-from-buffer` and `recursive-search-from-current-buffer`, they appear to do the same thing. Could you please explain?

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+            ;; Computed on every cycle
+            for url2 = (quri:uri (fetch-link i))
+            ;; Computed a single time
+            with url1 = (slot-value buffer 'url)
+            collect (when (and (shared-domain-p url1 url2)
+                               (distinct-web-page-p url1 url2)
+                               (shared-scheme-p url1 url2))
+                      url2)))))
+
+(defun 1th-neighbours-from-list (url-list)
```
jmercouris, line 254: I see, so you are:

+ given a list of URLs
+ collect all URLs contained within the pages at the list of URLs

So if we had url list ("a.com" "b.com")

You would visit a, and b, collect all of their lines, and return it as a list? Is that correct?

jmercouris:
Thanks for sharing Andre. I can see that you have been very hard at work! It seems to me that you have thought of all possible angles!

I think I understand your approach. Seems logical. I would still encourage you if possible to submit a single depth example so that we can try out the UI and see what it feels like. We could figure out what the performance is like, and any possible issues so that we can remedy them before implementing arbitrary depth search.

source/search-buffer.lisp
aadcg: Good question. The recursion mechanism I wrote is very general. It lets you start off from a single URL or a bunch of them!  This is the key observation.  How does that translate in terms of interfaces?

`recursive-search-from-current-buffer` passes a single URL, the current one. This is, by far, the most common case. Think of it as starting a recursive search with the current URL as the base of the tree.

Now we bring 2 more layers of chaos. We can think of buffers, instead of a single one (the current one).  And we can think of passing SOME links in any given buffer, via the hinting mechanism!  I thought this was what you had in mind when you mentioned the hinting mechanism. Perhaps I got it wrong.  To me mind, it makes little sense to bring the hinting mechanism into play.  But it is very possible. I'd stick with one source of chaos only, that of buffers.

aadcg:
> It seems to me that you have thought of all possible angles!

That's me!

> I think I understand your approach. Seems logical. I would still encourage you if possible to submit a single depth example so that we can try out the UI and see what it feels like. We could figure out what the performance is like, and any possible issues so that we can remedy them before implementing arbitrary depth search.

That was indeed my plan, but it got frustrated due to a big issue I'm currently working on. Basically, I can only fetch the links of URLs when they're fully loaded (since I need JS). I didn't have this issue before because I was working on depth 1. Notice the hack on `1th-neighbours-from-list` using `sleep`. I know that `on-signal-load-finished` exists, but that means turning research search into a mode, which I couldn't make much sense of. I'd like to be able to run something when a web buffer is fully loaded. What are your thoughts?

I'm currently cleaning this up.

jmercouris:
I would keep polling the web buffer on a separate thread until it is ready as an interim hack. The correct thing would be to implement some sort of listener model as you've suggested. I also don't have ideas beyond the usage of a hook :-\.

aadcg:
My first entry on this pull request was updated and might be worth reading.  It's crucial for me, at least, to understand where I am and where I'm going. Plenty of things to do, in progress. Thanks for the feedback @jmercouris, I'll address it.

source/search-buffer.lisp
aadcg: Right.

source/search-buffer.lisp
aadcg: I see, thanks!

source/search-buffer.lisp
aadcg: I was aware of it but it doesn't suit the case at hand.  

```
(equal-url (quri:uri "https://en.wikipedia.org/wiki/Potato#Etymology")
           (quri:uri "https://en.wikipedia.org/wiki/Potato#Characteristics"))
```

returns `nil`, which makes sense for bookmarks, but not in this case.  

It would be a good idea to have a `url-utils.lisp` perhaps and to move `equal-url` and company there.

source/search-buffer.lisp
Ambrevar: What's "1th"?

source/search-buffer.lisp
Ambrevar: As mentioned in #1283 I think it should rather be `scheme=` for consistency with the standard.

source/search-buffer.lisp
Ambrevar, line 222: Looks like you are callin `defun` inside a `defun`.  If you want to define functions locally in Common Lisp, this won't work, because the function `distinct-web-page-p` will be defined globally when the function gets run! :/ 

(Note that it works in Scheme.)

Instead, consider using the standard `flet`/`labels` or wrapping the body of your function in a `(sera:local ...)` form.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+      ;; The filtration must happen here.
+      (loop for i from 0 to (1- (parse-integer (nlinks)))
+            ;; Computed on every cycle
+            for url2 = (quri:uri (fetch-link i))
+            ;; Computed a single time
+            with url1 = (slot-value buffer 'url)
+            collect (when (and (shared-domain-p url1 url2)
+                               (distinct-web-page-p url1 url2)
+                               (shared-scheme-p url1 url2))
+                      url2)))))
```
Ambrevar, line 252: Are you sure you when to put the `when` condition here?
It will collect `nil` for URLs that don't satisfy the condition.
If this is not what you intend, use `loop`'s `when` form instead.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+
+(defun 1th-neighbours-from-list (url-list)
+  ;; Gets me rid of both the nils and the "nestedness".
+  (alexandria:flatten
```
Ambrevar, line 256: While this works, you may be able to avoid this call if you don't collect the `nil` in the first place and use `alex:mappend` at the right spot.

source/search-buffer.lisp
```diff
@@ -203,3 +203,129 @@
...
+;; At depth k, the k-1 level must accessible (i.e. buffers up to that level
+;; must be made).
+
+(defun recursive-links (url-list depth)
+  "TODO"
+  ;; In dynamical systems, f^n denotes the f composed with itself n times.  That's
+  ;; the inspiration for the name.
+  (defun nth-power-function (function n arg)
+  "Apply FUNCTION on ARG for N times."
+  (if (zerop n)
```
Ambrevar, line 281: What happens when n is negative? :p

source/search-buffer.lisp
aadcg: First. This will be renamed though. 

I thought there would be 1th, 2nd, 3rd, etc, but then I understood that that wouldn't be the case.

source/search-buffer.lisp
aadcg: Makes sense.

source/search-buffer.lisp
aadcg: Yes.

source/search-buffer.lisp
Ambrevar: OK.
Note that "first" is abbreviated "1st" ;)

Ambrevar: Also "first" is better than "1st", abbreviations are frowned upon because then the user would have to guess the name when searching.

aadcg: Ah, now I understand the question. I had "N-th" in mind hence the mistake.

source/search-buffer.lisp
aadcg: I understand your point. But I actually would like to get the nils for debugging purposes. At least for now.

source/search-buffer.lisp
aadcg: I see. I got carried away since I've been looking at some `define-parenscript` instances, but there it's obviously fine since you're actually defining a JS function.

source/search-buffer.lisp
aadcg: I see. Thanks.

source/search-buffer.lisp
aadcg: Fair enough.

source/search-buffer.lisp
aadcg: The difference between `foo=` and `foo-equal` is that the former is case sensitive and the latter isn't, correct? At least, when `foo` is `string`. Following the standard, shouldn't it be `scheme=` instead of `scheme-equal`?

source/search-buffer.lisp
jmercouris: That seems sound.

source/search-buffer.lisp
jmercouris: string-equal string=, etc are all just aliases. They were integrated into Common Lisp to consolidate the different Lisp vendor permutations.

Ambrevar: No there are not, `string-equal` is the case-insensitive version of `string=`.

Here since case does not matter for schemes, I suggest to just have `scheme=`.

jmercouris: Hm, that is inconsistent with my other knowledge then... well, good to know!

Ambrevar: Same with `char-equal` and `char=` by the way.

aadcg:
@Ambrevar, I'd like to ask for your advice. I've implemented the hook mechanism to run code when buffers are done loading. For my particular case, I need to run some JS when a buffer is loaded, but I also need the values returned by that function. At the moment I'm not able to do that.  @jmercouris hinted that this could perhaps be achieves with calispel.  I'm wondering if you could give me a hand.


Take a look at commit 4761adcc, at the function `get-links-from-url`. I'm able to run `nyxt::get-links`, but I need the values returned by that procedure. How would you approach this? Thanks.

Ambrevar:
OK, will look at it tomorrow!

aadcg:
Thank you @Ambrevar.

If you want to see it perform you can already do it. Call `recursive-search-from-current-buffer` from the prompt-buffer with depth 1 [here](https://nyxt.atlas.engineer/article/diff-mode.org).

All of the RELEVANT issues are yet to be addressed, but it's getting real close as soon as the issue with running js as soon as a buffer gets loaded will be sorted out.

aadcg:
My first entry was again edited.

Ambrevar:
Hey André, sorry I was unexpected out all day, I didn't have time to
look at this.  I'll come back to it as soon as possible.

aadcg:
> Hey André, sorry I was unexpected out all day, I didn't have time to
> look at this. I'll come back to it as soon as possible.

No worries Pierre!  I always have plenty of things to do anyway, so this
can be addressed whenever you find some time.  À tout à l'heure!

Ambrevar:
> The prompt buffer has all the functionality needed. Is it easy to put its contents into a window/buffer/pane where it would use all of the screen height available? Probably Pierre could help.

This is actually our oldest issue: #55.
I want to work on it soon, but probably not before 2.0.
Once we have window management, we can display the prompt buffer in arbitrary ways, including vertically left or right.

source/browser.lisp
```diff
@@ -414,7 +414,7 @@ Deal with REQUEST-DATA with the following rules:
          ;; webkit-policy-decision-ignore on a download requestion.
          ;; To work around this, we set the `load-status' to a value other than
          ;; `:loading'.
-         (setf (slot-value buffer 'load-status) :finished)
+         (setf (load-status buffer) :finished)
```
Ambrevar, line 417: `load-status` does not have a setter on purpose, because it's a fragile API detail that we will change in the future.  You should reverse this hunk.

source/buffer.lisp
```diff
@@ -231,7 +236,7 @@ The mode instances are stored in the `modes' slot.")
                               :finished
                               :unloaded
                               :failed)
-                :accessor nil
+                :accessor load-status
```
Ambrevar, line 239: If you define the setter yourself below, then don't define the accessor here, only define the reader.
But considering this is an implementation detail, I would still not define the reader.  Leave `accessor` to nil and define the writer, then use `(setf (load-status ...))` as you did.

source/browser.lisp
Ambrevar: See below my comment on the accessor.

source/search-buffer.lisp
```diff
@@ -206,3 +206,92 @@
...
                                        :buffer buffer))
                       buffers))))
+
+(defun make-search-buffers (url)
+  "Make buffer corresponding to URL with as little overhead as possible."
+  ;; TODO These should be made without loading CSS.
+  (make-buffer :modes '(web-mode
+                        base-mode
+                        noimage-mode)
+               :url (nyxt::ensure-url url)))
```
Ambrevar, line 216: OK for testing, but in the future we should include other modes like proxy-mode, etc., as per the current buffer.  Otherwise it may fail or leak users private data.

source/search-buffer.lisp
```diff
@@ -206,3 +206,92 @@
...
+(defun get-links-from-url (url)
+  "Return a list of links from URL.
+If there is a buffer corresponding to URL, then fetch the links from it.
+Otherwise, create a new buffer to fetch the links."
+  (let ((buffer-exists-p (first (buffer-list :url (nyxt::ensure-url url)))))
+    (if buffer-exists-p
+        (nyxt::get-links buffer-exists-p)
+        (let ((buffer (make-search-buffers url)))
+          (sera:add-hook (buffer-loaded-hook buffer)
+								         (nyxt::make-handler-buffer
```
Ambrevar, line 227: Untabify.

source/search-buffer.lisp
```diff
@@ -206,3 +206,92 @@
...
+                          (lambda (buffer)
+                            (nyxt::get-links buffer)
+                            ;; Doesn't seem to work...
+                            ;; How to run the hook handler just once?
+                            (sera:remove-hook (buffer-loaded-hook buffer)
+	  								                          'get-links))
```
Ambrevar, line 233: Untabify.

Ambrevar, line 233: You can't modify the hook while it's running.
Can you detail why you want to modify it?

To stop the crawl,
it's probably easier to store the status in a buffer/mode local variable and have get-links look at this variable to decide whether it should proceed or not.

source/search-buffer.lisp
```diff
@@ -206,3 +206,92 @@
...
+
+(define-command recursive-search-from-current-buffer (&key case-sensitive-p)
+  "Start a search on URLs that are linked in the current buffer, with arbitrary
+DEPTH."
+  (let* ((url-list (list (url (current-buffer))))
+         (depth (parse-integer
+                 (first (prompt
+                         :prompt "Depth"
+                         :sources (make-instance 'prompter:raw-source)))
+                 :junk-allowed t))
```
Ambrevar, line 276: Should probably default to some sensible value, like 1 or 2.

source/search-buffer.lisp
```diff
@@ -206,3 +206,92 @@
...
+         (last-level-url-list (prompt
+                               :prompt "Choose URLs"
+                               :sources (make-instance 'prompter:source
+                                                       :constructor (recursive-links depth url-list)
+                                                       :name "Choose URls"
```
Ambrevar, line 281: Typo: `URLs`.

Source names should not contain verbs though.

Ambrevar:

> Take a look at commit 4761adcc, at the function `get-links-from-url`. I'm able to run `nyxt::get-links`, but I need the values returned by that procedure. How would you approach this? Thanks.

Commit 4761adcc is gone, but I had a look at get-links-from-url
nonetheless.
`ffi-buffer-evaluate-javascript`, `pflet` and thus `get-links` all run
synchronously, and thus return a value as expected.

So is your problem to get the return value of get-links _when run from a
hook_?

If so, I think what you need to do after sera:add-hook is wait on a
channel with `calispel:?` and in the handler write to this channel with
`calispel:!`.

You can create a channel with `make-channel`.

Does that make sense?

I've just tested: I tried with a depth of 1, it seems to work already!
Of course there are a few things to iron out, but we can discuss this
progressively.

Good job so far!

aadcg:
> Commit 4761adc is gone
 
Indeed.  I didn't update it, sorry about that!

> So is your problem to get the return value of get-links _when run from a
hook_?

> If so, I think what you need to do after sera:add-hook is wait on a
channel with `calispel:?` and in the handler write to this channel with
`calispel:!`.

Ok, this helps me.  

There's still a question: is it possible to run a handler once only?

Ambrevar:
> There's still a question: is it possible to run a handler once only?

Have you tried `sera:once`?

aadcg:
> Have you tried `sera:once`?

Thanks.  I solved this issue, but others got in the way.

Can you think of a simple website, i.e. with very few links? Think Nyxt's website (which is actually quite simple) but yet simpler.

Ambrevar:
You mean the entire host or just a web page?
Maybe

https://0x0.st/

?

Otherwise you could write your own HTML local file.

aadcg:
> Otherwise you could write your own HTML local file.

Yes, I did smth along those lines using the `lisp` scheme. I have to figure it out if I can do this with http(s).

```
(in-package :nyxt)

(defun page1 ()
  (with-current-html-buffer (buffer "*Page1*" 'base-mode)
    (markup:markup
     (:p (:a :href (lisp-url '(page2)) "link-to-page2")))))

(defun page2 ()
  (with-current-html-buffer (buffer "*Page2*" 'base-mode)
    (markup:markup
     (:p (:a :href (lisp-url '(page1)) "link-to-page1")))))
```

Ambrevar:

>> Otherwise you could write your own HTML local file.
>
> Yes, I did smth along those lines using the `lisp` scheme. I have to figure it out if I can do this with http(s).

Why do you ask about HTTP?  Is the scheme relevant to recursive search?

To write HTML local files, you actually write the HTML in a file, then
load it from Nyxt, e.g. with the `file://` scheme.

aadcg:
There are things I still need to address here, namely:

- [ ] address @Ambrevar input;
- [ ] write some unit tests I have in mind;
- [ ] undo some half-assed things I kept in the code for the sake of the demo;
- [ ] `get-links` must filter links to downloads.

Before you isn't just a recursive search, but a powerful engine that computes URLs at a given distance or depth (see the definition above).

One of the applications is indeed recursive search. Try out `recursive-search-from-current-buffer`. Notice the endless and interesting possibilities that `recursive-links` unfolds.

source/buffer.lisp
aadcg: I'm trying to follow but I'm afraid I'm lost at this point.  I tried to to what you suggest -- take a look at 76403fc5. But, if I keep things this way, I can't get the links as soon as a buffer gets loaded. 

However, when I bring the assessor back and change all occurrences of `(slot-value buffer 'load-status)` into `(load-status buffer)`, I get the desired result.

What am I missing?

@jmercouris, isn't my approach aligned with what you suggested me?

source/search-buffer.lisp
aadcg: What does it mean?

source/buffer.lisp
aadcg: Ok, I think I have understood it. Take a look at `05968e00`. Is that you meant @Ambrevar ?

source/search-buffer.lisp
aadcg: Within the context of making buffers for the recursive-search, I agree that it makes sense to pass the modes of the current buffer to the new ones.

aadcg:
It's clear that the real bottleneck here is the fact that buffers need to be made to compute the links of a given URL. What can be done in that regard?

The following computes all of URLs that are at distance (length) 0, 1, 2 and 3 from `https://nyxt.atlas.engineer/`. 78 URLs were computed in those conditions.

```$ (time (nth-value 1 (recursive-links 3 (list (ensure-url "https://nyxt.atlas.engineer/")))))```

```
Evaluation took:
  75.608 seconds of real time
  22.345493 seconds of total run time (8.832316 user, 13.513177 system)
  [ Run times consist of 0.090 seconds GC time, and 22.256 seconds non-GC time. ]
  29.55% CPU
  166,942,078,582 processor cycles
  960 page faults
  334,371,424 bytes consed
```

**Comment by jmercouris** [comment]
There is no way around this bottleneck. You can maybe lazily load the buffers so that the user can begin typing immediately and the results are populated. You can also fetch the HTML content only (which will not work for many websites). You could fetch the HTML only as a first measure and then lazily make the buffers.

**Comment by aadcg** [comment]
> You can maybe lazily load the buffers so that the user can begin typing immediately and the results are populated. 

That's a separate issue. I'm not thinking about the search, but about computing URLs at depth N. In the search, the issue is that one can't have many buffers at the same time. In computing URLs, I need to open buffers (only to close them as soon as loaded) fast.

> You can also fetch the HTML content only (which will not work for many websites). 

Do we have that?  I think we can only get the HTML when the buffer exists.

**Comment by aadcg** [comment]
Another question regarding filtering URLs that correspond to downloads.  It's possible to check if a given URL ends with a given file extension, but that's actually naive -- take `https://nyxt.atlas.engineer/article/sharing-files.org`.

Any ideas?

File: source/buffer.lisp (hunk 13 above)
Comment (aadcg) [reply source/buffer.lisp]: Something weird is happening when I load web buffers from the initial buffer.

```
When attempting to
set the slot's value to :LOADING (SETF of SLOT-VALUE), the slot
NYXT::LOAD-STATUS is missing from the object
#<USER-INTERNAL-BUFFER {100A29B233}>.
```

Internal buffers don't have a load-status, since it's a slot of `web-buffer`. What's going on here?

**Comment by jmercouris** [comment]
It is not possible to know by the URL itself alone whether it corresponds to a download. You would have to probe the file, and open the file to make sure it is what it claims to be.

Additionally, whether a URL corresponds to a download is a vague notion. When we visit a web page, we are in fact downloading the page to disk and then viewing it in our browser. The key difference is that our browser only "downloads" files into the download directory that it cannot natively render.

**Comment by jmercouris** [comment]
With regard to getting the HTML only when the buffer exists: You can use dexador to retrieve the page contents: https://github.com/fukamachi/dexador .

**Comment by jmercouris** [comment]
To clarify; you do not need to create a buffer to use dexador :-)

**Comment by jmercouris** [comment]
Hopefully that helps!

**Comment by aadcg** [comment]
> The key difference is that our browser only "downloads" files into the download directory that it cannot natively render.

Right.

> Hopefully that helps!

Yes, thanks!

File: source/search-buffer.lisp (hunk 17 above)
Comment (aadcg) [reply source/search-buffer.lisp:218]: When you say default, you mean that some sensible value should be available in the prompt buffer to choose from?

I tried the following, but it doesn't work. What am I missing?

```
(prompt
 :prompt "Choose Depth"
 :sources (make-instance 'prompter:raw-source
                         :constructor (list 1 2))
 :hide-suggestion-count-p t)
```

source/buffer.lisp
aadcg: This seems to be related to commit `43d0600e8177ce4dfa190ee7414a84cea7c2b3ed`, that @Ambrevar authored. Maybe you can help me figure this out before this gets merged. I'm ignoring this for now and moving on with my work.

Ambrevar: Indeed, I introduced a bug in this commit, but @aartaka fixed it yesterday.  Can you rebase against master and try again?

Ambrevar: @aadcg Sorry, your previous message fell under my radar.  You asked me to take a look at 
https://github.com/atlas-engineer/nyxt/commit/fd7f96f6bcfc56a186e84e65304f94c4981b347d but it seels to still have the reader.  What exactly should I be looking at?

aadcg: Sure, no worries!

aadcg: So you suggest having no reader and no accessor, right?

Ambrevar: For now, until we can figure out a proper data flow / structure for the buffer status.

Ambrevar: If you think this is too limiting, we can start working it out now.  It's been in this limbo state for too long.

aadcg:
> If you think this is too limiting, we can start working it out now. It's
> been in this limbo state for too long.

It might be a good idea since I'm bringing a hook that runs as soon as a
buffer gets loaded.

But, exactly right now, I'm focusing on wrapping up the more pressing
issues surrounding recursive links/search.

jmercouris:
Will review soon!

aadcg:
> Will review soon!

Thanks. I'd prefer to go through the review process live rather than textually. 

If you agree, I'm available whenever you feel comfortable with the material.  

My first post was again updated.

jmercouris:
OK, we can do it on Monday then :-)

aadcg:
What's new (roughly):

- Add `buffer-loaded-hook`.  A hook enabling the execution of arbitrary
  code as soon as a web buffer gets loaded;

I hardly understand the comments given by @ambrevar, so feel free to
take the lead.  There's smth fishy going on, since I had to use
`bt:make-thread` to run parenscript when the buffer gets loaded.  Maybe
the thread doesn't need to be there in other cases.  Or maybe I could
make this thread at `renderer-get-links`.  Not tested.

Needed to fetch links as soon as a given web buffer was loaded
(`renderer-get-links`).

- Extend `buffer-list`.  Check if a buffer corresponding to a given URL
  exists.  Needed for `renderer-get-links`.

  I suggest that `buffer-list` should be refactored.  Optionally, a
  function could be passed that would filter through `buffer-list`,
  i.e. `(alex:hash-table-values (buffers *browser*))`.

- Add `get-links`.  A method that uses the renderer to fetch the links
  of any buffer.  It's possible to filter the fetched links arbitrarily.

- Add `renderer-get-links`.  A method that gets the links of a given URL
  using the web render, i.e. using `get-links`.  It's possible to filter
  the fetched links arbitrarily.

- Add `http-get-links`.  A method that gets the links of a given URL
  using the dexador library.

  There's an issue with the quri library, see
  https://github.com/fukamachi/quri/issues/47 that results in a minor
  bug in this method.

  It also doesn't handle at all urls that can't be reached or urls that
  download things.  This is also the case for `http-get-links`.

- Add `distinct-url-p`, `scheme=`, `domain=` and `host=`.  These are url
  utilities that compare two URLs and return a boolean.  Needed for
  filtering in `get-links`, `http-get-links` and `renderer-get-links`.

  There's a problem in `renderer-get-links`.  It doesn't directly make
  use of the optional argument `filtration-predicate`.

- Add `recursive-links`.  A rigorous engine to compute URLs at arbitrary
  depth.

  Notice that it relies on the methods that fetch links,
  `http-get-links` and `renderer-get-links`, which are shaky for the
  reasons mentioned above.

- Add `interactive-recursive-links`.  A method that specialises
  `recursive-links`.  It lets the user filter through the URLs at each
  depth level.

- New file `recursive-search.lisp` that tests `recursive-links`.

- Add `recursive-search`.  A method that lays the foundations for
  methods similar to `search-buffers`, coupled with the powerful engine
  of `interactive-recursive-links`.

- Add `recursive-search-from-current-buffer`.  A specialisation of
  `recursive-search` that covers a frequent use-case.

- Add `active-modes`.  A method that returns the enabled modes of a
  given buffer.  The mode symbols, not the class instances of those
  modes.  Needed for `recursive-search`.

- Move `match-*` from `browser.lisp` to `urls.lisp`.

--

The foundation, `recursive-links`, is solid.  Because of that, the new
command `recursive-search-from-current-buffer`, works reasonably well.
Give it a try with arbitrary depth, on any URL.  You might be surprised.

@Ambrevar @jmercouris @aartaka 

There are lots of half-assed things too.  I'm not either motivated or
the right person to solve some of those issues so help is welcome.

I need to go through things again, write proper docstrings, cleanup
comments, fit things on my mind, make decisions, etc.

A manual section about this feature could be made from my first post
with some effort.

aadcg:
This can't be merged since master is broken. See [this](https://github.com/atlas-engineer/nyxt/pull/1314#issuecomment-826711246).

Ambrevar:
Sorry for answering a bit late here.  One thing I didn't understand:

> I hardly understand the comments given by @Ambrevar, so feel free to take the lead. 

Which comments and take the lead on what?  Please let me know how I can help.

aadcg:
> Which comments and take the lead on what? Please let me know how I can help.

No worries! You've helped already. My doubts were related to the implementation of #1332.

jmercouris:
I'm closing this for now, there are many conflicts and it would be unreasonably difficult to merge this. This work is saved on a branch titled `recursive-search`.

PR #1253, opened 2021-03-29
buffer.lisp: Fix command to focus the first input field.

aadcg:
Input tags have types.

In order to focus the first (in the visual sense) input field, some of
the tags need to be skipped according to their type.

--

@edgar-vincent continue reporting bugs please :)

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1097,10 +1097,9 @@ ARGS are passed to the mode command."
   (pflet ((print-buffer () (print)))
          (print-buffer)))
 
-(define-command focus-first-input-field (&key (buffer (current-buffer))
-                                         (type-blacklist '("hidden"
-                                                           "checkbox"
-                                                           "button")))
+(define-command focus-first-input-field (&key (type-blacklist '("hidden"
+                                                                "checkbox"
+                                                                "button")))
   "Move the focus to the first input field of `buffer'."
   ;; The list of input types can be found below.
   ;; https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input
@@ -1112,9 +1111,8 @@ ARGS are passed to the mode command."
                                          (get-elements-by-tag-name "INPUT")
                                          (item (ps:lisp i))
                                          (focus))))
-    (with-current-buffer buffer
-      (nth-input-focus (do ((i 0 (1+ i)))
-                           ((notany
-                             (lambda (type) (equalp (nth-input-type i) type))
-                             type-blacklist)
-                            i))))))
+    (nth-input-focus (do ((i 0 (1+ i)))
+                         ((notany
+                           (lambda (type) (equalp (nth-input-type i) type))
+                           type-blacklist)
+                          i)))))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1021,9 +1021,9 @@ ARGS are passed to the mode command."
    (buffers :initarg :buffers :accessor buffers :initform nil)
    (prompter:multi-selection-p t)
    (prompter:constructor (lambda (source)
-                           (delete-duplicates 
-                            (alex:mappend 
-                             #'modes 
+                           (delete-duplicates
+                            (alex:mappend
+                             #'modes
                              (uiop:ensure-list (buffers source)))
                             :test (lambda (i y) (equal (mode-name i)
                                                        (mode-name y)))))))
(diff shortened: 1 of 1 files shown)


source/buffer.lisp
```diff
@@ -1097,11 +1097,25 @@ ARGS are passed to the mode command."
...
+                                                           "checkbox"
+                                                           "button"
+                                                           "search")))
   "Move the focus to the first input field of `buffer'."
-  (pflet ((focus () (ps:chain document
-                              (get-elements-by-tag-name "INPUT")
-                              (item 0)
-                              (focus))))
+  ;; The list of input types can be found below.
+  ;; https://www.w3schools.com/html/html_form_input_types.asp
```
Ambrevar, line 1107: In case you don't know this resource, there is also: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input
I tend to prefer MDN over w3schools, but no strong opinion either :)

source/buffer.lisp
```diff
@@ -1097,11 +1097,25 @@ ARGS are passed to the mode command."
...
+  (pflet ((nth-input-type (i) (ps:chain document
+                                        (get-elements-by-tag-name "INPUT")
+                                        (item (ps:lisp i))
+                                        type))
+          (nth-input-focus (i) (ps:chain document
+                                         (get-elements-by-tag-name "INPUT")
+                                         (item (ps:lisp i))
```
Ambrevar, line 1114: Looks like you are traversing the DOM multiple times.  Shouldn't you collect the inputs tags only once?

source/buffer.lisp
```diff
@@ -1097,11 +1097,25 @@ ARGS are passed to the mode command."
...
+                                         (focus))))
     (with-current-buffer buffer
-      (focus))))
+      (do ((i 0 (1+ i)))
+          ((notany
+            (lambda (type) (equalp (nth-input-type i) type))
+            type-blacklist))
+        (nth-input-focus i)))))
```
Ambrevar, line 1121: Opinion: I prefer `sera:range` (or `series`) over `do`. which allows me to apply functions over a sequence of numbers, instead of writing the body directly.

Ambrevar:
Looks good to me!
Can you just confirm that the double-traversal of the DOM is necessary or not?

source/buffer.lisp
aadcg: I wasn't aware, but thank you for sharing.  I hardly know anything about the web browsers :)

source/buffer.lisp
aadcg: I'm not.  I loop over the inputs tags, check the type, and when I find the right index, I focus it.

source/buffer.lisp
aadcg: I see.  I'm not comfortable with external libraries for the time being but I'll keep that in mind for the future.

source/buffer.lisp
aadcg: Ah, maybe I understood what you mean. I can try to figure it out.

aadcg: Ah, maybe I understood what you mean. I can try to figure it out.

Ambrevar: To clarify, I think it's done multiple time because `get-elements-by-tag-name` appears twice, in `nth-input-type` and `nth-input-focus`.

aadcg: Indeed, this annoyed me brutally and I was trying to work out some solution to no avail. I'm taking a second look.  My issue is that I can't pass a symbol inside `ps:chain` so that I'd have only one method inside `flet`.  Smth like `(nth-input-action i action)` where `action` would be `type` or `(focus)`.

Ambrevar: Passing the action wouldn't remove the double-traversal issue.
It may be that you'd have to do it all in Parenscript if you don't want to traverse twice.

On second read, are you focusing on _all_ inputs that don't match the blacklist?  Shouldn't you just focus on the first?

aadcg: > Passing the action wouldn't remove the double-traversal issue.

Indeed.

> On second read, are you focusing on all inputs that don't match the blacklist? Shouldn't you just focus on the first?

I focus the first input that _doesn't_ match the blacklist.

Ambrevar: Are you sure?

```lisp
(do ()
    (t 17)
    (print 'ok))
; "ok" is not printer.
; => 17
```

aadcg: Sorry, I made a mistake in my previous message indeed. But I don't see any mistake in my pull request.

source/buffer.lisp
Ambrevar: The `do` form stops as soon as the `(notany...)` return non-nil.  So if the first tag is not blacklisted, then nothing is done, it's not focused.

aadcg: It's clear now. See 1cd04f82.

aadcg:
@jmercouris can you provide me a high-level idea on how to avoid the double-traversal of the DOM?  Is re-writing it all in parenscript the only option? Thanks.

jmercouris:
You are right. You'll have to do the notany comparison within parenscript and then just invoke focus on the correct element. That is the simplest way.

jmercouris:
If you are unsure what your parenscript resolves to, you can always `(ps:ps ...)` within the REPL to see the computed javascript. You can then run said javascript within the inspector console of any browser (including Nyxt).

aadcg:
I'm very confused by all of this and I'd prefer not to spend so much time properly learning js and ps all at once. I feel that in the near future I could do this fix easily. For now, that's the best I can do.  Although it's sub optimal, it works.

Ambrevar:
No worries, it's merged!
I've added a comment about the double DOM traversal in
8fb78a4df57376745da911ddeb46d94d2cd64643.

PR #1260, opened 2021-03-30
Library move to build

### LINKED ISSUE #1250: Move librares to _build
I've just reverted the recommendation to store the repo in ~/.local/share/common-lisp/systems: Indeed, it would not find `nyxt.asd` since it looks it does not even recurse in the "nyxt" subdir :p

8ef171fd1eb62d168defe4a2d7115393230314d1

Instead, I suggest we store the Git submodules in a directory that has a name in `asdf/source-registry:*source-registry-exclusions*`.  `_build` seems to be the most appropriate here.

Thoughts?

This supersedes #1211.

Ambrevar:
This is required to avoid conflicting with user / system libraries.
Indeed, `_build` is ignored by ASDF by defaylt.

Supersedes #1211 and closes #1250.


jmercouris:
Thank you!

Ambrevar:
Also note that I've moved quicklisp-client to _build, I thought this
would make you happy!

jmercouris:
I also saw that, yes, thank you very much :-)

PR #1261, opened 2021-03-30
Fix auto mode previous url

### LINKED ISSUE #1095: Auto-mode does not handle restored buffers

Ambrevar:
@aartaka I've replaced the previous-url query with a dedicated slot.
Does that make sense to you?

Something else: can you explain a use case why we need last-active-modes-url, I don't understand the thing with the multiple applications of `auto-mode-handler`.

Should fix #1095 for good.

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -356,7 +356,7 @@ For the storage format see the comment in the head of your `auto-mode-rules-data
 (defmethod serialize-object ((rule auto-mode-rule) stream)
   (flet ((write-if-present (slot &key modes-p)
            (when (funcall slot rule)
-             (format t " :~a ~s"
+             (format t " :~a ~a"
                      slot
                      (let ((value (funcall slot rule)))
                        (if modes-p
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -162,7 +162,7 @@ non-new-page requests, buffer URL is not altered."
       (cond
         ((and (not rule) (new-page-request-p request-data))
          (reapply-last-active-modes auto-mode))
-        ((and rule (not (equalp rule previous-rule)))
+        ((and rule (not (eq rule previous-rule)))
          (enable-matching-modes (url request-data) (buffer request-data))))))
   request-data)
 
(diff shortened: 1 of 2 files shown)


aartaka:
> @aartaka I've replace the previous-url query with a dedicated slot.
> Does that make sense to you?

It does :)

> Something else: can you explain a use case why we need last-active-modes-url, I don't understand the thing with the multiple applications of `auto-mode-handler`.

I can't remember the exact reason, but after some re-reading it looks like a precaution against reloads with different sets of modes that can overwrite `last-active-modes`. Not sure how these reloads can harm us, though. I cannot find the reasoning in the original pull request either.

Ambrevar:
Hmm, I'll think about it.

Anyways, OK to merge this?  Does it work for you with `prompt-on-mode-toggle` set to T?

Ambrevar:
@aartaka Friendly ping in case you missed this :)

aartaka:
I didn't miss it, but am unable to test properly, sorry :sweat_smile:

I'm encountering weird hangs when navigating out of `proxy-mode`-enabling pages to the ones where it's not enabled. I'll open an issue in a day.

All in all, this looks mergeable to me :)

Ambrevar:
> I'm encountering weird hangs when navigating out of `proxy-mode`-enabling pages to the ones where it's not enabled. I'll open an issue in a day.

Is this also an issue on master?

aartaka:
> Is this also an issue on master?

Just checked, it is :(

Ambrevar:
I conclude we can merge this then, right?

Ambrevar:
While on the topic, I've also experienced a similar bug for a while:

- Wake up computer from sleep.
- Resume Nyxt (wasn't shut down).
- Load a URL in an existing buffer.
- *If the Tor node is down*, the load will hang.
  - Then I restart the Tor.

Result: The webkit process hogs the CPU.  The view will never load
anything.
I can open a new buffer and load anything just fine.

aartaka:
> I conclude we can merge this then, right?

Right :)

Ambrevar:
Thanks for your feedback!

PR #1268, opened 2021-03-31
Add plaintext-editor-mode.

aartaka:
This is a simplest possible text editor that we can use by default -- one huge textarea :)

How does it look to you?


jmercouris:
Looks great, just some minor style tweaks, I will handle them though. Thank you Artyom!

PR #1276, opened 2021-04-02
Add :mode property to command-properties.

aartaka:
This adds a "Mode" column to the `execute-command` prompt, as it was before `prompter` merge. It seems this mode column is the most reliable and simple way to understand which mode command belongs to, as [this Reddit post](https://www.reddit.com/r/Nyxt/comments/mhil6l/keybindings_problem_where_to_find_the_module/) says.

Does that look alright?
What was the reason to remove mode names, in the first place?

### DIFF
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -36,7 +36,11 @@
              (first (str:split +newline+ string))))
       (list :name (string-downcase (name command))
             :bindings (format nil "~{~a~^, ~}" bindings)
-            :docstring (first-line (nyxt::docstring command))))))
+            :docstring (first-line (nyxt::docstring command))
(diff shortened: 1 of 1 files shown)


Ambrevar:
Doh!  Good catch, I had accidentally left that out! :/

About the code: you could if-let bind the command package, but that's a nit.

I suggest we seize this opportunity to start using `sera:in` instead of `member :test 'string=`.

Otherwise looks good to me!

aartaka:
I haven't found a way to use `alex:if-let` conveniently, so I resorted to a plain old `let`. `sera:in` is there too :)

Ambrevar:
You are right, `if-let` wouldn't do here.

Ambrevar:
Excellent!

Issue #1219, opened 2021-03-22
Restore file-manager-mode and open-file as an extension (for now)


Ambrevar:
We've removed file-manager-mode when working on the prompt buffer.

aartaka:
affcf5278f0049340ad88e3315f826874145b81b is the commit it was removed in(just to not get lost).

Ambrevar:
With your recent work I don't think there is much left to restore.

What missing, if I'm not mistaken:

- A binding to go back.  I need to work on the prompt-buffer keymaps beforehand.
- An action to open files externally, e.g. with xdg-open.  The download source already does that, but we may want to make this accessible everywhere.

Bonus: Add a source the prompt for a program to open the file with.
Programs can be looked up in PATH.
Maybe add another source for desktop applications, e.g. those with a .desktop file as listed in GNOME, etc.

Issue #1213, opened 2021-03-17
Prompt buffer polishing


Ambrevar:
Follow-up to #1157.

Left to be done:

- [ ] Ensure paging works as expected.  Have C-p recenter on suggestion?

  See if we can keep
  2eeafbac2bea481b04fce0e7e5f8dbfbdcb70acc ("prompt-buffer.lisp: remove
  element-in-view-port code"), or else revert it.
  
- [ ] Remove `make-source`?

  Looks like we are only using `make-instance 'prompter:source`.

- [ ] Remove `constructor` / `destructor` slots, see #1005.
  But do we need instance-specific overrides?

- [ ] Should we have `C-h m` to explain how the prompt buffer works?
  
- [ ] VI bindings

- [ ] Extend support for input history.

  - `set-url` has it, add support for all commands.
  - Make sure duplicates are not inserted.
  
- [ ] Highlight portion of matching text.

- [ ] Enable follow-mode in multi-buffer search.
	  
- [ ] Make meta command which allows for selecting sources, then drop into new
  prompt buffer with given sources.

- [ ] Rename `properties` to `attributes` to avoid confusion with symbol.
      Use strings as keys.

	  Suggestion properties should be an _alist_, so that properties can have
      arbitrary names and casing.
	  I suggest un-dotted alists since they are shorter, more readable, more
      beginner-friendly.

	  So replace `(:name "foo" :desc "bar")` by 
	  `(("Name" "foo") ("My description" "bar"))`.
	  
- [ ] Cosmetic: Option to not display the column names.  For instance in "New
      URL" source.  Simply don't display when there is only one attribute?
	  
- [ ] Add `with-prompt` helper to avoid having to `(let ((result (first (prompt
      ...)))) (when result ...))` for each prompt.  We could even support
      chained prompts, like `sera:and-let*`.
	  
- [x] Echo message when no source has any suggestion (nor do they allow for raw
      user input) instead of showing a useless prompt-buffer.
	  
	  Actually, it may be useful to show an empty prompt buffer to emphasize
      what's happening (echo messages are not always seen), and to display the
      list of sources.
	  
- [ ] Add action to sort by column.
      Would be great if we had mouse support, i.e. click on the column name to
      sort by this column in ascending / descending order.
	  
- [ ] Add action to only show (un)marked suggestions.

- [ ] Add column filters.

  This one is tricky, because we want to make it both flexible and
  explorable.
  We could add actions, but it can be a bit tedious to go through multiple
[...]

PR #1281, opened 2021-04-03
Add file-source.

aartaka:
This adds a universally useful `file-source`. It searches for the file in the current directory, changing the directory properly.

# Things to Discuss
- Maybe add additional actions -- file deletion, touching, copying, opening in editor?
- Maybe add `C-l` for movement up the directories and other keybindings?

# How Has This Been Tested
I've tested it in my config with the function requiring file input. It worked.

How does this look? I'm still at the basic level of `prompter` understanding, but, hopefully, working on this will make me grok it :)

EDIT: It's fuzzy-sortable, yay!

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -397,6 +397,25 @@ call.")))
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
   (:documentation "Prompt for user input words."))
 
+(defun make-file-suggestions (suggestions source input)
+  (declare (ignore suggestions))
+  (let* ((pathname (pathname input))
+         (directory (if (uiop:directory-pathname-p pathname)
+                        pathname
+                        (uiop:pathname-directory-pathname pathname))))
+    (mapcar (lambda (file)
+              (funcall (suggestion-maker source) (namestring file) source))
+            (append (uiop:subdirectories directory) (uiop:directory-files directory)))))
+
+(define-class file-source (source)
+  ((name "Files")
+   (filter nil)
+   (filter-preprocessor 'make-file-suggestions)
+   (multi-selection-p t))
+  (:export-class-name-p t)
+  (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
+  (:documentation "Prompt for file(s) on the disk."))
+
 (defmethod ensure-suggestions-list ((source source) elements
                                     &key input &allow-other-keys)
   (mapcar (lambda (suggestion-value)
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -397,30 +397,6 @@ call.")))
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
   (:documentation "Prompt for user input words."))
 
-(defun make-file-suggestions (suggestions source input)
-  (declare (ignore suggestions))
-  (let* ((pathname (pathname input))
-         (directory (if (uiop:directory-pathname-p pathname)
-                        pathname
-                        (uiop:pathname-directory-pathname pathname))))
-    (mapcar (lambda (file)
-              (make-instance 'suggestion
-                             :value file
-                             :match-data (namestring file)
-                             :properties (object-properties file)
-                             :source source
-                             :input input))
-            (append (uiop:subdirectories directory) (uiop:directory-files directory)))))
-
-;; TODO: `must-match-p'?
-(define-class file-source (source)
-  ((name "Files")
-   (filter-preprocessor 'make-file-suggestions)
-   (multi-selection-p t))
-  (:export-class-name-p t)
-  (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
-  (:documentation "Prompt for file(s) on the disk."))
-
 (defmethod ensure-suggestions-list ((source source) elements
                                     &key input &allow-other-keys)
   (mapcar (lambda (suggestion-value)
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -412,6 +412,7 @@ call.")))
                              :input input))
             (append (uiop:subdirectories directory) (uiop:directory-files directory)))))
 
+;; TODO: `must-match-p'?
 (define-class file-source (source)
   ((name "Files")
    (filter-preprocessor 'make-file-suggestions)
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -404,12 +404,16 @@ call.")))
                         pathname
                         (uiop:pathname-directory-pathname pathname))))
     (mapcar (lambda (file)
-              (funcall (suggestion-maker source) (namestring file) source))
+              (make-instance 'suggestion
+                             :value file
+                             :match-data (namestring file)
+                             :properties (object-properties file)
+                             :source source
+                             :input input))
             (append (uiop:subdirectories directory) (uiop:directory-files directory)))))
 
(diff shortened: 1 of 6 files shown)


aartaka:
Fuzzy-matching works!

aartaka:
- [ ] Another desirable thing: display only the basenames, without obvious clutter of absolute current directory name.

Example:
If the user are in the `/home/user/docs/` directory and are aware of that, they probably want to have a choice between `foo.txt`, `bar.md`, and `quux.org`, instead of `/home/user/docs/foo.txt`, `/home/user/docs/bar.md`, and `/home/user/docs/quux.org`.

libraries/prompter/prompter-source.lisp
```diff
@@ -397,6 +397,30 @@ call.")))
...
+    (mapcar (lambda (file)
+              (make-instance 'suggestion
+                             :value file
+                             :match-data (namestring file)
+                             :properties (object-properties file)
+                             :source source
+                             :input input))
+            (append (uiop:subdirectories directory) (uiop:directory-files directory)))))
+
+;; TODO: `must-match-p'?
```
Ambrevar, line 415: I'm going to remove `must-match-p`, see #1213.

Ambrevar:
This may be a starting point for #1219! ;)

I think we need to look at the old file manager code see what we can save.

Question: Why did you make the source a part of libraries/prompter?  Any specific reason?  True, it's quite general, but I'm concerned that a file manager is a lot of code (see Helm Find Files :p) and having this as part for a library may not ever be good enough.

So I'd suggest first moving it to a new file (e.g. `source/file-manager`), then adding `C-l`, actions, etc.

Ambrevar:
About the basename display: Maybe you can find some inspiration in Helm Find Files. (Do you use Helm?)
So only display the basename and display the current working directory as part of the input.

jmercouris:
This is really cool! Finally a file navigation that works! :-D

With this, it becomes quite feasible to actually use Nyxt as a Notepad replacement :-D

I agree with Pierre, my only comment is just about moving things to a new source file :-P

P.S. sorry about the repetition of my emoticons, I have yet to install the Kaomoji plugin :-O

aartaka:
> About the basename display: Maybe you can find some inspiration in Helm Find Files. (Do you use Helm?)
> So only display the basename and display the current working directory as part of the input.

That's exactly what I want, and Helm is the source of inspiration there :D

aartaka:
> This may be a starting point for #1219! ;)
>
> I think we need to look at the old file manager code see what we can save.

It actually is a starting point :)

> Question: Why did you make the source a part of libraries/prompter? Any specific reason? True, it's quite general, but I'm concerned that a file manager is a lot of code (see Helm Find Files :p) and having this as part for a library may not ever be good enough.

It's a suspicious practice to use a part of a particular mode everywhere. Modes should be as decoupled as possible. That's why putting this source in a mode is not a good idea -- it's already used in three other places/modes with more to come. 

We better ensure mode independence and put `file-source` either in a separate file in the core of Nyxt, or in the `prompter` itself (this I believe to be right, because it's as universally useful as `raw-input`). In any case, it can be extended and inherited afterwards, and `file-manager-mode` will most probably do that, much like `download-mode` does.

Having `open-file` in `download-mode` is confusing, so we better move it somewhere else.

> So I'd suggest first moving it to a new file (e.g. `source/file-manager`), then adding `C-l`, actions, etc.

Creating this file is all fine by me. Small nits, though:
- `file-source` should better be a part of `prompter` (see above).
- The `file-manager-mode` makes more sense as `prompt-buffer` mode, rather than a plain buffer mode. I can change my opinion on that when I get to implementation, though.
- Some actions (like file deletion) can already be part of `file-source`, as they only require UIOP, with no `file-manager-mode`-specific hacks.

Ambrevar:
>> Question: Why did you make the source a part of libraries/prompter? Any specific reason? True, it's quite general, but I'm concerned that a file manager is a lot of code (see Helm Find Files :p) and having this as part for a library may not ever be good enough.
>
> It's a suspicious practice to use a part of a particular mode everywhere. Modes should be as decoupled as possible. That's why putting this source in a mode is not a good idea -- it's already used in three other places/modes with more to come. 

I didn't mean to make a mode, just to move it to `source`.

> Having `open-file` in `download-mode` is confusing, so we better move it somewhere else.

Agreed!

>> So I'd suggest first moving it to a new file (e.g. `source/file-manager`), then adding `C-l`, actions, etc.
>
> Creating this file is all fine by me. Small nits, though:
> - `file-source` should better be a part of `prompter` (see above).

I disagree: I think we should move it to `source` because it's a
non-trivial source that we will expand tremendously in the future.

aartaka:
Sounds fine to me until it's not in a separate mode :)

aartaka:
Okay, now there's `file-manager.lisp` in `source` (is there a better name?)

I have also extracted some helpers to simplify the code -- `directory-elements` and `ensure-directory-pathname`.

aartaka:
Should be mergeable now.

jmercouris:
Thank you Artyom. I too believe it is ready. If you like I can test it and merge.

Is there a better name? I was thinking `file-system.lisp`, though `file-manager.lisp` is also OK!

source/download-mode.lisp
```diff
@@ -147,10 +147,11 @@ download."
   "Download the page or file of the current buffer."
   (download (current-buffer) (url (current-buffer))))
 
-(define-class downloaded-files-source (prompter:source)
-  ((prompter:name "Files")
-   (prompter:must-match-p t)
+(define-class downloaded-files-source (file-source)
+  ((prompter:must-match-p t)
    (prompter:constructor (mapcar #'destination-path (downloads *browser*)))
```
jmercouris, line 152: I've thought that perhaps instead the files-source should have a slot with the path which determines the starting point.

source/file-manager.lisp
```diff
@@ -0,0 +1,33 @@
...
+                             :value file
+                             :match-data (namestring file)
+                             :properties (prompter:object-properties file)
+                             :source source
+                             :input input))
+            (directory-elements (if (uiop:directory-pathname-p pathname)
+                                    pathname
+                                    (uiop:pathname-directory-pathname pathname))))))
+
+(define-class file-source (prompter:source)
```
jmercouris, line 26: here is where the `path` slot would go representing the current path. Not entirely necessary, just a thought :-)

jmercouris:
I have to say, I am REALLY digging the ivy like completion. This interface is actually better than my current emacs config! Very nice Artyom!!! Very nice!

jmercouris:
Ah, one other thing, I would add a "Raw file" source so that the user can open new files. I think it should be enough to add a raw-source to the open-editor-with-new-file prompter instance.

here is what I am saying:
```
(define-command editor-open-file (&key (buffer (current-buffer)))
  "Open a file in the internal editor."
  (let ((file (first (prompt
                      :prompt "Open file"
                      :sources
                      (list (make-instance 'prompter:raw-source)
                            (make-instance 'file-source
                                           :name "Absolute file path"))))))
    (open-file buffer file)
    ;; TODO: Maybe make `editor-mode' and `editor-buffer' pathname-friendly?
    (setf (file buffer) (namestring file))
    (setf (title buffer) (namestring file))
    (setf (url buffer) (quri::make-uri-file :path (namestring file)))))
```

jmercouris:
Please let me know if you are OK with these changes, we can always revert them!

jmercouris:
If you are OK, then let's merge!

aartaka:
Yes, that looks alright to me :)

aartaka:
How about `files.lisp` instead of `file-manager.lisp`?

aartaka:
`files.lisp` is too vague. Let it be `file-manager.lisp` for historical reasons :)

jmercouris:
OK! Thank you once again Artyom!

PR #1288, opened 2021-04-06
Allow for interactive password-interface fixing.

aartaka:
This adds `fill-interface` method used in `with-password` to interactively fix the password interface in case of the wrong password. Currently makes sense for KeePassXC only.

# Things to Discuss
- Can there be a better design? I feel that redefining `fill-interface` in `source/password.lisp` is quite wrong
  - It involves the `password:keepassxc-interface`, while we strive for universality in `source/password.lisp`.
  - It should ideally be in password-interface library.
    - But it cannot be there, as it relies on `prompter`.
- `:invisible-input t` doesn't hide inputed characters at my machine. Is that intended?

# How Has This Been Tested
- `copy-password` with `keepassxc-interface` prompts you for the password file and master password in case it's not set. All as intended.

How does it look?


Ambrevar:
What about defining the generic in `source/` instead?

What happens if you specific a non-existing file or a non-password file?

> :invisible-input t doesn't hide inputed characters at my machine. Is that intended?

Known issue with the new prompter, will work on it as soon as possible! :p

aartaka:
> What about defining the generic in `source/` instead?

Hmmmm, but it seems to belong the the `password-interface`...

> What happens if you specific a non-existing file or a non-password file?

First is not an option, as there's not `raw-source`.
Second is an option, let me fix it.

> > :invisible-input t doesn't hide inputed characters at my machine. Is that intended?
> 
> Known issue with the new prompter, will work on it as soon as possible! :p

Thanks :)

aartaka:
Here you go -- user can only choose files with kdbx extension :)

aartaka:
I believe it's mergeable (unless we move `fill-interface`).

aartaka:
Fixed an empty prompt for password, now it works all fine :)

Ambrevar:
> Here you go -- user can only choose files with kdbx extension :)

What happens if the file has a kdbx extension, but is not actually a
KeepassXC file?

Ambrevar:
(I know, I'm being picky :p)

Ambrevar:
Invisible password input is now fixed with 8576ee9408687387fe2bd6b0c5b19b344d66c9e2.

aartaka:
> Here you go -- user can only choose files with kdbx extension :)
> What happens if the file has a kdbx extension, but is not actually a KeepassXC file?

Well, there is a limit to what we can do. The most solid way to whether the file is right is fetching passwords from it, but that is not possible without master password and vice versa. Because of this interdependence, we need to trust the user at least in one thing. File choice (bounded by extension) is the safest thing to trust to.

Ambrevar:
Maybe what I meant is a bit different: if KeepassXC fails at decrypting
the kbdx file, it may report that the file is not a database, not that
the password is wrong.
In this case, we should not keep prompting for a password but instead
report to the user to select a different file.

aartaka:
> Maybe what I meant is a bit different: if KeepassXC fails at decrypting the kbdx file, it may report that the file is not a database, not that the password is wrong. In this case, we should not keep prompting for a password but instead report to the user to select a different file.

That sounds sane, but that's much more code than this patch was planned to have. Maybe later? I can open an issue about it.

Setting the password file in the initfile is what one will most probably do at some point in time.

aartaka:
Renamed `fill-interface` to `complete-interface` to avoid ambiguities about password field autofills that one can have seeing the previous name.

All the other things look alright to me. Maybe I missed something, can we merge?

source/password.lisp
```diff
@@ -66,9 +66,37 @@ for which the `executable' slot is non-nil."
...
+                      (string-equal "kdbx"
+                                    (pathname-type (pathname (password::password-file
+                                                              password-interface)))))
+          :do (setf (password::password-file password-interface)
+                    (namestring
+                     (first (prompt
+                             :prompt "Password file"
+                             :sources (list (make-instance
+                                             'file-source
+
```
Ambrevar, line 85: Spurious line?  Also you could probably put `'file-source` on the same line as `make-instance`.

source/password.lisp
```diff
@@ -66,9 +66,37 @@ for which the `executable' slot is non-nil."
...
+                                             :filter-preprocessor #'directory-or-kdbx-file))))))
+          :until (password:password-correct-p password-interface)
+          :do (setf (password::master-password password-interface)
+                    (first (prompt
+                            :prompt "Password"
+                            :sources (list (make-instance 'prompter:raw-source))
+                                   :invisible-input-p t))))))
```
Ambrevar, line 92: Bad indentation.

Ambrevar:
Two formatting nits.

One last question: what happens when `with-password` goes through the "else" branch for other password managers that don't implement the method?

source/password.lisp
aartaka: 'Tis a spurious line, right! Moving `'file-source` is not right, though, as the next line will exceed the fill-column after re-indentation.

aartaka:
> One last question: what happens when `with-password` goes through the "else" branch for other password managers that don't implement the method?

It works the same as before this patch (i.e., continues running `body`) -- see default method for `complete-interface`. It just returns the password manager unchanged.

aartaka:
> Two formatting nits.

Fixed!

Ambrevar:
Thanks!

PR #1292, opened 2021-04-08
Prompter attributes

Ambrevar:
Two major changes:

- Rename "properties" to "attributes" in prompter library.  This is to avoid ambiguity with CL symbol properties or plists.

- Turn the attributes into a string-keyed alist.
  This has the benefit that we can now customize how we want to display they attributes keys in the attribute HTML header.  For instance, we now display "URL" instead of "Url".

Please have a a quick test! :)

### DIFF
--- a/build-scripts/guix.scm
+++ b/build-scripts/guix.scm
@@ -151,7 +151,9 @@
     ;; The official Guix package should use `sbcl-*' inputs though.
     (native-inputs
      `(("prove" ,cl-prove)
-       ("sbcl" ,sbcl)))
+       ("sbcl" ,sbcl)
+       ;; Only for development, unneeded for the upstream Guix package:
+       ("cl-trivial-benchmark" ,cl-trivial-benchmark)))
     (inputs
      `(("alexandria" ,cl-alexandria)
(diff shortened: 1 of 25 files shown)


Ambrevar:
What's missing: `describe-os-package` is now even slower than before, this must be fixed.

jmercouris:
Thank you!

jmercouris:
Could the performance penalty be a direct result of using alists instead of plists? I believe alists are significantly slower in many respects.

Ambrevar:
Maybe, but the plist version is also slow so there is definitely room
for improvement.

Should we need plists to restore performance, we could use a plist of
either VALUE or (VALUE KEY-DISPLAY).

aartaka:
Tested is superficially, looks alright!

Why can't we use hash-tables or vectors to speed up the lookup? Search should be faster in both of these.

Ambrevar:
Attributes are rarely numerous, in the ballpark of 5-20 elements.
The cost of initializing the hash-table would be prohibitive I think.

I don't think that searching strings in a vector is faster than in a
list.  Do you know of a fast way with vectors?

Anyways, I'll do some benchmarking and report.

PR #1296, opened 2021-04-14
Open-file

aartaka:
Finally, a `find-file` in Nyxt xD

# What's New
- `download-mode` manages files no more.
- `file-mode` to store all the settings (`find-file-function`, `find-file-in-new-buffer`, `supported-media-types`).
- `find-file` command (and underlying `default-find-file-function`).

# Things to Discuss
- I don't like things being over-entangled with `file-mode`.
  - Consequently, maybe we move mode slots to `*browser*` to make it reliant on global state?
    - Doesn't sound like a solution, though.
- @Ambrevar, how does one get the last `prompt-buffer` in it's own actions? I.e., how do I get the access to the current prompt inside `:actions` of it?
``` lisp
(prompt
     :input (namestring default-directory)
     :prompt "Open file"
     :sources (list (make-instance
                     'file-source
                     :actions (list (make-command open* (files)
                                      (dolist (file files)
                                        (funcall
                                         (find-file-function
                                          (find-submode
                                           ;; This does not work, as the current prompt-buffer is not yet there
                                           ;; Neither it is in active buffers 0_o
                                           (first (old-prompt-buffers *browser*)) 'file-mode))
                                         file))))
                     :filter-preprocessor #'supported-media-or-directory-filter)))
```

DOES NOT YET WORK for the reason above.

How does this look to you?

### DIFF
--- a/source/file-manager.lisp
+++ b/source/file-manager.lisp
@@ -43,3 +43,85 @@ It's suitable for `prompter:filter-preprocessor'."
   (:export-class-name-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
   (:documentation "Prompt source for file(s) on the disk."))
+
+;; TODO: Separate package?
+(define-mode file-mode (nyxt/prompt-buffer-mode:prompt-buffer-mode)
+  "Prompt-buffer mode forthe Nyxt-openable file filtering."
+  ((find-file-in-new-buffer t :documentation "If nil, don't open files and directories in a new buffer.")
+   (supported-media-types '("mp3" "ogg" "mp4" "flv" "wmv" "webm" "mkv")
+                          :type list-of-strings
+                          :documentation "Supported media types.
+Used for file filtering in `find-file'.")
+   (find-file-function #'default-find-file-function)))
+
+(defun supported-media-or-directory (filename)
+  "Return T if this filename's extension is a media that Nyxt can open (or a directory).
+See `supported-media-types' of `file-mode'."
+  (or (and (uiop:directory-pathname-p filename)
+           (uiop:directory-exists-p filename))
+      (sera:and-let* ((extension (pathname-type filename))
+                      (mode (find-submode (current-prompt-buffer) 'file-mode))
+                      (extensions (supported-media-types mode)))
+        (find extension extensions :test #'string-equal))))
+
+#+linux
+(defvar *xdg-open-program* "xdg-open")
+
+(export-always 'default-find-file-function)
+(defun default-find-file-function (filename &key (new-buffer-p
+                                                  (find-file-in-new-buffer
+                                                   (find-submode (current-prompt-buffer)
+                                                                 'file-mode))))
+  "Open FILENAME in Nyxt if supported, or externally otherwise.
+FILENAME is the full path of the file (or directory).
+
+See `supported-media-types' to customize the file types that are opened in
+Nyxt and those that are opened externally.
+
+NEW-BUFFER-P defines whether the file/directory will be open in a new buffer.
+
+Can be used as an `find-file-function'."
+  (handler-case
+      (if (supported-media-or-directory filename)
+          (if new-buffer-p
+              (make-buffer-focus :url (format nil "file://~a" filename))
+              (buffer-load (format nil "file://~a" filename)))
+          (uiop:launch-program
+           #+linux
+           (list *xdg-open-program* (namestring filename))
+           #+darwin
+           (list "open" (namestring filename))))
+    ;; We can probably signal something and display a notification.
+    (error (c) (log:error "Opening ~a: ~a~&" filename c))))
+
+
+(define-command find-file (&key (default-directory (user-homedir-pathname)))
+  "Open a file from the filesystem.
+
+The user is prompted with the prompt-buffer, files are browsable with
+fuzzy suggestion.
+
+DEFAULT-DIRECTORY specifies which directory to start from. Defaults to user home
+directory.
+
+By default, it uses the `xdg-open' command. The user can override the
+`find-file-function' of `file-mode' which takes the filename (or
+directory name) as parameter."
+  ;; TODO: How do we set the current directory permanently?
+  (flet ((supported-media-or-directory-filter (suggestions source input)
+           (remove-if-not #'supported-media-or-directory
+                          (make-file-suggestions suggestions source input)
+                          :key #'prompter:value)))
+    (prompt
+     :input (namestring default-directory)
+     :prompt "Open file"
+     :sources (list (make-instance
+                     'file-source
+                     :actions (list (make-command open* (files)
+                                      (dolist (file files)
+                                        (funcall
+                                         (find-file-function
+                                          (find-submode
+                                           (first (old-prompt-buffers *browser*)) 'file-mode))
+                                         file))))
+                     :filter-preprocessor #'supported-media-or-directory-filter)))))
--- a/source/file-manager.lisp
+++ b/source/file-manager.lisp
@@ -40,12 +40,12 @@ It's suitable for `prompter:filter-preprocessor'."
   ((prompter:name "Files")
    (prompter:filter-preprocessor (make-file-source-preprocessor))
    (prompter:multi-selection-p t)
-   (find-file-in-new-buffer-p t :documentation "If nil, don't open files and directories in a new buffer.")
+   (open-file-in-new-buffer-p t :documentation "If nil, don't open files and directories in a new buffer.")
    (supported-media-types '("mp3" "ogg" "mp4" "flv" "wmv" "webm" "mkv")
                           :type list-of-strings
                           :documentation "Media types that Nyxt can open.
 Others are opened with OS-specific mechanisms.")
-   (find-file-function #'default-find-file-function))
+   (open-file-function #'default-open-file-function))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
@@ -68,22 +68,22 @@ See `supported-media-types' of `file-mode'."
 (defmethod initialize-instance :after ((source file-source) &key)
   (setf (slot-value source 'prompter:actions)
         (list (make-command open-file* (files)
-                (let* ((new-buffer-p (find-file-in-new-buffer-p source)))
-                  ;; Open first file according to `find-file-in-new-buffer-p'
-                  (funcall (find-file-function source) (first files)
+                (let* ((new-buffer-p (open-file-in-new-buffer-p source)))
+                  ;; Open first file according to `open-file-in-new-buffer-p'
+                  (funcall (open-file-function source) (first files)
                            :new-buffer-p new-buffer-p
                            :supported-p (supported-media-or-directory (first files) source))
                   ;; Open the rest of the files in new buffers unconditionally.
                   (dolist (file (rest files))
-                    (funcall (find-file-function source) file
+                    (funcall (open-file-function source) file
                              :new-buffer-p t
                              :supported-p (supported-media-or-directory file source))))))))
 
 #+linux
 (defvar *xdg-open-program* "xdg-open")
 
-(export-always 'default-find-file-function)
-(defun default-find-file-function (filename &key supported-p new-buffer-p)
+(export-always 'default-open-file-function)
+(defun default-open-file-function (filename &key supported-p new-buffer-p)
   "Open FILENAME in Nyxt if supported, or externally otherwise.
 FILENAME is the full path of the file (or directory).
 
@@ -93,7 +93,7 @@ Nyxt and those that are opened externally.
 NEW-BUFFER-P defines whether the file/directory is opened in a new buffer.
 SUPPORTED-P says whether the file can be opened by Nyxt.
 
-Can be used as a `find-file-function'."
+Can be used as a `open-file-function'."
   (handler-case
       (if supported-p
           (if new-buffer-p
@@ -107,7 +107,7 @@ Can be used as a `find-file-function'."
     ;; We can probably signal something and display a notification.
     (error (c) (log:error "Opening ~a: ~a~&" filename c))))
 
-(define-command find-file (&key (default-directory (user-homedir-pathname)))
+(define-command open-file (&key (default-directory (user-homedir-pathname)))
   "Open a file from the filesystem.
 
 The user is prompted with the prompt-buffer, files are browsable with
@@ -117,7 +117,7 @@ DEFAULT-DIRECTORY specifies which directory to start from. Defaults to user home
 directory.
 
 By default, it uses the `xdg-open' command. The user can override the
-`find-file-function' of `file-mode' which takes the filename (or
+`open-file-function' of `file-mode' which takes the filename (or
 directory name) as parameter."
   (prompt
    :input (namestring default-directory)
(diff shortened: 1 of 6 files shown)


aartaka:
Another thing: It's called `find-file` for more than Emacs similarity -- `open-file` already exists in `nyxt/editor-mode`. I used `find-file` to avoid collisions. Ideally it should be `open-file` for this patch command, and something like `edit-file` for the editor one.

jmercouris:
Not sure yet, I will have to think about it a little bit. Sounds OK, though.

source/file-manager.lisp
```diff
@@ -31,3 +31,85 @@
   (:export-class-name-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
   (:documentation "Prompt source for file(s) on the disk."))
+
+;; TODO: Separate package?
+(define-mode file-mode (nyxt/prompt-buffer-mode:prompt-buffer-mode)
+  "Prompt-buffer mode forthe Nyxt-openable file filtering."
```
jmercouris, line 37: small little typo :-)

source/file-manager.lisp
```diff
@@ -31,3 +31,85 @@
...
+See `supported-media-types' of `file-mode'."
+  (or (and (uiop:directory-pathname-p filename)
+           (uiop:directory-exists-p filename))
+      (sera:and-let* ((extension (pathname-type filename))
+                      (mode (find-submode (current-prompt-buffer) 'file-mode))
+                      (extensions (supported-media-types mode)))
+        (find extension extensions :test #'string-equal))))
+
+#+linux
+(defvar *xdg-open-program* "xdg-open")
```
jmercouris, line 56: Could this be a slot of the file-mode instead?

Ambrevar: It could _also_ be a slot, but all references to external programs should be defvars, so that packagers can easily find and patch them.

source/file-manager.lisp
```diff
@@ -31,3 +31,85 @@
...
+           #+linux
+           (list *xdg-open-program* (namestring filename))
+           #+darwin
+           (list "open" (namestring filename))))
+    ;; We can probably signal something and display a notification.
+    (error (c) (log:error "Opening ~a: ~a~&" filename c))))
+
+
+(define-command find-file (&key (default-directory (user-homedir-pathname)))
+  "Open a file from the filesystem.
```
Ambrevar, line 87: "file system"

source/file-manager.lisp
```diff
@@ -31,3 +31,85 @@
...
+                                                  (find-file-in-new-buffer
+                                                   (find-submode (current-prompt-buffer)
+                                                                 'file-mode))))
+  "Open FILENAME in Nyxt if supported, or externally otherwise.
+FILENAME is the full path of the file (or directory).
+
+See `supported-media-types' to customize the file types that are opened in
+Nyxt and those that are opened externally.
+
+NEW-BUFFER-P defines whether the file/directory will be open in a new buffer.
```
Ambrevar, line 69: "is opened"

source/file-manager.lisp
jmercouris: Then it should be in global.lisp ideally!

source/file-manager.lisp
```diff
@@ -31,3 +31,85 @@
...
+          (uiop:launch-program
+           #+linux
+           (list *xdg-open-program* (namestring filename))
+           #+darwin
+           (list "open" (namestring filename))))
+    ;; We can probably signal something and display a notification.
+    (error (c) (log:error "Opening ~a: ~a~&" filename c))))
+
+
+(define-command find-file (&key (default-directory (user-homedir-pathname)))
```
Ambrevar, line 86: Shouldn't `default-directory` be `*default-pathname-defaults*` instead?  This way, we can "cd" into other places and later `find-file` will start from there.

Ambrevar:
About referring to the calling prompt: see `toggle-properties-display`.

Ambrevar:
Can't we conflate `find-file` and `edit-file` then?  Maybe the way a file is opened could depend on the current mode?
Also we could add an action to edit the file instead of using xdg-open on it.

Ambrevar:
@aartaka Note that I'm currently working on an overhaul of `file-source`: the definition is not right because you should not have to rely on `make-file-suggestions` and instead of `object-properties`, but I see now that there are API deficiencies in the prompter, so I need to work on it.

source/file-manager.lisp
jmercouris: Artyom is correct: https://en.wikipedia.org/wiki/File_system

aartaka:
> @aartaka Note that I'm currently working on an overhaul of `file-source`: the definition is not right because you should not have to rely on `make-file-suggestions` and instead of `object-properties`, but I see now that there are API deficiencies in the prompter, so I need to work on it.

Sure thing, `find-file` pathc can wait :) Looking forward for a cool API change!

aartaka:
Finally, we can open files with `find-file` in Nyxt! Is that okay to modify the keymap of `file-mode` in `initialize-instance :after`? I want to inherit all the keybindings (except RET) from `prompt-buffer-mode`, so that seems the way to go.

What's left to do:
- Integrate with `editor-mode` (`supported-text-types` slot?).
  - Rename to `open-file` then.
- `find-file` keybindings (`C-x C-f` is obvious to me, but what are CUA (`C-o`?) and VI ones?).
- Fix CI :D

Ambrevar:
> Finally, we can open files with `find-file` in Nyxt! Is that okay to modify the keymap of `file-mode` in `initialize-instance :after`? I want to inherit all the keybindings (except RET) from `prompt-buffer-mode`, so that seems the way to go.

There should be no need to do anything, prompt-buffer-mode bindings are
looked up after those of find-mode.

> - Integrate with `editor-mode` (`supported-text-types` slot?).

Why not.  We can also leverage MIME types.

>   - Rename to `open-file` then.
> - `find-file` keybindings (`C-x C-f` is obvious to me, but what are CUA (`C-o`?) and VI ones?).

VI would be `o` I suppose.

aartaka:
> Finally, we can open files with `find-file` in Nyxt! Is that okay to modify the keymap of `file-mode` in `initialize-instance :after`? I want to inherit all the keybindings (except RET) from `prompt-buffer-mode`, so that seems the way to go.
> There should be no need to do anything, prompt-buffer-mode bindings are looked up after those of find-mode.

This wasn't the case somewhy, so I decided to set it explicitly just in case. Won't hurt, I guess :)

> - Integrate with `editor-mode` (`supported-text-types` slot?).
> Why not. We can also leverage MIME types.
> - Rename to `open-file` then. - `find-file` keybindings (`C-x C-f` is obvious to me, but what are CUA (`C-o`?) and VI ones?).
> VI would be `o` I suppose.

Got it! Wait a push in a minute :)

aartaka:
All resolved, except text files. Should we hard-code the list of text file formats, and, if so, where do we get one? There are lots of test file formats...

aartaka:
> - Integrate with `editor-mode` (`supported-text-types` slot?).
> Why not. We can also leverage MIME types.

Can you expand on how we can use MIME types there, Pierre? Maybe that's the solution to my problem of text types' list :)

aartaka:
Oh, and `o` is already taken in VI keymap, so I went with `g o`. Feel free to alter, 'cause I'm terrible at understanding VI keybindings!

Ambrevar:
> Oh, and `o` is already taken in VI keymap, so I went with `g o`. Feel free to alter, 'cause I'm terrible at understanding VI keybindings!

Shouldn't be.  Taken by what?

Ambrevar:
WebKitGTK has a MIME function I think.

Otherwise, we can use https://shinmera.github.io/trivial-mimes/.
There is also https://common-lisp.net/project/magicffi/.

Ambrevar:
Short review:

- https://shinmera.github.io/trivial-mimes/:

  No CFFI, fastest, no MIME description (only the type), hard-coded, external
  database (could be a problem when producing the binary).
  
- https://common-lisp.net/project/magicffi/:
  CFFI, a bit slower, MIME type + desc + encoding, depends on libmagic
  (from the `file` package).

aartaka:
> Oh, and `o` is already taken in VI keymap, so I went with `g o`. Feel free to alter, 'cause I'm terrible at understanding VI keybindings!
> Shouldn't be. Taken by what?

It's taken by `set-url` in `base-mode`. Or it was taken by it the last time I checked it :)

aartaka:
> WebKitGTK has a MIME function I think. 

WebKitGTK specifically doesn't, but GTK (GIO) does have these:
- `g_file_info_get_content_type()`
- `g_file_info_get_mime_type()`

> Otherwise, we can use https://shinmera.github.io/trivial-mimes/. There is also https://common-lisp.net/project/magicffi/.

Using `trivial-mimes` with [fresh IANA MIME collection](https://www.iana.org/assignments/media-types/media-types.xhtml) sounds like a good way in case GTK methods don't work/aren't bound/are too slow.

`libmagic` FFI is good too, except that it's going to be a one-shot dependency :)

Ambrevar:
> It's taken by `set-url` in `base-mode`. Or it was taken by it the last time I checked it :)

But there is no base mode in the prompt-buffer.

aartaka:
Oh, sorry for being confusing :) I meant the keybinding for `find-file` (the command that opens the file that user picks via prompt buffer) while not in the prompt-buffer.

And that's where the problem appears: `o` is already bound (to `set-url`) in `base-mode`, and `base-mode` is enabled in all the usual buffers (non-prompting ones, I mean). Your suggestion to use `o` for `find-file` is colliding with the `set-url` keybinding. That is why I used `g o` to call the `find-file`. Is that a sane keybinding?

See eeee549 for the keybinding in question.

Ambrevar:
Thanks for the clarification, indeed I had misunderstood.
`g o` is fine by me!

aartaka:
Note: I'll rewrite this yet another time once #1340 is merged. I believe all the `file-mode` tinkering can be elegantly moved to `file-source`, that's why the rewrite.

Ambrevar:
#1340 is merged!

aartaka:
Merged `file-mode` into `file-source` and renamed `find-file` to `open-file`. Maybe we should call the file file-source.lisp instead of vague file-manager.lisp? Seems like `file-source` can fit anything inside itself :)

Are sources `define-configuration`-friendly? When I `define-user-class` on file-source and try to configure it in init.lisp, it's only configured when I use `nyxt::` prefix everywhere, even though the class name and accessor names are exported. Is that an intended behavior?

aartaka:
It's merge-ready otherwise :)

Ambrevar:

> Are sources `define-configuration`-friendly? When I `define-user-class` on file-source and try to configure it in init.lisp, it's only configured when I use `nyxt::` prefix everywhere, even though the class name and accessor names are exported. Is that an intended behavior?

Nope!  Can you give a complete example?

aartaka:
> Are sources `define-configuration`-friendly? When I `define-user-class` on file-source and try to configure it in init.lisp, it's only configured when I use `nyxt::` prefix everywhere, even though the class name and accessor names are exported. Is that an intended behavior?
> Nope! Can you give a complete example?

Sure! This works:
``` lisp
(define-configuration nyxt::file-source
  ((nyxt::supported-media-types `("jpeg" "png" ,@%slot-default%))))
```
While this doesn't:
``` lisp
(define-configuration file-source
  ((supported-media-types `("jpeg" "png" ,@%slot-default%))))
```

Even though `file-source` should be exported to `nyxt-user`.

aartaka:
By "works" I mean that files with .png and .jpeg extensions become viewable in Nyxt. It doesn't happen with the second config snippet.

Ambrevar:
Maybe because the source is not in the `nyxt` package?

aartaka:
> Maybe because the source is not in the `nyxt` package?

0_o

It is. There is `(in-package :nyxt)` in file-manager.lisp.

aartaka:
Maybe merge this as part of 2-pre-release-7? It's pretty small, polished, and relatively backwards-compatible (we used to have file-opening until recently) :)

aartaka:
Once we understand the prefixing issue, of course :upside_down_face:

Ambrevar:
I'd like to test this a bit more.  I hope to find the time :)

jmercouris:
Artyom, I'm dissapointed, what is this regular emoji?

jmercouris:
In case it isn't clear from above, I am joking :-)

Ambrevar:
:)

aartaka:
xD

source/file-manager.lisp
```diff
@@ -39,7 +39,87 @@ It's suitable for `prompter:filter-preprocessor'."
...
+      (if supported-p
+          (if new-buffer-p
+              (make-buffer-focus :url (quri::make-uri-file :path filename))
+              (buffer-load (quri::make-uri-file :path filename)))
+          (uiop:launch-program
+           #+linux
+           (list *xdg-open-program* (namestring filename))
+           #+darwin
+           (list "open" (namestring filename))))
+    ;; We can probably signal something and display a notification.
```
Ambrevar, line 107: See notification.lisp.

source/file-manager.lisp
```diff
@@ -39,7 +39,87 @@ It's suitable for `prompter:filter-preprocessor'."
...
+                                     &optional (file-source (make-instance 'user-file-source)))
+  "Return T if this filename's extension is a media that Nyxt can open (or a directory).
+See `supported-media-types' of `file-mode'."
+  (or (and (uiop:directory-pathname-p filename)
+           (uiop:directory-exists-p filename))
+      (sera:and-let* ((extension (pathname-type filename))
+                      (extensions (supported-media-types file-source)))
+        (find extension extensions :test #'string-equal))))
+
+(defmethod initialize-instance :after ((source file-source) &key)
```
Ambrevar, line 68: Smart, I hadn't though of this.  Maybe I can leverage this technique to unexport suggestion-maker.

source/file-manager.lisp
```diff
@@ -39,7 +39,87 @@ It's suitable for `prompter:filter-preprocessor'."
...
   ((prompter:name "Files")
    (prompter:filter-preprocessor (make-file-source-preprocessor))
-   (prompter:multi-selection-p t))
+   (prompter:multi-selection-p t)
+   (open-file-in-new-buffer-p t :documentation "If nil, don't open files and directories in a new buffer.")
+   (supported-media-types '("mp3" "ogg" "mp4" "flv" "wmv" "webm" "mkv")
+                          :type list-of-strings
+                          :documentation "Media types that Nyxt can open.
+Others are opened with OS-specific mechanisms.")
+   (open-file-function #'default-open-file-function))
```
Ambrevar, line 48: This function expects a specific lambda list, i.e. the `supported-p` and `new-buffer-p` arguments.
This should be documented.

Ambrevar:
Weirdly enough, opening a video works from the binary, but from the REPL it triggers the download manager instead.  Any idea?

Ambrevar:
I've fixed the define-class export issue on master:
17bff7d867222ea211ecf859efed0655688042e2

Ambrevar:
Beside the REPL issue, this looks good to merge for me.

@jmercouris Let's try to get it in for 2-pre-release-7?
User feedback on this could be useful.

jmercouris:
I think we'll get user feedback anyways, many people are running master...!

jmercouris:
If you are OK to merge, I am OK to merge.

Ambrevar:
I'm undecisive... :p @aartaka, any last minute opinion on this?

jmercouris:
I'm not, done!

Ambrevar:
:D

aartaka:
Things I'll do now, as a follow-up to merge:
- Use `*default-pathname-defaults*` as a default directory for `open-file`.
- Extend documentation of `open-file-function` with mentions of it's keyword arguments.

aartaka:
Done with 3b8fa1592426465093dd2ebbd7d7b60708ed5f9d and c65cfa48b5460befa2de46725766bf5c981761eb!

PR #1297, opened 2021-04-14
Element frame & Expedition Mode

jmercouris:
Select a set of elements with a box. For post 2.0.


jmercouris:
The future is now: https://www.youtube.com/watch?v=lML2UH3kbuY

Ambrevar:
To clarify, the difference with the previous video is that you are now
running actions on the results, right?

In all cases: super cool!  Kudos!

jmercouris:
Well, turning it into a source was rather complex, but I've managed to do so. I am now running the actions against a source.

jmercouris:
Thank you!

Ambrevar:
Fantastic, this is a good test for the prompt API! :)

jmercouris:
@aartaka , @Ambrevar 

auto mode interferes with expedition mode, any way to ensure that it is not enabled in a newly created buffer? I can explicitly set the modes in the new buffer I create, but that does not seem to be the correct solution as it ignores the users default modes.

jmercouris:
In the interim I've removed auto-mode from my default modes :'(

Ambrevar:
How does it interfere?

What you could do is instantiate the buffer on a blank URL, remove
auto-mode from there, then load the URL.

jmercouris:
It interferes by disabling expedition mode every time the url is changed

jmercouris:
Go ahead and try this branch, it's a lot of fun, but remove auto mode from your default modes first

jmercouris:
No need to remove auto mode now to try this branch, I have fixed it!

source/expedition-mode.lisp
```diff
@@ -44,20 +44,21 @@
...
+                                               :channel (make-instance 'calispel:channel)))
+                 :after-destructor
+                 (lambda ()
+                   (with-current-buffer buffer
+                     (nyxt/web-mode::frame-element-clear))))))
+         (buffer (make-buffer :title "" :url (first urls))))
+    (expedition-mode :urls urls :buffer buffer)
+    ;; Auto mode interferes with expedition mode, automatically disabling it
+    ;; when viewing the next URL in an expedition
+    (disable-modes 'auto-mode buffer)
```
aartaka, line 63: This does not look right.

How about adding `nyxt/expedition-mode:expedition-mode` to the `non-rememberable/modes` of `auto-mode` so that auto-mode does not see expedition-mode?

Hopefully, I'll make apull request allowing to set rememeberability as a mode slot, in a couple of days.

jmercouris: In time, yes. After that is merged, after 2.0, I will then make the change as you've suggested!

aartaka: Here you go: #1304! Setting `rememberable-p` of `expedition-mode` to `nil` will do the job.

PR #1298, opened 2021-04-14
Update bindings

jmercouris:
I've made a branch, please feel free to push changes to the bindings you think are necessary.

### DIFF
--- a/source/base-mode.lisp
+++ b/source/base-mode.lisp
@@ -13,11 +13,8 @@ This mode is a good candidate to be passed to `make-buffer'."
                      "C-q" 'quit
                      "C-[" 'switch-buffer-previous
                      "C-]" 'switch-buffer-next
-                     "C-x b" 'switch-buffer
-                     "C-x C-b" 'list-buffers
-                     "C-x k" 'delete-buffer
+                     "M-down" 'switch-buffer
                      "C-w" 'delete-current-buffer
-                     "C-x C-k" 'delete-current-buffer
                      "C-shift-tab" 'switch-buffer-previous
                      "C-tab" 'switch-buffer-next
                      "C-pageup" 'switch-buffer-previous
@@ -60,13 +57,14 @@ This mode is a good candidate to be passed to `make-buffer'."
                      "C-/" 'reopen-buffer
                      "C-shift-t" 'reopen-buffer
                      "C-T" 'reopen-buffer
-                     "C-p" 'print-buffer
-                     "M-i" 'focus-first-input-field)
+                     "M-i" 'focus-first-input-field
+                     "C-p" 'print-buffer)
 
                     scheme:emacs
                     (list
                      "C-x C-c" 'quit
-                     "C-x k" 'delete-buffer ; Emacs' default behaviour is to query.
+                     "C-x k" 'delete-buffer
+                     "C-x C-k" 'delete-current-buffer
                      "C-x left" 'switch-buffer-previous
                      "C-x right" 'switch-buffer-next
                      "C-x b" 'switch-buffer
@@ -84,6 +82,7 @@ This mode is a good candidate to be passed to `make-buffer'."
                      "C-h s" 'describe-slot
                      "C-h k" 'describe-key
                      "C-h b" 'describe-bindings
+                     "C-d" 'list-downloads
                      "M-x" 'execute-command
                      "C-x r j" 'set-url-from-bookmark
                      "C-x r J" 'set-url-from-bookmark-new-buffer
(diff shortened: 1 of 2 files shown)


jmercouris:
At the end we can rebase everything into a single commit...!

jmercouris:
I can't think of much else to change, I've been through most of the keybindings and I have no ideas. I set them to what I thought were sensible defaults at the time of creation.

source/base-mode.lisp
```diff
@@ -66,11 +64,13 @@ This mode is a good candidate to be passed to `make-buffer'."
...
                     (list
                      "C-x C-c" 'quit
-                     "C-x k" 'delete-buffer ; Emacs' default behaviour is to query.
+                     "C-x k" 'delete-buffer
+                     "C-x C-k" 'delete-current-buffer
                      "C-x left" 'switch-buffer-previous
                      "C-x right" 'switch-buffer-next
                      "C-x b" 'switch-buffer
                      "C-x C-b" 'list-buffers
+                     "C-x C-b" 'list-buffers
```
Ambrevar, line 73: Dupplicate?

source/base-mode.lisp
```diff
@@ -13,11 +13,9 @@ This mode is a good candidate to be passed to `make-buffer'."
                      "C-q" 'quit
                      "C-[" 'switch-buffer-previous
                      "C-]" 'switch-buffer-next
-                     "C-x b" 'switch-buffer
-                     "C-x C-b" 'list-buffers
-                     "C-x k" 'delete-buffer
+                     "M-[" 'switch-buffer
+                     "M-]" 'switch-buffer
```
Ambrevar, line 17: Not sure about this one, it does not seem intuitive.  What about `C-\`?  It's often next to `[` or `]`.

jmercouris: I don't see how C-\ is any more intuitive :-D

source/base-mode.lisp
jmercouris: Duplicate!

source/base-mode.lisp
jmercouris: This is the one binding I really struggled with, I don't see a parallel in any other CUA programs to draw inspiration. Maybe "C-." would be a good one? or "C-/" or "C-?"?

jmercouris: We could take a hint from macOS expose binding which would be "C-down"

Ambrevar: OK for `C-down`.
My thinking behind `C-\` is that it's next to `C-[` and `C-]`.
What I don't like about `M-[` and `M-]` is that the brackets imply a symmetry which isn't there.  `\` does not suffer from this (unless you consider the symmetry of `\` to be `/`).

jmercouris: I had to switch to "M-down" to avoid collision with scroll to bottom of page.

jmercouris:
@aartaka any other suggestions? @aadcg ? @Ambrevar ?

aadcg:
Nothing worth sharing. All good for me.

aartaka:
No suggestions from me — it seems all good :)

source/base-mode.lisp
```diff
@@ -60,13 +57,14 @@ This mode is a good candidate to be passed to `make-buffer'."
                      "C-/" 'reopen-buffer
                      "C-shift-t" 'reopen-buffer
                      "C-T" 'reopen-buffer
-                     "C-p" 'print-buffer
-                     "M-i" 'focus-first-input-field)
```
Ambrevar, line 64: I think this M-i binding should stay, no?

source/base-mode.lisp
```diff
@@ -60,13 +57,14 @@ This mode is a good candidate to be passed to `make-buffer'."
...
+                     "C-p" 'print-buffer)
 
                     scheme:emacs
                     (list
+                     "M-i" 'focus-first-input-field
```
Ambrevar, line 64: And I'd remove it from here.

source/base-mode.lisp
Ambrevar: OK.

source/base-mode.lisp
jmercouris: I moved it to the emacs section, because it seems an emacs like binding a CUA user would not use

jmercouris: I tried M-i in Firefox and it did nothing

Ambrevar:
True, it's not a CUA binding, but it's not an Emacs binding either.

I suggest the following approach: set all Nyxt specific bindings in the
CUA scheme since the other schemes inherit from CUA.

jmercouris:
OK, I will change that and then merge.

source/base-mode.lisp
aadcg: > I tried M-i in Firefox and it did nothing

But that wouldn't be a good analogy anyway. Inspiration would have to come from browser extensions like tridactyl (that I have never uses and know nothing about).

I seem to agree with @Ambrevar.

PR #1301, opened 2021-04-15
Review docstrings

aadcg:
This an attempt to standardize docstrings, fix some issues/typos, and add some key but nonexistent docstrings (such as important classes, modes, etc.)


jmercouris:
Thank you Andre, I will take a look!

source/browser.lisp
```diff
@@ -104,21 +104,22 @@ identifiers in `get-unique-buffer-identifier'.  We can't rely on the windows
 count since deleting windows may result in duplicate identifiers.")
    (startup-function (make-startup-function)
                      :type (or function null)
-                     :documentation "The function run on startup.  It takes a
+                     :documentation "The function runs on startup.  It takes a
 list of URLs (strings) as optional argument (the command line positional
 arguments).  It is run after the renderer has been initialized, after the
 `*after-init-hook*' has run and after a first window has been spawned.")
+   ;; This is a hook to run arbitrary code, is it needed?  If needed, isn't
```
jmercouris, line 111: The startup function slot will be removed. Pierre and I have already conclued it must be removed.

aadcg: I will drop this commit then.

jmercouris: Please let me know when that is done and I will merge, I have seen everything else :-)

aadcg: So you agree with all of my the other changes? I'm surprised. Do you find the docstrings on buffer, browser and window good enough?

jmercouris: I find them good enough, yes. I don't think they need to be perfect, in my opinion it is an interative process.

jmercouris: If you must know all specifics, I was debating whether I should mention the usage of commas in some of your changes :-)

jmercouris: The startup function slot has now been removed.

source/window.lisp
```diff
@@ -139,7 +141,7 @@ The handlers take the window as argument."))
    :sources (make-instance 'window-source)))
 
 (define-command delete-current-window (&optional (window (current-window)))
-  "Delete WINDOW, or the currently active window if unspecified."
+  "Delete WINDOW or the current window, when omitted."
```
jmercouris, line 144: This is the comma I was referring to. i would stil write:

"Delete WINDOW, or the current window, when omitted."

source/window.lisp
```diff
@@ -154,7 +156,7 @@ The handlers take the window as argument."))
     (values window buffer)))
 
 (define-command toggle-fullscreen (&optional (window (current-window)))
-  "Fullscreen WINDOW, or the currently active window if unspecified."
+  "Fullscreen WINDOW or the current window, when omitted."
```
jmercouris, line 159: This is the second comma I was referring to. I would still write:

"Fullscreen WINDOW, or the current window, when omitted."

aadcg:
I removed the commit regarding the obsolete slot, and changed the commas (I agree with you). 

You can merge.

aadcg:
Ready to merge @jmercouris.

aadcg:
I noticed that predicates are sometimes documented as "If non-nil..." and other times "Whether". Actually, I did this myself. Maybe we could standardize this?

jmercouris:
I don't think it will determine the success of Nyxt whether we use wether or "if non-nil" :-).

Ambrevar:
Just had a look: awesome work, @aadcg!  Thank you so much!

aadcg:
Thanks @Ambrevar. There's more coming, I'm filling up my git stash as I work on other things.

PR #1304, opened 2021-04-17
Make modes rememberable.

aartaka:
This adds `rememberable-p` slot to `root-mode` and (all) the inheriting
modes. It's set to nil for those modes that we don't want to be
remembered by `auto-mode`, like `internal-buffer`-styling modes or
one-shot modes, like `visual-mode` or `expedition-mode`.

# Things To Discuss
This, however, shifts some of the responsibilities of `auto-mode` back
to modes, instead of universal `mode-invocation`s. Should it be that
way? I'm not sure.

# How Has This Been Tested
- Nyxt compiles and runs with this. Haven't tested all the non-rememberable modes, though.

How does that look?

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -51,16 +51,14 @@ Package prefix is optional.")
 If the mode specifier is not known, it's omitted from the results."
   (delete nil (mapcar #'mode-invocation mode-list)))
 
-(declaim (ftype (function (list (or auto-mode null))
+(declaim (ftype (function ((or (cons mode-invocation *) null))
                           (or (cons mode-invocation *) null))
                 rememberable-of))
-(defun rememberable-of (modes auto-mode)
-  "Filter MODES based on rememberability by AUTO-MODE."
-  (if auto-mode
-      (set-difference (mode-invocations modes)
-                      (non-rememberable-modes auto-mode)
-                      :test #'equals)
-      modes))
+(defun rememberable-of (mode-invocations)
+  "Filter MODES based on `rememberable-p'."
+  (remove-if (complement #'rememberable-p) mode-invocations
+             :key #'(lambda (mode-invocation)
+                      (make-instance (name mode-invocation)))))
 
 (define-class auto-mode-rule ()
   ((test (error "Slot `test' should be set.")
@@ -96,18 +94,17 @@ If the mode specifier is not known, it's omitted from the results."
 
 (declaim (ftype (function (quri:uri buffer)) enable-matching-modes))
 (defun enable-matching-modes (url buffer)
-  (let ((rule (matching-auto-mode-rule url buffer))
-        (auto-mode (find-mode buffer 'auto-mode)))
+  (let ((rule (matching-auto-mode-rule url buffer)))
     (dolist (mode-invocation (set-difference
                                (included rule)
-                               (rememberable-of (modes buffer) auto-mode)
+                               (rememberable-of (mode-invocations (modes buffer)))
                                :test #'equals))
       (check-type mode-invocation mode-invocation)
       (enable-modes (list (name mode-invocation)) buffer (arguments mode-invocation)))
     (disable-modes (mapcar #'name
                            (if (exact-p rule)
                                (set-difference
-                                (rememberable-of (modes buffer) auto-mode)
+                                (rememberable-of (mode-invocations (modes buffer)))
                                 (included rule) :test #'equals)
                                (excluded rule)))
                    buffer)))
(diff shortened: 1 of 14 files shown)


source/auto-mode.lisp
```diff
@@ -51,16 +51,12 @@ Package prefix is optional.")
...
-(defun rememberable-of (modes auto-mode)
-  "Filter MODES based on rememberability by AUTO-MODE."
-  (if auto-mode
-      (set-difference (mode-invocations modes)
-                      (non-rememberable-modes auto-mode)
-                      :test #'equals)
-      modes))
+(defun rememberable-of (modes)
+  "Filter MODES based on `rememberable-p', turn them into `mode-invocation's."
+  (mode-invocations (remove-if-not #'rememberable-p modes)))
```
Ambrevar, line 59: To be picky, `remove-if-not` is deprecated.
You can use `remove-if (complement ...)` or `sera:filter` instead.

source/auto-mode.lisp
```diff
@@ -310,11 +295,11 @@ For the storage format see the comment in the head of your `auto-mode-rules-data
       (setf url (url url)))
     (add-modes-to-auto-mode-rules
      (url-infer-match url)
-     :include (set-difference (mode-invocations (modes (current-buffer)))
+     :include (set-difference (rememberable-of (modes (current-buffer)))
```
Ambrevar, line 298: I don't underestand why this `mode-invocations` is replaced by `rememberable-of`.
Same below.
Can you elaborate?

Ambrevar:
Looks good to me!
I think it's actually a very welcome change.

About moving the feature from auto-mode to mode, I think it's actually better this way, because "rememberable-p" is not a name that necessarily refers to auto-mode and can potentially be used by other modes in the future.  In other words, it's a slot that composes well.
Should we ever use rememberable-p for something other than auto-mode, all we would have to do is updat the documentation ;)

source/auto-mode.lisp
aartaka: Sure I can. 
1. We cannot check rememberability of the modes in `add-modes-to-auto-mode-rules`, as we cannot call `rememberable-p` on `mode-instance`s passed to `add-modes-to-auto-mode-rules` (actually, we can, but that requires calling/instantiating the mode, and that's too much of state change). Because of that, we need to filter modes by rememberability _before_ they are passed to `add-modes-to-auto-mode-rules`.
2. We need to pass `mode-instance` list to `add-modes-to-auto-mode-rules`, and I've made `rememberable-of` to call `mode-instances` implicitly. Is that the right choice? I'm not sure.

Both of these can easily be solved if we check rememberability in `add-modes-to-auto-mode-rules`. Let me test how the mode instantiation works for that and I'll push it.

source/auto-mode.lisp
aartaka: Didn't know about that 0_o Will fix!

aartaka:
Okay, instantiating modes works for `rememberable-p` check. All fixed now :)

aartaka:
Should this be merged now or after 2.0? `root-mode` slots changed, after all -- and it's an API change.

Ambrevar:
I'd say now.

source/auto-mode.lisp
Ambrevar: When you wrote `mode-instance` I suppose you meant `mode-invocation`s?

Anyways, the new patch introduces less changes and is clearer, so I guess we are all good!

Ambrevar:
Wait, actually I just realized that we may be instantiating modes for nothing here, which is probably not desirable.

First thing, the easy part: in enable-matching-modes you call `rememberable-of` over a list of modes, so no need to wrap with `mode-invocation`, right?

Then, in `add-modes-to-auto-mode-rules` it depends: sometimes we have the modes, sometimes just the invocations.
What shall we do when we only have the invocations?  What about getting the default value of the mode slot with moptilities?

aartaka:
> Wait, actually I just realized that we may be instantiating modes for nothing here, which is probably not desirable.
> 
> First thing, the easy part: in enable-matching-modes you call `rememberable-of` over a list of modes, so no need to wrap with `mode-invocation`, right?

Hmmm... Yes, good point. How about this: make `rememebrable-of` a simple `remove-if` using `rememberable-p` as filtering function, while defining additions `rememberable-p` over `mode-invocation`?

> Then, in `add-modes-to-auto-mode-rules` it depends: sometimes we have the modes, sometimes just the invocations.
> What shall we do when we only have the invocations? What about getting the default value of the mode slot with moptilities?

1. We always have `mode-invocation`s in `add-modes-to-auto-mode-rules`, according to its type declaration.
2. In case of invocations, they can also have arguments changing mode behavior, so we _need_ to instantiate them to know for sure whether they are rememberable or not.
3. Moptilities can help us there, but they don't concider the change that additional arguments bring (see 2). It's the same as with buffer instantiation that we do to know the exact value of the buffer slot at runtime.

To sum up, what I'll do:
- Add `rememberable-p` method over `mode-invocation` and simplify `rememebrable-of`.
  - Use arguments to mode instantiation in `rememberable-p`.

aartaka:
See fae5c2c1db01c95bc59802109e3cba0a5458da96.

PR #1308, opened 2021-04-18
Ups! Typo.

santiagopim:
Just tiny typo, but crashed my config loading because I created the file by myself.
Thanks for your work!

### DIFF
--- a/documents/README.org
+++ b/documents/README.org
@@ -393,11 +393,11 @@ Follow the following steps:
 
 1. Add the following line to your Nyxt init file
   #+begin_src lisp
-    (load-after-system :slynk "~/.config/nyxt/my-slink.lisp")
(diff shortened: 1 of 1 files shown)


Ambrevar:
Good catch!
Looks like @jmercouris used the wrong spelling here :)

Ambrevar:
When you say it "crashed your config", did it crash the browser, or did Nyxt correctly report the error?

santiagopim:
Nyxt correctly reported the error, very accurate, but was an ugly white page :) Can not reproduce here.
Thanks

Ambrevar:
Indeed, the error page has no styling at all.
We will address this when we work on functional theme management.

Ambrevar:
Thanks for the feedback!

PR #1312, opened 2021-04-20
Remove must-match-p

Ambrevar:
must-match-p was never used and it seems that raw sources do a better job.

With sources like yes/no that must match, there are only two options: either "yes" or "no" is selected, or the prompt is canceled.

In the future, I'd also like to add a prompt-buffer option that displays a message instead of the prompt buffer when there are no suggestions.  This implies that the prompt buffer suggestions are determined synchronously at creation time, so we must add an option for that.  But this is a feature different from "must-match-p".

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -15,9 +15,6 @@
 ;; TODO: Performance: Consider computing the attribute values only when active.
 ;; Lazy evaluation would make it easy.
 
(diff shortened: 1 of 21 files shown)


jmercouris:
I can agree with this change! :-) Less lines, the better!

Ambrevar:
Thanks!
Merged with 61d7ffbb8f9ad6ea73663e3bc0d4d793b591b11b.

PR #1314, opened 2021-04-21
Remove object-display and object-string

Ambrevar:
Both methods were used in mostly 3 cases:

- The minibuffer (which does not exist anymore).
- Inserting a string value of the object in the minibuffer input.
- Rendering URLs as unicode strings.

The first 2 are fully superseded by the new prompter buffer architecture.

So I've replaced the last one by a `render-url` function, which by the way fixes a few spots where we didn't render the URL puni codes.

Let me know what you think and test around a little bit.


jmercouris:
Will test tomorrow!

jmercouris:
This has introduced a few instances of the following:

```
The value
  "https://news.ycombinator.com/"
is not of type
  QURI.URI:URI
when binding URL
   [Condition of type TYPE-ERROR]

Restarts:
 0: [ABORT] abort thread (#<THREAD "Anonymous thread" RUNNING {100A0CBFC3}>)

Backtrace:
  0: (RENDER-URL "https://news.ycombinator.com/") [external]
  1: ((:METHOD PROMPTER:OBJECT-ATTRIBUTES (NYXT/WEB-MODE::HINT)) #<NYXT/WEB-MODE::LINK-HINT {1009BF5A53}>) [fast-method]
  2: ((:METHOD PROMPTER:MAKE-SUGGESTION (T T)) #<NYXT/WEB-MODE::LINK-HINT {1009BF5A53}> #<NYXT/WEB-MODE::HINT-SOURCE {100A0CB7D3}> NIL) [fast-method]
  3: ((:METHOD PROMPTER::ENSURE-SUGGESTIONS-LIST (PROMPTER:SOURCE T)) #<NYXT/WEB-MODE::HINT-SOURCE {100A0CB7D3}> (#<NYXT/WEB-MODE::LINK-HINT {1009BF5A53}> #<NYXT/WEB-MODE::IMAGE-HINT {1009CB3883}> #<NYXT/W..
  4: ((LAMBDA NIL :IN INITIALIZE-INSTANCE))
  5: ((LAMBDA NIL :IN BORDEAUX-THREADS::BINDING-DEFAULT-SPECIALS))
  6: ((FLET SB-UNIX::BODY :IN SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE))
  7: ((FLET "WITHOUT-INTERRUPTS-BODY-4" :IN SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE))
  8: ((FLET SB-THREAD::WITH-MUTEX-THUNK :IN SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE))
  9: ((FLET "WITHOUT-INTERRUPTS-BODY-1" :IN SB-THREAD::CALL-WITH-MUTEX))
 10: (SB-THREAD::CALL-WITH-MUTEX #<CLOSURE (FLET SB-THREAD::WITH-MUTEX-THUNK :IN SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE) {7F50CCDD6D8B}> #<SB-THREAD:MUTEX "thread result lock" owner: #<SB-THREAD:THREAD "Ano..
 11: (SB-THREAD::NEW-LISP-THREAD-TRAMPOLINE #<SB-THREAD:THREAD "Anonymous thread" RUNNING {100A0CBFC3}> NIL #<CLOSURE (LAMBDA NIL :IN BORDEAUX-THREADS::BINDING-DEFAULT-SPECIALS) {100A0CBF6B}> NIL)
 12: ("foreign function: call_into_lisp")
 13: ("foreign function: new_thread_trampoline")
 --more--

```

The above occurred when using element hints to navigate to a page. I was also able to trigger this same condition by `(make-buffer :url "https://www.duckduckgo.com")`.

Another thing I may suggest is not replacing `object-string` with `render-url`. The only difference between the two is the name. Since `object-string` is a method that specializes on a `quri:uri` object, it feels a little more natural.

Ambrevar:
Thanks for the report.
Actually, the errors are good because they highlight parts where we are not consistently using the quri:uri type, while we should.

The rationale behind removing object-string is that the method is used for only one type, quri:uri, and the name is not as meaningful as render-uri (which actually _renders_ puni codes to unicode).

jmercouris:
Well, the argument was actually a URL, not a URI. URL being a string, and URI being a QURI object.

Ambrevar:
In the past we had discussed to
- always use the quri:uri type
- always use the "URL" nomenclature.

Ambrevar:
I've just fixed the issues you've reported.

jmercouris:
Ah, I had understood that when we use URL, we mean a string, and when we use URI we mean a QURI object. Is that not correct?

Ambrevar:
This was the informal practice some months ago, until we had a
discussion together with @aartaka and we decided the above.

Some more points about the rational that I can remember:

- We should only manipulate QURI objects because they are more
structured and allow for method specialization (unlike strings).

- Sticking to just one format removes a whole class of confusions.

- Plus it saves us back back and forth conversions.

- https://en.wikipedia.org/wiki/Uniform_Resource_Identifier and
  https://en.wikipedia.org/wiki/Uniform_Resource_Locator have different
  meanings indeed, but this has nothing to do with the data
  representation of it.
  Making such distinction as we used to is confusing because it
  has nothing to do with the original meaning or URI and URL.

Second thoughts?

Ambrevar:
Another reason I like `render-url` more than `object-string` is that the name highlights that the function is more costly (FFI call!) than just turning the struct into a string.

jmercouris:
OK, I can accept that.

Ambrevar:
OK to merge then?

jmercouris:
Give me a few more minutes with it, I want to look a little bit more :-)

jmercouris:
Where is the FFI in `render-url`?

```
(defun render-url (url)
    "Return decoded URL.
If the URL contains hexadecimal-encoded characters, return their unicode counterpart."
  (let ((url (if (stringp url)
                 url
                 (quri:render-uri url))))
         (or (ignore-errors url)
             url))))
```

Ambrevar:
Oops!
Thanks for the catch, I've just force-pushed the correct fix.

jmercouris:
Most everything appears to work... I haven't found any problems.

Ambrevar:
Merged with eb418c140b12defc0617b90862cf2b3a0769673b.

aadcg:
This breaks `match-regex` and `match-url`.  Could someone please take a look?

jmercouris:
Is `match-regex` and `match-url` reliant on the existence of object-string for URLs?

aadcg:
Wait. I misunderstood the problem. Everything is ok.

jmercouris:
No problem!

PR #1316, opened 2021-04-22
Rename: %slot-default -> %slot-default%

jmercouris:
%xyz% is for special local variables

for some reason, this does not work, I do not believe to have missed any instances. Any idea why?


Ambrevar:
What's the error?

jmercouris:
It fails on my configuration saying %slot-default% is undefined.

jmercouris:
I see the problem now! Projectile replace failed me!

aadcg:
> Projectile replace failed me!

A suggestion if you're on emacs 27. Run `project-find-regexp`, an `*xref*` buffer pops up, and then run `xref-query-replace-in-results` (which by default is bound to `r`).

PR #1317, opened 2021-04-22
Add element-hint-mode package.

Ambrevar:
Indeed, toggle-hints-transparency belongs to this mode and thus the package.
Without this, "f1 b" in the prompt buffer would fail to list toggle-hints-transparency.


jmercouris:
OK!

jmercouris:
Thank you :-)

PR #1318, opened 2021-04-22
libraries/prompter: Remove unused keymap support from the prompter li…

Ambrevar:
…brary.

Nyxt uses its own mode-associated keymaps.

This is a trivial pull request but I'm opening it to start a discussion: should prompters support keymaps?
Can anyone see some use to it?


jmercouris:
I do not feel it is the job of the prompter library.

PR #1319, opened 2021-04-22
libraries/prompter: Fix behaviour with case sensitivity.

Ambrevar:
Previously, non-downcase input would not update the match-data properly.
This is because match-data was computed just one, on initialization, before any
input is given.

This is wrong, we must obviously recompute the match-data on every `update`.
Most importantly, we must leave full control the match-data to the filter
functions.  Indeed, it does not have to be just strings.

To that add, we introduce ensure-match-data-string which is called from our
default filter functions when they need it.

Instead of calculating if the input is upcased of downcased for each
suggestion (it could be slow since it's a performance bottleneck), we do it only
when `update` is called and store the result in `current-input-downcase-p`.

Finally:

- The automatic match-data initialization is removed from `initialize-instance`
and the dedicated `match-data` reader, since NIL is actually an acceptable
value.

- `filter` functions follow the same argument logic as `filter-preprocessor`
and `filter-postprocessor`, so they now take `(SUGGESTION SOURCE INPUT)`.

- `to-unicode` has been renamed to the more appropriate
`ensure-non-base-string`.  It is called only when needed:
  - Once per `update` for `input`.
  - When setting match-data in `ensure-match-data-string`.

### DIFF
--- a/libraries/prompter/filter-preprocessor.lisp
+++ b/libraries/prompter/filter-preprocessor.lisp
@@ -28,11 +28,12 @@ If any input substring matches exactly (but not necessarily a whole word),
 then all suggestions that are not exactly matched by at least one substring are removed.
 
 Suitable as a `source' `filter-preprocessor'."
-  (declare (ignore source))
   (unless (str:empty? input)
     (let ((exactly-matching-substrings (find-exactly-matching-substrings
                                         input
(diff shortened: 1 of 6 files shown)


Ambrevar:
Merged with 5956ca41635b6d0af4900500892490a4e481b98d.

PR #1324, opened 2021-04-25
Prompter keymaps (emacs, VI!)

Ambrevar:
The long awaited VI bindings for the prompt buffer!

This induces some significant changes:

- The prompt buffer mode list is now displayed to the right of the input.  It's empty by default.
- I've emptied the prompt-buffer override map because `escape` would conflict with vi-insert-mode `escape`.  Can you think of a better way?
- Mode togglers now apply to `current-prompt-buffer`, if any, with higher priority than `current-buffer`.  This may be quite a disruptive change.

To do: test with the `enable-mode-*` and `disable-mode-*` commands still behave as expected.

Thoughts?

Tests are welcome! :)

### DIFF
--- a/source/base-mode.lisp
+++ b/source/base-mode.lisp
@@ -58,7 +58,6 @@ This mode is a good candidate to be passed to `make-buffer'."
                      "C-/" 'reopen-buffer
                      "C-shift-t" 'reopen-buffer
                      "C-T" 'reopen-buffer
-                     "M-i" 'focus-first-input-field
                      "C-p" 'print-buffer)
 
                     scheme:emacs
--- a/source/mode.lisp
+++ b/source/mode.lisp
@@ -58,7 +58,7 @@ Example:
          ,(unless (eq name 'root-mode)
             `(define-command ,name (&rest args
                                     &key
-                                    (buffer (current-buffer))
+                                    (buffer (or (current-prompt-buffer) (current-buffer)))
                                     (activate t explicit?)
                                     &allow-other-keys)
                ,docstring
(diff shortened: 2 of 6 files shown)


Ambrevar:
To enable emacs or VI bindings in the prompt buffer, you can use a configuration snippet like the following:

```lisp
(define-configuration (prompt-buffer)
  ((default-modes (append '(vi-normal-mode) %slot-default%))))
```

jmercouris:
Thank you Pierre :-) ! I will take a look at this on Monday when I am working again.

Ambrevar:
Let's do the following: restore the override, and instead bind `C-c` in VI mode to exit the prompt buffer.
Benefit: no risk of double-pressing escape and accidentally quit the prompt buffer.

Ambrevar:
About `(or (current-prompt-buffer) (current-buffer))`, maybe we should have a third function, like `(focused-buffer)`.

By the way, how does it interact with `with-current-buffer`?

jmercouris:
I've made two variations. Please see the branches: prompt-buffer1 and prompt-buffer2.

I am OK with either, though am personally in favor of prompt-buffer2.

Ambrevar:
Thanks, I've opted for prompt-buffer2 as well, first one was confusing in my opinion.

Ambrevar:
Update: I was wrong, we can't restore the override because `escape` would not be usable to leave insert mode.  And we don't want to use `C-c` to leave insert mode, it's too challenging for VI users muscle memory.

Ambrevar:
`enable-modes` works.

Ambrevar:
`with-current-buffer` also seems unaffected by this change.

Ambrevar:
Merged with 
58429f44778e1b74be35c95cf3bccb7e3ad99949.

PR #1325, opened 2021-04-26
Add support for lisp:// URLs in buffer-load.

Ambrevar:
This fixes an important issue:

- Start Nyxt.
- `follow-hint` "common settings".

Tadah!  Indeed, `follow-hint` goes through buffer-load, which used to create a new web buffer if the current buffer was internal, and from there `request-resource` would refuse to evaluate the Lisp URL because the new current buffer is a web buffer.

@aartaka @jmercouris We've discussed the scheme handling a lot recently.
This is not as good as proper scheme handling like we are discussing for cl-webkit, but I think it's a good workaround for now.
What do you think?

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -900,13 +900,19 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
     (when new-url
       (check-type new-url quri:uri)
       (setf url new-url)
-      (if (internal-buffer-p buffer)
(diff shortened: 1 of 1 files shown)


jmercouris:
Merci monsieur! Yes, I think it could be factored out as well.

Ambrevar:
Merged with 18c5d9825676d7c281dd13e45e6759faa4f07cc7.

PR #1326, opened 2021-04-26
Add cruise-control-mode.lisp

jmercouris:
For post 2.0.0.


source/cruise-control-mode.lisp
```diff
@@ -0,0 +1,63 @@
...
+(define-command velocity-incf (&key (cruise-control (current-cruise-control)))
+  "Increase the velocity."
+  (incf (velocity cruise-control)))
+
+(define-command velocity-decf (&key (cruise-control (current-cruise-control)))
+  "Decrease the velocity."
+  (decf (velocity cruise-control)))
+
+(define-command velocity-zero (&key (cruise-control (current-cruise-control)))
+  "Zero the velocity."
```
Ambrevar, line 62: Does it mean scrolling stops?

source/cruise-control-mode.lisp
```diff
@@ -0,0 +1,63 @@
...
+                        (when (not (zerop (velocity mode)))
+                          (with-current-buffer (buffer mode)
+                            (nyxt/web-mode::scroll-down
+                             :scroll-distance (velocity mode))))))))))
+   (destructor
+    (lambda (mode)
+      (bt:destroy-thread (thread mode))))))
+
+(defun current-cruise-control (&key (buffer (current-buffer)))
+  (find-submode buffer 'cruise-control-mode))
```
Ambrevar, line 51: Should we add a `current-mode` function?
Could be called this way: `(current-mode 'cruise-control)` and automatically append "-mode" to the symbol.

source/cruise-control-mode.lisp
```diff
@@ -0,0 +1,63 @@
...
+       "k" 'velocity-decf
+       "j" 'velocity-incf)))
+   (constructor
+    (lambda (mode)
+      (setf (thread mode)
+            (bt:make-thread
+             (lambda ()
+               (loop while t
+                     do (sleep (poll-sleep-time mode))
+                        (when (not (zerop (velocity mode)))
```
Ambrevar, line 42: `(when (not...))` -> `(unless ...)`

source/cruise-control-mode.lisp
```diff
@@ -0,0 +1,63 @@
...
+    (define-scheme "cruise-control-mode"
+      scheme:cua
+      (list
+       "0" 'velocity-zero
+       "up" 'velocity-decf
+       "down" 'velocity-incf)
+      scheme:emacs
+      (list
+       "p" 'velocity-decf
+       "n" 'velocity-incf)
```
Ambrevar, line 30: This means `p` and `n` can't be inserted while cruise control mode is enabled.  I guess that OK since the mode is meant for viewing only, right?

source/cruise-control-mode.lisp
```diff
@@ -0,0 +1,63 @@
...
+      scheme:vi-normal
+      (list
+       "k" 'velocity-decf
+       "j" 'velocity-incf)))
```
Ambrevar, line 34: Don't we want to leave the scroll key accessible so that use can accelerate on top of the auto scrolling?

Ambrevar:
Nice!

What about adding an escape key to disable the mode?

source/cruise-control-mode.lisp
jmercouris: Yes, that would be good.

source/cruise-control-mode.lisp
jmercouris: Correct!

source/cruise-control-mode.lisp
jmercouris: Correct!

source/cruise-control-mode.lisp
Ambrevar: Done in ae156f5c785d7a22385171b4ca59dd3597c309b6.

source/cruise-control-mode.lisp
```diff
@@ -0,0 +1,63 @@
...
+      scheme:emacs
+      (list
+       "p" 'velocity-decf
+       "n" 'velocity-incf)
+      scheme:vi-normal
+      (list
+       "K" 'velocity-decf
+       "J" 'velocity-incf)))
+   (constructor
+    (lambda (mode)
```
aartaka, line 36: `cruise-control-mode` seems to be perfectly subclassable from `process-mode` (#1208). Any reason to not use it?

aartaka: I mean, it's a thread-based automated action mode, so some things (like thread creation and timing) can be abstracted away.

PR #1328, opened 2021-04-26
urls.lisp: Add utils to compare urls.

aadcg:
As discussed with @jmercouris.

This is an abstraction that should be leveraged to avoid `quri:uri=`, `url-equal`, `schemeless-uri=` and others. Equivalence relations are our friends :)

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -126,6 +126,18 @@ This way, HTTPS and HTTP is ignored when comparing URIs."
   (string< (schemeless-url uri1)
            (schemeless-url uri2)))
 
+(declaim (ftype (function (quri:uri quri:uri) boolean) url-equal))
+(defun url-equal (url1 url2)
+  "URLs are equal if the URIs are equal, scheme excluded.
+Empty paths are also excluded from the comparison.
+For instance, these are equal:
+- http://example.org
+- https://example.org/"
+  (if (and (quri:uri-http-p url1)
+           (quri:uri-http-p url2))
+      (schemeless-uri= url1 url2)
+      (the (values boolean &optional) (quri:uri= url1 url2))))
+
 (declaim (ftype (function (string) (values (or quri:uri null) t &optional)) parse-url))
 (defun parse-url (input-url)
   "From user input, return the full URL to visit.
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -185,3 +185,28 @@ functionality into internal-buffers."
        (apply #'str:concat "lisp://"
               (mapcar (alex:compose #'quri:url-encode #'write-to-string)
                       (cons lisp-form more-lisp-forms)))))
+
+(declaim (ftype (function (quri:uri quri:uri) boolean) path=))
+(defun path= (url1 url2)
+  "Return non-nil when URL1 and URL2 have the same path."
+  ;; See https://github.com/fukamachi/quri/issues/48.
+  (equalp (string-right-trim "/" (or (quri:uri-path url1) ""))
+          (string-right-trim "/" (or (quri:uri-path url2) ""))))
+
+(declaim (ftype (function (quri:uri quri:uri) boolean) scheme=))
+(defun scheme= (url1 url2)
+  "Return non-nil when URL1 and URL2 have the same scheme.
+HTTP and HTTPS belong to the same equivalence class."
+  (or (equalp (quri:uri-scheme url1) (quri:uri-scheme url2))
+      (and (quri:uri-http-p url1) (quri:uri-http-p url2))))
+
+(declaim (ftype (function (quri:uri quri:uri) boolean) domain=))
+(defun domain= (url1 url2)
+  "Return non-nil when URL1 and URL2 have the same domain."
+  (equalp (quri:uri-domain url1) (quri:uri-domain url2)))
+
+(declaim (ftype (function (quri:uri quri:uri) boolean) host=))
+(defun host= (url1 url2)
+  "Return non-nil when URL1 and URL2 have the same host.
+This is a more restrictive requirement than `domain='."
+  (equalp (quri:uri-host url1) (quri:uri-host url2)))
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -128,15 +128,19 @@ This way, HTTPS and HTTP is ignored when comparing URIs."
 
 (declaim (ftype (function (quri:uri quri:uri) boolean) url-equal))
 (defun url-equal (url1 url2)
-  "URLs are equal if the URIs are equal, scheme excluded.
-Empty paths are also excluded from the comparison.
-For instance, these are equal:
-- http://example.org
-- https://example.org/"
-  (if (and (quri:uri-http-p url1)
-           (quri:uri-http-p url2))
-      (schemeless-uri= url1 url2)
-      (the (values boolean &optional) (quri:uri= url1 url2))))
+  "Like `quri:uri=' but ignoring the scheme.
+URLs are equal up to `scheme='."
+  (url-eqs url1
+           url2
+           (list #'scheme= 
+                 (lambda (url1 url2) (equal (or (quri:uri-path url1) "/")
+                                       (or (quri:uri-path url2) "/")))
+                 (lambda (url1 url2) (equal (quri:uri-query url1)
+                                       (quri:uri-query url2)))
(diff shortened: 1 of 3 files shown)


source/urls.lisp
```diff
@@ -126,6 +126,54 @@ This way, HTTPS and HTTP is ignored when comparing URIs."
...
+      (and (quri:uri-http-p url1) (quri:uri-http-p url2))))
+
+(defun domain= (url1 url2)
+  "Return non-nil when URL1 and URL2 have the same domain."
+  (equalp (quri:uri-domain url1) (quri:uri-domain url2)))
+
+(defun host= (url1 url2)
+  "Return non-nil when URL1 and URL2 have the same host.
+This is a more restrictive requirement than `domain='."
+  ;; What's the difference between quri:uri-host and quri:uri-authority?
```
Ambrevar, line 174: uri-authority includes the user:password information.

source/urls.lisp
```diff
@@ -126,6 +126,54 @@ This way, HTTPS and HTTP is ignored when comparing URIs."
...
+  ;; nil, unlike the solution below.
+  (every #'identity (mapcar (lambda (fn) (funcall fn url1 url2)) eq-fn-list)))
+
+(declaim (ftype (function (quri:uri quri:uri) boolean)
+                distinct-url-path-p scheme= domain= host=))
+(defun distinct-url-path-p (url1 url2)
+  "Return non-nil when URL1 and URL2 have distinct paths."
+  ;; See https://github.com/fukamachi/quri/issues/48.
+  (not (equalp (string-right-trim "/" (or (quri:uri-path url1) ""))
+               (string-right-trim "/" (or (quri:uri-path url2) "")))))
```
Ambrevar, line 159: This is unconventional logic.  What about `path=` instead, and then in the caller you can either `(not (path=...))` or `(complement #'path=)`?

source/urls.lisp
```diff
@@ -126,6 +126,54 @@ This way, HTTPS and HTTP is ignored when comparing URIs."
...
+(defun url-equal (url1 url2)
+  "URLs are equal if the URIs are equal, scheme excluded.
+Empty paths are also excluded from the comparison.
+For instance, these are equal:
+- http://example.org
+- https://example.org/"
+  (if (and (quri:uri-http-p url1)
+           (quri:uri-http-p url2))
+      (schemeless-uri= url1 url2)
+      (the (values boolean &optional) (quri:uri= url1 url2))))
```
Ambrevar, line 139: Don't forget to remove `equal-url` from bookmark.lisp ;)

source/urls.lisp
```diff
@@ -126,6 +126,54 @@ This way, HTTPS and HTTP is ignored when comparing URIs."
...
+EQ-FN-LIST is a list of functions that take URL1 and URL2 as arguments and
+return a boolean.  It defines an equivalence relation induced by EQ-FN-LIST.
+`quri:uri=', `url-equal' and `schemeless-uri=' are examples of equivalence
+relations."
+  ;; (and (fn1 url1 url2) (fn2 url1 url2) ...) stops as soon as any fn returns
+  ;; nil, unlike the solution below.
+  (every #'identity (mapcar (lambda (fn) (funcall fn url1 url2)) eq-fn-list)))
+
+(declaim (ftype (function (quri:uri quri:uri) boolean)
+                distinct-url-path-p scheme= domain= host=))
```
Ambrevar, line 154: Nit: I prefer to put the type declaration right above the function definition, this helps us prevent the declaim from going out ot sync with the actual function signature when it's updated.

source/urls.lisp
```diff
@@ -126,6 +126,54 @@ This way, HTTPS and HTTP is ignored when comparing URIs."
...
+(defun eq-uri-p (url1 url2 eq-fn-list)
+  "Return non-nil when URL1 and URL2 are \"equal\" as dictated by EQ-FN-LIST.
+
+EQ-FN-LIST is a list of functions that take URL1 and URL2 as arguments and
+return a boolean.  It defines an equivalence relation induced by EQ-FN-LIST.
+`quri:uri=', `url-equal' and `schemeless-uri=' are examples of equivalence
+relations."
+  ;; (and (fn1 url1 url2) (fn2 url1 url2) ...) stops as soon as any fn returns
+  ;; nil, unlike the solution below.
+  (every #'identity (mapcar (lambda (fn) (funcall fn url1 url2)) eq-fn-list)))
```
Ambrevar, line 151: Why not, but do you have example uses?

I would name this function `url-eqs` for consistency with `sera:eqs`.
Also we adopted the convention to always use "URL" and not "URI".

Ambrevar:
Very welcome changes, thank you!

source/urls.lisp
aadcg: Ah, thanks!

source/urls.lisp
aadcg: Sure, I agree.

source/urls.lisp
aadcg: Good point!

source/urls.lisp
aadcg: > Why not, but do you have example uses?

I noticed that there were some functions using this pattern (think of `quri:uri=`, `url-equal`, `schemeless-uri=`, etc).  Since a URL an entity composed by many sub-entities, the concept of equivalence is vague.  In short, one should be able to define their own equivalence relation easily.  

I noticed it while working on #1246. Take a look [here](https://github.com/atlas-engineer/nyxt/blob/21fa6e8cbe9f380a7b1a397d74d70afbe03ba48f/source/renderer-script.lisp#L101).  

Perhaps, this could be included in the quri library, but this is a decision that I leave up to you of course.

I will rewrite the functions on Nyxt's side for the time being.

> I would name this function url-eqs for consistency with sera:eqs.
Also we adopted the convention to always use "URL" and not "URI".

Ok, I agree. I think @jmercouris will agree too.

Ambrevar: OK, thanks!

aadcg:
In general, should look good now.

There's a compiler note I'm not understanding tho:

```

; file: [...]/nyxt/source/urls.lisp
; in: DEFUN URL-EQUAL
;     (DEFUN NYXT::URL-EQUAL (NYXT::URL1 NYXT::URL2)
;       "Like `quri:uri=' but ignoring the scheme.
;   URLs are equal up to `scheme='."
;       (NYXT::URL-EQS NYXT::URL1 NYXT::URL2
;                      (LIST #'NYXT::SCHEME=
;                            (LAMBDA (NYXT::URL1 NYXT::URL2) (EQUAL # #))
;                            (LAMBDA (NYXT::URL1 NYXT::URL2) (EQUAL # #))
;                            (LAMBDA (NYXT::URL1 NYXT::URL2) (EQUAL # #))
;                            (LAMBDA (NYXT::URL1 NYXT::URL2) (EQUALP # #)))))
; --> PROGN SB-IMPL::%DEFUN SB-IMPL::%DEFUN SB-INT:NAMED-LAMBDA 
; ==>
;   #'(SB-INT:NAMED-LAMBDA NYXT::URL-EQUAL
;         (NYXT::URL1 NYXT::URL2)
;       (DECLARE (SB-C::TOP-LEVEL-FORM))
;       "Like `quri:uri=' but ignoring the scheme.
; URLs are equal up to `scheme='."
;       (BLOCK NYXT::URL-EQUAL
;         (NYXT::URL-EQS NYXT::URL1 NYXT::URL2
;                        (LIST #'NYXT::SCHEME= (LAMBDA # #) (LAMBDA # #)
;                              (LAMBDA # #) (LAMBDA # #)))))
; 
; (VALUES &OPTIONAL BOOLEAN &REST T).
; (VALUES &OPTIONAL BOOLEAN).
```

aadcg:
`url-equal` is not the best name, so feel free to give suggestions.

jmercouris:
It seems your declaim statement is misformed- or impossible to check.

Ambrevar:
SBCL is not smart enough to deduce the type of the return value.
Wrap it with `(the (values TYPE &optional) FORM)`.
We do this in a few places in the code base, look up `(the`.

Ambrevar:
Merged with d0ba4457c4512bb1f4408a051c5b15b8fe347aee.
Thanks!

aadcg:
Thanks @Ambrevar.

PR #1332, opened 2021-04-27
buffer.lisp: Add buffer-loaded-hook.

aadcg:


aartaka:
This looks like a duplicate of `on-signal-load-finished`. Do you have some specific use-case in mind?

jmercouris:
There was a specific use case, Andre was loading buffers, but did not want to add a mode to them. He just wanted to add a hook to know when they were finished loading. It was part of recursive search.

aadcg:
Indeed, this seems to come out of the blue. I'm splitting #1246 into digestible hunks.

`on-signal-load-finished` works when you have a class, or a mode. But this doesn't suffice for my application.

aadcg:
There could be a mode, say `buffer-temporary-mode`, to be enabled in those buffers that are solely created so that links can be fetched.  But it seems like a terrible idea. Isn't it legit to have a hook when a web buffer is loaded?

aartaka:
Fair points!

Side note: I'm really surprised that recursive search needs no mode to function, because it seems necessary to store data somewhere, and mode is best for that ~_~

aadcg:
> Side note: I'm really surprised that recursive search needs no mode to function, because it seems necessary to store data somewhere, and mode is best for that ~_~

A mode is a behavior of a buffer, correct? In the way I conceptualized the problem, I don't see how could a mode help me with the problem. Think of `search-buffer`. No modes are involved. Same goes for a recursive search.  You just need to be smart about computing the buffers or URLs that will be searched for.

source/buffer.lisp
```diff
@@ -272,6 +277,12 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
   (when (expand-path (cookies-path buffer))
     (ensure-parent-exists (expand-path (cookies-path buffer)))))
 
+(defmethod (setf load-status) (status (buffer web-buffer))
```
aartaka, line 280: How about moving the hook call to the default `on-signal-load-finished` method a couple hundred lines down from here? `on-signal...` is called only when WebKit actually says that the loading is finished, while this custom setter relies on the slot value that can be altered for reasons other than page loading. Am I over-engineering here?

aadcg: Your rationale makes sense to me. My implementation follows @jmercouris advice, since I was a bit lost at that time and asked for his advice. What do you think @jmercouris?

aadcg: Wait... but that also relies on the modes. So it's of no use, right?

aadcg: Ok @aartaka, your suggestion works indeed for me.  There are two mysteries I can't wrap my head around though. What does 
```
(dolist (mode (modes buffer))
    (on-signal-load-finished mode url))
```

do to `on-signal-load-finished`? 

The second is a complete mystery to me.

Why does this fail?
```
(defmethod on-signal-load-finished ((buffer buffer) url)
  (dolist (mode (modes buffer))
    (on-signal-load-finished mode url))
  (print
      (with-current-buffer buffer
        (pflet ((nlinks () (ps:chain document links length))
                (fetch-link (i) (ps:chain document links (item (ps:lisp i)) href)))
          (loop for i from 0 to (1- (parse-integer (nlinks)))
                collect (quri:uri (fetch-link i)))))))
```

And this works?

```
(defmethod on-signal-load-finished ((buffer buffer) url)
  (dolist (mode (modes buffer))
    (on-signal-load-finished mode url))
  (bt:make-thread
   (lambda ()
     (print
      (with-current-buffer buffer
        (pflet ((nlinks () (ps:chain document links length))
                (fetch-link (i) (ps:chain document links (item (ps:lisp i)) href)))
          (loop for i from 0 to (1- (parse-integer (nlinks)))
                collect (quri:uri (fetch-link i)))))))))
```

aartaka: The second I can't answer, as I don't understand the reason myself.

As for the first — this method call is our own way of event-passing. What we do is call the `on-signal-load-finished` of every enabled mode. Of course, if the mode implements this method. 
- If it doesn't — nothing happens, because there's a default implementation somewhere that just returns nil.
- If it does implement the method, then this method is a useful way to do things on page loading. So, the sequence of actions is:
  - WebKit loads the page.
  - WebKit  sends `load-changed` signal with finished status.
  - We catch the signal and call the `on-signal-load-finishes` on buffer.
  - Buffer propagates this signal to all the modes (with this `dolist`.)
  - Modes react to the signal in whatever way they want, like setting custom HTML or running Parenscript scripts :)

------------------------------------------------
Is the code you've shown an implementation you have in mind? If so, I may have confused you. What I meant was calling a hook in the `on-signal-load-finished`, i.e., adding
``` lisp
    (bt:make-thread
     (lambda () (hooks:run-hook (buffer-loaded-hook buffer) buffer)))
```

After the `dolist` in `on-signal-load-finished`. This way we can ensure both configurability (that is indeed necessary in case of recursive search) and proper place/event to call the hook at. Win-win :)

aadcg: > The second I can't answer, as I don't understand the reason myself.

No worries. I just took the opportunity to get another pair of eyeballs in the hope that we could figure something out.  Notice that I'm bringing this up, since I have the following in the pull request:

```
(bt:make-thread
     (lambda () (hooks:run-hook (buffer-loaded-hook buffer) buffer)))
```

But, a priori, one would expect that the following would suffice:

```
(hooks:run-hook (buffer-loaded-hook buffer) buffer)
```

---

Thanks for your explanation regarding `on-signal-load-finished`.  I was confused since modes don't inherit from buffers, but now all is clear. 

> If it doesn't — nothing happens, because there's a default implementation somewhere that just returns nil.

Indeed. That's `on-signal-load-finished` at `mode.lisp`, where the method is implemented for `root-mode`, from which all modes inherit.

---

> Is the code you've shown an implementation you have in mind? If so, I may have confused you.

Here I'm afraid that I have confused you too! I was indeed showing a particular application I had in mind. I implicitly was following your suggestion here, while at the same time showing you a pathological case I came across.

aadcg:
Looks good to me @aartaka.

Example:

```lisp
(sera:add-hook (buffer-loaded-hook (current-buffer))
	       (make-handler-buffer (lambda (buffer) (print "hello"))
                                    :name 'hello))
```

```lisp
(sera:remove-hook (buffer-loaded-hook (current-buffer)) 'hello)
```

source/buffer.lisp
```diff
@@ -517,7 +521,9 @@ BUFFER's modes."
 (export-always 'on-signal-load-finished)
 (defmethod on-signal-load-finished ((buffer buffer) url)
   (dolist (mode (modes buffer))
-    (on-signal-load-finished mode url)))
+    (on-signal-load-finished mode url))
+  (bt:make-thread
```
Ambrevar, line 525: Never use `bt:make-thread` in the Nyxt source unless you know what you are doing.
Prefer `run-thread`, which is protected.

Ambrevar:
Please replace make-thread with run-thread and we can merge :)

source/buffer.lisp
aadcg: Ok, will keep in mind!

aadcg:
> Please replace make-thread with run-thread and we can merge :)

I would, but then the hook doesn't work anymore... What shall we do?

aadcg:
Ok to merge.

PR #1335, opened 2021-04-28
buffer.lisp: Add function that returns enabled modes.

aadcg:
The modes are returned as mode symbols.

This differs from the modes slot of the buffer class, since that returns
the instances of the enabled modes.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1156,6 +1156,13 @@ ARGS are passed to the mode command."
           (apply #'funcall command :buffer buffer :activate t args)
           (log:warn "Mode command ~a not found." mode)))))
 
+(export-always 'active-modes)
+(defun active-modes (&optional (buffer (current-buffer)))
+  "Return the enabled modes of BUFFER.
+The modes are the mode symbols, as returned by `mode-name'."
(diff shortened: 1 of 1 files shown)


Ambrevar:
The function can be shortened:

```lisp
(mapcar #'mode-name (modes buffer))
```

But then it's almost as short as calling the wrapper:

```lisp
(active-modes buffer)
```

By the way, a "mode" generally refers to the class instance, not the symbol, so
I'd rather call this wrapper `mode-names`.

So is it worth creating a wrapper?  Do you foresee widespread use?

aadcg:
> So is it worth creating a wrapper? 

I don't think so, so I agree with you.

> so I'd rather call this wrapper `mode-names`

Ok, I can make that change.

Ambrevar:
>> So is it worth creating a wrapper? 
>
> I don't think so, so I agree with you.
>
>> so I'd rather call this wrapper `mode-names`
>
> Ok, I can make that change.

I'm confused: if you think making a wrapper is not useful, then why do
you want to make that change?

aadcg:
> I'm confused: if you think making a wrapper is not useful, then why do
you want to make that change?

Sorry, it's my confusion that confuses you!

PR #1336, opened 2021-04-28
Consistent uri typing

Ambrevar:
Trivial enough it should be OK to merge, I just want to double-check with you @jmercouris if we are on the same page.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -920,13 +920,13 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
          (ffi-buffer-evaluate-javascript buffer (quri:url-decode (quri:uri-path url))))
         (t (ffi-buffer-load buffer url))))))
 
-(defun new-buffer-load (url)
+(defun new-buffer-load (suggestion)
   "Load a URL in a new buffer."
-  (make-buffer-focus :url (url url)))
+  (make-buffer-focus :url (url suggestion)))
 
-(defun new-nosave-buffer-load (url)
+(defun new-nosave-buffer-load (suggestion)
   "Load a URL in a new nosave buffer."
-  (make-buffer-focus :url (url url) :nosave-buffer-p t))
+  (make-buffer-focus :url (url suggestion) :nosave-buffer-p t))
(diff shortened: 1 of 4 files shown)


aadcg:
`(cons quri:uri *)` - I have been wondering how to say this, thanks!

Ambrevar:
> `(cons quri:uri *)` - I have been wondering how to say this, thanks!

I also discovered this embarassingly late :p
Note that type checking is only done on the first element, not the whole
list.

To type-check a whole list, we must use the `satisfies` type specifier.

source/buffer.lisp
```diff
@@ -920,13 +920,13 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
          (ffi-buffer-evaluate-javascript buffer (quri:url-decode (quri:uri-path url))))
         (t (ffi-buffer-load buffer url))))))
 
-(defun new-buffer-load (url)
+(defun new-buffer-load (suggestion)
```
jmercouris, line 923: instead of taking a suggestion, why not take a quri:uri? I think it should be the responsibility of `set-url` to ensure that we are returning a quri:uri object.

Another idea: we could view all objects that represent URLs in some way as conforming to an API.

We could have them inherit from a class, something like `uri-protocol`. This class would have:

```
(defgeneric uri (uri-protocol)
"Return the string URL representation of this object"
(error))
```

As an example, a bookmark could inherit from `uri-suggestion`, and we would implement the above generic method for it. This would allow us to treat all objects that deal with URLs in the same manner.

What do you think? Does that make sense?

Ambrevar: The idea of a protocol is fine by me, but we can leave this for later I think.

To ansewr your first question, it's not set-url here because we haven't returned from the prompt buffer yet when we are running an action!
Actions take suggestions and there is no way around it :)

jmercouris: OK! An agreeable conclusion!

Ambrevar:
OK to merge then?

jmercouris:
OK!

PR #1337, opened 2021-04-29
Buffer list filtering

aadcg:
This a first draft to discuss `buffer-list`.

In my view, the way I have implemented `buffer-list` is adequate and flexible.  It can merely return `(alex:hash-table-values (buffers *browser*))`; or functions, taking it as argument, are passed in such a way that the transformations are composed/chained.

This flexibility is overkill if one just wants to pass a single transformation.

Say we'd want to sort by time:

Up to know we had a nice syntax, `(buffer-list :sort-by-time t)`, but `buffer-list` was defined poorly.

With the new definition, these would be equivalent:

1)

```lisp
(sort (buffer-list) #'local-time:timestamp> :key #'last-access)
```

2)

```lisp
(buffer-list :fn-list (list (lambda (x) (sort x
                                              #'local-time:timestamp>
                                              :key #'last-access))))
```

3)

```lisp
(defun sort-by-time (x)
  (sort x #'local-time:timestamp> :key #'last-access))

(buffer-list :fn-list (list #'sort-by-time))
```

1) is by far the sanest solution, IMO.

---

Now another question comes when one wants to pass 2 or more transformations. We want something akin to 2) or 3).  In other words, does one want to define beforehand the functions passed in `:fn-list`? I'd say no.  We can do that when, and where we need it. I'd stick with 2). What sense would it make to define such functions when thinking about `buffer-list`, when they're generally defined for sequences?

The most complicated case we have is `(buffer-list :sort-by-time t :domain domain)`, which now would become:

(notice that `domain` is obviously unbound.)

```lisp
(buffer-list :fn-list (list (lambda (x) (sort x
                                              #'local-time:timestamp>
                                              :key #'last-access))
                            (lambda (x) (remove-if-not
                                         (lambda (buffer) (equal domain (quri:uri-domain (url buffer))))
                                         x))))
```

---

When we agree on something, I can propagate the changes.

@jmercouris @Ambrevar @aartaka

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -686,15 +686,12 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
     (store (data-profile buffer) (history-path buffer))))
 
 (export-always 'buffer-list)
-(defun buffer-list (&key sort-by-time domain)
-  (let* ((buffer-list (alex:hash-table-values (buffers *browser*)))
-         (buffer-list (if sort-by-time (sort
-                                        buffer-list #'local-time:timestamp> :key #'last-access)
-                          buffer-list))
-         (buffer-list (if domain (remove-if-not
-                                  (lambda (i) (equal domain (quri:uri-domain (url i)))) buffer-list)
-                          buffer-list)))
-    buffer-list))
+(defun buffer-list ()
+  (alex:hash-table-values (buffers *browser*)))
+
+(defun sort-by-time (sequence &key (key #'last-access))
+  "Return a timely ordered SEQUENCE by KEY.  More recent elements come first."
+  (sort sequence #'local-time:timestamp> :key key))
 
 (defun buffers-get (id)
   (gethash id (slot-value *browser* 'buffers)))
@@ -759,8 +756,7 @@ proceeding."
 
 (defun last-active-buffer ()
   "Return buffer with most recent `last-access'."
-  (first (sort (buffer-list)
-               #'local-time:timestamp> :key #'last-access)))
+  (first (sort-by-time (buffer-list))))
 
 (defun get-inactive-buffers ()
   "Return inactive buffers sorted by last-access timestamp, or NIL if none."
@@ -769,7 +765,7 @@ proceeding."
         (buffers (buffer-list)))
     (alex:when-let ((diff (set-difference buffers active-buffers)))
        ;; Display the most recent inactive buffer.
-       (sort diff #'local-time:timestamp> :key #'last-access))))
+       (sort-by-time diff))))
 
 (define-command copy-url ()
   "Save current URL to clipboard."
@@ -782,7 +778,7 @@ proceeding."
   (echo "~a copied to clipboard." (title (current-buffer))))
 
 (defun buffer-initial-suggestions (&key current-is-last-p domain)
-  (let ((buffers (buffer-list :sort-by-time t :domain domain)))
+  (let ((buffers (apply (match-domain domain) (sort-by-time (buffer-list)))))
     (when current-is-last-p
       (setf buffers (alex:rotate buffers -1)))
     buffers))
@@ -819,13 +815,13 @@ proceeding."
     (prompt
      :prompt "Switch to buffer in current domain:"
      :sources (make-instance 'buffer-source
-                             :constructor (buffer-list :sort-by-time t
-                                                       :domain domain)))))
+                             :constructor (apply (match-domain domain)
+                                                 (sort-by-time (buffer-list)))))))
 
 (defun switch-buffer-or-query-domain (domain)
-  "Switch to a buffer if it exists for a given domain, otherwise query
+  "Switch to a buffer if it exists for a given DOMAIN, otherwise query
   the user."
-  (let* ((matching-buffers (buffer-list :domain domain)))
+  (let ((matching-buffers (apply (match-domain domain) (buffer-list))))
     (if (eql 1 (length matching-buffers))
         (set-current-buffer (first matching-buffers))
         (switch-buffer-domain :domain domain))))
@@ -1093,7 +1089,7 @@ generate a new URL query from user input.
 That is to say, the one with the most recent access time after the current buffer.
 The current buffer access time is set to be the last so that if we keep calling
 this command it cycles through all buffers."
-  (let* ((buffers (buffer-list :sort-by-time t))
+  (let* ((buffers (sort-by-time (buffer-list)))
          (last-buffer (alex:last-elt buffers))
          (current-buffer (current-buffer)))
     (when (second buffers)
@@ -1105,7 +1101,7 @@ this command it cycles through all buffers."
 
 (define-command switch-buffer-next ()   ; TODO: Rename switch-buffer-oldest
   "Switch to the oldest buffer in the list of buffers."
-  (let* ((buffers (buffer-list :sort-by-time t))
+  (let* ((buffers (sort-by-time (buffer-list)))
          (oldest-buffer (alex:last-elt buffers)))
     (when (eq oldest-buffer (current-buffer))
       ;; Current buffer may already be the oldest, e.g. if other buffer was
--- a/source/status.lisp
+++ b/source/status.lisp
@@ -42,19 +42,16 @@
 
 (export-always 'format-status-tabs)
 (defun format-status-tabs ()
-  (flet ((buffer-domains ()
-           (remove-duplicates
-            (remove nil
-                    (mapcar #'(lambda (i) (quri:uri-domain (url i)))
-                            (buffer-list :sort-by-time t)))
-            :test #'equal)))
-    (markup:markup
-     (:span
-      (loop for domain in (buffer-domains)
-            collect (markup:markup
-                     (:a :class "tab"
-                         :href
-                         (lisp-url `(nyxt::switch-buffer-or-query-domain ,domain)) domain)))))))
+  (markup:markup
+   (:span
+    (loop for domain in (remove-duplicates
+                         (sera:filter-map #'quri:uri-domain
+                                          (mapcar #'url (sort-by-time (buffer-list))))
+                         :test #'equal)
+          collect (markup:markup
+                   (:a :class "tab"
+                       :href
+                       (lisp-url `(nyxt::switch-buffer-or-query-domain ,domain)) domain))))))
 
 (defun format-status (window)
   (let ((buffer (current-buffer window)))
@@ -62,13 +59,13 @@
      (:div :id "container"
            (:div :id "controls"
                  (markup:raw (format-status-buttons)))
-           (:div :class "arrow arrow-right" 
+           (:div :class "arrow arrow-right"
                  :style "background-color:rgb(80,80,80)" "")
            (:div :id "url"
                  (markup:raw
                   (format-status-load-status buffer)
                   (format-status-url buffer)))
-           (:div :class "arrow arrow-right" 
+           (:div :class "arrow arrow-right"
                  :style "background-color:rgb(120,120,120)" "")
            (:div :id "tabs"
                  (markup:raw
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -19,9 +19,7 @@
   `(("Name" ,(str:downcase (hooks:name handler)))))
 
 (defun get-commands (&optional (buffer (current-buffer)))
-  (sort (apply #'list-commands
-               (mapcar #'mode-name (modes buffer)))
-        #'local-time:timestamp> :key #'last-access))
+  (sort-by-time (apply #'list-commands (mapcar #'mode-name (modes buffer)))))
 
 (defun command-attributes (command &optional (buffer (active-buffer (current-window :no-rescan))))
   (let ((scheme-name (keymap-scheme-name buffer))
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -175,14 +175,14 @@ lot."
 (defun history-html-list (&key (limit 100) ; Export?
                             (separator " → "))
   (with-data-unsafe (history (history-path (current-buffer)))
-    (let* ((history (when history
-                      (mapcar #'first
-                              (sort (alex:hash-table-alist (htree:entries history))
-                                    #'local-time:timestamp>
-                                    :key (lambda (entry-nodes)
-                                           (let ((nodes (rest entry-nodes)))
-                                             (apply #'local-time:timestamp-maximum
-                                                    (mapcar #'htree:last-access nodes)))))))))
+    (let* ((history
+             (when history
+               (mapcar #'first
+                       (sort-by-time (alex:hash-table-alist (htree:entries history))
+                                     :key (lambda (entry-nodes)
+                                            (let ((nodes (rest entry-nodes)))
+                                              (apply #'local-time:timestamp-maximum
+                                                     (mapcar #'htree:last-access nodes)))))))))
       (loop for entry in (sera:take limit history)
             collect (markup:markup
                      (:li (title entry) (unless (str:emptyp (title entry)) separator)
@@ -353,9 +353,8 @@ We keep this variable as a means to import the old format to the new one.")
                  (htree:set-current-owner history (htree::fallback-owner history))))
              (alex:when-let ((latest-id (first
                                          (first
-                                          (sort (alex:hash-table-alist (htree:owners history))
-                                                #'local-time:timestamp>
-                                                :key (alex:compose #'htree:last-access #'rest))))))
+                                          (sort-by-time (alex:hash-table-alist (htree:owners history))
+                                                        :key (alex:compose #'htree:last-access #'rest))))))
                (unless (equal latest-id htree:+default-owner+)
                  (switch-buffer :id latest-id))))
 
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -194,8 +194,8 @@ return a boolean.  It defines an equivalence relation induced by EQ-FN-LIST.
 (export-always 'match-scheme)
 (defun match-scheme (scheme &rest other-schemes)
   "Return a predicate for URLs matching one of SCHEME or OTHER-SCHEMES."
-  #'(lambda (url)
-      (some (alex:curry #'string= (quri:uri-scheme url))
+  #'(lambda (buffer-or-url)
+      (some (alex:curry #'string= (quri:uri-scheme (url buffer-or-url)))
             (cons scheme other-schemes))))
 
 (declaim (ftype (function (string &rest string) (function (quri:uri) boolean))
@@ -203,8 +203,8 @@ return a boolean.  It defines an equivalence relation induced by EQ-FN-LIST.
 (export-always 'match-host)
 (defun match-host (host &rest other-hosts)
   "Return a predicate for URLs matching one of HOST or OTHER-HOSTS."
-  #'(lambda (url)
-      (some (alex:curry #'string= (quri:uri-host url))
+  #'(lambda (buffer-or-url)
+      (some (alex:curry #'string= (quri:uri-host (url buffer-or-url)))
             (cons host other-hosts))))
 
 (declaim (ftype (function (string &rest string) (function (quri:uri) boolean))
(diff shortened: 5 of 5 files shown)


source/buffer.lisp
```diff
@@ -667,18 +667,25 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
...
-                                        buffer-list #'local-time:timestamp> :key #'last-access)
-                          buffer-list))
-         (buffer-list (if domain (remove-if-not
-                                  (lambda (i) (equal domain (quri:uri-domain (url i)))) buffer-list)
-                          buffer-list)))
-    buffer-list))
-
-(defun buffers-get (id)
-  (gethash id (slot-value *browser* 'buffers)))
+(defun buffer-list (&key (fn-list nil fn-list-supplied-p))
```
Ambrevar, line 670: Parameter names should tell the user what they do, not what type they are.  Types can be instructed in a declaim form instead, and the user can consult it from the function documentation.

So instead of `fn-list` I'd name it `filters` for instance.

Also for convenience I'd allow single functions, not necessarily list of functions.  Just wrap the argument use with `(uiop:ensure-list ...)`.

Ambrevar:
I'm OK with this change, but I don't understand what you are asking.  Can you rephrase?

aartaka:
I may be out of context here, but why make a list of functions instead of one function? The caller can use `alex:compose` themselves, while most frequent use-case is a single lambda one. In this situation, a single function argument (`predicate`/`filter`?) fits more than list-of-functions one.

An added benefit of implying one function is that you can use `#'identity` by default and get rid of the argument checking.

What I can definitely agree with, though, is that `buffer-list` needs more generality in sorting/filtering. Thanks for initiating the discussion, André!

aadcg:
You do make a very good point @aartaka.

aadcg:
And which one would you use @aartaka?

```
(buffer-list :filter (lambda (x) (sort x #'local-time:timestamp> :key #'last-access)))
(sort (buffer-list) #'local-time:timestamp> :key #'last-access)
```

aartaka:
Most probably the second one, but that depends on the context :)

aadcg:
Exactly. That's why I don't fully agree with you. Passing the filter/predicate key only makes sense when you need something more convoluted, which will always imply using `alex:compose`.

@Ambrevar suggestion regarding `uiop:ensure-list` is also insightful. I will draft smth.

aadcg:
This is how I see it. 

> I don't understand what you are asking. Can you rephrase?

Look at the changes. It results in more lines then what we had before, right? Basically we need to write the same predicate several times. Do you get my questions and the difference between alternatives 2) and 3) in my initial post?

Ambrevar:
The repeated can be extract to top-level functions, no problem.
Does this bother you?

aadcg:
> Does this bother you?

It does.  For instance, where should a function like this one be located? 

```lisp
(defun sort-by-time (x)
  (sort x #'local-time:timestamp> :key #'last-access))
```

The other functions are even more problematic. It seems silly to write functions to be used EXCLUSIVELY for `buffer-list`. For instance, turning the following into a function has several problems. If `domain` becomes an argument, then I can't pass it to `buffer-list` with the `filter` key. Why? Because the functions that get composed are functions that only take `buffer-list`, i.e. `(alex:hash-table-values (buffers *browser*))` as argument. Any other parameters, such as `domain`, must be in the lexical scope but can't be passed directly as an argument.

```lisp
(lambda (buffer-list) (remove-if-not
                       (lambda (buffer) (equal domain (quri:uri-domain (url buffer))))
                       buffer-list))
```

aartaka:
> The other functions are even more problematic. It seems silly to write functions to be used EXCLUSIVELY for `buffer-list`. For instance, turning the following into a function has several problems. If `domain` becomes an argument, then I can't pass it to `buffer-list` with the `filter` key. Why? Because the functions that get composed are functions that only take `buffer-list`, i.e. `(alex:hash-table-values (buffers *browser*))` as argument. Any other parameters, such as `domain`, must be in the lexical scope but can't be passed directly as an argument.
> 
> ```lisp
> (lambda (buffer-list) (remove-if-not
>                        (lambda (buffer) (equal domain (quri:uri-domain (url buffer))))
>                        buffer-list))
> ```

[`alex:curry`&`alex:rcurry`](https://common-lisp.net/project/alexandria/draft/alexandria.html#Data-and-Control-Flow)?

I'm still on the side of a more conservative&conventional filter argument, as there are tons of functions/marcos supporting this idea in e.g., Alexandria :)

Ambrevar:
You can also use the convenient `sera:equals` to generate the predicate.

aadcg:
You're giving me ideas on how to express the filtering, but that's not the main point. Filtering is a bad word btw. We have a list, and we apply transformation to it. We compose functions that act on it, on a certain order (and the order matters). Filtering does little justice to such a thing, but I can't think of a better word too.

The topic at hand here (for me) is: we need to rewrite calls such as `buffer-list :sort-by-time t :domain domain` without being too verbose and replicating code. Notice that as things stand in master, there's no code replication since the logic behind these keys (`sort-by-time` or `domain`) are part of `buffer-list`.


My solution thus far is:

```lisp
(buffer-list
 :filter (list
          (lambda (buffer-list) (remove-if-not
                            (lambda (buffer)
                              (equal domain
                                     (quri:uri-domain (url buffer))))
                            buffer-list))
          (lambda (buffer-list) (sort buffer-list
                                 #'local-time:timestamp>
                                 :key #'last-access))))
```

But that means I have to write all of this each time I need `buffer-list :sort-by-time t :domain domain`, which sucks. And, as mentioned above, I have doubts that writing these `lambda`s as functions is a good idea. 


Show me your solutions.

Ambrevar:
As I mentioned, we can define these lambdas as top-level functions.

Your concern was "where to store them".  Anywhere it fits, buffer.lisp
is fine, we can even store them in a dedicated file full of these
transformers.

About the naming: true, "filter" is misleading, what about "transform" then?

aadcg:
> As I mentioned, we can define these lambdas as top-level functions.

I have mentioned that defining a function that expects as argument `(alex:hash-table-values (buffers *browser*))` seems too much of a specific case to deserve such a treatment.

I took your suggestion and used "transform". Tho, honestly, I wouldn't even use a `key`. I would put it as an optional argument.

Ambrevar:
> I have mentioned that defining a function that expects as argument `(alex:hash-table-values (buffers *browser*))` seems too much of a specific case to deserve such a treatment.

Can you give examples?

aadcg:
> Can you give examples?

```lisp
(remove-if-not (match-domain domain)
               (mapcar #'url (buffer-list)))
```

This could be called with smth along the lines of `(buffer-list :transform #'match-domain-buffer-list)`.  But that seems silly, since you'd be defining a function that expects `(alex:hash-table-values (buffers *browser*))` as input. That's too much of a specific case to define as a top-level function.

Ambrevar:
I'm confused by your last paragraph.

What could be called?  I don't see the relation between the `remove-if-not` snippet and `(buffer-list :transform #'match-domain-buffer-list)`.  Can you provide the complete code you have in mind?

Which function would expect `(alex:hash-table-values (buffers *browser*))`?

aadcg:
> Which function would expect (alex:hash-table-values (buffers *browser*))?

The functions that you pass in the `:transform` key. They take `(alex:hash-table-values (buffers *browser*))` as argument right?

I think that if you'll look at the complete pull request you'll understand me.  Notice that I've defined `sort-by-time`, but didn't define something along those lines to match the domain. I think you'll understand why.

Ambrevar:
Every time you said `(alex:hash-table-values (buffers *browser*))`, you meant they take a list of buffers, right?
If so, what's the problem?
The code seems completely fine to me.

The definition of `sort-by-time` seems good to me, and I don't understand why you didn't define a domain matcher.  Can you explain, I'm still confused.

Can you highlight what bothers you explicitly?

aadcg:
> Every time you said (alex:hash-table-values (buffers *browser*)), you
> meant they take a list of buffers, right?

Right.

> If so, what's the problem?
> The code seems completely fine to me.
>
> The definition of sort-by-time seems good to me, and I don't understand
> why you didn't define a domain matcher. Can you explain, I'm still
> confused.

Because `sort-by-time` can be used generically with any sequence.  It is
true that when the key is `last-access`, then the sequence must be of
objects that have a `last-access` slot.  But anyway, `sort-by-time` is
useful in contexts other than `buffer-list` (as you can see in this
merge request).

But that wouldn't be the case for a domain matcher!  For the required
sequences are very specific ones, i.e. sequences of objects that must
have a URL slot.  So writing a function along the lines of:

```
(defun domain-matcher (x domain)
  (remove-if-not (match-domain domain)
                 (mapcar #'url x)))
```

Is utterly misleading since `x` is nothing but `(alex:hash-table-values
(buffers *browser*))`.

It could be written in the following way, tho I also find it misleading
since we would be defining a top-level function for a VERY specific
case.  So, to my mind, the only proper way to do this (minus the way
I've done it with lambdas) is to be able to define functions in the
"lexical scope" of buffer-list.  That is, is it possible to define
functions that could only be used (that is passed as argument) to
`buffer-list`?  I think that what I need is smth resembling a closure
but not quite, since I don't want to call a certain function within
`buffer-list`.  I want to say that a function can only be called when
passed as argument to `buffer-list`!

```
(defun domain-matcher (buffer-list domain)
  (remove-if-not (match-domain domain)
                 (mapcar #'url buffer-list)))
```

I hope this makes sense.

Ambrevar:
The answer is that the matcher must return the lambda.
See the `match-domain` function that's already in the source.

So you'd write something along those lines:

```lisp
(defun match-domain (domain &rest other-domains)
  "Return a predicate for URLs matching one of DOMAIN or OTHER-DOMAINS."
  #'(lambda (buffer-or-url)
      (some (alex:curry #'string= (quri:uri-domain (url buffer-or-url))
            (cons domain other-domains))))
```

Then you can just call it this way:

```lisp
(buffer-list :transform (match-domain "foo.xyz"))
```

Does that help?

aadcg:

> The answer is that the matcher must return the lambda.
> See the match-domain function that's already in the source.

I'm making use of match-domain.

> So you'd write something along those lines:
>
> (defun match-domain (domain &rest other-domains)
>   "Return a predicate for URLs matching one of DOMAIN or OTHER-DOMAINS."
>   #'(lambda (buffer-or-url)
>       (some (alex:curry #'string= (quri:uri-domain (url buffer-or-url))
>             (cons domain other-domains))))
>
> Then you can just call it this way:
>
> (buffer-list :transform (match-domain "foo.xyz"))

This is not what I want.  The functions passed to buffer-list take
buffer-list, i.e. (alex:hash-table-values (buffers *browser*)), as
argument.

But I get the idea, so I could do:

```lisp
(defun domain-match (domain)
  #'(lambda (buffer-list) (remove-if-not (match-domain domain)
                                         (mapcar #'url buffer-list))))
```

and then call (buffer-list :transform (domain-match "foo.xyz")).

Seems ok to me.

Ambrevar:
It seems we are on the same page.

In my example, I conflated match-domain to work both on URLs and buffers.
Indeed, the `url` method works on both types.

aadcg:
Take a look @Ambrevar. Thanks.

source/buffer.lisp
```diff
@@ -667,15 +667,25 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
...
+(defun buffer-list (&optional (transform #'identity))
+  "Return the list of buffers of `*browser*'.
+TRANSFORM is a list of functions, or a single function, that take a list of
+buffers as argument.  These functions are composed from right-to-left."
+  (funcall (apply #'alex:compose (alex:ensure-list transform))
+           (alex:hash-table-values (buffers *browser*))))
+
+(defun domain-match (domain)
+  #'(lambda (buffer-list) (remove-if-not (match-domain domain)
+                                    (mapcar #'url buffer-list))))
```
Ambrevar, line 679: Having both `domain-match` and `match-domain` is confusing.
I suggest just having this in place of the existing `match-domain`:

```lisp
(defun match-domain (domain)
   #'(lambda (buffers-or-urls) (remove-if-not (match-domain domain)
                                     (mapcar #'url buffers-or-urls))))
```

source/buffer.lisp
```diff
@@ -740,8 +750,7 @@ proceeding."
 
 (defun last-active-buffer ()
   "Return buffer with most recent `last-access'."
-  (first (sort (buffer-list)
-               #'local-time:timestamp> :key #'last-access)))
+  (first (buffer-list #'sort-by-time)))
```
Ambrevar, line 753: As was suggested, this is equivalent to `(sort-by-time (buffer-list))`, so I suggest we keep the latter which is more unversal, and we drop the `transform` argument since you deprecated it.

source/buffer.lisp
```diff
@@ -763,7 +772,7 @@ proceeding."
   (echo "~a copied to clipboard." (title (current-buffer))))
 
 (defun buffer-initial-suggestions (&key current-is-last-p domain)
-  (let ((buffers (buffer-list :sort-by-time t :domain domain)))
+  (let ((buffers (buffer-list (list (domain-match domain) #'sort-by-time))))
```
Ambrevar, line 775: Then this would be `(funcall (match-domain domain) (sort-by-time (buffer-list))`.

Ambrevar: And similar for the other calls.

source/buffer.lisp
```diff
@@ -862,7 +871,7 @@ set of useful URLs or preparing a list to send to a someone else."
 
 (define-command delete-all-buffers (&key confirmation-p)
   "Delete all buffers, with confirmation."
-  (let ((count (length (buffer-list))))
+  (let ((count (buffer-list #'length)))
```
Ambrevar, line 874: I prefer the former `(length (buffer-list))`, it's more idiomatic CL.

source/status.lisp
```diff
@@ -42,15 +42,16 @@
...
-                    (mapcar #'(lambda (i) (quri:uri-domain (url i)))
-                            (buffer-list :sort-by-time t)))
-            :test #'equal)))
+  (let ((buffer-domains
+          (buffer-list (list (lambda (buffer-list)
+                               (remove-duplicates
+                                (sera:filter-map #'quri:uri-domain
+                                                 (mapcar #'url buffer-list))
+                                :test #'equal))
+                             #'sort-by-time))))
```
Ambrevar, line 51: We can call `match-domain` here, no?

Ambrevar:
OK, I like where this is going.  If this is alright with everyone, I suggest dropping the `transform` argument and just use regular function calls, it's just as short and regular CL anyways.

source/buffer.lisp
aadcg: > Having both domain-match and match-domain is confusing.

This is a fair concern.

> I suggest just having this in place of the existing match-domain

These are two different things anyway. I agree `match-domain` and friend could handle both a URL or a buffer, but one thing is the lambda that returns the predicate that does the matching, and other thing is passing a list and getting the list back with the matches of the filtering.

source/buffer.lisp
aadcg: Fair enough.

Ambrevar:
> These are two different things anyway. I agree `match-domain` and friends could handle both a URL or a buffer, but one thing is the lambda that returns the predicate that does the matching, and other thing is passing a list and getting the list back with the matches of the filtering.

Sorry, you are right, replace `funcall` by `remove-if` or `sera:filter` in my
suggestions, thus

```lisp
(sera:filter (match-domain "foo") (buffer-list))
```

is good enough, no?

source/buffer.lisp
aadcg: But this completely defeats the way I defined `buffer-list`. This would only make sense if `buffer-list` would be nothing else but `(alex:hash-table-values (buffers *browser*))`.

How can `(funcall (match-domain domain) (sort-by-time (buffer-list))` be more sensible than `(buffer-list (list (domain-match domain) #'sort-by-time))`?  

The second syntax is plain simple. It says: take `buffer-list` as argument and pass it to the functions from the list. These functions are composed from right-to-left.

source/buffer.lisp
aadcg: Also fair.

source/status.lisp
aadcg: No. Here you want a list with all the unique domains from `buffer-list`.

aadcg:
> I suggest dropping the transform argument and just use regular function calls, it's just as short and regular CL anyways.

I disagree. All we're discussing here are the notations below to denote composition of functions `f`, `g` and `h`. 

The left hand side notation is far superior than the one on the right. In the Lisp world, I'm just putting these functions in a list, choosing my argument (`x`) and happily composing them from right-to-left.

Ambrevar:

> How can `(funcall (match-domain domain) (sort-by-time (buffer-list))` be more sensible than `(buffer-list (list (domain-match domain) #'sort-by-time))`?  
>
> The second syntax is plain simple. It says: take `buffer-list` as argument and pass it to the functions from the list. These functions are composed from right-to-left.

There are two ways to do this:

- either have the transforms happen inside buffer-list;

- or have no transforms in buffer-list at all, and just do them on the
call site.

The latter has the benefit that the reader who does not know the details
of buffer-list would know immediately what going on (and does not need
to read the docstring of buffer-list to know in which order the
transforms are processed).
This approach is also less code.

This is why I'm in favour of having `buffer-list` that just returns 
`(alex:hash-table-values (buffers *browser*))`.

Can you see more
benefits to having transforms inside buffer-list?

aadcg:
> This approach is also less code.

I have doubts about this one.  But I understood the other reasons.

> 
This is why I'm in favour of having `buffer-list` that just returns
`(alex:hash-table-values (buffers *browser*))`.

Ok, I will refactor it using this approach.

aadcg:
Is master broken? I get 

```
The value
  #<USER-INTERNAL-BUFFER {100A944643}>
is not of type
  QURI.URI:URI
```

when I start it.

Ambrevar:
Can you open a new issue with the full stacktrace?
(When you run it from a REPL.)

aadcg:
> Can you open a new issue with the full stacktrace?
(When you run it from a REPL.)

I can't reproduce it anymore. Might have been some mistake on my side.

aadcg:
I think things now stand with what you had in mind @Ambrevar. 

Thoughts? @jmercouris @aartaka

aartaka:
Looks all good to me!

jmercouris:
I have nothing to add. Looks OK to me!

Ambrevar:
Thanks!
Looks good to me.  I've just tested it, and I've spotted two issues.
I've fixed them on the https://github.com/atlas-engineer/nyxt/tree/buffer-list-filtering-fixes branch.

What do you think?

jmercouris:
Why protect against NIL? should that be the responsibility of these functions?

Ambrevar:
Since match-domain and friends return a predicate, and predicates are
likely to be called against anything.

jmercouris:
OK. Seems consistent with the rest of CL.

Ambrevar:
@aadcg OK to merge the fixed branch?

aadcg:
@ambrevar, OK to merge, thanks. What's the advantage of `sera:filter` over `apply`?

Ambrevar:
`apply` does not do what you want here.

- sera:filter == remove-if-not.
- apply: calls the function over the arguments found in the list in last
  position.

In this case, `(apply (match-domain domain) URLS)` would fail because
the predicate returned by match-domain expects only one argument.

Ambrevar:
`apply` can return a list, it actually returns the return value of the function in first position.

Ambrevar:
`remove-if-not` is deprecated by CLHS.  `sera:filter` is not :)

Ambrevar:
Merged with 645e5ab0a7a41030fd832f497810cd145b911580.  Thank you André for keeping through this :)

aadcg:

> Merged with 645e5ab. Thank you André for keeping through this :)

Thank you.

I finally understood why I did this silly thing of using apply.  It was
because match-domain used to me smth else and then I didn't take action!

PR #1338, opened 2021-04-29
renderer-script.lisp: Add fetch-links.

aadcg:
Another hunk from #1246.

### DIFF
--- a/source/renderer-script.lisp
+++ b/source/renderer-script.lisp
@@ -84,6 +84,23 @@ The function can be passed ARGS."
                       (|insertAdjacentHTML| "afterbegin"
                                             (ps:lisp style)))))))
 
+(defun collect-links (&key (buffer (current-buffer))
+                        (filtering-rules (list #'host=
+                                               (complement #'path=)
+                                               #'scheme=)))
+  "Return a list of links from BUFFER.
+FILTERING-RULES is a list of functions that take two URLs and return a boolean.
+A link is collected when, for all elements of FILTERING-RULES, the return value
+is non-nil."
+  (with-current-buffer buffer
+    (pflet ((nlinks () (ps:chain document links length))
+            (fetch-link (i) (ps:chain document links (item (ps:lisp i)) href)))
+      (loop for i below (parse-integer (nlinks))
+            with url = (url buffer)              ; computed once
+            for link = (quri:uri (fetch-link i)) ; computed on every iteration
+            when (url-eqs url link filtering-rules)
+              collect link))))
+
 (export-always 'with-current-html-buffer)
 (defmacro with-current-html-buffer ((buffer-var title mode
                                      &key no-history-p)


source/renderer-script.lisp
```diff
@@ -84,6 +84,23 @@ The function can be passed ARGS."
...
+                                             (lambda (x y) (not (path= x y)))
+                                             #'scheme=)))
+  "Return a list of links from BUFFER.
+FILTERING-RULES is a list of functions that take two URLs and return a boolean.
+A link is collected when, for all elements of FILTERING-RULES, the return value
+is non-nil."
+  (with-current-buffer buffer
+    (pflet ((nlinks () (ps:chain document links length))
+            (fetch-link (i) (ps:chain document links (item (ps:lisp i)) href)))
+      (loop for i from 0 to (1- (parse-integer (nlinks)))
```
Ambrevar, line 98: Replace `to (1- ...` with `below`.

source/renderer-script.lisp
```diff
@@ -84,6 +84,23 @@ The function can be passed ARGS."
...
                                             (ps:lisp style)))))))
 
+(defun fetch-links (&key (buffer (current-buffer))
+                      (filtering-rules (list #'host=
+                                             (lambda (x y) (not (path= x y)))
+                                             #'scheme=)))
+  "Return a list of links from BUFFER.
+FILTERING-RULES is a list of functions that take two URLs and return a boolean.
+A link is collected when, for all elements of FILTERING-RULES, the return value
+is non-nil."
```
Ambrevar, line 94: Maybe name it `collect-links`?

source/renderer-script.lisp
```diff
@@ -84,6 +84,23 @@ The function can be passed ARGS."
                       (|insertAdjacentHTML| "afterbegin"
                                             (ps:lisp style)))))))
 
+(defun fetch-links (&key (buffer (current-buffer))
+                      (filtering-rules (list #'host=
+                                             (lambda (x y) (not (path= x y)))
```
Ambrevar, line 89: Use `(complement #'path=)` instead.

Ambrevar:
OK to merge this, but in general I'd prefer to only add helper functions as part of a patch set where there are used.
What do you think?  OK to keep it as part of #1246 or do you think it's easier to merge it now?

aadcg:
Actually #1246 doesn't directly use this.  When I started working on #1246 I was using the renderer to collect the links but then moved to another strategy.  I see this change as an orthogonal change, but might be useful for other applications.

Also, @jmercouris mentioned that #1246 could benefit from being smaller.

Ambrevar:
OK then I suggest to keep this here as a pull request for a little
while, see if we ever need it.

Thoughts?

aadcg:
Seems fair to me @Ambrevar.

source/renderer-script.lisp
aartaka: Adding to that: `from 0` is optional, AFAIK.

aadcg:

> @aartaka commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In source/renderer-script.lisp:
>
>> @@ -84,6 +84,23 @@ The function can be passed ARGS."
>                        (|insertAdjacentHTML| "afterbegin"
>                                              (ps:lisp style)))))))
>  
> +(defun fetch-links (&key (buffer (current-buffer))
> +                      (filtering-rules (list #'host=
> +                                             (lambda (x y) (not (path= x y)))
> +                                             #'scheme=)))
> +  "Return a list of links from BUFFER.
> +FILTERING-RULES is a list of functions that take two URLs and return a boolean.
> +A link is collected when, for all elements of FILTERING-RULES, the return value
> +is non-nil."
> +  (with-current-buffer buffer
> +    (pflet ((nlinks () (ps:chain document links length))
> +            (fetch-link (i) (ps:chain document links (item (ps:lisp i)) href)))
> +      (loop for i from 0 to (1- (parse-integer (nlinks)))
>
> Adding to that: from 0 is optional, AFAIK.

Indeed, I got rid of that too.

jmercouris:
This information is also stored on the `recursive-search` branch.

Ambrevar:
OK, but we haven't cleaned up the branches in a long time so it will be
hard to find it in this mess :p

Time for a summer clean up?

jmercouris:
It's time for a paaaaartayyyy! I'll clean up all of my branches I'm no longer using, I guess if we all do this we will have a squeaky clean repository!

Ambrevar:
Thanks, will do as well!

Ambrevar:
This should give a clear outlook of which branch is "yours":

https://github.com/atlas-engineer/nyxt/branches/yours

Then you can delete the branches from your shell as follows:

git push origin --delete BRANCHES...

@aartaka Can you also do some clean up?

aartaka:
Will do!

Issue #1109, opened 2021-01-26
Internal buffers should have a history


Ambrevar:
It would allow for documentation navigation for instance.

PR #1343, opened 2021-04-30
Draft of the skeleton of Nyxt's manual.

aadcg:
This a first step towards the Nyxt's manual.

Manuals are always important, be it a microwave manual or a software
manual.  In particular, programs that empower their users must handle
documentation with special care.

Manuals help users fit the program in their brain.  They don't provide a
faithful technical description -- there's Nyxt's help system, or the
source code for that.

Some of the information already exists, it's scattered in many places
(README, Dev README, patreon, website, articles, etc).  For that reason,
the style and tone of this material varies.  It needs to be compiled
into a meaningful unified banner.  Maintaining all of those place up to
date is costly.  Today, there are already some discrepancies.  Since the
manual is HTML, it should be easy to replicate that information in other
places (such as the website).

Nyxt's manual will primarily be read within Nyxt itself.  Whether there
will ever any printed copies is a subject that I find pointless to
discuss now.  Therefore, we can be a bit bold and make it interactive.
Notice that we can run any piece of lisp by passing a link with the lisp
scheme.

Nyxt's "fundamental unit of data" is HTML, for it uses a web renderer at
all(?) times.  But we never write HTML ourselves.  We use the markup
library to generate HTML.  Or we use the org format that ultimately gets
translated into HTML (on GitHub, or Nyxt's website).

I'm bringing an org file because it was easier for me to organise my
ideas.  I feel that writing a manual using the markup library might be
too painful.  Org exports to html, or a new exporter to that markup
library syntax could be written but that introduces a dependency on org
mode, which is bad.  Maybe there's no alternative, but to write it in
the lispy markup.

### DIFF
--- a/documents/MANUAL2.org
+++ b/documents/MANUAL2.org
@@ -1,0 +1,182 @@
+#+TITLE: Nyxt User Manual
+
+* Introduction
+Nyxt is designed to meet the needs of those who wish to navigate the
+vast source of information of the Internet, unrestrained.
+
+Nyxt is an extensible web browser.  You can expect all the features of
+other web browsers you may have used; some unique more advanced
+features; and the ability to extend it as you wish.  While some
+programming expertise is needed to achieve the last point, all
+non-programmers can benefit from their work.  Nyxt sets the foundations
+so that programmers can write extensions with ease.
+
+* Installation
+
+** On MacOS
+# TODO
+
+** On GNU/Linux
+# https://nyxt.atlas.engineer/download
+Get it from your distribution package manager
+
+Nyxt is currently packaged for:
+
+- Alpine
+- Debian and derivatives (Ubuntu, LinuxMint)
+- [[https://source.atlas.engineer/view/repository/macports-port][MacPorts]]
+- [[https://aur.archlinux.org/packages/nyxt][Arch Linux AUR]] (and the [[https://aur.archlinux.org/packages/nyxt-browser-git/][-git PKGBUILD]])
+- [[https://nixos.org/nix/][Nix]]: Install with =nix-env --install nyxt=.
+- [[https://guix.gnu.org][Guix]]: Install with =guix install nyxt=.
+
+If your distribution provides a Nyxt package, reach out to us and we
+will include it in this list. If the package is outdated, please ping
+the package maintainers for an update.
+
+If your distribution does not package Nyxt, read on.  Ubuntu 18.04
+package (x86-64)
+
+It should work on other distributions based on this version of Ubuntu.It
+will probably not work on other Debian-based distributions.
+
+All distributions (x86-64)
+
+SHA-256 checksum:
+
+14d68e8621d049d8a2ba640ddfccad208e60ae507a2a0e4a98f9724fcd0524cb
+
+This is a complete bundle that includes all the recursive dependencies and requires nothing but a recent-enough Linux kernel, hence its rather big size! As soon as you download Nyxt for GNU/Linux,
+
+    extract the archive (e.g. with tar xf /path/to/download/file)
+    execute ./usr/local/bin/nyxt to get started!
+
+The archive does not have to be extracted to the filesystem root
+(/). This pack relies either on the "user namespace" feature of the
+Linux kernel or on PRoot to set up a virtual root folder for Nyxt in the
+root of the extracted archive.
+
+If Nyxt does not work for you at this point, you might have to enable
+user namespaces. Try to run the following in a shell:
+
+sudo sysctl -w kernel.unprivileged_userns_clone=1
+
+    If the previous command did not work
+    You might have to issue the following instead:
+
+    sudo sh -c 'echo 0 > /proc/sys/kernel/userns_restrict'
+
+    or, depending on your kernel compilation options:
+
+    sudo sh -c 'echo 1 > /proc/sys/kernel/unprivileged_userns_clone'
+
+    If none of the above works, your kernel might not support user namespaces. You might have to compile a custom kernel to enable this feature, but at this point it's probably easier to install Guix and run
+
+    guix install nyxt
+
+    or even build Nyxt from source. Refer to your distribution for more help.
+
+You can produce this Guix pack yourself with
+
+guix pack -RR --no-grafts --compression=xz --symlink=/usr/local/bin/nyxt=bin/nyxt nyxt
+
+which should yield:
+
+/gnu/store/jycbbqgziaj8sqh7gwh15ifjd1hkpjcx-tarball-pack.tar.xz
+
+on the Guix commit:
+
+a2f3a456bec357394fe550d927c26bbbdaa7ba6e
+
+You can find the documentation for Nyxt within Nyxt. If you need
+support, please find us on #nyxt on the Freenode IRC network, on our
+forums, or open up an issue on our GitHub:
+https://github.com/atlas-engineer/nyxt.
+
+** From Source
+[[file:~/common-lisp/nyxt/documents/README.org::+TITLE: Nyxt Developer Readme][info here]]
+
+The latest source is available from https://github.com/atlas-engineer/nyxt.
+
+You can find the documentation for Nyxt within Nyxt. If you need
+support, please find us on #nyxt on the Freenode IRC network, on our
+forums, or open up an issue on our GitHub:
+https://github.com/atlas-engineer/nyxt.
+
+* Getting started
+Describe the screen an basic concepts.  Make it interactive.
+
+** Buffers
+
+** Prompt buffer
+
+** Keys
+
+** Commands
+
+** Modes
+
+** Message Area
+
+** Status Area
+
+** Window
+
+* Getting Around
+Describe how to move around in buffers, and how to navigate the web.
+
+* Basic Features
+
+** Navigation
+
+** Bookmarks
+
+** Search
+
+** Hinting System
+
+** Spell Checker
+
+** Noimage/sound/sript
+
+** Download Manager
+
+* Advanced Features
+
+** Repl
+
+** Auto mode
+
+** Visual Mode
+
+** OS package manager mode
+
+* Help
+
+** Getting Help
+
+** Nyxt Help System
+
+* Customisation
+
+** UI
+
+** Init File
+
+** Hooks
+[[https://nyxt.atlas.engineer/article/hooks.org][here]]
+
+** Theme
+
+* Contributing
+[[https://nyxt.atlas.engineer/article/technical-design.org][here]]
+
+* Acknowledgements
+
+* Licence
+
+* COMMENT Sources of information
+- [[https://www.patreon.com/nyxt][Patreon]]
+- [[https://nyxt.atlas.engineer/][Nyxt's website]]
+- [[https://nyxt.atlas.engineer/articles][Articles]]
+- [[file:~/common-lisp/nyxt/README.org::*Nyxt browser][Nyxt's README]]
+- [[file:~/common-lisp/nyxt/documents/README.org::+TITLE: Nyxt Developer Readme][Nyxt's Developer README]]


Ambrevar:
Thanks for starting this discussion.

First question: why not stick to the current manual?
Note that our .asd and Makefile can export it to a stand alone HTML file.

jmercouris:
Pierre,

I believe that Andre would like to write a new manual and did not want to just delete the existing manual file. Not sure the reasoning behind that.

jmercouris:
To clarify my previous comment: Andre and I have spoken and the next thing he would like to work on is improving the manual.

aadcg:
> First question: why not stick to the current manual?
Note that our .asd and Makefile can export it to a stand alone HTML file.

I didn't want to interfere with anything that was in place -- `manual.org` contains valuable information. I only noticed now that there used to be a manual. I will have to analyze it.

Ambrevar:
> I only noticed now that there used to be a manual. I will have to analyze it.

Oops!  If you didn't notice the manual before, it probably means we
didn't highlight it well enough.

Would you have any suggestions to make its existence and access more obvious?

aadcg:
> Oops! If you didn't notice the manual before, it probably means we
> didn't highlight it well enough.

It seems to me that everything is as it should be.  My lack of attention
shouldn't be taken as a rule of thumb.  You used to have a manual
(i.e. an org file), but then you deprecated it and provided a manual
accessible from M-x manual.

> Would you have any suggestions to make its existence and access more
> obvious?

There's a reference in the source code.  I don't think you want to have
it referenced in the project's README, right?

Ambrevar:
> There's a reference in the source code.

Which reference?

> I don't think you want to have
> it referenced in the project's README, right?

The project's README is mostly seen by developers / hackers.
But it does not cost much to add it.

@jmercouris Thoughts?

aadcg:
> Which reference?

source/documents/MANUAL.org:

- For Nyxt 1.5 (then called Next), see https://github.com/atlas-engineer/nyxt/blob/1.5.0/documents/MANUAL.org.

jmercouris:
I think things are as they should be as well. Let us not be so reactive :-)

aadcg:
I'm more interested in the following: what motivates the decision?  It seems to me that `M-x manual` doesn't replace `MANUAL.org`.

I suspect that it might be due to "the documentation are the docstrings" but that's a mistake to my mind. Docstrings help with specific technicalities. But users read manuals when they're looking for general accounts of how something works.  

That's the difference between having a person teach you a language and reading the language's grammar. People (or manuals) are good at giving inexact but general mental pictures.  Docstrings (or technical specifications) are good at defining things unequivocally.

Ambrevar:
> But users read manuals when they're looking for general accounts of
> how something works.

This is what the embedded `manual' aims to do.
Do you think it's not done well?  If so, what would you suggest?

edgar-vincent:
I understand why `MANUAL.org` was moved to `M-x manual`, but I still think that some form of documentation should be available online (much like Emacs documentation which is available both through `M-x info` and online). One might need to be able to read it when Nyxt is not running, for instance.

I understand that URLs may change, but is that really a problem? One can assume that the reference manual is the one included in Nyxt, and an online (and possibly versioned?) version could be generated from the source.
It is also unclear why some parts of the existing manual tell the user to refer to a specific mode/slot's docstring. I think this could be done in the case where additional information can be retrieved form the docstring, but I believe the manual should at least give general information (eg. `foo-mode does this and that. You can configure it like this and that. For more information, refer to its docstring.`).

On another matter, there might be a need, IMHO, of a stricter separation between the tutorial and the manual. I often find myself not knowing which document I should be referring to.

I'd like to learn enough basic CL to write (or help someone write by giving my layman's feedback) a general guide about CL concepts in the context of Nyxt (what's a class? a slot? how do I write my own configuration?), but I am no programmer and it will take me a long time to do it.

I imagine the following set of documentation, all available online and within Nyxt. 
* a manual, for an overview of what Nyxt can do (like the current manual). The manual should list *all* user-facing modes and their basic usage. I don't think one should have to look at the articles to learn about new modes, or manually browse though available functions with `describe-function`. It should all be available centrally.
* a tutorial, for a set of quick, basic configuration choices (like the current tutorial) to get oneself running when first introduced to Nyxt
* an general introduction to "Nyxt Common Lisp" (:-p), i.e. CL as applied to Nyxt (similarly to https://www.gnu.org/software/emacs/manual/eintr.html for Emacs)
* a "cookbook", i.e a compilation of snippets which *work* in the current version of Nyxt (like the website's articles, but kept up-to-date). I don't think one should have to browse through forum posts or old articles to know how to load an extension, for example.

aadcg:
@edgar-vincent, thank you for sharing your thoughts. I read it with enthusiasm, and it seems we're on the same page.

> One might need to be able to read it when Nyxt is not running, for instance.

This is paramount! And this why I think that interactive documentation is interesting but then it lacks this key aspect. I think a good interactive tutorial could be made (that builds on the existing one) and a manual should also be provided.  I see you're also a fan of the `info` format as much as I do.  So indeed, it would be nice to have a manual available in say HTML and `info`.  Of course that, if the manual is written in org, in can be exported to those format or to whatever one might need.

> I understand that URLs may change, but is that really a problem? One can assume that the reference manual is the one included in Nyxt, and an online (and possibly versioned?) version could be generated from the source.

Exactly. The online documentation of Nyxt could be available on Nyxt's website. And indeed, the manual is an output of any Nyxt version.  That is, at any given time, it should be possible to "compile" the manual from source and that would become "Nyxt Manual version x.x.x".

> It is also unclear why some parts of the existing manual tell the user to refer to a specific mode/slot's docstring.

Indeed, and I have been advocating in this direction too. For the developers it's easy to say: "it's all in there, the code is the documentation". In a previous post above I've distinguished between docstrings and manuals.  One doesn't replace the other. I have a more radical stance than you and I'd argue that a manual that refers to docstrings doesn't do its job well. Docstrings and manuals live in different worlds. And they should be apart from each other. 

> On another matter, there might be a need, IMHO, of a stricter separation between the tutorial and the manual. I often find myself not knowing which document I should be referring to.

Exactly.  The reality is that a manual doesn't really exist. And even if we take what exists as a tutorial, it can improved. 

> I'd like to learn enough basic CL to write (or help someone write by giving my layman's feedback) a general guide about CL concepts in the context of Nyxt (what's a class? a slot? how do I write my own configuration?), but I am no programmer and it will take me a long time to do it.

Those concepts must be explained in the manual.  But they have to be defined from the perspective of the Nyxt user, not from the perspective of a person who is learning CL. Btw, regardless of you being a programmer or not, I think it would take any person some time to wrap their heads around Nyxt.

> I don't think one should have to look at the articles to learn about new modes

Exactly. I've been advocating in that direction too.

> or manually browse though available functions with describe-function. It should all be available centrally.

True. `describe-funtion` is for people that want to study the internals of Nyxt. But not everyone will want/need that.

> a tutorial, for a set of quick, basic configuration choices (like the current tutorial) to get oneself running when first introduced to Nyxt

I'd say that this is in place.  Though, as mentioned, the tutorial can be greatly improved. Nyxt can also be configured graphically (though it could still be argued that it can be daunting). 

> an general introduction to "Nyxt Common Lisp" (:-p), i.e. CL as applied to Nyxt (similarly to https://www.gnu.org/software/emacs/manual/eintr.html for Emacs)

Couldn't agree more.  That's not only needed but actually unavoidable! It could be argued that all programs need DSLs (Domain-Specfic Languages). Lisp, due to its nature, is a very good choice to achieve that that. I.e. Nyxt defines lots of macros (sort of defines its own programming language) in order to reason about the problems it aims to solve. And a manual for that is needed. Indeed, learning CL is important to understand Nyxt. But it doesn't suffice! 

> a "cookbook", i.e a compilation of snippets which work in the current version of Nyxt (like the website's articles, but kept up-to-date). I don't think one should have to browse through forum posts or old articles to know how to load an extension, for example.

Here I'm afraid we disagree.  I have to admit that I'm against the very nature of "cookbooks".  I never learnt anything meaningful from them so, in retrospective, it was always a waste of time. Tho, one does get the feeling that "oh, I got a recipe and I solved my problem". Anyway, I think this is useless when things as documented properly. And "properly" alludes to the "centralized documentation sources" you alluded to! Which contrasts to the way scattered way we do things today (articles, etc). Notice that users can get plenty of snippets from sources that aren't directly Nyxt.  You can look at the config files of "advanced" users, there's a Reddit channel, a discourse forum, you can post issues on Github. To my view, it's more than enough.

@edgar-vincent, once again, thanks for taking the time to write your thoughts and experiences. Your views are very valuable!

edgar-vincent:
My phrasing was not good enough. What I meant by cookbook was something similar to [mu4e's documentation](https://www.djcbsoftware.nl/code/mu/mu4e/), which contains useful snippets **in context** (i.e a feature is described, and common ways in which it can be extended or configured are listed after that) which are relevant to most users' needs.

What I think should be addressed is the clarification of the notion of (**α**) extensibility: (**β**) Nyxt can be extended by writing CL *or* (**γ**) by using code others have written. 
Choosing **γ** does not make **α** untrue. I don't think users should know any Lisp (or only very, very little) to be able to extend Nyxt. But I don't think they should be browsing through pages and pages of forums threads or git repositories to find what they need, at least for the common use cases.

Incidentally, I think that copy-pasting snippets is a great way to get a very basic introduction to Lisp. It's what I did with Emacs: I don't have the skills to write proper packages or the time to learn Elisp thoroughly, so I've been copy-pasting snippets without having the faintest idea how they worked and *then* learning basic Elisp. This doesn't allow me to write any serious programs, but it does allow me to adapt Emacs to my needs.

aadcg:
edgar-vincent ***@***.***> writes:

> What I think should be addressed is the clarification of the notion of
> (α) extensibility: (β) Nyxt can be extended by writing CL or (γ) by
> using code others have written.
> Choosing γ does not make α untrue. I don't think users should know any
> Lisp (or only very, very little) to be able to extend Nyxt. But I don't
> think they should be browsing through pages and pages of forums threads
> or git repositories to find what they need, at least for the common use
> cases.

Couldn't agree more.

> Incidentally, I think that copy-pasting snippets is a great way to get a
> very basic introduction to Lisp. It's what I did with Emacs: I don't
> have the skills to write proper packages or the time to learn Elisp
> thoroughly, so I've been copy-pasting snippets without having the
> faintest idea how they worked and then learning basic Elisp. This
> doesn't allow me to write any serious programs, but it does allow me to
> adapt Emacs to my needs.

That's totally fair.  The questions I'd like to ask you now are the
following:

Where are you finding those snippets?  Do you get them exclusively from
Emacs sources (its manual, etc)?  In other words, do you find that the
snippets provided by Emacs are enough?  For an example, look at (info
"(emacs) Init Examples").

edgar-vincent:
I don't really use snippets anymore, because I have a good enough knowledge of Elisp to write my own simple functions and bugfixes. When I did, I mostly used snippets found on Reddit or individual packages' documentation. However, this takes a lot of time, and I'd rather have used a central source, because my needs were pretty basic (I often get my own snippets/bigfixes upstreamed for that very reason).

I don't think the snippets provided by the Emacs manual are good enough, because I find them too generic or too trivial. I have the impression that they illustrate general concepts more than what lay people may actually need (that's fine, but it's not what I'm looking for). I think the target should be power users, not programmers. Customising Emacs takes too much time - I don't want to be doing that, hence my use of Doom Emacs. I have the feeling that most of what I need, as a regular user, is what most users need, which is why I love Doom Emacs (which, in spirit, is pretty much exactly what I was trying to achieve when I was writing my own configuation from scratch, only much better and *much* less time-consuming).

I understand that Doom Emacs can be perceived as too opinionated, in which case I think providing good, useful snippets is a good middle ground.

edgar-vincent:
> I'm afraid I strongly disagree with you. As a user, you have the
> freedom to use the tools as you wish. The question here is: is it
> Nyxt's burden to provide the snippets/the experience you mention? IMO,
> it is not.

Not necessarily, no. But I think the very basic uses should be covered. There are many beginners (including myself) who turn up on IRC to ask very trivial questions - I don't think this should be the case. For instance, I've never been able to use the RSS reader, the `ace-mode` editor or `nx-freestance-handler` (please note that I am not at all pointing fingers here, and I'm a *huge* admirer of the team's work - and some of the extensions I'm referring to weren't even written by the main team). I do also know that this is probably due to the fact that they are WIP.

> I could go on and explain why. But @protesilaos has been writing
> and thinking a lot about Emacs, and what it means to be an Emacs user.
> When I came across his work, it felt as if he has been reading my mind
> so, I apologise, but I'd rather point to his work:

I read his post when he published and enjoyed it very much. I even emailed him to discuss it further as I also had the feeling he'd been reading my mind :) I don't think it contradicts my point. The importance of collective autonomy is exactly what I'm emphasizing. I just think it should be a little more accessible :)

Anyway, this is all very abstract, I should be working on concrete suggestions and improvements :)

Ambrevar:
Allow me to clarify some points that were mentioned in this thread. :)

@aadcg wrote:

> I suspect that it might be due to "the documentation are the docstrings" but that's a mistake to my mind.

Agreed that would be a mistake and it's not the intention of the manual.

If the manual is sparse at the moment, it's because we need more time to work on
it ;)

As was hinted elsewhere, the biggest reason behind this conciseness is that
Nyxt's API has been such a moving target for the past 2 years that it was hard
to follow up on documentation.

With 2.0, things should settle down a bit and now is a great time to finally
overhaul the manual.

Contributions are welcome! :)

@edgar-vincent wrote:

> I still think that some form of documentation should be available online (much
> like Emacs documentation which is available both through `M-x info` and
> online).

We've been able to export the documentation to HTML for a while.
Now all that remains to do is some automatic way to host it on
https://nyxt.atlas.engineer.  It's not hard, we just need to get down to it.

@edgar-vincent wrote:

> It is also unclear why some parts of the existing manual tell the user to
> refer to a specific mode/slot's docstring. I think this could be done in the
> case where additional information can be retrieved form the docstring, but I
> believe the manual should at least give general information.

The rationale is that the manual should not repeat the docstrings, lest the doc
goes out-of-sync and contradicts itself.

True, it's currently not very comfortable to read the manual while querying the
docstrings to access the details.

I can think of two possible improvements:

- Convert the docstring references to links, so that the user just has to click
  on the link, then can go back to the manual (this means we need to add history
  support to internal buffers).
  
- Add the possibility to substitute docstring references with their content, so
  that they are embedded in the manual.
  
@edgar-vincent wrote:

>  On another matter, there might be a need, IMHO, of a stricter separation between
> the tutorial and the manual. I often find myself not knowing which document I
> should be referring to. 

The manual _embeds_ the tutorial.

You are not the first one being confused by this.
First line of the manual says: "This manual first includes the tutorial, then
covers the configuration of Nyxt."
Can we phrase this differently to remove all sources of confusion?

The tutorial is meant to be read only once.  After the first reading, the user
can stick to the manual, since all the tutorial content is also in there.


@edgar-vincent wrote:

> I'd like to learn enough basic CL to write (or help someone write by giving my
> layman's feedback) a general guide about CL concepts in the context of Nyxt

The manual includes a "Learn Lisp" link to
https://nyxt.atlas.engineer/learn-lisp.
Should we emphasize it more?

Teaching Lisp is outside the scope of Nyxt.

However, we could package the offline version of the Common Lisp cookbook and
use it as well as the offline CLHS package as an optional dependency for Nyxt.

Nyxt could add commands to browse the offline CLHS / Cookbook, when locally installed.

@edgar-vincent wrote:

> I don't think one should have to look at the articles to learn about new
> modes, 

Agreed, this is not the goal of articles.  Articles are meant for public
communication, not as a documentation reference.


*About the question of a "cookbook":* I agree with @aadcg that it's hard to
maintain and keep reliable.

Instead, I'd rather include good examples in the manual.

*About the `info` format*: Ideally, we would display the documentation in a
virtually-sectioned pager like Info does, e.g. one page per `<h1>` or `<h2>` header.
In particular, this would mean the `tutorial` command would just bring the user
to the "Tutorial" chapter of the manual, and they could keep browsing the manual
from there.  It would remove the ambiguity about what the tutorial is.


@edgar-vincent wrote:

> To summarise, I think that extensibility should be effective in *practice* for
> non-programmers. 

There may be even better ways than snippets, e.g. a graphical approach to
configuration.

Here I'm thinking Emacs' `customize` interface, but not just that.  We could
have a look at what Pharo and Gtoolkit do.

This will be for post 2.0 though :p

> For instance, I've never been able to use the RSS reader, the `ace-mode`
> editor or `nx-freestance-handler`

Until today (oops :p) there was a bug which prevented vanilla Nyxt from loading
extensions automatically.  This is now fixed, so it should work as expected in
the coming version.

@aadcg wrote:

> I have raised some concern, but it didn't get much attention.  I take it humbly,
> so I believe that they're in a better position to judge :)

Sorry about that.  The attention things get is not well correlated to their
importance, sadly.

I'm only too aware how important documentation is, and at the same time one has
to chose their battles! :p

Documentation rework is high on my priority list now for 2.0, but this will need
community reviews as well!

Ambrevar:
edgar-vincent writes:

> I agree with you, but there seems to be a difference between Lisp and Lisp as used within Nyxt (i.e its architecture: which objects/classes/slots might I be interested in? how do I write my own extension?). The manual may already be good enough, but I'm not sure.

Nyxt's Lisp is regular Common Lisp, really.  We have a few macros (like
define-configuration), which is expected from Common Lisp, and these
should be well documented with examples.

Classes and slots are all documented and explorable.  The interesting
objects such as buffers should be covered in the manual.  Right now it's
a bit lacking, so contributions are welcome.

This is not a concern about the _language_, but rather about the API.

>> Nyxt could add commands to browse the offline CLHS / Cookbook, when locally installed.
>> 
> That would be nice. It's more or less what Emacs does, and I think that works well.

Emacs does not rely on separate packages though, it embeds Eintr and
other manuals.

> I'd be very happy to help :)

Thank you! :)

kssytsrk:
> For instance, I've never been able to use <...> `nx-freestance-handler`

Hi, sorry if it's off-topic (but since you mentioned this here, it makes sense to talk about it here), but `nx-freestance-handler` didn't work on the newer versions of Nyxt until now. Oof. It worked flawlessly on the older versions, and since I haven't been keeping up with Nyxt's updates recently, I didn't know that the codebase changed a lot. Had to rewrite some stuff, now it works (at least for me!). If you are still interested in using it, please `git pull` from the repo and do an `(asdf:clear-source-registry)` in SBCL (I also updated the README, so you might want to check that out). If there's something broken, please open an issue in the `nx-freestance-handler`'s repo.

Ambrevar:
@kssytsrk Thanks for updating it, it's a most useful extension ;)

Ambrevar:
Closing this, I've opened many doc-related issues here: https://github.com/atlas-engineer/nyxt/labels/doc.
See also https://github.com/atlas-engineer/nyxt/issues/1109.

PR #1344, opened 2021-04-30
Toggle modes

Ambrevar:
This answers one of my primary itches with auto-mode: when one wants to override the behaviour of auto-mode _for the current buffer_, and _permanently_, we had to toggle the desired modes _and_ disable auto-mode, which was cumbersome.

With `toggle-modes`, it's now easier to select the desired modes and run the secondary action which disables auto-mode in the process.

I've seized this opportunity to remove `enable-modes-for-current-buffer`, etc. since they are superseded by `toggle-modes`.

Thoughts?

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1245,6 +1245,43 @@ ARGS are passed to the mode command."
     (loop for buffer in buffers
           do (enable-modes (uiop:ensure-list modes) buffer))))
 
+(defun all-mode-names ()
+  (mapcar #'name (sera:filter #'mode-toggler-p (list-commands))))
+
+(defun make-mode-suggestion (mode &optional source input)
+  "Return a `suggestion' wrapping around ATTRIBUTE. "
+  (declare (ignore source input))
+  (make-instance 'prompter:suggestion
+                 :value mode
+                 :attributes `(("Mode" ,(string-downcase (symbol-name mode)))
+                               ("Documentation" ,(or (first (sera:lines (documentation mode 'function)))
+                                                     "")))))
+
+(define-class mode-source (prompter:source)
+  ((prompter:name "Modes")
+   (prompter:multi-selection-p t)
+   (prompter:constructor (sort (all-mode-names) #'string< :key #'symbol-name))
+   (prompter:suggestion-maker 'make-mode-suggestion))
+  (:export-class-name-p t))
+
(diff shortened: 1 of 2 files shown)


jmercouris:
So, to be clear, the user will be prompted if they invoke `style-mode` manually, but if they invoke `toggle-modes` they will not be prompted. Is this correct and/or intended?

jmercouris:
Seems a little roundabout, but we weren't really using enable and disable-mode anyways :-D

source/buffer.lisp
```diff
@@ -1245,6 +1228,43 @@ ARGS are passed to the mode command."
...
+                                                                  (make-command force-disable-auto-mode (modes)
+                                                                    "Return selection but force disabling auto-mode.
+This is convenient when you use auto-mode by default and you want to toggle a
+mode permanently for this buffer."
+                                                                    (delete (read-from-string "nyxt/auto-mode:auto-mode" )
+                                                                            modes)))
+                                                   :marks (mapcar #'mode-name (modes (current-buffer))))))
+         (modes-to-disable (set-difference (all-mode-names) modes-to-enable
+                                           :test #'string=)))
+    (disable-modes (uiop:ensure-list modes-to-disable) (current-buffer))
```
jmercouris, line 1265: I prefer to capture (current-buffer) in the arguments to the command. So I would write `(define-command toggle-modes (&key (buffer (current-buffer)))`. What do you think?

Ambrevar: Good catch, fixed!

jmercouris:
Thank you!

Ambrevar:

> So, to be clear, the user will be prompted if they invoke `style-mode` manually, but if they invoke `toggle-modes` they will not be prompted. Is this correct and/or intended?

Prompted what?

- If they invoke style-mode manually, then auto-mode remains on.
  This means that on next browse, style-mode will be reverted.

- If they invoke toggle-modes to toggle style-mode, then auto-mode is
  disabled.
  This means that style-mode state will persist when browsing this buffer.

jmercouris:
Sorry, my head was in another space when I wrote that. Apply the following operations:

`s/not be prompted/optionally not be disabled by auto-mode`

jmercouris:
Anyways, your post provides the confirmation I wanted. All good!

Issue #1330, opened 2021-04-27
Remove search engines from bookmarks


Ambrevar:
Currently bookmarks have the option to store a "search-url" (and a "shortcut"): this effectively adds the capability for each bookmark to be a search engine entry.

Thus we have a dedicated function `all-search-engines` to return the list of search engines from the 2 main sources: the buffer `search-engines` slot and the bookmarks.

But bookmark search engines don't have any interface right now, they must be added manually to the file.
So I doubt that anyone is using this feature.

In the past, I had introduced this feature because we didn't have multiple source in the minibuffer.
But now that we do, I believe that bookmark search engines are obsolete: instead, the user should define all search engines in the `search-engines` buffer slot, and we can add a new source to `set-url-from-bookmarks` to also list the fallback URLs of the search engines.

Thoughts?

Ambrevar:
Bonus: it would simplify some code and the logic in the `search-engine.lisp` file.

Ambrevar:
Now that #1331 is merged, I'd like to apply this change.
Should I send a patch?  @jmercouris @aadcg @aartaka ?

aadcg:
All good from me. I don't use bookmarks, and all I need from search engines are the bangs from duckduckgo.

aartaka:
Sound alright!

jmercouris:
OK!

Ambrevar:
See #1346.

PR #1346, opened 2021-04-30
Search engine source

Ambrevar:
This remove search engine related slots from bookmarks as per #1330.


jmercouris:
Looks ready to me, thank you!

PR #1347, opened 2021-04-30
Prompter persistent actions

Ambrevar:
This is hopefully the last part of the prompter API that I want to edit before 2.0.

It renames `persistent-action` to "follow-mode-functions" and allows a list, even though it's not used at the moment (it will in the future).

The last part that I want to discuss: shall we rename `prompter:actions`?
It's not completely clear to newcomers what `actions` are.  In fact, they are the functions that are run over the result upon returning from the prompt buffer.

So what about `returning-functions`?  A bit clumsy maybe.
`result-processor`?

Or I can just leave it to "actions"...

Thoughts?


jmercouris:
I would leave it to `actions`. I think it is a quite servicable term :-). I will take a look at this pull request shortly!

aartaka:
`result-processor` sounds quite meaningful, but `actions` has an unbeatable simplicity to it, so I'm on the side of `actions`, especially given that we ensure good documentation.

libraries/prompter/prompter-source.lisp
```diff
@@ -405,13 +393,24 @@ See `resume-sources'.")
...
+                 "Execute the first `follow-mode-functions' after this delay
+when `follow-p' is non-nil.")
+
+   (follow-mode-functions '()
+                          :type (or null
+                                    (or function function-symbol)
+                                    (cons (or function function-symbol) *))
+                          :documentation
+                          "The first function of this list is called
+automatically when on the selection when it's changed.  It does not interrupt or
```
jmercouris, line 410: "when on the"

PR #1348, opened 2021-04-30
Fix nosave buffer

### LINKED ISSUE #1345: set-url-nosave-buffer should not set URL of non-nosave buffer, right?
Current, if I'm in a regular buffer and call `set-url-nosave-buffer`, confirm on a URL, it loads this URL in the current, non-nosave buffer.

This does not seem to be the intended behaviour.
Instead, I think the first action should test whether the current URL is a nosave buffer:
- if it is, call buffer-load.
- if it is not, call `new-nosave-buffer-load`.

Thoughts? @aartaka

Ambrevar:
Fixes #1345.
@aartaka @jmercouris What do you think?

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -3,6 +3,23 @@
 
 (in-package :nyxt)
 
+(export-always 'url)
+(defmethod url ((uri quri:uri))
+  uri)
+
+(defmethod url ((url-string string))
+  (quri:uri url-string))
+
+(defun has-url-method-p (object)
+  "Return non-nil if OBJECT has `url' specialization."
+  (some (lambda (method)
+          (subtypep (type-of object) (class-name
+                                      (first (closer-mop:method-specializers method)))))
+        (closer-mop:generic-function-methods  #'url)))
+
+(deftype url-designator ()
+  `(satisfies has-url-method-p))
+
 (export-always 'render-url)
 (declaim (ftype (function ((or quri:uri string)) string) render-url))
 (defun render-url (url)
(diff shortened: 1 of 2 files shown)


source/urls.lisp
```diff
@@ -3,6 +3,23 @@
 
 (in-package :nyxt)
 
+(export-always 'url)
```
jmercouris, line 6: A very welcome addition!

jmercouris: All of the changes in this file.

jmercouris:
It looks good to me. As an aside, why are all of our tests periodically failing? The tests should be fully reproducible, no?

aartaka:
This is just my assumption based on CI error logs, though.

aartaka:
I'm still on the side of not using `buffer-load` in `set-url-nosave-buffer` (see https://github.com/atlas-engineer/nyxt/issues/1345#issuecomment-830029757). However, that could be instructed by my own (possibly flawed) usage of `set-url` and friends, so no universal truths behind my opinion :)

Ambrevar:
@aartaka I'm not sure I understand, you mean replace `buffer-load` by what?

@jmercouris: Tests are failing because of a race condition in the `prompter` library.  It needs t be fixed.
The reason it's not reproducible is because race conditions depend on time, which depends on hardware and resources which vary constantly in the CI.

aartaka:
> @aartaka I'm not sure I understand, you mean replace `buffer-load` by what?

By nothing :) So that `set-url-nosave-buffer` always opens a new buffer. In this situation, opening URL in the current buffer (be it a web or nosave one) is only possible with `set-url`. This separation (`set-url`—current buffer, other `set-url-*` commands—new buffer) seems clearer to me.

Ambrevar:
OK, got it, I think you are right.

Shall we rename set-url-nosave-buffer to set-url-new-nosave-buffer then?

aartaka:
Yes, that's clearer :)

jmercouris:
I agree with the conclusion.

Ambrevar:
Merged with a5f30ec2417a50f8683cf9acea550a63755b1107.

Issue #1295, opened 2021-04-13
Leverage Plump for link hinting & virtual DOM?


aartaka:
[Plump](https://github.com/Shinmera/plump) is a good parser that can allow us to produce page models really fast, and traverse these without resorting to Parenscript. Several applications there:
- Element hinting, the procedure is:
  - Call Parenscript function that will add `nyxt-identifier` attribute to _all_ the nodes in the DOM.
  - Fetch the document with `ffi-buffer-get-document` and parse it with `plump:parse`.
  - Recursively traverse Plump-produced tree, matching the hintable elements by tag name and attributes (can use [CLSS](https://shinmera.github.io/CLSS/) for that).
  - Hint the elements by the `nyxt-identifier`s stored in Plump nodes.
- Visual mode, the procedure is the same as with element hints, what changes are actions and matching criteria.
- Building a virtual DOM for page prettification / change tracking / assistive tech / anything else. We can call Parenscript functions to add/delete/modify elements with matching `nyxt-identifier`s.

The main advantage is that we'll stay in Lisp land both syntactically and semantically (not the case with Parenscript) while managing all these cases. Code will most probably shorten too.

EDIT: ad -> add.
EDIT: Add links to Plump and CLSS.

aartaka:
#1290 can be a part of this almost-pure-Lisp transition.

Ambrevar:
Looks like a very good idea to me!

Indeed, Parenscript can be a bit complex to grasp, while being a niche language that not many people understand well.

If Plump could cut down the amount of Parenscript and simplify our code base, it could be a net win for Nyxt!

aadcg:
When is the white smoke coming out from your chimney, @jmercouris? I think you might have some justified concerns about this. 

If my understanding is correct, does it make sense to fetch the document with `ffi-buffer-get-document` (which runs JS) and then manipulating it with an HTML parser? What's the benefit? You could just run the JS you need.  Not to mention that web pages are a zoo of dynamic content, so perhaps we'd have to run `ffi-buffer-get-document` before each hinting/visual-mode/whatever call. 

I have no idea what I'm talking about, so these are just impressions. It is also my impression that parenscript is a jewel that I don't understand it, not even superficially.

aartaka:
[...]

PR #1351, opened 2021-05-02
Make qs and qsa Parenscript macros.

aartaka:
This turns our ubiquitous `qs` and `qsa` definitions into Parenscript macros, so that we have them defined in one place (renderer-script.lisp) and called everywhere, without writing any additional code.

# What's new
- `qs` and `qsa` rewritten as Parenscript macros.
- `ps:defpsmacro` used for the first time in Nyxt code, it seems (see Things to Discuss).
- Much less code duplication :D 

# Things To Discuss
Was there a reason to not define these as Parenscript macros before? I feel it's quite cool to leverage what Parenscript provides, especially with such universally useful utilities as `qs` and `qsa`, but am I missing some part of the picture?

# How Has This Been Tested
- Compiled.
- Ran.
- Tried `follow-hint`.
- It worked :)

What do you think of it? What else can we define as PS macros?

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -83,6 +83,7 @@
                (:file "data-storage")
                (:file "configuration")
                (:file "command")
+               (:file "parenscript-macro")
                (:file "renderer-script")
                (:file "buffer")
                (:file "window")
--- a/source/element-hint-mode.lisp
+++ b/source/element-hint-mode.lisp
@@ -10,7 +10,7 @@
 (define-command toggle-hints-transparency (&key (buffer (current-buffer)))
   "Toggle the on-screen element hints transparency."
   (pflet ((toggle-transparent ()
-            (ps:dolist (element (nyxt:qsa document ".nyxt-hint"))
+            (ps:dolist (element (nyxt/ps:qsa document ".nyxt-hint"))
               (if (or (= (ps:chain element style opacity) "1")
                       (= (ps:chain element style opacity) ""))
                   (setf (ps:chain element style opacity) "0.2")
--- a/source/element-hint-mode.lisp
+++ b/source/element-hint-mode.lisp
@@ -10,10 +10,7 @@
 (define-command toggle-hints-transparency (&key (buffer (current-buffer)))
   "Toggle the on-screen element hints transparency."
   (pflet ((toggle-transparent ()
-            (defun qsa (context selector)
-              "Alias of document.querySelectorAll"
-              (ps:chain context (query-selector-all selector)))
-            (ps:dolist (element (qsa document ".nyxt-hint"))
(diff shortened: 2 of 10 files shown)


aartaka:
Another question: which package should these be defined in?
- `nyxt` is sane, as this is Nyxt-specific extension to Parenscript. The Parenscript code starts getting too prefixful, though.
- `ps` has a shorter prefix, and Parenscript macros can be said to belong to Parenscript exclusively—we're kind of doing an extension to Parenscript. Namespace pollution, though.

What does sound better?

Ambrevar:
To the question "why we didn't do it before": it's been on my todo-list almost forever, but I never took the time to dig deeper into PS ;)

As for the package: I think we had already discussed it (where?) but I can't remember the conclusion.
`nyxt/parenscript` (with the `nyxt/ps` nickname) is an option.

Finally, other factorable ps:

- insert-at
- element-drawable-p
- element-in-view-port-p
- strings-generate
- hints-generate
- hints-determine-chars-length
- hints-add

jmercouris:
I agree with the nyxt/ps. It will ensure no conflicts in future revisions of parenscript. As per why not? My answer is slightly different to Pierre's, I did not know such functionality existed! Or if I did, I had forgot about it! :-D

I'll go ahead and clean this up and merge it. I'll also open up an issue about the other parenscript macros we can implement.

aartaka:
> To the question "why we didn't do it before": it's been on my todo-list almost forever, but I never took the time to dig deeper into PS ;)
> 
> As for the package: I think we had already discussed it (where?) but I can't remember the conclusion.
> `nyxt/parenscript` (with the `nyxt/ps` nickname) is an option too.

Hmmmm. Let's just go with `nyxt` for the time being.

> Finally, other factorable ps:
> 
>     * insert-at
>     * element-drawable-p
>     * element-in-view-port-p

These I can agree with!

>     * strings-generate
>     * hints-generate
>     * hints-determine-chars-length
>     * hints-add

These are element-hints-specific. I'm currently working on #1295 (this pull request is a side-effect of this work), so let's see -- maybe we can come to a better design having less Parenscript to abstract away :) 

Let me write Parenscript macros for `insert-at`, `element-drawable-p`, and `element-in-view-port-p`, and we can merge!

aartaka:
> I agree with the nyxt/ps. It will ensure no conflicts in future revisions of parenscript. As per why not? My answer is slightly different to Pierre's, I did not know such functionality existed! Or if I did, I had forgot about it! :-D
> 
> I'll go ahead and clean this up and merge it. I'll also open up an issue about the other parenscript macros we can implement.

Go ahead! Seems like we're writing simultaneous comments again xD

jmercouris:
OK! Will do :-D Ay yai yai!

jmercouris:
Thank you Artyom!

Issue #1168, opened 2021-02-21
Make Nyxt configurable as headless


aartaka:
Headless (GUI-less) browsers are the most reliable way to script the Web-based activity (one use-case is web scraping, as in #1163), so Nyxt can benefit from having a headless option. For this, we just need to not create GUI frames and run in the background, while rendering webpages to the non-displayable WebView.

Ambrevar:
Great idea!

Ambrevar:
On master!

PR #1354, opened 2021-05-03
WIP: Headless mode with prompt buffer tests

Ambrevar:
This is a draft that shows how we can implement headless mode (see #1168) and makes use of this feature to test:

- The renderer;
- The interactive prompt buffer.

And it works!

A few points to address:

- "set-url" hangs on and off at `prompter:all-ready-p`.  Race condition?
- We need to add a few event signals, like:
  - when Nyxt is started and ready for interactions;
  - when URL is done loading (we already have this via methods, maybe tests could just add a default-mode to the tested buffer).
 
Thoughts?

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -290,7 +290,15 @@
                cl-gobject-introspection
                bordeaux-threads)
   :pathname "source/"
-  :components ((:file "renderer-gobject-gtk")))
+  :components ((:file "renderer-gobject-gtk"))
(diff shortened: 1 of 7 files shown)


Ambrevar:
Also try in a TTY, without X / Wayland.

aartaka:
What's the status of this PR?

Ambrevar:
Still very high priority in my opinion!

I haven't found the time to work on it, hopefully I can get back to it
soon.
Unless someone beats me to it ;)

aartaka:
Closing in favor of #2054 (rebase).

Issue #1352, opened 2021-05-03
How to select nothing in the prompt buffer?


Ambrevar:
There is a little "hole" in our prompt buffer user experience: some prompt buffer are meant to return a list of things, possibly an empty one, but there is no way to return "nothing".

An example of this can be seen with `toggle-attributes-display` (`C-]` in a prompt-buffer).

Another example, with a workaround this time, is when selecting a tag for bookmarks: I've added a special source for new tags, including the empty one.  If the user select just the empty tag, since the caller discard the empty string (`""`) it amounts to no tag at all.

We need a better UX.  Even the workaround seems a bit clunky. 
Any suggestion?

aadcg:
Thanks for writing this up.  I noticed it when working on recursive search but didn't write it up.

I'm afraid I don't have any suggestions.

jmercouris:
I think we could add a source called the "empty source" that has an option "return nothing". What do you think?

Ambrevar:
That could do, but this would only work if there is no mark.

I'm afraid this use case would be a bit confusing:

- C-space
- C-]
- It shows 2 sources:
  - The empty source at the top
  - The source with some marked attributes at the bottom.
- The user selects the first element empty source element.

Expectation: disable all attributes but the default one.
Result: no change, since some attributes were marked.

Can we avoid this ambiguity and make the behaviour explicit?

jmercouris:
What if the text said "unmark all and select nothing"?

Ambrevar:
Ha!  That's interesting: a source with an action that discards all other
elements.

I'll send a pull request so we can test this.

PR #1356, opened 2021-05-03
Empty action

Ambrevar:
This answers #1352.
In the end, I couldn't add a separate source because if there are marks in _any_ source, then all unmarked elements are ignored.
The action of unmarked elements is ignored.

Here I took another approach: add an action which discard the selection if there are no marks.  I've made sure that the prompt is explicit to the user.

Thoughts?

### DIFF
--- a/source/prompt-buffer-mode.lisp
+++ b/source/prompt-buffer-mode.lisp
@@ -190,12 +190,24 @@ If STEPS is negative, go to next pages instead."
 (define-class attribute-source (prompter:source)
   ((prompter:name "List of prompter attributes")
    (prompter:multi-selection-p t)
-   (prompter:suggestion-maker 'make-attribute-suggestion)))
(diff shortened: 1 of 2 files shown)


aartaka:
I like this movement of `(funcall* return-function)` higher in the `hide-prompt-buffer`! Wanted this for a long time :)

Everything is fine, it seems.

aartaka:
How about extracting `return-marks-only` to a top-level function and exporting it? Maybe make a function-maker. Then it can be used in much more contexts than the original (attributes choice) one.

Ambrevar:
Good point, I'll try to extract it.

Ambrevar:
Done and merged with 7c29efddb3fffadf75ddea734bd3a127216f2607.

jmercouris:
Thank you!

Issue #1342, opened 2021-04-29
Replace URL with URI?


jmercouris:
All URLs are URIs. I suggest we replace all references to "URL" in the codebase with "URI" for consistency. I suggest this for post 2.0.0.

aadcg:
Not long ago @Ambrevar told me:

> Also we adopted the convention to always use "URL" and not "URI".

I'm confused.

jmercouris:
I don't remember that, perhaps it happened :-D

Ambrevar:
John, we decided the other way around ;)

jmercouris:
Ah, so we replace all references to URI with URL? Something something getting old :-D

Ambrevar:
Which ones?  We've been using "URL" for a while now, there should be
very few "URI" occurrences left.

jmercouris:
For example, ffi-buffer-load, uses URI instead of URL in the defgeneric.

jmercouris:
To be clear this is *not* about replacing URL strings with Quri:URI objects. I am talking about the usage of the word "URL" within the codebase vs "URI".

Ambrevar:
Yes, that's what I had understood.

Let's fix ffi-buffer-load then :)

jmercouris:
I am a little bit surprised that we voted for URL. Quri makes URIs and we are using it. They may be URLs, but they are ALWAYS URIs. Oh well, we can always change it later.

Ambrevar:
If I recall correctly, one of the elements of the rationale was that URL
is more widespread among browsers and a more familiar terms for newcomers.

jmercouris:
That is a good rationale. I think that at the time we were not making extensive use of `quri:uri`. Now that we are, it strikes me as a logical inconsistency!

Ambrevar:
Yup, but there is little we can do about external libraries.
I guess it's the price to pay for leveraging other projects :)

jmercouris:
I am still unsettled by this. I keep wanting to use a singular term, and while URI is less known, it is the semantically correct term to use. This can wait until after 2.0.0.

Ambrevar:
URL are specific kinds of URI.
In the context of Nyxt, almost all (if not all?) are URLs.

jmercouris:
You're correct, but we can say URI to be more general.

Ambrevar:
If we only targeted tech-minded users, we should go with URI.

But "URI" may put back our audience a bit, so if we have to choose
[...]

PR #1358, opened 2021-05-04
Rename URI to URL.

aadcg:
See #1342.


source/urls.lisp
```diff
@@ -29,12 +29,12 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
                  (quri:render-uri url))))
     (the (values (or string null) &optional)
-         (or (ignore-errors (ffi-display-uri url))
+         (or (ignore-errors (ffi-display-url url))
              url))))
 
 (export-always 'valid-url-p)
 (defun valid-url-p (url)
-  ;; List of URI schemes: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
+  ;; List of URL schemes: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
```
Ambrevar, line 37: This is actually URI.

aadcg: Yes, I noticed this mistake too. I will fix it. Thanks.

Ambrevar:
The `url<` type needs its return type to be fixed:

```lisp
(declaim (ftype (function (quri:uri quri:uri) (or null fixnum)) uri<))
```

Ambrevar:
Great job, thanks!

aadcg:
> The url< type needs its return type to be fixed

Indeed. I wrote it as `(or boolean integer)`.  I get the difference between integer and fixnum. What about the difference boolean and null?

aadcg:
There are some other things I need to address here. I'll fix it.

Ambrevar:

> Indeed. I wrote it as `(or boolean integer)`.  I get the difference between integer and fixnum. What about the difference boolean and null?
>

Only NIL satisfies the NULL type.
Both NIL and T satisfy the BOOLEAN type.

aadcg:
Ok, this look good to me @Ambrevar @jmercouris @aartaka.

source/buffer.lisp
```diff
@@ -478,29 +478,29 @@ See `buffer-make'."
   (and (web-buffer-p buffer)
        (eq (slot-value buffer 'load-status) :failed)))
 
-(export-always 'on-signal-notify-uri)
-(defmethod on-signal-notify-uri ((buffer buffer) no-uri)
-  "Set BUFFER's `url' slot, then dispatch `on-signal-notify-uri' over the
+(export-always 'on-signal-notify-url)
```
aartaka, line 481: Actually, this one better stay with URI, because it's a callback for WebKit signal called `notify::uri`.

source/history.lisp
```diff
@@ -51,12 +51,12 @@ class."
     (setf (slot-value he 'url) (ensure-url (slot-value he 'url))))
   (slot-value he 'url))
 
-(defmethod s-serialization::serialize-sexp-internal ((uri quri:uri)
+(defmethod s-serialization::serialize-sexp-internal ((url quri:uri)
```
aartaka, line 54: This one is directly related to the type, so maybe it's better to leave it as URI too?

source/buffer.lisp
aartaka: Same applies to other `on-signal-notify-uri` methods in mode.lisp and web-mode.lisp.

source/history.lisp
aadcg: Fair.

source/buffer.lisp
aadcg: Fair.

aadcg:
@aartaka, I've addressed your remarks.

Ambrevar:
Thanks a lot!

PR #1359, opened 2021-05-04
WIP: Automatic RPM package build for Fedora

Ambrevar:
The long-awaited!

It's still a work-in-progress.

Next things to be done:
- Include all the deps properly (webkit2gtk)
- See if linux-packaging works.
- Bundle libfixposix in the RPM.
- Test!

### DIFF
--- a/.github/workflows/package-fedora.yml
+++ b/.github/workflows/package-fedora.yml
@@ -1,0 +1,15 @@
+name: Build .rpm package for latest Fedora
+
+on:
+  push:
+    # tags: '*'
+    ## To test, uncomment the line below with branch name used for review.
+    branches: [ rpm ]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build the Docker image
+      run: docker build . --file Dockerfile --tag ${{ github.repository }}:$(date +%s)
--- a/.github/workflows/package-fedora.yml
+++ b/.github/workflows/package-fedora.yml
@@ -12,12 +12,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Build the Docker image
-      run: docker build -t example:${{ github.sha }} . --file Dockerfile --tag ${{ github.repository}}:$(date +%s)
+      run: docker build -t example:${{ github.sha }} . --file Dockerfile
     - uses: shrink/actions-docker-extract@v1
       id: extract
       with:
         image: example:${{ github.sha }}
-        path: /root/nyxt/*.rpm
+        path: /foo
     - name: Upload Dist
       uses: actions/upload-artifact@v2
       with:
--- a/.github/workflows/package-fedora.yml
+++ b/.github/workflows/package-fedora.yml
@@ -15,7 +15,10 @@ jobs:
       id: buildrpm
       run: |
         docker build -t example:${{ github.sha }} . --file Dockerfile
-        docker cp example:${{ github.sha }}:$(docker exec -it example:${{ github.sha }} bash -c 'ls -1 /root/nyxt/*.rpm') .
+        echo "list"
+        docker images
+        echo "CP"
+        docker cp $(docker ps --latest --quiet):$(docker exec -it $(docker ps --latest --quiet) bash -c 'ls -1 /root/nyxt/*.rpm') .
     # - name: Copy output
     #   id: rpmoutput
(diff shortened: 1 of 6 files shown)


jmercouris:
Should I test this on Fedora? I can install Fedora on a VM as well if you like.

jmercouris:
Well, I've installed Fedora, when you'd like me to test, please let me know!

Ambrevar:
Thanks a lot, I was hesitant to ask because I know it's not so much fun!
The .rpm is not ready yet, I'll try to get it to work as soon as
possible :)

jmercouris:
OK! Let me know when!

Ambrevar:
I've made progress: it now reaches to the linux-packaging step, which fails because it cannot find the fedora package behind libfixposix (built manually).

So what are out options?  I can think of the following:
- Ask Fedora upstream to package libfixposix.  We should certainly do this in all cases, but it will take a while before it reaches the stable repository.
- Make an RPM ourselves.
- Have linux-packaging handle errors when a library cannot be found.

I'll go for the last option for now, see what upstream thinks.

Ambrevar:
See https://gitlab.com/ralt/linux-packaging/-/issues/9.

jmercouris:
Understood, thank you!

Ambrevar:
Progress!  it turns out that libfixposix is in Fedora after all (their package search is a bit unreliable... :p) so this blocker is gone.

Next: `fpm` fails for a reason I don't really get (missing ruby dep?):
https://gitlab.com/ralt/linux-packaging/-/issues/10

dtw-waleee:
I skimmed the build log above. Don't you need to install some *-devel rpms? gtk3-devel seems like a possible suspect.

Ambrevar:
The build log indicates an issue with `fpm` (and ruby) itself.  If it were a GIR problem, it would only show up on startup, not when building the .rpm.

Ambrevar:
Great news, the .rpm just got built!
Hooray, we're (probably) going to have a .rpm for the 2.0 release!

jmercouris:
Fantastic, I'll try this soon!

jmercouris:
I can't seem to find the artifact. I've checked in multiple locations, is there something you have to set?

Ambrevar:
It's not working yet, patience :)

Ambrevar:
To clarify: the .rpm is built _inside the docker_, now I need to
understand how it's meant to be extract and published as a GitHub artifact.

jmercouris:
I see!

jacksongoode:
Is the rpm locatable?

Ambrevar:
By "locatable", are you asking if it's currently published?
If so, then no, since this pull request hasn't been merged.

aartaka:
Reminder for this PR: `guix pack rpm` may work better than any CI generation action. Also, now that we have Flathub, is it that much relevant?

Maybe close?

aadcg:
>  Reminder for this PR: guix pack rpm may work better than any CI generation action. Also, now that we have Flathub, is it that much relevant?

It would be simpler to generate the rpm via `guix pack` than the approach drafted in this PR indeed. But that would need to be triggered by a CI action, just like we do today with the `guix pack` that outputs a tarball (the self-contained bundle).

> Maybe close?

Yes, I'd close it since the Flatpak is superior in every aspect.

It works under all GNU/Linux distributions. One format suffices to reach all users, in contrast to generating .deb or .rpm packages. Even if it's effortless to generate those specific package formats, we'd need to maintain the each of the testing environments. With the flatpak, it's trivial to create the environments suitable to generate and run it.

I'm closing it. CC @jmercouris 

@Ambrevar Is it ok to delete this branch?

jmercouris:
I wonder if package mantainers can wrap our flatpak into a 'deb' or whatever.

Ambrevar:
Before deleting this branch, I'll move the content together with the .deb generating one.

I'd like to clarify a few things:

- `guix pack -f rpm` is not a replacement for RPM generation: the generated RPM contains all recursive deps, while this branch tries to generate an RPM that only contains Nyxt and uses the systems deps.
- Flatpak and friends are not strictly superior: they embed the deps, and as such:
  - they are bigger
  - the deps may not be updated separately (e.g. security update for OpenSSL).
 
All that said, ultimately each distribution should provide their tailored Nyxt package, it's too much work for us to maintain the `.deb`, the `.rpm`, etc.

aadcg:
Acknowledged @Ambrevar. All points are clear, thanks.

Ambrevar:
Done, see https://github.com/atlas-engineer/build-script-archive/commit/894aacf561bba0ffab48b71692b1c275dda4cf4e.

PR #1360, opened 2021-05-04
Search autocompletion.

### LINKED ISSUE #79: Search engine dynamic suggestions (alias "webjumps")

aartaka:
Behold! Search suggestions in Nyxt! (see an attached screenshot)

autocompletion

This adds search autocompletion requests to the default `search-engine`s, based on the new extensible mechanism:
- `completion-function` slot of `search-engine', and
- `make-search-completion-function` helper to make completion functions.

Works, loads things (although takes some time with both Wikipedia and DuckDuckGo), shows fancy search suggestions!

(I'm going to add completion functions to `nx-search-engines` right after we merge this, as this is too exciting to ignore!)

How does it look for you?

Closes #79.

EDIT: Small readability fixes.
EDIT: "deafult" typo :)
EDIT: Add a linked issue.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -47,11 +47,22 @@ after the mode-specific hook.")
    (search-engines (list (make-instance 'search-engine
                                         :shortcut "wiki"
                                         :search-url "https://en.wikipedia.org/w/index.php?search=~a"
-                                        :fallback-url (quri:uri "https://en.wikipedia.org/"))
+                                        :fallback-url (quri:uri "https://en.wikipedia.org/")
+                                        :completion-function
+                                        (make-search-completion-function
+                                         :base-url "https://en.wikipedia.org/w/api.php?action=opensearch&format=json&search=~a"
+                                         :processing-function (alex:compose #'second #'json:decode-json-from-string)))
                          (make-instance 'search-engine
                                         :shortcut "ddg"
                                         :search-url "https://duckduckgo.com/?q=~a"
-                                        :fallback-url (quri:uri "https://duckduckgo.com/")))
+                                        :fallback-url (quri:uri "https://duckduckgo.com/")
+                                        :completion-function
+                                        (make-search-completion-function
+                                         :base-url "https://duckduckgo.com/ac/?q=~a"
+                                         :processing-function
+                                         #'(lambda (results)
+                                             (mapcar #'cdar
+                                                     (json:decode-json-from-string results))))))
                    :type (cons search-engine *)
                    :documentation "A list of the `search-engine' objects.
 You can invoke them from the prompt-buffer by prefixing your query with SHORTCUT.
@@ -1001,10 +1012,19 @@ Finally, if nothing else, set the `engine' to the `default-search-engine'."))
                                          (mapcar #'shortcut engines)
                                          :test #'string=))
               (list (make-instance 'new-url-query :query input)))
-            (mapcar (lambda (engine) (make-instance 'new-url-query
-                                                    :query (str:join " " (rest terms))
-                                                    :engine engine))
-                    engines))))
+            (alex:mappend (lambda (engine)
+                            (append
+                             (list (make-instance 'new-url-query
+                                                  :query (str:join " " (rest terms))
+                                                  :engine engine))
+                             ;; Some engines (I'm looking at you, Wikipedia!)
+                             ;; return garbage in response to an empty request.
+                             (when (and (completion-function engine) (rest terms))
+                               (mapcar (alex:curry #'make-instance 'new-url-query
+                                                   :engine engine :query)
+                                       (funcall (completion-function engine)
+                                                (str:join " " (rest terms)))))))
+                          engines))))
 
 (define-class new-url-or-search-source (prompter:source)
   ((prompter:name "New URL or search query")
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -44,6 +44,9 @@ after the mode-specific hook.")
 after the mode-specific hook.")
    (keymap-scheme-name scheme:cua
                        :documentation "The keymap scheme that will be used for all modes in the current buffer.")
+   (search-auto-complete-p t
+                           :type boolean
+                           :documentation "Whether search suggestions are requested and displayed.")
    (search-engines (list (make-instance 'search-engine
                                         :shortcut "wiki"
                                         :search-url "https://en.wikipedia.org/w/index.php?search=~a"
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -470,7 +470,7 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
                             nil)))
 
 (declaim (ftype (function (buffer &key (:downloads-only boolean))) proxy-adress))
-(defun proxy-address (buffer &key (downloads-only nil))
+(defun proxy-url (buffer &key (downloads-only nil))
   "Return the proxy address, nil if not set.
 If DOWNLOADS-ONLY is non-nil, then it only returns the proxy address (if any)
 when `proxied-downloads-p' is true."
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -463,7 +463,7 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
   (setf (slot-value buffer 'proxy) proxy)
   (if proxy
       (ffi-buffer-set-proxy buffer
-                            (server-address proxy)
+                            (url proxy)
                             (allowlist proxy))
       (ffi-buffer-set-proxy buffer
                             (quri:uri "")
@@ -478,7 +478,7 @@ when `proxied-downloads-p' is true."
          (proxied-downloads (and proxy (proxied-downloads-p proxy))))
     (when (or (and (not downloads-only) proxy)
               proxied-downloads)
-      (server-address proxy))))
+      (url proxy))))
 
 (defmethod initialize-modes ((buffer buffer))
   "Initialize BUFFER modes.
--- a/source/search-engine.lisp
+++ b/source/search-engine.lisp
@@ -36,6 +36,16 @@ Can be built via `make-search-completion-function'"))
                  :search-url search-url
                  :fallback-url fallback-url))
 
+;; TODO: Move elsewhere?
+;; TODO: Add `proxied-post'?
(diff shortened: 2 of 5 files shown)


jmercouris:
Absolutely blows my mind! Fantastic work Artyom!

That is just so damn cool!

Ambrevar:
Awesoooooooooooooooooooome!!! Thank you so much for working on this, will review as soon as possible!

P.S.: I see you had fun with theming :)

jmercouris:
It seems Pierre is not familiar with the venerable Laconia theme!

Ambrevar:
The related issue is #79, one of our oldest! ;)

jmercouris:
Resources: https://github.com/aartaka/nyxt-config/blob/master/style.lisp
https://github.com/aartaka/laconia-theme

Ambrevar:
I've tested it, it's absolutely amazing.
I suggest we merge this now and include it in 2-pre-release-7.
@jmercouris ?

source/search-engine.lisp
```diff
@@ -32,6 +36,28 @@ the empty string."))
...
                  :fallback-url fallback-url))
 
+(export-always 'make-search-completion-function)
+(declaim (ftype (function (&key (:base-url string)
+                                (:request-function (function (string &rest *) *))
+                                (:processing-function (function (*) list-of-strings)))
+                          (function (string) list-of-strings))
+                make-search-completion-function))
+(defun make-search-completion-function (&key base-url
+                                          (request-function #'dex:get)
```
Ambrevar, line 46: dex:get must use some proxy settings, no?
But which ones?  Does it make sense for the prompt buffer to use proxy settings of the current-buffer?
Should it have its own proxy settings?

Here is my suggestion: make a wrapper around `dex:get` that's uses the prompt buffer proxy.
Then the user can enable the proxy by adding proxy-mode to the prompt-buffer or not.
Does that make sense?

jmercouris:
I think we best leave it for actual 2.0.0. It is too much of a last minute thing in my opinion. Rather, let Artyom fix it in peace :-). What do you think?

Ambrevar:
Agreed!

aartaka:
Things left to do before merging:
- auto-completion using default search engine, e.g., when you type `salad` and it gives you `salad recipe`, `salad dressing` etc. even without the search engine shortcut.
- Proxying.
- Making auto-completion togglable: there are people as paranoidal as myself in regards to auto-completion, so we better make it disableable :D

Ambrevar:
Agreed on all points! :)

aartaka:
- Implemented:
  - Proxied auto-completion.
  - A buffer-located switch for auto-completion.
  - Auto-completion by default, for any query.
- Rebased over master to use the fix for the bug found in the process (11425b2300cbe295fc25beef7c52ebaf6a67cc20).
- Tested proxied downloads and disabled auto-completion. Works fine.

Mergeable, I believe :)

aartaka:
+ CI passes. Phew...

source/search-engine.lisp
```diff
@@ -32,6 +36,38 @@ the empty string."))
...
 
+;; TODO: Move elsewhere?
+;; TODO: Add `proxied-post'?
+(defun proxied-get (url &rest args)
+  "A version of `dex:get' that enables proxy in case `current-buffer' has one enabled."
+  (let ((args (append args (list :proxy (and (web-buffer-p (current-buffer))
+                                             (proxy-address (current-buffer))
+                                             (quri:render-uri
+                                              (proxy-address (current-buffer))))))))
+    (apply #'dex:get url args)))
```
Ambrevar, line 47: What about adding a `(proxy-url-string BUFFER)` function, then at the call site use

```lisp
(dex:get :proxy (proxy-url-string PROMPT-BUFFER))
```

?

By the way, shouldn't we rename `proxy-address` to `proxy-url` for consistency?

source/buffer.lisp
```diff
@@ -992,17 +1006,30 @@ Finally, if nothing else, set the `engine' to the `default-search-engine'."))
...
-                   (all-search-engines)
-                   :key #'shortcut)))
-    (append (unless (and engines (string= (first terms) (shortcut (first engines))))
-              (list (make-instance 'new-url-query :query input)))
-            (mapcar (lambda (engine) (make-instance 'new-url-query
-                                                    :query (str:join " " (rest terms))
-                                                    :engine engine))
-                    engines))))
+  (flet ((make-engine-completions (engine input)
+           (when (and (and (current-buffer) (search-auto-complete-p (current-buffer)))
```
Ambrevar, line 1010: Nested `and` can be merged.

Ambrevar:
So how does one configure the prompt-buffer to be proxied?

aartaka:
There's no way as of now: I use `current-buffer` (which prompt-buffer cannot be) proxy to decide on proxying auto-completion. So, prompt-buffer is not involved. 

Should prompt-buffer have a proxy? I'm not sure -- prompt-buffer is made for internal interface use, after all.

source/search-engine.lisp
aartaka: Hmmm. I'm not sure about the `proxy-url-string` helper. Let's leave it the way it is and see whether we need it in the future :)

Renaming `proxy-adress` to `proxy-url` sounds intuitive! I'll do that.

jmercouris:
I agree with Artyoms assessment

jmercouris:
I understood that the prompt buffer will use the proxy settings of the current buffer

Ambrevar:
> I understood that the prompt buffer will use the proxy settings of the current buffer

But that's not correct.  For instance it will fail to do the right thing
when run from the *Help* buffer.

aartaka:
Fair point! While I don't feel right about adding proxy to the prompt-buffer, I see the problem.

jmercouris:
I would say don't add it to the prompt buffer, add it to the browser, that makes more sense. It can then be used for downloads with the lisp engine etc.

jmercouris:
I don't see that as likely, but a fair point. Being more specific has benefits.

It also has costs. Will the user know to set their proxy for every single aspect of the system?

Ambrevar:
> It also has costs. Will the user know to set their proxy for every single aspect of the system?

We can set the default value to that of (proxy *browser*).

Done! :)

Ambrevar:
Sorry, I wasn't very clear: We can set the default value _of the various local proxy slots_ to that of `(proxy *browser*)`.

jmercouris:
I see, that seems a good solution.

aartaka:
Hmmmm... How about yet another approach:
- Don't define auto-completion functions in Nyxt core.
- Leave proxied completion functions to be implemented by extension writers (because this logic actually belongs to extensions).
- Add `:request-args` keyword argument to the `make-search-completion-function` and use `dex:get` by default for `:request-function`. Example of usage:
```lisp
(make-search-completion-function
 :base-url "https://duckduckgo.com/ac/?q=~a"
 :request-args '(:proxy "socks5://localhost:9050")
 :processing-function
 #'(lambda (results)
     (mapcar #'cdar
             (json:decode-json-from-string results))))
```

Given that `nx-search-engines` is already destined to be a testbed for completion, I suggest completions functions (including proxied ones) are written there.

Rationale: 
- `prompt-buffer` is merely a display for search suggestions, so it doesn't make sense to add proxies to it.
- `buffer` makes more sense, given that most search-related slots (`search-engines`, `search-auto-complete-p`) are there. Why not use `buffer`'s `proxy` then? Because it's not every buffer that has proxies. Maybe instantiate buffers and enable `proxy-mode`? That's unreliable -- too much variables. Adding search-dedicated proxy slot doesn't sound right either.
- Adding `*browser*`-based proxy and inheriting that configuration is cool, but that's another global variable that supercedes a carefully crafted mode. Doesn't sound right.
- The mechanism is intended to be extensible, so why impose any implementation? If user wants to have things proxied, they can always override the `completion-function`. `make-search-completion-function` is intended for that, especially so with `:request-args`.

jmercouris:
I concur that the `make-search-completion-function` is the appropriate way for the user to customize this. It does not feel right that the `prompt-buffer` contain proxy information. It seems to me that would be outside of the prompt buffers concerns.

Ambrevar:
I agree with your last proposal, @aartaka! :)

aartaka:
> I agree with your last proposal, @aartaka! :)

I'll push it in an hour then!

aartaka:
Here it is: `:request-args` and `dex:get` used by default. Two things to discuss:
- Should I remove the default search engines' (DDG, Wiki) completion?
- `proxy-url` in `buffer` conflicts with `proxy-url` in `renderer-gtk`. Maybe rename GTK one?

jmercouris:
Thank you Artyom! I would leave the default search engine completion for now. I will look into what to do about the naming conflict!

jmercouris:
Perhaps it would make sense to name it %proxy-url in renderer-gtk.lisp, since it is an implementation detail.

jmercouris:
Or is it...? I guess it isn't. Decisions decisions...

jmercouris:
No, definitely it should be set to %proxy-url in `renderer-gtk.lisp`.

jmercouris:
@aartaka if you approve, I shall merge!

jmercouris:
I decided to just call it `gtk-proxy-url` in the end.

Ambrevar:
Looks good to me, except that I think we should expand add an example on how to add a proxy, and refer to this in the manual.

aartaka:
> Looks good to me, except that I think we should expand add an example on how to add a proxy, and refer to this in the manual.

Fair point. Done in 65092494ed1356baced050a207e76803db99cb7a. Feels all good to merge otherwise :)

PR #1363, opened 2021-05-05
Rename gobject-gtk renderer to gi-gtk.

### LINKED ISSUE #1361: Rename gobject-gtk to gi-gtk or gigtk
gobject-gtk is inaccurate, since the renderer is not about gobject but rather "GObject Introspect" (the relationship between the two is like what MOP is to CLOS).

What about renaming it to `gi-gtk` instead or even shorter, `gigtk`?

Besides, the system naming is inconsistent: the original renderer is named "gtk" but the gobject-gtk renderer is named "gobject/gtk" (notice the slash).  Beside being inconsistent, the slash means "subsystem" for ASDF, but there is no "gobject" parent system here.

Thoughts?

Ambrevar:
Also rename the gobject/gtk system to gi-gtk.

Fixes #1361.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -37,12 +37,12 @@ to, use your local install of Quicklisp.  For example:
 But in this case, you don't have to use `make' at all.  Instead, you could build
 Nyxt with an invocation along those lines:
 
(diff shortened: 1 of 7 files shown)


jmercouris:
Aha, the tests fail on CCL because we only handle the SBCL condition for joining a thread that is stopped! @Ambrevar

Ambrevar:
OK to merge?

PR #1364, opened 2021-05-05
Nyxt quicklisp asd

Ambrevar:
The Quicklisp-in-asd setup has given me enough headaches...

There is a whole class of problems we can get rid of if we just move the setup to its own .asd file.  This is what this pull request does.

WARNING: Don't merge before removing the .deb test commit!

### DIFF
--- a/.github/workflows/package-ubuntu.yml
+++ b/.github/workflows/package-ubuntu.yml
@@ -2,9 +2,9 @@ name: Build .deb package for latest Ubuntu
 
 on:
   push:
-    tags: '*'
+    # tags: '*'
     ## To test, uncomment the line below with branch name used for review.
-    # branches: [ MY-WORK-BRANCH ]
+    branches: [ nyxt-quicklisp-asd ]
 
 jobs:
   build:
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,8 @@ endif
 
 lisp_eval:=$(LISP) $(LISP_FLAGS) \
 	--eval '(require "asdf")' \
+	--eval '(when (string= "$(NYXT_INTERNAL_QUICKLISP)" "true") (setf asdf:*default-source-registries* nil) (asdf:clear-configuration) (asdf:load-asd "$(makefile_dir)/nyxt-quicklisp.asd") (asdf:load-system :nyxt-quicklisp))' \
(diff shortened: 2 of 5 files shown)


Ambrevar:
@jmercouris Thoughts?

jmercouris:
I don't understand enough to contribute anything meaningful. Could you please explain why keeping everything in a single ASD causes issues?

jmercouris:
I'm not saying I disagree with the change (if it does fix things of course :-)), I just would like to understand as well!

Ambrevar:
The .asd file is known to ASDF by the means of `load-asd`, and it's not necessarily in any ASDF registry.

When `ql:update-dist` is run, the ASDF state is somehow reset (not sure how exactly), and ASDF fails to find nyxt.asd again, which means the execution of the `nyxt/quicklisp` system fails!

The workaround is to load the .asd twice.

There has been other issues related to this, and the workarounds are very global and stateful.

Separating the files removes a whole class of problems and these global workarounds, which means we can be much more confident with our build system in the future.

jmercouris:
OK. I see. No idea why this would behave this way. Please feel free to merge it whenever you are ready (i.e. removing the test commit).

Ambrevar:
The change does not fix anything, but it simplifies the `nyxt-quicklisp` system a bit.  So hopefully it will be more robust.

In particular, these 2 chunks are no longer significant:

```lisp
                       ;; When running `ql:update-dist' from an .asd system
                        ;; that's not in an ASDF registry, it will complain that
                        ;; it cannot find the "nyxt" system which is required by
                        ;; "nyxt/quicklisp".  So we must reload the .asd before
                        ;; running `ql:update-dist'.
                        (asdf:load-asd (asdf:system-source-file c)
```

```
                              ;; Register this directory so that nyxt.asd is
                               ;; included.  It's important since
                               ;; `ql:update-dist' may clear the ASDF
                               ;; configuration and this file could be forgotten
                               ;; if it was loaded with `asdf:load-asd'.
                               (namestring (uiop:truenamize (uiop:pathname-directory-pathname (asdf:system-source-file c))
```

It's 2 5-line long comments we can get rid of :)

Ambrevar:
Merged with 3ed976b1ecc16fe8977b94ba75cca254c4fedbfd.

jmercouris:
Merci monsieur!

Issue #1133, opened 2021-02-08
Hover option for follow-hint


g1smo:
It would be useful to have a follow-hint that hovers instead of clicking elements; some websites have menus that need to be hovered to click sub-entries.

Additionally (this may belong to a separate issue?), I've noticed that some clickable elements are not hinted. An example is the file/folder list view on nextcloud or the feature list at https://nyxt.atlas.engineer/features

That is, the features cannot be toggled using the keyboard; vimium extension for chromium/firefox properly hints them.

Version: Nyxt version 2-pre-release-5 on debian testing

aartaka:
Yes, we don't (yet!) hint the `<details>` tags. Thanks for reporting!

aartaka:
Regarding hoverable elements, I'm not that sure they are even actionable from keyboard. Does Vimium work with these?

ivangrozni:
Nope, vimium doesn't work with hoverable elements.

g1smo:
Sorry I was a bit unclear in my issue; one thing is hintable details and another hover hinting.
As ivan mentioned, no, vimium does not support it. It's a feature that would be very nice but I haven't seen elsewhere. Not sure how technically feasible it is, either.

aartaka:
Alright, we now hint `<details>` tags and have a hovering action. Closing this.

Ambrevar:
I tried to hover over the "Princing" menu on GitHub top bar and got
this:

The value
  "true"
is not of type
  LIST
   [Condition of type TYPE-ERROR]

Restarts:
 0: [ABORT] abort thread (#<THREAD "Nyxt anonymous thread" RUNNING {100B412333}>)

Backtrace:
 0: ((LAMBDA (NYXT/WEB-MODE::RESULTS) :IN NYXT/WEB-MODE:FOLLOW-HINT) "true")
 1: (NYXT/WEB-MODE:FOLLOW-HINT)
 2: ((LAMBDA NIL :IN NYXT::RUN-ASYNC))
 3: ((LAMBDA NIL :IN BORDEAUX-THREADS::BINDING-DEFAULT-SPECIALS))

aartaka:
The error is fixed now (see commits b5e8b0eceb2d565b3838262128e10189e6cad88e and 388423ed83aeae7dca6af689413fb9a3086da918), but the hovering is still unstable. I'll look into it tomorrow.

aadcg:
Is this still an issue @aartaka?

[...]

Issue #1290, opened 2021-04-07
Faster hint-generating function


wsgac:
I've recently been working on a simple hint generator for Eww and came up with a closure-based solution ([here](https://gist.github.com/wsgac/8893149c55d3e2684d87b622418be6bb)). I thought I'd have a look at how this is done in Nyxt. So I took a dive, refreshed my Parenscript slightly and found the relevant bit. I was curious which approach was faster and so I extracted the Nyxt code, converted it to pure CL and timed both on identical data. On a repeated sample of generating 250 hints my code turned out to be 3.5-4x faster. 

I don't know how important you deem performance in this area (keyboard-driven link selection is high on my personal priority list and among my favorite features of Nyxt). If you think it's worthwhile, I can prepare a PR incorporating my mechanism. The only thing I found troublesome is translating my code to Parenscript, so it would probably have to be called from within `(ps:lisp)`.

I'm attaching my code below, together with a usage example:
```
(defun gen-lexicographic-sequence (&key (starting-sequence "a")
				     (uppercase t))
  "Return a lambda for generating sequential, lexicographically ordered
  strings. When STARTING-SEQUENCE is provided, start with that, otherwise use
  \"a\" as a starting point. UPPERCASE controls the case of the sequences. The
  lambda produced by this function is meant to be invoked repetedly with
  FUNCALL."
  (labels ((string-to-dec (str)
	     (reduce #'(lambda (acc d) (+ (* acc 26) d))
		     (mapcar #'(lambda (c) (- (char-code c) (char-code #\a) -1))
			     (coerce str 'list))
		     :initial-value 0))
	   (dec-to-string (dec)
	     (loop
	       with dec = dec
	       with digit = nil
	       with chars = ()
	       while (plusp dec)
	       do (multiple-value-setq (dec digit)
		    (truncate (1- dec) 26))
	       do (push (code-char (+ (char-code #\a) digit)) chars)
	       finally (return (coerce chars 'string)))))
    (let ((current (string-to-dec (string-downcase starting-sequence))))
      #'(lambda ()
	  (prog1 (funcall (if uppercase #'string-upcase #'string-downcase)
			  (dec-to-string current))
	    (incf current))))))

;; Usage
(loop
 repeat 26
 with gen = (gen-lexicographic-sequence :starting-sequence "AB")
 collect (funcall gen))
[...]

Issue #1204, opened 2021-03-10
Search-buffer performance


aadcg:
Say we're searching in a web buffer with tons of lines. Since that requires running JS, I understand that there's not much to do.  

But perhaps in those cases, we could treat the source of the page as text so that the search will be faster? There might be some overhead into fetching the text but it should compensate nonetheless. 

I bumped into this while [here](https://packages.ubuntu.com/focal/allpackages?format=txt.gz).

aadcg:
Take a second look at this after #2529.

PR #1365, opened 2021-05-05
Plump DOM

### LINKED ISSUE #218: Support for drop-down buttons (<select> HTML tags) in element hinting
Hello! Great idea with emacs like browser.
Do you think about adding navigation with keys by inputs, as it was in Vimperator and Surfingkeys, for example?
In NEXT:
image
In Surfingkeys:
image

### LINKED ISSUE #1029: Support <select> HTML tags with follow-hint
`<select>` elements (i.e. dropdown menus) are not supported by `follow-hint` as of 2ed878cf85e0c68405e6dfc82ee992e12d8d40a2, so one needs mouse to fill these elements in.

Identifying these in `follow-hint` will bring us closer to mouse-less future :star_struck: 

Adding to that, using minibuffer to filter dropdown elements can be a cool feature in terms of Accessibility (if we are ever to focus on that) and convenient interfaces. Imagine: incrementally searching a 200-element dropdown (e.g. list of countries) with minibuffer instead of mouse-scrolling it...

### LINKED ISSUE #1295: Leverage Plump for link hinting & virtual DOM?

aartaka:
This is a terrible work in progress just to show out how Plump DOM can make our element-hints almost JS-less. Additionally, out own DOM means that we can learn much more about the displayed page, and modify it in way not possible (or quite non-trivial) before.

# What's New
- dom.lisp file with lots of auto-generated classes (`a-element`, `li-element` etc.). To be extended with methods on these.
- `document-model` slot of `web-mode` that we keep the parsed page model inside of.
  - `update-document-model` function to refresh it.
- Finally, element-hints rewritten in pure CL with occasional Parenscript isolated to small functions!
  - Much richer `prompter:object-properties`.
  - dispatching on auto-generated classes, like `nyxt/dom:a-element` for anchors.
  - Better handling of different types of `<input>`s.
  - `<details>` and `<select>` hinting.
- Jump to headings rewritten with Plump and CLSS. Looks much cleaner.

# What's Not (Yet?) There
- `clickable-hint` and `focusable-hint` are not there, maybe temporarily. Maybe rather introduce a multiple actions system on hintables?
- `annotate-visible-only-p` keyword argument is not there either. It won't get back, as it complicates lots of things, while existing for the sole performance purpose (and performance is a thing we can fix with this patch, I believe).

How does that look?

Closes #1295.
Closes #1029.
Closes #218.
Half of the solution for #1133.

EDIT: Mention jump to headings.
EDIT: Mention `<input>`, `<details>` and `<select>` improvements.
EDIT: Remove stale todos and rephrase `annotate-visible-only-p` mention -- it won't get back.
EDIT: Reference related issues, mention multiple actions on hintable elements.
EDIT: #218.


aadcg:
Could potentially solve #1204.

aartaka:
While I haven't yet fixed element hints (CI hints about that :D), I've tested Plump DOM on `jump-to-heading` and it makes code cleaner there! No slowdowns either. The only piece of JS explicitly used is a three-lines `define-parenscript` for universal page scrolling (i.e., it scrolls to things that are not necessarily headers!)

aartaka:
> Could potentially solve #1204.

Indeed it can. I've added an `inner-text` method on `plump:node` that allows to get the element text as an easily searchable string!

aartaka:
Element hints work! I've sped up the hinting function in ffa5d0983416be7b5e1aad390c521be07a8c1b6f, so the main bottleneck is gone. Plump version of element hinting is (seemingly) as fast as original, while being simpler and lispier. Win-win :)

Things left ot do:
- Ensure that `document-model` is fresh (this made me to add `when element` to `hint-elements` and I want to get rid of it).
- Use the new'n'shiny mechanism in `visual-mode` (I believe we can save lots of lines there!)
- Utilize #1290.
- Test, test, test.
- Benchmark (not that necessary now that the main performance problem is gone.)

Ambrevar:
Exciting stuff, congrats!
I'll have a look (hopefully) soon!

aartaka:
Okay, `visual-mode` re-uses what element hints have and this cuts more than 100 lines of code :)

aartaka:
# What's new:
Good:
+ Support for `<select>` and `<details>` tags.
+ Minor refactoring.

Bad:
- Page search is broken.

aartaka:
I really want to try rewriting buffer search using Plump DOM :upside_down_face: 

But the only actual breakage is `highlight-selected-hint` and that is trivial to fix. I'll push the fix in a minute and it should be more or less reviewable, unless I find more bugs :)

aartaka:
Search is fixed, should be testable.

aartaka:
Be wary that Plump-based hinting is still somewhat slower than JS-based one. It could be sped up, it's just that I haven't yet gotten myself to profile it :)

aartaka:
Small note on what we've come to while on the call with @jmercouris:
- We can merge some parts right now, as they are immediately usable, work well and are possibly better than the existing code. These parts are:
  - DOM library (dom.lisp, web-mode.lisp)
  - Jump to heading functions, as they work fast enough, can be extended with ease (using `plump:next-sibling` on a header element will most probably return us the paragraph following the header that can be listed in the prompt-buffer to ease search and comprehension), and are much cleaner with Plump and CLSS (Plump companion library allowing to select DOM elements with CSS selectors, like `"h1, h2, h3"` and `"[nyxt-identifier=\"1\"]"`).
- Some parts are imperfect (`visual-mode`, element hints), so these can wait.
- The biggest performance bottleneck in the case of element hints is fetching the page source (takes approximately 0.1 second out of the total 0.17 second for `add-element-hints` call). All in all, it's 5 times slower than JS version of `add-element-hints`, but JS version does not return Lisp-native objects and does not do parsing, so the comparison is uneven.

Ambrevar:
Can we merge after 2.0?  This is quite a big change and I'm afraid it
will introduce its load of maintenance work that would be quite a burden
right now.

Other than that, congrats on this behemoth achievement!

jmercouris:
That is correct, this is slated for post 2.0.

jmercouris:
At least per my understanding.

aartaka:
I believe that's it for the patch. Element hinting is fast enough (still somewhat slower than JS version), but it much simpler conceptually, thus reusable and readable.

I've tried implementing search on top of new DOM, but failed. Translating the existing JavaScript implementation is impossible due to its complexity (this piece of code seems to have a critical bus-factor). Rewriting search from scratch is an option, but that's too much for this patch.

I've also tried adding classes for [WAI-ARIA roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) to enhance the custom controls hinting, but it occured to be less trivial than I expected (reason: It's hard to come up with a simple and false-positive-less selector encompassing all of the necessary roles to hint). This needs a separate design, but can still be implemented on top of Plump DOM (my incomplete attempts are in 2a824d462d84400a673c7c4f1fba6d8b171478b5).

Merge whenever.

aartaka:
As much as I love this patch and hope to use it, it's a big regression. `visual-mode` reports only 44 text pieces on https://reactjs.org, compared to 398 reported by the current implementation   **o(TヘTo)**

This is not practical. 

The plan is:
- I close this and move hinting improvements to master.
- I'll most probably extract the parsing and tagging part to an extension. Something like `nx-dom`. Feels natural.
- If this extension (somehow) accumulates enough power to be better than JS implementation (maybe actually traverse the DOM with JS and collect all the attributes and tags **(¬‿¬)**?), then it could replace the JS-native implementation.

Ambrevar:
Too bad, I was really looking forward to it :(

Let's not waste this precious (and gigantic) patchset, so +1 on moving
this to a nx-dom extension.

> - I close this and move hinting improvements to master.

Which hinting improvements exactly?

aartaka:
> Too bad, I was really looking forward to it :( Let's not waste this precious (and gigantic) patchset, so +1 on moving this to a nx-dom extension.

Yeah, that's the plan.

> - I close this and move hinting improvements to master.
> Which hinting improvements exactly?

`<details>` and `<select>` hinting, better `prompter:object-attributes` for hints.

Ambrevar:
Roget that! :)

aartaka:
Good news! d7a6911451782eb219799d57b2f93e0e3904d61a enables JS-based page scraping algorithm. It's reasonable fast (although needs more testing) and produces desirable results.

As a benchmark, https://reactjs.org now has 400+ `visual-mode` hints!

Ambrevar:
I get 26 hints on Nyxt 2.0.0...
O.o

aartaka:
OK, a small disillusionment: `clss:select` was returning around 3-4 duplicates for each element, thus hinting was misreporting the number of hinted elements. Now that is fixed (well, the fix is not satisfactory -- we need to rebuild the DOM, while we're just do `remove-duplicates`), it's around 70-80 elements on the same React page.

Moreover, if we add `<span>` elements to the `visual-mode`-hintable ones, it jumps to several hundred!

All in all, this seems to be fully functional (except minor annoyances, like page not scrolling to the selected element and unsatisfactory fix to the problem of duplicates).

source/dom.lisp
```diff
@@ -0,0 +1,164 @@
...
+                     (cond
+                       ((string-equal (alex:assoc-value json-alist :name) "#text")
+                        (plump:make-text-node parent (alex:assoc-value json-alist :text)))
+                       ((string-equal (alex:assoc-value json-alist :name) "#comment")
+                        (plump:make-comment parent (alex:assoc-value json-alist :text)))
+                       (t (plump:make-element parent (alex:assoc-value json-alist :name))))))
+               (when (typep element 'plump:nesting-node)
+                 (setf (plump:children element)
+                       (plump:ensure-child-array
+                        (map 'vector (alex:rcurry #'json-to-plump parent)
```
aartaka, line 148: There should be an `element` instead of `parent`! This causes the duplicates problem!

aartaka:
OK, branch is as complete as it can be. I've rebased over master and it started crashing, though.

https://www.reddit.com is the URL that always crashes it for me (both with config and without it). It crashes even before loading finishes. I'm totally clueless why this branch breakes it, while master branch doesn't -- they are absolutely identical in what happens before page loading and a minor difference in what happens after (but it doesn't matter -- crashes happen before loading).

jmercouris:
Are you starting Nyxt with the same parameters in both cases?

aartaka:
Oh wait...

It crashes in dev environment (`nyxt-make-guix-sbcl-for-nyxt` from nyxt/build-scripts/nyxt-guix.el), but works all fine in a compiled version. Weird.

aartaka:
I've discovered some under-hinting on GitHub, will try to look into it tomorrow.

Ambrevar:
> It crashes in dev environment (`nyxt-make-guix-sbcl-for-nyxt` from nyxt/build-scripts/nyxt-guix.el), but works all fine in a compiled version. Weird.

Try regenerating the image.

aartaka:
> It crashes in dev environment (`nyxt-make-guix-sbcl-for-nyxt` from nyxt/build-scripts/nyxt-guix.el), but works all fine in a compiled version. Weird.
> Try regenerating the image.

That didn't help 0_o

aartaka:
Alright, the annoyance with under-hinting is solved by 219f1fb5d1868fba04df316d30504fad93065907. Now the document-model should be updated whenever it goes stale. The branch should be fully testable and shining.

Backwards-compatibility checks:
- [X] `allow-visible-only-p`.
- [ ] Image hinting (it was quite interrupting and Vimium doesn't do it, maybe we don't need it?)
- [ ] Focusable and clickable elements hinting (now that we hint most HTML actionable elements, it should be unnecessary).
- [X] No functions deleted, only rewritings and additions.

All in all, I can now use GitHub and other websites without resorting to mouse. Yay! Once we add ARIA roles support, it should be as complete as Vimium etc.

aartaka:
ARIA roles can wait until later, though. This branch is mergeable, I believe.

Ambrevar:
Fantastic, thanks for all the work!

> - [ ] Image hinting (it was quite interrupting and Vimium doesn't do it, maybe we don't need it?)

Image hinting can be important, say, if you want to download a picture
gallery.

Why is it interrupting in your experience?

We could also have a convenient way to select which _type_ of elements
to hint.

> - [ ] Focusable and clickable elements hinting (now that we hint most HTML actionable elements, it should be unnecessary).

I don't understand, why is it unnecessary?

aartaka:
> Fantastic, thanks for all the work!
> - [ ] Image hinting (it was quite interrupting and Vimium doesn't do it, maybe we don't need it?)
> Image hinting can be important, say, if you want to download a picture gallery. Why is it interrupting in your experience? 

- It's typical that links are augmented by images for visuality. In this case, both link and image have the same bounding box and their hints are ambiguously close.
- Images are often intended to be clicked for custom actions, instead of being opened directly. We don't want to interfere with this.
- Neither Trydactil nor Vimium do that (probably for reasons above). See #1400 for what selectors they use there.

> We could also have a convenient way to select which _type_ of elements to hint.
> - [ ] Focusable and clickable elements hinting (now that we hint most HTML actionable elements, it should be unnecessary).
> I don't understand, why is it unnecessary?

Well, in the ideal world, semantic elements of HTML should be enough for hinting... But, on the second thought, those are necessary. Trydactil has some clickable attributes in its selectors too.

Ambrevar:
>> Fantastic, thanks for all the work!
>> - [ ] Image hinting (it was quite interrupting and Vimium doesn't do it, maybe we don't need it?)
>> Image hinting can be important, say, if you want to download a picture gallery. Why is it interrupting in your experience? 
>
> - It's typical that links are augmented by images for visuality. In this case, both link and image have the same bounding box and their hints are ambiguously close.
> - Images are often intended to be clicked for custom actions, instead of being opened directly. We don't want to interfere with this.
> - Neither Trydactil nor Vimium do that (probably for reasons above). See #1400 for what selectors they use there.

What about only hinting images bigger than some threshold?

aartaka:
How about we merge this and I implement clickables' support on top of it in July?

aartaka:
> What about only hinting images bigger than some threshold?

Not an option either, because linked images can be big too, same for images with intended clicking actions.

Ambrevar:
> Not an option either, because linked images can be big too, same for images with intended clicking actions.

I'm confused about the problem.  Why wouldn't we want to hint at big
images?
Are you saying there is an ambiguity because an image can have two
actions: display image or follow link?

aartaka:
> Not an option either, because linked images can be big too, same for images with intended clicking actions.
> I'm confused about the problem. Why wouldn't we want to hint at big images? Are you saying there is an ambiguity because an image can have two actions: display image or follow link?

Yes.

Ambrevar:
What about drawing a different looking hint?

aartaka:
Not sure about different looking hint.

As an alternative course of action, we can define default action for `%follow-hint` to be clicking. This way, images are covered, but aren't hi-jacked with the action that wasn't indended for them initially.

Ambrevar:
The default action should be "do-what-I-mean" on pictures, no?
Because not all pictures are links.  Do a do-what-I-mean behaviour would
be:

- If it's a link, follow it.
- It no, open the picture.

aartaka:
I don't agree there. Given that keyboard navigation is an alternative to mouse navigation, we should follow the patterns that website developers established for pictures. 

Patrerns are usually:
- Doing nothing on image click (default HTML behavior).
- Doing a custom `onclick` action.

And that's basically it. To support this, we need to click the image as the default action. Everything else is additional actions.

Ambrevar:
OK, good point.

aartaka:
# What's new:
+ Image hinting (with click as default action).
+ Several actions for hints.
+ Minor refactoring.
+ An addition of `url` and `body` methods of hints for backwards-compatibility.
However...
- There's a major slowdown (1+ second) happening of pages with 1000+ hints. Interestingly, it happens on the side of prompt-buffer, because `ensure-suggestions-list` chokes of long lists. @Ambrevar, can we speed `ensure-suggestions-list` up or somehow avoid calling it?

Ambrevar:
We definitely can, although it shouldn't choke just on 1000 items.  Do you have a recipe so that I can witness this myself?

aartaka:
The recipe is: 
- build this branch,
- put sticker on `(ensure-suggestions-list ...)` form on line 492 in libraries/prompter/prompter-source.lisp,
- run Nyxt with breaking on stickers, 
- and use `follow-hint` on sufficiently big page (like the page of this PR -- there are 2081 hints there with all the images).

Should break on the sticker, and, after you press continue, it'll hang for a second or two.

Ambrevar:
OK, I could reproduce and narrowed it down to `img-element' `url'
method.

`valid-url-p` turns out to be slow because it calls to
`iolib/sockets:lookup-hostname`.

I could work around this by memoizing the function this way:

```lisp
(defmacro defmemo (name params &body body)
  (alex:with-gensyms (memo-table args result result?)
    `(let ((,memo-table (make-hash-table :test 'equal)))
       (defun ,name (&rest ,args)
         (multiple-value-bind (,result ,result?)
             (gethash ,args ,memo-table)
           (if ,result?
               ,result
               (setf (gethash ,args ,memo-table)
                     (apply (lambda ,params
                              ***@***.***)
                            ,args))))))))

(defmemo lookup-hostname (name)
  "Memoized verison of `iolib/sockets:lookup-hostname'."
  (iolib/sockets:lookup-hostname name))
```

Does it work for you?

Ambrevar:
See https://github.com/atlas-engineer/nyxt/pull/1605.

aartaka:
Yes, that should do the trick!

source/web-mode.lisp
```diff
@@ -187,6 +192,50 @@ and to index the top of the page.")
...
+         (ps:chain node (set-attribute "nyxt-identifier" (ps:stringify nyxt-identifier-counter))))
+       (incf nyxt-identifier-counter)
+       (dolist (child (ps:chain node children))
+         (add-nyxt-identifiers child))
+       nyxt-identifier-counter)
+     (setf nyxt-identifier-counter (add-nyxt-identifiers (ps:chain document body)))))
+  (setf (document-model mode)
+        (nyxt/dom::named-json-parse (nyxt/dom::get-document-body-json))))
+
+(define-parenscript %count-dom-elements ()
```
jmercouris, line 212: Only used in the command `count-dom-elements`, hints that it should probably be a pflet :-)

aartaka:
- `%count-dom-elements` is a `pflet` now.
- @jmercouris, the bug with Wikipedia search field is fixed.
- Images with `alt=""` are excluded from the `hint-source`, as these are usually intended to be non-interactive and purely decorative.
- Still, some pages have slow-ish hints even though memoization. Maybe we don't even need to check URL validity for images?

All in all, this patch is complete. Feel free to merge (after giving it a test, of course)!

jmercouris:
OK! I will do some more testing, thank you Artyom!

aartaka:
Another optimization: image URLs are not checked for validity anymore, as it's not necessary (I wonder why I've added it in the first place).

aartaka:
Final optimization: I've profiled the time lags and found that `add-element-hints` was a bit too slow because of huge JavaScript calls. I've made separate JS calls to run on separate threads, and it cut the time lag to being almost unnoticeable. Please look at it and tell me how fishy it is! 

Other than that, if feels more merge-ready than ever.

jmercouris:
Thank you for the updates Artyom!

source/web-mode.lisp
```diff
@@ -187,6 +192,47 @@ and to index the top of the page.")
...
+           (defun count-dom-elements (node)
+             (incf dom-counter)
+             (dolist (child (ps:chain node children))
+               (count-dom-elements child))
+             dom-counter)
+           (setf dom-counter 0)
+           (count-dom-elements (nyxt/ps:qs document "html"))))
+    (let ((value (call-next-method))
+          (element-count (parse-integer (%count-dom-elements) :junk-allowed t)))
+      (if (and value
```
jmercouris, line 225: 11 I like this

jmercouris: "significant" is magically 10 :-D

aartaka: Had no better value to pick there. It was totally arbitrary. We need to do a statistical research there :)

jmercouris: I have a better way to write this, will push :-)

Issue #1080, opened 2020-12-22
The nx-freestance does not work with the new deb file


danrobi11:
[nx-freestance-handler](https://github.com/kssytsrk/nx-freestance-handler) does not work with the new deb file version 2-pre-release-5. 
Bummer cuz I wanted to promote the new nyxt release / the new nx-freestance in a few subreddit.
The nx-freestance is working fine when compiling nyxt from source.
Which makes me think that the deb version by default might not be listening to whats in the ```common.lisp``` directory. 

Is it possible to manually add a few lines of code somewhere to create a sort of work around for new users that would like to install the deb file, instead of having to compile nyxt from source in order to run the nx-freestance-handler?

Although, thank you much for the new deb file release!

P.S Possible to add a note somewhere about the nx-freestance-handler so new users could know about it? In the readme maybe? idk.

aartaka:
Regarding the note about the nx-freestance-handler: It's already there, in [EXTENSIONS.org](https://github.com/atlas-engineer/nyxt/blob/master/documents/EXTENSIONS.org) :heavy_check_mark: 

Regarding extension compatibility: A workaround could be a good old 
``` lisp
(push #P"/path/to/nx-freestance-handler/" asdf:*central-registry*)
```

But I'm not fully aware of how either of nx-freestance-handler or Nyxt extension handling works, so @Ambrevar might be in a better position to help you :)

danrobi11:
> It's already there, in EXTENSIONS.org heavy_check_mark

Nice!
I've tried adding ```(push #P"/home/danrobi/common-lisp/nx-freestance-handler/" asdf:*central-registry*)``` in my ```init.lisp``` and unfortunately that does not work.

jmercouris:
@danrobi11 thank you for your interest in Nyxt!

You can try `(asdf:load-asd "/path/to/asd/file")` then you should be able to load the system with `(asdf:load-system :system-name)`.

Of course substituting the above with your correct information.

jmercouris:
If it doesn't work in your init file, please try it in the `lisp-repl` and report any errors/messages in `list-messages`. Thank you!

danrobi11:
@jmercouris 
With ```(asdf:load-asd "/home/danrobi/common-lisp/nx-freestance-handler")``` I get this warning in nyxt:
```
[...]

PR #1366, opened 2021-05-05
Allow Nyxt to load extensions

jmercouris:
### Synopsis
I've added the following to the end of `start.lisp`:

```
(map () 'asdf:register-preloaded-system (asdf:already-loaded-systems))
(map () 'asdf:register-immutable-system (asdf:already-loaded-systems))
```

This prevents attempted recompilation of already loaded systems in an image of Nyxt. This is critical for loading on NixOS. It prevents ASDF from attempting to recompile systems that it cannot (they are located in non-writable directories).

This therefore allows you to do:

```
(asdf:load-asd "/home/jmercouris/Source/Lisp/nx-kaomoji/nx-kaomoji.asd")
(asdf:load-system :nx-kaomoji :force-not t)
```

notice the key argument of `:force-not`. This is important. You can read more about it in the ASDF source code. Here is the documentation from `asdf::*immutable-systems*`:

```
"A hash-set (equal hash-table mapping keys to T) of systems that are immutable,
i.e. already loaded in memory and not to be refreshed from the filesystem.
They will be treated specially by find-system, and passed as :force-not argument to make-plan.

For instance, to can deliver an image with many systems precompiled, that *will not* check the
filesystem for them every time a user loads an extension, what more risk a problematic upgrade
 or catastrophic downgrade, before you dump an image, you may use:
   (map () 'asdf:register-immutable-system (asdf:already-loaded-systems))

Note that direct access to this variable from outside ASDF is not supported.
Please call REGISTER-IMMUTABLE-SYSTEM to add new immutable systems, and
contact maintainers if you need a stable API to do more than that."
```

### Questions
Please let me know if this also works for Guix. I believe it should also alleviate any problems on Ubuntu as well, though I do not have a deb to test against.

Also, please let me know if you can think of a better place to add (perhaps it belongs somewhere in the asd?):

```
(map () 'asdf:register-preloaded-system (asdf:already-loaded-systems))
(map () 'asdf:register-immutable-system (asdf:already-loaded-systems))
```

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -311,6 +311,11 @@
                      :executable t
                      :compression (uiop:getenv "NYXT_COMPRESS"))))
 
+(defmethod perform :before ((o image-op) (c system))
+  "Register immutable systems to prevent compiled images of Nyxt from
+trying to recompile dependencies"
+  (map () 'asdf:register-immutable-system (asdf:already-loaded-systems)))
+
 (defsystem "nyxt/install"
   :depends-on (alexandria
                str
--- a/.github/workflows/package-ubuntu.yml
+++ b/.github/workflows/package-ubuntu.yml
@@ -2,9 +2,9 @@ name: Build .deb package for latest Ubuntu
 
 on:
   push:
-    tags: '*'
+    # tags: '*'
     ## To test, uncomment the line below with branch name used for review.
-    # branches: [ MY-WORK-BRANCH ]
+    branches: [ nx-load ]
 
 jobs:
   build:


jmercouris:
If you complete this quest, the nectar that awaits you at the end of this journey is unfettered access to `nx-kaomoji`. I can think of no sweeter prize.

Screenshot_20210505_172755

Ambrevar:
Indeed the code belongs to the dump hook I think, as mentioned in the documentation.

For instance we could set `uiop:*image-dump-hook*` to this in the `nyxt.asd` file.
Needs some testing.

This could be a solution for Nix.
The problem with the Guix package is different if I recall what @aartaka reported.  I'll investigate on that end.

source/start.lisp
```diff
@@ -532,3 +532,6 @@ Finally,run the `*after-init-hook*'."
 (define-command nyxt-init-time ()
   "Return the duration of Nyxt initialization."
   (echo "~,2f seconds" (slot-value *browser* 'init-time)))
+
+(map () 'asdf:register-preloaded-system (asdf:already-loaded-systems))
+(map () 'asdf:register-immutable-system (asdf:already-loaded-systems))
```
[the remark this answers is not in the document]
jmercouris: I'm just using the example provided by ASDF, not sure why they wrote it that way.

jmercouris:
In any case, I think this is important. Compiled images of Nyxt should not be recompiling already loaded dependencies!

Ambrevar:
Instead of the hook, we could also add a `:before` method to `image-op`'s `perform`.  See what we have already for an example:

```lisp
#+sb-core-compression
(handler-bind ((warning #'muffle-warning))
  (defmethod perform ((o image-op) (c system))
    (uiop:dump-image (output-file o c)
                     :executable t
                     :compression (uiop:getenv "NYXT_COMPRESS"))))
```

jmercouris:
OK, I'll try that.

jmercouris:
Well it seems that works: (-‿◦☀)

jmercouris:
we are now ready to rock: `ʕ•̫͡•ʔ♬✧`

nyxt.asd
```diff
@@ -311,6 +311,12 @@
                      :executable t
                      :compression (uiop:getenv "NYXT_COMPRESS"))))
 
+(defmethod perform :before ((o image-op) (c system))
+  ;; Register preloaded/immutable systems to prevent compiled images
+  ;; of Nyxt from trying to recompile dependencies
```
Ambrevar, line 316: Cool!
Can you turn the comment into a docstring, and add a reference to the ASDF docstring / documentation where you found the code?

Ambrevar:
I'll test on Guix.
I'll also look at how StumpWM does it.

Ambrevar:
For past references, see #1080.

Ambrevar:
In the meantime, can you trigger the deb build on this branch and test it?

Ambrevar:
Nit: Please squash the commits :)

Ambrevar:
Ma, look what I found! :)

https://github.com/stumpwm/stumpwm/blob/master/make-image.lisp.in

Ambrevar:
Why call asdf:register-preloaded-system?
It's not in the doc, nor used by StumpWM.  I suggest to remove it.

Ambrevar:
It works in Guix!

If I'm not mistaken, all that remains to be test is with Ubuntu.
The results could be slightly different because of the use of
linux-packaging.
Fingers crossed.

Should I generate the .deb?

jmercouris:
Please generate the .deb, I'm not sure how to trigger the artifact generation.

nyxt.asd
jmercouris: I can't do that, ASDF is a single file. I can make it a docstring and squash the commits.

jmercouris:
> Why call asdf:register-preloaded-system? It's not in the doc, nor used by StumpWM. I suggest to remove it.

I was advised to do it, and it makes things work on NixOS for some reason :-)

Ambrevar:
>> Why call asdf:register-preloaded-system? It's not in the doc, nor used by StumpWM. I suggest to remove it.
>
> I was advised to do it, and it makes things work on NixOS for some reason :-)

Can you confirm it does not work without it?
Then please add a comment explaining that it's required for NixOS and
that it needs more investigation.

jmercouris:
I tried without it, and it did not work, but I will try again :-D

nyxt.asd
Ambrevar: Can't do what?  Refer to the asdf docstring?  Of course, just write `asdf::*immutable-systems*` in the perform docstring.

Ambrevar:
Thanks!

Ambrevar:
I'll trigger the .deb build, don't push anything.

jmercouris:
No idea what changed, perhaps there was some state in emacs, but you are correct, that is indeed not necessary...!

jmercouris:
OK, I won't push anything.

Ambrevar:
See https://github.com/atlas-engineer/nyxt/pull/1366/commits/bdaef5f38a000b700f293cdad3b87cb7951e3e1b for how to trigger the .deb build.

jmercouris:
`(>^_^)><(^o^<)`

jmercouris:
I see, thanks!

jmercouris:
Works on Ubuntu!

Screenshot_20210505_212534

jmercouris:
Merged with 5386223c

Ambrevar:
Fantastic!  And thanks again for taking the pain to deal with the VM!

jmercouris:
No problem!

aartaka:
It work! Thanks John! `＼（＠￣∇￣＠）／`

danrobi11:
@jmercouris @Ambrevar  ❤️

jmercouris:
Thank you :-)

PR #1371, opened 2021-05-06
Actions fix

jmercouris:
What do you think?

### DIFF
--- a/source/file-manager.lisp
+++ b/source/file-manager.lisp
@@ -37,6 +37,8 @@ It's suitable for `prompter:filter-preprocessor'."
 
 (define-class file-source (prompter:source)
   ((prompter:name "Files")
+   (prompter:actions nil :documentation "Actions will be calculated/populated in
+   initialize-instance :after.")
    (prompter:filter-preprocessor (make-file-source-preprocessor))
    (prompter:multi-selection-p t)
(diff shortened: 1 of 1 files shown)


aartaka:
That seems like an absolute overkill. What stops us from extracting `open-file*` to a top-level function instead?

Ambrevar:
Is something broken?
@jmercouris I don't understand the intention.

Ambrevar:
@aartaka We could extract `open-file*`, but not that actions don't take a source as argument, so you'd still need the initialize-instance method and to curry the source into the extracted function.

jmercouris:
@Ambrevar that's exactly the problem I identified as well, hence my proposed solution. I just force pushed, please take a look.

jmercouris:
@Ambrevar Yes, something was broken. Editor open file would constantly open up libre office instead of actually opening the built-in editor. It turns out that the `initialize-instance :after` of the `file-source` was changing the `actions`.

aartaka:
Looks alright. Maybe call it `openable-file-source` or `open-file-source`? Also, it should be used in `download-open-file`.

aartaka:
@Ambrevar, maybe we should add a `(setf actions)` method to mirror `actions`?

aartaka:
Thanks John! That was quick! :)

jmercouris:
No problem!

aartaka:
> @aartaka You mean a setf-method?

Exactly.

PR #1372, opened 2021-05-06
start/configuration: Move socket and init expansion logic to expand-d…

Ambrevar:
…ata-path.

There is no good reason *socket-path*, *init-path* and *auto-config-file-path*
should have a special setup local to the `start` function.  Because of this
mistake, --init was left unhandled for months :/

So make things right and consistent again!

WARNING: Needs testing!

### DIFF
--- a/source/configuration.lisp
+++ b/source/configuration.lisp
@@ -6,17 +6,37 @@
 (define-class init-data-path (data-path)
   ()
   (:export-class-name-p t)
-  (:export-accessor-names-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
+
+(define-class auto-init-data-path (init-data-path)
+  ()
+  (:export-class-name-p t)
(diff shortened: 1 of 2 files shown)


Ambrevar:
@jmercouris @aartaka Thoughts on this?

jmercouris:
I'm not sure I understand what this has achieved. Can you please explain?

Ambrevar:
Consistency.

Instead of doing the path expansion logic at an arbitrary point in the
program (the `start` function), we do it in `expand-data-path` as for
every other data-path.

The behaviour remains the same, it's just that
- it's shorter
- it's more consistent
- it's less fragile since the logic is factored out.

jmercouris:
I see, that makes sense. I strongly agree with this change.

jmercouris:
Feel free to merge this whenever you are ready.

Ambrevar:
OK, I'll test a bit more first.

jmercouris:
Thank you!

PR #1384, opened 2021-05-09
browser.lisp: remove ffi-prompt-buffer-evaluate-javascript

jmercouris:
This distinction between evaluating javascript within the prompt
buffer and within a regular buffer is no longer meaningful.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -602,8 +602,6 @@ sometimes yields the wrong result."
 (define-ffi-generic ffi-buffer-load (buffer url))
 (define-ffi-generic ffi-buffer-evaluate-javascript (buffer javascript))
 (define-ffi-generic ffi-buffer-evaluate-javascript-async (buffer javascript))
-(define-ffi-generic ffi-prompt-buffer-evaluate-javascript (window javascript))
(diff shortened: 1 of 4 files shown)


Ambrevar:
Looks good to me, but do you remember why we had the distinction in the first place?

Congrats on the 20-line clean up, always welcome :)

jmercouris:
The original rationale was that the callbacks were handled by us, rather than cl-webkit. They had distinct callback stacks to avoid any contamination of data/callbacks between regular buffers and prompt buffers. Additionally they were accessed via distinct ffi methods to create a forced isolation. We still have that isolation, but not because of distinct `ffi-prompt-buffer*` methods.

Ambrevar:
I remember something like this, but I'm not sure this was the main
reason.

Going back in time, e.g. at commit
19c673879539c408419439b9e2645cd32b21626c, we see that we have roughly
the same body for ffi-buffer-evaluate-javascript and
ffi-minibuffer-evaluate-javascript, the main difference being that the
former binds `%callback` while the other does not.

Anyways, this distinction seems irrelevant now that we have the -sync
and -async versions.

PR #1386, opened 2021-05-10
Remove notification.lisp

jmercouris:
needs better design, for example, different notification programs may
have different command arguments, this does /not/ consider that.

this will also allow us to remove `notify-osd`.


Ambrevar:
To fix the CI, I suggest you put `(echo-warning message)` in place of the remove `notify`.

Ambrevar:
I'm OK with the removal, it can be restored later more properly.

I don't use notifications though, so I may be biased.
Thoughts anyone?
@aartaka?

jmercouris:
We are only using them in a single place for now. When we restore them, we will restore them properly. We can add support for all kinds of notifications, page finished loading, download complete, etc.

aartaka:
> I'm OK with the removal, it can be restored later more properly.
> 
> I don't use notifications though, so I may be biased.
> Thoughts anyone?
> @aartaka?

Not using notifications either ¯\\\_(ツ)_/¯

PR #1391, opened 2021-05-11
Info buffer

jmercouris:
For post 2.0.0


Ambrevar:
Woooooow this is super cool!!! :D 

I wonder why `q` for an info buffer?! :p

The first thing I'd suggest is to rename the concept of "info buffer", as I don't find it obvious enough.
Besides, these buffers may not really carry "info", like here the list of buffers is hardly "information".

What about "side buffer"?

Another point: shall we first implement proper window management (binary splitting à-la Emacs), then these "info buffers" on top of them?  Intuitivement, I'd say that it would not require much more work while simplifying the info buffer creation + allowing for vertical prompt buffers.  Thoughts?

jmercouris:
> Woooooow this is super cool!!! :D

Thank you!

> I wonder why `q` for an info buffer?! :p

When developing in the REPL I pick single character defparameters to represent objects. so I will write something like `(defparameter x ...)`. Just a habit of mine :-)

> The first thing I'd suggest is to rename the concept of "info buffer", as I don't find it obvious enough.
> Besides, these buffers may not really carry "info", like here the list of buffers is hardly "information".

Well, I thought a long time about what to call them. The point is this, they are /additional/ information about a specific buffer, or window. They are like a little inspector. So in this case, since I've only implemented them on the window level, they provide additional information for the window.

> What about "side buffer"?

Too ambiguous. What do you think about `inspector-buffer`?

> Another point: shall we first implement proper window management (binary splitting à-la Emacs), then these "info buffers" on top of them? Intuitively, I'd say that it would not require much more work while simplifying the info buffer creation + allowing for vertical prompt buffers. Thoughts?

These are orthogonal concepts. I would like info buffers to remain outside of the window splitting framework. I think we will have to support both window splitting and info buffers. For example, we may have the following situation:

Buffer X which has info buffer 1, 2, and 3

when you set the active buffer to Buffer X, info buffers 1 2 and 3 appear or disappear. With traditional Emacs like window splititng, this is not possible. In other words, the window splitting treats a buffer + info buffers as a SINGLE UNIT.

Ambrevar:
> Well, I thought a long time about what to call them. The point is
> this, they are /additional/ information about a specific buffer, or
> window.

I'd like to go further: Why not generalize it?

We can go beyond adding information about a specific buffer, and simply
add general information.

For instance, we could very well display the weather :)

>> What about "side buffer"?
>
> Too ambiguous.

Why?

> What do you think about `inspector-buffer`?

We already have an inspector that's unrelated to this.

Also I don't think many users would make the link between the word
"inspector" and displaying the buffer list.

>> Another point: shall we first implement proper window management (binary splitting à-la Emacs), then these "info buffers" on top of them? Intuitively, I'd say that it would not require much more work while simplifying the info buffer creation + allowing for vertical prompt buffers. Thoughts?
>
> These are orthogonal concepts. I would like info buffers to remain outside of the window splitting framework. I think we will have to support both window splitting and info buffers. For example, we may have the following situation:
>
> Buffer X which has info buffer 1, 2, and 3
>
> when you set the active buffer to Buffer X, info buffers 1 2 and 3 appear or disappear. With traditional Emacs like window splititng, this is not possible. In other words, the window splitting treats a buffer + info buffers as a SINGLE UNIT.

Misunderstanding: I'm not talking about splitting web buffers, but
implementing the _window management logic_ at the FFI level, a bit
like you did but with methods / parameters to split in the other
direction as well, to shrink / enlarge, etc.

jmercouris:
> Well, I thought a long time about what to call them. The point is this, they are /additional/ information about a specific buffer, or window.
> I'd like to go further: Why not generalize it? We can go beyond adding information about a specific buffer, and simply add general information. For instance, we could very well display the weather :)

Sure, in that case it is info, no? Therefore `info-buffer` makes sense :-D

> > What about "side buffer"? Too ambiguous.
> Why?

Well, the status buffer is on the side of the active buffer, is it also a side buffer?

> Misunderstanding: I'm not talking about splitting web buffers, but implementing the _window management logic_ at the FFI level, a bit like you did but with methods / parameters to split in the other direction as well, to shrink / enlarge, etc.

Ah, I see. I think that is a very challenging problem. I would like to save that for a later time. I have some things in my head, but nothing concrete.

Ambrevar:

> Sure, in that case it is info, no? Therefore `info-buffer` makes sense
> :-D

Following this logic, everything is information.

That's the problem actually: "info" is too unspecific.

>> > What about "side buffer"? Too ambiguous.
>> Why?
>
> Well, the status buffer is on the side of the active buffer, is it also a side buffer?

Actually, could be, no?

Would be nice to have all elements of the UI customizable under a
unified interface.

>> Misunderstanding: I'm not talking about splitting web buffers, but implementing the _window management logic_ at the FFI level, a bit like you did but with methods / parameters to split in the other direction as well, to shrink / enlarge, etc.
>
> Ah, I see. I think that is a very challenging problem. I would like to save that for a later time. I have some things in my head, but nothing concrete.

I'm not sure yet because I haven't tried, but it might not be that
challenging.

I'd like to give it a try.

jmercouris:
OK, but I cannot hold this change back until you give it a try :-D that would be crazy!

edgar-vincent:
This is great! Thanks @jmercouris! I agree with @Ambrevar though, I find the name a bit vague. I quite like `side-buffer`. The `status-buffer` is also situated on the side, but its name conveys the fact that it has a specific purpose - `side-buffers` don't :)

jmercouris:
No problem monsieur!

Unfortunately I really do not like `side-buffer`, we'll have to think of something else :-D

Ambrevar:
I think we need to find a name that stands on its own on does not need
to have " buffer"  appended.

Indeed, so far we've often had a 1-to-1 mapping between buffers and web
views.

When we say "switch to buffers", we mean the web view buffer.

(By the way: we don't have a name for the web view buffers.
"View buffers" sounds weird.  "Core buffer"?)

The drawback with a term like "info buffer" (or "whatever buffer") is
that it would introduce an ambiguity between  the type of buffers and
the web view buffers.

For instance, we could later introduce a command "info" that shows a web
view buffer with system information.  Tadah!  An "info buffer"! :p



So what about a term that focuses on the UI aspect?
Quick suggestion: "panel".
I think it's a term commonly used in UIs, for instance Qt applications
often have side panels.

Thoughts?

jmercouris:
Panel could work. Why not call them panel-buffers then? Will we ever have a "panel" command?

Ambrevar:
Oh absolutely, what I meant is have the full name "X buffer" where "X"
would also make sense on its own.

So "panel buffer" is the full, technical name, but it's also fine to
refer to them as "panels", it makes sense in English.

jmercouris:
je suis d'accord!

Ambrevar:
Très bon français !

jmercouris:
@Ambrevar Can you please review this, I think I am ready to merge. I have a demo command I am building using it (show-bookmarks-panel), but I don't think that is necessary for this itself.

Ambrevar:
Will have a look hopefully tomorrow.

source/browser.lisp
```diff
@@ -613,6 +613,9 @@ sometimes yields the wrong result."
 (define-ffi-generic ffi-window-set-title (window title))
 (define-ffi-generic ffi-window-active (browser))
 (define-ffi-generic ffi-window-set-buffer (window buffer &key focus))
+(define-ffi-generic ffi-window-add-panel-buffer (window buffer side))
+(define-ffi-generic ffi-window-remove-panel-buffer (window buffer))
+(define-ffi-generic ffi-window-set-panel-buffer-width (window buffer width))
```
Ambrevar, line 618: Don't we want both width and heigth here?  Panels could be displayed with any kind of splitting in the future, better be flexible.

source/buffer.lisp
```diff
@@ -376,6 +376,14 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
 
 (define-user-class internal-buffer)
 
+(define-class panel-buffer (internal-buffer)
+  ((width 250 :documentation "The width in pixels."))
```
Ambrevar, line 380: Same comment about height.

source/renderer-gtk.lisp
```diff
@@ -42,7 +42,11 @@ want to change the behaviour of modifiers, for instance swap 'control' and
 
 (define-class gtk-window ()
   ((gtk-object)
-   (box-layout)
+   (root-box-layout)
+   (horizontal-box-layout)
+   (panel-buffer-container-left)
+   (panel-buffer-container-right)
+   (active-buffer-container)
```
Ambrevar, line 49: I would unexport all those slots if we want to implement vertical boxes later.

source/renderer-gtk.lisp
```diff
@@ -235,9 +243,21 @@ Such contexts are not needed for internal buffers."
                                        :type :toplevel
                                        :default-width 1024
                                        :default-height 768))
-       (setf box-layout (make-instance 'gtk:gtk-box
+       (setf root-box-layout (make-instance 'gtk:gtk-box
                                        :orientation :vertical
```
Ambrevar, line 247: Indentation is off.

source/renderer-gtk.lisp
```diff
@@ -235,9 +243,21 @@ Such contexts are not needed for internal buffers."
...
                                        :spacing 0))
+       (setf horizontal-box-layout (make-instance 'gtk:gtk-box
+                                                  :orientation :horizontal
+                                                  :spacing 0))
+       (setf panel-buffer-container-left (make-instance 'gtk:gtk-box
+                                                  :orientation :horizontal
+                                                  :spacing 0))
+       (setf panel-buffer-container-right (make-instance 'gtk:gtk-box
+                                                  :orientation :horizontal
+                                                  :spacing 0))
```
Ambrevar, line 257: Indentation is off for both panel buffer instantiations.

source/window.lisp
```diff
@@ -11,6 +11,14 @@
    (active-prompt-buffers '()
                           :export nil
                           :documentation "The stack of current prompt buffers.")
+   (panel-buffers-left (list)
+                      :accessor nil
+                      :documentation "A list of info buffers appearing on the
```
Ambrevar, line 16: Indentation is off.

Ambrevar: info buffers -> panel buffers

source/renderer-gtk.lisp
Ambrevar: Not sure about `active-buffer` because a prompt buffer that's showing is also an active buffer.  What about `main buffer` or `view buffer` instead?

source/window.lisp
```diff
@@ -85,6 +93,25 @@ The handlers take the window as argument."))
...
 
+(defmethod window-panel-buffers ((window window))
+  (append (slot-value window 'panel-buffers-left)
+          (slot-value window 'panel-buffers-right)))
+
+(defmethod window-add-panel-buffer ((window window) (buffer panel-buffer) side)
+  "Add a panel buffer to a window. Side can either be :right or :left."
+  (cond
+    ((eql side :left) (push buffer (slot-value window 'panel-buffers-left)))
+    ((eql side :right) (push buffer (slot-value window 'panel-buffers-right))))
```
Ambrevar, line 104: Move the `cond` around `'panel-buffers-left` and `'panel-buffers-right` to reduce code duplication.

source/window.lisp
```diff
@@ -85,6 +93,25 @@ The handlers take the window as argument."))
 
 (define-user-class window)
 
+(defmethod window-panel-buffers ((window window))
+  (append (slot-value window 'panel-buffers-left)
+          (slot-value window 'panel-buffers-right)))
```
Ambrevar, line 98: What about removing this and just having `panel-buffers` in `window`?

The `window-add-panel-buffer` function is in charge for chosing where to put it.
Optionally, if we want the panel to know _itself_ where it is, we could store the position in a slot.  But that's redundant information, so I'd rather have a dynamic function to retrieve the panel.

Having just `panel-buffers` would also make `window-remove-panel-buffer` simpler.

source/renderer-gtk.lisp
jmercouris: hm, OK. I will change it!

source/window.lisp
jmercouris: Yes, but I want to hide the implementation details behind a method. We need to know to which "container" they belong to (to the left or to the right) for the purposes of GTK.

source/browser.lisp
jmercouris: Yes, but they don't support it right now, so we don't add it yet. That is a backwards compatible change.

source/buffer.lisp
jmercouris: Same response as above.

source/browser.lisp
Ambrevar: Not if we have just 1 method, `ffi-window-set-panel-buffer-size` that takes 2 keyword arguments, width and height.
Later, having 2 functions to set the size would be annoying.  Having just one with `:width` for now will make sure we will be backward compatible.

source/window.lisp
jmercouris: Not sure, what do you mean here?

source/renderer-gtk.lisp
Ambrevar: Wait, I think we need to decide for an "official name" of this central buffer.  If we change it here, it would be great to make it a policy that this will be the term used everywhere to refer to the central buffer.

source/window.lisp
Ambrevar: That's what I'm missing: why to we need to know to which container they belong?  Seems to me this information is never used in the current code.

jmercouris: It is used in the GTK code.

jmercouris: Perhaps I should modify the GTK window class instead, that would make things simpler indeed.

jmercouris: I have an idea!

source/window.lisp
Ambrevar: This:

```lisp
(push buffer (slot-value window (cond ((eq side :left) 'panel-buffer-left)) (eq side :right) 'panel-buffer-right))
```

source/window.lisp
jmercouris: Actually weight, might the user not want to know which panel buffers are to the right and left? Rather than a contiguous list that they have to traverse?

source/window.lisp
jmercouris: Ah, I see. OK!

source/window.lisp
Ambrevar: I don't see where it's used in the GTK code.  Can you point it out?

jmercouris:
A little bit more styling, and a realistic panel example:


Screenshot_20210705_135541

jmercouris:
No need to do dynamic dispatch. Click is not captured in the panel buffer, and we can leave it as is. I would rather leave the input handling changes for another branch. I find that it is unrelated to this change.

I've also included a test command for you: `show-bookmarks-panel` and `delete-panel-buffer` that allows you to delete a panel buffer.

jmercouris:
In other words, I believe this is ready.

source/window.lisp
```diff
@@ -85,6 +87,31 @@ The handlers take the window as argument."))
...
+
+(defmethod window-remove-panel-buffer ((window window) (buffer panel-buffer))
+  "Remove a panel buffer from a window."
+  (setf (panel-buffers window)
+        (remove buffer (panel-buffers window)))
+  (ffi-window-remove-panel-buffer window buffer))
+
+(define-class panel-buffer-source (prompter:source)
+  ((prompter:name "Panel buffers")
+   (window :accessor window :initarg :window)
```
Ambrevar, line 103: The initform of `window` should probably be `(error "Window required...")`.

Ambrevar:
Tested, it's great!  Here is my last-minute feedback:

- Margin / padding missing on top of "Bookmarks".
- When there is no bookmark, the panel is empty which looks like a bug.  I
  suggest to echo a message.
- Unexport window panel slots.
- Does not work in vi mode (cannot click bookmark links), must be fixed before
  next release.

jmercouris:
This has been merged!

Ambrevar:
What about my previous comment:

> When there is no bookmark, the panel is empty which looks like a bug. I suggest to echo a message.

?

Ambrevar:
I also suggest to not show the panel when there is no bookmarks.

jmercouris:
I didn't do it because it was inconsistent with the list bookmarks command. I forgot to comment, sorry!

Ambrevar:
Then I suggest we fix both commands :)

Issue #1392, opened 2021-05-11
Scrolling with mouse-wheel in prompter crashes nyxt


franburstall:
**Describe the bug**
Scrolling with mouse in prompter crashes nyxt.

**Precise recipe to reproduce the issue**
In a web-buffer, I do:
* M-x history and see a list of possible completions
* Scroll with mouse-wheel to see more
* Nyxt crashes

**Information**
- OS name+version: Debian-10
- Graphics card and driver:
- Desktop environment / Window manager name+version: EXWM
- How you installed Nyxt (Guix pack, package manager, build from source): Source
- Information from command copy-system-information:
Nyxt version: 2-pre-release-7-23-g1ca2a43c
Renderer: GI-GTK
Operating system kernel: Linux 4.19.0-16-amd64
Lisp implementation: SBCL 2.1.1
Features: (SLYNK WEBKIT2 WEBKIT2-2.30 WEBKIT2-TRACKING WEBKIT2-MUTE
           WEBKIT2-EMOJI WEBKIT2-MEDIA WEBKIT2-SANDBOXING GTK-3-22 GTK-3-20
           GTK-3-18 GTK-3-16 GTK-3-14 GTK-3-12 GTK-3-10 GTK-3-8 GTK-3-6 GTK-3-4
           GTK GDK-3-22 GDK-3-20 GDK-3-18 GDK-3-16 GDK-3-14 GDK-3-12 GDK-3-10
           GDK-3-8 GDK-3-6 GDK-3-4 CAIRO-1-10 CAIRO-1-12 GDK-PIXBUF GLIB-2-30
           GLIB-2-32 GLIB-2-34 GLIB-2-36 GLIB-2-38 GLIB-2-40 GLIB-2-42
           GLIB-2-44 GLIB-2-46 GLIB-2-48 GLIB-2-50 GLIB-2-52 GLIB-2-54
           GLIB-2-56 GLIB-2-58 GLIB FSET-EXT-STRINGS CUSTOM-HASH-TABLE-NATIVE
           SWANK PLUMP-UTF-32 GLOBAL-VARS DECLARE-TYPES PARENSCRIPT
           NAMED-READTABLES 21BIT-CHARS CHUNGA FLEXI-STREAMS CLOSER-MOP
           CL-PPCRE-UNICODE CL-UNICODE CL-PPCRE BORDEAUX-THREADS SPLIT-SEQUENCE
[...]

Issue #399, opened 2019-09-26
Allow scrolling when link hints are active


alexherbo2:

Ambrevar:
Scrolling with the mouse wheel / touchpad works when links are active.
Is this what you meant?

Ambrevar:
Wait, actually I can only scroll the minibuffer, I cannot scroll the current web view.
@jmercouris Can you reproduce?

jmercouris:
I cannot reproduce

Ambrevar:
Recipe:

- Press `C-g` on this page.
- Click on the web view (hopefully to focus it).
- Scroll the mouse wheel or press the up/down arrows.

Expected result: it should scroll the web view.

Actual result: It scrolls the minibuffer.

jmercouris:
ah yes, I think we should unbind the scroll wheel

Ambrevar:
OK, I'll fix this as part of the minibuffer overhaul.

Ambrevar:
Actually, better than unbinding the scroll wheel, it'd be nice to support both: if the web view is focused, scroll it, otherwise scroll the minibuffer.

PR #1394, opened 2021-05-12
Scroll prompt buffer entries with mouse only when focused

Ambrevar:
This is a natural follow-up to #1392 and finally a fix to #399.

### DIFF
--- a/source/input.lisp
+++ b/source/input.lisp
@@ -32,9 +32,10 @@ Example:
     keymap))
 
 (export-always 'current-keymaps)
-(defun current-keymaps (&optional (buffer (if (active-prompt-buffers (current-window))
-                                              (current-prompt-buffer)
-                                              (current-buffer))))
+(defun current-keymaps (&optional (buffer (let ((prompt-buffer (current-prompt-buffer)))
+                                            (if (and prompt-buffer (ffi-focused-p prompt-buffer))
+                                                prompt-buffer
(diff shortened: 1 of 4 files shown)


jmercouris:
Ah, so the user will now be able to click and scroll, instead of hijacking all of the scroll because of the addition of the `is-focused` generic function!

source/input.lisp
```diff
@@ -32,9 +32,10 @@ Example:
     keymap))
 
 (export-always 'current-keymaps)
-(defun current-keymaps (&optional (buffer (if (active-prompt-buffers (current-window))
-                                              (current-prompt-buffer)
-                                              (current-buffer))))
+(defun current-keymaps (&optional (buffer (let ((prompt-buffer (current-prompt-buffer)))
```
jmercouris, line 35: Why not generalize this and just retrieve the focused buffer from the window? Have a method of the window `focused-buffer` or something like that?

jmercouris: Note that this change would also be very helpful when we have buffer splitting.

jmercouris: On the other hand, it adds extra complexity to this PR, and we still have `active-buffer`, which would need to be dealt with. I am thinking this does not belong in this PR, and that this technique suffices.

Ambrevar: All good points!

PR #1395, opened 2021-05-12
History tree dummy buffer fix

### LINKED ISSUE #1393: history-tree errors out
**Describe the bug**
From a web-buffer, I do `M-x history-tree` and nothing happens except an error:
````
<WARN> [19:04:51] Warning: Error on separate thead: The value
                           NIL
                         is not of type
                           NYXT:BUFFER
                         when binding BUFFER
````

**Information**
- OS name+version: Debian-10
- Graphics card and driver:
- Desktop environment / Window manager name+version: EXWM
- How you installed Nyxt (Guix pack, package manager, build from source): Source
- Information from command copy-system-information:
Nyxt version: 2-pre-release-7-23-g1ca2a43c
Renderer: GI-GTK
Operating system kernel: Linux 4.19.0-16-amd64
Lisp implementation: SBCL 2.1.1
Features: (SLYNK WEBKIT2 WEBKIT2-2.30 WEBKIT2-TRACKING WEBKIT2-MUTE
           WEBKIT2-EMOJI WEBKIT2-MEDIA WEBKIT2-SANDBOXING GTK-3-22 GTK-3-20
           GTK-3-18 GTK-3-16 GTK-3-14 GTK-3-12 GTK-3-10 GTK-3-8 GTK-3-6 GTK-3-4
           GTK GDK-3-22 GDK-3-20 GDK-3-18 GDK-3-16 GDK-3-14 GDK-3-12 GDK-3-10
           GDK-3-8 GDK-3-6 GDK-3-4 CAIRO-1-10 CAIRO-1-12 GDK-PIXBUF GLIB-2-30
           GLIB-2-32 GLIB-2-34 GLIB-2-36 GLIB-2-38 GLIB-2-40 GLIB-2-42
           GLIB-2-44 GLIB-2-46 GLIB-2-48 GLIB-2-50 GLIB-2-52 GLIB-2-54
           GLIB-2-56 GLIB-2-58 GLIB FSET-EXT-STRINGS CUSTOM-HASH-TABLE-NATIVE
           SWANK PLUMP-UTF-32 GLOBAL-VARS DECLARE-TYPES PARENSCRIPT
           NAMED-READTABLES 21BIT-CHARS CHUNGA FLEXI-STREAMS CLOSER-MOP
           CL-PPCRE-UNICODE CL-UNICODE CL-PPCRE BORDEAUX-THREADS SPLIT-SEQUENCE
           GRAY-STREAMS FAST-IO-SV FAST-IO SBCL-USES-SB-ROTATE-BYTE
           ASDF-SYSTEM-CONNECTIONS CL-JSON-CLOS CL-JSON THREAD-SUPPORT
           FLAT-NAMESPACE X86-64 UNIX CFFI FLAT-NAMESPACE SEQUENCE-EMPTYP
           QUICKLISP ASDF3.3 ASDF3.2 ASDF3.1 ASDF3 ASDF2 ASDF OS-UNIX
           NON-BASE-CHARS-EXIST-P ASDF-UNICODE X86-64 GENCGC 64-BIT ANSI-CL
           COMMON-LISP ELF IEEE-FLOATING-POINT LINUX LITTLE-ENDIAN
           PACKAGE-LOCAL-NICKNAMES SB-LDB SB-PACKAGE-LOCKS SB-THREAD SB-UNICODE
           SBCL UNIX)

### PR DESCRIPTION
**Pull request by Ambrevar**
Fixes #1393.

@aartaka Why did you originally go for a dummy buffer to retrieve the history data?  Am I doing something wrong by getting the current buffer history path?

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -354,8 +354,9 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
 (define-user-class editor-buffer)
 
 (defun make-dummy-buffer ()
-  ;; Internal buffers are lighter than full-blown buffers which can have a
-  ;; WebKit context, etc.
+  "Internal buffers are lighter than full-blown buffers which can have a
+WebKit context, etc.
+Delete it with `ffi-buffer-delete'."
   (make-instance 'user-internal-buffer))
 
 (define-class status-buffer (user-internal-buffer)
@@ -670,6 +671,7 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
 
 (declaim (ftype (function (buffer)) buffer-delete))
 (defun buffer-delete (buffer)
+  "For dummy buffers, use `ffi-buffer-delete' instead."
   (hooks:run-hook (buffer-delete-hook buffer) buffer)
   (let ((parent-window (find buffer (window-list) :key 'active-buffer)))
     (when parent-window
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -424,9 +424,8 @@ Otherwise go forward to the only child."
   (nyxt::with-current-html-buffer (output-buffer "*History*"
                                                  'nyxt/history-tree-mode:history-tree-mode)
     (with-data-unsafe (history (let ((dummy-buffer (nyxt::make-dummy-buffer)))
-                                 (prog1
-                                     (history-path dummy-buffer)
-                                   (delete-buffer :id (id dummy-buffer)))))
+                                 (prog1 (history-path dummy-buffer)
+                                   (ffi-buffer-delete dummy-buffer))))
       (let ((markup:*auto-escape* nil)
             (mode (find-submode output-buffer 'nyxt/history-tree-mode:history-tree-mode))
             (tree `(:ul ,(htree:map-tree
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -423,9 +423,7 @@ Otherwise go forward to the only child."
   "Open a new buffer displaying the whole history tree."
   (nyxt::with-current-html-buffer (output-buffer "*History*"
                                                  'nyxt/history-tree-mode:history-tree-mode)
-    (with-data-unsafe (history (let ((dummy-buffer (nyxt::make-dummy-buffer)))
-                                 (prog1 (history-path dummy-buffer)
-                                   (ffi-buffer-delete dummy-buffer))))
+    (with-data-unsafe (history (history-path (current-buffer)))
       (let ((markup:*auto-escape* nil)
             (mode (find-submode output-buffer 'nyxt/history-tree-mode:history-tree-mode))
             (tree `(:ul ,(htree:map-tree

### CODE REVIEW

**Comment by aartaka** [comment]
> @aartaka Why did you originally go for a dummy buffer to retrieve the history data? Am I doing something wrong by getting the current buffer history path?

Well, it was because one might want to look at the whole global history tree from `nosave` buffer. That's why we need get data from the non-configured dummy buffer.

But that's not enough of a reason, I guess. I'm alright with using `current-buffer`. Thanks for the fix!

**Comment by Ambrevar** [comment]
> But that's not enough of a reason, I guess. I'm alright with using `current-buffer`. Thanks for the fix!

Isn't it?  This means that calling `history-tree` from a nosave buffer
won't do anything.

Also using `current-buffer` means that it will have different results
depending on the current buffer, if the user uses multiple history
trees.

The dummy buffer approach has the benefit of working reproducibly,
independently of the current buffer.

Thoughts?

**Comment by aartaka** [comment]
Maybe use MOP to get initform?

But maybe we need more there, because `epxpand-path` also relies on `current-buffer`, so we need to use `with-current-buffer` on dummy one. So, the proper way is something along the (untested!) lines of:
``` lisp
(define-command history-tree ()         ; TODO: Factor this with `buffer-history-tree'.
  "Open a new buffer displaying the whole history tree."
  (nyxt::with-current-html-buffer (output-buffer "*History*"
                                                 'nyxt/history-tree-mode:history-tree-mode)
    (let ((dummy-buffer (nyxt::make-dummy-buffer)))
      (with-current-buffer dummy-buffer
        (with-data-unsafe (history (history-path (current-buffer)))
          ;; More code there.
          ))
      (ffi-delete-buffer :id (id dummy-buffer)))))
```

Ambrevar:
Let's take a step back and think about the semantics.
We are trying to get the global history, which reference is only to be
found in buffers.

- The buffer `history-path` initform may hold a path that's not
  actually used, e.g. the user configured the buffers to use a GHT
  stored in a mode.

- Some buffers maybe have no history.  (I.e. (expand-path (history-path
  buffer)) expands to nil.)

- Some buffers may have a history different from "default" GHT.

So which history are we talking about?

Indeed, there may be multiple GHTs.

So what I suggest we do is the following: use

```lisp
(with-data-unsafe (history (history-path (current-buffer) ...)))
```

and in the resulting GHT buffer, print somewhere the buffers that use
this GHT ((current-buffer) at least).

For buffers without history, just echo a message.

Result: this is more accurate and would work in all cases, whatever user
configuration.
It's also "do what I mean": it works in buffers with GHTs, while calling
`history-tree` in a nosave buffer would not confuse the user by
showing a history that has nothing to do with the current buffer.

Thoughts?

aartaka:
Yes, that sounds sane, especially given that we don't modify the data — `with-data-unsafe` fits well :)

aartaka:
Oh wait, it was there all along xD

Anyway, I agree with your proposal, Pierre.

Maybr we should add a `has-data-p` predicate for `data-paths` so that it's testable whether there's any data?

Ambrevar:
Doh, just understood something: `history-tree` does not display the
whole tree, only the whole _branch_ the current buffer is on.

So it makes perfect sense to user the current buffer here! :)

Ambrevar:
Merged with 7210ce34a5affdebac04316d6d88a4e5afd40a31.

aartaka:
> Doh, just understood something: `history-tree` does not display the whole tree, only the whole _branch_ the current buffer is on. So it makes perfect sense to user the current buffer here! :)

Does it? 0_o

I though it's for all the tree.

Ambrevar:
> I though it's for all the tree.

For this, we would need to loop over all the roots.

The history-tree function runs `map-tree`, which iterates on the nodes
starting from the _current root_ only.

aartaka:
> I though it's for all the tree.
> For this, we would need to loop over all the roots. The history-tree function runs `map-tree`, which iterates on the nodes starting from the _current root_ only.

Yes, you're right :) Just checked the code, and `(current-buffer)` fits fine here.

Ambrevar:
We've never printed the entire tree since we've moved to the GHT, so
there is no regression here.

We could add this feature in the future though.

Issue #1294, opened 2021-04-13
Rename enable-mode-for-buffer, etc.


Ambrevar:
I suggest the following renames for consistency:

- enable-mode-for-buffer -> enable-mode
- disable-mode-for-buffer -> disable-mode
- enable-mode-for-current-buffer -> enable-mode-current-buffer
- disable-mode-for-current-buffer -> disable-mode-current-buffer

Thoughts?

jmercouris:
I don't see the advantage. I see these as simply less verbose. Can you explain how that would be improving consistency? If anything I see it as the opposite.

aartaka:
Modes are already tied to buffers, and commands are often working on them too, so no need for extraneous `-for-buffer` suffix there. I agree with Pierre's suggestion here, as it results in less typing without meaning loss.

I can guess that consistency is about name of the command (`enable-mode`) matching the name of the underlying function (`enable-modes`). @Ambrevar, is that so?

jmercouris:
Are modes tied to buffers though? This is a question that I do not think we have thoroughly answered!

Ambrevar:
By consistency, I meant with almost all other commands:

- reload-buffer / reload-current-buffer
- delete-buffer / delete-current-buffer
- bookmark-url / bookmark-current-page (should we rename these ones too?)
- delete-window / delete-current-window

Another inconsistency is with

`history-tree` / `buffer-history-tree`.  I believe this last command
should be `history-tree-current`.

Ambrevar:
Second thoughts?

jmercouris:
My thoughts remain the same, I disagree with this change, but I am willing to accept it! :-)

Ambrevar:
You haven't explained why you disagree.  Can you elaborate?

jmercouris:
I think it is less verbose. Consider this:

`enable-mode`. Does this simply enable a mode? enable a mode for a buffer? Allow a mode to be enabled? What does it mean? However `enable-mode-for-buffer` is very clear.

Ambrevar:
I find it very clear, considering modes only make sense in the context
of a buffer.

If the user does not know what a mode is, `enable-mode-for-buffer` won't
make more sense.

Finally, the docstring is displayed in M-x and says "Enable queried mode(s) for
select buffer(s)".

My approach to command naming is that the priority is on consistency
with names, while documentation is done in the docstring.

jmercouris:
[...]

PR #1398, opened 2021-05-12
Rename several commands

jmercouris:
bookmark-current-page -> bookmark-current-url
enable/disable-mode-for-buffer -> enable/disable-mode


Ambrevar:
See #1294.

Ambrevar:
To get the list of commands, sorted alphabetically:

```lisp
(in-package :nyxt)
(sort (mapcar #'write-to-string  (mapcar #'name (list-commands))) 'string<)
```

Other suggestions:

- bookmark-delete -> delete-bookmark (consistency)
- search-buffer -> shouldn't it be `search` for consistency?
- reload-buffers -> reload-buffer (consistency)
- search-selection -> `query-selection-in-search-engine` (lengthy, but more accurate)
- All prompt buffer commands should remove the "prompt-buffer-" prefix.
- "*-backward-char" vs. "*backwards*" (notice the "s").  Same with "forwards".  We need to pick one and stick to it.  Without the "s" is shorter, but doesn't fuzzy-match as well as with an "s".  I guess it's OK considering we have live-narrowing fuzzy matching.
- `jump-to-heading-across-buffers`: Not sure about this one, should we remove "across"?  It's the only multi-buffer command with "across".

jmercouris:
> To get the list of commands, sorted alphabetically:
> 
> ```lisp
> (in-package :nyxt)
> (sort (mapcar #'write-to-string  (mapcar #'name (list-commands))) 'string<)
> ```
> 
> Other suggestions:
> 
> * bookmark-delete -> delete-bookmark (consistency)

OK!

> * search-buffer -> shouldn't it be `search` for consistency?

No. Search buffer is clear. It is searching the buffer, not anything else. I have a very strong opinion about this one!

> * reload-buffers -> reload-buffer (consistency)

Reload buffer alread exists. Reload-buffers allows you to reload multiple buffers.

> * search-selection -> `query-selection-in-search-engine` (lengthy, but more accurate)

OK1

> * All prompt buffer commands should remove the "prompt-buffer-" prefix.

Why?

> * "*-backward-char" vs. "_backwards_" (notice the "s").  Same with "forwards".  We need to pick one and stick to it.  Without the "s" is shorter, but doesn't fuzzy-match as well as with an "s".  I guess it's OK considering we have live-narrowing fuzzy matching.
> * `jump-to-heading-across-buffers`: Not sure about this one, should we remove "across"?  It's the only multi-buffer command with "across".

`jump-to-heading-buffers` works, I guess :-D

jmercouris:
After going through the list of commands, the only suggestion I have is `copy` -> `copy-text`.

jmercouris:
Also, I assume you want to remove the `prompt-buffer` prefix from prompt-buffer commands because they are in their own package?

Ambrevar:
>> * reload-buffers -> reload-buffer (consistency)
>
> Reload buffer alread exists. Reload-buffers allows you to reload multiple buffers.

We don't have a reload-buffer command though.

For all other buffer-related command, we have the following pattern:

- FOO-current-buffer (for current buffer)
- FOO-buffer (for multiple buffers -- no "s" here)

Example: delete-buffer and delete-current-buffer.

>> * All prompt buffer commands should remove the "prompt-buffer-" prefix.
>
> Why?

It's redundant, they already belong to the prompt-buffer-mode package.
The names are thus extra long:

nyxt/prompt-buffer-mode:prompt-buffer-toggle-mark-backwards

I'd much rather have

nyxt/prompt-buffer-mode:toggle-mark-backwards

It's also more consistent with the rest of the code base (we don't
prefix web-mode commands with "web-".)

>> * "*-backward-char" vs. "_backwards_" (notice the "s").  Same with "forwards".  We need to pick one and stick to it.  Without the "s" is shorter, but doesn't fuzzy-match as well as with an "s".  I guess it's OK considering we have live-narrowing fuzzy matching.
>> * `jump-to-heading-across-buffers`: Not sure about this one, should we remove "across"?  It's the only multi-buffer command with "across".
>
> `jump-to-heading-buffers` works, I guess :-D

As above, without "s" for consistency?

If you prefer with "s", fine by me, but let's rename all multi-buffer
commands with an "s" then.

Ambrevar:
> After going through the list of commands, the only suggestion I have is `copy` -> `copy-text`.

Hmm... Can you elaborate?  What about `paste` then?

jmercouris:
> We don't have a reload-buffer command though. For all other buffer-related command, we have the following pattern: - FOO-current-buffer (for current buffer) - FOO-buffer (for multiple buffers -- no "s" here) Example: delete-buffer and delete-current-buffer.

I see, you wanted it to be a command! OK, done!

> > * All prompt buffer commands should remove the "prompt-buffer-" prefix. Why?
> It's redundant, they already belong to the prompt-buffer-mode package. The names are thus extra long: nyxt/prompt-buffer-mode:prompt-buffer-toggle-mark-backwards I'd much rather have nyxt/prompt-buffer-mode:toggle-mark-backwards It's also more consistent with the rest of the code base (we don't prefix web-mode commands with "web-".)
> > * "*-backward-char" vs. "_backwards_" (notice the "s"). Same with "forwards". We need to pick one and stick to it. Without the "s" is shorter, but doesn't fuzzy-match as well as with an "s". I guess it's OK considering we have live-narrowing fuzzy matching. 

Let's keep the S, it is more correct.

> * `jump-to-heading-across-buffers`: Not sure about this one, should we remove "across"? It's the only multi-buffer command with "across". `jump-to-heading-buffers` works, I guess :-D
> As above, without "s" for consistency? If you prefer with "s", fine by me, but let's rename all multi-buffer commands with an "s" then.

OK! Will do!

jmercouris:
By my reckoning everything looks OK. I can't figure out which commands you think should have an "S" appended :-D

documents/CHANGELOG.org
```diff
@@ -431,9 +431,9 @@ If possible!
 
   - highlight-regexp
 
-  - zoom-images-in
+  - zoom-images
 
-  - zoom-images-out
+  - unzoom-images
```
Ambrevar, line 436: This should not be renamed.

documents/CHANGELOG.org
```diff
@@ -148,7 +148,7 @@ If possible!
   - Fix WM class and name on both Xorg and Wayland.
 
   - Fix many commands that broke with the new prompt buffer API
-    (=reload-buffers=, =bookmark-url=, =bookmark-page=, =search-buffer=...).
+    (=reload-buffers=, =bookmark-url=, =bookmark-buffer-url=, =search-buffer=...).
```
Ambrevar, line 151: This should not be renamed.

libraries/prompter/tests/fuzzy.lisp
```diff
@@ -11,14 +11,14 @@
...
-    "BOOKMARK-CURRENT-PAGE" "NEW-BUFFER" "COMMAND-INSPECT"
+    "BOOKMARK-CURRENT-URL" "NEW-BUFFER" "COMMAND-INSPECT"
     "ADD-SEARCH-HINTS" "KILL" "REMOVE-SEARCH-HINTS" "LOAD-FILE"
     "KEYMAP" "NYXT-VERSION" "NAME" "SCROLL-LEFT" "ACTIVATE"
     "SCROLL-PAGE-DOWN" "SCROLL-RIGHT" "DESTRUCTOR"
     "SCROLL-TO-BOTTOM" "SWITCH-BUFFER-NYXT" "COMMAND-EVALUATE"
     "DID-FINISH-NAVIGATION" "BOOKMARK-ANCHOR" "SCROLL-DOWN"
     "SCROLL-UP" "VI-BUTTON1" "RELOAD-CURRENT-BUFFER"
-    "COPY-ANCHOR-URL" "BOOKMARK-DELETE" "GO-ANCHOR-NEW-BUFFER"
+    "COPY-ANCHOR-URL" "DELETE-BOOKMARK" "GO-ANCHOR-NEW-BUFFER"
```
Ambrevar, line 21: This may alter the test and it's not relevant anyways, you can leave it out.

Ambrevar:
> Let's keep the S, it is more correct.

Then `reload-buffer` should be `reload-buffers`, right?

Ambrevar:
Beside the last few nits in my review, this change it looking good, I'm very happy with the improved concistency!

As they say, "naming is hard" :p

documents/CHANGELOG.org
jmercouris: Why not? I tried to make it consistent with the new zoom command names.

libraries/prompter/tests/fuzzy.lisp
jmercouris: OK

documents/CHANGELOG.org
jmercouris: Why not?

Ambrevar: Because it's in the change log of older versions :)
The name was bookmark-page for this version.

documents/CHANGELOG.org
Ambrevar: It's the changelog :)
The command renaming is fine, it's just that the old changelog entries should not be editted.  We don't want to rewrite history :p

jmercouris:
> Beside the last few nits in my review, this change is looking good, I'm very happy with the improved concistency!
> 
> As they say, "naming is hard" :p

1111111111

documents/CHANGELOG.org
jmercouris: Ah, I see, you are saying to not change the changelog! Understood!

documents/CHANGELOG.org
jmercouris: Aha indeed, tut mir leid!

jmercouris:
As soon as the tests pass, I will merge!

PR #1402, opened 2021-05-13
Recursive describe

Ambrevar:
Try describing `*interfaces*` or any global variable that's a list of things.  Now you can click on the element to describe them.

I've used the trick to store the value being inspected in the mode-local variable `target`, otherwise the `lisp://` URL may not be able to refer to the value when it's unprintable like with structures.

This may not be the most general nor future-proof design.

Thoughts?

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -26,12 +26,16 @@ If URL-STRING cannot be converted to a `quri:uri' object, return an empty `quri:
 If a URL string cannot be converted to a `quri:uri', it is discarded from the result."
   (remove-if #'url-empty-p (mapcar #'string->url url-strings)))
 
-(defun has-url-method-p (object)
+(defun has-method-p (object generic-function)
   "Return non-nil if OBJECT has `url' specialization."
   (some (lambda (method)
           (subtypep (type-of object) (class-name
                                       (first (closer-mop:method-specializers method)))))
-        (closer-mop:generic-function-methods  #'url)))
+        (closer-mop:generic-function-methods generic-function)))
+
+(defun has-url-method-p (object)
+  "Return non-nil if OBJECT has `url' specialization."
+  (has-method-p object #'url))
 
 (deftype url-designator ()
   `(satisfies has-url-method-p))
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -27,7 +27,7 @@ If a URL string cannot be converted to a `quri:uri', it is discarded from the re
   (remove-if #'url-empty-p (mapcar #'string->url url-strings)))
 
 (defun has-method-p (object generic-function)
-  "Return non-nil if OBJECT has `url' specialization."
+  "Return non-nil if OBJECT has GENERIC-FUNCTION specialization."
   (some (lambda (method)
           (subtypep (type-of object) (class-name
                                       (first (closer-mop:method-specializers method)))))
--- a/source/help-mode.lisp
+++ b/source/help-mode.lisp
@@ -10,6 +10,5 @@
 (define-mode help-mode ()
   "Mode for displaying documentation."
   ((rememberable-p nil)
-   (target nil
-           :type t
-           :documentation "The value being inspected.")))
+   (inspected-value nil
+                    :type t)))
--- a/source/help-mode.lisp
+++ b/source/help-mode.lisp
@@ -9,4 +9,7 @@
 
 (define-mode help-mode ()
   "Mode for displaying documentation."
-  ((rememberable-p nil)))
+  ((rememberable-p nil)
+   (target nil
+           :type t
+           :documentation "The value being inspected.")))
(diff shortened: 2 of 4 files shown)


jmercouris:
It is interesting! Very cool! It is basically a workaround for the fact that we cannot directly pass Lisp objects/references to HTML. Perhaps we can solve that problem in a generalized way, and then we won't need a workaround.

I also think that this is definitely not for before 2.0 though. I believe there are too many unknowns.

Ambrevar:
> It is interesting! Very cool! It is basically a workaround for the fact that we cannot directly pass Lisp objects/references to HTML. Perhaps we can solve that problem in a generalized way, and then we won't need a workaround.

Note that this "workaround" I've implemented... might just be the clean
way to do it actually!

Need to think a bit more about it.

> I also think that this is definitely not for before 2.0 though. I believe there are too many unknowns.

I'm fine with merging now or later.
It does not change anything beside `describe-variable`, so the scope is
limited and as such I think it's fine to merge.

Thouhgts?

jmercouris:
I would rather wait!

aartaka:
This `(nth ...)` hints me at the idea: how about passing sexps we got the value with to `lisp-url`? This way we need no variables.

Example (I'm pretty sure it'll fail because I forgot our buffers API, but you've got the idea)
``` lisp
(lisp-url `(nyxt:buffer-history-tree :buffer (nyxt:buffers-get :id ,(id (nyxt:current-buffer))))
```

The downside is that it requires much more discipline and is not always easy/possible to do.

Ambrevar:
Not sure to understand what you are trying to do?

Are you saying we could chain-wrap the s-exps instead of storing the
intermediate value in `target`?

Example:

- (lisp-url `(search-engines (current-buffer)))

- The link to each search engine will be
  (nth I (search-engines (current-buffer)))

  instead of 

  (nth I (target (current-buffer)))

  ?

The problem in the above is that the lisp-url code is stateful:
current-buffer does not have the same value in the second instance.

So I don't think it would work, unless we only had stateless URLs, which
seems difficult to do.

Storing values in a `target` slot is an easy and reliable way to proceed.

aartaka:
> Not sure to understand what you are trying to do? Are you saying we could chain-wrap the s-exps instead of storing the intermediate value in `target`? 

Exactly.

> Example: - (lisp-url `(search-engines (current-buffer))) - The link to each search engine will be (nth I (search-engines (current-buffer))) instead of (nth I (target (current-buffer))) ? The problem in the above is that the lisp-url code is stateful: current-buffer does not have the same value in the second instance. So I don't think it would work, unless we only had stateless URLs, which seems difficult to do. Storing values in a `target` slot is an easy and reliable way to proceed.

Thanks for this problematizing example! Yes, this is deeply state-dependent, but the treatment of `current-buffer` that I suggested is somewhat different: we take its id at the time of `lisp-url` creation, we insert it into `buffers-get`, and this will always return us the same buffer (unless it's deleted, but that causes problems with anything, including `target`).

What I mean is that we don't need `target` if we can clearly trace where our objects come from. I.e., using expressions that'll guide us to the object (insert \`(buffers-get ,(id (current-buffer)), instead of putting object somewhere (`(setf (target (current-mode 'help)) (current-buffer))`).

And what if we want to make a help buffer related to several objects? Make a list? Make another slot? This doesn't scale well, so  it seems better to just learn the path/way/source-code-formula to the objects we use and insert it wherever (and if) necessary.

Ambrevar:
> And what if we want to make a help buffer related to several objects?

We can always make `target` a hash table for instance.

> Make a list? Make another slot?

Do you have examples?

What you are suggesting seems very interesting on the theoretical level,
it's just that I'm not sure of what the practical benefits would be.

At first glance, it seems to be a hard problem.  I wonder if anyone has
ever derived a theory around this.

Ambrevar:
I've retested this, seems to work well.

OK to merge?
We can open a new issue for @aartaka's suggestion if we decide to implement it.

Once this is merged, I'd like to implement cross-references.

source/urls.lisp
```diff
@@ -26,12 +26,16 @@ If URL-STRING cannot be converted to a `quri:uri' object, return an empty `quri:
 If a URL string cannot be converted to a `quri:uri', it is discarded from the result."
   (remove-if #'url-empty-p (mapcar #'string->url url-strings)))
 
-(defun has-url-method-p (object)
+(defun has-method-p (object generic-function)
   "Return non-nil if OBJECT has `url' specialization."
```
aartaka, line 30: Irrelevant docstring?

aartaka (commented):
Alright to me, except for the forgotten docstring!

source/urls.lisp
Ambrevar: Good catch, I think I let that one slip through during the rebase conflict resolution.

PR #1403, opened 2021-05-13
package: Unuse trivia, remove alex and sera nicknames.

Ambrevar:
I believe this is essential to guarantee a stable API.

First, we should not `:use` trivia.  At best we should `import-from` `match` and `match-lambda`.
But better leave this to the user, really.

Indeed, should we `use` package which we don't control, and should these packages API change, an update to the offending package may break the user configuration.

Same with local nicknames: `sera` effectively binds us to Serapeum forever.  What if we want to get rid of it?
What if we decide someday to use `a:` instead of `alex`?

For all these reasons, I believe it's better to leave it to the user to decide which symbols they want to import, and which nicknames they want to use.

### DIFF
--- a/source/package.lisp
+++ b/source/package.lisp
@@ -22,16 +22,30 @@
   (:export :repl-mode))
 
 (uiop:define-package nyxt-user
-  (:use #:common-lisp #:trivia #:nyxt)
+  (:use #:common-lisp #:nyxt)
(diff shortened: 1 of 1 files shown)


Ambrevar:
OK to merge this?

jmercouris:
I concur!

PR #1404, opened 2021-05-13
auto-mode: Add reload-with-modes command.

Ambrevar:
Some time ago @jmercouris expressed an itch with auto-mode which had also scratched me for a long time: that it was too painful to disable auto-mode and enable / disable other modes for the current buffer.

https://github.com/atlas-engineer/nyxt/pull/1344

This fixes another itch of mine with auto-mode, possibly a bigger one: to bypasse auto mode for the current URL only.
Indeed, only to often when you use noscript-mode, etc. by default, you navigate to a page that does not load properly.  Then I want to disable the offending mode, but keep auto-mode enabled for future navigations in this buffer.

`reload-with-modes` does just that.

@aartaka What do you think of the implementation?

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -4,6 +4,7 @@
 (uiop:define-package :nyxt/auto-mode
   (:use :common-lisp :nyxt)
   (:import-from #:class-star #:define-class)
+  (:import-from #:keymap #:define-key #:define-scheme)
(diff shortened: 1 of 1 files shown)


Ambrevar:
@aartaka Fine merging this?

aartaka:
Yes, I'm totally fine with this! Seems like the comment I wrote before wasn't sent 0_o

Issue #1406, opened 2021-05-13
Move bookmarklets to nyxt/web-mode?


Ambrevar:
By definition they are javascript, so they run over the _content_ of the web page, thus they should belong to nyxt/web-mode, nah?

jmercouris:
Hm, you could make that argument, I'm not very convinced

Ambrevar:
Can you elaborate?

The issue with putting them in `nyxt` is that they become global, so
it's a bit polluting.

Secont point is consistency: all bookmarklets related to content
manipulation, like zooming.  That's what web mode commands do
(zoom-in-page, etc.).

The rationale of the separation between base-mode and web-mode is the
following:

- Base mode is for anything global that does not relate to the _content_
  of a buffer.

- Web mode is for the (HTML) content of a buffer.

jmercouris:
The rationale is clear. If they belong to a mode, web mode is the logical choice. I guess I am just not as concerned about polluting the namespace. It is our own namespace after all :-O!

Ambrevar:
See my suggestion in #1408.

Ambrevar:
>  It is our own namespace after all 

`:nyxt` is `:use`d by `:nyxt-user`, so we must be careful about what we export in there.  This is the critical API part that carries all the technical debts of Nyxt :p

PR #1408, opened 2021-05-14
bookmarklets: Move all to nyxt/web-mode package.

Ambrevar:
Indeed, they manipulate the content of the buffer, they should not be global
commands.

Bonus: Move bookmarklets-related commands from data-storage (why were they
there?) to bookmarklets.lisp.


Ambrevar:
Addresses #1406.

jmercouris:
I can accept this change EXCEPT for moving the functions out of data storage. I don't see why they would belong to web-mode, they have use in a broader context.

jmercouris:
Additionally the `define-bookmarklet-command` macro should belong to the Nyxt package.

Ambrevar:
> Additionally the `define-bookmarklet-command` macro should belong to the Nyxt package.

Good catch!

Ambrevar:
> I can accept this change EXCEPT for moving the functions out of data storage. I don't see why they would belong to web-mode, they have use in a broader context.

No strong opinion on where they belong.  Here was my thinking process:

- They don't belong to data-storage :p  Maybe configuration.lisp?
- They are not exported, so it does not really matter where they are.
- They are only used in bookmarklets so far.  We can always move them to
  a more general location when we get a need for them.
  Since they are unexported, we are free to move them around as we wish.

Thoughts?

jmercouris:
OK, `configuration.lisp` is fine.

Ambrevar:
Just force-pushed, what do you think?

jmercouris:
Thank you!

Issue #1024, opened 2020-11-04
Improve status and CSS customizability


Ambrevar:
At the moment customizing the CSS of various elements (mostly the status bar) is tedious because the user needs to repeat the existing CSS.

Besides, it's hard to access a CSS property value, e.g. the background color of the status bar.

What we could do is only apply the `cl-css:css` function when creating the GTK objects.  Thus we would store an alist/plist for the current CSS slots.

The status buffer definition would look like this:

```lisp
(define-class status-buffer (user-internal-buffer)
  ((height 20
           :type integer
           :documentation "The height of the status buffer in pixels.")
   (style '((body
             :background "rgb(200, 200, 200)"
             :font-size "14px"
             :color "rgb(32, 32, 32)"
             :padding 0
             :margin 0
             :line-height "20px")
            ;; ...
            (|.button:hover|
             :color "black"))))
  (:export-class-name-p t)
  (:export-accessor-names-p t)
  (:accessor-name-transformer #'class*:name-identity))
```

For instance, to access the background value of a STATUS-BUFFER object, we can write (untested)

```lisp
(getf (rest (assoc 'body (style STATUS-BUFFER))) :background)
```

To access the default value, replace `(style STATUS-BUFFER)` with

```lisp
(getf (mopu:slot-properties 'status-buffer 'style) :initform)
```

Since the status buffer is a user-class, the user can customize the existing CSS this way (untested):

```lisp
(define-configuration status-buffer
  ((style (setf (alex:assoc-value %slot-default ...) new-value))))
```

jmercouris:
One minor correction: they do not need to repeat the CSS, only need to define the delta, CSS cascades and subsequent definitions override previous ones.

jmercouris:
Therefore they can just append to the slot default value in order to see the CSS to whatever they wish, or they can start from scratch.

Ambrevar:
Good point.

That said the drawback with CSS is that the user does not get language
support (at the Lisp level) against symbols for the CSS path.

For instance, if the user overrides "body" with "boyd" (intended typo)
by mistake, the rule won't apply but the user won't get warned.

With `assoc` this would not be a problem.

jmercouris:
Indeed!

aadcg:
I like your idea @Ambrevar.

jmercouris:
I am still not sold on this. I am concerned it may become too complex.

aadcg:
[...]

PR #1409, opened 2021-05-14
status-buffer: vi mode indicator

edgar-vincent:
This adds a vi mode indicator in the status buffer when `vi-normal-mode` or `vi-insert-mode` are enabled. There is some flickering in the status buffer when changing modes - I'm not sure how to fix that.
Also, my branch's history is messed up, but I assume that you'll squash the commits :)

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -395,7 +395,7 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
                 :padding-left "10px"
                 :text-align "center")
                (".overlap"
-                :grid-column-start "span 2")
+                :grid-column "span 3")
                ("#url"
                 :background-color "rgb(120,120,120)"
                 :min-width "100px"
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -382,14 +382,20 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
                 :clip-path "polygon(0 50%, 100% 100%, 100% 0)")
                ("#container"
                 :display "grid"
-                ;; Columns: controls, arrow, url, arrow, tabs, arrow, modes
-                :grid-template-columns "115px 10px 1fr 10px 2fr 10px 250px"
+                ;; Columns: controls, arrow, vi-status, arrow, url, arrow, tabs, arrow, modes
+                :grid-template-columns "115px 10px 20px 10px 1fr 10px 2fr 10px 250px"
                 :overflow-y "hidden")
                ("#controls"
                 :background-color "rgb(80,80,80)"
                 :padding-left "5px"
                 :overflow "hidden"
                 :white-space "nowrap")
+               ("#vi-mode"
+                :margin-left "-10px"
+                :padding-left "10px"
+                :text-align "center")
+               (".overlap"
+                :grid-column-start "span 2")
                ("#url"
                 :background-color "rgb(120,120,120)"
                 :min-width "100px"
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -281,7 +281,7 @@ instance must be non-nil.")
         (:code "~/.local/share/nyxt/extensions")".")
     (:p "Extensions are regular Common Lisp systems.")
     (:p "A catalogue of extensions is available in the "
-        (:code "document/EXTENSIONS.org") "file in the source repository.")
+        (:code "document/EXTENSIONS.org") " file in the source repository.")
 
     (:h2 "Troubleshooting")
     (:h3 "Playing videos")
--- a/source/status.lisp
+++ b/source/status.lisp
@@ -24,6 +24,17 @@
    (:a :class "button" :title "Execute" :href (lisp-url '(nyxt:execute-command)) "⚙")
    (:a :class "button" :title "Buffers" :href (lisp-url '(nyxt::list-buffers)) "≡")))
 
+(export-always 'format-status-vi-mode)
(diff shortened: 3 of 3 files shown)


Ambrevar:
Thank you for your contribution!
If I'm not mistaken, this is your first contribution, so welcome to the world of Nyxt hackers :)

You are addressing an important point: per-mode status buffer customization.
Unfortunately we can't hard-code the mode customization.  Instead, we should derive a framework that's extensible and where the core functions are mode-agnostic.

If you look at `format-status-modes`, you'll see that it calls `glyph` which is a method that can be specialized for each mode.

This is a bit limited and I believe we can do better.  In particular:
- Change the appearance of the HTML span of the mode button in the status bar.
- A mode should be able to tell whether it should appear first.  How do we resolve conflicts then?
- Should a mode be able to tweak the appearance of the whole mode list?  E.g. change the background color?  If we want to keep it simple, I'd say no, the above point about mode reordering is enough.

This generalization could integrate the glyph feature.

Thoughts?

edgar-vincent:
> Thank you for your contribution!
> If I'm not mistaken, this is your first contribution, so welcome to the world of Nyxt hackers :)

Thank you!

> You are addressing an important point: per-mode status buffer customization.
> Unfortunately we can't hard-code the mode customization. Instead, we should derive a framework that's extensible and where the core functions are mode-agnostic.

I was wondering about that, and you're absolutely right. How to do you think this should be done?
This is somewhat unrelated, but what about styling? The styling definitions seem to be scattered in different places (moreover, shouldn't the blue "accent" colour be inherited from somewhere rather than re-specified every time it is used?).

> If you look at `format-status-modes`, you'll see that it calls `glyph` which is a method that can be specialized for each mode.

I'd missed that! Thanks.

> This is a bit limited and I believe we can do better. In particular:
> 
>     * Change the appearance of the HTML span of the mode button in the status bar.

I agree. It is probably a good thing, especially for "visual" people (which I'm not BTW ^^). For example, it should be very obvious that one is using visual-mode or, indeed, vi-insert-mode. Quite a few people have been confused with the latter, even though the mode is listed in the status buffer.
 
>     * A mode should be able to tell whether it should appear first.  How do we resolve conflicts then?

Specifically about the vi-mode indicator, I think it should probably in the left, like in powerline. Perhaps there could be several "mode" areas in the status buffer? Tridactyl has it on the right though, but I think that vim users would expect it to be on the left.
Not sure about conflicts.

>     * Should a mode be able to tweak the appearance of the whole mode list?  E.g. change the background color?  If we want to keep it simple, I'd say no, the above point about mode reordering is enough.

I can't think of why a mode would need to change the appearance of the whole list, but it might be useful.

Ambrevar:
> I was wondering about that, and you're absolutely right. How to do you
> think this should be done?

I'm not completely sure yet, but I think a generalized version of
format-status-modes seems closer to the right direction.

> This is somewhat unrelated, but what about styling? The styling
> definitions seem to be scattered in different places

"Functional" theming has been on my roadmap for a while.
We've already started laying down the foundations: most styling elements
are found in a slot that's consistently called `style`.

It should be quite easy to go from there.

> (moreover, shouldn't the blue "accent" colour be inherited from
> somewhere rather than re-specified every time it is used?).

Which "blue accent colour" do you mean?

> Specifically about the vi-mode indicator, I think it should probably
> in the left, like in powerline.

This is a separate issue, and not really an issue.
With the WINDOW `status-formatter` slot you can fully customize the
appearance of the status area, including the position of the different parts.

> Perhaps there could be several "mode" areas in the status buffer?

By default I'd rather have just one mode area, but we can discuss it of
course.

And obviously, styling and custom status bars would make for great
extensions! :)

edgar-vincent:
> I'm not completely sure yet, but I think a generalized version of format-status-modes seems closer to the right direction.

> "Functional" theming has been on my roadmap for a while. We've already started laying down the foundations: most styling elements are found in a slot that's consistently called `style`. It should be quite easy to go from there.

Okay - I trust the expert :)


> Which "blue accent colour" do you mean?

I mean `#37a8e4` as in https://github.com/atlas-engineer/nyxt/blob/d50f819372f8c781147ad52fe4d31e020ae50168/source/prompt-buffer.lisp#L111 and https://github.com/atlas-engineer/nyxt/blob/17f7b7f2d903082b5b0308b07cdad8311a81dadc/source/web-mode.lisp#L47 :)


> This is a separate issue, and not really an issue. With the WINDOW `status-formatter` slot you can fully customize the appearance of the status area, including the position of the different parts.

I mean I think it should be there by default :) But I suppose if we make it more obvious that `vi-mode` is enabled by other means, I imagine it would be fine :)

Ambrevar:
Related issue: https://github.com/atlas-engineer/nyxt/issues/1024

Ambrevar:
OK to merge this as an intermediary solution until we come up with something more general.
Let's just unexport the `format-*` function.

Ambrevar:
Rebased and merged with 08b36b4ae42db1209798ea9c160e53ee11dc3f9d.
I've fixed a few things to accommodate with the recent changes in the status bar CSS.

Thanks a lot for this very demanded change, sorry for the delay and thanks for your patience!

jmercouris:
Pierre, this is broken with no init, AKA people who do not use vi mode... I will fix it.

Ambrevar:
Oops, I was thinking I should test this case, then I forgot... :(

jmercouris:
It's OK, it happens to the best of us :-)

Issue #1355, opened 2021-05-03
Non-web-buffer (editor, lisp REPL) should re-use bindings


Ambrevar:
Currently enabling VI or Emacs bindings for web-browsers does not enable them for other buffers, like the editor or the Lisp REPL.

Maybe we should make that more "do what I want".  Thoughts?

aadcg:
I think we might want to discuss smth more general. The modes we have (which we can think of them as `minor-modes` in Emacs' parlance), could be local or global. What we have, afaik, are local minor modes.

Indeed, some modes (like this one) could be global, i.e. enabled for all buffers.

Ambrevar:
Indeed, this could be one way to handle it.
Sadly I'm not sure we have time to introduce it before 2.0.

Another option would be to have `nyxt::lisp-repl', `make-editor-buffer`
or `editor-buffer`, etc., _append to the list of modes / default modes_
on instantiation.

This would, it would properly inherit from the `buffer` default-modes which would result
in the intended behaviour.

Does that make sense?

aadcg:
Given the time constraint, that seems a sound way to approach it.

edgar-vincent:
Note: this should include the prompt buffer :)

Ambrevar:
> In other words, allow the user to configure the general case, and optionally override it in the more specific case. What do you think?

This is sound, but isn't it what we already do?

The problem is that web-buffers, prompt-buffers, REPL buffers, etc. all
_specify_ their own default modes, thus overriding the default.

jmercouris:
I've been thinking about this, and I've come up with no resolution beyond removing the default-modes.

jmercouris:
That is not to say I've not come up with other solutions, but they've been quite unsatisfactory...

jmercouris:
Perhaps we need another new slot like `user-modes` or something?

Ambrevar:
> Perhaps we need another new slot like `user-modes` or something?

I think that would only postpone the problem.  Example:

- A set default-modes.
- B inherits from A and sets user-modes
- C inherits from B.

At this point, C must set either default modes or user modes, thus
overriding the user configuration of B or A.

I have an idea in mind, I'll send a proposal later.

Ambrevar:
See #1411 for my proposal using regular CLOS.

jmercouris:
I believe this has been fixed :-)

PR #1411, opened 2021-05-15
Replace default-modes with initialize-modes method.

### LINKED ISSUE #1383: Make VI and Emacs mode work in follow-hint & friends
It's about `element-hint-mode` that we hard-code in `query-hints` (element-hint.lisp, L133):
``` lisp
(let ((result (prompt
                   :prompt prompt
                   ;; Here!         vvvvvvvvvvvvvvvvv
                   :default-modes '(element-hint-mode prompt-buffer-mode)
                   :history nil
                   :sources
                   (make-instance
                    'hint-source
                    :constructor (add-element-hints))
                   :after-destructor
                   (lambda ()
                     (with-current-buffer buffer
                       (remove-element-hints))))))
      (funcall function result))
```

This overrides any mode configuration one might have had for `prompt-buffer` including keybindings.

How do we go around this? MOP? 

This can be a wider problem, though: how do we add a specific mode to the prompt-buffer while retaining user configuration? Maybe add an `:additional-modes` argument to `prompt` and use it instead of `:default-modes`?

Ambrevar:
Fixes #1383, #1355.

The user cannot use `define-configuration` because the `default-modes` slot is not longer writable.

Instead, you are expected to use regular CLOS.  Example to enable VI bindings everywhere:

```lisp
(defmethod initialize-modes :after ((buffer buffer) &key)
  (nyxt::make-mode 'vi-normal-mode buffer))
```

Pretty cool, nah?

Now why can't we leverage `define-configuration`?  Because this macro works over the initforms of user classes.

Here what we want is different: we want a given slot list value to cascade down the subclasses.  CLOS is perfect for that thanks to `call-next-method`.

EDIT: This implements the technique mentioned in #1005, but in a different context.  So this is useful ground work for #1005.

### DIFF
--- a/source/auto-mode.lisp
+++ b/source/auto-mode.lisp
@@ -230,7 +230,7 @@ The rules are:
 (defun initialize-auto-mode (mode)
   (unless (last-active-modes mode)
     (setf (last-active-modes mode)
-          (mode-invocations (default-modes (buffer mode)))))
+          (mode-invocations (modes (buffer mode)))))
   (when (prompt-on-mode-toggle mode)
     (hooks:add-hook (enable-mode-hook (buffer mode))
                     (nyxt::make-handler-mode
--- a/source/blocker-mode.lisp
+++ b/source/blocker-mode.lisp
@@ -131,8 +131,8 @@ Example:
   \"Blocker mode with custom hosts from `*my-blocked-hosts*'.\"
   ((nyxt/blocker-mode:hostlists (list *my-blocked-hosts* nyxt/blocker-mode:*default-hostlist*))))
 
-\(define-configuration buffer
-  ((default-modes (append '(my-blocker-mode) %slot-default%))))"
+\(defmethod initialize-modes :after ((buffer buffer))
+  (make-mode 'my-blocker-mode buffer))"
   ((hostlists (list *default-hostlist*))
    (destructor
     (lambda (mode)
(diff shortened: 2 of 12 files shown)


Ambrevar:
To do:
- Export `make-mode`.
- Test if `element-hint-mode` works as expected with vi-bindings.

Ambrevar:
Question for post 2.0: Can we leverage the same technique to replace `define-configuration`?

Ambrevar:
The drawback of this approach is that we don't encode the configuration as data, but as functions.

Here, instead of a list of mode symbols, we have a specialized method that enables the desired mode. 

This has downsides for debugging, inspection, etc.

Ambrevar:
Another option is to define our own `default-mode` generic function using an `append` method combination.  Example:

```lisp
(defgeneric default-modes (obj)
  (:method-combination append)
  (:method append ((buffer buffer))
    '(base web))
  (:method append ((buffer web-buffer))
    '(base web certificate-exception))
  (:method append ((buffer prompt-buffer))
    '(prompt-buffer-mode)))
```

This is very neat and then the user would keep using `define-configuration`.

But how do we exclude modes from the superclasses then?
In the above, `prompt-buffer` would be wrongly initialized with `base` and `web` modes.

One option is to rething our buffer hierarchy, i.e. something like this:

- buffer (no modes)
  - internal-buffer (base-mode web-mode)
    - web-buffer (certificate-exception-mode)
  - prompt-buffer (prompt-buffer-mode)

Then adding emacs-mode to `buffer` would work.

Ambrevar:
See #1413.

I'd rather drop this one in favour of #1413.

jmercouris:
Concurred, closing in favor of #1413

PR #1413, opened 2021-05-16
Combine-append buffer default-modes.

Ambrevar:
This is an alternative to #1411.

It uses an `append` method combination.
Unlike what I commented in #1411, this does not allow for using `define-configuration` either.  But still, it's cleaner and less intrusive than #1411 I believe, so I'm in favour of this one.

As per `define-configuration` support, we could expand the macro to add support for generic functions.

For instance,

```lisp
(define-configuration buffer
  ((default-modes '(foo)))
```

would automatically detact that default-modes is a defgeneric with a non-standard method combination and thus expand to

```lisp
(defmethod default-modes append ((buffer buffer)
  '(foo)))
```

Thoughts?

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -30,12 +30,6 @@ See the `data-path' class and the `expand-path' function.")
 Modes are instantiated after the `default-modes' slot, with `initialize-modes'
 and not in the initform so that the instantiation form can access the
 initialized buffer.")
-   (default-modes '()
-                  :writer t
-                  :export t
-                  :type list-of-symbols
-                  :documentation "The symbols of the modes to instantiate on buffer creation.
-The mode instances are stored in the `modes' slot.")
    (enable-mode-hook (make-hook-mode)
                      :type hook-mode
                      :documentation "Hook run on every mode activation,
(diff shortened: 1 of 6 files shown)


Ambrevar:
Great news, I've just implemented support for `define-configuration`.

What's really cool about it is that it should be syntactically backward compatible, meaning that the user config should now behave as expected, without change!

Ambrevar:
Element-hint is now also working properly.

This does it I believe :)
Please confirm that it works with your configuration (hopefully without change), and this should be ready to merge.

Ambrevar:
As with #1411, this fixes #1383 and #1355.

jmercouris:
Let me give this  a try. It is a most interesting proposal!

jmercouris:
Well, it does work with my configuration! That is good news. This is very very clever, and that concerns me a little bit. I must agree with you though, this is a better solution that #1411.

I would like to talk about something you hinted at though: the problem of potentially having a `user-modes` slot. You talked about this being overridden in the hierarchy of inheritance. However, if we do not ever set this slot, and confirm that it is for user use only, would that not be OK?

PR #1416, opened 2021-05-17
Global commands

### LINKED ISSUE #1415: Defining global commands
Currently, global commands are commands that are defined in the `nyxt` or `nyxt-user` packages.

There is one huge downside to this approach: extensions cannot define global commands without polluting the `nyxt` namespace, which would inevitably lead to conflicts (this is why we don't use `:use` after all -- except that here it's worse).

Last but not least, `list-commands` is very ugly code that was broken a few times at least.

So instead of this crazyness, I suggest 
- we don't make `base-mode` special;
- we add `define-global-command` (or `define-command-global`, or `define-command*`?) that "marks" the command as global.
- We make mode togglers global.

To "mark" the command as global, I see a few simple options:

- Add a `*global-command-list*`.
- Set the command symbol property (from the symbol property-list).
- Add a `global-p` slot to the `command` class.

I'd go for the latter since it's more consistent with our style.

Thoughts?

Ambrevar:
Fixes #1415.

Discussion: The last code commit changes the way commands are seen from `nyxt-user`: unless `define-command-global` is used, they user commands won't be listed in `execute-command`.

This is consistent and what we want I believe, but it breaks backward compatibility.
I believe it'd be good to merge this before 2.0.


source/command.lisp
```diff
@@ -166,6 +175,16 @@ Example:
...
 
+(export-always 'define-command-global)
+(defmacro define-command-global (name (&rest arglist) &body body)
+  "Like `define-command' but mark the command as global.
+This means it will be listed in `command-source' when the global option is on.
+This is mostly useful for third-party packages to define globally-accessible
+commands without polluting Nyxt packages."
+  `(progn
+    (define-command ,name (,@arglist) ,@body)
+    (setf (global-p (find-command ',name)) t) ))
```
jmercouris, line 186: extra space in the parens.

PR #1419, opened 2021-05-17
Replace some commands / options with prompt-buffer actions

Ambrevar:
First commit replace focus-on-reopended-buffer-p with an action.
It also dusts of the old recent-buffer code.

More commits to come where I'll move bookmark and history related commands to actions.

### DIFF
--- a/source/base-mode.lisp
+++ b/source/base-mode.lisp
@@ -27,7 +27,6 @@ This mode is a good candidate to be passed to `make-buffer'."
                      "C-r" 'reload-current-buffer
                      "C-R" 'reload-buffers
                      "C-m o" 'set-url-from-bookmark
-                     "C-m C-o" 'set-url-from-bookmark-new-buffer
                      "C-m s" 'bookmark-current-url
                      "C-d" 'bookmark-current-url
                      "C-m C-s" 'bookmark-buffer-url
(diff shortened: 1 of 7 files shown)


Ambrevar:
- History-forwards-query: Use all sources from         `history-forwards-direct-children` and `history-forwards-all`.

- `history-backwards-query`: Use all sources from        `history-backwards-query` and `history-all`.

- `history-forwards-maybe-query`: Same but use `history-forwards-query`        instead.

Ambrevar:
Bookmark  command replaced.

jmercouris:
I still don't think this makes sense with the emacs syntax. Well, anyways, we shall discuss!

Ambrevar:
Note to myself: Add an `action` parameter to set-url-from-bookmark, so that the user can easily bind to a command with a specialized action.

jmercouris:
I will work on this, OK?

Ambrevar:
Please do :)

jmercouris:
I don't really like the way it works with actions as a key argument. Unwieldy!

jmercouris:
I've tried a couple of ways, and both are feeling wrong.

jmercouris:
Well, I will remove it for now...

jmercouris:
Actually, I will keep the actions as the key argument.

Issue #1425, opened 2021-05-18
Cannot configure buffer-source 


franburstall:
**Describe the bug**
I try to set the slot `follow-p` of `buffer-source` to nil by hitting the *Configure* button for that slot after `C-h C buffers-source`.  This writes the following form to `auto-config.lisp`:
````common-lisp
(DEFINE-CONFIGURATION BUFFER-SOURCE
  ((FOLLOW-P nil)))
````
I restart nyxt and this silently fails to have an affect.  Well, almost silently: there is nothing in the messages buffer but you can see a truncated attempt to complain in the terminal output:
````
<INFO> [17:41:23] Loading Lisp file "/home/fran/.config/nyxt/auto-config.lisp".
While evaluating the form starting at line 1, column 0
  of <INFO> [17:41:23] Loading Lisp file "/home/fran/.config/nyxt/init.lisp".
SLYNK's ASDF loader finished.<INFO> [17:41:26] Loading history of 636 URLs from "/home/fran/.local/share/nyxt/history/default.lisp".
````
If I put the same form in my `init.lisp` instead, the error is apparent (and predictable): I get:
````
Error occured:

Could not load the init file: define-configuration argument USER-BUFFER-SOURCE is not a known class.
````
(though still nothing in the message buffer and truncated output on the terminal).

Looking at the code, I see that I cannot configure class `foo` unless class `user-foo` exists and this is not the case for many values of `foo`!  (Have I misunderstood something here?)

**Information**
- OS name+version: Debian-10
- Graphics card and driver:
- Desktop environment / Window manager name+version: EXWM
- How you installed Nyxt (Guix pack, package manager, build from source): Source
- Information from command copy-system-information:

Nyxt version: 2-pre-release-7-98-g0248af96
Renderer version: GI-GTK
Operating system kernel: Linux 4.19.0-16-amd64
Lisp implementation: SBCL 2.1.1
Features: (SLYNK WEBKIT2 WEBKIT2-2.30 WEBKIT2-PASTE-PLAINTEXT WEBKIT2-TRACKING
 WEBKIT2-MUTE WEBKIT2-EMOJI WEBKIT2-MEDIA WEBKIT2-SANDBOXING GTK-3-22 GTK-3-20
 GTK-3-18 GTK-3-16 GTK-3-14 GTK-3-12 GTK-3-10 GTK-3-8 GTK-3-6 GTK-3-4 GTK
[...]

PR #1428, opened 2021-05-19
Fix configure slot

Ambrevar:


Ambrevar:
Fixes #1425.

Ambrevar:
These `define-user-class` are annoying to maintain.

Instead, I think we should automatically define the user class when the class is exported.

Our `define-class` is currently a copy-paste of `defclass*`.

Ambrevar:
But this change belongs to #1102 I think.  For after 2.0.

jmercouris:
I still would prefer a class option rather than an implication that an exported class is to be configured

jmercouris:
Tested, thank you!

Issue #1094, opened 2021-01-07
Block and redirect internal-buffer web queries properly


Ambrevar:
Currently when we try loading an external resource from an internal buffer, the query is intercepted in `request-resource`, blocked and a new web-buffer is spawned to load that query.

I see 2 problems here:

- `request-resource` is a candidate for `request-resource-hook`, which means it's optional and can be removed.  If the user decides to replace it with a function of theirs which does not do the dispatch, they might end up with an internal buffer showing an external URL and thus allowing `lisp://` schemes, which is a vulnerability.

- `notify::uri` is sent _before_ `request-resource-hook` is triggered.  
  Example problem: say you are on the help buffer and `set-url` http://example.org.  Then `notify::uri` will be emitted, the new URL will be set and `on-signal-notify-uri` will be called for the internal buffer!  It's only after the query is blocked that the URL is reset.

To fix two birds with one stone, I suggest we move the internal-buffer blocking to either `notify::uri` or a signal that could be caught even before that.
The problem with `notify::uri` is that it cannot block the loading of the resource.  So if there is no better signal before `notify::uri`, then I suggest we handle external URLs in `notify::uri` to do nothing, and block the query in `on-signal-decide-policy` before the `request-resource-hook` is run.

Thoughts?

aartaka:
`webkit_web_context_register_uri_scheme`? It can block requests, return any type of data, and it's a conventional way of custom schemes handling. I just can't stop pushing this idea xD

Needs `GInputStream`, though (as mentioned in https://github.com/joachifm/cl-webkit/pull/46).

jmercouris:
Would it also be fine to just include everything in `on-signal-decide-policy` and ignore `notify::uri`? If `notify::uri` has no control over the flow, why would we even involve it at all?

Ambrevar:
John Mercouris <notifications@github.com> writes:

> Would it also be fine to just include everything in `on-signal-decide-policy` and ignore `notify::uri`? If `notify::uri` has no control over the flow, why would we even involve it at all?

Because it's the only reliable way to get signaled that the URL
changed.  In particular, when you click on an HTML anchor or when the
URL changes via Javascript (like on GitHub).

Ambrevar:
Artyom Bologov <notifications@github.com> writes:

> `webkit_web_context_register_uri_scheme`? It can block requests, return any type of data, and it's a conventional way of custom schemes handling. I just can't stop pushing this idea xD
>
[...]

PR #1429, opened 2021-05-19
Force request-resource on every request, don't allow users to customize it out

Ambrevar:
Addresses the most urgent part of #1094, namely to force-run `request-resource`.

Breaking news: Scheme handling is done _after_ `on-decide-policy`, so unfortunately @aartaka's approach won't cut it to fix the load order problem.  As such, I've decided to not handle the schemes at all since it seemingly does not add any benefit compared to the `on-decide-policy` signal.
This is not critical for security though.

So I guess the only way forward to properly intercept requests before the load starts is to leverage `send-request` from a WebKit extension, as mentioned here: https://github.com/atlas-engineer/nyxt/issues/1094#issuecomment-832454345.

Anyways, this pull request should be good enough for 2.0.  Thoughts?

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -394,10 +394,8 @@ view.")
   (:export-accessor-names-p t)
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name)))
 
-(export-always 'request-resource)
-(defun request-resource (request-data)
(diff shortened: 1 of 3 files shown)


Ambrevar:
Please test with your config and modes that leverage the request-resource-hook, like
- auto-mode
- blocker-mode
- force-https-mode

PR #1448, opened 2021-05-23
Update SOURCES.org

mjsir911:
A few of the packages under "Renderer-specific dependencies" are vendored too.

Additionally, `osicat` is mentioned in a few places with `#-sbcl` prefixing it. Does this mean install on non-sbcl installs? Should this be documented?

### DIFF
--- a/documents/SOURCES.org
+++ b/documents/SOURCES.org
@@ -12,6 +12,7 @@ which external dependencies we are directly relying on.
 - bordeaux-threads
 - calispel
 - cl-css
+- cl-containers
 - cl-custom-hash-table
 - cl-html-diff
 - cl-json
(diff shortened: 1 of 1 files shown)


aartaka:
> A few of the packages under "Renderer-specific dependencies" are vendored too.

Let's leave them where they are :)

> Additionally, `osicat` is mentioned in a few places with `#-sbcl` prefixing it. Does this mean install on non-sbcl installs? Should this be documented?

SBCL is the only implementation we officially support (Nyxt on CCL is still somewhat unstable AFAIK), so I'm not sure about that. Let's not include it for now.

aartaka:
Thank you for this patch! :)

Ambrevar:
CCL works fine from the REPL, but the binary fails to start.

PR #1451, opened 2021-05-24
Fixes xdg-open from using the improper application when opening a downloaded file

mariari:
Nyxt uses the `namestring` function which does the following

```lisp
NYXT-USER> (namestring #p"file:///home/katya/Downloads/124065.pdf")
"file:/home/katya/Downloads/124065.pdf"
```

When this is fed into xdg-open it uses the `x-www-browser` application instead of the proper pdf application. the following code fixes that.

```lisp
NYXT-USER> (inject-proper-xdg-name (namestring #p"file:///home/katya/Downloads/124065.pdf"))
"file:///home/katya/Downloads/124065.pdf"
```

### DIFF
--- a/source/file-manager.lisp
+++ b/source/file-manager.lisp
@@ -94,6 +94,13 @@ See `supported-media-types' of `file-mode'."
 #+linux
 (defvar *xdg-open-program* "xdg-open")
 
+(defun inject-proper-xdg-name (name-file)
+  "injects the proper xdg name for files. Common lisp gives back file:/ from a
+namestring, however for xdg-open to properly function it must have file:///"
+  ;; check for zerop to assure we don't have an edge case in the file name
+  (if (zerop (search "file:/" name-file))
+      (format nil "file:///~a" (subseq name-file (length "file:/")))))
+
 (export-always 'default-open-file-function)
 (defun default-open-file-function (filename &key supported-p new-buffer-p)
   "Open FILENAME in Nyxt if supported, or externally otherwise.
@@ -113,7 +120,7 @@ Can be used as a `open-file-function'."
               (buffer-load (quri::make-uri-file :path filename)))
           (uiop:launch-program
            #+linux
-           (list *xdg-open-program* (namestring filename))
+           (list *xdg-open-program* (inject-proper-xdg-name (namestring filename)))
            #+darwin
            (list "open" (namestring filename))))
     ;; We can probably signal something and display a notification.
--- a/source/file-manager.lisp
+++ b/source/file-manager.lisp
@@ -94,14 +94,6 @@ See `supported-media-types' of `file-mode'."
 #+linux
 (defvar *xdg-open-program* "xdg-open")
 
-(defun inject-proper-xdg-name (name-file)
-  "injects the proper xdg name for files. Common lisp gives back file:/ from a
-namestring, however for xdg-open to properly function it must have file:///"
-  ;; check for zerop to assure we don't have an edge case in the file name
-  (if (zerop (search "file:/" name-file))
-      (format nil "file:///~a" (subseq name-file (length "file:/")))
-      name-file))
-
 (export-always 'default-open-file-function)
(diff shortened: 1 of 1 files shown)


aartaka:
Thank you for this contribution! I've changed the printing to the `quri`-provided `render-uri`, but it all wouldn't be possible if not for your initiative :D

mariari:
> Thank you for this contribution! I've changed the printing to the `quri`-provided `render-uri`, but it all wouldn't be possible if not for your initiative :D

It seems that doesn't solve the issue I've hooked up some paramaters outside of the code

```lisp

            (progn
              (setf nyxt-user::*file-name* filename)
              (setf nyxt-user::*test* file-url)
              (echo (quri:render-uri file-url))
              (uiop:launch-program
               #+linux
               (list *xdg-open-program* (quri:render-uri file-url))
```

When run on `ugh.pdf` I get the following

```lisp
NYXT-USER> *test*
#<QURI.URI.FILE:URI-FILE file://file:///home/katya/Downloads/ugh.pdf>

NYXT-USER> (quri.uri.file:make-uri-file :path *file-name*)
#<QURI.URI.FILE:URI-FILE file://file:///home/katya/Downloads/ugh.pdf>
NYXT-USER> (quri.uri.file:make-uri-file :path #p"file:///home/katya/Downloads/ugh.pdf")
#<QURI.URI.FILE:URI-FILE file://file:/home/katya/Downloads/ugh.pdf>
NYXT-USER> (quri:render-uri (quri.uri.file:make-uri-file :path #p"file:///home/katya/Downloads/ugh.pdf"))
"file://file:/home/katya/Downloads/ugh.pdf"
```

I didn't realize it, but the input coming into this file isn't a #P"...", but rather a plain string, so the following works

```lisp
(defun default-open-file-function (filename &key supported-p new-buffer-p)
  "Open FILENAME in Nyxt if supported, or externally otherwise.
FILENAME is the full path of the file (or directory).

See `supported-media-types' to customize the file types that are opened in
Nyxt and those that are opened externally.

NEW-BUFFER-P defines whether the file/directory is opened in a new buffer.
SUPPORTED-P says whether the file can be opened by Nyxt.

Can be used as a `open-file-function'."
  (let ((file-url (quri::make-uri-file :path filename)))
    (handler-case
        (if supported-p
            (if new-buffer-p
                (make-buffer-focus :url file-url)
                (buffer-load file-url))
            (uiop:launch-program
             #+linux
             (list *xdg-open-program* filename)
             #+darwin
             (list "open" (namestring filename))))
      ;; We can probably signal something and display a notification.
      (error (c) (log:error "Opening ~a: ~a~&" filename c)))))
```

Should I open another PR with this change?

aartaka:
Wait a second, what are you trying to do in the first place? Can you please provide a recipe for the problem that bothers you? Do you do `open-file` and input a `file://...` url afterwards?

mariari:
> Wait a second, what are you trying to do, in the first place? Can you please provide a recipe for the problem that bothers you? Do you do `open-file` and input a `file://...` url afterwards?

The input `default-open-file-function` is `"file:///home/katya/Downloads/ugh.pdf"` from the `*Downloads` buffer. When I was first investigating this problem I thought it was a file-path because of the `namestring` call.

However when you call 

```lisp
(quri:render-uri (quri.uri.file:make-uri-file :path *file-name*))
```

it adds an extra file:// to the link

```lisp
"file://file:///home/katya/Downloads/ugh.pdf"
```

Steps to reproduce:

have nyxt open: http://web.mit.edu/~simsong/www/ugh.pdf

Go to the downloads buffer and click the open file button.

it shouldn't open anything as the file shouldn't exist with the extra file:/ in it

aartaka:
Oh, I see now. You're right, `default-open-file-function` should be smarter about this. 5fdabca9023628bbc0109aaf6ca0e5de1651646e should fix that. However, the way we represent files is something to [think about](https://github.com/atlas-engineer/nyxt/issues/1454). Thanks for this revelation :)

Issue #1138, opened 2021-02-09
[FR] Integrate autofills and password management


edgar-vincent:
It would be particularly handy if one could combine autofills with the password manager integration. I'm thinking of something similar to https://github.com/browserpass/browserpass-extension, which inserts logins and passwords in the correct fields at the press of a keychord. Entries are identified by their names, which match a given site's URL.

If I'm not mistaken, one cannot currently copy usernames in Nyxt, for instance, which is a bit of a drag.

jmercouris:
Correct, and we have a autofill issue about that, but I think we should provide some sort of abstraction for this integration!

fabian-thomas:
## Options

I see 4 options here:

### Using the existing autofill framework
One could add one autofill target for password and username respectively. Then the user could focus those fields manually and autofill the pw/username. This is already possible with the following really hacky snippet:
```
(make-autofill :name "username" :fill (lambda () (copy-username) (trivial-clipboard:text)))
```
The hacky part could easily be fixed by providing some get-username and get-password function in the APIs. This has the downside of having to manually select both username and password fields and triggering an autofill for both. The worst part about this triggering twice, is that one has to select the correct account twice. To fix this latter part a hacky solution until this is implemented nicely is:
```
(make-autofill :name "both" :fill (lambda () (copy-username) (let* ((username (trivial-clipboard:text))) (copy-password) username)))
```
Let the autofill framework fill the username, and copy the password to clipboard.

### Extending the autofill framework to support element matching
We could implement some way to specify multiple autofill targets in one and specify the HTML element that should be filled by XPath or Jspath. This would be the cleanest, but worst to maintain solution. This is what the browser extensions like bitwarden do.

### Misusing the existing autofill framework
As I understand it, currently one autofill target just fills/pastes the data that is specified into the currently selected field. That means we can't fill two fields nicely. As with the above case, we could extend the autofill framework to perform multiple fills with actions in between. I think about, fill username, then fake a tab key press, then fill password. I know this sounds hacky again, but that is what I used with qutebrowsers [`qute-pass` userscript](https://github.com/qutebrowser/qutebrowser/blob/65aa61f33fbada24e07a30fe78b69232573aa069/misc/userscripts/qute-pass) for over one year now, and it worked like a charm. I had 3 keybindings: one for username only and password only fill, and one for filling both. Like that I could fill any login form pretty easily.

### Above option but not misusing the autofill framework
[...]

PR #1455, opened 2021-05-25
Add username copying to password manager.

aartaka:
This adds username copying to KeePassXC and Password Store interfaces, together with a generic to implement for other interfaces. A step towards #1138.

EDIT: `command` is installed on my machine, but isn't installed in development environment xD
EDIT: Mention #1138.

### DIFF
--- a/libraries/password-manager/password-pass.lisp
+++ b/libraries/password-manager/password-pass.lisp
@@ -39,13 +39,26 @@
   (execute password-interface (list "show" "--clip" password-name)
     :output '(:string :stripped t)))
 
+(defmethod clip-username ((password-interface password-store-interface) &key password-name service)
+  (declare (ignore service))
+  (let ((username-line (1+ (position "[Ll]ogin|[Uu]ser(name)?:"
+                                     (ppcre:split "\\n"
+                                                  (execute password-interface
+                                                      (list "show" password-name)
+                                                    :output '(:string :stripped t)))
+                                     :test #'ppcre:all-matches))))
+    (execute password-interface (list "show" (format nil "--clip=~d" username-line) password-name)
+      :output '(:string :stripped t))))
+
 (defmethod save-password ((password-interface password-store-interface)
-                          &key password-name password service)
+                          &key password-name username password service)
   (declare (ignore service))
   (if (str:emptyp password)
       (execute password-interface (list "generate" password-name))
-      (with-open-stream (st (make-string-input-stream password))
-        (execute password-interface (list "insert" "--echo" password-name)
+      (with-open-stream (st (make-string-input-stream (format nil "~a~&username:~a"
+                                                              password
+                                                              username)))
+        (execute password-interface (list "insert" "--multiline" password-name)
           :input st))))
 
 (defmethod password-correct-p ((password-interface password-store-interface))
--- a/libraries/password-manager/password-pass.lisp
+++ b/libraries/password-manager/password-pass.lisp
@@ -39,16 +39,48 @@
   (execute password-interface (list "show" "--clip" password-name)
     :output '(:string :stripped t)))
 
+(defvar *multiline-separator* ": *"
+  "A regular expression to separate keys from values in the `pass' multiline format.")
+
+(defun parse-multiline (content)
+  "Return an alist of the multiple entries.
+An entry is a sequence of
+- a key string,
+- a colon,
+- optional spaces,
+- a value string.
+
+This is meant to handle the organization suggestion from
+http://www.passwordstore.org/#organization.
+
+Lines that don't match the format are ignored.
+The first line (the password) is skipped."
+  (let ((lines (str:split (string #\newline)
+                          content)))
+    (delete nil (mapcar (lambda (line)
+                          (let ((entry (ppcre:split *multiline-separator* line :limit 2)))
+                            (when (= 2 (length entry))
+                              entry)))
+                        ;; Skip first line to ignore password:
+                        (rest lines)))))
+
+(defvar *username-keys* '("login" "user" "username")
+  "A list of string keys used to find the `pass' username in `clip-username'.")
+
 (defmethod clip-username ((password-interface password-store-interface) &key password-name service)
+  "Save the multiline entry that's prefixed with on of the `*username-keys*' to clipboard.
+Case is ignored.
+The prefix is discarded from the result and returned."
   (declare (ignore service))
-  (let ((username-line (1+ (position "[Ll]ogin|[Uu]ser(name)?:"
-                                     (ppcre:split "\\n"
-                                                  (execute password-interface
-                                                      (list "show" password-name)
-                                                    :output '(:string :stripped t)))
-                                     :test #'ppcre:all-matches))))
-    (execute password-interface (list "show" (format nil "--clip=~d" username-line) password-name)
-      :output '(:string :stripped t))))
+  (let* ((entries (parse-multiline (execute password-interface
+                                       (list "show" password-name)
+                                     :output '(:string :stripped t))))
+         (username-entry (some (lambda (key)
+                                 (find key entries :test #'string-equal :key #'first))
+                               *username-keys*)))
+    (when username-entry
+      (trivial-clipboard:text (second username-entry))
+      (second username-entry))))
 
 (defmethod save-password ((password-interface password-store-interface)
                           &key password-name username password service)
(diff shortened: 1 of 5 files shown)


aartaka:
Any `pass` users to test it?

aartaka:
Important pointers to learn from:
[Browserpass conventions for passwords and usernames](https://github.com/browserpass/browserpass-extension#organizing-password-store)
[auth-source-pass](https://github.com/DamienCassou/auth-source-pass)
[passff](https://github.com/passff/passff#readme)

edgar-vincent:
Just to keep a note of this: in its current state, `copy-username` copies the entire line containing the username, i.e `login:  mylogin` instead of `mylogin`.

libraries/password-manager/password-security.lisp
```diff
@@ -29,4 +29,8 @@
                         "-a" password-name "-s" service "-g")
                   :error-output '(:string :stripped t)))))))
 
+(defmethod clip-username ((password-interface security-interface) &key password-name service)
+  (declare (ignore password-name service))
+  (error "Username clipping is not supported by security interface."))
```
Ambrevar, line 34: Ambiguous use of "security" here.
Maybe `by the 'security' interface`?

libraries/password-manager/password.lisp
```diff
@@ -23,9 +23,15 @@
 (defgeneric clip-password (password-interface &key password-name service)
   (:documentation "Retrieve specific password by name."))
 
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (export 'clip-username))
```
Ambrevar, line 27: We should probably define the `export-always` macro locally at this point :)

libraries/password-manager/password-pass.lisp
```diff
@@ -39,13 +39,26 @@
...
     :output '(:string :stripped t)))
 
+(defmethod clip-username ((password-interface password-store-interface) &key password-name service)
+  (declare (ignore service))
+  (let ((username-line (1+ (position "[Ll]ogin|[Uu]ser(name)?:"
+                                     (ppcre:split "\\n"
+                                                  (execute password-interface
+                                                      (list "show" password-name)
+                                                    :output '(:string :stripped t)))
+                                     :test #'ppcre:all-matches))))
```
Ambrevar, line 49: username-line may be NIL, `1+` will raise a condition then.
Furthermore, the rest of the body should not be run if `username-line` is NIL.

libraries/password-manager/password-pass.lisp
```diff
@@ -39,13 +39,26 @@
   (execute password-interface (list "show" "--clip" password-name)
     :output '(:string :stripped t)))
 
+(defmethod clip-username ((password-interface password-store-interface) &key password-name service)
+  (declare (ignore service))
+  (let ((username-line (1+ (position "[Ll]ogin|[Uu]ser(name)?:"
```
Ambrevar, line 44: The username maybe not be prefixed with anything at all, and just be on the second line by convention.  I suggest you default to `2` if there is no result.  Bear in mind there may be no second line.

libraries/password-manager/password-pass.lisp
```diff
@@ -39,13 +39,26 @@
...
+                                     (ppcre:split "\\n"
+                                                  (execute password-interface
+                                                      (list "show" password-name)
+                                                    :output '(:string :stripped t)))
+                                     :test #'ppcre:all-matches))))
+    (execute password-interface (list "show" (format nil "--clip=~d" username-line) password-name)
+      :output '(:string :stripped t))))
```
Ambrevar, line 51: You need to strip the matching "[Ll]ogin|[Uu]ser(name)?:" as reported by @edgar-vincent.

Ambrevar:
Great addition, thanks!

What about going even more generic and add a command to copy _any_ field?
In the case of `pass`, it could simply parse the entry and offer each line as a suggestion in the prompt buffer.
If a line has `:` or `=`, we could offer to copy only the right-hand side.

libraries/password-manager/password-pass.lisp
edgar-vincent: I don't think having an unprefixed field is encouraged: what if an entry doesn't have a user associated to it but does have a port or a URL, stored on the second line? Moreover,  a password could contain a `:` or a `=`. I think we should stick to a specific organisational style, like what `auth-source-pass` for Emacs uses.

Ambrevar:
Shall we fix and merge this before the next release? @aartaka Do you have time to fix it?  Let me know if you need help! :)

aartaka:
I'll be glad to have help there, because I'm not a `pass` user and I barely understand the conventions behind it :)

All in all, this seems to be too draft-ish to merge just now.

edgar-vincent:
The conventions for storing additional data in `pass` entries are not standardised. I suggest we follow the following organisational rules:

- the first line is the password except if any of the subsequent lines is prefixed with `password: `, `pass: ` or `secret: ` (case-insensitive)
- the username entry should be prefixed with `login: `, `username: `, or `user: ` (case-insensitive)
- additional fields, such as `port: `, are stored in subsequent lines and are always prefixed by a case-insensitive `keyword: `.
- an entry's name (eg: `My_websites/example.com`) should contain the full domain name for a given website
- if there are two or more accounts for a given domain name, then they can be specified in the entry's path: 
  * either `My_websites/example.com/myuser`
  * or `My_websites/myuser@example.com`
- a port number can be specified after a colon in a given entry's name: `My_websites/example.com:443`

This should provide acceptable compatibility with https://github.com/DamienCassou/auth-source-pass https://github.com/passff/passff#readme and https://github.com/browserpass/browserpass-extension#organizing-password-store

aartaka:
@edgar-vincent, thanks for this summary!

@Ambrevar, feel free to take this issue over -- I don't have enough time to do it well :(

Ambrevar:
I've just pushed some commits which should address most of the comments in this thread.

@edgar-vincent Can you test it and report if this cuts it for you?  Thanks in advance :)

> the first line is the password except if any of the subsequent lines is prefixed with password: , pass: or secret: (case-insensitive)

I haven't implemented this because it adds some complexity and I'm not sure this is really in use.  `password-store` home page really insists that the password ought to be on the first line.

About the port: I don't have a use case, but is Nyxt missing something here?

libraries/password-manager/password-pass.lisp
```diff
@@ -39,13 +39,58 @@
...
+                            (when (= 2 (length entry))
+                              entry)))
+                        ;; Skip first line to ignore password:
+                        (rest lines)))))
+
+(defvar *username-keys* '("login" "user" "username")
+  "A list of string keys used to find the `pass' username in `clip-username'.")
+
+(defmethod clip-username ((password-interface password-store-interface) &key password-name service)
+  "Save the multiline entry that's prefixed with on of the `*username-keys*' to clipboard.
```
edgar-vincent, line 71: s/on of/one of

edgar-vincent:
> I've just pushed some commits which should address most of the comments in this thread.
> 

Thank you *very* much for this! It seems to work wonderfully.

> > the first line is the password except if any of the subsequent lines is prefixed with password: , pass: or secret: (case-insensitive)
> 
> I haven't implemented this because it adds some complexity and I'm not sure this is really in use. `password-store` home page really insists that the password ought to be on the first line.

I agree. I just mentioned it for the sake of compatibility with other add-ons, but your choice makes a lot of sense.

> About the port: I don't have a use case, but is Nyxt missing something here?

If might be useful if, for example, one has service A running on https://localhost:123 and service B on https://localhost:456, both services requiring different credentials.

Ambrevar:
@edgar-vincent Thanks for your feedback.

I've merged this as 2f9f500ec60f0001530a8dd5bd841daf5fa28ac5 since it's working and has been hanging for too long.

About the port question: for now Nyxt makes no attempt at guessing the right password entry based on the current buffer, the URL / port has no impact on the UX at the moment.  

If I'm mistaken, can you detail a precise recipe where parsing the port would be useful for the Nyxt UI?

edgar-vincent:
Thanks for the merge.
You are right, parsing the port is not needed now. It might be in the future if Nyxt gets to guess the right password entry.

Ambrevar:
OK, see https://github.com/atlas-engineer/nyxt/issues/1553 for future discussions.
Closing this then.

aartaka:
Thank you Pierre!

Ambrevar:
Thank you for the initial patch!

PR #1456, opened 2021-05-25
Remember page scroll positions

aartaka:
This adds a simple scroll position saving to history. It works lightning fast and most probably does what you mean.

# What's New:
- `document-scroll-position` method (see "Things To Discuss" about it being a method) that calls internal `%document-scroll-position` parenscript function.
- Small additions to `nyxt/web-mode:on-signal-notify-uri` and `nyxt/web-mode:on-signal-load-finished` to store and restore scroll position respectively.
- `list-of-numbers` type for convenience (how about we actually write a `list-of` type with a placeholder for anything?).

# Things To Discuss:
- The current architecture, although effective, rings some imperfection bell to me. Maybe:
  - Make `document-scroll-position` an FFI method?
  - Make it a `buffer` slot?
- Maybe use current DOM element instead of current scroll position? Feels right semantically, but it's a pain to implement, even with #1365.

# How Has This Been Tested:
- Compiled.
- Ran.
- Moved in between several pages in history, back and forth.
- Position is restored up to a pixel!

How does it look to you?

### DIFF
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -25,7 +25,11 @@ include implicit visits.")
                     :type integer
                     :documentation "
 Number of times the URL was visited by following a link on a page.  This does
-not include explicit visits."))
+not include explicit visits.")
+   (scroll-position '()
+                    :type list-of-numbers
+                    :documentation "The scroll position user was at when last visiting the page.
+Is a list of a form (Y &OPTIONAL X)."))
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
   (:documentation "
 Entry for the global history.
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -29,7 +29,7 @@ not include explicit visits.")
    (scroll-position '()
                     :type list-of-numbers
                     :documentation "The scroll position user was at when last visiting the page.
-Is a list of a form (Y &OPTIONAL X)."))
+It's a list of a form (Y &OPTIONAL X)."))
   (:accessor-name-transformer (hu.dwim.defclass-star:make-name-transformer name))
   (:documentation "
 Entry for the global history.
--- a/source/renderer-script.lisp
+++ b/source/renderer-script.lisp
@@ -35,11 +35,11 @@ The function can be passed ARGS."
 
 (export-always 'document-scroll-position)
 (defmethod document-scroll-position (&optional (buffer (current-buffer)))
-  "Get current scroll position of set it.
-If passed no arguments, returns a list of two elements: vertical (Y) and
-horisontal (X) offset.
-If `setf'-d to a single value (or a single list) -- sets Y to it.
-If `setf'-d to a list of two values -- sets Y to `first' and X to `second' element."
+  "Get current scroll position or set it.
+If passed no arguments, return a list of two elements: vertical (Y) and
+horizontal (X) offset.
+If `setf'-d to a single value (or a single list) -- set Y to it.
+If `setf'-d to a list of two values -- set Y to `first' and X to `second' element."
   (with-current-buffer buffer
     (mapcar #'parse-integer
             (str:split "," (%document-scroll-position)))))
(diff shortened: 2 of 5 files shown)


aartaka:
OK, it's good enough to merge it, unless there are ideas about how we can redesign it.

source/history.lisp
```diff
@@ -25,7 +25,11 @@ include implicit visits.")
                     :type integer
                     :documentation "
 Number of times the URL was visited by following a link on a page.  This does
-not include explicit visits."))
+not include explicit visits.")
+   (scroll-position '()
+                    :type list-of-numbers
+                    :documentation "The scroll position user was at when last visiting the page.
+Is a list of a form (Y &OPTIONAL X)."))
```
Ambrevar, line 32: Typo: `It's`

source/renderer-script.lisp
```diff
@@ -25,6 +25,31 @@ The function can be passed ARGS."
...
+  (let ((x (ps:lisp x))
+        (y (ps:lisp y)))
+    (if (or (ps:lisp x-provided-p) (ps:lisp y-provided-p))
+        (ps:chain window (scroll-to x y))
+        (list (ps:chain window page-y-offset)
+              (ps:chain window page-x-offset)))))
+
+(export-always 'document-scroll-position)
+(defmethod document-scroll-position (&optional (buffer (current-buffer)))
+  "Get current scroll position of set it.
```
Ambrevar, line 38: Typo?  `or set it`

source/renderer-script.lisp
```diff
@@ -25,6 +25,31 @@ The function can be passed ARGS."
...
+If passed no arguments, returns a list of two elements: vertical (Y) and
+horisontal (X) offset.
```
Ambrevar, line 40: Typo: `horizontal`

source/renderer-script.lisp
```diff
@@ -25,6 +25,31 @@ The function can be passed ARGS."
...
+        (y (ps:lisp y)))
+    (if (or (ps:lisp x-provided-p) (ps:lisp y-provided-p))
+        (ps:chain window (scroll-to x y))
+        (list (ps:chain window page-y-offset)
+              (ps:chain window page-x-offset)))))
+
+(export-always 'document-scroll-position)
+(defmethod document-scroll-position (&optional (buffer (current-buffer)))
+  "Get current scroll position of set it.
+If passed no arguments, returns a list of two elements: vertical (Y) and
```
Ambrevar, line 39: Prefer the imperative tense: `return`

source/renderer-script.lisp
```diff
@@ -25,6 +25,31 @@ The function can be passed ARGS."
...
+horisontal (X) offset.
+If `setf'-d to a single value (or a single list) -- sets Y to it.
+If `setf'-d to a list of two values -- sets Y to `first' and X to `second' element."
```
Ambrevar, line 42: Imperative tense: `set`

Ambrevar:
> Maybe use current DOM element instead of current scroll position?

As you said, it will take a while to get right, so in the mean time I suggest we go with this approach and implement the DOM version later.

aartaka:
> > Make document-scroll-position an FFI method?
> 
> Why? Is cl-webkit related to this? Are you saying you'd like to implement it without Javascript?
> If so, is there a benefit in using the WebKitGTK alternative?

It doesn't realy bring much difference, I guess. And it's substantially harder to use :)
 
> > list-of-numbers
> 
> Can't we just use `(cons number *)`? The type-checking is not as thorough, but I guess it's good enough, nah?

Yes, we can. It's just me being really strict about types :)

> > Make it a buffer slot?
> 
> Make what a buffer slot?

`scroll-position`. Doesn't make much sense now that I think of it :)

aartaka:
All typos fixed :)

Ambrevar:
Thanks!

PR #1465, opened 2021-05-30
Nyxt-native dialogs!

aartaka:
This adds handing of `script-dialog` WebKit signal and opens our pretty prompt-buffer instead of WebKit menus, and echoes JS alerts to the message-buffer instead of creating small alerts getting in the way.

### DIFF
--- a/source/renderer-gtk.lisp
+++ b/source/renderer-gtk.lisp
@@ -820,6 +820,29 @@ See `gtk-browser's `modifier-translator' slot."
    (gtk-object buffer) "scroll-event"
    (lambda (web-view event) (declare (ignore web-view))
      (on-signal-scroll-event buffer event)))
+  (gobject:g-signal-connect
+   (gtk-object buffer) "script-dialog"
+   (lambda (web-view dialog) (declare (ignore web-view))
+     (let ((dialog (gobject:pointer dialog)))
+       (webkit:webkit-script-dialog-ref dialog)
+       (run-thread
+         (case (webkit:webkit-script-dialog-get-dialog-type dialog)
+           (:webkit-script-dialog-alert (echo (webkit:webkit-script-dialog-get-message dialog)))
+           (:webkit-script-dialog-prompt
+            (webkit:webkit-script-dialog-prompt-set-text
+             dialog
+             (first
+              (prompt
+               :input (webkit:webkit-script-dialog-prompt-get-default-text dialog)
+               :prompt (webkit:webkit-script-dialog-get-message dialog)
+               :sources (list (make-instance 'prompter:raw-source))))))
+           ((:webkit-script-dialog-confirm :webkit-script-dialog-before-unload-confirm)
+            (webkit:webkit-script-dialog-confirm-set-confirmed
+             dialog (if-confirm
+                     ((webkit:webkit-script-dialog-get-message dialog))
+                     t nil))))
+         (webkit:webkit-script-dialog-close dialog)))
+     t))
   ;; TLS certificate handling
   (gobject:g-signal-connect
    (gtk-object buffer) "load-failed-with-tls-errors"
--- a/source/renderer-gtk.lisp
+++ b/source/renderer-gtk.lisp
@@ -843,6 +843,42 @@ See `gtk-browser's `modifier-translator' slot."
                      t nil))))
          (webkit:webkit-script-dialog-close dialog)))
      t))
+  (gobject:g-signal-connect
+   (gtk-object buffer) "run-file-chooser"
+   (lambda (web-view file-chooser-request)
+     (declare (ignore web-view))
+     (gobject:g-object-ref (gobject:pointer file-chooser-request))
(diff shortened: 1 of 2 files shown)


aartaka:
A small screenshot to get you teased :)
script-dialog2

aartaka:
Keep in mind that it only works on the latest version of cl-webkit.

aartaka:
Mergeable. Thing to do:
- Before merging this:
  - Update submodule version for `cl-webkit`.
- After merging:
  - Add mouse support to prompt-buffer, because we need support equal amount of input devices as built-in dialogs (#1213).
  - Kill web-views properly, so that user gets "Are you sure you want to close this page and lose your input there?" messages.

aartaka:
I've added file-chooser handing with Nyxt-native interface. It handles multiple file input elegantly, fetches an already existing value and works nice in general. I believe something can be done to color-chooser too, but for that we need to somehow add color parsing (is there any CSS-compatible library for that?).

In general, lots of WebKit UI can be replaced with prompt-buffer (including permission requests and notifications). Maybe introduce a `browser`-based switch for Nyxt-native UIs? Something like `nyxt-native-dialogs`.

jmercouris:
I think a browser based switch is a good idea! I can't think of a color picker outside of the one when you invoke open-inspector. There are many JS based ones of course... Maybe we could provide a prompt buffer that includes a few sources for different 'standard' colors?

jmercouris:
Nothing is to stop us for providing an emacs-like color picker experience :-D

aartaka:
> I think a browser based switch is a good idea! I can't think of a color picker outside of the one when you invoke open-inspector. There are many JS based ones of course... 

What we can do is compile user input to CSS, insert it to a page and fetch its color with JavaScript. Terribly hacky, but really convenient. We can even use prompt-buffer's input line as this styled element!

> Maybe we could provide a prompt buffer that includes a few sources for different 'standard' colors?

Yes, that could be an option too.

jmercouris:
I don't think most people are comfortable translating from hex to a color. I can imagine what a color would be, but I can't be certain, and I've been doing it a long time. I guess it could be the first 'raw' user input source in the prompt buffer.

aartaka:
> I don't think most people are comfortable translating from hex to a color. I can imagine what a color would be, but I can't be certain, and I've been doing it a long time. I guess it could be the first 'raw' user input source in the prompt buffer.

I mean, CSS supports [lots of color names](https://www.w3schools.com/cssref/css_colors.asp). You don't need to necessarily use `#HEX` or `rgb(35, 100, 20)` to set the color. That's why CSS is a good way to go here :)

jmercouris:
That works too :-)

jmercouris:
Hello Artyom, I am taking a look at this now :-)

aartaka:
Be aware that there's no toggle to turn these Nyxt dialogs on/off yet!

jmercouris:
Hm, then I will not merge for now.

aartaka:
And I also have a color-chooser somewhere in stashes :)

Ambrevar:
Sorry for the delay, I'll have a look as soon as possible.

aartaka:
OK, here's the `native-dialogs` switch! We can merge if it doesn't break on CI :)

Color-chooser proved to be quite complex, because getting the color of DOM element is a non-obvious task, while I need it to leverage all the power of CSS. Let's leave it for later and open an issue on it.

aartaka:
It doesn't seem to break on CI :D

Ambrevar:
Just tested: it seems to hang the current page when the user exits with `escape`
instead of confirming the dialog.

Other than that, congrats on this, it's a fantastic change :)
I'll update the cl-webkit Guix package just now.

source/renderer-gtk.lisp
```diff
@@ -820,6 +820,67 @@ See `gtk-browser's `modifier-translator' slot."
...
+                                                         file-chooser-request)
+                                                        (mapcar #'cffi:foreign-string-alloc files)
+                                                        (list (cffi:foreign-string-alloc (first files))))
+                                  :count (if (webkit:webkit-file-chooser-request-select-multiple
+                                              file-chooser-request)
+                                             (length files)
+                                             1)
+                                  :null-terminated-p t)))
+           (webkit:webkit-file-chooser-request-cancel file-chooser-request)
+           t)))))
```
Ambrevar, line 883: This is getting a tad long :)  Extract the handler to separate functions?

aartaka: Sure!

aartaka:
All adressed! (I resorted to a suspicious `ignore-errors` to return `nil` from the cancelled `prompt`, but it's all smooth otherwise.)

aartaka:
The last suspicious thing (the `ignore-erros`) is addressed in 25dbdc5141ee6cca0e799e84786e8f269ede9b03. Should be mergeable, I believe.

jmercouris:
Thanks Artyom!

PR #1467, opened 2021-06-03
Add gobject-introspection to shell.nix

alejandrogallo:
Using the nix shell to compile nyxt
I was missing the `gobject-instropection` library, this fixed the compilation
for me.


jmercouris:
We have a gi-shell.nix for that purpose. I'll go ahead and just rename gi-shell.nix to shell.nix! Thank you for pointing this out though. Now that the default renderer is gi-gtk, I can see how that would be confusing!

PR #1486, opened 2021-06-08
Makes search-buffer-source have an user instance

mariari:
This allows the user to customize the search buffer behavior, I've made the following extension in my configuration

```lisp
(define-configuration nyxt/web-mode:search-buffer-source
  ((nyxt/web-mode:minimum-search-length 1)))
```

however one can use this for other purposes


jmercouris:
Thank you!

Issue #1380, opened 2021-05-08
Prevent history corruptions


Ambrevar:
It happened to me twice already that the history would get corrupted.
I think this happens when the browser is terminated prematurely while history is being written to.  Indeed, we are writing the history file in a buffered fashion.

Instead, we should serialize the history to a temp file and rename it to the final file.

PR #1492, opened 2021-06-08
Serialize data-paths (e.g. history) safely

Ambrevar:
In the course of the last 2 weeks, my global history file got corrupted twice.
This happens, I believe, when an error occurs while the file is persisted.

To protect against this, we can persist the file to a temporary one and rename it to the final path when done.

@jmercouris @aartaka Can you test if this works for you?
(Considering backing up your files!)

### DIFF
--- a/source/blocker-mode.lisp
+++ b/source/blocker-mode.lisp
@@ -54,7 +54,7 @@ If HOSTLIST has a `path', persist it locally."
         (when path
           (handler-case
               (alex:write-string-into-file hosts (ensure-parent-exists path)
-                                           :if-exists :overwrite
+                                           :if-exists :supersede
                                            :if-does-not-exist :create)
             (error (c)
               (log:error "Could not persist hostlist ~s: ~a" path c))))
--- a/source/data-storage.lisp
+++ b/source/data-storage.lisp
@@ -494,3 +494,35 @@ Parent directories are created if necessary."
        (when ,path
          (with-maybe-gpg-file (,stream ,path ,@options)
            ,@body)))))
+
+(defun ensure-unique-file (file)
+  "Return FILE if unique or suffix it with a number otherwise.
+Warning: This is not atomic, no lock is made on any file, so it's not safe from
+race conditions."
+  (loop with original-name = file
+        with suffix = 1
+        while (uiop:file-exists-p file)
+        do (setf file (format nil  "~a.~a" original-name suffix) )
+        do (incf suffix))
+  file)
+
+(export-always 'with-data-file-output)
+(defmacro with-data-file-output ((stream data-path) &body body)
+  "Open DATA-PATH for writing.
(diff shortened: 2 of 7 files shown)


Ambrevar:
Fixes #1380.

jmercouris:
I'll try to give this a try tomorrow!

source/data-storage.lisp
```diff
@@ -494,3 +494,35 @@ Parent directories are created if necessary."
...
+(export-always 'with-data-file-output)
+(defmacro with-data-file-output ((stream data-path) &body body)
+  "Open DATA-PATH for writing.
+The stream is written to a temporary file and renamed to the final DATA-PATH
+path upon closing.
+This measure protects against half-written files when an error occurs in the BODY."
+  (alex:with-gensyms (temp-path final-path)
+    `(let* ((,final-path (expand-path ,data-path))
+            (,temp-path (ensure-unique-file ,final-path)))
+       (unwind-protect (prog1 (with-data-file (,stream ,temp-path
```
aartaka, line 518: `,temp-path` here evaluates to a string, while `with-data-file` expects a `data-path` object. 

This provokes two problems:
- Type mismatch that simply errors out (at least on my machine).
- `data-path`s are expanded earlier than they should (i.e. earlier than in `with-data-file`).

To solve that, how about rewriting `ensure-unique-file` to accept a `data-path` and return its copy with `basename` uniquified? It'll be `ensure-unique-data-path` then. 

This, however, can complicate further renaming and stuff.

source/blocker-mode.lisp
```diff
@@ -54,7 +54,7 @@ If HOSTLIST has a `path', persist it locally."
         (when path
           (handler-case
               (alex:write-string-into-file hosts (ensure-parent-exists path)
-                                           :if-exists :overwrite
+                                           :if-exists :supersede
```
aartaka, line 57: So that's the difference between `:overwrite` and `:supersede` that I didn't get :D 

Thanks for this patch -- I learned a new thing today :)

aartaka (commented):
I've just tested and it errors on type mismatch -- `with-data-file` expects a `data-path`, while `with-data-file-output` passes a string to it.

source/blocker-mode.lisp
Ambrevar: Me too! :)

source/data-storage.lisp
Ambrevar: Oops!  Thanks for the catch, will fix next week!

Ambrevar: Should be fixed now.

Ambrevar:
I've tested it more properly this time, should be good to merge.

aartaka:
The only error I encountered is gone. Code is clear. Let's merge :)

Issue #1426, opened 2021-05-19
M-i binding conflict in VI mode


Ambrevar:
When VI mode is enabled in the prompt buffer, by default we can't toggle hint transparency in `follow-hint` because `M-i` is bound to `vi-focus-first-input-field`, which has priority since `vi-normal-mode` comes first.

Any clue how to work around this?

Rebind `M-i`?  Any suggestion?

jmercouris:
Just rebind it to anything at all :-D. There is no precedent here!

Ambrevar:
So rebind "M-i"?

Note that I would have to rebind the CUA binding as well.

To what?

jmercouris:
I don't know, just pick something :-D

Ambrevar:
Fixed with #1499.

PR #1499, opened 2021-06-10
Fix automatic VI focus

### LINKED ISSUE #1462: Enable vi-insert-mode automatically on input fields
I find myself typing in an input field and wondering why my keys don't work all of a sudden then I find out I forgot to switch to vi-insert-mode. So I think it would be easier if this happens automatically,
* When opening prompt buffer.
* After selecting an input area or a text area through element-hint-mode.
#### What I tried
I tried enabling vi-insert-mode automatically in the prompt buffer by hooking it to prompt-buffer-mode-after-hook, but it only turns on vi-insert-mode globally with the prompt still on vi-normal-mode.
#### Related issues
* This issue #1461 talks about how the "g" key is not usable. This is because the user is on vi-normal-mode.
* This issue #739 talks about not forwarding unbound keys to avoid people typing in vi-normal-mode.

Ambrevar:
Quite a few things happened here, looks at the individual commits.

Do you like the strategy of introducing a method to customize the behaviour per mode?

Questions:

- What do you think of the use of `define-deprecated-command`?  By the way, I just realized the naming is inconsistent with `define-command-global`, so it should be `define-command-deprecated` instead.
- Should we have a `element-focused` hook instead of the method?  Hooks are more powerful but a bit less convenient to use.  Thoughts?

Fixes #1462, #1426.


Ambrevar:
I recently saw a few complaints about vi-insert-mode not being the default in the prompt buffer when the user would use the customize interface.  I can't find the related issue.  Anyone?

Anyways, should be fixed by the first commit of this patchset.

Ambrevar:
@aartaka Can you test and review this?

jmercouris:
I don't believe Artyom is a vi user

aartaka:
> @aartaka Can you test and review this?

Will do, hopefully in a day!

Ambrevar:
Oh, @aartaka I thought you were a VI user, aren't you?

Ambrevar:
Regardless of being a VI user or not, please let me know what you think of this architecture, I have some doubts.

aartaka:
I'm not a VI user (although I plan to move to VI bindings at some point). 

> Regardless of being a VI user or not, please let me know what you think of this architecture, I have some doubts.

Sure!

aartaka:
Sorry, Pierre, didn't have time to test it today! Will test and review tomorrow!

jmercouris:
@aartaka No need, Pierre and I have spoken about it already!

Ambrevar:
`custom-settings` was completely broken in 2.0 (my bad :/), because `configure-slot` actually only works on single classes, not on class list, unlike what the docstring claimed.

I've just fixed it.

@jmercouris @aartaka Can you double check this works for you?  It's rather critical for the next release :p

jmercouris:
Well, I take back what I said :-D

Ambrevar:
:D

jmercouris:
Works for me. Does not appear to work on iFrames though, like at https://nyxt.atlas.engineer/start .

May be a limitation we will have to live with for now.

source/web-mode.lisp
```diff
@@ -522,6 +524,50 @@ Otherwise go forward to the only child."
    :prompt "Autofill"
    :sources (make-instance 'user-autofill-source)))
 
+(export-always 'element-focused)
+(defgeneric element-focused (mode)      ; TODO: Make hook instead?
+  (:method ((mode t))
+    nil)
+  (:documentation "Method run when `focus-element' is called."))
+
+(defmacro focus-element ((&optional buffer) &body element-script)
```
aartaka, line 533: This actually points at an important shortcoming of `define-parenscript`: we often need to do Lisp actions alongside the JS ones. Maybe introduce some way to add Lisp code around/before/after (methods are most suitable there) the JS action is completed? This way we can use good old `focus-element (&optional nyxt-identifier)` while still having an arbitrary Lisp code ran around it.

source/web-mode.lisp
```diff
@@ -522,6 +524,50 @@ Otherwise go forward to the only child."
...
+  "Select the element pointed to by ELEMENT-SCRIPT.
+ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
+  (alex:with-gensyms (element)
+    `(progn
+       (ffi-buffer-evaluate-javascript (or ,buffer (current-buffer))
```
aartaka, line 538: Maybe use `(&optional (buffer (quote (current-buffer))))` in the arglist instead?

source/web-mode.lisp
```diff
@@ -522,6 +524,50 @@ Otherwise go forward to the only child."
...
+                                       (ps:ps (let ((,element (progn ,@element-script)))
+                                                (ps:chain ,element (focus))
+                                                (ps:chain ,element (select)))))
+       (dolist (mode (modes (or ,buffer (current-buffer))))
```
aartaka, line 542: Multiple argument evaluation. It's unlikely to shoot us in the feet, but better to err on the side of caution and use `alex:once-only`.

source/web-mode.lisp
```diff
@@ -522,6 +524,50 @@ Otherwise go forward to the only child."
    :prompt "Autofill"
    :sources (make-instance 'user-autofill-source)))
 
+(export-always 'element-focused)
+(defgeneric element-focused (mode)      ; TODO: Make hook instead?
```
aartaka, line 528: Using methods is alright, and we already do so. Hooks are better left for the user to configure.

What we _can_ do, though, is to call the hook in the default method of this generic. Best of both options: we use flexible methods in the code, and expose hooks for users to tinker with :)

aartaka (commented):
Deleting deprecated commands is quite harsh, but it feels right. The architecture is all fine, except the fact that the `define-parenscript` is not strong enough (was that the reason for the `focus-element` rewriting?)

Ambrevar:
> Works for me. Does not appear to work on iFrames though, like at https://nyxt.atlas.engineer/start .

Can you reproduce on master?  It seems that I can, so this is probably not related to this pull request.  Can you confirm?

source/web-mode.lisp
Ambrevar: Good point, I'll either add a comment or add a hook.

source/web-mode.lisp
Ambrevar: Good catch!

source/web-mode.lisp
Ambrevar: Maybe for another patch?

Ambrevar:
Updated.
If it's OK with you, let's merge and open 2 new issues:

- iframe link hints don't work, as on the start page.
- Expand define-parenscript as @aartaka mentioned.

jmercouris:
> > Works for me. Does not appear to work on iFrames though, like at https://nyxt.atlas.engineer/start .
> 
> Can you reproduce on master? It seems that I can, so this is probably not related to this pull request. Can you confirm?

I can confirm.

aartaka:
> Updated.
> If it's OK with you, let's merge and open 2 new issues:
> 
>     * iframe link hints don't work, as on the start page.
> 
>     * Expand define-parenscript as @aartaka mentioned.

Sounds like a plan! :D

Ambrevar:
See https://github.com/atlas-engineer/nyxt/issues/1527 and https://github.com/atlas-engineer/nyxt/issues/1526.

PR #1502, opened 2021-06-11
Update CSS colors for coherence.

aartaka:
This changes our hex-based CSS to named colors.

# Why?
- Named colors are easier to comprehend.
- I've started doing [systematic Nyxt theming](https://github.com/aartaka/nyxt-config/commit/4afc3d3618839bca5251756a316dc46eec7d4a92) and understood that the set of colors we use is quite small, but one cannot see ~the forest for the trees~ all this because of hex-valued colors. Thus, this patch is an improvement to make theming somewhat easier cognitively.
  - Adding to that, I plan to simply search for color names we use after we discuss and merge this patch. This will allow to make a default Nyxt color theme to derive other themes from, e.g., easily do themes like in the macro I've done.

# Things To Discuss
- @jmercouris, I've adjusted some colors 2-20 RBG points away from initial hex values. Is that alright? I feel it is, because it yields more coherence in the long term, but you can prove me wrong :)
- Maybe we should include the aforementioned bit of configuration into the upstream? This can even become a cool Common Settings part to quickly style Nyxt to one's liking :)

# How Has This Been Tested

It wasn't. Colors are not too far, though. Shouldn't cause much trouble.

How's that looking?

### DIFF
--- a/source/buffer-listing-mode.lisp
+++ b/source/buffer-listing-mode.lisp
@@ -56,7 +56,7 @@
        (:br "")
        (:div
         (if cluster
-            (append (list (internal-buffers-markup)) 
+            (append (list (internal-buffers-markup))
                     (loop for cluster-key being the hash-key
                           using (hash-value cluster) of (cluster-buffers)
                           collect (cluster-markup cluster-key cluster)))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -366,7 +366,9 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
                (|.button:visited|
                 :color "white")
                (|.button:active|
-                :color "white")))))
+                :color "white")
+               (a
+                :color "gray")))))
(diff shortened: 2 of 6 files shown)


jmercouris:
I kind of think that we should use a CSS class to describe all colors. So we can have slots for each color, or each type of element we want to color. This could then bubble down to the default values of the style slots, unless explicitly overridden. 

With regards to these colors changed, I'll have to take a look at them. I selected the colors very carefully.

jmercouris:
Something like

```
(defclass theme ()
  ((color-of-x "xyz"))))
```

jmercouris:
To effectively do this though, we would probably need a global, or to at least leverage the `*browser*`, is that smart? I don't know.

aartaka:
@jmercouris, mode names' color has more contrast now! Should be mergeable :)

jmercouris:
Thank you!

PR #1505, opened 2021-06-14
[backwards-incompatible] Move (&rename) gtk-extensions-path to XDG_DATA_HOME.

aartaka:
This changes the WebKit extensions directory to be next to the Lispy extensions one -- in .local/share/nyxt (or other XDG_DATA_HOME value). This makes things more consistent. However, this is a backwards-incompatible change and thus something to keep until 3.0.

### DIFF
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -204,12 +204,20 @@ not return."
   "We shouldn't enable (possibly) user-identifying extensions for `nosave-data-profile'."
   nil)
 
+
 (define-class gtk-download (download)
   ((gtk-object)
    (handler-ids
     :documentation "See `gtk-buffer' slot of the same name."))
   (:accessor-name-transformer (class*:make-name-transformer name)))
 (define-user-class download (gtk-download))
+
+(defmethod expand-data-path ((profile data-profile) (path gtk-extensions-data-path))
+  "Return finalized path for gtk-extension directory."
+  (expand-default-path path :root (namestring (if (str:emptyp (namestring (dirname path)))
+                                                  (uiop:xdg-data-home +data-root+ "gtk-extensions")
+                                                  (dirname path)))))
+
 
 (defun make-web-view (&key context-buffer)
   "Return a web view instance.
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -200,10 +200,8 @@ not return."
   (:accessor-name-transformer (class*:make-name-transformer name)))
 
 (defmethod expand-data-path ((profile nosave-data-profile) (path gtk-extensions-data-path))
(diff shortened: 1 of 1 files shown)


Ambrevar:
Wouldn't it be possible to support both paths, to make it backward compatible?

jmercouris:
Hm, then it will have to wait until 3.0.0.

aartaka:
> Hm, then it will have to wait until 3.0.0.

No problem :)

jmercouris:
Thank you Artyom!

aartaka:
Hmmmmm, interesting. I get this warning on master:

```
** (nyxt:2571): WARNING **: 15:52:49.141: GApplication is required for xdg-desktop-portal access in the WebKit sandbox. Actions that require xdg-desktop-portal will be broken.
```

It breaks loading WebKit extensions. Seems like we need to either move things back to where they were (`.config/nyxt/extensions`) or create a `GApplication` for Nyxt.

Ambrevar:
I'm confused, why did it work with .config/nyxt/extensions?

aartaka:
> I'm confused, why did it work with .config/nyxt/extensions?

Oh wait. Now I wonder too.

aartaka:
> I'm confused, why did it work with .config/nyxt/extensions?

Oh wait. Now I wonder too.

Ambrevar:
Open a new issue about this?

aartaka:
Done: #1781.

aartaka:
I've just tried removing the `expand-data-path` for `gtk-extensions-data-path` defined in this patch. Surprisingly, WebKitWebExtensions do load from `.config/nyxt/extensions` without any warnings and work well! 

Do we remove this method and move back to `.config/nyxt/extensions/` as the WebKitWebExtensions storage in 2.2.3/3.0?

aartaka:
That's a hillarious bug that I've introduced with this patch, sorry about it :)

aartaka:
Oh wait. I've tested it with `WEBKIT_FORCE_SANDBOX=0` and it errors if I don't set this variable \(\~_\~\;)

`.config/` is not safe either :/

Ambrevar:
Any idea then why the error _started_ occurring after 2.2.0 then?

aartaka:
> Any idea then why the error _started_ occurring after 2.2.0 then?

No idea (・_・ヾ

Something else should've changed.

Ambrevar:
I guess we need to bisect now :p

Ambrevar:
I think it's the WebKitGTK 2.34 update.  To be confirmed.

Ambrevar:
Note that everyone complaining or Void / Arch users who have received the WebKitGTK update.

Ambrevar:
The plot thickens! :)

Issue #1503, opened 2021-06-11
Add command `show-clipboard`


pdelfino:
While @jmercouris was instructing me on how to solve issue #765 , a new idea for a command emerged: the `show-clipboard`.

The command should indicate in the Nyxt mini-buffer what is inside the clipboard. The step-by-step to solve this _might_ be similar to PR #1501. For instance, the `(echo ...)` expression could be used.

The tags "good first issue" and "feature" might be appropriate.

aartaka:
Given our reliance on `trivial-clipboard`, it can be hard to do -- the only thing this library provides in copying the top-most thing from the clipboard. But this is a nice feature to have (and, in fact, the one I dreamt about for a long time)!

pdelfino:
PR #1506 was merged. Thus, I am closing this issue :)

Issue #765, opened 2020-05-29
Add copy-placeholder-text command


Ambrevar:
An itch I want to scratch: placeholder text cannot be copied (beside from the inspector) which can be frustrating if one wants to, say, copy a form including the suggestions.

A dedicated command could help.
Maybe C-c could be bound to selectively copy the placeholder text or the text, depending on which one is present.

pdelfino:
Leave a comment

pdelfino:
@Ambrevar, I think you can close this one. John helped me solve it.

Ambrevar:
Fantastic, thanks a lot!

PR #1506, opened 2021-06-14
browser.lisp: create the command to show the last element on the clipboard stack

pdelfino:
Hi guys,

This seems to solve issue #1503. The new command shows the *last* element on the clipboard stack. It does not show all elements but it is already good enough useful (imho). In the future, I bet this could be improved to show all the clipboard stack. 
I did it mainly for learning reasons since this was close to #765.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -568,6 +568,14 @@ sometimes yields the wrong result."
                    :if-does-not-exist :create
                    :if-exists :append))))))
 
+
+
+(define-command show-clipboard ()
+  "Show the current clipboard."
+  (echo (ring-insert-clipboard (clipboard-ring *browser*))))
+
+
+
 (defmacro define-ffi-generic (name arguments &body options)
   `(progn
      (export-always ',name)


source/web-mode.lisp
```diff
@@ -510,6 +510,10 @@ Otherwise go forward to the only child."
         (progn (copy-to-clipboard current-value)
                (echo "Placeholder copied.")))))
 
+(define-command show-clipboard ()
+  "Shows the current clipboard."
```
Ambrevar, line 514: Use the imperative tense `Show`.

source/web-mode.lisp
```diff
@@ -510,6 +510,10 @@ Otherwise go forward to the only child."
         (progn (copy-to-clipboard current-value)
                (echo "Placeholder copied.")))))
 
+(define-command show-clipboard ()
```
Ambrevar, line 513: If this is meant to be used globally, i.e. regardless of the buffer web view, then make it a global command with `define-command-global`.

Ambrevar: And if it's not related to the web view, then it does not belong in web-mode and should be move to `browser.lisp` for instance.

pdelfino: Thanks for the feedback, @Ambrevar! I think I did not **fully** understand your comment due to my lack of knowledge about Nyxt.

According to the documentation web mode is the "Base mode for interacting with documents.".

How do I know if something should be related or not to the web view? Could you please give more examples with commands that already exist to illustrate that?

I think `show-clipboard` is pretty general. If Nyxt is running, people could be able to use `show-clipboard` even if they are using in Dr. Racket in another window. So, should I move it to browser.lisp?

Ambrevar: We haven't defined any commands in `browser.lisp` yet, but why not.

By "related to the web view", I mean anything that interacts with the content of the web view (HTML, css, javascript).  Examples: search, hints and pasting.

Does that make sense?

pdelfino: Great examples. It is crystal clear now for me what you said before. Thanks!

> We haven't defined any commands in browser.lisp yet, but why not.

I was unsure about where to put it in the file. So I put in the end as the last thing. Is this a good/bad/neutral approach?

Moreover, I noticed 1 tests was not successful as shown by Github. I do not know how to solve it or why it happened.

Could you give me a hint?

Ambrevar:
To place things in a file, try to find a related context.  If it's not related to anything but still belongs to the file, place it at the end.

The GitHub action issue is a false negative (I need to work on it), don't worry.

jmercouris:
I would put the command further up in the file, the FFI-* functions are at the bottom.

pdelfino:
> I would put the command further up in the file, the FFI-* functions are at the bottom.

Ok, John. I did it! I also squashed all commits into 1 with the magit trick you taught me yesterday. I hope it is correct. Thanks!

jmercouris:
It's correct, thank you!

Issue #1088, opened 2021-01-02
Question : how to set the proxy


deb75:
I have settled `privoxy` listening on 10.10.1.1:8118.

How do I tell `nyxt` to use my proxy ? Does it as to do with the `proxy-mode` ?

Regards

aartaka:
Yes, it's about `proxy-mode`! You can see an example of configuration if you will do `describe-class nyxt/proxy-mode:proxy-mode` in Nyxt. I'm implying you're familiar with how commands are executed in Nyxt and how minibuffer works. If you're not, you better say it so I can explain ;)

For what you want, it should be something along the lines of:
``` lisp
(define-configuration nyxt/proxy-mode:proxy-mode
  ((nyxt/proxy-mode:proxy (make-instance 'proxy
                                         :server-address (quri:uri "socks5://0.10.1.1:8118")
                                         :allowlist '("localhost" "0.10.1.1:8118")
                                         :proxied-downloads-p t))))
```
in your init file.

aartaka:
@deb75, did the code I suggested work for you?

deb75:
Hi,

Yes ! thanks very much for your kind and detailed answer.

I am ok with minibuffer, using emacs for several years.

I would also like to make the bottom bar a little bigger, how can I do that ?

Regards

aartaka:
Oh, wait, I've had a mistake in the code I suggested -- `"10.10.1.1:8118"` shouldn't be in `allowlist`.

Regarding the error in logs: is it only this, or is there some other message after that? it looks like error message is incomplete.

Regarding `status-buffer` (i.e. bottom bar, as you call it) size -- try something like:
``` lisp
(define-configuration status-buffer
  ((height 40)
   (style (str:concat %slot-default
                      (cl-css:css
                       '((body
                          :font-size "30px"
                          :line-height "40px")))))))
```

And adjust numbers as you like :)

deb75:
Thanks for your reply.

Here is the configuration is use : 
```lisp
(define-configuration (buffer web-buffer)
  ((default-modes (append '(emacs-mode) %slot-default))))

(define-configuration nyxt/proxy-mode:proxy-mode
  ((nyxt/proxy-mode:proxy (make-instance 'proxy
                                         :server-address (quri:uri "socks5://10.10.1.1:8118")
                                         :allowlist '("127.0.2.1")
                                         :proxied-downloads-p t))))

(define-configuration buffer
[...]

PR #1507, opened 2021-06-14
proxy-mode: fix proxy setting example

Zwo1in:
Hi,
as mentioned in #1088, I got an error launching nyxt with example from proxy-mode help. I've managed to configure it like in this commit. I don't know lisp and have no idea what am I doing, so please correct me if the problem was on my side :)

### DIFF
--- a/source/proxy-mode.lisp
+++ b/source/proxy-mode.lisp
@@ -14,10 +14,10 @@ a proxy for all buffers, add it to the list of default modes.
 Example to use Tor as a proxy both for browsing and downloading:
 
 \(define-configuration nyxt/proxy-mode:proxy-mode
-  ((nyxt/proxy-mode:proxy (make-instance 'proxy
-                                         :url (quri:uri \"socks5://localhost:9050\")
-                                         :allowlist '(\"localhost\" \"localhost:8080\")
-                                         :proxied-downloads-p t))))
+  ((proxy (make-instance 'proxy
+                               :url (quri:uri \"socks5://localhost:9050\")
+                               :allowlist '(\"localhost\" \"localhost:8080\")
+                               :proxied-downloads-p t))))
 
 \(define-configuration buffer
   ((default-modes (append '(proxy-mode) %slot-default%))))"


Ambrevar:
Thanks for the feedback!

The issue is not with the documentation however, but with a symbol not being exported properly.
This should be fixed with 3fd79a0639f0fe897ade563badaa4feca2614260, can you confirm?

Ambrevar:
Actually, 3fd79a0639f0fe897ade563badaa4feca2614260 might be broken, so test the latest commit instead.

Zwo1in:
Yes, indeed, on latest master I can see the config based on original documentation parsed with no issues. It does not work as I would expect tho.
What I was expecting was that after turning vpn on, I would need to manually enable `proxy-mode` for each buffer that I'd like to use proxy to be able to reach desired websites.
What I found is that I don't need to enable `proxy-mode`. Each website I was trying to reach was accessible, it didn't matter whether the website was internal to the in-corpo network or external like youtube or this github etc. Maybe this is a desired behavior, but I wasn't aware of it (cool feature if it is by design btw :)).
Just for completenes, I didn't enable any addition of `proxy-mode` to `auto-mode` etc, the change that's visible in this PR is the only thing I have in my config (reverted to origin for this test puropses ofc).

Zwo1in:
One more thing that I thing I should mention. When not under vpn, this proxy setting would result in no webste being accessible, and this is what happens when I explicitely set `proxy-mode` with no vpn. Not setting it explicitely results in each website accessible anytime, so it acts 'smartly'. That's why I said that's a cool feature if it's by design.
I'm sure that this proxy setting is used automatically under vpn, as normally my corpo blacklists sites like youtube and the only way of accessing those under vpn is using socks5 proxy.

Zwo1in:
I've closed this, but please let me know whether this was desired :)

Zwo1in:
In my config there is literally only this proxy setting that I was changing in this PR.

It makes sense, that means that I have some 'system network settings' that allows this connection. Setting `proxy-mode` results in much faster connections, so everythings seems to be working. I have no idea what are my 'system network settings', maybe I did there something years ago that make this working :) Anyway, PEBKAC, let's leave this topic, I'm completely fine with the outcome, thanks!

Ambrevar:
Which desktop environment are you using?
Also check the output of `env | grep -E 'HTTPS?_PROXY'`.

Zwo1in:
I'm using awesome wm.
My environment is clean:
```
~
❯ env | rg -i proxy

~
❯                  
```

Indeed when in firefox I check `Use system proxy settings`, the youtube is working also.

Ambrevar:
Are you using the system of some corporate environment?  If so, the
proxy settings might have been preconfigured somehow.  You may want to
ask your system administrator.

PR #1509, opened 2021-06-15
Fix broken link in README

eihli:
Fixing a typo where the URL was omitted.


aartaka:
That's a good catch! Thank you!

PR #1510, opened 2021-06-15
tutorial.lisp: add information about enable and disable modes on the manual

pdelfino:
Hi,

As discussed in a previous private email, here is a suggestion to insert the commands `enable-mode` and `disable-mode` on the manual. I tried to keep the style similar to the other parts. Not sure if I did manage to do that, though. I can iterate the content. Thanks!

### DIFF
--- a/source/tutorial.lisp
+++ b/source/tutorial.lisp
@@ -176,7 +176,12 @@ removed from the input, they are also removed from the existing bookmark.")
     (:li (command-markup 'delete-bookmark) ": Delete queried bookmarks.")
     (:li (command-markup 'list-bookmarks) ": Display a new buffer containing the
 list of all bookmarks."))
-
+   (:h3 "Enable and Disable modes")
+   (:p "The command " (:code "enable-mode") " allows the user to apply a
+mode (such as " (:code "nosound-mode") " or " (:code "dark-mode") ") to multiple
+buffers at once. Conversely, it is possible to revert this action
+executing "(:code "disable-mode") " while choosing exactly the same buffers and
+the same mode previously selected.")
    (:h3 "Visual mode")
    (:p "Select text without a mouse. Nyxt's "
        (:code "visual-mode") " imitates Vim's visual mode (and comes with the


source/tutorial.lisp
```diff
@@ -176,7 +176,8 @@ removed from the input, they are also removed from the existing bookmark.")
     (:li (command-markup 'delete-bookmark) ": Delete queried bookmarks.")
     (:li (command-markup 'list-bookmarks) ": Display a new buffer containing the
 list of all bookmarks."))
-
+   (:h3 "Enable and Disable modes")
+   (:p "Unlike other modes, " (:code "enable-mode") " represents an upper level of abstraction having a meta profile. It allows the user to apply other modes (e.g., " (:code "nosound-mode") " or " (:code "dark-mode") ") to multiple buffers at once. Conversely, it is possible to revert this action executing " (:code "disable-mode")" while choosing exactly the same buffers and the same mode previously selected.")
```
aartaka, line 180: Wow, that's a long one!

Try to make sure that every line is less than 100 chars long — it's more readable this way. Don't be afraid to insert line breaks — they all will disappear once this code is converted to HTML.

Regarding content: it looks fine, but the length on the line makes it really inconvenient to read, so sonething could've escaped my attention :)

pdelfino: Sorry! I was editing this in Emacs with two windows like a vertical split. I did not notice this fact because of the word wrapping. I will pay intention on it and update the PR.

source/tutorial.lisp
```diff
@@ -176,7 +176,13 @@ removed from the input, they are also removed from the existing bookmark.")
     (:li (command-markup 'delete-bookmark) ": Delete queried bookmarks.")
     (:li (command-markup 'list-bookmarks) ": Display a new buffer containing the
 list of all bookmarks."))
-
+   (:h3 "Enable and Disable modes")
+   (:p "Unlike other modes, " (:code "enable-mode") " represents an upper 
+level of abstraction having a meta profile. It allows the user to apply other
```
Ambrevar, line 181: What do you mean with "meta profile"?

source/tutorial.lisp
```diff
@@ -176,7 +176,13 @@ removed from the input, they are also removed from the existing bookmark.")
...
+ modes (e.g., " (:code "nosound-mode") " or " (:code "dark-mode") ") to multiple
+ buffers at once. Conversely, it is possible to revert this action executing "
+ (:code "disable-mode")" while choosing exactly the same buffers and the same
```
Ambrevar, line 184: Remove the leading space in the above lines.

source/tutorial.lisp
```diff
@@ -176,7 +176,13 @@ removed from the input, they are also removed from the existing bookmark.")
     (:li (command-markup 'delete-bookmark) ": Delete queried bookmarks.")
     (:li (command-markup 'list-bookmarks) ": Display a new buffer containing the
 list of all bookmarks."))
-
+   (:h3 "Enable and Disable modes")
+   (:p "Unlike other modes, " (:code "enable-mode") " represents an upper 
+level of abstraction having a meta profile. It allows the user to apply other
+ modes (e.g., " (:code "nosound-mode") " or " (:code "dark-mode") ") to multiple
```
Ambrevar, line 182: Prefer "such as" to the more cryptic "e.g.".

source/tutorial.lisp
pdelfino: Sorry, not sure if I get it. 

Are you referring to remove the line breaks I inserted after the suggestion made by @aartaka on this PR yesterday?
Should I restore it to the way it was?

jmercouris: Notice that there is whitespace on line 182 before 'modes' and on 183 before 'buffers' etc.

Ambrevar:
> By "meta" I mean that enable-**mode** is _a mode_ built to activate _other modes._ 
> Does it sound inappropriate?

enable-mode is not a mode, it's a command.

pdelfino:
> By "meta" I mean that enable-**mode** is _a mode_ built to activate _other modes._ Does it sound inappropriate?
> enable-mode is not a mode, it's a command.

Sorry. You are correct. I am pushing a similar text. This time making clear enable-mode is a command and without the 'meta' part.

source/tutorial.lisp
pdelfino: This how I see things in emacs:
image

Please, take a look at the cursor. Should this space exist between "a" and the "line break"? 

The word on the next line called "mode" does not have a whitespace by its left side (in emacs).

pdelfino:
Hey, @jmercouris ,

After we applied `fill-paragraph` and  used `whitespace-mode`, I added and committed the change. When I pushed, one of the tests started to fail. I tried reading the message but it says something about "global history". I do not know why this is happening.

Before talking to you, I removed a line that I thought could be an extra space. Initially, I thought this could be the root of the problem. So I reverted the commit (even though it seems harmless deleting a blank line).

Despite reverting the commit, the test still fails. Could you help me understand what happened?

jmercouris:
The test is unrelated to your changes, our CI is having some problems currently, I wouldn't worry about it!

PR #1528, opened 2021-06-18
Correct mode's package in proxy-mode documentation

Zwo1in:
Hi,
I'm here again messing with proxy-mode documentation :) Example copied from docs hasn't worked for me, and this is how I fixed that. As before, I have no idea what I'm doing as I'm at the begining of my journey with lisp, so if the problem was on my side, please just correct me.

### DIFF
--- a/source/proxy-mode.lisp
+++ b/source/proxy-mode.lisp
@@ -24,7 +24,7 @@ Example to use Tor as a proxy both for browsing and downloading:
                                          :proxied-downloads-p t))))
 
 \(define-configuration buffer
-  ((default-modes (append '(proxy-mode) %slot-default%))))"
+  ((default-modes (append '(nyxt/proxy-mode) %slot-default%))))"
   ((proxy (make-instance 'nyxt:proxy
                          :url (quri:uri "socks5://localhost:9050")
                          :allowlist '("localhost" "localhost:8080")


Zwo1in:
While this seems to not cause errors on startup, it doesn't look like it actually enables proxy-mode by default :(
The error I'm facing with the original example is:
```
❯ nyxt
<INFO> [20:14:29] Listening to socket "/run/user/1000/nyxt/nyxt.socket".
Nyxt version 2.0.0-131-g7ae151bc
<INFO> [20:14:29] Loading Lisp file "/home/mzwolins/.config/nyxt/auto-config.lisp".
<INFO> [20:14:29] Loading Lisp file "/home/mzwolins/.config/nyxt/init.lisp".
WARNING:
   redefining DEFAULT-MODES :AROUND (#<STANDARD-CLASS NYXT:USER-BUFFER>) in DEFMETHOD
<WARN> [20:14:29] Warning: Error on separate thread: When attempting to
                          set the slot's value to #<PROXY {1008FD7A23}> (SETF of SLOT-VALUE),
                          the slot PROXY is missing from the object
                          #<USER-INTERNAL-BUFFER {1008886DB3}>.
```

aartaka:
What this error message means is that there's no place for proxy in `internal-buffer`. I assume you tried adding proxy to `buffer`?

Add it to `web-buffer` instead, because that's the only class that has proxy slot.

aartaka:
Oh wait, our documentation is actually wrong here -- it lists `buffer` instead of `web-buffer`! Thanks for  flagging that, should be fixed with 4f9365288367664b5c13ce682f9a57fed3e993d5 :)

Zwo1in:
Indeed, it works with `web-buffer`, thanks!

PR #1530, opened 2021-06-19
Added instruction for installing Nyxt on Void Linux.

e-eight:
Added that Void Linux also provides a package for Nyxt, and how to install it.


Ambrevar:
Thanks a lot!

Issue #947, opened 2020-09-10
Feature Request: Make it easy to navigate to "next"/"prev" elements


jellelicht:
I've been a fan of vimium's `goNext` and `goPrevious` for the longest time; it basically boils down to finding and following a `<link>`, `<a>` or `<area>` tag with `rel` attribute of "next"/"prev" (or any user customizable regex, really).

Could something similar be implemented in nyxt?

Ambrevar:
Like on a GNU manual page where goNext would go the the next node?

I did implement something similar in the past for some browser, it's
very easy.

Would you like to send a patch? :)

jellelicht:
If I can get some pointers on a nice interactive workflow for developing parenscript, sure :+1:

Ambrevar:
Parenscript homepage:

https://common-lisp.net/project/parenscript/

Official tutorial:

https://common-lisp.net/project/parenscript/tutorial.html

For interactivity, you'll probably want Emacs with SLIME or SLY.
SLIME needs some configuration, SLY is more functional out-of-the-box.

In the REPL, you can expand macros (e.g. sly-macro-expand-all), inspect
objects (sly-inspect) and much more.

A REPL debugging tutorial that I like:

https://malisper.me/debugging-lisp-part-1-recompilation/

To run Parenscript in Nyxt, you need to use `define-parentscript`, then
call the resulting function in a `with-result` form.

Let me know if you need more pointers!

jellelicht:
Thanks!

Still one question: (how) can I set up the nyxt from guix
(2-pre-release-1) to be hackable with sly? Or would that rather be a
nyxt-from-git-only experience?

Ambrevar:
See the comments in build-scripts/guix.scm to load the dependencies.

To get the Nyxt source from Guix, you can do

    (uiop:run-program '("guix" "build" "-S" "nyxt") :output '(:string :stripped t))

which yields a path.

In your .sbclrc:

    (push #P"/path/to/nyxt" asdf:*central-registry*)


You could even get the path dynamically like this (untested):


    (push (pathname (uiop:run-program '("guix" "build" "-S" "nyxt") :output '(:string :stripped t)))
          asdf:*central-registry*)

Restart the REPL, then you should be able to

    ,load-system RET nyxt RET

jmercouris:
Please look at Pierre's dotfiles for Sly configuration.

Start swank in Nyxt, and then in Emacs, slime-connect :-). Port 4006 by default.

[...]

PR #1532, opened 2021-06-19
Feature: mimic vimium `goNext`/`goPrevious` inside Nyxt

pdelfino:
Hi guys,

I tried solving this [issue](https://github.com/atlas-engineer/nyxt/issues/947) . Since I was struggling with parenscript, I decided to attack the problem from a different perspective.

Instead of parenscript, I defined 2 new commands in Nyxt using the `define-bookmarklet-command` macro. It does not seem to make any difference in comparison to `define-command` for a GUI-only user perspective.

It works with [GNU manual](https://www.gnu.org/software/emacs/manual/html_node/emacs/Keys.html) pages and [XKCD](https://xkcd.com/2472/). This is a gif showing it in action: go-next and go-previous in Nyxt

This approach could be good for power users that lack knowledge of Common Lisp. I think the `define-bookmarklet-command` macro is way easier to understand than parenscript itself combined with `define-parenscript` macro. But this can be just my perspective.

Since my approach was different and I did not know if you would like it, I stopped before making it more general. It only works with HTML `rel` attribute being `next` or `prev`. If you think this approach is cool, I can put synonyms such as the ones `("next" "more" "newer" ">" "›" "→" "»" "≫" ">>")` available on @jellelicht [first try](https://gist.github.com/jellelicht/99e8b0dbb2533bea1888276d3fee39b5). 

This approach **will not** work on websites [like this](https://thelocalyarn.com/excursus/secretary/flatland/07-chapter.html). But [Vimium](https://chrome.google.com/webstore/detail/vimium/dbepggeogbaibhgnhhndojpepiihcmeb?hl=en) (the inspiration) **also does not work** on it.

### DIFF
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -222,6 +222,13 @@ and to index the top of the page.")
   "Scroll to top if no input element is active, forward event otherwise."
   (call-non-input-command-or-forward #'scroll-to-top :buffer buffer))
 
+(define-command go-next ()
+  "Navigate to the previous element according to the HTML rel attribute."
+  (pflet ((go-next ()
+                   (ps:chain document
+                             (query-selector-all "[rel=next]") 0 (click))))
+    (go-next)))
+
 (defun load-history-url (url-or-node
                          &key (buffer (current-buffer))
                               (message "History entry is already the current URL."))
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -229,6 +229,13 @@ and to index the top of the page.")
                              (query-selector-all "[rel=next]") 0 (click))))
     (go-next)))
 
+(define-command go-previous ()
+  "Navigate to the previous element according to the HTML rel attribute."
+  (pflet ((go-previous ()
+                       (ps:chain document
+                                 (query-selector-all "[rel=prev]") 0 (click))))
+    (go-previous)))
+
 (defun load-history-url (url-or-node
                          &key (buffer (current-buffer))
                               (message "History entry is already the current URL."))


jmercouris:
This is a good start. It should still be in parenscript though. Keep playing with the `ps:ps` function, you will get it!

pdelfino:
> This is a good start. It should still be in parenscript though. Keep playing with the `ps:ps` function, you will get it!

Thanks, John! Just for me to learn, why is it important to have it as parenscript? Why is that relevant from the technical perspective?

I guess the reason is because parenscript might be more extensible or powerful. I am not sure, though.

jmercouris:
Our codebase is Lisp, all of us are proficient as Lisp developers. Using parenscript means that we are easily able to understand what is going on, and to extend it.

For example, our parenscript functions can accept arguments. This code cannot.

pdelfino:
> Keep playing with the `ps:ps` function, you will get it!

`ps:ps` is a life saver, thanks for the hint :blush:

I've forced pushing a few squashed commits to make the feature using parenscript instead of bookmarklets.
This approach is way shorter than the [sketch](https://gist.github.com/jellelicht/99e8b0dbb2533bea1888276d3fee39b5) from @jellelicht. This fact makes me think if my approach is too naive.

Again, I have tested it with a GNU manual (emacs) and XKDC. I have noticed he uses things as ">, >, ›, →, »,≫, >>" in his implementation. But I could not find a web page with these rel attributes. Are they really necessary?

Do you recommend more pages for me to test this?

Should I extend this solution?

Ambrevar:
Many websites have "paging", like GitHub issues :)
You can look around and see if the "next" and "previous" links work.  

In any case, it's fine to merge a command that works on some websites for now.

pdelfino:
Hi @jmercouris ,

You did not write here about this here, but I think that on the live code review yesterday you suggested for me to change the code from using `define-command` such as:

```lisp
(define-command go-previous ()
  "Navigate to the previous element according to the HTML rel attribute."
  (pflet ((go-previous ()
                       (ps:chain document
                                 (query-selector-all "[rel=prev]") 0 (click))))
    (go-previous)))
```

Into  parenscript definition such as:

```lisp
(define-parenscript go-prev ()
  "Navigate to the previous element according to the HTML rel attribute."
  (ps:chain document
            (query-selector-all "[rel=prev]") 0 (click)))
```

I did it. I can call the `define-parenscript` in the REPL and it works as expected:

```lisp
CL-USER> (nyxt/web-mode::go-prev)
"undefined"
T
```

However, it **does not** show as an option to be executed on the prompt buffer.

1 - Is your intention that this `go-prev` and `go-next` will **not** be shown as an optional command in the prompt buffer? (I guess it is not your intention)

2 - If not, how can I turn this `define-parenscript` into a **visible** command for the GUI user?

I tried reading other examples in code base with parenscript but could not figure it out. Thanks

jmercouris:
I see, that's because define-parenscript converts to a `define-command` rather than a `define-command-global`. This is a problem indeed. Perhaps we would need an equivalent macro of `define-parenscript-global`.

jmercouris:
Ah, my apologies, define-parenscript creates a defun, not a define-command! Well then, the previous way /is/ the way to go. Ignore what I had said.

pdelfino:
> Ah, my apologies, define-parenscript creates a defun, not a define-command! Well then, the previous way /is/ the way to go. Ignore what I had said.

No problem, @jmercouris! 

> Many websites have "paging", like GitHub issues :)
> You can look around and see if the "next" and "previous" links work.

Thanks, @Ambrevar. This was **not** under my radar. Actually, *Vimium* (the OP inspiration) also works for a paging situation, like *github issues*. Unfortunately, my current implementation does not work.

> In any case, it's fine to merge a command that works on some websites for now.

Right. I will open a new issue so that this implementation can be extended/generalized, ok?

pdelfino:
> Many websites have "paging", like GitHub issues :)

Actually,  I realized this implementation works on the pagination of Github issues :) I can go to the next and the previous pages. 

I tried on [hacker news ](https://hn.algolia.com/?q=github)  but my implementation and vimium do not work there.

In Vimium, there is a binding for this:

> ]], [[  Follow the link labeled 'next' or '>' ('previous' or '<')
>           - helpful for browsing paginated sites

Should we use the same bindings? As far as I know, those are free. 

Is there any criteria/principle for leaving some command as bounded or unbounded?

Finally, if you were developing this, how would you test if the `next` on pagination works? How many websites would you try? 

How would you pick them? 

Better put this on master, and see if the issues arise?


I do not know what would be considered a **good practice** in Software Engineering for this feature situation. I lack common sense.

PR #1533, opened 2021-06-21
Fix history

Ambrevar:
This fixes the `implicit-visit` slot that was broken since forever.

It does not fix the scroll position however: indeed, `(nyxt:document-scroll-position (current-buffer))` should be run _before_ loading a new page.  The only spot where we can do this is in `preprocess-request`, but at this point it's too late, the request has already started and `document-scroll-position` returns nil.

I believe the only way to store the scroll position is in https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebPage.html#WebKitWebPage-send-request, but this requires that we switch from `on-decide-policy` to the WebKit extension architecture.

I could be missing something.  Thoughts?

@jmercouris @aartaka


Ambrevar:
Note that this may also fix the history navigation bug that would sometime get us to the wrong page when going back / forward.

Ambrevar:
If I'm right about scrolling, I suggest we remove the scroll saving from the code since it's a no-op and could be confusing to future readers.

jmercouris:
I see what you are doing here. I don't think you are missing anything. I suggest for 2.3.0 or something we try a minimal example with the minibrowser and an extension.

aartaka:
Wait, but how did scroll-position saving work in the first place?

jmercouris:
I never had it work for me :-D so no idea!

aartaka:
0_o

aartaka:
It clearly did work for me when I implemented it...

Ambrevar:
The bad news is that you might have tested it wrong.
From
https://github.com/atlas-engineer/nyxt/pull/1456:

> How Has This Been Tested:
> 
> Compiled.
> Ran.
> Moved in between several pages in history, back and forth.
> Position is restored up to a pixel!

When moving back and forth, WebKit is in charge for restoring the scroll
position, this worked in 2.0 already :p

To really test if it works, you can try one of these things:

- Delete a buffer then reopen it.
- Browse in a "Y" shape: Go to A, then B, back to A, then C, back to A,
  then B.

aartaka:
Yes, I tested it wrong xD

PR #1543, opened 2021-06-22
helper: define-command-prompt

Ambrevar:
The main purpose of this change is to check for the availability of the prompt buffer, and warn when it's not up.

Can we do better?

What do you think about inserting the implicit symbol `prompt-buffer` in the arg list?  Shall we name it `%prompt-buffer%` instead?

### DIFF
--- a/source/prompt-buffer-mode.lisp
+++ b/source/prompt-buffer-mode.lisp
@@ -102,6 +102,25 @@ Actions can be listed and run with `return-selection-over-action' (bound to
        "C-f" 'select-next-page
        "C-b" 'select-previous-page)))))
 
+(defmacro define-command-prompt (name (&rest arglist) &body body)
+  "Like `define-command' but automatically inserts a `prompt-buffer' argument
+and warn when its nil."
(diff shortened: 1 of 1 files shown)


Ambrevar:
@jmercouris Thoughts on this?

jmercouris:
What about still keeping `&key prompt-buffer` within the lambda list? Makes it a little more flexible?

Ambrevar:
And automatically replace it with `&key (prompt-buffer (current-buffer))`?

Ambrevar:
But then how would the macro tell the different between the key argument
that designate the prompt-buffer, and the other ones?

jmercouris:
Nevermind, you are right, there would be ambiguity. No other suggestions.

Ambrevar:
Got an idea, we can make the first argument special.

I've updated the pull request to reflect this.
@jmercouris @aartaka What do you think?

jmercouris:
I like this approach a lot. I concur. No other comments from me :-)

aartaka:
Feels alright :)

Issue #1310, opened 2021-04-19
Automatically bookmark frequently used pages


Ambrevar:
Add a customizable visit frequency threshold after which Nyxt automatically bookmarks the URL.
Optionally, have Nyxt query the user for tags.

This would be a good middle-ground between the global history and manual bookmark management.  Indeed, manual bookmarks can quickly get overwhelming which may lead users to not use them at all.

Another use case is for when the user simply forgets to bookmark a page.

jmercouris:
this is a great idea!

aartaka:
I'd optionally add a prompt like "Do you want to bookmark this page?" to make sure there are no unwanted bookmarks.

pdelfino:
Hi guys,

I am trying this one. Let me ask one thing, what should be considered for the threshold? 

1- Only **explicit** visits (URL visited by prompt-buffer request)? 
2 - Only **implicit** visits (URL visited by following links)? 
3 - Both of them (explicit and implicit) as a sum?
4 - Different thresholds for each kind of visit?    [the best in my IMHO]

I think this is not a straightforward decision. Curious about your thoughts since you have a great experience with Nyxt as users and coders.

aartaka:
> Hi guys,
> 
> I am trying this one. Let me ask one thing, what should be considered for the threshold?

Currently, the only type of visits we store is explicit ones (implicit visits are always zero.) Thus, the choice is lacking here — just rely on explicit visits.

pdelfino:
Thanks for the help, @aartaka!

> Currently, the only type of visits we store is explicit ones (implicit visits are always zero.) Thus, the choice is lacking here — just rely on explicit visits.

About your comment on implicit visits being always zero, something weird just happened. First, I run on the REPL:
`
NYXT> (prompt :sources (list (make-instance 'history-disowned-source :buffer (current-buffer)))
`
The code above is *part* of the ` is not delete-history-entry` command definition. After running it, Nyxt's GUI presents a prompt.
So, I decided to delete an irrelevant history query. The REPL then returned the following:

`
(#<HISTORY-ENTRY {10025864F3}>)
`
Just because I am trying to learn, I inspected it:
```
A proper list:
0: #<NYXT::HISTORY-ENTRY {10025864F3}>
```
And now the odd part:

```
#<HISTORY-ENTRY {10025864F3}>
--------------------
Class: #<STANDARD-CLASS NYXT::HISTORY-ENTRY>
--------------------
[...]

PR #1548, opened 2021-06-23
Accidentally useful feature: Bookmark frequently used pages looping over selected history entries

pdelfino:
Hi guys,

This the first version for issue #1310. I coded a function called `bookmark-frequently-visited-urls`, it glues together several `%auxiliary` functions. When I called this function on my Slime's REPL, the bookmark based on high implicit visits worked out as expected. 

In order to make it an automatic process as pages are loading, I tried inserting the function `bookmark-frequently-visited-urls` as a function call `(bookmark-frequently-visited-urls)` on `history-add`. This stands as a commentary, I will explain _why._ bellow.

Unfortunately, this insertion of a function call by the end of `history-add` crashes Nyxt when some page reaches the threshold. Nyxt freezes and I need to kill it using my shell `xkill`. I do not know why this crash is happening. Probably, something on my side - not a bug. 

I hope we will figure this out on the live code review with @jmercouris tomorrow (June, the 24th). Hopefully, this at least a good start.


source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
                          history))
       (let* ((entry (htree:data (htree:current (htree:owner history (id buffer))))))
         (setf (title entry) title)
-        (incf (implicit-visits entry))))))
+        (incf (implicit-visits entry))
+       ; (bookmark-frequently-visited-urls) crashes Nyxt completely. If I run on REPL it works
+        ))))
+
+(defparameter *threshold-frequent-url-to-bookmark* 20)
```
jmercouris, line 102: Please convert this to a slot of theh browser class. Then you can access it like this:

`(url->bookmark-visit-threshold *browser*)`

source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
...
+
+(defun %urls-visited-over-threshold (threshold)
+  "Return a list '(number-of-visits url-instance url-title) 
+of all URLs which were visited more times than the threshold."
+  (let*
+      ((history-entries-raw
+         (with-data-unsafe (history (history-path (current-buffer)))
+           (alex:hash-table-keys (htree:entries history))))
+       (pair-frequency-url (mapcar #'(lambda (e) (list (implicit-visits (htree:data e))
```
jmercouris, line 111: No need to construct a list of lsits here, just use the objects directly.

source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
...
+                                                       (url (htree:data e))
+                                                       (title (htree:data e))))
+                                   history-entries-raw)))
+    (remove-if-not #'(lambda (e) (> (first e) threshold)) pair-frequency-url)))
```
jmercouris, line 115: `(remove-if-not #'(lambda (e) (> (implicit-visits (htree:data e)) threshold)) history-entries-raw)`

parenthesis may be incorrect...!

source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
...
+
+(defun %list-bookmarks ()
```
jmercouris, line 117: If you think it is useful you may want to add a #+nil above it so that the user must manually evaluate it to use it.

source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
...
+  "List all instances of URLs currently bookmarked."
+  (mapcar #'(lambda (e) (url e))
+          (with-data-access (bookmarks (bookmarks-path (current-buffer)))
+            bookmarks)))
+
+(defun %%concatenate-url (url-instance)
```
jmercouris, line 123: You can use `(quri:render-uri url-instance)` instead of this function.

source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
...
+  "After receiving an instace of an URL, concatenate
+scheme, host and path to form a fully complete URL address."
+  (let ((scheme (quri:uri-scheme url-instance))
+        (host   (quri:uri-host   url-instance))
+        (path   (quri:uri-path   url-instance)))
+    (concatenate 'string scheme "://" host path)))
+
+(defun %is-url-new-to-bookmarks (url)
```
jmercouris, line 131: Name should `%is-url-new-to-bookmarks-p` if it is a predicate, always use a `-p` suffix.

source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
...
+  "Return NIL if URL is already bookmarked or
+return the URL's full address if it is new to the bookmark list."
+  (let ((bookmarks-list (mapcar #'(lambda (e) (%%concatenate-url (url e)))
```
jmercouris, line 134: You could replace the whole function with a single line using `find` and `:key`.

source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
...
+
+(defun bookmark-frequently-visited-urls ()
+  "Add websites frequently visited that are not
+included on the bookmarklist."
+  (dolist (url-data  (%urls-visited-over-threshold *threshold-frequent-url-to-bookmark*))
+    (let* ((url-instance (second url-data))
+           (url-address (%%concatenate-url url-instance))
+           (title        (third url-data)))
+      (if (%is-url-new-to-bookmarks url-address)
+          (%bookmark-a-frequently-visited-url url-instance title)))))
```
jmercouris, line 162: You could instead `(bookmark-url :url (quri:render-uri url-instance))`

source/history.lisp
```diff
@@ -95,7 +95,71 @@ The `implicit-visits' count is incremented."
...
+  "Return NIL if URL is already bookmarked or
+return the URL's full address if it is new to the bookmark list."
+  (let ((bookmarks-list (mapcar #'(lambda (e) (%%concatenate-url (url e)))
+                                (with-data-access (bookmarks (bookmarks-path (current-buffer)))
+                                  bookmarks))))
+    (if (member url bookmarks-list :test #'string=)
+        nil
+        url)))
+
+(defun %bookmark-a-frequently-visited-url (url-instance title-data)
```
jmercouris, line 141: You can delete this function.

source/history.lisp
pdelfino: John corrected this during our live meeting. But I think he forgot the typo. For future reference, in case someone of the team/or/and an user trying to contribute sees this after searching for something similar, the function to receive an  url-instance and to return its full formed address such as `https://www.example.com` is `NYXT>(render-url url-instance)` inside the `nyxt` package. 

It is also possible to solve this with the `quri` package as `(quri:render-uri url-instance)` but you need to use with the **i** as the end.  The `quri` package only treats renderization of URIs, since URIs are a superset of URLs they also work for this concrete case.

source/history.lisp
pdelfino: Hi John,
I have just realized there is a possible downside on your suggestion. Currently, `bookmkar-url` does **not** get the title of the page (which is a very important info to bookmarks, imho). My approach above does get the title. 

1 - What do you think? Do you agree that the title is also an important data to bookmarks?

2 - If so:
 (a) should I change `bookmark-url` in source code so that it **gets** the title of the page automatically in my case and other applications?; or,  
 (b) should I keep the current approach using my implementation?
 (c) should I use your suggestion with `bookmark-add` and do **not** include the title **for now** even though you think title would be a good data (we can fix that later)?

source/history.lisp
pdelfino: Hi John,

While improving the code I realized something. On one hand, predicates return `T` or `NIL`. 
 On the other hand, this function returns the URL address if is unprecedented to the bookmark list or returns `NIL` if the URL is already bookmarked. 
 
 Thus, should the name really change to end with a `-p` indicating a predicate?  
 
 I reckon the name change can be misleading! People may think it will always return a boolean.

source/history.lisp
pdelfino: John,

I was thinking about this. Maybe I did something wrong but as I tried doing it with `find` and `:key` the final solution was, actually, worse than this one (imho). 

Ultimately, abstracting some stuff, what we want here is:
```lisp
NYXT> (is-url-new-to-bookmarks    "url1" '("url1" "url2" "url3"))
NIL  ; URL address is already bookmarked

NYXT> (is-url-new-to-bookmarks    "url1" '( "url2" "url3"))
"url1"  ; URL address is new! Returning the fresh one to bookmark it later
```

My current approach is:
```
(defun is-url-new-to-bookmarks (str str-list)
  (if (member str str-list :test #'string=)
      nil
      str))
```

When I tried using `find`, a `not` wrapping became a necessity. Looking only to this abstracted tiny problem and the info provided, do you still think `find` and `:key` would be better, here?

If so, how would you do it in this **tiny** abstracted problem described above?

pdelfino:
Hi guys,

I am sending a new updated version of this. Basically, I incorporated the suggestions of @jmercouris. His comments above were the guidelines. 

There is no comment about using `labels` but he mentioned it during the live meeting. Thus, I also implemented. I have also squashed some commits through an interactive rebase. Magit is really cool, thanks for introducing it to me :)

Some doubts about the suggestions appeared while implementing the advice. I put the doubts as replies to the suggestions above.

There still **one important thing to do**: learn how to make my `bookmark-frequently-visited-urls` run *everytime a webpage is loaded on Nyxt* (without crashing it).

source/history.lisp
jmercouris: Aha! Good to know.

source/history.lisp
jmercouris: You are technically right. However, the function is very much a predicate. It returns a truthy value. Whether they choose to actually discard said value doesn't matter. They could treat it like a boolean, and everything would still work as expected.

source/history.lisp
jmercouris: That's only true if your list of bookmarks is already a string. If your list of bookmarks is just the actual bookmark objects then it would work. In that case you would do something like:
```
(find url bookmarks :key #'url)

```

in this hypothetical example it would call the URL function on each BOOKMARK and see if it matches the URL. I don't remember exactly what one needs to invoke against a bookmark object to get the URL, however that is the gist. Is that clear?

source/history.lisp
jmercouris: Getting the title automatically would actually be probably pretty hard.
I would suggest option C. Not retrieving the title is not the end of the world. It can always be obtained later, or fetched via another function.

jmercouris:
Should it happen EVERY time a page is loaded in Nyxt? Or should it happen on startup?

jmercouris:
Or perhaps on some interval of the user's choosing?

source/history.lisp
```diff
@@ -78,12 +78,43 @@ class."
...
+                                   history-entries-raw))
+                  (urls-frequently-visited
+                    (mapcar #'(lambda (e) (url (htree:data e))) history-entries-above-threshold)))
+               urls-frequently-visited))
+           (is-url-new-to-bookmarks (url-address)
+             (let ((bookmarks-address-list
+                     (mapcar #'(lambda (e)  (render-url (url e)))
+                             (with-data-access (bookmarks (bookmarks-path (current-buffer)))
+                                                bookmarks)))) 
+               (if (member url-address bookmarks-address-list :test #'string=)
```
jmercouris, line 105: this approach is perfectly valid, no need to convert it to use `find`.

jmercouris:
Looking at the source once again, a very significant improvement. Well done! There are a few minor questions to be answered, like when do we run this function?

source/history.lisp
pdelfino: Ok, it makes sense. I guess this is is the "if is not false, it work as true" approach in `if` and `cond` clauses on Lisp. Already implemented :ok_hand:

source/history.lisp
pdelfino: :crystal_ball: clear 
It was new for me this approach with `find` and `:key`. Thanks for teaching it.

source/history.lisp
pdelfino: > I would suggest option C. 

Luckily, this is the one already implemented :four_leaf_clover:

> Getting the title automatically would actually be probably pretty hard.

Could you briefly please educate me on why this would be hard?

On my first try, I was able to do it. I guess it was *easier* because I used the context of the *history objects* being part of the process. The **urls instances** on history have this data with easy access.

pdelfino:
> Should it happen EVERY time a page is loaded in Nyxt? Or should it happen on startup?

I tried some stuff. When, you have multiple URLs above the threshold, this command is a bit annoying. Because it is running the loop (dolist) and the prompt buffer will be sequentially asking the user for tags *without letting him know for which URL will the tags be associated with*. Obviously, we can tweak that. But, considering the current design and keeping everything else constant, it would be better to run it every time a page is loaded.

I am just affraid about performance issues. It looks expensive to run it every time a page is loaded. "premature optimization is the rool all evil", though... 

> Looking at the source once again, a very significant improvement. Well done! 

I was really happy to receive this compliment. Thank you!

jmercouris:
It is far too expensive to run on every page load, hence my suggestion. On startup it may make sense, or as a command the user can run like "bookmark-frequently-used-urls".

pdelfino:
> It is far too expensive to run on every page load, hence my suggestion. On startup it may make sense, or as a command, the user can run like "bookmark-frequently-used-URLs".

ok, I think transforming this into a command would be the best option. I will adapt the code so that the user will know via an `echo` command which page is being bookmarked, is that ok?

Moreover, I will insert the suggestion made by @aartaka so that there is a question if the user wants to bookmark it or not.

I think a **blocklist of non-bookmarked URLs ** (the pages the user decides to not bookmark) should be stored somewhere. Hence, the user won't be asked about the same URL multiple times. Maybe this could be stored as a list inside a slot value.

Would you suggest a specific class for this slot value to be created? 

Browser class maybe?

jmercouris:
I think that could certainly be a future enhancement. I would not suggest a slot be added for this purpose. If there is a URL that they don't wish to bookmark, they will just have to specify every time they ignore it, I don't think it is that big of a problem.

Everything else sounds OK to me!

pdelfino:
> I think that could certainly be a future enhancement. I would not suggest a slot be added for this purpose. If there is a URL that they don't wish to bookmark, they will just have to specify every time they ignore it, I don't think it is that big of a problem.

Ok, noted as a future enhancement :memo:

> Everything else sounds OK to me!

I mentioned the use of `echo` and the `mini-buffer` to tell the user which URL was being bookmarked, and you said it was OK.
However, it turns out that @aartaka suggested the use of the macro `if-confirm` for his own feature suggestion in a private email. This  `if-confirm` macro makes this new feature really cool! Check-it out :wink:

source/history.lisp
```diff
@@ -78,12 +78,44 @@ class."
...
   (htree:make :key 'history-tree-key :current-owner-id (id buffer)))
 
+(define-command bookmark-frequently-visited-urls ()
+  "Add websites frequently visited that are not included on the bookmarklist.
+
+The local function urls-visited-over-thresold returns all URLs instances which
+were visited more times than the threshold.
+
+The local function is-url-new-to-bookmarks-p returns the URL address itself if it
+new to the bookmark list and NIL if it is already there "
```
jmercouris, line 88: please move from documentation string to comments as it is not relevant for users.

jmercouris:
Please remove the submodule commit.

source/history.lisp
```diff
@@ -78,12 +78,44 @@ class."
...
+  (labels ((urls-visited-over-threshold (threshold)
+             (let*
+                 ((history-entries-raw
```
jmercouris, line 91: I would move this up to the previous line.

source/history.lisp
```diff
@@ -94,6 +126,7 @@ The `implicit-visits' count is incremented."
                                         :title title)
                          history))
       (let* ((entry (htree:data (htree:current (htree:owner history (id buffer))))))
+
```
jmercouris, line 129: Please remove the whitespace here.

jmercouris:
Hi Pedro, final final thing, and then we merge :-)

can you please remove the commits at the beginning that update the submodule and are unrelated to these changes?

source/browser.lisp
```diff
@@ -72,6 +72,12 @@ Most recent messages are first.")
 This history is used if no history is specified for a given prompt buffer.")
    (set-url-history (make-ring)
                     :documentation "The history of all URLs set via set-url")
+   (url->bookmark-visit-threshold :initform 20
+                                  :export t
+                                  :documentation "The threshold definition for
+                                  what makes an URL be considered frequently
+                                  visited")
```
Ambrevar, line 79: "a URL"
"definition" is not needed.
Period is missing.

source/browser.lisp
```diff
@@ -72,6 +72,12 @@ Most recent messages are first.")
 This history is used if no history is specified for a given prompt buffer.")
    (set-url-history (make-ring)
                     :documentation "The history of all URLs set via set-url")
+   (url->bookmark-visit-threshold :initform 20
```
Ambrevar, line 75: Why the `->`?  The arrow is usually meant for conversion.

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
   "Return a new global history tree for `history-entry' data."
   (htree:make :key 'history-tree-key :current-owner-id (id buffer)))
 
+(define-command bookmark-frequently-visited-urls ()
+  "Add websites frequently visited that are not included on the bookmarklist."
+  ;; The local function urls-visited-over-thresold returns all URLs instances which
+  ;; were visited more times than the threshold.
```
Ambrevar, line 84: You can move this comment to the docstring of `urls-visited-over-threshold`.

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
   "Return a new global history tree for `history-entry' data."
   (htree:make :key 'history-tree-key :current-owner-id (id buffer)))
 
+(define-command bookmark-frequently-visited-urls ()
+  "Add websites frequently visited that are not included on the bookmarklist."
+  ;; The local function urls-visited-over-thresold returns all URLs instances which
```
Ambrevar, line 83: You can surround the function name with "`" and "'" so that it gets highlighted in Emacs.
Also there is a typo ;)

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
   "Return a new global history tree for `history-entry' data."
   (htree:make :key 'history-tree-key :current-owner-id (id buffer)))
 
+(define-command bookmark-frequently-visited-urls ()
+  "Add websites frequently visited that are not included on the bookmarklist."
```
Ambrevar, line 82: Grammar: "Add frequently visited websites..."
"included on" -> "included in".

Technically we are not adding websites though :p 


"bookmarklist" -> "bookmarks".

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+           ;; new to the bookmark list and NIL if it is already there 
+           (is-url-new-to-bookmarks-p (url-address)
+             (let ((bookmarks-address-list
+                     (mapcar #'(lambda (e)  (render-url (url e)))
+                             (with-data-access (bookmarks (bookmarks-path (current-buffer)))
+                                                bookmarks)))) 
+               (if (member url-address bookmarks-address-list :test #'string=)
+                    nil
+                    url-address))))
+    (dolist (url-instance  (urls-visited-over-threshold (url->bookmark-visit-threshold *browser*)))
```
Ambrevar, line 105: Spurious double-space after `url-instance`.

Ambrevar: `url-instance` is redundant, you can just call it `url`.

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+                      (alex:hash-table-keys (htree:entries history))))
+                  (history-entries-above-threshold  
+                    (remove-if-not #'(lambda (e) (> (implicit-visits (htree:data e)) threshold))
+                                   history-entries-raw))
+                  (urls-frequently-visited
+                    (mapcar #'(lambda (e) (url (htree:data e))) history-entries-above-threshold)))
+                    urls-frequently-visited))
+           ;; The local function is-url-new-to-bookmarks-p returns the URL address itself if it
+           ;; new to the bookmark list and NIL if it is already there 
+           (is-url-new-to-bookmarks-p (url-address)
```
Ambrevar, line 97: Shorter: `bookmarked-url-p`

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+                    (with-data-unsafe (history (history-path (current-buffer)))
+                      (alex:hash-table-keys (htree:entries history))))
+                  (history-entries-above-threshold  
+                    (remove-if-not #'(lambda (e) (> (implicit-visits (htree:data e)) threshold))
+                                   history-entries-raw))
+                  (urls-frequently-visited
+                    (mapcar #'(lambda (e) (url (htree:data e))) history-entries-above-threshold)))
+                    urls-frequently-visited))
+           ;; The local function is-url-new-to-bookmarks-p returns the URL address itself if it
+           ;; new to the bookmark list and NIL if it is already there
```
Ambrevar, line 96: Same, you can move this comment to the docstring.

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+  (labels ((urls-visited-over-threshold (threshold)
+            (let* ((history-entries-raw
+                    (with-data-unsafe (history (history-path (current-buffer)))
+                      (alex:hash-table-keys (htree:entries history))))
+                  (history-entries-above-threshold  
+                    (remove-if-not #'(lambda (e) (> (implicit-visits (htree:data e)) threshold))
+                                   history-entries-raw))
+                  (urls-frequently-visited
+                    (mapcar #'(lambda (e) (url (htree:data e))) history-entries-above-threshold)))
+                    urls-frequently-visited))
```
Ambrevar, line 94: Indentation is off, this does not read right.

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+             (let ((bookmarks-address-list
+                     (mapcar #'(lambda (e)  (render-url (url e)))
+                             (with-data-access (bookmarks (bookmarks-path (current-buffer)))
+                                                bookmarks)))) 
+               (if (member url-address bookmarks-address-list :test #'string=)
+                    nil
+                    url-address))))
+    (dolist (url-instance  (urls-visited-over-threshold (url->bookmark-visit-threshold *browser*)))
+      (if (is-url-new-to-bookmarks-p (render-url url-instance))
+          (if-confirm ("Do you want to bookmark ~a?" (render-url url-instance))
```
Ambrevar, line 107: Maybe just "Bookmark ~a?"?

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+                      (bookmark-url :url (render-url url-instance)))))))
```
Ambrevar, line 108: Let-bind `(reader-url url-instance)` to avoid rendering it twice.

source/history.lisp
Ambrevar: Maybe just return the `mapcar` directly, no need to let-bind `urls-frequently-visited`.

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+                    (remove-if-not #'(lambda (e) (> (implicit-visits (htree:data e)) threshold))
+                                   history-entries-raw))
+                  (urls-frequently-visited
+                    (mapcar #'(lambda (e) (url (htree:data e))) history-entries-above-threshold)))
+                    urls-frequently-visited))
+           ;; The local function is-url-new-to-bookmarks-p returns the URL address itself if it
+           ;; new to the bookmark list and NIL if it is already there 
+           (is-url-new-to-bookmarks-p (url-address)
+             (let ((bookmarks-address-list
+                     (mapcar #'(lambda (e)  (render-url (url e)))
```
Ambrevar, line 99: Spurious double-space.

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+                             (with-data-access (bookmarks (bookmarks-path (current-buffer)))
+                                                bookmarks))))
```
Ambrevar, line 101: Don't use `with-data-access` if you don't need read-only access to the data.  Prefer `with-data-unsafe` or `get-data`.

source/history.lisp
```diff
@@ -78,12 +78,41 @@ class."
...
+                    (mapcar #'(lambda (e) (url (htree:data e))) history-entries-above-threshold)))
+                    urls-frequently-visited))
+           ;; The local function is-url-new-to-bookmarks-p returns the URL address itself if it
+           ;; new to the bookmark list and NIL if it is already there 
+           (is-url-new-to-bookmarks-p (url-address)
+             (let ((bookmarks-address-list
+                     (mapcar #'(lambda (e)  (render-url (url e)))
+                             (with-data-access (bookmarks (bookmarks-path (current-buffer)))
+                                                bookmarks)))) 
+               (if (member url-address bookmarks-address-list :test #'string=)
```
Ambrevar, line 102: You are comparing the bookmark URLs by comparing the renderer strings.
Just compare them using `quri=` without rendering them.

Ambrevar:
Sorry, I went for another round :)

If I get it right, for now we just have a command.
So what remains to be done (possibly in another pull request) is to hook this in automatically, e.g. on URL load.

source/browser.lisp
jmercouris: I suggested the name, and it makes sense to me, it is the threshold at which we convert a URL to a bookmark :-)

source/history.lisp
jmercouris: Ah, true, I didn't even think of them being docstrings for the labels!

jmercouris:
I believe it should just be a command, it is WAY too expensive to run on every URL load. Also possibly very annoying to be prompted as you are randomly browsing the internet.

Ambrevar:
Why expensive?  It's a single hash-table lookup, no?

jmercouris:
It is not a single hash table look-up, it is doing many comparisons for equality, and it is doing a look up of the history.

jmercouris:
Depending on the amount of history items and bookmarks you have, it grows exponentially.

source/browser.lisp
Ambrevar: Hmmm, for a user-facing option I'd rather have something less ambiguous.
What about "frequent-visit-bookmark-threshold"?

Ambrevar:
Why history items?  You only check if the current URL is in the bookmarks.
Misunderstanding?

(Nit: it would grow polynomially if you were to cross all history entries and bookmark entries :p)

Ambrevar:
OK, just saw the mapcar over history, this is wrong, you only need to query the current URL implicit visit.

Ambrevar:
To sum up, the algorithm is pretty simple:

- Query the implicit-visit of the current URL (zero-cost)
- Compare with the threshold.
- If above threshold and not already bookmarked (single hash-table lookup), bookmark.

That's it :)

Ambrevar:
@pdelfino If you are starting to get tired with these rounds of review, I totally understand, don't hesitate to tell me if you'd rather that I take over to merge this.  As you prefer :)

source/history.lisp
pdelfino: > You can move this comment to the docstring of `urls-visited-over-threshold`.

It took me awhile to understand what you meant. I decided to mimick this snippet on history.lisp:

```lisp
(labels ((restore-buffers (history)
             "For each owner, make buffer, swap owner identifier for buffer id ...
```

I hope it is what you mean.

source/browser.lisp
pdelfino: > What about "frequent-visit-bookmark-threshold"?

My first attempt was `threshold-frequent-url-to-bookmark`. I also thought it was strange the `->`. But I am new to the code base and its conventions, so I followed John's suggestion.

`frequent-visit-bookmark-threshold` or `threshold-frequent-url-to-bookmark` both seem better than `url->bookmark-visit-threshold` to my **naive** eyes. 

Actually, I preferer `threshold` first. Important things first. Being a threshold seems the most important attribute. But it is up to you guys. Not a crucial thing at the end of the day.

source/history.lisp
pdelfino: Ok, I fixed everything. Thanks for the feedback.

> Technically we are not adding websites though :p

"Add frequently visited **URLs** that are not included in the bookmarks."

source/history.lisp
pdelfino: Implemented.

source/history.lisp
pdelfino: Interesting trick! I did not know about it. Thanks for teaching it :)

source/history.lisp
pdelfino: Good advice.  `(render-url url-instance)` was being called and re-rendered 3 times, actually... It was even worse...

source/history.lisp
pdelfino: Actually, I am getting an error message

If I try:

`(define-command "`bookmark-frequently-visited-urls"`

Or

`(define-command "'bookmark-frequently-visited-urls" ()`

Error message

```lisp

Execution of a form compiled with errors.
Form:
  (DEFINE-COMMAND "'bookmark-frequently-visited-urls"
    NIL
  "Add frequently visited URLs that are not included in the bookmarks."
  (LABELS ((URLS-VISITED-OVER-THRESHOLD (THRESHOLD)
             "The local function urls-visited-over-thresold returns all URLs
             instances which were visited more times than the threshold."
             (LET* (# # #)
               URLS-FREQUENTLY-VISITED))
           (IS-URL-NEW-TO-BOOKMARKS-P (URL-ADDRESS)
             "The local function is-url-new-to-bookmarks-p returns the URL
             address itself if it new to the bookmark list and NIL if it is
             already there "
             (LET (#)
               (IF #
                   NIL
                   URL-ADDRESS))))
    (DOLIST
        (URL-INSTANCE
         (URLS-VISITED-OVER-THRESHOLD
          (URL->BOOKMARK-VISIT-THRESHOLD *BROWSER*)))
      (LET ((URL-ADDRESS #))
        (IF (IS-URL-NEW-TO-BOOKMARKS-P URL-ADDRESS)
            (IF-CONFIRM # #))))))
Compile-time error:
  (during macroexpansion of (DEFINE-COMMAND "'bookmark-frequently-visited-urls"
    ...))
The value
  "'bookmark-frequently-visited-urls"
is not of type
  SYMBOL
  ```

File: source/history.lisp (hunk 25 above)
Comment (pdelfino) [reply source/history.lisp:99]: Thanks, do you use some tool to catch those things? Or is it just a trained eye?

Comment (pdelfino) [reply source/history.lisp:99]: `white-space-mode` maybe?

**Comment by pdelfino** [comment]
> @pdelfino If you are starting to get tired with these rounds of review, I totally understand, don't hesitate to tell me if you'd rather that I take over to merge this. As you prefer :)

It is fine for me! If I had always had thorough reviews like this over my code, by now I would be way better as a programmer. 

I am just afraid that you guys could run out of patience :sweat_smile: . I think this process is helping me understand better the code base and how you guys think.

File: source/history.lisp (hunk 26 above)
Comment (pdelfino) [reply source/history.lisp:101]: Pierre, just to be clear. I am confused. On `with-data-unsafe` documentation:

> Beware: this is not a thread-safe macro, so data may be altered while
> you use this macro!

You said: 

> Don't use with-data-access if you don't need read-only access to the data.

Did you mean: "Do not use with-data-access it you **only** need read-only-access to data, [since it would be a waste of resources]"
?

In this case, I am just reading the bookmarks data, correct?

After this code snipet, by the end of the code, it will write it the bookmark data adding stuff.

Comment (pdelfino) [reply source/history.lisp:101]: Anyway, I have changed to `with-data-unsafe`

File: source/history.lisp (hunk 27 above)
Comment (pdelfino) [reply source/history.lisp:102]: I could not find `quri=` using `git grep "quri="` on the codebase. 
However, I **did** find `quri:uri=`. Just to be sure, were you thinking about this one?

**Comment by pdelfino** [comment]
Guys,

Something occurred to me.

> To sum up, the algorithm is pretty simple:
> 
>     * Query the implicit-visit of the current URL (zero-cost)
> 
>     * Compare with the threshold.
> 
>     * If above threshold and not already bookmarked (single hash-table lookup), bookmark. 
> 
> 
> That's it :)

If my understanding is correct, it turns out I have built something different from the issue original intention. **Sory** :sweat_smile:

But, I think it has some value :point_up:. Both things the original intention and what I have written with John's help are complementary!

For instance, suppose user Plato has a big history file. A few months ago, he was doing exploratory readings on the web for his PhD thesis. Now, he is in a more advanced phase. Thus, afraid you losing interesting stuff, he decides to reduce the threshold value to 5 implicit and **run** the **command** (already written) `bookmark-frequently-visited-urls`. 

The prompt-buffer reads all the history-file and shows multiple webpages over this threshold, Plato decides as a `No` for most because they are not helpul for his thesis, but some of the webpages really deserve a `Yes` with cool tags.

After mister Plato (almost Doctor Plato) has done this big bookmarking event about the past in a bulk, he starts writing his PhD thesis. He is already burdened with references and material. He needs to focus. 

Thus, he decides to increase the threshold putting it up to 50. Every time he reaches 50 implicit visits on the same page _and_ it was not bookmarked yet, the prompt buffer asks if he wants to bookmark it. To do so I need to write Pierre's original intention as a light code to run everytime a page is loaded.

Nyxt is so powerful that it allows people to *manipulate* what they consider important (favorities/bookmarks) transcending time haha. This is like *back-to-the future* for bookmarks :zap: :watch: :zap:

Curious about your thoughts!

File: source/history.lisp (hunk 17 above)
Comment (Ambrevar) [reply source/history.lisp:83]: I meant the function name _in the docstring_.

File: source/history.lisp (hunk 16 above)
Comment (Ambrevar) [reply source/history.lisp:84]: Yes!

File: source/browser.lisp (hunk 15 above)
Comment (Ambrevar) [reply source/browser.lisp:75]: Another important aspect is that symbols read well in English.

Most other programming languages don't do this, but in Lisp it's an important idiom.

File: source/history.lisp (hunk 25 above)
Comment (Ambrevar) [reply source/history.lisp:99]: Could use white-space-mode, but in this case it's just the trained eye.

In the end you should not have to handle this manually, you can use automatic indentation which also removes unnecessary spaces.  I use Lispy but I believe Paredit or Parinfer can also do this.

File: source/history.lisp (hunk 26 above)
Comment (Ambrevar) [reply source/history.lisp:101]: Sorry, you are right, I mean "if you only need read-only access".

File: source/history.lisp (hunk 27 above)
Comment (Ambrevar) [reply source/history.lisp:102]: Yup, sorry about the typo!

**Comment by Ambrevar** [comment]
If you want to loop over all history entries and bookmark those above some threshold, this works too:

- Implement the aforementioned algorithm in a function (not a command), say `bookmark-frequent-visit`.
- Add a `bookmark-frequent-history-entries` command which prompts the user for history entries (the user can select them all), and then call `bookmark-frequent-visit` over the selection.

Voilà :)

**Comment by pdelfino** [comment]
> Hi Pedro,

Hi John

> can you please remove the commits at the beginning that update the submodule and are unrelated to these changes?

I think I successfully removed them. Some weird things happened while trying an interactive rebase maybe because there was a **merge** between the submodule stuff (my initial change) and the additions necessary for the feature (last changes).

When I pushed things, I was surprised to see the arise of merge conflicts. Not sure what is the best way to solve this. Would be cool to receive advice.

**Comment by pdelfino** [comment]
Hi guys,

1 - On my last comment I asked advice about solving the best way to solve the merge conflicts. I solved them following a tutorial. 

2 - The command works in the small sample of tests I did. I tried to follow @Ambrevar recipe:

```
    Implement the aforementioned algorithm in a function (not a command), say bookmark-frequent-visit.
    Add a bookmark-frequent-history-entries command which prompts the user for history entries (the user can select them all), and then call bookmark-frequent-visit over the selection.
```

3 - Those `_build/submodules/cl-webkit` changes really annoy me. I wrongly committed them along with useful stuff. It was hard to fix it. I was almost opening a new branch/PR. But the Gods of Git on the internet helped me.

4 - Should I squash some commits? Should I keep things as they are? Is this good to go?

**Comment by Ambrevar** [comment]
Yes, you can squash all the commits that you deem irrelevant for merging.

**Comment by pdelfino** [comment]
> Yes, you can squash all the commits that you deem irrelevant for merging.

Unfortunately, I still lack the common sense for this matter. From my naive perspective, these commits below seem atomic and useful if I need to get back at the development:

```
history.lisp: define a function to check all URLs implicitly visited above a certain threshold.

browser.lisp: make the threshold for frequent implicit visits a slot value.

history.lisp: add a function to bookmark URLs that are frequently visited.

history.lisp: change name of the function to indicate it is a predicate.

history.lisp:  transform bookmark-frequently-visited-urls from a function into a command.

history.lisp: on bookmark-frequently-visited-urls command, insert a `if-confirm` macro.

history.lisp: remove the comment with an exploratory attempt to run bookmark-frequently-visited-urls everytime a page is loaded.

browser.lisp: improve documentation (length, style, and grammar) of the new slot.

history.lisp: transform documentation as comments into docstrings for the local functions

history.lisp: improve grammar.

```

Would you have the same opinion? Would you keep all of them? Which would you squash?

pdelfino:
Guys,

As I was doing this task, some bugs were found and fixed involving implicit visits. Bringing the fixes from master, I merged them on this branch.

Unfortunately, I did not rebase to put my changes on top before merging.  There was a problem with the submodules too that I accidentally added/staged and committed the unwanted changes blended with useful stuff having to revert them in isolation later. Hence, it was not a fast-forward merge and it is a mess now to squash commits.

Thus, I am following @jmercouris advice, closing this PR, and opening a new "cleaner" PR transposing **only** the appropriate changes. 

This situation reminds me of this [XKCD](https://xkcd.com/1597/) cartoon :laughing:. Sorry for the inconvenience.

Issue #1560, opened 2021-06-26
Nyxt buffer's after crash lose owner 


mariari:
**Describe the bug**

After Nyxt crashes  and you restart it. You may get a message like

```
<WARN> [13:13:02] Warning: Failed to restore history from /home/katya/.local/share/nyxt/history/default.lisp: Owner with identifier "11" does not exist
```

When you try to change buffers to any previous one you get something akin to

```
<WARN> [13:13:45] Warning: Error on separate thread: Owner with identifier "16" does not exist
```

The only solution is to kill your history. Any attempt to delete the buffer with `delete-buffers` Will result in this

```lisp
Unhandled SIMPLE-ERROR in thread #<SB-THREAD:THREAD "Prompter follow selection thread" RUNNING
                                    {1008721133}>:
  Owner with identifier "16" does not exist

Backtrace for: #<SB-THREAD:THREAD "Prompter follow selection thread" RUNNING {1008721133}>
0: (SB-DEBUG::DEBUGGER-DISABLED-HOOK #<SIMPLE-ERROR "Owner with identifier ~s does not exist" {10087B2743}> #<unused argument> :QUIT T)
1: (SB-DEBUG::RUN-HOOK SB-EXT:*INVOKE-DEBUGGER-HOOK* #<SIMPLE-ERROR "Owner with identifier ~s does not exist" {10087B2743}>)
2: (INVOKE-DEBUGGER #<SIMPLE-ERROR "Owner with identifier ~s does not exist" {10087B2743}>)
3: (ERROR "Owner with identifier ~s does not exist" "16")
4: (HISTORY-TREE:SET-CURRENT-OWNER #<HISTORY-TREE:HISTORY-TREE {10065EDF53}> "16")
5: ((FLET SB-THREAD::WITH-RECURSIVE-LOCK-THUNK :IN WINDOW-SET-BUFFER))
6: ((FLET "WITHOUT-INTERRUPTS-BODY-11" :IN SB-THREAD::CALL-WITH-RECURSIVE-LOCK))
[...]

Issue #1550, opened 2021-06-25
"Owner with identifier x does not exist" - constant crashing


ncryptid:
**Describe the bug**

Nyxt crashes on loading a session while trying to switch to one of my open buffers and will give an error like this (the number varies):

```
<WARN> [22:32:03] Warning: Failed to restore history from /home/nyx/.local/share/nyxt/history/default.lisp: Owner with identifier "8" does not exist
Unhandled SIMPLE-ERROR in thread #<SB-THREAD:THREAD "Prompter follow selection thread" RUNNING
                                    {10070E8D23}>:
  Owner with identifier "8" does not exist
```

If I manually restore the session by running `restore-history-by-name`, I can continue to browse and switch buffers etc., but it seems noticeably unstable and will crash frequently and seemingly randomly while trying to load a new URL. The site in question doesn't seem to affect it, but it returns an error like this, which seems to indicate that it's related to this problem:

```
<WARN> [22:32:59] Warning: Error on GTK thread: There is no applicable method for the generic function
                       #<STANDARD-GENERIC-FUNCTION HISTORY-TREE:DATA (2)>
                     when called with arguments
                       (NIL).
```

This may have been related to an issue I had where Nyxt was trying to load a URL and taking a long time to do so (I think it was Reddit), and then became unresponsive to the point where I could open the prompt buffer but it wouldn't let me run any comands and would stay on the same page it was trying to load. I had to kill the process manually but that's when I started to notice this issue. 

I could probably just delete my history file and start fresh, but I would prefer not to lose my entire browsing history.

**Precise recipe to reproduce the issue**

1. Go to ~some site that Nyxt can't load
2. Kill Nyxt while it's hung up and can't run any commands
3. Restore the session, receiving an error about an owner with identifier x not existing
4. Attempt to switch buffers by scrolling through the open ones
5. crash

**Information**
- OS name+version: Devuan Beowulf
- Graphics card and driver: whatever integrated Intel GPU a Thinkpad X230 has
- Desktop environment / Window manager name+version: StumpWM 20.11
- How you installed Nyxt (Guix pack, package manager, build from source): build from source
- Information from command copy-system-information:

```
Nyxt version: 2.1.0
Renderer version: GI-GTK
Operating system kernel: Linux 4.19.0-16-amd64
Lisp implementation: SBCL 2.1.4
[...]

PR #1564, opened 2021-06-28
history: Extract buffer restoration out of `restore' to avoid side ef…

Ambrevar:
…fects and corruption.

`restore' should not have side effect.  Otherwise it may accidentally trigger
nested `restore' calls (via get-data) which may corrupt everything.  (Like
having 2 histories!)

The buffer restoration should thus be done outside of the context of data-path
restoration.


Ambrevar:
Shoud fix #1560 and #1550.

jmercouris:
I'm running an older commit, and can't really test. I've also never had history problems. Perhaps it is something about my setup.

Ambrevar:
It's a race condition, so it depend on how many buffers you restore and
how fast your hardware is.

Also the history-related issues usually arise when

```
(define-configuration browser
  ((session-restore-prompt :always-restore)))
```

Anyways, can you have a look at the code and test it on you machine, see
if it loads the history only once (look at the shell output) and if it
works probably on restart?

jmercouris:
I will test on my machine. I guess the big difference here is that I have `:never-restore` in my config, which is probably why I am not having these errors. I'll test with no config then.

jmercouris:
I'll also take a look at the code, of course.

Ambrevar:
> I guess the big difference here is that I have `:never-restore` in my
> config, which is probably why I am not having these errors.

Totally, the issue at hand only arises when session restoration! :p

jmercouris:
I've tried to force kill the program at critical moments to corrupt the history with a ton of buffers, still appears to work.

Ambrevar:
Thanks!

Issue #967, opened 2020-09-25
Cross-references in docstrings / *describe* buffers


Ambrevar:
It would be nice to turn docstring cross references to `lisp://` URLs.
For instance, the docstring of `buffer-load` mentions `buffer-load-hook`: we would turn it into a clickable link which would open the appropriate description page.

This was discussed in #745.

It shouldn't be too hard: parse "`[^[:space:]]*'" or something like that and replace it with `lisp://(describe MATCH)`.
`describe` would be a smart function which finds out which description function to call depending on the type.
If a symbol has multiple types (e.g. both a function and a variable), then we would query the user.

Thoughts?

jmercouris:
I feel like 1984 editing your messages :-D

jmercouris:
I really like this idea. I assume the idea is kind of like how when clicking M-. on a object if Lisp finds multiple things like several methods with the name, you select the one that you are interested in?

Ambrevar:
Exactly!

aadcg:
> If a symbol has multiple types (e.g. both a function and a variable), then we would query the user.

I like the idea but there should be a smarter way to deal with the above.

jmercouris:
There is no smarter way. A symbol in a Lisp-2 is ambiguous and has to be resolved. An alternative may be displaying both a listing for the symbol as a function as and as a variable.

For example:

`xyz (variable)`
`xyz (function)`

Ambrevar:
This is the different topic of _symbol namespaces_ (whether we want to
resolve a symbol to a function, a class, a variable, etc.).

What the "target" slot solves is the problem of how to refer to _values_
(unprintable objects).
Note that the problem of symbol namespaces does not occur in the context
of this patch.

Ambrevar:
About symbol namespaces: your suggestion could work.
In case of ambiguity (e.g. because the parentheses were forgotten), we
can prompt the user for the namespace.

Ambrevar:
Would be nice to have bi-directional linking, i.e. list all symbols where the current symbol help page is referenced.

Ambrevar:
Partially done in https://github.com/atlas-engineer/nyxt/commit/ed97f71e66c7a4b3eea481d3bb7e5951fb88df76.

But I'd like to address a few points:

- The regex could be "`[^'\\s]+'" to include any kind of whitespace.
- Multiple definitions are not supported.  Try with `nyxt:name`.

For this last point, I suggest the following:
[...]

PR #1566, opened 2021-06-29
tutorial.lisp: insert an explanation about application-mode on built-in tutorial

### LINKED ISSUE #1516: Documentation: manual should mention `application-mode`
I am working on it

pdelfino:
Hopefully, this fixes #1516.

Inside emacs, before commiting, I used `fill-paragraphs` and `white-space-mode`.  Depending on how `fill-paragraph` goes, it breaks the compilation because some line breaks are set in "empty" places (between strings and `(:code ....)`). I am not sure if `fill-paragraph` is the best way to keep it short. I am open to advice :wink:

I hope there will be no problem concerning whitespaces or the length of the content.

### DIFF
--- a/source/tutorial.lisp
+++ b/source/tutorial.lisp
@@ -176,6 +176,14 @@ removed from the input, they are also removed from the existing bookmark.")
     (:li (command-markup 'delete-bookmark) ": Delete queried bookmarks.")
     (:li (command-markup 'list-bookmarks) ": Display a new buffer containing the
 list of all bookmarks."))
+   (:h3 "Application mode")
+   (:p "The command " (:code "application-mode") " forwards all keys to the
+renderer. For instance, using the default binding of Nyxt (" (:code "web-cua-map") ") the
+key-binding " (:code "C-i") " executes " (:code "autofill") ". Suppose
+a user is using their email client which also uses " (:code "C-i") " for the italic command. Thus, after
+executing " (:code "application-mode") " the " (:code "C-i") " binding is associated
+with the webpage's italic command instead of " (:code "autofill") ". Finally, the
+user can return to their configuration just by executing " (:code "application-mode") " again.")
    (:h3 "Enable and Disable modes")
    (:p "The command " (:code "enable-mode") " allows the user to apply a
 mode (such as " (:code "nosound-mode") " or " (:code "dark-mode") ") to multiple


edgar-vincent:
That's a very good idea!
I think the tone might need to be a little more formal or abstract to blend in better with the rest of the manual. Also, I think the possessive adjective "their" should be used instead of "his".

pdelfino:
>   "their" should be used instead of "his".

Fixed. I am not a native speaker, so, thanks a lot! :uk:

>  think the tone might need to be a little more formal or abstract to blend in better with the rest of the manual

**On being more formal:**
Could you give me another English hint? What exactly sounds informal to you?

**On being abstract:**
I noticed that too. But I was not able to come up with something better. If we use `descibe-command` on `application-mode` we get: _Mode that forwards all keys to the renderer._

I even used that as part of my explanation. It is abstract. But I am not sure if this is clear to a lot of people . It was not clear for me the power of `application-mode` when I first read the definition. It was clear after seeing an example.

Having a look at the Common Lisp official hyper spec [documentation](http://clhs.lisp.se/Body/f_find_.htm), they provide abstract definitions _and_ examples. To be honest here, the examples are the first thing I usually read :laughing:

Maybe we could consider inserting more examples on the `tutorial`! Instead of removing this one to keep it consistent.

jmercouris:
Hi Pedro, I think it is more or less ready, @edgar-vincent has left the same comments I would :-). I also noticed though that there is still yet an instance of "his" in the string. Perhaps you forgot to push?

pdelfino:
>  also noticed though that there is still yet an instance of "his" in the string. Perhaps you forgot to push?

Oh, no. I actually forgot about it. Using `their` with the *singular* is counter-intuitive for my Brazilian brain haha. I missed the second one. My bad! It was good to learn the grammar rule abstracting the gender, thanks.

Ambrevar:
See https://en.wikipedia.org/wiki/Singular_they :)

source/tutorial.lisp
```diff
@@ -176,6 +176,16 @@ removed from the input, they are also removed from the existing bookmark.")
     (:li (command-markup 'delete-bookmark) ": Delete queried bookmarks.")
     (:li (command-markup 'list-bookmarks) ": Display a new buffer containing the
 list of all bookmarks."))
+   (:h3 "Application mode")
+   (:p "The command " (:code "application-mode") " forwards all keys to the
+renderer. For instance, using the default configuration of Nyxt (" (:code "web-cua-map") ") the
+key-binding " (:code "C-i") " executes a command called "(:code "autofill") ". Suppose
```
Ambrevar, line 182: Missing space before `(:code`

Ambrevar:
This pull request removes the cl-webkit submodule.  Can you leave this out?

The addition is welcome, however it feels longish for a simple mode.
You can shorten it.  Some tips:

- `Coincidentally, ` -> `where`
- `will be ` -> `is`
- `and not " (:code "autofill") ` -> `instead`.
- `After successfully using the keyboard shortcut for italic provided by this specific website,` -> remove since it does not convey anything new.

source/tutorial.lisp
pdelfino: I have implemented all of the suggestions. Thanks. 
 
 On this one, I would like to ask a question:

>  "After successfully using the keyboard shortcut for italic provided by this specific website," -> remove since it does not convey anything new.

I have already removed it but the I think the **tutorial** does not mention at any point the **fact** that to revert a mode you can execute the mode again. You can also use `disable-mode` (I did a PR to include that recently).

What do you think about adding a phrase on the **Modes** section of the tutorial to inform this?

Just to illustrate:
**Modes**
[...] _Each mode has an associated mode toggler which is a command of the same name that toggles the mode for the current buffer._ Finally, modes are enabled after they are executed and disabled if executed again. It is also possible to enable and disable modes in one or more buffers using `enable-mode` and `disable-mode` (see below).


Btw, it would be nice to have cross-references using hyperlinks on the tutorial instead of "see below". Have you guys ever thought about it?

pdelfino:
I have implemented all of the suggestions. Thanks. 
 
 On this one, I would like to ask a question:

>  "After successfully using the keyboard shortcut for italic provided by this specific website," -> remove since it does not convey anything new.

I have already removed it but the I think the **tutorial** does not mention at any point the **fact** that to revert a mode you can execute the mode again. You can also use `disable-mode` (I did a PR to include that recently).

What do you think about adding a phrase on the **Modes** section of the tutorial to inform this?

Just to illustrate:
**Modes**
[...] _Each mode has an associated mode toggler which is a command of the same name that toggles the mode for the current buffer._ Finally, modes are enabled after they are executed and disabled if executed again. It is also possible to enable and disable modes in one or more buffers using `enable-mode` and `disable-mode` (see below).


Btw, it would be nice to have cross-references using hyperlinks on the tutorial instead of "see below". Have you guys ever thought about it?

jmercouris:
Yes we've thought about it, but have not yet implemented. We would also like links for internal buffers and pagination :-D

Ambrevar:
The issue about cross references: https://github.com/atlas-engineer/nyxt/issues/967

source/tutorial.lisp
```diff
@@ -176,6 +176,15 @@ removed from the input, they are also removed from the existing bookmark.")
     (:li (command-markup 'delete-bookmark) ": Delete queried bookmarks.")
     (:li (command-markup 'list-bookmarks) ": Display a new buffer containing the
 list of all bookmarks."))
+   (:h3 "Application mode")
+   (:p "The command " (:code "application-mode") " forwards all keys to the
+renderer. For instance, using the default configuration of Nyxt (" (:code "web-cua-map") ") the
```
Ambrevar, line 181: web-cua-map is not a configuration, you mean the default binding.  This sentence can be shortened then.

source/tutorial.lisp
```diff
@@ -176,6 +176,15 @@ removed from the input, they are also removed from the existing bookmark.")
...
+key-binding " (:code "C-i") " executes a command called " (:code "autofill") ". Suppose
```
Ambrevar, line 182: Remove `a command called`.

source/tutorial.lisp
```diff
@@ -176,6 +176,15 @@ removed from the input, they are also removed from the existing bookmark.")
...
+a user is using their email client which has a keyboard shortcut for
```
Ambrevar, line 183: Shorter: "which also uses `C-i' for the italic command."

PR #1572, opened 2021-06-30
command: Forbid required parameters in commands.

Ambrevar:
They don't make sense since commands are meant to be run interactively.

### DIFF
--- a/source/command.lisp
+++ b/source/command.lisp
@@ -136,6 +136,9 @@ Example:
     (unless documentation
       (warn (make-condition 'command-documentation-style-warning
                             :name name)))
+    (alex:when-let ((req-args (alex:parse-ordinary-lambda-list arglist)))
+      (error "Commands cannot have required parameters (~s), consider optional or keyword parameters instead."
+             req-args))
     (let ((before-hook (intern (str:concat (symbol-name name) "-BEFORE-HOOK")
                                (symbol-package name)))
           (after-hook (intern (str:concat (symbol-name name) "-AFTER-HOOK")


Ambrevar:
It can be backward-incompatible if a user used `define-command` instead of `defun`, but then the error will be justified because that's not how `define-command` should be used.

jmercouris:
I disagree, what if you want a command to HAVE to accept some user input via `execute-extended-command`? In fact, you could argue that many commands that accept input via the prompt buffer could have been programmed in that way, no?

Ambrevar:
This is already what we support, isn't it?

A command that requires an argument simply prompts for it, doesn't it?

jmercouris:
No, a command that requires an argument does not prompt for it.

jmercouris:
If you have a command like

```
(define-command xyz (qre)
   (echo qre))
```

and you bind it to C-x C-y

it will not prompt you for `qre`.

jmercouris:
Now, if you did `execute-extended-command` and try to run `xyz`, then it /would/ prompt you.

Ambrevar:
> If you have a command like
>
> ```
> (define-command xyz (qre)
>    (echo qre))
> ```
>
> and you bind it to C-x C-y
>
> it will not prompt you for `qre`.

I'm confused, because it would fail.  Presently you can't call a command with
required arguments.

jmercouris:
Exactly, you should be able to do that.

jmercouris:
Closing since we can now automatically use required parameters in commands.

Issue #1586, opened 2021-07-02
Shoudn't execute-extended-command run command asynchronously?


Ambrevar:
I noticed that `execute-extended-command` runs commands directly instead of leveraing `run-async` like `execute-command` does.

Is this intentional?  A mistake?
@jmercouris ?

jmercouris:
I can't remember. Perhaps a mistake. We would need to test to be sure.

Ambrevar:
I'm on it.

I think know what it was not run asynchronous: there was a bug in
run-async, we didn't rebind the arguments, but since it's a separate
thread, the binding would run stale which would cause the function to be
called over a random memory location.

I have a fix, will send soon.

Ambrevar:
Closing since the fix is in #1587.

PR #1582, opened 2021-07-01
command: Hide deprecated commands (don't add them to *command-list*).

Ambrevar:
While we are at it, factor out some code, it's even shorter now.

PR #1587, opened 2021-07-02
Automatic required arguments for commands

Ambrevar:
This is an alternative to https://github.com/atlas-engineer/nyxt/pull/1582 which wanted to forbid required arguments in commands.

As @jmercouris suggested in #1582, we could instead prompt the user to fill in the required argument when the command is called.  This would make command writing must faster when the input is trivial (raw-sources).
We could even infer the prompt buffer source from the argument type.

To this end, I started overhauling `execute-extended-command`.  It now prefills the values and does type-checking!
It also fixes  https://github.com/atlas-engineer/nyxt/issues/1586.

To do:
- Once `execute-extended-command` is approved, implement the above idea for general command calls!

### DIFF
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -74,13 +74,11 @@ keyword parameters."
                            :prompt "Execute extended command"
                            :sources (make-instance 'user-command-source)
                            :hide-suggestion-count-p t)))
-          (argument-list (swank::arglist (fn command)))
-          (required-arguments (nth-value 0 (alex:parse-ordinary-lambda-list
-                                            argument-list)))
-          (optional-arguments (nth-value 1 (alex:parse-ordinary-lambda-list
-                                            argument-list)))
-          (key-arguments (nth-value 3 (alex:parse-ordinary-lambda-list
-                                       argument-list))))
+          (argument-list (swank::arglist (fn command))))
+     (multiple-value-bind (required-arguments optional-arguments rest key-arguments
+                           aok? aux key?)
+         (alex:parse-ordinary-lambda-list argument-list)
+       (declare (ignore rest aok? aux key?)))
      (apply command
             (append
              (when required-arguments
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -69,7 +69,6 @@
 (define-command execute-extended-command (&optional command)
    "Execute a command by name, also supply required, optional, and
 keyword parameters."
-  ;; TODO: prefill default-values when prompting optional/key arguments
   ;; TODO: Display type and catch type errors.
   (let* ((command (or command
                       (first (prompt
@@ -82,27 +81,30 @@ keyword parameters."
          (alex:parse-ordinary-lambda-list lambda-list)
        (declare (ignore rest aok? aux key?))
        (setf (last-access command) (local-time:now))
-       (labels ((prompt-arg (prompt)
+       (labels ((prompt-arg (prompt &optional input)
                   (read-from-string
                    (first (prompt
                            :prompt prompt
+                           :input (write-to-string input)
                            :sources (make-instance 'prompter:raw-source)))
                    nil nil))
-                (prompt-args (params)
-                  (mapcar (lambda (param)
-                            (prompt-arg param))
-                          params))
-                (prompt-kwargs (params)
-                  (alex:mappend (lambda-match
-                                  ((cons key _)
-                                   (copy-list (list (first key)
-                                                    (prompt-arg (second key))))))
-                                params)))
+                (parse-args (params)
+                  (alex:mappend
+                   (lambda-match
+                     ((and param (type symbol))
+                      (list (prompt-arg param)))
+                     ((list (list keyword name) default _)
+                      (list keyword
+                            (prompt-arg name default)))
+                     ((list name default _)
+                      (list (prompt-arg name default))))
+                   params)))
          (apply #'run-async
                 command
-                (append (alex:mappend #'prompt-args
-                                      (list required-arguments optional-arguments))
-                        (prompt-kwargs keyword-arguments)))))))
+                (alex:mappend #'parse-args
+                              (list required-arguments
+                                    optional-arguments
+                                    keyword-arguments)))))))
 
 (defun get-hooks ()
   (flet ((list-hooks (object)
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -66,10 +66,29 @@
       (setf (last-access command) (local-time:now))
       (run-async command))))
 
+(defun parse-function-lambda-list-types (fn)
+  #-sbcl
+  (declare (ignore fn))
+  #-sbcl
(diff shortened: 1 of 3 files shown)


Ambrevar:
Done!

You can test with the following dummy function:

```lisp
(declaim (ftype (function (number)) adder))
(define-command adder (number)
  "Switch the active buffer in the current window."
  (echo "(1+ ~s) = ~s" number (1+ number)))
```

Call the command interactively, it will prompt for a number.

source/command-commands.lisp
```diff
@@ -66,43 +66,79 @@
...
-                                      :sources (make-instance 'prompter:raw-source))))))
-             (when key-arguments
-               (loop for argument in key-arguments
-                     collect (first (car argument))
-                     collect (read-from-string
-                              (first (prompt
-                                      :prompt (second (car argument))
-                                      :sources (make-instance 'prompter:raw-source))))))))
-     (setf (last-access command) (local-time:now))))
+(defun parse-function-lambda-list-types (fn)
```
jmercouris, line 69: No way to do this cross-implementation? What do you think of the idea of extending our `define-command` to allow storing type information? Then we could store that information within the command object.

source/command-commands.lisp
```diff
@@ -66,43 +66,79 @@
...
+  "Query the user for the arguments to pass to a given COMMAND.
+User input is evaluated Lisp."
+  ;; TODO: Add support for &rest arguments.
+  (let* ((command (or command
+                      (first (prompt
+                              :prompt "Execute extended command"
+                              :sources (make-instance 'user-command-source)
+                              :hide-suggestion-count-p t))))
+         (lambda-list (swank::arglist (fn command))))
+    (multiple-value-bind (required-arguments optional-arguments rest keyword-arguments
```
jmercouris, line 117: why this multiple-value-bind for things that you are not using? why not utilize the approach in the previous version of this code? 

```
(required-arguments (nth-value 0 (alex:parse-ordinary-lambda-list
                                             argument-list)))
```

source/command-commands.lisp
```diff
@@ -66,43 +66,79 @@
...
+                                               types))
+         (keyword-type-pairs (pairlis keywords (rest type-batches))))
+    (alex:nreversef keywords)
+    (values (first type-batches)
+            (alex:assoc-value keyword-type-pairs '&optional)
+            (alex:assoc-value keyword-type-pairs '&rest)
+            (alex:assoc-value keyword-type-pairs '&key))))
+
+(defun prompt-argument (prompt &optional type input)
+  (let ((value
```
jmercouris, line 89: I would suggest `result` in lieu of `value`.

jmercouris: or perhaps `input`.

source/command-commands.lisp
jmercouris: of course you could optimize things and not reparse the lambda list for every single value you want to extract, but the same approach. I would rather use `nth-value` and grab exactly what I need than having `(declare (ignore ...))`

source/command-commands.lisp
Ambrevar: Could be an excellent idea!
Not sure how to do it in a backward-compatible way though.

source/command.lisp
```diff
@@ -347,11 +352,15 @@ This is blocking, see `run-async' for an asynchronous way to run commands."
 See `run' for a way to run commands in a synchronous fashion and return the
 result."
   (run-thread
-    (with-current-buffer (current-buffer) ; See `run' for why we bind current buffer.
-      (handler-case (apply #'funcall command args)
-        (nyxt-prompt-buffer-canceled ()
-          (log:debug "Prompt buffer interrupted")
-          nil)))))
+    ;; It's important to rebind `args' since it may otherwise be shared with the
```
jmercouris, line 355: I'm not sure I understand, can you please expand? Are you saying the caller may manipulate args leading to unintended behavior?

source/command-commands.lisp
jmercouris: Perhaps it is overkill for now, but just something to make a note of.

source/command-commands.lisp
```diff
@@ -66,43 +66,79 @@
...
+                              :hide-suggestion-count-p t))))
+         (lambda-list (swank::arglist (fn command))))
+    (multiple-value-bind (required-arguments optional-arguments rest keyword-arguments
+                          aok? aux key?)
+        (alex:parse-ordinary-lambda-list lambda-list)
+      (declare (ignore rest aok? aux key?))
+      (multiple-value-bind (required-types optional-types rest-types keyword-types)
+          (parse-function-lambda-list-types (fn command))
+        (declare (ignore rest-types))
+        (flet ((parse-args (params)
```
jmercouris, line 124: Do you have any ideas of a simpler way to express `parse-args`?

source/command-commands.lisp
Ambrevar: I believe what you are suggesting does not work without recalling `parse-ordinary-lambda-list` multiple times :)

jmercouris:
Looks great, I like it! :-)

source/command-commands.lisp
Ambrevar: But it really _is_ the argument value, isn't it?

source/command.lisp
Ambrevar: Read `bt:make-threads` docstring, it will clarify a few things.
So yes, the `args` binding coming from the caller is shared between threads, and then the caller exits, which leaves the callee with `args` in undefined state, leading to the error.

source/command-commands.lisp
jmercouris: I've not tried it, perhaps you are right.

source/command-commands.lisp
jmercouris: Sure, but to anyone reading it, it /is/ the user input to the argument value.

source/command.lisp
jmercouris: OK, I understood the problem. Just wanted to verify.

source/command-commands.lisp
Ambrevar: Funny you asked, I've spent some time simplifying `parse-args` and I'm happy with this clear and concise result (in my opinion).  Do you have something else in mind?

source/command-commands.lisp
jmercouris: OK!

source/command-commands.lisp
jmercouris: I don't have something else, but this is certainly difficult to understand. Took me some minutes!

Ambrevar: I suppose it's a matter of habbit.  When you get used to lambda-match, it starts shining with clarity in my experience.

source/command-commands.lisp
Ambrevar: There is actually a better way: with `multiple-value-match` we can bind ignore arguments to `_`.

PR #1591, opened 2021-07-05
Original desire of issue 1310: As a page is loaded, cheaply check it is frequently visited and not already bookmarked

pdelfino:
Hi guys,

This an implementation for the **original** desire on issue #1310.

I have tried doing the prescribed "algorithm":

```text
Query the implicit-visit of the current URL (zero-cost)
Compare with the threshold.
If above threshold and not already bookmarked (single hash-table lookup), bookmark.
```

In addition to the prescription, I inserted a prompt asking the user to confirm the bookmarking. 

Currently, the implementation works but as an asynchronous process to be run on REPL.

I need to learn how to insert this function to run every time a webpage is loaded on Nyxt. 
It would be nice to receive hints on how to do it :)

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -106,6 +106,7 @@
                #+quicklisp
                (:file "lisp-system")
                ;; Core Modes
+               (:file "bookmark-frequent-visits-mode")
                (:file "editor-mode")
                (:file "plaintext-editor-mode")
                (:file "buffer-listing-mode")
--- a/source/bookmark-frequent-visits-mode.lisp
+++ b/source/bookmark-frequent-visits-mode.lisp
@@ -1,0 +1,51 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/bookmark-frequent-visits-mode
+    (:use :common-lisp :trivia :nyxt :alexandria)
+  (:documentation "Mode to bookmark frequently visited URLs."))
+(in-package :nyxt/bookmark-frequent-visits-mode)
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (trivial-package-local-nicknames:add-package-local-nickname :alex :alexandria)
+  (trivial-package-local-nicknames:add-package-local-nickname :sera :serapeum))
+
+(define-mode bookmark-frequent-visits-mode ()
+  "Mode to bookmark frequently visited URLs while navigating the web."
+  ((threshold 20)
+   (constructor
+    (lambda (mode)
+      (nyxt:on-signal-load-finished mode (url (current-buffer)))))))
+
+(defun bookmark-frequent-visit (threshold)
+  "Check if current URL is frequently visited and not included in the
+bookmarks. If this is the case, prompt the user about bookmarking it."
+  (labels ((bookmarked-url-p (url-address)
+             "The local function `bookmarked-url-p' returns the URL
+           address itself if it is new to the bookmark list or NIL if it is
+           already there."
+             (let ((bookmarks-address-list
+                     (mapcar #'(lambda (e) (render-url (url e)))
+                             (with-data-unsafe (bookmarks (bookmarks-path (current-buffer)))
+                               bookmarks))))
+               (if (member url-address bookmarks-address-list :test #'string=)
+                   nil
+                   url-address))))
+    (sera:and-let* ((history-entries (with-data-unsafe (history (history-path (current-buffer)))
+                                        (mapcar #'htree:data (alex:hash-table-keys (htree:entries history)))))
+                    (current-url-history
+                      (find (url (current-buffer)) history-entries :test #'equalp :key #'url))
+                    (implicit-visits-value
+                      (nyxt::implicit-visits current-url-history))
+                    (current-url-address
+                      (render-url (url current-url-history)))
+                    (threshold threshold))
+                   (run-thread
+                    (if (and (> implicit-visits-value threshold)
+                           (bookmarked-url-p current-url-address))
+                      (if-confirm ("Bookmark ~a?" current-url-address)
+                                  (bookmark-url :url current-url-address)))))))
+
+(defmethod nyxt:on-signal-load-finished ((mode bookmark-frequent-visits-mode) url)
+  (bookmark-frequent-visit (threshold mode))
+  url)


pdelfino:
Hi @Ambrevar,

I talked about this with @jmercouris in a live meeting:

> Currently, the implementation works but as an asynchronous process to be run on REPL.
> I need to learn how to insert this function to run every time a webpage is loaded on Nyxt.

John suggested (i) to transform this into a mode and (ii) to remove the slot value addition. The approach would be something similar to `style-mode`.

It would be nice to listen to your opinion about this before I make more advances implementing this, especially because this implementation is supposed to solve your idea following the recipe you coined:

```
Query the implicit-visit of the current URL (zero-cost)
Compare with the threshold.
If above threshold and not already bookmarked (single hash-table lookup), bookmark.
```

pdelfino:
The previous message was supposed to be sent yesterday night. For some reason, it was not sent.

jmercouris:
It's not similar to style mode, but it uses the 

```
(defmethod nyxt:on-signal-load-finished ((mode style-mode) url)
  (declare (ignore url))
 ...)
```

like style mode does :-)

pdelfino:
> like style mode does :-)

Ok. Thanks for correcting me, John. Besides being a more precise message for Pierre, it will help me.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,53 @@
...
+               (bookmarked-url-p current-url-address))
+          (if-confirm ("Bookmark ~a?" current-url-address)
+                      (bookmark-url :url current-url-address))))))
+
+(defun print-on-nyxt ()
+  (echo "HELLO WORLD"))
+
+(defmethod nyxt:on-signal-load-finished ((mode bookmark-frequent-visits-mode) url)
+  (print-on-nyxt)
+  ;(bookmark-frequent-visit)
```
pdelfino, line 52: This line does not work. It has to be something related to the packages hierarchy. Could you guys give me a hint on how to make it work?

pdelfino:
Hi guys,

I am close but stuck. The mode created works in some ways:

1 -  I can see it on GUI as a command and after executing, I can see it on the status bar. Executing it again works as expected disabling it.

2 - The mode works with the toy/pet function `print-on-nyxt` synchronously to page loading. Thus, everytime a page is loaded the `echo` works:

```lisp
NYXT/WEB-MODE> (buffer-load "https://ambrevar.xyz")
NIL
T
<INFO> [00:38:09] Loading "https://ambrevar.xyz/".
<INFO> [00:38:11] HELLO WORLD
<INFO> [00:38:11] Finished loading "https://ambrevar.xyz/".
```
3 - The real deal of the feature is the function  `bookmark-frequent-visit`, though. I can run the function being inside the package:
```lisp
NYXT/BOOKMARK-FREQUENT-VISITS-MODE> (bookmark-frequent-visit)
NIL
```
It works as expected.

However, I **cannot** insert this function on [this](https://github.com/atlas-engineer/nyxt/pull/1591/files#r665848097) line. Hence, it does not work every time a webpage is loaded. I need to fix this.

Doubts:

1 -  When I load the `nyxt/gi-gtk` system, the mode I created is not compiled. I need to manually compile it with Slime (`C-c`) to see things. I thought it would compile all files in the source code. Why does this happen? How can I fix this?  

Since creating modes is something (really) new for me, this is probably my fault not a bug.

2 - For some reason, just after loading the system, I cannot move between packages from toplevel  `CL-USER` directly to the package `nyxt/bookmark-frequent-visits-mode`.
I only managed to successfully achieve `(in-package :nyxt/bookmark-frequent-visits` being inside `nyxt/web-mode`. It did not work being inside `nyxt` also. Why?

3 - Apparently, `(nyxt::implicit-visits` does not work inside `(defmethod nyxt:on-signal-load-finished` but it does work inside the same file and package when I call it. The relationship between package hierarchy, methods, and functions definitions confuse me.

What do I need to change to use appropriately invoke `bookmark-frequent-visit` function **_inside_** the `defmethod` expression?


Thanks.

Ambrevar:
It works for me.

You need to add the mode to your .asd.

Then note that current-url-history may be nil.
Try replacing the `let*` with a `serapeum:and-let*`, it should avoid
executing the body if any of the values is nil.

pdelfino:
> You need to add the mode to your .asd. Then note that current-url-history may be nil. Try replacing the `let*` with a `serapeum:and-let*`, it should avoid executing the body if any of the values is nil.

:heavy_check_mark: Done.

> It works for me. 

For me, Nyxt crashes when I reach the threshold for frequent visits in a specific webpage. 
Did you try to re-load a light web page ([for instance](https://fr.wikipedia.org/wiki/Br%C3%A9sil)) more than 20 times (this web page should not be already in your bookmarks)?

If I try this, Nyxt crashes for me.

nyxt.asd
```diff
@@ -106,6 +106,7 @@
                #+quicklisp
                (:file "lisp-system")
                ;; Core Modes
+               (:file "bookmark-frequent-visits-mode")
```
pdelfino, line 109: @jmercouris , would you consider this a **core mode**? Should I insert it somewhere else?

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,55 @@
...
+               (bookmarked-url-p current-url-address))
+          (if-confirm ("Bookmark ~a?" current-url-address)
+                      (bookmark-url :url current-url-address))))))
+
+(defun print-on-nyxt ()
+  (echo "HELLO, WORLD!!"))
+
+(defmethod nyxt:on-signal-load-finished ((mode bookmark-frequent-visits-mode) url)
+  (print-on-nyxt)
+  (bookmark-frequent-visit)
```
pdelfino, line 54: This line makes Nyxt crashes. Probably, the reason is something related to `(nyxt::implicit-visits`

Ambrevar: "Crash" is too vague.  When you run from the REPL, you should get a stacktrace.  Can you share this?

nyxt.asd
jmercouris: I would not consider it a core mode, but since it would be part of Nyxt, it is in fact, a core mode :-)

source/bookmark-frequent-visits-mode.lisp
pdelfino: crashes-nyxt

source/bookmark-frequent-visits-mode.lisp
pdelfino: Just in case someone trying to contribute to the project read this, there was a deadlock problem happening involving `sera:and-let*` and `nyxt:on-signal-load-finished`.

Thanks to @jmercouris help, the problem was solved using `bordeaux-threads` with `bt:make-thread`.

pdelfino:
Hi guys,

I think this is good to go! 

Thanks.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,53 @@
...
+             (with-data-unsafe (history (history-path (current-buffer)))
+               (mapcar #'htree:data (alex:hash-table-keys (htree:entries history)))))
+            (current-url-history
+          (find (url (current-buffer)) history-entries :test #'equalp :key #'url))
+          (implicit-visits-value
+           (nyxt::implicit-visits current-url-history))
+          (current-url-address
+           (render-url (url current-url-history)))
+         (threshold threshold))
+      (bt:make-thread (lambda ()
```
Ambrevar, line 45: Never use `make-thread` unless you know what you are doing, prefer `run-thread` instead which protects Nyxt against crashes.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,53 @@
...
+           address itself if it is new to the bookmark list or NIL if it is
+           already there."
+             (let ((bookmarks-address-list
+                    (mapcar #'(lambda (e) (render-url (url e)))
+                              (with-data-unsafe (bookmarks (bookmarks-path (current-buffer)))
+                                 bookmarks))))
+               (if (member url-address bookmarks-address-list :test #'string=)
+                   nil
+                   url-address))))
+    (sera:and-let*
```
Ambrevar, line 34: Indentation issue below.

Ambrevar:
Isn't the core function redundant with that of #1649?
I thought we would implement and merge #1649 first, then base this implementation on top of it.  This way, this mode would be completely trivial.

Git: This change probably needs only one commit, you can squash them (unless you believe some of them really belongs to `master` history).

source/bookmark-frequent-visits-mode.lisp
jmercouris: Ah.. that was my mistake :-D

pdelfino:
> Git: This change probably needs only one commit, you can squash them (unless you believe some of them really belongs to `master` history).

Ok. Everything was squashed :tennis:

> Isn't the core function redundant with that of #1649?
I thought we would implement and merge #1649 first, then base this implementation on top of it. This way, this mode would be completely trivial.

Oh, sorry. I did not realize this order should have been followed.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,51 @@
...
+  ((threshold 20)
+   (constructor
+    (lambda (mode)
+      (nyxt:on-signal-load-finished mode (url (current-buffer)))))))
+
+(defun bookmark-frequent-visit (threshold)
+  "Check if current URL is frequently visited and not included in the
+bookmarks. If this is the case, prompt the user about bookmarking it."
+  (labels ((bookmarked-url-p (url-address)
+             "The local function bookmarked-url-p returns the URL
```
pdelfino, line 24: I have just realized the docstring trick: bookmarked-url-p ->  `bookmarked-url-p'. Fixing it, commit amend as --no-edit, and pushing it now.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,51 @@
...
+bookmarks. If this is the case, prompt the user about bookmarking it."
+  (labels ((bookmarked-url-p (url-address)
+             "The local function `bookmarked-url-p' returns the URL
+           address itself if it is new to the bookmark list or NIL if it is
+           already there."
+             (let ((bookmarks-address-list
+                     (mapcar #'(lambda (e) (render-url (url e)))
+                             (with-data-unsafe (bookmarks (bookmarks-path (current-buffer)))
+                               bookmarks))))
+               (if (member url-address bookmarks-address-list :test #'string=)
```
Ambrevar, line 31: The `if` is not necessary here.  Just return `(member ...)`.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,51 @@
...
+    (lambda (mode)
+      (nyxt:on-signal-load-finished mode (url (current-buffer)))))))
+
+(defun bookmark-frequent-visit (threshold)
+  "Check if current URL is frequently visited and not included in the
+bookmarks. If this is the case, prompt the user about bookmarking it."
+  (labels ((bookmarked-url-p (url-address)
+             "The local function `bookmarked-url-p' returns the URL
+           address itself if it is new to the bookmark list or NIL if it is
+           already there."
```
Ambrevar, line 26: The docstring documents the opposite return value of what the name implies.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,51 @@
...
+    (sera:and-let* ((history-entries (with-data-unsafe (history (history-path (current-buffer)))
+                                        (mapcar #'htree:data (alex:hash-table-keys (htree:entries history)))))
+                    (current-url-history
+                      (find (url (current-buffer)) history-entries :test #'equalp :key #'url))
+                    (implicit-visits-value
+                      (nyxt::implicit-visits current-url-history))
+                    (current-url-address
+                      (render-url (url current-url-history)))
+                    (threshold threshold))
+                   (run-thread
```
Ambrevar, line 43: Why run a thread?  If there is a good reason, you should probably document it.
Or move it to the signal, it does not seem that this function should be in charge of running a thread.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,51 @@
...
+                    (if (and (> implicit-visits-value threshold)
```
Ambrevar, line 44: Indentation is off.

source/bookmark-frequent-visits-mode.lisp
```diff
@@ -0,0 +1,51 @@
...
+               (if (member url-address bookmarks-address-list :test #'string=)
+                   nil
+                   url-address))))
+    (sera:and-let* ((history-entries (with-data-unsafe (history (history-path (current-buffer)))
+                                        (mapcar #'htree:data (alex:hash-table-keys (htree:entries history)))))
+                    (current-url-history
+                      (find (url (current-buffer)) history-entries :test #'equalp :key #'url))
+                    (implicit-visits-value
+                      (nyxt::implicit-visits current-url-history))
+                    (current-url-address
```
Ambrevar, line 40: `current-url`.

Ambrevar:
> Oh, sorry. I did not realize this order should have been followed.

So what's your plan?  It still seems to me that #1649 should be addressed first, what about you?

source/bookmark-frequent-visits-mode.lisp
jmercouris: There is a reason to run a thread, if you do NOT run a thread there is a deadlock.

Ambrevar: This is a symptom but not the explanation.  Why is there a dead lock?
Looks like a leaky abstraction that we should fix at the core.

source/bookmark-frequent-visits-mode.lisp
pdelfino: The problem is actually the name of the function. 
If it the URL is a member of the set, then it returns NIL:

```lisp
(if (member url-address bookmarks-address-list :test #'string=)
                   nil
                   url-address)
```
On line 45 of this file, there is this expression:

```lisp
(if (and (> implicit-visits-value threshold)
             (bookmarked-url-p current-url))
    (if-confirm ("Bookmark ~a?" current-url)
                (bookmark-url :url current-url)))
```
If the URL visited is above the threshold **and** if it is **not** bookmarked yet, then proceed to the `if-confirm` macro.

This will make it more clear:

```lisp
(unless (member url-address bookmarks-address-list :test #'string=)
                       url-address)
```

What name do you recommend while keeping this body and this docstring?

source/bookmark-frequent-visits-mode.lisp
pdelfino: Please, see [here](https://github.com/atlas-engineer/nyxt/pull/1591#discussion_r676947802). I picked a misleading name for the function.

source/bookmark-frequent-visits-mode.lisp
Ambrevar: The other way around: keep the name which is perfectly fine, reverse the logic including at the call site.

source/bookmark-frequent-visits-mode.lisp
Ambrevar: You can wrap `member` in `not` to reverse the logic.

pdelfino:
> > Oh, sorry. I did not realize this order should have been followed.
> 
> So what's your plan? It still seems to me that #1649 should be addressed first, what about you?

I was actually having a hard time testing the implementation on #1649. I waited until the morning today to discuss my struggle with @jmercouris. Until the meeting, I was focusing on other tasks, such as (this) issue #1591 and others.

I understand your concerns about the code modularity and they make sense to avoid repetitions.

During the meeting today, I confirmed there were some problems in the source code around history on Nyxt (not on the current implementation). I was suspicious about them. These problems have been documented on issue #1677.

John suggested closing or putting this PR on hold until things in Nyxt's history are better.

So, I will:

1 - wait until things on #1677 are fixed;

2 - finish #1649; and,

3 - come back to #1591.

Ambrevar:
Which of the bugs in #1677?
It seems to me that this change is unrelated to the history bugs, am I
missing something?

pdelfino:
> Which of the bugs in #1677? 
>It seems to me that this change is unrelated to the history bugs, am I missing something?

It was hard to test the implementation. For instance,  [this](https://discourse.atlas.engineer/t/why-having-a-history-disowned-and-a-global-history-source/229) was happening.

Ambrevar:
I suppose you can test this by starting from an empty history, and loading a URL
multiple times.
No need to delete history entries if I understand correctly.

pdelfino:
> I suppose you can test this by starting from an empty history, and loading a URL multiple times. 

Ok, I can try again. Just to be clear: issue #1591 (this original one) is ok to test.
The problem is the accidentally useful one (#1649) which you suggested for me to focus first.

There, on issue #1649, after loading an URL multiple times, I need to use them as history entries. However, the "refresh" of history entries considering what I have just visited did not seem to be stable, even closing Nyxt, and re-starting PC (which is in theory not necessary).

>No need to delete history entries if I understand correctly.

jmercouris:
I think we can revisit this now that I have solved some bugs with the `on-signal-load-finished` handler. Would you care to retry this in a new branch?

pdelfino:
> I think we can revisit this now that I have solved some bugs with the `on-signal-load-finished` handler. Would you care to retry this in a new branch?

No problem at all :)

pdelfino:
Closing this PR!

Issue #1573, opened 2021-06-30
Provide a feature to group or manage sets of buffers


novoid:
Hi, I am copying some thoughts from [this reddit discussion](https://www.reddit.com/r/Nyxt/comments/oa69oe/whats_preventing_you_from_using_nyxt_as_your/) about "Window Titler" (FF add-on) to here.

The underlying requirement I address here is that my current Firefox has 14 Windows with 96 tabs. It would be too cumbersome to switch between all those buffers without some kind of grouping by task/purpose/topic/whatever.

# How I'm currently handling this issue in Firefox

By giving each FF window a name via [the small FF add-on "Window Titler"](https://addons.mozilla.org/en-US/firefox/addon/window-titler/), each window gets an optional prefix like "[News] " which is clearly visible in my task bar. Instead of "What's preventing ..." of this page, I get a window title like "[News] What's pre...". The prefix stays independent of the selected tab within the window.

That way, I separate "sets of tabs" according to their shared relationship usually for a task. It is noteworthy that I use this all the time with very similar web pages. For business use, I may have multiple windows for different tasks that all contain tabs of our Confluence system. Therefore, buffer management by domain does not solve the issue I'm addressing here because I could not differ between Confluence pages related to task X and a different task Y that both require Confluence pages.

Example: a FF window gets the title "ISO" and it contains all tabs I'm working with for ISO compliance. Another window gets "ProjX" which contains all tabs for a specific project. Both do contain Confluence pages of the same space.

There are many different solutions to handle large numbers of tabs. [TreeTabs](https://gitlab.com/kroppy/TreeTabs/-/wikis/home) seems to be a popular one.

# How Nyxt could address the issue of managing many buffers

With Nyxt, it would require the user to remember (and know) which buffer is associated to which task. I would get 96 buffers that are on the same level, without any grouping. From this perspective, "Window Titler" offers a similar functionality than projectile: jumping between things within the same "project".

Nyxt should offer at least some kind of answer to this requirement. Maybe built-in, maybe as an add-on.

[...]

PR #1592, opened 2021-07-05
window: Add `title' slot to allow customizing the window title.

Ambrevar:
Should fix one of the requirements of #1573 about customizing the window title.

Question: Should we name the slot `titler` instead?

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -350,17 +350,25 @@ current buffer."
 (declaim (ftype (function (&optional window buffer)) set-window-title))
 (export-always 'set-window-title)
 (defun set-window-title (&optional (window (current-window)) (buffer (current-buffer)))
-  "Set current window title to 'Nyxt - TITLE - URL.
-If Nyxt was started from a REPL, use 'Nyxt REPL...' instead.
+  "Set current window title to the return value of (title window). "
+  (declare (ignore buffer)) ; TODO: BUFFER is kept for backward compatibility.  Remove with 3.0.
+  (ffi-window-set-title window (funcall (title window) window)))
+
+(declaim (ftype (function (window) string) window-default-title))
+(export-always 'window-default-title)
+(defun window-default-title (window)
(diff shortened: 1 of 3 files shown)


Ambrevar:
While this allows for programmatic customization of the window title, what about an interactive setting?  

It's easy to add a command to set the window title, but what if we just want to set a prefix and still have the current page URL appended?  Then we would need a `title-prefix` slot or similar.

jmercouris:
WHAT IF, instead we had the user supply a format string?

jmercouris:
Less flexible, sure, but also easier to do.

Ambrevar:
But what would be the arguments?

jmercouris:
The arguments would be the URL, the buffer title, etc.

Ambrevar:
But you can't predict what the user will want.
A function is far more flexible and useful here.

jmercouris:
Why don't we just add `cl-crystal-ball` into our dependencies? That way we won't need to be so flexible as provide a function.

Ambrevar:
Is this an actual CL meme? :p

jmercouris:
No no, I was just joking :-)

Ambrevar:
What about the question and call it `titler` instead?  I'm in favour of `titler` to emphasize it's a function.

jmercouris:
If you want to emphasize it is a function, I suggest the `-fn` suffix. Therefore `title-fn`. In any case, it seems just fine to me as `title`.

Ambrevar:
I don't like `fn`, it's not Lispy, it does not read out well.

It's also not very consistent since we don't use `fn` anywhere in the
code base beside the command slot (which was done out of consistency
with Serapeum's hooks, maybe not the wisest thing).
We have many function symbols that are either verbs or "doers".

jmercouris:
OK. Either way :-)

Ambrevar:
Renaming done with 
82fa5ef770de869e46e7763dc8842e85dbdd1253.

Issue #854, opened 2020-07-18
Turn buffer list into a tree


Ambrevar:
In commit 793bb39a8c4ea2b3140e48f8e36763e9cc46bcb0 I fixed yet another bug related to buffer switching.  Our current structure holding the list of buffers is brittle.

I believe this is because we are conflating  2 independent buffer switching features: by access time and by order.  On top of this, our current code is very stateful which makes it hard to debug.

We should have 3 commands:

- "left" buffer.
- "right" buffer
- next most recent buffer.

Note that Firefox recently acquired an option to change the behaviour of C-tab exactly like the above.

We already store the last-access time so the implementation of "next most recent buffer" is trivial  (we could even simplify our code).  However, we don't store the relative "position" of the buffers.

I remember someone mentioned we stored our buffers in a tree to preserve the information of which is the parent buffer (e.g. when opening a link in a new buffer).  I cannot find the issue, feel free to share it if you can find it back.

To implement the "left" and "right" command, we need to flatten the tree in a way that's "do what I mean".  I'd say "append flat left branch and flat right branch" (recursive algorithm).

When deleting a parent buffer, I think we need to move the oldest child in place of the parent.

With a tree, we would also have the possibility to show the tree of all buffers in an IBuffer-like buffer.  See #565.

remedialchaos:
If I'm understanding correctly, this is about making buffer list and switching more aware of last opened/switched times.

Or does this mean something more like "Tree Style Tabs" extension for Firefox? I would personally vote for TST. TST in traditional FF is beneficial because it uses vertical space for tab lists and so it is easier to identify and select. This is beneficial in Nyxt as a persistent bird's eye view for currently open buffers.

But even more benefit is for tracking train of thought. It is like the global history tree, but for current state of all open buffers. It helps me track which tab belongs to which group/subgroup.

Ambrevar:
This is about the implementation details to finally get to "tree style tabs".
The actual tree display is discussed in #565.

Note that this issue is a bit old and now that we've got the Global History Tree, we incidentally also have tree-like relationships between buffers.  We just need to leverage this in our code base.

Ambrevar:
Done!

Issue #565, opened 2020-02-18
Implement IBuffer


Ambrevar:
It would be nice to offer an `ibuffer` command that would display the list of buffers in a special web page.
The list could be sorted, filtered, etc.

Finally, every buffer could be switched to.  This last feature requires some sort of HTML/JS <-> Next protocol.

jmercouris:
> This last feature requires some sort of HTML/JS <-> Next protocol.

Absolutely, this is most certainly a prerequisite to that. A good idea!

Ambrevar:
Also see #745.

jmercouris:
Yes, fair enough. This is also not really a *core* priority at the moment. We have plenty of buffer management tools.

PR #1593, opened 2021-07-05
Buffer tree!

Ambrevar:
This addresses #854 as well as some points mentioned in #1573 and in #565.

This changes the logic of `switch-buffer-previous` and `switch-buffer-next` to finally fix the long-standing issue of "jumping-around buffer last-access".

Now to go to the last visited buffer, use the new `switch-buffer-last`.

`switch-buffer-previous` visits the previous siblings (siblings are sorted by ID), `switch-buffer-next` to the next.

Questions before we can merge: can we browse the whole tree (including children and parents) with just `switch-buffer-previous` and `switch-buffer-next`?
To do so with just 2 commands we'd need to keep track of the visited parents (this is OK, can store it as a buffer slot).  Is this what we want?

Example:

- root
  - childA
    - subchildA1 
    - subchildA2
  - childB
    -  subchildB1
    - subchildB2

If we are on childA, successive calls to switch-buffer-next should visit, in order:
- subchildA1 
- subchildA2
- childB
-  subchildB1
- subchildB2
- root
- childA

and we loop.

Does that make the most sense?

### DIFF
--- a/source/base-mode.lisp
+++ b/source/base-mode.lisp
@@ -18,6 +18,7 @@ This mode is a good candidate to be passed to `make-buffer'."
                      "C-w" 'delete-current-buffer
                      "C-shift-tab" 'switch-buffer-previous
                      "C-tab" 'switch-buffer-next
+                     "C-`" 'switch-buffer-last
                      "C-pageup" 'switch-buffer-previous
                      "C-pagedown" 'switch-buffer-next
                      "C-l" 'set-url
--- a/source/buffer-listing-mode.lisp
+++ b/source/buffer-listing-mode.lisp
@@ -84,3 +84,37 @@
                       (:a :class "button" :href (lisp-url '(nyxt/buffer-listing-mode::show-buffers-panel)) "Update ↺")
                       (loop for buffer in (buffer-list)
                             collect (buffer-markup buffer)))))))
+
+(define-command-global list-buffer-trees ()
+  "Draw the list of buffer trees.
+Buffers have relationships.  When a buffer is spawned from another one (e.g. by
+middle-clicking on a link), the new buffer is a child buffer.
+This kind of relationships creates 'trees' of buffers."
+  (labels ((buffer-markup (buffer)
+             "Present a buffer in HTML."
+             (markup:markup
+              (:p (:a :class "button"
+                      :href (lisp-url `(nyxt::delete-buffer :id ,(id buffer))) "✕")
+                  (:a :class "button"
+                      :href (lisp-url `(nyxt::switch-buffer :id ,(id buffer))) "→")
+                  (:span (title buffer) "  "
+                         (:u (render-url (url buffer)))))))
+           (buffer-tree->html (root-buffer)
+             "Render a single buffer tree to HTML."
+             (markup:markup
+              (:div (markup:raw (buffer-markup root-buffer)))
+              (:ul
+               (loop for child-buffer in (nyxt::buffer-children root-buffer)
+                     collect (markup:markup
+                              (:li
+                               (markup:raw (buffer-tree->html child-buffer)))))))))
+    (with-current-html-buffer (buffer "*Buffers*" 'nyxt/buffer-listing-mode:buffer-listing-mode)
+      (markup:markup
+       (:style (style buffer))
+       (:h1 "Buffers")
+       (:a :class "button" :href (lisp-url '(nyxt/buffer-listing-mode::list-buffers)) "Update")
+       (:br "")
+       (:div
+        (loop for buffer in (buffer-list)
+              unless (nyxt::buffer-parent buffer)
+                collect (buffer-tree->html buffer)))))))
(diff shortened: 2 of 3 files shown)


jmercouris:
Looks great. I would like to make some tweaks to the appearance of this and list-buffers. When the tree depth gets too great, it can become visually hard to read.

jmercouris:
I'll push some stylistic changes to this branch and then merge.

Ambrevar:
Don't merge yet, I'd like to tweak some other things.

In particular, I'd like to merge the command into the original
list-buffers command.

Ambrevar:
> When the tree depth gets too great, it can become visually hard to read.

What about making the roots collapsable?

Ambrevar:
Also what's your take on my question in the original post about the tree traversal?
Shall we add more commands or just have 2 commands that traverse the whole tree in a depth-first fashion?

jmercouris:
I believe few people have the sufficient mental model to keep in mind the tree structure. I think they will be very surprised by `switch-buffer-next` and `switch-buffer-previous` if we make it utilize the tree.

I suggest keeping this model as simple as possible and leaving it the way it is.

jmercouris:
With regards to adding two more commands, I don't see the advantage it provides over simply utilizing `switch-buffer`.

jmercouris:
Lastly, with regards to making roots collapsible; I will work on that CSS, it might be a little bit tricky though.

Ambrevar:
> I believe few people have the sufficient mental model to keep in mind the tree structure. I think they will be very surprised by switch-buffer-next and switch-buffer-previous if we make it utilize the tree.

> I suggest keeping this model as simple as possible and leaving it the way it is.

I think there is a misunderstanding.  I believe that the current implement _is confusing_, because `switch-buffer-previous` and `switch-buffer-next` does cycle through all buffers!

So what I'm suggesting instead is to flatten the tree in a reproducible way and have the commands rotate over the result.  This way the commands will behave like `C-tab` and `C-shift-tab` in other browsers.

Ambrevar:
To clarify: I'm precisely trying to abstract the tree away from the user.

jmercouris:
I am trying to say this:

I believe a list sorted by last access is more intuitive than a flattened tree.

Ambrevar:
> Now, why does our C-tab equivalent not work as in other browsers? I believe it does not work because we treat each keypress as an individual command and update the access time on EACH press for EACH buffer. In a traditional browser or GUI system, the last access for a program does not get updated until you release control and select your new program or tab.

I don't think that's correct: in other browsers, tabs are ordered, so C-tab can trivially cycle across all tabs.  Last access is not used. This is essentially what I'm trying to do by flattening the tree.

Ambrevar:
> I believe a list sorted by last access is more intuitive than a flattened tree.

This is the implementation on master, and it does not work as you've indicated in your other answer.  As soon as we switch to the next buffer, the last access time is updated, thus preventing the user from cycling through all buffers.  This is not the right approach.

jmercouris:
OK, so we wish to emulate a traditional browser and NOT a WM application switcher. Then your flattened tree approach is appropriate.

jmercouris:
My confusion arises from the fact that you wanted C-tab and C-shift-tab to cycle in the past, which is the behavior one expects from a WM application switcher, and not a browser. Now I see that you would like to switch it to the behavior of a traditional browser. That is fine with me!

Ambrevar:
I'm confused, it the same thing for me.  Can you give a precise example of a "WM application
switcher" use?

jmercouris:
Pressing Cmd + Tab in macOS uses most recency to sort applications.

Pressing Cmd + Tab in Safari presents the list in a fixed order.

jmercouris:
Please let me know when I can make changes to this branch. (The style/collapsing functionality).

Ambrevar:
Can you give more details?

I believe that what you mean is that `Cmd+Tab` shows a _list_ of
applications sorted by recency.  It does not switch to the application
before the user has selectted and confirmed.

Note that this is different from `switch-buffer-next` and
`switch-buffer-previous` which go directly to the next buffer, without
showing a menu.  Thus ordering by frequency cannot work.

Does that make sense?

Ambrevar:
Please go ahead, I'm on init-file recovery right now.

Ambrevar:
@aartaka Thoughts on this?

aartaka:
Looks right! I haven't digged deep enough into the code, though.

Ambrevar:
I've just pushed the full implementation of the buffer tree traversal.

So now if you've got a tree as in the original post, say you are on ChildA, pressing `switch-buffer-next` will _always_ bring you to `subchildA1`, and `switch-buffer-previous` will always bring you to ChildA, regardless of the last access time.

This essentially emulates "fixed tab positions" from popular browsers.

Ready to merge for me, let me know what you think.

jmercouris:
I suggest: Let's first merge the spinneret changes, then we convert this to spinneret, and then merge it.

Ambrevar:
OK, I'm on it!

Ambrevar:
Merged with b1e244338c17780cea023fddcb5485ce78259f38.

PR #1595, opened 2021-07-06
Avoid the generation of two final dots "." while inserting command docstrings automatically inside the documentation

pdelfino:
I guess the current macro generates an extra final dot "." since the docstring already has a "."
This is what I see in my manual under version `Version 2.1.1-37-g42946c64` on Master commit  `42946c6459`:

image

If I remove the last part of the string concatenation, I think it will work appropriately. Not sure if this happened **only** in few cases.

P.S.: Please double check before merging it, macros in CL are still not clear for me. Any change on macro definitions should be carefully reviewed :)

### DIFF
--- a/source/input.lisp
+++ b/source/input.lisp
@@ -23,8 +23,7 @@
   `(markup:raw
     (format nil "<span>~a</span>"
             (str:concat (or (first (str:split ". " (documentation ,fn 'function)))
-                            (error "Undocumented function ~a." ,fn))
-                        "."))))
+                            (error "Undocumented function ~a." ,fn))))))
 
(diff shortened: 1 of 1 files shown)


Ambrevar:
The problem is with docstrings that contain only 1 sentence.  With 2 sentences, the code was correct and your fix would be missing a period then.

Fixed with 78c6cb09f6b247db5487ed80710fc04bed419320.

Note: Please only share images when necessary.  Here you could have quoted the manual by prefixing with `>` or by surround with a triple-backquote block.  
Images are heavier, they can't be read properly on all interfaces and the text cannot be copied.  Thanks!

PR #1596, opened 2021-07-06
GTK: Add connect-signal helper to systematically protect signal threads.

Ambrevar:
And yet another protection against crashes! :)


jmercouris:
An unhandled condition in a GTK signal handler leads to a crash I take it?

Ambrevar:
Indeed!

jmercouris:
Thank you!

Issue #1431, opened 2021-05-20
Overhaul user-classes


Ambrevar:
Commits a276091fa0f55daf1400c9a0ee1380982cccf676 and 17f7b7f2d903082b5b0308b07cdad8311a81dadc show how painful it is to consistently define and use "user classes".

We can do better.
Here is what I suggest:

- Remove our `libraries/class-star` and use an up-to-date Git submodule of `defclass-star` instead.
- Define a `define-class` macro which just adds the `:user-class-p` option.
- User classes should be remembered in some unexported list, like we do with `*command-list*`.  We should not rely on the name to identify user classes (this is what we do now and it's brittle).  Add a predicate `user-class-p`.
- Reverse the logic: instead of having the virtual class prefix with "user-", name the virtual class using the given name and prefix the defining parent class with something like `%`.  The parent class should be unexported even if the user class is exported.
- `define-configuration` generated classes can follow the same logic and be named `%foo17` for the `foo` user class.
- For the rare cases where we need to redefine the superclasses of a user class, we can use `defclass` manually, e.g.

  ```lisp
  (defclass browser (gtk-browser %browser))
  ```

  But then `browser` would lose its docstring (which is in `%browser`).
  So maybe keep `define-user-class` (unexported) for these cases?  It expands to

  ```lisp
  > (macroexpand-1  '(define-user-class browser (gtk-browser)))
  (PROGN
   (EXPORT-ALWAYS 'USER-BROWSER (SYMBOL-PACKAGE 'USER-BROWSER))
   (SB-MOP:ENSURE-CLASS 'USER-BROWSER :DIRECT-SUPERCLASSES '(GTK-BROWSER BROWSER)
                        :DOCUMENTATION (DOCUMENTATION 'BROWSER 'TYPE)))
  ```

  Or just call the above `ensure-class` manually, no need for `define-user-class`.

- `describe-class` would no longer list both user-classes and the defining parent class.

Ambrevar:
After some thinking here is a revised proposal: what about using metaclasses to add a user-reserved method to customize instances?

The original reason why we created this user-class macro is because users can't use `initialize-instance :after` since it's already used by the code.  But then, I propose this (now) obvious fix; leverage MOP to add a new method executed after initialization, which leaves full control to the user what to do when the instance has been initialized.

In practice, it would look like this from the `init.lisp`:
```lisp
(defmethod user-class-init ((buffer buffer))
  (do whatever)
[...]

PR #1598, opened 2021-07-06
Recover from broken init

Ambrevar:
This is where functional style and user classes really pay off!

Previously, if an init file was syntactically correct but would break Nyxt otherwise, the browser would fail to start in the weidest ways (e.g. blank window).

Now startup errors are caught and the state is almost completely reinitialized, then the error is displayed as usual using the reported.

Notes:
- The user classes implementation is completely upside down, but I don't think we can easily overhaul this without breaking backward compatibility.  We can do this for 3.0.  See https://github.com/atlas-engineer/nyxt/issues/1431.

- If you want to test, in particular compare the behaviour with master, try adding this to your init file:

```lisp
(define-configuration window
  ((status-formatter 17)))
```

From KABOOM! to swooosh! :)

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -225,6 +225,17 @@ prevents otherwise."))
   "Get the window containing a buffer."
   (find buffer (alex:hash-table-values (windows browser)) :key #'active-buffer))
 
+(defun reset-browser (browser &key message)
+  (setf *browser*
+        (make-instance 'user-browser
+                       :startup-error-reporter-function (if message
+                                                            (lambda ()
+                                                              (echo-warning "Resetting state and restarting due to `startup' error: ~a" message)
+                                                              (error-in-new-window "*Initialization errors*" message))
+                                                            (startup-error-reporter-function browser))
+         :startup-timestamp (startup-timestamp browser)
+         :socket-thread (socket-thread browser))))
+
 (defmethod finalize ((browser browser) urls startup-timestamp)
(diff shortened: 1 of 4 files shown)


Ambrevar:
OK to merge? @jmercouris @aartaka

jmercouris:
I don't know how I feel about a 'reset' function rather than a functional start. Hence why I haven't commented yet! Could you please walk me through some of your decision making?

Ambrevar:
A functional (re)start would be great, but... Can we?

GTK is heavily non-functional.  Maybe we can join the GTK thread, thus
stopping it, then completely start over passing `:no-init t` to `start`.
Seems more elegant indeed.
I'll give it a try.

Note that we have to "reset" the user classes anyways.  Does that make sense?

jmercouris:
Yes, that makes sense :-)

Ambrevar:
Done, we are now restarting the Nyxt _executable_ on error.  Only once, to avoid looping forever.  This should be much more bullet proof.

OK to merge now?

jmercouris:
Ok!

jmercouris:
Works well! Thank you!

Ambrevar:
Note for future reference: 

> Note that we have to "reset" the user classes anyways.  Does that make sense?

This is only true when run from the REPL.

Thank you John for testing and suggesting to restart the browser, it's much better now!

PR #1605, opened 2021-07-07
urls: Memoize lookup-hostname.

Ambrevar:
iolib/sockets:lookup-hostname is particularly slow.  Memoizing it works around
this issue.  In practice, many URLs tend to come from the same hosts.

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -52,6 +52,23 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
          (or (ignore-errors (ffi-display-url url))
              url))))
 
+(defmacro defmemo (name params &body body) ; TODO: Replace with https://github.com/AccelerationNet/function-cache?
+  (alex:with-gensyms (memo-table args result result?)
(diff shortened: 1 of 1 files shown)


aartaka:
All good, although I feel that `defmemo` (or something like it) should have been part of some utility library...

Ambrevar:
I could find a definition in
https://github.com/AccelerationNet/function-cache and fare-memoization.

Ambrevar:
https://gitlab.common-lisp.net/frideau/fare-memoization

aartaka:
I mean, it feels likebelonging to Alexandria or Serapeum. Anyway, thanks for the patch and references!

Ambrevar:
Should we add the fare-memoization or function-cache dependency to Nyxt
then?

The defmemo definition I've used here is very basic.
Among others, it does not time out and it's not thread-safe.

jmercouris:
I would suggest let's wait on introducing a new library. First let's see if both of those problems arise.

Ambrevar:
It's not urgent for sure and I've already put a TODO: comment so we remember.  I'll
merge then.

Issue #1576, opened 2021-06-30
Feature request: create a command so that the user can edit their bookmark data (URL, title, tags) - `edit-bookmarks`


pdelfino:
It is already possible to edit bookmarks data in two ways. First, the user can edit the "raw" bookmarks.lisp local file. It is also possible to do it in another way @jmercouris showed me earlier today. I cannot remember, but he agreed that there could be a more direct and easier command for it.

Since I have learnt a few things about bookmarks while working on #1548, I will try to solve this :muscle:

Ambrevar:
Agreed.

Currently we can edit tags by simply re-bookmark.  We can't edit the title without reloading the page though.

I think `list-bookmarks` would be the most convenient places to perform these edits.

Since we have support for external editors, we could also add a command that opens the bookmark file externally for editing.

Ambrevar:
Eric gave me an idea (https://discourse.atlas.engineer/t/config-file-path/202/3): what if we add a command `edit-user-file` that prompts for any user file (init.lisp, bookmarks, history, etc.) and open it in the external editor?

This would be a great way to solve the issue of the user wondering where the files are.

Ambrevar:
To do so, we need a way to list the data-path instances.  I believe that storing them in an internal `*data-path-list*` on instantiation should do it.

jmercouris:
Is this list mutable? Based on my understanding it should be `+data-path-list+`.

Ambrevar:
It's the list of instances.  The user can also create their own
data-path, so yes, the list would be mutable, just like the command list
is mutable.

jmercouris:
OK!

ericdrgn:
Same Eric from the forum btw, like this idea a lot @Ambrevar, it definitely makes sense to do it that way.

Ambrevar:
Remove the `high` tag since we now have `edit-user-file-with-external-editor`.

We could still improve bookmark listing to make it editable.

PR #1606, opened 2021-07-07
Edit data paths with external editor

Ambrevar:
This addresses https://github.com/atlas-engineer/nyxt/issues/1576.

What's left to do:

- [x] Add button on `help` page to run this command.
- [x] Allow creating the file if it does not exist.  Does it always make sense?  Just no if file is editable I guess.
- The weak hash table highlights that prompt buffer don't get garbage collected properly.  Recipe:

```lisp
> (hash-table-count *data-paths*)
N
> ; open a prompt buffer, then close it.
> (setf (old-prompt-buffers *browser*) nil)
> (tg:gc :full t)
NIL
> (hash-table-count *data-paths*)
P
```

Result: `P > N` while we should have `P=N`.

Any idea where the reference to old prompt buffers is kept?


jmercouris:
Perhaps it is kept in a window object?

Ambrevar:
@jmercouris @aartaka Do you know a way to keep track of instantiated
object in Common Lisp?

I suppose we could set `initialize-instance :after (t)` and record
everything that belongs to the Nyxt package.

Anything better?

Can we somehow track references?

jmercouris:
I'll ask on IRC. I don't know of a way to do this.

jmercouris:
Some great advice from IRC:

```
<Bike> what the thing i'm recalling does is give you a path from roots
    to a given object. since the object you're looking for is already
    alive for some reason, you're not making it any more impossible to
    collect
<Bike> here we go, sb-ext:search-roots
<Bike> "Find roots keeping the targets of [a weak pointer] alive"
<Bike> sounds useful for finding leaks, but i've never used it myself

```

Ambrevar:
Great find!

With this, all that's missing is a systemic indexation of Nyxt objects
with weak pointers.

Would initialize-instance suffice there?

jmercouris:
I don't see why not...!

Ambrevar:
Alright, I tried this:

```lisp
(defvar *all-nyxt-objects* '()
  "Weak list.")

(defmethod initialize-instance :after ((object t)  &key)
  (when (member (symbol-package (sera:class-name-of object))
                (list (find-package :nyxt)
                      (find-package :nyxt-user)))
    (push (tg:make-weak-pointer object) *all-nyxt-objects*)))

(defun cleanup-all-nyxt-objects ()
  "`*all-nyxt-objects*' must not be changed while this is running."
  (let ((all-nyxt-objects *all-nyxt-objects*)) ; Just make sure we are handling a consistent object.
    (maplist (lambda (l)
               (when (and (second l)
                          (null (tg:weak-pointer-value (second l))))
                 (setf (rest l) (rest (rest l)))))
             all-nyxt-objects)
    (setf *all-nyxt-objects*
          (if (tg:weak-pointer-value (first all-nyxt-objects))
              (rest all-nyxt-objects)
              all-nyxt-objects))))
```

Then you can print the root of the object with

```lisp
(sb-ext:search-roots (second *all-nyxt-objects*) :print :verbose )
```

Or all roots in one go with

```lisp
(sb-ext:search-roots  *all-nyxt-objects* :print :verbose)
```

Awesome!

It helps, but it won't automatically reveal _why_ an object is not collected, it requires some deeper investigation.

You can force a garbage collection with `tg:gc`.  You can run `cleanup-all-nyxt-objects` manually to free the list of the collected objects.  I could have used a weak hash table to do this automatically, but then I don't know how to access the weak pointers in the hash table to feed them to `sb-ext:search-roots`.

I'll polish this and include it in Nyxt when `debug` is on or something like that.

Ambrevar:
I believe this is ready to be merged.  The memory leak issue is unrelated to this anyways and will be worked out as part of #1612.

PR #1613, opened 2021-07-09
Macroedit

jmercouris:
Ability for the user to create and edit macros!

see attached screenshot to get excited!
Screenshot_20210709_150537

### DIFF
--- a/source/macro-edit-mode.lisp
+++ b/source/macro-edit-mode.lisp
@@ -1,0 +1,19 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/macro-edit-mode
+  (:use :common-lisp :trivia :nyxt)
+  (:documentation "Mode for editing macros."))
+(in-package :nyxt/macro-edit-mode)
+
+(define-mode macro-edit-mode ()
+  "Mode for creating and editing macros."
+  ((rememberable-p nil)))
+
+(define-command-global edit-macro ()
+  "Edit a macro."
+  (with-current-html-buffer (buffer "*Macro edit*" 'nyxt/macro-edit-mode:macro-edit-mode)
+    (markup:markup
+     (:style (style buffer))
+     (:h1 "Macro edit")
+     (:p "Hello world!"))))
--- a/source/macro-edit-mode.lisp
+++ b/source/macro-edit-mode.lisp
@@ -2,13 +2,15 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (uiop:define-package :nyxt/macro-edit-mode
-  (:use :common-lisp :trivia :nyxt)
+    (:use :common-lisp :trivia :nyxt)
   (:documentation "Mode for editing macros."))
 (in-package :nyxt/macro-edit-mode)
 
 (define-mode macro-edit-mode ()
   "Mode for creating and editing macros."
-  ((rememberable-p nil)))
+  ((functions
+    (make-hash-table)
+    :documentation "A list of functions the user has added to their macro.")))
 
 (define-command-global edit-macro ()
   "Edit a macro."
@@ -16,4 +18,12 @@
     (markup:markup
      (:style (style buffer))
      (:h1 "Macro edit")
-     (:p "Hello world!"))))
+     (:p "Name")
+     (:input :type "text")
+     (:p "Commands")
+     (:p (:a :class "button"
+             :href "xyz" "+ Add command"))
+     (:br)
+     (:hr)
+     (:a :class "button"
+         :href "xyz" "Save macro"))))
--- a/source/macro-edit-mode.lisp
+++ b/source/macro-edit-mode.lisp
@@ -8,9 +8,16 @@
 
 (define-mode macro-edit-mode ()
   "Mode for creating and editing macros."
-  ((functions
+  ((function-sequence
+    0
+    :documentation "A variable used to generate unique identifiers for user
+    added functions.")
+   (functions
     (make-hash-table)
     :documentation "A list of functions the user has added to their macro.")))
+
+(defmethod get-unique-function-identifier ((mode macro-edit-mode))
+  (incf (function-sequence mode)))
 
 (define-command-global edit-macro ()
   "Edit a macro."
@@ -22,8 +29,23 @@
      (:input :type "text")
      (:p "Commands")
      (:p (:a :class "button"
-             :href "xyz" "+ Add command"))
+             :href (lisp-url '(nyxt/macro-edit-mode::add-command)) "+ Add command"))
      (:br)
      (:hr)
      (:a :class "button"
-         :href "xyz" "Save macro"))))
+         :href (lisp-url '(nyxt/macro-edit-mode::save-macro)) "Save macro"))))
+
+(define-command add-command (&optional (macro-editor (current-mode 'macro-edit-mode)))
+  "Add a command to the macro."
+  (let ((command 
+          (first
+           (prompt
+            :prompt "Describe command"
+            :sources (make-instance 'user-command-source)))))
+    (setf (gethash (get-unique-function-identifier macro-editor) 
+                   (functions macro-editor))
+          command)))
+
+(define-command save-macro (&optional (macro-editor (current-mode 'macro-edit-mode)))
+  "Save the macro to the auto-config.lisp file."
+  (echo "Macro saved."))
--- a/source/macro-edit-mode.lisp
+++ b/source/macro-edit-mode.lisp
@@ -26,7 +26,7 @@
      (:style (style buffer))
      (:h1 "Macro edit")
      (:p "Name")
-     (:input :type "text")
+     (:input :type "text" :id "macro-name")
      (:p "Commands")
      (:p (:a :class "button"
              :href (lisp-url '(nyxt/macro-edit-mode::add-command)) "+ Add command"))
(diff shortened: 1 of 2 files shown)


aartaka:
What about prompt-buffer programmable input? No macro system is complete without recording keys and replaying them :/

jmercouris:
That's a record function not a create and edit function!

jmercouris:
One step at a time!

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+  (:documentation "Mode for editing macros."))
+(in-package :nyxt/macro-edit-mode)
+
+(define-mode macro-edit-mode ()
+  "Mode for creating and editing macros."
+  ((name
+    ""
+    :documentation "The name used for the macro."
+    :accessor nil)
+   (function-sequence
```
aartaka, line 15: The name is somewhat confusing. `function-order`, `current-function-id`?

aartaka (approved):
All good, except for confusing counter slot.

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+           (ps:ps
+             (ps:chain document (get-element-by-id "macro-name") value)))))
+    (cond ((not (str:emptyp name)) (setf (slot-value macro-editor 'name) name))
+          ((slot-value macro-editor 'name) (slot-value macro-editor 'name))
+          (t nil))))
+
+(defmethod generate-macro-form ((macro-editor macro-edit-mode))
+  (let ((name (intern (name macro-editor)))
+        (commands (mapcar
+                   (lambda (command) (write `(,(name command)) :stream nil))
```
Ambrevar, line 96: Am I mistaken or this `write` is a no-op and you can just return `(,(name command))`?

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+                   (alexandria:hash-table-values (functions macro-editor)))))
+    `(define-command-global ,name () "User generated macro form" ,@commands)))
+
+(define-command add-command (&optional (macro-editor (current-mode 'macro-edit-mode)))
+  "Add a command to the macro."
+  (let ((command 
+          (first
```
Ambrevar, line 103: Style: Maybe fold some lines.

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+
+(defmethod macro-form-valid-p ((macro-editor macro-edit-mode))
+  (and (name macro-editor)
+       (alexandria:hash-table-values (functions macro-editor))))
+
+(define-command save-macro (&optional (macro-editor (current-mode 'macro-edit-mode)))
+  "Save the macro to the auto-config.lisp file."
+  (if (macro-form-valid-p macro-editor)
+      (nyxt::append-configuration (generate-macro-form macro-editor))
+      (echo "Your macro form is invalid; check if you have a title and functions.")))
```
Ambrevar, line 117: "it has a title"

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+
+(define-command evaluate-macro (&optional (macro-editor (current-mode 'macro-edit-mode)))
+  "Evaluate the macro for testing."
+  (if (macro-form-valid-p macro-editor)
+      (progn (eval (generate-macro-form macro-editor)) (echo "Macro evaluated."))
```
Ambrevar, line 122: Maybe insert a line break in the progn?

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+      (echo "Your macro form is invalid; check if you have a title and functions.")))
```
Ambrevar, line 123: "it has a title"

source/macro-edit-mode.lisp
Ambrevar: Agreed, but see below for alternatives.

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+   (function-sequence
+    0
+    :documentation "A variable used to generate unique identifiers for user
+    added functions.")
+   (functions
+    (make-hash-table)
+    :documentation "A list of functions the user has added to their macro.")))
+
+(defmethod get-unique-function-identifier ((mode macro-edit-mode))
+  (incf (function-sequence mode)))
```
Ambrevar, line 24: It seems that the ID is never user facing, right?  Then why not use`gensym`?  After all, it's here for exactly this, to generate unique identifiers :)

Ambrevar: Alternatively, you could use a closure as a slot value to avoid the defmethod altogether:

```lisp
(defclass foo ()
  ((counter :initform (let ((c 0))
                       (lambda () (incf c)))
           :accessor counter)))

(let ((foo (make-instance 'foo)))
  (values (funcall (counter foo))
          (funcall (counter foo))))
```

Note in this case that I chose to call it `counter` to be consistent with the standard `*gensym-counter*`.

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/macro-edit-mode
+    (:use :common-lisp :trivia :nyxt)
```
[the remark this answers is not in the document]
jmercouris: Emacs strikes again...!

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+  (:documentation "Mode for editing macros."))
+(in-package :nyxt/macro-edit-mode)
+
+(define-mode macro-edit-mode ()
+  "Mode for creating and editing macros."
+  ((name
+    ""
+    :documentation "The name used for the macro."
+    :accessor nil)
```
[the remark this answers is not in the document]
jmercouris: No, I actually want 'name exported because it saves me having to export the method of the same 'name

source/macro-edit-mode.lisp
```diff
@@ -0,0 +1,123 @@
...
+        command)
+  (render-functions macro-editor))
+
+(defmethod remove-function ((macro-editor macro-edit-mode) command-id)
+  (remhash command-id (functions macro-editor))
+  (render-functions macro-editor))
+
+(defmethod name ((macro-editor macro-edit-mode))
+  (let ((name
+          (ffi-buffer-evaluate-javascript
```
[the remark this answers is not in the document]
jmercouris: Agreed!

source/macro-edit-mode.lisp
jmercouris: 2

source/macro-edit-mode.lisp
jmercouris: I don't want to use gensym because it will make the lisp urls look like nonsense. I am quite happy with the current approach!

source/macro-edit-mode.lisp
jmercouris: Good catch, thank you!

source/macro-edit-mode.lisp
jmercouris: If you have experience with SQL databases, the name is quite logical :-D

source/macro-edit-mode.lisp
Ambrevar: Why?  Gensym can generate unique numbers too, e.g.

```lisp
(symbol-name (gensym ""))
```

jmercouris: Also I can't remove the defmethod, who will set the key to the hash table?

jmercouris: Oh I see you meant `get-unique-function-identifier`

Ambrevar: Actually

```lisp
(parse-integer (symbol-name (gensym "")))
```

source/macro-edit-mode.lisp
Ambrevar: But this is not related to SQL, nor should we use SQL idioms in Lisp code.
I would expect a sequence from "function-sequence", not a number or an identifier.

source/macro-edit-mode.lisp
Ambrevar: Why not?  3 s-exps on the same line is a bit hard to read.

source/macro-edit-mode.lisp
Ambrevar: > I can't remove the defmethod, who will set the key to the hash table?

You can remove `get-unique-function-identifier` and replace it with either the closure or the gensym trick.

jmercouris: After thinking for a little, I still like the original approach more, I will name the slot something more obvious to non-SQL users :-)

Ambrevar: What's your rationale?

jmercouris: wow these messages are so out of order, thanks Bill Gates!

jmercouris: Also feels wrong to intern a symbol for this... polluting our namespace. Hm...

jmercouris:
Well I ended up using the gensym, felt better than the closure, but still felt wrong to be using gensym for such a purpose! Well, I'll let it sink in, worst case we can always change it :-)

Ambrevar:
gensym generate uninterned symbols.

In my opinion, it feels perfectly fine to use gensyms for this :)
It generate syms, which are identifiers, exactly what we want, no?

jmercouris:
Ah, they are uninterned? then I have no problem with it! I should read the spec more carefully!

Ambrevar:
Run `gensym` in a REPL, it returns something like `#:G893`.

The `#:` is the reader macro for uninterned symbols.

Issue #904, opened 2020-08-24
Add buttons to modes in status view


Ambrevar:
I suggest we add 2 buttons to the mode list in the status view:

- 1 button that wraps over the full list of modes and that triggers `disable-mode-for-current-buffer`
- 1 button over a new `+` sign that triggers `enable-mode-for-current-buffer`.

Thoughts?

jmercouris:
An interesting idea! My counter points is the following: will a new user be enabling/disabling modes?

I see modes as an advanced feature, and I was even thinking about a way of reducing their size in the default format-status function. I was thinking of abbreviating them somehow.

A question that may help us resolve this is the following: who are the buttons for, what purpose do they serve? As an advanced user, do you find yourself still using the buttons? If so, it may be wise, as you've suggested to add these buttons.

Ambrevar:
A good point.  We can expect most new users to be unfamiliar with modes.
In particular, new users start with "[web base]" and this might not
speak to them much.

However, when they see the list of modes, if they can click on them (and
the "+"), it will become immediately obvious what they can do with it.

So I'm in favour of adding buttons because it's a door to the path of
more customization for newcomers.

jmercouris:
I have concluded that I agree with you, we can add a plus and minus for the modes area.

Ambrevar:
Thanks!  Will do later then, unless someone beats me to it ;)

Ambrevar:
Buttons got added by John in https://github.com/atlas-engineer/nyxt/commit/e131585fd5574cfe5b1e1d7d67c4508d68bbf470.
Ellipsis for mode listing and toggling in https://github.com/atlas-engineer/nyxt/pull/1615.

jmercouris:
The user can now add and remove modes via the "+" symbol in the far right. I almost want to make it a Δ because that would be cool, but I think the "+" is more obvious. Anyways, the criteria here are fulfilled. Closing for now. Please feel free to reopen!

PR #1615, opened 2021-07-12
status: Add ellipsis after mode list to toggle modes.

Ambrevar:
It's also useful to list the enabled modes when too many of them cannot be displayed.

This is particularly important since e131585fd5574cfe5b1e1d7d67c4508d68bbf470: hovering over the status modes won't display the full list anymore.  The ellipsis reintroduces this behaviour.

Should close https://github.com/atlas-engineer/nyxt/issues/904.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -471,13 +471,13 @@ Delete it with `ffi-buffer-delete'."
                 :margin-left "-10px")
                ("#container"
                 :display "grid"
-                ;; Columns: controls, url, tabs, modes
-                :grid-template-columns "120px 2fr 3fr 240px"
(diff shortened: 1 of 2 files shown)


jmercouris:
How about just putting a "i" at the front of the modes list with all of the modes instead of introducing a new column?

jmercouris:
can I show you an example of what I mean?

Ambrevar:
Sure!

Why an "i"?
What do you mean "with all the modes"?

jmercouris:
This is what I mean: https://github.com/atlas-engineer/nyxt/pull/1616

why an "i"? because it is the universal symbol for information.

jmercouris:
It also adds our other desire to allow the user to toggle modes on and off!

PR #1616, opened 2021-07-12
status.lisp: show enabled modes

jmercouris:
additionally allow clicking to toggle enable/disable modes

### DIFF
--- a/source/status.lisp
+++ b/source/status.lisp
@@ -7,6 +7,9 @@
 (defun format-status-modes (buffer window)
   (markup:markup
    (:span (when (nosave-buffer-p buffer) "⚠ nosave"))
+   (:span (:a :class "button"
+              :href (lisp-url '(nyxt:toggle-modes))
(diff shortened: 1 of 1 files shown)


Ambrevar:
OK, it's better, the only thing is that the circled `i` only displays if you've got the font.  I don't! :p 
Any alternative?

@jmercouris Which fonts do you have on your system?  Maybe we should have Nyxt depend on some lightweight fonts, like the awesome fonts.

jmercouris:
I use noto font on my system, let me look for a more common symbol. I'm hesitant to include a font in Nyxt, maybe we will have to at some point!

Ambrevar:
Unfortunately, Noto is 1.5 GiB strong, so it's a big no-no for lots of
people :p

jmercouris:
Do you have 'ℹ'?

jmercouris:
Do you have '⊕'?

jmercouris:
Do you have '⏼'?

Ambrevar:
I have `⊕`.

Issue #1588, opened 2021-07-04
.deb is broken on Pop!_OS


TheCatster:
When attempting to install the `.deb` on Pop!_OS, it fails saying that enchant is a requirement that cannot be fulfilled. The exact message is:
```bash
❯ sudo apt install ./nyxt_2.1.1_amd64.deb
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'nyxt' instead of './nyxt_2.1.1_amd64.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 nyxt : Depends: enchant but it is not installable
E: Unable to correct problems, you have held broken packages.
```
Any ideas on what might be the cause of this? I am running Pop!_OS 21.04, and have updated apt sources to ensure I have the latest.

MasterEnoc:
I was going write the same issue. The 'enchant' package is not available anymore in some distros (all distros base in Debian testing, maybe other distros with newer software). 

Nyxt team, would you consider add a 'debian' directory to make a .deb package instead or running a common-lisp script? I could help making a PR with the 'debian' directory and all the needed files.

Ambrevar:
Thanks for the feedback.  This was actually a typo, fixed with 71ee37c4170ab43c9b8d1fece6dff4f1645f8f81.

@MasterEnoc Always happy to help get Nyxt packaged on more distros, but out of curiosity, how would a `debian` directory help here?  Can you elaborate?

MasterEnoc:
* It could help to possible add nyxt to the debian package repository (and other debian base repositories), but this will need an active debian maintainer or a prospect debian maintainer.

* I noticed a comment in 'nyxt-ubuntu-package.asd' about having problems with the current building script

It would not help a lot in the current workflow, but you can have the 'debian' directory for a future debian maintainer and the 'common-lisp script' for you.

Ambrevar:
> I noticed a comment in 'nyxt-ubuntu-package.asd' about having problems with the current building script

Which one?  The only comment is not really a problem.

Before adding a `debian` directory, I suggest we first find a Debian maintainer.  That person will be in a good position to guide us through what's needed to make Nyxt part of Debian.

MasterEnoc:
> Which one? The only comment is not really a problem.

My bad, I misundestood the comment.

[...]

PR #1618, opened 2021-07-13
Updated README for debian based systems when building from source

MarkusSagen:
On Debian-based distros, `enchant` has been removed.
Using enchant-2 however, resolves the issue, and building from source works then (issue #1588).
Tested on Ubuntu and Pop!_OS


Ambrevar:
Good catch, thanks!

PR #1620, opened 2021-07-14
Add link to QR code

ag91:
**Note: I am new to Common Lisp and Nyxt development practices and I didn't manage to build this locally (I got some errors). Some guidance would be very welcome (please @Ambrevar !), because I would like to contribute more once I get how this works :)**

This lets you easily pass the link of the page you are viewing to your
mobile phone. Thanks to Karl Voit for the idea.

For reference, I wrote more about the reasoning here:
https://ag91.github.io/blog/2021/06/29/extending-nyxt-via-emacs-how-to-leverage-common-lisp-wealth-to-get-your-links-as-qr-codes/

### DIFF
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -697,13 +697,6 @@ ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
   (with-current-buffer buffer
     (let* ((pathname (expand-path
                      (make-instance 'data-path :basename "qrcode.png")))
-          (url (quri:render-uri (url (current-buffer))))
-          (title (str:concat "QRcode for " url)))
-      (unwind-protect
-           (progn
-             (cl-qrencode:encode-png url :fpath pathname)
-             (make-buffer-focus :url (quri::make-uri-file :path pathname))
-             (sleep 0.1) ;; TODO it seems I cannot open the buffer in time if I delete the file at the end
-             (setf (title (current-buffer)) title)
-             )
-        (uiop:delete-file-if-exists pathname)))))
+          (url (quri:render-uri (url (current-buffer)))))
+      (cl-qrencode:encode-png url :fpath pathname)
+      (make-buffer-focus :url (quri::make-uri-file :path pathname)))))
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -695,6 +695,15 @@ ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
 (define-command show-qrcode-of-current-url (&optional (buffer (current-buffer)))
   "Show the QR code containing the URL for the current buffer."
   (with-current-buffer buffer
-    (let ((pathname (str:concat (namestring uiop:*temporary-directory*) "qrcode.png")))
-        (cl-qrencode:encode-png (quri:render-uri (url (current-buffer))) :fpath pathname)
-        (uiop:run-program (list "nyxt" pathname)))))
+    (let* ((pathname (expand-path
+                     (make-instance 'data-path :basename "qrcode.png")))
+          (url (quri:render-uri (url (current-buffer))))
+          (title (str:concat "QRcode for " url)))
+      (unwind-protect
+           (progn
+             (cl-qrencode:encode-png url :fpath pathname)
+             (make-buffer-focus :url (quri::make-uri-file :path pathname))
+             (sleep 0.1) ;; TODO it seems I cannot open the buffer in time if I delete the file at the end
+             (setf (title (current-buffer)) title)
+             )
+        (uiop:delete-file-if-exists pathname)))))
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -691,3 +691,10 @@ ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
   ;; Need to force document-model re-parsing.
   (setf (document-model mode) nil)
   url)
+
+(define-command show-qrcode-of-current-url (&optional (buffer (current-buffer)))
+  "Show the QR code containing the URL for the current buffer."
+  (with-current-buffer buffer
+    (let ((pathname (str:concat (namestring uiop:*temporary-directory*) "qrcode.png")))
+        (cl-qrencode:encode-png (quri:render-uri (url (current-buffer))) :fpath pathname)
+        (uiop:run-program (list "nyxt" pathname)))))
--- a/build-scripts/guix.scm
+++ b/build-scripts/guix.scm
@@ -159,6 +159,7 @@
        ("cl-markup" ,cl-markup)
        ("cl-ppcre" ,cl-ppcre)
        ("cl-prevalence" ,cl-prevalence)
+       ("cl-qrencode" ,cl-qrencode)
        ("closer-mop" ,cl-closer-mop)
        ("cluffer" ,cl-cluffer)
        ("dexador" ,cl-dexador)
--- a/documents/SOURCES.org
+++ b/documents/SOURCES.org
@@ -20,6 +20,7 @@ which external dependencies we are directly relying on.
 - cl-ppcre
 - cl-ppcre-unicode
 - cl-prevalence
+- cl-qrencode
 - closer-mop
 - dexador
 - enchant
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -33,6 +33,7 @@
                cl-ppcre
                cl-ppcre-unicode
                cl-prevalence
+               cl-qrencode
                closer-mop
                cl-containers
                moptilities


jmercouris:
No help from me :'(? OK :'(

JK, if you want Pierre, no problem, or I can help too if that is OK!

Ambrevar:
Which errors did you get?  Can you provide a recipe (the list of the commands you ran) and the output?

source/web-mode.lisp
```diff
@@ -691,3 +691,10 @@ ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
   ;; Need to force document-model re-parsing.
   (setf (document-model mode) nil)
   url)
+
+(define-command show-qrcode-of-current-url (&optional (buffer (current-buffer)))
+  "Show the QR code containing the URL for the current buffer."
+  (with-current-buffer buffer
+    (let ((pathname (str:concat (namestring uiop:*temporary-directory*) "qrcode.png")))
```
Ambrevar, line 698: We should avoid touching directories outside of ~/.cache/nyxt, ~/.local/share/nyxt and ~/config/nyxt.

I suggest storing the qrcode in ~/.local/share/nyxt/qrcode.png instead.
Look at how `bookmarks-data-path` is defined and used.  Let me know if you need help.

ag91:
Ah, lol! No, I just summoned Pierre because he suggested to open a
PR with this change in the first place XD

I would much appreciate any help, so thanks John!!

source/web-mode.lisp
```diff
@@ -691,3 +691,10 @@ ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
...
+        (cl-qrencode:encode-png (quri:render-uri (url (current-buffer))) :fpath pathname)
```
Ambrevar, line 699: Are you sure you want to use `quri:render-uri` with puni encoding?
If you want to decode the puni codes, use `render-url`.

source/web-mode.lisp
```diff
@@ -691,3 +691,10 @@ ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
...
+        (uiop:run-program (list "nyxt" pathname)))))
```
Ambrevar, line 700: Don't we want to open the QR code in a new buffer instead?

Ambrevar:
Thank you very much for working on this, it is going to be a great addition!

To do:
- Add cl-qrencode to the Guix dependencies.
- Add a hint command for this.

jmercouris:
A hint command? Hint to url encode? Why not just add an action to element-hints that are URLs?

source/web-mode.lisp
jmercouris: I disagree, here I think `uiop:*temporary-directory*` is good for this purpose. Why would we not want to use `/tmp`?

source/web-mode.lisp
jmercouris: That could work :-).

ag91: Indeed, that is what I wanted to do but I do not know how. So I called Nyxt as an external program to open a new buffer.
I was trying `uiop:with-temp-file` to write the image there and open the file in a Nyxt buffer, but I surrendered after a few attempts (because I need some more practice of CL for that to work apparently). (That is also how I got to the *temporary-directory* thingy)

source/web-mode.lisp
ag91: Totally not sure, it was the first thing that worked to get a string out of URI object. Is `render-url` [this one](https://github.com/atlas-engineer/nyxt/blob/master/source/urls.lisp#L45)? It seems to do call `quri:render-uri` as well: how does it change the encoding?

source/web-mode.lisp
ag91: My rationale was to have a throwaway picture, so if I can even skip to create the file would be perfect (I saw I can get a stream from `cl-qrencode`, but I don't know how to pass it to Nyxt to display a volatile buffer).

Ambrevar:

> A hint command? Hint to url encode? Why not just add an action to element-hints that are URLs?

Yes, that's what I meant, sorry for the confusing request :)

source/web-mode.lisp
Ambrevar: A temp file may always be left behind, e.g. if Nyxt crashes or gets killed before it gets cleaned up.
You don't want to leave anything in /tmp lest this directory is readable by other users, which could compromise your privacy.

Finally, `data-path` is great for defining paths because it gives full customizability to the user.  So if the user cares where they write temporary stuff, they can customize it this way.

ag91:
- Add cl-qrencode to the Guix dependencies.
  I thought guix.scm was doing the trick! When I tried to build from Guix I got indeed an error that cl-qrencode was not available. It is on Guix repo though.

- Add a hint command for this.
  What are hints in Nyxt? I think I didn't notice a feature! Do you have a screenshot? Not sure how to add one, but if I see one I can investigate the code :)

source/web-mode.lisp
Ambrevar: Yes, this one.  Notice it _also_ call ffi-render-url, which defers to WebKitGTK to decode the puni codes.
This is useful for instance when the URL contains non latin characters.  But in this case, I guess the right thing to do is to keep the puni codes because not all browsers understand decoded URLs.

Ambrevar:
Try the `follow-hint` and `copy-hint-url` commands :)

source/web-mode.lisp
jmercouris: OK, a good point!

ag91:
Ah! This is interesting. I originally thought this feature to create a QR code of the URL for the page you are viewing only.
Maybe I should add a second command that let you choose one of the link within the page via hints?

Ambrevar:
> Maybe I should add a second command that let you choose one of the link within the page via hints?

Yup, but don't worry about it for now, we can always address this
later.  I wrote this down so we don't forget about it.

source/web-mode.lisp
ag91: It worked! Although I am not sure how to avoid the `sleep`: Nyxt take a little to open a new buffer, just the right time to delete the target!

source/web-mode.lisp
```diff
@@ -695,6 +695,15 @@ ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
...
-        (uiop:run-program (list "nyxt" pathname)))))
+    (let* ((pathname (expand-path
+                     (make-instance 'data-path :basename "qrcode.png")))
+          (url (quri:render-uri (url (current-buffer))))
+          (title (str:concat "QRcode for " url)))
+      (unwind-protect
+           (progn
+             (cl-qrencode:encode-png url :fpath pathname)
+             (make-buffer-focus :url (quri::make-uri-file :path pathname))
+             (sleep 0.1) ;; TODO it seems I cannot open the buffer in time if I delete the file at the end
```
jmercouris, line 706: add a hook to the buffer to delete the file when it is closed instead

jmercouris: that way you won't have to contend with potentially deleting the file while the buffer is active, if the user presses refresh etc.

ag91: Great idea! That way I can also skip the `unwind-protect`!

ag91: Mmm, I am running the Guix 2.1.1 version of Nyxt and in  the Slime repl I cannot find `hooks:add-hook`, but I can find `serapeum:add-hook`. Still that does not seem to do anything when I kill the buffer. Do you know if I have to import something for using `hooks:add-hook` (it does not seem so looking at the code base).

jmercouris: You have to be in the Nyxt package, as you have guessed, it is the same as serapeum add-hook :-). Our work was merged into Serapeum.

jmercouris: If you look at our source/package.lisp you'll see the following;

```
(trivial-package-local-nicknames:add-package-local-nickname :hooks :serapeum/contrib/hooks :nyxt)
```

that way we can just say `hooks:add-hook` and it will mean `serapeum/contribs/hooks:add-hook`.

jmercouris: I should point out that lastly, you have an option of just not deleting the file. It will get overwritten every time you generate a new QR code, so perhaps it is not necessary to delete it?

ag91: I will probably go for no hooks then because I am still wrapping my head around them.
For example, I thought this should work:
```
NYXT> (serapeum/contrib/hooks:add-hook
         (buffer-delete-hook (current-buffer))
         (serapeum/contrib/hooks:make-handler-any
                 (lambda (buffer) 'example)
                 :name 'some-name)
         )
```
But instead I get:
```
  #<STANDARD-GENERIC-FUNCTION SERAPEUM:ADD-HOOK (14)>
when called with arguments
  (#<HOOK-BUFFER {10063A7CA3}>
   #<HOOKS:HANDLER-ANY {100BBF1883}>).
   [Condition of type SB-PCL::NO-APPLICABLE-METHOD-ERROR]
See also:
  Common Lisp Hyperspec, 7.6.6 [:section]
```
It seems that `buffer-delete-hook` is not recognized as a valid hook. How far I am from interpreting that error?

Ambrevar: Hooks are typed and only accept handlers (functions) with a specific signature.

In this case, `buffer-delete-hook` is of type `hook-buffer`, so it only accepts handlers made with `make-handler-buffer`.  Does that make sense?

ag91: Now it makes more sense! How comes that `hooks:make-handler-any` is not fine though? The "buffer" type is not seen as an "any"? Thanks for the explanation though, I am building up my knowledge!
I am not adding the hooks because they seem to suffer the same initialization problem: until the buffer is not created, I cannot attach the hooks (at least the hooks don't work when I create them in the function, but they do when I create them in a REPL for an existing buffer).


For reference, this works:
```
(let ((buffer (make-buffer-focus :url (quri::make-uri-file :path pathname))))
        (serapeum/contrib/hooks:add-hook
         (buffer-delete-hook buffer)
         (make-handler-buffer
          (lambda (buffer) (uiop:delete-file-if-exists pathname))
          :name 'delete-qr-code-file)
         )
        (serapeum/contrib/hooks:add-hook
         (buffer-loaded-hook buffer)
         (make-handler-buffer
          (lambda (buffer) (setf (title buffer) (str:concat "QrCode for " pathname)))
          :name 'rename-qr-code-buffer)
         ))
```

jmercouris:
Is this ready for a final review?

ag91:
I think I should still address those TODOs @Ambrevar suggested. I can look into the hint command, but I am pretty unsure of what is needed for Guix, I thought to have added the dependency on guix.scm..

jmercouris:
Don't worry, Pierre can take care of the Guix changes. Please let me know when you are ready for everything else :-).

ag91:
Mmm, since the hints thing seems to be broken on my Nyxt, could you just review this for now?
In the meanwhile I can explore better how to build from source, so later I can open a new PR for the hint command.

jmercouris:
The HINT command was not a TODO for you :-) it was for anybody.

jmercouris:
I will try this out on my machine and see if it is OK!

q3cpma:
Unless there's a misundersating from my part, shouldn't this kind of stuff be made optional at build time?

ag91:
> Unless there's a misundersating from my part, shouldn't this kind of stuff be made optional at build time?

@q3cpma  I need help at getting your comment: "stuff" as in dependencies or as in feature?

q3cpma:
Both. I mean features that are very situational and/or unused by almost everybody AND that adds a dependency.

ag91:
> Both. I mean features that are very situational and/or unused by almost everybody AND that adds a dependency.

Ah okay. Well, I would be happy to add this somewhere in Nyxt that is optional. I am thinking of a "contributed/extras" sort of space that users can decide to include. I don't know if that exists already though.
As you say, for now I was playing around with this kind of ideas here: https://github.com/ag91/emacs-with-nyxt

source/web-mode.lisp
Ambrevar: The type is for the _handler_, not the _handler argument_.  These are 2 different types:

- `(function (t))`
- `(function (buffer))`

Can you share the code that does not work for you?

Ambrevar:
@q3cpma We could very well make this optional.  But is it worth it?

```sh
$ guix size nyxt
...
total: 1728.4 MiB
```

```sh
$ guix size nyxt sbcl-cl-qrencode
...
total: 1729.6 MiB
```

It adds less than 0.07% to the size of Nyxt (the majority of which is due to WebKitGTK).  `cl-qrencode` is a very light dependency, there is practically no cost in including it I believe.

jmercouris:
I don't think qrlencode should be optional. That said, I /DO/ think standalone/non-core modes could have feature flags in the ASD. That is however part of a bigger mission that involves completely separating the modes from the core.

As it stands, I am basically OK to merge in this change and make some small tweaks.

Ambrevar:
We already have optional parts (e.g. Quicklisp and OSPM).  One drawback
is that it makes it harder for maintainers to know what packages the
user's Nyxt has loaded.

We should include this information in `--system-information`.

source/web-mode.lisp
ag91: Oh. I think I get it now. I will practice a little more with typed hooks :)

> Can you share the code that does not work for you?

Last week I was trying to make the following code to work. The idea was to rename the buffer and delete the temporary file.
```
(define-command-global show-qrcode-of-current-url (&optional (buffer (current-buffer)))
  "Show the QR code containing the URL for the current buffer."
  (with-current-buffer buffer
    (let* ((pathname (expand-path
                      (make-instance 'data-path :basename "qrcode.png")))
           (url (quri:render-uri (url (current-buffer)))))
      (cl-qrencode:encode-png url :fpath pathname)
      (let ((buffer (make-buffer-focus :url (quri::make-uri-file :path pathname))))
        (serapeum/contrib/hooks:add-hook
         (buffer-delete-hook buffer)
         (make-handler-buffer
          (lambda (buffer) (uiop:delete-file-if-exists pathname))
          :name 'delete-qr-code-file)
         )
        (serapeum/contrib/hooks:add-hook
         (buffer-loaded-hook buffer)
         (make-handler-buffer
          (lambda (buffer) (setf (title buffer) (str:concat "QrCode for " pathname)))
          :name 'rename-qr-code-buffer)
         )))))
```
Since it was a bit long and the two hooks weren't set (because I guess the buffer was not up and running yet), I didn't push this.
Anyway, I liked to learn about typed hooks! I shall write about that XD

anon00000000:
> @q3cpma We could very well make this optional. But is it worth it?
> 
> ```shell
> $ guix size nyxt
> ...
> total: 1728.4 MiB
> ```
> 
> ```shell
> $ guix size nyxt sbcl-cl-qrencode
> ...
> total: 1729.6 MiB
> ```
> 
> It adds less than 0.07% to the size of Nyxt (the majority of which is due to WebKitGTK). `cl-qrencode` is a very light dependency, there is practically no cost in including it I believe.

This is more by principle (to avoid the slippery slope that is node.js' NPM hell, for example) than the indeed negligible bloat of this specific feature, such a feature isn't needed at all for a normal browser operation. I am myself way more pained by webkit-gtk depending on Ruby to build.
It may also help when distributions try to package it and the various Lisp packages (e.g. Gentoo really doesn't like it when the network sandbox must be disabled).

Ambrevar:
anon00000000 ***@***.***> writes:

> This is more by principle (to avoid the slippery slope that is
> node.js' NPM hell, for example) than the indeed negligible bloat of
> this specific feature,

We are very wary of this danger, and we've got good provisions for it,
starting with a bootstrappable Guix package for Nyxt.  This gives us
strong guarantees that all our dependencies remain under control.

> such a feature isn't needed at all for a normal browser operation.

I would probably never use it myself, but I see that popular browsers
support this feature now, so I can understand why a sizeable portion of
users would like this feature by default.

> I am myself way more pained by webkit-gtk depending on Ruby to build.

Interestingly, Guix' webkitgtk package does not require ruby at all:

```sh
$ guix graph --path webkitgtk ruby
guix graph: error: no path from ***@***.***' to ***@***.***'
```

Which distribution do you use?

> It may also help when distributions try to package it and the various
> Lisp packages

As far as I know, only Guix provides Lisp packages for now.
All other distributions defer to our Quicklisp setup, so Lisp
dependencies come at zero cost for them.

> (e.g. Gentoo really doesn't like it when the network
> sandbox must be disabled).

Sorry, I don't understand how this is related to the network sandbox.
Can you give more details?

q3cpma:
On Mon, Jul 19, 2021 at 05:15:15AM -0700, Pierre Neidhardt wrote:
> anon00000000 ***@***.***> writes:
>
> > This is more by principle (to avoid the slippery slope that is
> > node.js' NPM hell, for example) than the indeed negligible bloat of
> > this specific feature,
>
> We are very wary of this danger, and we've got good provisions for it,
> starting with a bootstrappable Guix package for Nyxt.  This gives us
> strong guarantees that all our dependencies remain under control.
>
Thanks for giving attention to this. To be honest, I expect Lisp packages' average quality to be
magnitudes higher than node.

> > such a feature isn't needed at all for a normal browser operation.
>
> I would probably never use it myself, but I see that popular browsers
> support this feature now, so I can understand why a sizeable portion of
> users would like this feature by default.
>
Good point, didn't know it was widespread. I have no objections, then; even if as a Gentoo autist,
I can feel a bit dirty knowing I have a dependency useless to me =).

> > I am myself way more pained by webkit-gtk depending on Ruby to build.
>
> Interestingly, Guix' webkitgtk package does not require ruby at all:
>
> ```sh
> $ guix graph --path webkitgtk ruby
> guix graph: error: no path from ***@***.***' to ***@***.***'
> ```
>
> Which distribution do you use?
>
Gentoo. I see ruby in https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/webkit.scm in the
native-input section (seems equivalent of Gentoo's BDEPEND), too.

> > It may also help when distributions try to package it and the various
> > Lisp packages
>
> As far as I know, only Guix provides Lisp packages for now.
> All other distributions defer to our Quicklisp setup, so Lisp
> dependencies come at zero cost for them.
>
Indeed, it looks like Gentoo didn't care for Lisp the same way it did for Perl, Python,
Haskell, OCaml, etc...

> > (e.g. Gentoo really doesn't like it when the network
> > sandbox must be disabled).
>
> Sorry, I don't understand how this is related to the network sandbox.
> Can you give more details?
>
Wasn't very clear. I mean that Gentoo usually blocks network access during build time to ensure
that you get what's in the specified sources with matching checksum and nothing more.
Official Gentoo packages usually don't violate that rule; same for library bundling.

But since there are no Lisp packages on Gentoo, it's probably okay. At least, I can do a package in
my overlay.

(hope it won't get too weird going from Tcl to Lisp)

source/web-mode.lisp
Ambrevar: We've implemented typed hooks after learning the pain points of Emacs.  Typed Hooks are a bit harder to understand but much more practical in the end.  For the complete story, see https://github.com/ruricolist/serapeum/issues/41.

Ambrevar: If you dislike the long-winded `serapeum/contrib/hooks` package name, you can do something like this in your `init.lisp`

```lisp
(serapeum:eval-always
 (trivial-package-local-nicknames:add-package-local-nickname :hooks :serapeum/contrib/hooks :nyxt-user))
```

Ambrevar: You code seems correct.  To be clear, what does not work?  I suspect that the delete hook works as expected, but not the load hook, is that right?

If so, it's because `make-buffer-focus` automatically loads the given URL, so by the time the hook is modified, the buffer is already loaded and thus the hook is not run.

Instead, you can call `make-buffer` with `:load-url-p nil`, then focus and load manually _after_ modifying the hook.

This is arguably not a great interface.  We will overhaul `make-buffer` and friends at some point, probably for 3.0.

ag91: > If so, it's because `make-buffer-focus` automatically loads the given URL, so by the time the hook is modified, the buffer is already loaded and thus the hook is not run.
> 
> Instead, you can call `make-buffer` with `:load-url-p nil`, then focus and load manually _after_ modifying the hook.
> 

Indeed, that is the same explanation I gave myself by intuition. Both hooks fail to set because the buffer is just not there yet. I am sure there is a way around it, but I thought to keep the code simple. The functionality is very small and readable as is. It seems that if I add the hooks I am muddling a bit the waters. (If I should go that way, I would split things in utility functions to keep things readable.)

jmercouris:
@Ambrevar shall I merge this in? I don't think it matters whether it is packaged for Guix, it can be packaged when we do the 2.3.0 release.

Ambrevar:
It does because then I (and other guix.scm users) cannot build :)

Ambrevar:
But wait, what isn't packaged in Guix?
I don't think we talked about unpackaged Guix deps here, did we?

jmercouris:
I believe you wanted to package qrencode for Guix, is it already packaged or not? Can we merge this in?

Ambrevar:
You misunderstood: it's already packaged, I was merely talking about
updating the input list in the build-scripts/guix.scm file.

Yes, we can merged, I'll apply a few changes later :)

jmercouris:
Thank you @ag91 !!!

PR #1611, opened 2021-07-08
Replace tutorial and manual markup:markup with `who' markup.

Ambrevar:
It seems that markup:markup is poorly optimized and requires a high
dynamic-space-size just to expand and compile.
`who' does not seem to suffer from this issue.

@pdelfino @jmercouris Can you test and confirm it fixes #1600?

Issue #1600, opened 2021-07-06
Possible bug: Changes on `tutorial.lisp` with `command-docstring-first-sentence` crash Nyxt


pdelfino:
**Describe the bug**

After inserting changes on manual using `command-docstring-first-sentence` and recompiling the system, Nyxt crashes.
This happened while doing this [PR](https://github.com/atlas-engineer/nyxt/pull/1589).

**Precise recipe to reproduce the issue**
- be on this [branch](https://github.com/atlas-engineer/nyxt/pull/1589) or use version  Version 2.1.1-41-g643fe92a (commit 42946c6459e3)
- insert on `tutorial.lisp` the following:
```lisp
   (:h3 "Copying and Pasting")
   (:p "Unlike other web browsers, Nyxt provides powerful ways of copying
   and pasting content via different commands. Starting from:")
   (:ul
    (:li (command-markup 'nyxt/web-mode:copy) ": " (command-docstring-first-sentence 'nyxt/web-mode:copy))
    (:li (command-markup 'nyxt/web-mode:paste) ": " (command-docstring-first-sentence 'nyxt/web-mode:paste)))
   (:p "Passing through webpage's data:")
   (:ul
    (:li (command-markup 'copy-url) ": " (command-docstring-first-sentence 'copy-url))
    (:li (command-markup 'copy-title) ": " (command-docstring-first-sentence 'copy-title))
    (:li (command-markup 'nyxt/web-mode:copy-placeholder) ": " (command-docstring-first-sentence 'nyxt/web-mode:copy-placeholder))
    (:li (command-markup 'nyxt/web-mode:copy-hint-url) ": " (command-docstring-first-sentence 'nyxt/web-mode:copy-hint-url)))
   (:p "Leveraging password managers: ")
   #+nil
   (:ul 
    (:li (command-markup 'copy-username) ": " (command-docstring-first-sentence 'copy-username))
    (:li (command-markup 'copy-password) ": " (command-docstring-first-sentence 'copy-password))
    (:li (command-markup 'copy-password-prompt-details) ": " (command-docstring-first-sentence 'copy-password-prompt-details)))
   (:p "And more: ")
   #+nil 
   (:ul
    (:li (command-markup 'nyxt/web-mode:paste-from-clipboard-ring) ": " (command-docstring-first-sentence 'nyxt/web-mode:paste-from-clipboard-ring))
[...]

PR #1621, opened 2021-07-14
Markup to spinneret

Ambrevar:
Follow up to #1611 and addresses #1600.

This time I've used spinneret.
It seems to be as fast as `cl-who` but much more convenient.

I tried it on a complet markup generation like `list-bookmarks` and it seems to be even more  convenient than `cl-markup`.

All in all, conversion from cl-markup to Spinneret should be trivial.

@jmercouris Can you test if you can compile Pedro's documentation addition on this branch?


jmercouris:
I confirm the fix.

jmercouris:
I'll convert to spinneret, this will take a while I see 247 instances of `markup` in the codebase...

Ambrevar:
Most of them should work with query-replace.

Nested `markup:markup` can be removed, as in `list-bookmarks`.

jmercouris:
Wow, removing the nested `markup:markup` and having it "just work" is amazing.

Ambrevar:
Yup, spinneret is a net win!

Still until today, I struggled to nest markup:markup / markup:markup*
and markup:raw calls, now this class of complexity should be gone altogether!

jmercouris:
I did the easy ones, some of the hard ones remain... I keep running into stupid issues.

jmercouris:
All that remains is OSPM and Prompt buffer. I may do prompt buffer, but I leave OSPM to you, as I cannot test.

jmercouris:
There are also two usages of `eval` in the code I had to use, I don't like it, but could not think of a simpler way.

jmercouris:
Many were simple replacements, many were not.

source/bookmark.lisp
```diff
@@ -112,49 +112,50 @@ In particular, we ignore the protocol (e.g. HTTP or HTTPS does not matter)."
...
-                             :href (lisp-url `(nyxt::delete-bookmark ,url-href)) "Delete"))
-                     (:hr "")))))
-          (format nil "No bookmarks in ~s." (expand-path (bookmarks-path (current-buffer)))))))))
+      (:style (style bookmarks-buffer))
+      (:h1 "Bookmarks")
+      (:body
+       (or (with-data-unsafe (bookmarks (bookmarks-path (current-buffer)))
+             (loop for bookmark in bookmarks
+                   collect
+                      (let ((url-display (render-url (url bookmark)))
```
Ambrevar, line 121: Looks like we still have different indentation styles :)
What's your `lisp-indent-function`?

jmercouris:
```
lisp-indent-function is a variable defined in ‘lisp-mode.el’.
Its value is ‘common-lisp-indent-function’
Original value was 
‘lisp-indent-function’
Local in buffer start.lisp; global value is the same.

```

source/status.lisp
```diff
@@ -5,96 +5,87 @@
...
 (export-always 'format-status-load-status)
 (defun format-status-load-status (buffer)
-  (markup:markup
-   (:div :class (when (and (web-buffer-p buffer)
-                           (eq (slot-value buffer 'load-status) :loading))
-                  "loader") "")))
+  (spinneret:with-html-string
+    (:div :class (if (and (web-buffer-p buffer)
+                          (eq (slot-value buffer 'load-status) :loading))
+                     "loader" ""))))
```
Ambrevar, line 47: Wait, is this correct?  The way I understand the original code is that the div content is `""` (empty string), while the class is NIL if the condition is not met.

With this new change, the class can be `""` while the content is always empty.

source/command.lisp
```diff
@@ -179,7 +179,7 @@ Example:
                                            :name ',name
                                            :docstring ,documentation
                                            :fn (symbol-function ',name)
-                                           :sexp '(define-command (,@arglist) ,@body)
+                                           :sexp '(define-command ,name (,@arglist) ,@body)
```
Ambrevar, line 182: Unrelated to this patchset, right?

Probably a welcome change anyways, just wanted to know why you fixed it here in particular.

source/status.lisp
jmercouris: This is correct.

Ambrevar: I guess it's equivalent then, no?

source/command.lisp
jmercouris: Unrelated, was just something I noticed while testing :-)

source/bookmark.lisp
Ambrevar: So what about setting `lisp-indent-function` to a specific value in dir-locals.el?  Which one?
By the way, why did you set it to this value?

source/status.lisp
jmercouris: This code works as expected, the results are equivalent. Yes :-)

source/bookmark.lisp
jmercouris: I don't know why I set it to this value. Probably we differ because I have this in my config:

```
(setq lisp-indent-function 'common-lisp-indent-function)
(setq common-lisp-style-default "modern")
```

jmercouris: I suggest you set the dir locals to reflect my config :-D

Ambrevar: Where is `common-lisp-style-default` defined?

jmercouris:
To clarify: it would be nice if you could do the OSPM -> spinneret conversion because I cannot test it.

Ambrevar:
Understood.

Ambrevar:
I've found a way to get rid of the 2 `eval` calls.

source/bookmark.lisp
Ambrevar: I tried setting `(lisp-indent-function . common-lisp-indent-function)` in `.dir-locals.el`, but the variable is marked as risky, which means Emacs will prompt "Do you want to apply it? ..." on each file visit.

Any idea?

Otherwise, I suggest you remove your `lisp-indent-function` setting.  As far as I can tell, it mainly impacts `loop`, it should not change much beside this.

I also may be that `common-lisp-style-default` does nothing :p

Ambrevar:
As suggested in https://github.com/ruricolist/spinneret/issues/55#issuecomment-882767191, we may very just wrap the body of all `user-interface:object-expression` with `spinneret:with-html`, then object-string can simply call `spinneret:with-html-string` and we no longer need `eval`.

Is this a satisfying solution in your opinion, @jmercouris?

Ambrevar:
I'm done with OSPM.

Ambrevar:
I'm OK to merge this if we add a "TODO:" next to the user interface `eval`, or go as ruricolist suggested.

jmercouris:
> I'm OK to merge this if we add a "TODO:" next to the user interface `eval`, or go as ruricolist suggested.

OK, I will investigate and do either /or

jmercouris:
I don't like wrapping everything as ruricolist suggested, I may make a macro to deal with this macro... how many macros until we find the base turtle? :-D?

PR #1641, opened 2021-07-20
Show system information

Ambrevar:


jmercouris:
I'm not sure I see the benefit of this change. Can you please explain why the user would need to see their own system information?

Ambrevar:
They don't necessarily know them.  It's a centralized way to display all
kinds of information the user may be looking for, such as which
compiler, which kernel, which version of Nyxt, etc.

jmercouris:
OK!

Issue #1630, opened 2021-07-18
Drawing prompt takes a lot of time.


shamazmazum:
Hi! I tried nyxt on FreeBSD. Drawing prompt takes a lot of time (about 20 seconds). The interface hangs up for those 20 seconds and you cannot kill nyxt with ^C. It's enough to press `C-l` to reproduce the bug on my machine. By placing debugging messages I tracked the error to be somewhere between 
~~~~
(apply #'make-instance 'user-prompt-buffer
    (append args
                  (list
                   :window (current-window)
                   :result-channel result-channel
                   :interrupt-channel interrupt-channel)))
~~~~
in prompt-buffer.lisp and `(defmethod initialize-instance :after ((prompter prompter) &key) ...)` in prompter.lisp

Also, if you remove `prompter:prompter` from superclasses of `prompt-buffer`, the interface will not hang (but prompts will not work, of course).

Additional info:

* *OS*: FreeBSD 13.0
* *Video environment*: amdgpu driver from graphics/drm-kmod + mesa 20.2.3 + wayland (tried on X11 too)
* SBCL 2.1.6 from `lang/sbcl`
* Installed nyxt from ultralisp: `(ql:quickload :nyxt/gtk-application)` and `(asdf:make nyxt/gtk-application)`.
* webkit2-gtk3-2.30.5_2 from `www/webkit2-gtk3`

~~~~
vasily@vonbraun:~ % nyxt --system-information
Nyxt version: 2.1.1
Renderer version: GTK
Operating system kernel: FreeBSD 13.0-RELEASE-p1
Lisp implementation: SBCL 2.1.6 (Dynamic space size: 1073741824)
Features: (WEBKIT2 WEBKIT2-2.30 WEBKIT2-PASTE-PLAINTEXT WEBKIT2-TRACKING WEBKIT2-MUTE
 WEBKIT2-EMOJI WEBKIT2-MEDIA WEBKIT2-SANDBOXING GTK-3-22 GTK-3-20 GTK-3-18
 GTK-3-16 GTK-3-14 GTK-3-12 GTK-3-10 GTK-3-8 GTK-3-6 GTK-3-4 GTK GDK-3-22
 GDK-3-20 GDK-3-18 GDK-3-16 GDK-3-14 GDK-3-12 GDK-3-10 GDK-3-8 GDK-3-6 GDK-3-4
[...]

PR #1642, opened 2021-07-20
Replace UIOP:RUN-PROGRAM with SERAPEUM:RESOLVE-EXECUTABLE

shamazmazum:
UIOP:RUN-PROGRAM introduces a long delay on some platfroms
(e.g. on FreeBSD 13.0 with SBCL 2.1.6 built from lang/sbcl port).

This causes any new prompt buffer to take up to 30 seconds to show
up. SERAPEUM:RESOLVE-EXECUTABLE does not have such issues.

See #1630


Ambrevar:
Thanks a lot for this patch, and thanks a lot for keeping up with the investigation, it's not easy on my end when I can't reproduce locally :)

Issue #1590, opened 2021-07-05
Feature request: Prompt the user for history-entries, loop over them and bookmark those above some threshold


pdelfino:
This feature request is similar to #1310.

#1310 is cheap and should run every time a page is loaded.

This one is **expensive** and should run as **a command executed** by the user.

This emerged as an **accident** and **misunderstanding** while trying to solve #1310.

Awkwardly enough, this issue is born having an ongoing pull request #1548 

The algoritm is:

    Ask the user to define a sample space of selected history entries (can be all of them)
    Query the implicit-visit of selected history entries (high-cost)
    Compare with the threshold.
    If above threshold, not already bookmarked (single hash-table lookup), and 
    allowed by the user to be bookmarked, bookmark it.

jmercouris:
This feature has now been fulfilled! If you feel there is something missing, please feel free to reopen!

pdelfino:
@jmercouris, I am afraid this is slightly different from #1330.

What we have now: a mode (`bookmark-frequent-visit-mode`) that when enabled will check if the **current web page** is frequently visited. if so, it will ask the user if they want to bookmark it or not.

What we still need to do (with this issue): a **command** in which the user selecst a subset (or the whole set) of history entries. Then, the function loops through the implicit visits of all websites in the selected history entries checking which of them are beyond the threshold. For all of those above it, there will be a question if the user wants to bookmark it or not.

pdelfino:
Did I miss something?

jmercouris:
Ah yes, for some reason I thought we had merged that in!

PR #1649, opened 2021-07-22
 Accidentally useful feature: Bookmark frequently used pages looping over selected history entries

pdelfino:
This PR is a substitute for PR #1548. It is, hopefully, a solution for feature request #1590.

The code tries to implement this NL recipe:

```
Ask the user to define a sample space of selected history entries (can be all of them)
Query the implicit-visit of selected history entries (high-cost)
Compare with the threshold.
If above threshold, not already bookmarked (single hash-table lookup), and 
allowed by the user to be bookmarked, bookmark it.
```

It would be interesting to see this being tested by other people. It has worked for me in some tests I did. It is critical that the history file was built entirely **after** some fixes were made on implicit-visits counting.


Ambrevar:
What is NL? :p

source/browser.lisp
```diff
@@ -72,6 +72,10 @@ Most recent messages are first.")
 This history is used if no history is specified for a given prompt buffer.")
    (set-url-history (make-ring)
                     :documentation "The history of all URLs set via set-url")
+   (url->bookmark-visit-threshold :initform 20
```
Ambrevar, line 75: Nit: No need for `:initform`.  The 2nd argument is automatically understood as the initform.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
   "Return a new global history tree for `history-entry' data."
   (htree:make :key 'history-tree-key :current-owner-id (id buffer)))
 
+(defun bookmark-frequent-visit (history-entries-selected)
+  "Add frequently visited URLs that are not included in the bookmarks.  The
+   sample space of visited URLs is received as an argument and previously
+   defined by the user. "
```
Ambrevar, line 84: Style: Don't indent the docstring, or else they will display badly in Nyxt Help buffers.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
   "Return a new global history tree for `history-entry' data."
   (htree:make :key 'history-tree-key :current-owner-id (id buffer)))
 
+(defun bookmark-frequent-visit (history-entries-selected)
```
Ambrevar, line 81: Just `history-entries`.

source/history.lisp
Ambrevar: You can greatly simplify the docstring: "Bookmark frequently visited HISTORY-ENTRIES URLs.".

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+  "Add frequently visited URLs that are not included in the bookmarks.  The
+   sample space of visited URLs is received as an argument and previously
+   defined by the user. "
+  (labels ((urls-visited-over-threshold (threshold)
+             "The local function "`urls-visited-over-thresold" returns all URLs
```
Ambrevar, line 86: Syntax: remove the double quotes around `urls-visited-over-thresold`, add a single quote afterwards.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+             instances which were visited more times than the threshold."
+             (let* ((history-entries-above-threshold
+                      (remove-if-not #'(lambda (e) (> (implicit-visits e) threshold))
+                                     history-entries-selected)))
+               (mapcar #'(lambda (e) (url e)) history-entries-above-threshold)))
```
Ambrevar, line 91: This lambda is equivalent to `#'url`.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+           (bookmarked-url-p (url)
+             "The local function  "`bookmarked-url-p" returns the URL
```
Ambrevar, line 93: Same as above about the double quotes.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+            itself if it is new to the bookmark list and NIL if it is
+            already there."
+             (let ((bookmarks-address-list
+                     (mapcar #'(lambda (e) (url e))
```
Ambrevar, line 97: Same for the lambda.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+                             (with-data-unsafe (bookmarks (bookmarks-path (current-buffer)))
+                               bookmarks))))
```
Ambrevar, line 99: For such a simple statement, it's enough to just call `get-data` over the data path.
The result may be NIL, which is OK since we are `mapcar`-ing over it.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+             instances which were visited more times than the threshold."
+             (let* ((history-entries-above-threshold
+                      (remove-if-not #'(lambda (e) (> (implicit-visits e) threshold))
+                                     history-entries-selected)))
+               (mapcar #'(lambda (e) (url e)) history-entries-above-threshold)))
+           (bookmarked-url-p (url)
+             "The local function  "`bookmarked-url-p" returns the URL
+            itself if it is new to the bookmark list and NIL if it is
+            already there."
+             (let ((bookmarks-address-list
```
Ambrevar, line 96: Better name: `bookmark-urls`.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+                     (mapcar #'(lambda (e) (url e))
+                             (with-data-unsafe (bookmarks (bookmarks-path (current-buffer)))
+                               bookmarks))))
+               (if (member url bookmarks-address-list :test #'quri:uri=)
+                   nil
+                   url))))
```
Ambrevar, line 102: The `if` and its body are unnecessary, since `member` already does the logic you want.
Except... You are returning the opposite intention of "bookmarked-url-p", it return NIL when it's bookmarked :p

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+    (dolist (url (urls-visited-over-threshold (url->bookmark-visit-threshold *browser*)))
+      (let ((url-address (render-url url)))
+        (if (bookmarked-url-p url)
```
Ambrevar, line 105: As suggested above, here the logic is reversed: if should be `(unless (bookmarked-url-p ...))`.

source/history.lisp
Ambrevar: You should compute the bookmark URLs just once: let-bind them at the top of this function.  Them you can substitute the `bookmarked-url-p` call for the `member` call.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
   (htree:make :key 'history-tree-key :current-owner-id (id buffer)))
 
+(defun bookmark-frequent-visit (history-entries-selected)
+  "Add frequently visited URLs that are not included in the bookmarks.  The
+   sample space of visited URLs is received as an argument and previously
+   defined by the user. "
+  (labels ((urls-visited-over-threshold (threshold)
+             "The local function "`urls-visited-over-thresold" returns all URLs
+             instances which were visited more times than the threshold."
+             (let* ((history-entries-above-threshold
```
Ambrevar, line 88: Don't use `let*` if you can use `let`.

source/history.lisp
```diff
@@ -78,6 +78,43 @@ class."
...
+                   url))))
+    (dolist (url (urls-visited-over-threshold (url->bookmark-visit-threshold *browser*)))
+      (let ((url-address (render-url url)))
+        (if (bookmarked-url-p url)
+            (if-confirm ("Bookmark ~a?" url-address)
+                        (bookmark-url :url url-address)))))))
+
+(define-command bookmark-frequent-history-entries (&key (buffer (current-buffer)))
+  "Prompts the user for history entries (the user can select them all), and then
+call bookmark-frequent-visit over the selection."
```
Ambrevar, line 111: The second part is a bit too specific, it's not very user friendly to mention function names for high-level commands.

What about "and bookmark those that are frequently visited.  See also the `url->bookmark-visit-threshold' browser slot."

pdelfino:
> What is NL? :p

Sorry, I will avoid abbreviations from now on. It was the power of old habits. 

It means Natural Language (NL).

source/browser.lisp
jmercouris: This is where Pierre and I disagree, I am not a fan of the implicit magic of define-class!

source/history.lisp
pdelfino: I also need to fix **"\`**bookmarked-url-p**"** -> **\`**bookmarked-url-p**'**

source/history.lisp
pdelfino: You have a good point. I would just add:

  "Prompts the user for history entries (the user can select them all), and then
bookmark those that are frequently visited **but not already bookmarked.** See also the
`url->bookmark-visit-threshold' browser slot."

What do you say about this addition?

source/browser.lisp
Ambrevar: Nothing magic here, it's simple lambda list processing.

source/history.lisp
Ambrevar: From my perspective it does not add useful information, because re-bookmarking something already bookmarked does not change the bookmark in a meaningful way.  It's mostly an optimization detail.

Something else: we use the imperative tense in docstrings, so it's "Prompt" instead of "Prompts".

source/browser.lisp
jmercouris: OK sure, no magic, but magic :-)

pdelfino:
Following John's advice, I will wait until the problems reported on issue #1677 are fixed to work on this again.

jmercouris:
I'm closing this since the functionality has been now published/superseded by a more recent pull request :-)

Issue #1635, opened 2021-07-19
Some web application keybindings are captured by Nyxt instead of the web-buffer in `vi-insert-mode`


Kabouik:
**Describe the bug**
When on a webpage that supports keybindings, some are not properly captured by the web buffer, even when in `vi-insert-mode` (I would **not** expect them to be captured when in `vi-normal-mode`). 

For instance, with the mattermost web client, `<C-i>` or `<C-n>` which should be formatting keybindings instead call the Autofill and Bookmark functions in Nyxt, despite `vi-insert-mode` being active. `<C-u>` should open the "upload file" dialog, but does nothing (it did work once when I pressed it twice in a row, but I couldn't reproduce it). 

Some of them are captured however, immediately like `/` to focus the search field on Github, or when pressed twice like `<C-u>` on Discord to toggle the user list.

Those keybindings work fine in Epiphany. I have not tried with CUA or Emacs mode but wouldn't be surprised if this was a consequence of the vi modal system.

I understand that this is a difficult thing to implement without compromises. For instance, I really like that I can still switch buffers with `<C-[>` and `<C-]>` without opening the prompt-buffer or without leaving `vi-insert-mode` when I am composing, and I am not sure how this could still work if all keybindings were interpreted in priority by the web-buffer when in insert mode. Similarly, it would become impossible to call the promp buffer when in `vi-insert-mode`, while this can be done currently with `<C-space>`; one would have to hit `<Esc>` and then `:` or `<C-space>`.

**Information**
- OS name+version: Solus (rolling), Debian Sid (rolling)
- Graphics card and driver: Intel, Nvidia MX150 with nvidia-glx-driver, or none
- Desktop environment / Window manager name+version: i3wm
- How you installed Nyxt (Guix pack, package manager, build from source): built from `master`
- Information from command copy-system-information:

```
Nyxt version: 2.1.1-227-g85bbe1c2
Renderer version: GI-GTK
Operating system kernel: Linux 5.13.2-188.current
Lisp implementation: SBCL 2.1.2 (Dynamic space size: 1073741824)
Features: (WEBKIT2 WEBKIT2-2.32 WEBKIT2-PASTE-PLAINTEXT WEBKIT2-TRACKING WEBKIT2-MUTE
 WEBKIT2-EMOJI WEBKIT2-MEDIA WEBKIT2-SANDBOXING GTK-3-22 GTK-3-20 GTK-3-18
[...]

PR #1652, opened 2021-07-22
vi-insert-mode and application mode

Ambrevar:
This addresses the main issue in https://github.com/atlas-engineer/nyxt/issues/1635, that vi-insert-mode should have a convenient way to switch to application-mode back and forth.

It also binds application-mode to C-z everywhere, including in application-mode itself to disable it.

Questions:

- Should we bind C-z in base-mode or web-mode?
- I'm realizaing that application-mode should have an indicator just like vi-*-mode.  Maybe rename and generalize the vi indicator so that it displays there as well?  For instance, when both application mode and vi-insert-mode are enabled, it would display `A I`.

### DIFF
--- a/source/application-mode.lisp
+++ b/source/application-mode.lisp
@@ -3,6 +3,7 @@
 
 (uiop:define-package :nyxt/application-mode
   (:use :common-lisp :nyxt)
+  (:import-from #:keymap #:define-key #:define-scheme)
   (:documentation "Forward all keybindings to the web view except those in the `override-map'."))
 (in-package :nyxt/application-mode)
 
@@ -14,11 +15,18 @@
 ;;; the keymaps of all modes (except the override-map).
 
 (define-mode application-mode ()
-  "Mode that forwards all keys to the renderer."
-  ((destructor
+  "Mode that forwards all keys to the renderer.
+See the mode `keymap-scheme' for special bindings."
+  ((keymap-scheme
+    (define-scheme "application"
+      scheme:cua
+      (list
+       "C-z" 'application-mode)))
+   (destructor
     (lambda (mode)
       (hooks:remove-hook (current-keymaps-hook (buffer mode))
-                         'keep-override-map)))
+                         'keep-override-map)
+      (echo "Application-mode disabled.")))
    (constructor
     (lambda (mode)
       (if (current-keymaps-hook (buffer mode))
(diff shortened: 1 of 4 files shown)


jmercouris:
I would bind "C-z" in `web-mode` because it is a binding for `web-applications`. If we are to call it `passthrough-mode`, I would bind it in `base-mode`.

jmercouris:
As per utilizing the vi mode indicator, I believe this is a bad idea. Suddenly we have two indicator areas. One on the far right, and one on the far left. Which is used for what, and when?

If anything we should eliminate the status indicator on the left and allow the status indicator on the right to reorder/emphasize things to increase its visibility.

Thoughts?

Ambrevar:
> If anything we should eliminate the status indicator on the left and allow the status indicator on the right to reorder/emphasize things to increase its visibility.

Absolutely.  We could split the mode listing in 2 parts: the modes the
must display, and the modes than can be hidden behind an ellipsis.

jmercouris:
How do we know if a mode should be prioritized for visibility?

1. A slot in the mode?
2. A slot in the status-buffer class?

I am leaning towards option 2.

jmercouris:
Any other ideas?

Ambrevar:
The benefit of storing the information in the mode is that modes
themselves can provide the information, it composes better because the
status bar cannot know the list of all possible modes in advance.

Thus I suggest a mix of both:

- A mode slot that gives a "priority suggestion".
- A status bar slot that places the modes as it wants, possibly
  leveraging the information in this "priority suggestion".

Does that make sense?

jmercouris:
Seems overly complex. I would suggest either/or rather than a hybrid model. I will make a PR with it being a slot of the mode and then we can investigate more!

Ambrevar:
Anyways, our status bar logic needs an overhaul, so it can wait.
The mode slot is certainly enough until we've figured it out.

jmercouris:
I think our status bar logic is quite good, what would you change about it?

Ambrevar:
It does not compose well and the CSS is too hard-coded.

jmercouris:
What do you mean it doesn't compose well? With what would it compose?

Ambrevar:
The various elements the compose the status bar.

For instance, if I want to remove the "tabs", I need to rewrite the CSS.

It shows well with how we had to add the VI indicator: we had to modify
multiple places.  At the moment, modes are unable to customize how they
appear in the status area.

jmercouris:
OK, fair enough. To achieve what you want we would need to generate the CSS dynamically. I'm not sure that it is worth the effort, but it can be done.

Ambrevar:
It's not urgent for sure.  Would be nice to have it for 3.0 for instance.

Ambrevar:
I'll open a new issue about this.

Anyways, back to this pull request.  OK to merge?

jmercouris:
Merci Pierre.

PR #1653, opened 2021-07-23
buffer: Don't update the last-access when previewing buffers.

### LINKED ISSUE #1491: Do-what-I-mean follow-mode in switch-buffer
In commit b3dcf8400d92d1fe12eb91043981b52d5bc32201 @jmercouris changed the behaviour of `switch-buffer` to not move the last buffer to the top.

While this is convenient when follow-mode is on, it removes the ability to easily "alt-tab" between the last two buffers.

@jmercouris Can you remind me of the rationale behind not shuffling buffers?

Ambrevar:
Also add option to show current buffer last in switch-buffer.
This makes it easier to define a custom command to not show the current buffer last.

Fixes #1491.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -942,7 +942,7 @@ proceeding."
 (define-user-class buffer-source)
 
 (-> switch-buffer (&key (:id string) (:current-is-last-p boolean)) *)
-(define-command switch-buffer (&key id (current-is-last-p t))
+(define-command switch-buffer (&key id (current-is-last-p nil))
   "Switch the active buffer in the current window.
 Buffers are ordered by last access.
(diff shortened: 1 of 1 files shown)


Ambrevar:
@jmercouris About your rationale:

1. Is "fixed" by allowing the user to define a custom commands to choose whether the current buffer should be prompter or not.  I'm fine with swapping the defaults.
2. Is fixed by not updating `last-access` on preview.

Ambrevar:
OK to merge?

jmercouris:
99% of the way there :-). I don't like how "current-is-last" got changed :-O

It really makes me forget exactly what I was doing every time. I think the default should be nil.

Ambrevar:
As I said, I don't mind setting the default to nil :)

PR #1655, opened 2021-07-23
web-mode: Use in-memory stream instead of intermediary file to displa…

Ambrevar:
…y QRcode.

@ag91 This removes a whole class of problem by not writing a file :)
Also note the use of the `with-current-html-buffer` helper, makes everything easier :)

I'm introducing new deps, namely cl-base64 and flexi-streams, but both of them are already indirect dependencies so it comes for free.

That said, if anyone has better suggestions for byte (octet) streams and base64 encoding, please share!

### DIFF
--- a/build-scripts/guix.scm
+++ b/build-scripts/guix.scm
@@ -150,6 +150,7 @@
     (inputs
      `(("alexandria" ,cl-alexandria)
        ("bordeaux-threads" ,cl-bordeaux-threads)
+       ("cl-base64" ,cl-base64)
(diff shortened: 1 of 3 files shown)


jmercouris:
Very nice, thank you :-)

ag91:
> @ag91 This removes a whole class of problems by not writing to a file :)

This is fantastic @Ambrevar ! I think I have made a new synapsis just now to connect that Nyxt treats HTML as first class presentation format and that file streams can become easily HTML. Very smart!

PR #1656, opened 2021-07-23
element-frame.lisp: poll the page instead of requiring user action

jmercouris:
I want the element frame results to update in the prompt buffer AS the user is dragging their selection. In order to achieve this I need to call `prompter::update` on some loop to invoke my `prompter:filter-preprocessor` which fetches the user selection.

Unfortunately, this does not work. I have to then press ANY key (not necessarily even in the prompt buffer), and then the sources update.

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -342,7 +342,7 @@ interface.")
    (update-notifier (make-channel)
                     :type calispel:channel
                     :documentation "A channel which is written to when `filter'
-commits a change to `suggetsions'.  A notification is only send if at least
+commits a change to `suggestions'.  A notification is only sent if at least
 `notification-delay' has passed.  This is useful so that clients don't have to
 poll `suggestions' for changes.")
 
--- a/source/element-frame.lisp
+++ b/source/element-frame.lisp
@@ -153,44 +153,36 @@
     (remove-overlay)))
 
 (define-class frame-source (prompter:source)
-  ((prompter:name "Selection Frame (confirm with C-s)")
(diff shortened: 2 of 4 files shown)


jmercouris:
@Ambrevar Do you know how I should approach this?

jmercouris:
Well, I did get a successful update to work doing:

```
(prompter::update-sources (current-prompt-buffer)) (run-thread (update-prompt-input (current-prompt-buffer)))
```

in a thread. Unfortunately, after doing so, the user couldn't scroll down because the results kept instantly updating. I think we'll need to fix the key dispatching per buffer before I can make a reasonable fix for the way that element-frame works.

jmercouris:
Closing for now.

libraries/prompter/prompter-source.lisp
```diff
@@ -342,7 +342,7 @@ interface.")
    (update-notifier (make-channel)
                     :type calispel:channel
                     :documentation "A channel which is written to when `filter'
-commits a change to `suggetsions'.  A notification is only send if at least
+commits a change to `suggestions'.  A notification is only sent if at least
```
Ambrevar, line 345: We could merge this :)

Ambrevar:
```
(prompter::update-sources (current-prompt-buffer))
```
is correct.

However don't call `update-prompt-input` on something that gets updated in parallel because it updates _until_ there is no more changes.

If you just want to redraw the prompt, call `(prompt-render-suggestions prompt-buffer)`.

Ambrevar:
Or `prompt-render` to redraw the entire prompt and not just the suggestions.

PR #1662, opened 2021-07-24
GTK: Use a single data-manager cache by default, instead of one per b…

### LINKED ISSUE #1645: Delete data-manager cache on buffer deletion?
If you look into ~/.cache/nyxt, you'll see a bunch of `nyxt-data-manager-N` directories, with `N` being a buffer ID.

It seems that we are never deleting these directories and they tend to grow rather big (the sum in the ballpark of hundreds of MiB).

Something else: when creating a buffer with an ID that matches an existing directory, we tell WebKitGTK to reuse this directory.  Is this the right thing to do?

We could delete the cache on buffer deletion.  Would we lose some cache optimization then?

Ambrevar:
…uffer.

This might fix the issue in which only some buffers when logged in to some web
sites, while others were not.

Fixes #1645.

### DIFF
--- a/source/renderer-gtk.lisp
+++ b/source/renderer-gtk.lisp
@@ -64,8 +64,7 @@ want to change the behaviour of modifiers, for instance swap 'control' and
   ((gtk-object)
    (gtk-proxy-url (quri:uri ""))
    (proxy-ignored-hosts '())
-   (data-manager-path (make-instance 'data-manager-data-path
-                                     :dirname (uiop:xdg-cache-home +data-root+))
+   (data-manager-path (make-instance 'data-manager-data-path)
                       :documentation "Directory in which the WebKitGTK
 data-manager will store the data separately for each buffer.")
    (gtk-extensions-path (make-instance 'gtk-extensions-data-path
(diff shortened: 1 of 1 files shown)


Ambrevar:
@aartaka Can you test and tell me if it still allows for all the per-buffer customization you had in mind?

Ambrevar:
While working on this I realized that we never set the `dirname` of the various data-path classes, we only set it at instantiation time.  I wonder why that is, any idea?

I think it's a mistake, because if we set the dirname in the class definition it would allow for inherited classes to get the same default if needed.

It's also more verbose to repeat the `:dirname ...` part on instantiation.

aartaka:
> While working on this I realized that we never set the `dirname` of the various data-path classes, we only set it at instantiation time. I wonder why that is, any idea?

Hysterical reasons, I guess ¯\_(ツ)_/¯
 
> I think it's a mistake, because if we set the dirname in the class definition it would allow for inherited classes to get the same default if needed.
> 
> It's also more verbose to repeat the `:dirname ...` part on instantiation.

Agreed here.

aartaka:
I can only test it tomorrow. It looks totally fine, though.

Ambrevar:
Let me know if it's OK to merge :)

aartaka:
It works perfectly :)

aartaka:
OK to merge.

Ambrevar:
Thanks for your feedback!

jmercouris:
Thank you!

PR #1663, opened 2021-07-24
Add support for GTK clipboard

shamazmazum:
Hi! I would like to add support for clipboard in GTK interface without help of any external programs. I googled and ended up with this solution. The most stupid thing about this solution is that `gtk-clipboard-request-text` does not have any timeout which specifies how long it waits for text in the clipboard, so I added a timeout to `calispel:?` to finish execution of the caller thread if there is no text in the buffer.

Can this be discussed and eventually merged?

### DIFF
--- a/source/clipboard.lisp
+++ b/source/clipboard.lisp
@@ -4,12 +4,24 @@
 (in-package :nyxt)
 
 (-> ring-insert-clipboard (containers:ring-buffer-reverse) string)
+
+(export-always 'clipboard-text)
+(defgeneric clipboard-text (browser)
+  (:documentation "Get text currently residing in BROWSER's clipboard")
+  (:method ((browser browser))
+    (ignore-errors (trivial-clipboard:text))))
+
+(defgeneric (setf clipboard-text) (text browser)
+  (:documentation "Set content of BROWSER's clipboard to TEXT")
+  (:method (text (browser browser))
+    (trivial-clipboard:text text)))
+
 (export-always 'ring-insert-clipboard)
 (defun ring-insert-clipboard (ring)
   "Check if clipboard-content is most recent entry in RING.
 If not, insert clipboard-content into RING.
 Return most recent entry in RING."
-  (let ((clipboard-content (ignore-errors (trivial-clipboard:text))))
+  (let ((clipboard-content (clipboard-text *browser*)))
     (when clipboard-content
       (unless (string= clipboard-content (unless (containers:empty-p ring)
                                            (containers:first-item ring)))
@@ -19,4 +31,6 @@ Return most recent entry in RING."
 (export-always 'copy-to-clipboard)
 (defun copy-to-clipboard (input)
   "Save INPUT text to clipboard, and ring."
-  (containers:insert-item (clipboard-ring *browser*) (trivial-clipboard:text input)))
+  (containers:insert-item
+   (clipboard-ring *browser*)
+   (setf (clipboard-text *browser*) input)))
--- a/source/renderer-gi-gtk.lisp
+++ b/source/renderer-gi-gtk.lisp
@@ -13,6 +13,18 @@
 
 (in-package :nyxt)
 
+(defmacro with-namespaces (bindings &body body)
+  "Bind symbols to GIR namespaces. BINDINGS must be a list with
+elements in the form (SYMBOL NAMESPACE-NAME)."
+  (let ((binding-forms
+          (loop for binding in bindings
+                collect
+                (destructuring-bind (symbol namespace-name)
+                    binding
+                  `(,symbol (gir:require-namespace ,namespace-name))))))
+    `(let ,binding-forms
+       ,@body)))
+
 (setf +renderer+ "GI-GTK")
 (handler-bind ((warning #'muffle-warning))
   (let ((renderer-thread-name "Nyxt renderer thread"))
@@ -44,3 +56,16 @@ interface. On Darwin, we must run the GTK thread on the main thread."
     (define-ffi-method ffi-kill-browser ((browser gtk-browser))
       (unless *run-from-repl-p*
         (gir:invoke ((gir:ffi "Gtk" "3.0") 'main-quit))))))
+
+;; Working with clipboard
+(define-ffi-method (setf clipboard-text) (text (gtk-browser gtk-browser))
+  (with-namespaces ((gtk "Gtk") (gdk "Gdk"))
+    (let* ((atom (gir:invoke (gdk "atom_intern") "CLIPBOARD" nil))
+           (clipboard (gir:invoke (gtk "Clipboard" 'get) atom)))
+      (gir:invoke (clipboard 'set_text) text -1))))
+
+(define-ffi-method clipboard-text ((gtk-browser gtk-browser))
+  (with-namespaces ((gtk "Gtk") (gdk "Gdk"))
+    (let* ((atom (gir:invoke (gdk "atom_intern") "CLIPBOARD" nil))
+           (clipboard (gir:invoke (gtk "Clipboard" 'get) atom)))
+      (gir:invoke (clipboard 'wait_for_text)))))


source/renderer-gtk.lisp
```diff
@@ -1284,3 +1284,19 @@ As a second value, return the current buffer index starting from 0."
...
+
+(defmethod clipboard-text ((gtk-browser gtk-browser))
+  (let ((channel (make-channel 1)))
+    (within-gtk-thread
+      (gtk:gtk-clipboard-request-text
+       (gtk:gtk-clipboard-get "CLIPBOARD")
+       (lambda (clipboard text)
+         (declare (ignore clipboard))
+         (calispel:! channel text))))
+    (calispel:? channel 1)))
```
Ambrevar, line 1302: Waiting 1 sec seems wrong, it might introduce a delay in the caller.

Maybe you could use https://developer.gnome.org/gtk3/stable/gtk3-Clipboards.html#gtk-clipboard-wait-for-text instead?  

You could also use 2 channels, one to synchronize the Lisp calling thread with the GTK thread, and the other one to pass the clipboard data and waiting 0 second.

Ambrevar:
This is a great contribution, thanks!

We can also bind `C-c` to insert each copy to the Nyxt clipboard.  It would not work when application-mode is enabled though.

Maybe we can do better.  What about this: in `renderer-gtk`'s `ffi-initialize` method we call https://developer.gnome.org/gtk3/stable/gtk3-Clipboards.html#gtk-clipboard-wait-for-text with a callback which first copies the text to the Nyxt clipboard, then calls gtk-clipboard-wait-for-text again.  This would result in an infinite loop of transfering the GTK copies to the Nyxt clipboard, thus effectively acting as a hook over the GTK clipboard.

Would that work?

shamazmazum:
I think I can just use `gtk_clipboard_wait_for_text` as it simply returns `NULL` if there is no text in the clipboard (I thought it would block, but no). Unfortunately, there is no `gtk-clipboard-wait-for-text` in `cl-cffi-gtk`. Should I contact developers of `cl-cffi-gtk` first to add it?

We need something like this:
~~~~
(defcfun ("gtk_clipboard_wait_for_text" gtk-clipboard-wait-for-text) :string
  (clipboard (g-object gtk-clipboard)))

(export 'gtk-clipboard-wait-for-text)
~~~~

When `clipboard-text` can be changed to 
~~~~
(define-ffi-method clipboard-text ((gtk-browser gtk-browser))
  (gtk:gtk-clipboard-wait-for-text
   (gtk:gtk-clipboard-get "CLIPBOARD")))
~~~~

jmercouris:
If you are to contact the developers of cl-cffi-gtk, please note that we are using this branch: https://github.com/Ferada/cl-cffi-gtk

which unfortunately appears to have been archived... I guess we will have to update our code to comply with: https://github.com/crategus/cl-cffi-gtk

anyways, good luck getting anything merged into: https://github.com/crategus/cl-cffi-gtk , it is more or less impossible.

jmercouris:
Maybe we should make our own fork of cl-cffi-gtk.

shamazmazum:
I updated my branch adding a file which adds needed gtk functionality. It seems like `crategus/cl-cffi-gtk` also has it, but quicklisp still points to old `Ferada/cl-cffi-gtk`. I do not know a solution to this situation. How to add needed functionality to an archived repo? :)

For me there are three choices:

* Add missing gtk functionality to nyxt sources directly (as I did just now)
* Fork `cl-cffi-gtk` as @jmercouris suggested (but what with installing with quicklisp? Should the fork has a different name?)
* Use `gtk-clipboard-request-text` instead.

I have a question to @Ambrevar regarding the third choice. You mentioned:
> You could also use 2 channels, one to synchronize the Lisp calling thread with the GTK thread, and the other one to pass the clipboard data and waiting 0 second.

Can you explain this? You still have to wait for a message on that channel (a separate thread can be created for this, but this seems complicated).

For me using `gtk_clipboard_wait_for_text` seems the most simple (and right) decision, I just need to put a binding somewhere

Ambrevar:
In the medium (short?) term, we would like to get rid of cl-cffi-gtk and use
cl-gobject-introspection.
As such, spending more time on cl-cffi-gtk might not be worth the effort.

@shamazmazum: There is an easy way around the lack of binding in our
cl-cffi-gtk: don't use it and leverage cl-gobject-introspection instead :)
You can see a tiny use of it in renderer-gi-gtk.

The code would go to `renderer-gi-gtk.lisp` and the old `gtk` renderer would not
have it, but that's OK since it's not the default.

What do you think?

jmercouris:
@Ambrevar I forgot to mention or perhaps I did and forgot. It is not possible to make a complete renderer using only GI, one has to employ some CFFI due to limitations in the typelibs produced by GTK.

Ambrevar:
Can you expand?
I did manage to make a simpmle GTK browser using
cl-gobject-instrospection, I didn't see any limitation there.

jmercouris:
There are limitations. I cannot expand specifically because it was many months ago-- I did confirm it though. It is impossible.

There are some functions that cannot be invoked by GI that require CFFI bindings. This is because these functions are not tied to a specific object (they are not methods). We need some of these functions for the operation of Nyxt. Therefore, a pure GI renderer is not possible.

Ambrevar:
I'd like to see with a concrete example.

jmercouris:
Let me just fire up my time machine :-D

In any case, we will run into the limitation again when the time comes. It is no big deal, we'll just have to make a few CFFI stubs.

Ambrevar:
Oh, I misunderstood you, I though you meant we had to use cl-cffi-gtk.

We depend on CFFI anyways (cl-gobject-instrospection does!), so no big deal.

shamazmazum:
I cannot do this with `cl-gobject-introspection`. Either I do not understand it or its functionality is restricted somehow:

~~~~
(gir:nget (gir:require-namespase "Gdk") "Atom" 'intern)
~~~~

signals SIMPLE-ERROR. I need it to get clipboard descriptor (GdkAtom)

Ambrevar:
"gdk_atom_intern" is the function, there is no "intern" method for the
"GdkAtom" object (I know, it's confusing, but hey, it's C :p).

So this should work:

```lisp
(gir:invoke ((gir:require-namespace "Gdk") "atom_intern") "foo_name" nil)
```

or

```lisp
(gir:invoke ((gir:require-namespace "Gdk") 'atom-intern) "foo_name" nil)
```

Ambrevar:
For reference:

https://developer.gnome.org/gdk3/stable/gdk3-Properties-and-Atoms.html#gdk-atom-intern

shamazmazum:
I think I finally did it via `cl-gobject-introspection`. All works for me. Can you confirm?

Ambrevar:
The Nyxt binary crashes for me when I call copy-url.
I see this in the shell output:

```
(nyxt:30564): GLib-GObject-WARNING **: 10:01:24.084: cannot register existing type 'GtkClipboard'

(nyxt:30564): GLib-CRITICAL **: 10:01:24.084: g_once_init_leave: assertion 'result != 0' failed

(nyxt:30564): GLib-GObject-CRITICAL **: 10:01:24.084: g_object_new_with_properties: assertion 'G_TYPE_IS_OBJECT (object_type)' failed
CORRUPTION WARNING in SBCL pid 30564 tid 30576:
```

Ambrevar:
From a REPL, I get this stacktrace:

``` common-lisp
Unhandled memory fault at #x18.
   [Condition of type SB-SYS:MEMORY-FAULT-ERROR]

Restarts:
 0: [ABORT] abort thread (#<THREAD "Nyxt renderer thread" RUNNING {100F3B8003}>)

Backtrace:
 0: (SB-SYS:MEMORY-FAULT-ERROR #<unused argument> #.(SB-SYS:INT-SAP #X00000018))
 1: ("foreign function: call_into_lisp")
 2: ("foreign function: funcall2")
 3: ("foreign function: handle_memory_fault_emulation_trap")
 4: ("foreign function: handle_trap")
 5: ("foreign function: #x41925E")
 6: ((LAMBDA (&REST GIR::ARGS-IN) :IN GIR::BUILD-FUNCTION) #<GIR::STRUCT-INSTANCE {10077C0263}>)
 7: ((LAMBDA NIL :IN NYXT:CLIPBOARD-TEXT))
[...]
```

Ambrevar:
What's your OS, GTK and gojbect-introspection version?

shamazmazum:
Weird. Do you see where the error could be?

* OS: FreeBSD 13.0
* gtk: gtk3-3.24.29
* gobject-introspection: gobject-introspection-1.66.1

Ambrevar:
Mine:

- OS: Guix 99ca1ae
- gtk: 3.24.24
- gobject-introspection: 1.62.0

No idea what the problem is, maybe it fails on empty clipboards or something?

Ambrevar:
Maybe we can merge the original patch with cl-cffi-gtk in the mean time,
and add a TODO comment explaining what should be done.

I suggest you open a second pull request so that we can compare both patches.

shamazmazum:
Did this #1680

Issue #1660, opened 2021-07-23
Naming: use hyphenating convention in command names pre-fixed with `no`, e.g.: changing from `nosound` to `no-sound`


pdelfino:
Currently, Nyxt has as command names:
- `nowebgl-mode`;
- `nosound-mode`;
- `noimage-mode`;
- `noscript-mode`;

Maybe, I missed something valuable about this approach, but it seems to be inconsistent with the Common Lisp conventions on [naming](https://www.cliki.net/naming+conventions) and inconsistent with everything else in Nyxt. 

Even the predicates in Nyxt's source code are written as `foo-p` although in classical books they use `foop`. [1] 

Thus, I would suggest using **hyphenating**:

- `nowebgl-mode` -> `no-webgl-mode`;
- `nosound-mode` -> `no-sound-mode`;
- `noimage-mode` -> `no-image-mode`;
- `noscript-mode` -> `no-script-mode`;

I would be glad to send a patch. The case of `noop` is special and treated [here](https://github.com/atlas-engineer/nyxt/issues/1659).

[1] - By the way, I think `foo-p` is better :smile: . hyphenating rocks :guitar:

Ambrevar:
Fair to me.

@jmercouris @aartaka?

Ambrevar:
Issues #1660 #1659 and #1658 could have been just one issue ;)  It's also helpful to gather things that belong together and not clutter the issue tracker too much :)

aartaka:
I prefer the no-hyphen version more here. It looks and reads shorter without the pause hyphen introduces. In case with `noscript` and `nosound` it's even a valid piece of language.

Ambrevar:
You mean "nosound" is in the English dictionary?

Can't find it :p
https://en.wiktionary.org/wiki/Special:Search?search=nosound&ns0=1

jmercouris:
I am ambivalent, I don't think the success of Nyxt hinges upon the presence of hyphens. :-D

pdelfino:
> Issues #1660 #1659 and #1658 could have been just one issue ;) It's also helpful to gather things that belong together and not clutter the issue tracker too much :)

Interesting. It was actually harder for me to split things. It would have been easier for me to write just one.

I was reading about good practices in FOSS and the recommendation was to try splitting things into atomic/independent and small changes.

It turns out, you guys said kind of a "yes" for all of suggestions. But since they are different, you could have said "no" to one or two without affecting the other. 

[...]

PR #1665, opened 2021-07-24
Use hyphenating in commands pre-fixed with "no", such as changing from `nosound` to `no-sound`

pdelfino:
Hi guys,

This PR address issue #1660. As a *general instruction*, John asked me to insert the name of the files being changed in commit messages. So, I did it.

I did 1 commit for each file changed. Sometimes, one file had something like 2, 3 or 4 lines changed due to the renaming.

I also did 1 commit for each `magit-file-rename` which means 4 commits. 

I know how to squash but I was just unsure if I should squash these commits. It would be nice to receive suggestions if these renaming should be in master history or not.

Since these changes could cause problems, I tested `no-sound`, `no-image`, `no-script`, and `no-webgl`. 
I did not break anything. Everything is working :wink:


documents/CHANGELOG.org
```diff
@@ -539,7 +539,7 @@ If possible!
 
   Help to implement additional backends is welcome!
 
-- New =nowebgl-mode=.
+- New =no-webgl-mode=.
```
Ambrevar, line 542: Don't update the ChangeLog previous entries, since older entries refer to older code.

source/dom.lisp
```diff
@@ -65,7 +65,7 @@
   img-element input-element ins-element kbd-element label-element legend-element
   (li-element list-element) link-element (main-element semantic-element) map-element
   (mark-element semantic-element) meta-element meter-element (nav-element semantic-element)
-  noscript-element object-element (ol-element list-element) optgroup-element
+  no-script-element object-element (ol-element list-element) optgroup-element
```
Ambrevar, line 68: Not sure this is correct.  @aartaka ?

aartaka: It's not! HTML is standardized, and we're borrowing from HTML here.

pdelfino:
OK. The changes on CHANGE.log and `dom.lisp` were reverted and then I squashed them. I think it is good to go! :wink:

jmercouris:
Thank you!

Ambrevar:
Wait, this should not have been merged, it breaks backward compatibility!

I suggest to revert it, we dont' have a good way to deal with deprecated modes.  We can keep this for 3.0, it's a trivial, non-urgent change anyways.

jmercouris:
Right, I forgot about that! I almost think it would be easier at this point to merge everything to the 2 series branch and work on that (except for this). What do you think?

jmercouris:
I can also revert it of course, I let me see if I could turn it into 1 revert commit somehow

Ambrevar:
> Right, I forgot about that! I almost think it would be easier at this point to
> merge everything to the 2 series branch and work on that (except for this). What
> do you think?

What do you mean with "everything" and "that"? :p

Ambrevar:
For now I suggest we revert as I'd like to prepare 2.2.

jmercouris:
> > Right, I forgot about that! I almost think it would be easier at this point to
> > merge everything to the 2 series branch and work on that (except for this). What
> > do you think?
> 
> What do you mean with "everything" and "that"? :p

It was clear in my head :-D. Reading it again, no idea... Ay yai yai. Anyways, I will revert.

Issue #1647, opened 2021-07-21
Documentation: mention `toggle-mode` on tutorial section about `enable` and `disable` modes


pdelfino:
On [this section](https://github.com/atlas-engineer/nyxt/blob/5122b228a122177a54dc7e0e97426769d747b161/source/tutorial.lisp#L208), the tutorial already covers `enable` and `disable` mode.

`toggle-mode` has an overlap with them. Enable and disable act on multiple buffers. Toggle mode only works for the present buffer.

I have just realized this recently. The tutorial could cover that info with a 1 line addition. The section would go from (:h3 "Enable and Disable modes") to (:h3 " Enable, Disable, and Toggle modes") with a phrase addition such as: "In addition, it is also possible to apply and/or remove multiple modes only for the current buffer using toggle-mode`.

I can send a patch if you think it is a valuable addition.

pdelfino:
I am closing this issue since the suggestions have now been merged. Awkwardly, PR #1668 is listed as closed and not merged. But the suggested changes were, indeed, merged to the master branch.

PR #1668, opened 2021-07-26
tutorial: insert a brief explanation about `toggle-mode` on section about Enable and Disable of multiple modes.

pdelfino:
This addresses issue #1647.

### DIFF
--- a/source/tutorial.lisp
+++ b/source/tutorial.lisp
@@ -5,8 +5,8 @@
 
 (defun tutorial-content ()
   (spinneret:with-html-string
-   (:h2 "Core Concepts")
-   (:h3 "Keybindings and Commands")
+   (:h2 "Core concepts")
+   (:h3 "Keybindings and commands")
    (:p "Commands are invoked by pressing specific keys or from
 the " (:code "execute-command") " menu (" (:code (binding-keys 'execute-command))
 ").")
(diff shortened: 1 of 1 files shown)


source/tutorial.lisp
```diff
@@ -205,12 +205,14 @@ a user is using their email client which also uses " (:code "C-i") " for the ita
 executing " (:code "application-mode") " the " (:code "C-i") " binding is associated
 with the webpage's italic command instead of " (:code "autofill") ". Finally, the
 user can return to their configuration just by executing " (:code "application-mode") " again.")
-   (:h3 "Enable and Disable modes")
-   (:p "The command " (:code "enable-mode") " allows the user to apply a
-mode (such as " (:code "nosound-mode") " or " (:code "dark-mode") ") to multiple
-buffers at once. Conversely, it is possible to revert this action
+   (:h3 "Enable, Disable, and Toggle multiple modes")
```
Ambrevar, line 208: We are a bit inconsistent with our title casing.  Only a few use "title case", so I suggest using sentence case everywhere.
```
Enable, Disable, and Toggle multiple modes
```

pdelfino: Ok, fixed sente case on `Enable, disable, and toggle multiple modes`.

I did another commit in this PR making the whole tutorial **sentence case** consistent.

In addition, I noticed a section I wrote on `Copying and Pasting` was the only one using gerund (ing). So, to make it consistent on grammar in comparison to the whole manual, I have changed it to `Copy and paste`

jmercouris:
This has been merged, thank you!

PR #1669, opened 2021-07-26
Replace all ftype declarations by sera:->.

Ambrevar:
It's less verbose and reads more naturally from left to right.  It also enforces
the declaration of the return type.  When unspecified, use `*'.


jmercouris:
Thank you!

Issue #1659, opened 2021-07-23
Naming: change name of command from `noop` to `no-operation` or `nooperation`


pdelfino:
The opinion below might be subjective/biased on background, since I have never used/heard/seen the "noop" word before. 
Still, I think the command  `noop` could have a better name.

1 -  "noop" or "nop" or "no-op" are all abbreviations for "no operations". Nyxt's parlance avoids abbreviations. It is probably the only command using one. Also having 3 possibilities shows the lack of consistency;

2 - abbreviations are usually written with capital letters. The Wikipedia entry for this word in Assembly domain (probably the origin) used capital letters and chose "NOP" as the main [abbreviation](https://en.wikipedia.org/wiki/NOP_(code)). The Nyxt command `noop`  is not written with capital letters and will not be since command names are lower case and fuzzy search is case sensitive.

Thus, I would suggest using `nooperation` (symmetric to `nosound`, `noimage`, `noscript`, `nowebgl`) or even `no-operation` (the clearest, imho). 

I would be happy to send a patch if you think it is valuable.

Ambrevar:
Maybe `void` or `nothing` would be better and shorter names.  Other suggestions?

pdelfino:
I really like the natural language expressiveness of nothing.

pdelfino:
Something has just occured to me. I have another suggestion.

What about `unbound`?

That would be expressive in the config file:

```lisp
(define-configuration buffer
  ((override-map (let ((map (make-keymap "override-map")))
                   (define-key map
                      "C-space" 'unbound
                      "M-x" 'execute-command)))))
```

Ambrevar:
True, `unbound` sounds good.

But note the subtlety here: with `unbound` the user might assume that
the key is unbound in this keymap and thus forwarded.  Which is not
true.  When the key is _bound_ to any command, including `noop`, it is
no longer forwarded to other keymaps nor the renderer.

So `nothing` might be clearer here.

pdelfino:
I am closing this issue since the suggestions have now been merged with  PR #1672.

Issue #1658, opened 2021-07-23
Documentation: insert a one-line explanation about `noop` on manual


pdelfino:
On the subsection *keybinding configuration* of manual, there should be a brief mention to `noop`.

A few weeks ago, when I was tweaking with config files I was craving for that, I just did not know it already existed and the thing had this name. Today, accidentally, I uncover this command `noop`:

> A command that does nothing. This is useful to override bindings to do nothing.

At least, it could be used as **part of the example already available**: 

[...] 
The override-map is a keymap _that_ has priority over all other keymaps. By default, it has few bindings like the one for execute-command. You can use it to set keys globally (last line of the expression):

```
(define-configuration buffer
  ((override-map (let ((map (make-keymap "override-map")))
                   (define-key map
                     "M-x" 'execute-command
                     "C-space" 'noop)))))
```

Beyond the insertion on the example, there could be a brief explanation/mention right below the code:
 
[...] _The_ `noop` _command is useful to override bindings to nothing_. _In addition_, a more flexible _approach_ is to create [...]

I would be happy to send a patch.

Ambrevar:
Sure thing, feel free to send a pull request.

If it makes it easier for you, especially for documentation, it's often
faster to send a pull request directly than opening an issue about it ;)

pdelfino:
> Sure thing, feel free to send a pull request. If it makes it easier for you, especially for documentation, it's often faster to send a pull request directly than opening an issue about it ;)

Wow, I am glad to hear this!
I was reading about open source projects and the recommendation was to open issues before sending a PR. But I was feeling this was somewhat too much bureaucratic, at least for minor changes or documentation.

I think just sending the PR will make me more productive. Thanks for the hint.

pdelfino:
I am closing this issue since the suggestions have now been merged with  PR #1672.

PR #1672, opened 2021-07-26
Change `noop` to `nothing`, use it to illustrate keybinding configuration example, and briefly explain it.

pdelfino:
This PR addresses issues #1659 (naming) and #1658 (documentation).

### DIFF
--- a/source/command.lisp
+++ b/source/command.lisp
@@ -366,3 +366,8 @@ result."
   "A command that does nothing.
 This is useful to override bindings to do nothing."
   (values))
+
+(define-deprecated-command noop ()                 ; TODO: Replace with ESCAPE special command that allows dispatched to cancel current key stack.
+  "A command that does nothing.
+This is useful to override bindings to do nothing."
+  (values))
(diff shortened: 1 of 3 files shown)


pdelfino:
I can easily `load-system` on my environment using this branch. I am not sure why the tests are braking. Not sure if the problem is my code or the test implementation itself.

source/command.lisp
```diff
@@ -362,7 +362,7 @@ result."
             (log:debug "Prompt buffer interrupted")
             nil))))))
 
-(define-command noop ()                 ; TODO: Replace with ESCAPE special command that allows dispatched to cancel current key stack.
+(define-command nothing ()                 ; TODO: Replace with ESCAPE special command that allows dispatched to cancel current key stack.
```
Ambrevar, line 365: We can't rename command without breaking backward compatibility.  You need to keep `noop` as well and redefine it with `define-deprecated-command`.

pdelfino: > We can't rename command without breaking backward compatibility. You need to keep `noop` as well and redefine it with `define-deprecated-command`.

Interesting, I've just fixed it.

> OR just wait until 3.0.0 to merge it.

Since it is going to take a while until 3.0.0 and it is an easy fix, I did the `define-deprecated-command` trick.

Issue #1643, opened 2021-07-20
Feature request: create commands `go-up` and `go-homepage` to improve Nyxt as a doc overview tool


pdelfino:
A few days ago (maybe 10 days), user @mariari was talking about their use of Nyxt as a documentation overview tool in the IRC channel.

It was mentioned how useful the command `jump-to-heading` was. I feel the same. I have been using GNU manuals a lot with Nyxt.

Actually, I have been using `go-next` and `go-previous` too. Not sure if @mariari already tested those "new" commands.

`go-next` and `go-previous` were suggested by @jellelicht based on his experience with Vimium.

Vimium also has 2 other commands that would be useful for me in Nyxt:

```
gu      go up one level in the URL hierarchy
gU      go up to root of the URL hierarchy

```

They could be new Nyxt commands as `go-up` and `go-root`. I think this would be a low-hanging fruit for me considering the previous effort with parenscript.

pdelfino:
After talking with @jmercouris, he endorsed the idea of this feature request. He just suggested using `try-to-go-home` instead of `go-root`.

Ambrevar:
I don't understand try-to-go-home, what is "home"?
Also why "try"?

pdelfino:
Oh, sorry for not being clear, Pierre.

The command `try-to-go-home` would navigate to the "most basic URL on that domain". Usually, this is the **home page**. Vimium calls this the **root** in URL hierarch (which seems to be technically imprecise).

For instance, suppose you are at [https://ambrevar.xyz/articles.html](https://ambrevar.xyz/articles.html). After invoking `try-to-go-home`, the buffer would navigate to [https://ambrevar.xyz/](https://ambrevar.xyz/). The string "articles.html" would be removed.

In general, this "basic address" (root in URL hierarchy) is the **home page**. Unfortunately, some websites use other patterns as home pages, such as domain-name/index.html.

> I don't understand `try-to-go-home`, what is "home"? Also why "try"?

"home" -> homepage
"try" -> in general, it does take you to the homepage, but sometimes it does not take you to the homepage

We were insecure about the name. Home seems to be widely used as a shortcut for the homepage. Do you have a better suggestion?

`try-to-go-homepage` would be better?  `maybe-go-to-homepage`? `probably-go-to-homepage`?

Ambrevar:
I prefer "go-root" or "go-homepage".

About "try": why doesn't it take you to the homepage?
[...]

PR #1673, opened 2021-07-27
Create command that navigates to the homepage and a command that goes up in the URL hiearchy.

pdelfino:
This solves half of issue #1643 with the creation of the  `go-to-homepage` command. 

I am still struggling with regular expressions `cl-ppcre` to build the other command (`go-up`). While I advance on the other, you guys can review this one.

Thanks.

### DIFF
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -300,10 +300,23 @@ This was useful before Nyxt 2.0 as a workaround for hangs that would occur on pa
 
 (define-command go-to-homepage ()
   "Navigate to the homepage."
-  (let* ((url-address (url (current-buffer)))
-         (host (quri:uri-authority url-address))
-         (scheme (quri:uri-scheme url-address)))
-    (buffer-load (uiop:strcat scheme "://" host))))
+  (let* ((url (url (current-buffer)))
+         (authority (quri:uri-authority url))
+         (scheme (quri:uri-scheme url)))
+    (buffer-load (str:concat scheme "://" authority))))
+
+(define-command go-up ()
+  "Navigate to the upper level in URL path hierarchy."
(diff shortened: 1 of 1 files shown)


source/web-mode.lisp
```diff
@@ -298,6 +298,13 @@ This was useful before Nyxt 2.0 as a workaround for hangs that would occur on pa
                                  (query-selector-all "[rel=prev]") 0 (click))))
     (go-previous)))
 
+(define-command go-to-homepage ()
+  "Navigate to the homepage."
+  (let* ((url-address (url (current-buffer)))
+         (host (quri:uri-host url-address))
+         (scheme (quri:uri-scheme url-address)))
+    (buffer-load (concatenate 'string scheme "://" host))))
```
Ambrevar, line 306: Don't use `concatenate` for string concatenation, prefer `str:concat`.
(There are other options, like `uiop:strcat`.)

source/web-mode.lisp
```diff
@@ -298,6 +298,13 @@ This was useful before Nyxt 2.0 as a workaround for hangs that would occur on pa
                                  (query-selector-all "[rel=prev]") 0 (click))))
     (go-previous)))
 
+(define-command go-to-homepage ()
+  "Navigate to the homepage."
+  (let* ((url-address (url (current-buffer)))
+         (host (quri:uri-host url-address))
```
Ambrevar, line 304: prefer `uri-authority` over `uri-host` since there may be credentials in the URI.

Ambrevar:
You don't need regexp (in fact, avoid regexp as much as possible in general) to manipulate URI.

Instead, you should leverage the `quri` library.  It parses the URI for you, including the path.

pdelfino:
Hi guys,

I think this is good to go!

PR #1675, opened 2021-07-27
Fix domain name resolution on SBCL.

### LINKED ISSUE #1674: Nyxt cannot open a page if the server's hostname cannot be resolved by DNS protocol.
Hi! I use my own multicast DNS system (https://github.com/shamazmazum/zero-dns) for my local network. It works system-wide in conjuction with NSS (name service switch) plugin. Unfortunately nyxt cannot show pages hosted by LAN hosts if I use domain names rather than IP addresses.

Look at this REPL session:

~~~~
CL-USER> (iolib:lookup-hostname "vonbraun.local")
; Evaluation aborted on #<IOLIB/SOCKETS:RESOLVER-NO-NAME-ERROR {102A711DC3}>.
CL-USER> (nyxt::valid-url-p "http://vonbraun.local/")
NIL
CL-USER> (require 'sb-bsd-sockets)
NIL
CL-USER> (sb-bsd-sockets:host-ent-address (sb-bsd-sockets:get-host-by-name "vonbraun.local"))
#(192 168 20 10)
~~~~

As you can see, `get-host-by-name` can resolve hostname `vonbraun.local`, but nyxt says `http://vonbraun.local/` is not a valid URL because iolib cannot resolve it.

**How to reproduce**:
Try to visit a site using a hostname which is not resolvable by DNS protocol (e.g. it can be resolved via mDNS only).

**OS name, graphics card, etc.**
Irrelevant

I am unable to find a portable replacement to `sb-bsd-sockets:get-host-by-name`. Can you suggest something? It can be used instead of `iolib/sockets:lookup-hostname` in `sources/urls.lisp`. If you do not know a portable replacement, I can write a function which uses `sb-bsd-sockets:get-host-by-name` if using SBCL and falls back to `iolib/sockets:lookup-hostname` otherwise.

shamazmazum:
This commit fixes situations when a domain name cannot be resolved
by DNS protocol but can be resolved by gethostbyname(3). These
situations can occur when the host is resolved by mDNS, for example.

This fixes #1674

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -2,7 +2,9 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 #+sbcl
-(sb-ext:assert-version->= 2 0 0)
+(progn
+  (sb-ext:assert-version->= 2 0 0)
+  (require 'sb-bsd-sockets))
 
 (defvar *prefix* (format nil "~a/~a"
                          (or (uiop:getenv "DESTDIR") "")
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -66,7 +66,10 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
                             ,args))))))))
 
 (defmemo lookup-hostname (name)
-  "Memoized verison of `iolib/sockets:lookup-hostname'."
+  "Resolve hostname NAME and memoize the result"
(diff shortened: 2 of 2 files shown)


Ambrevar:
Thanks a lot!

PR #1678, opened 2021-07-27
Git submodules

### LINKED ISSUE #1427: Replace Quicklisp with Git submodules?
Building using Quicklisp has proven quite a pain to maintain.

What's more, a few issues remain:
- Quicklisp idiosyncrasis (no HTTPS support, may break in case of download errors, etc.)
- The build may break if Quicklisp updates a non-backward-compatible library in the distribution.

The last point is a deal-breaker for reproducibility.

So what about leveraging Git submodules instead?  This would allow us to mark the exact commit of each dependency.

- We already use submodules it, so no extra work required beside removing the Quicklisp chunk in `nyxt-quicklisp.asd`.
- Nyxt has about 100+ CL dependencies in its transitive closure.  It's quite something, but still maintainable considering most of them don't move much.
- It's easy to update the submodules thanks to the `git submodule update` command.

Possible drawback: it may make the Nyxt repo really big, thus adding lots to how much the user must download.
Set `submodule.<name>.shallow` to true, or use `--depth 1` option when initializing the submodules?

Ambrevar:
Fixes #1427.

To do:

- [ ] Add libraries that are not in Git repos in Guix.
- [ ] Remove Quicklisp snippet in the .asd.
- [ ] Test with the Ubuntu package.


Ambrevar:
Looks like the `shallow = true` option does not work like I want: the `cl-cffi-gtk` module is still fully checked out.

Ambrevar:
Turns out that `shallow` _does_ work, but only as far as Git can handle it.  Depending on the Git tree and if the distance of the requested commit to the history HEAD, Git may pull more of the history (but not all of it).

Looks like the only way to get rid of the 15 MiB of cl-cffi-gtk is to stop using it :p 

All in all, a Nyxt clone is pretty expensive using this method.  But is it a problem?

- OS packages can probably afford it.  (Guix and Nix won't use the submodules anyways.)
- Common Lisp hackers can use their own Quicklisp if they want, at the cost of reproducibility.
- It's expensive for Nyxt contributors who don't use Quicklisp, but only once.  After the first clone, subsequent updates should be very fast.

Thoughts?

jmercouris:
Seems the trade offs are reasonable. Could one parallelize the clone operations?

Ambrevar:
Thanks for asking, this got me to discover the `--jobs` option!

I suppose we could default to 4 jobs and expose the number of jobs as an environment variable option.

I'll test.

jmercouris:
Indeed, I found the same thing, I think that could speed things up a lot!

Ambrevar:
Some benchmarks with at least 2Mbps bandwitdh:

- -j1: 5m16.832s
- -j4: 2m50.719s
- -j8: 2m21.093s

jmercouris:
2 minutes at potato bandwidth seems OK. Let's do -j8?

Ambrevar:
Sorry, it's not 2Mbps but 2MB/s, so about 16Mbps.
I was leaning towards -j4 then.

jmercouris:
Why would we use less jobs? There is no performance gain. I say we saturate the cores! :-D

also 16 mbps is a little less potato :-D

Ambrevar:
Just being cautious, I have no clue how `git clone -j` behaves on
different hardware.

The gain with -j8 is rather low.

Ambrevar:
Closing in favour of #1802.

PR #1679, opened 2021-07-27
Fix `control c`  `control v` as the key binding for `copy` in cua, add `paste` in CUA, and add `copy` in `vi-normal-mode`

pdelfino:
The keybinding for `copy` (`C-c`) was not working due to the overshadow from `edit-from-external-editor` (`C-c '`). I have changed it to be `C-e '`. The "e" is a mnemonic reference to "edit". I spotted the problem happening and @jmercouris figured out its root.

Also, there was no keybinding for `paste` in CUA. So, I added it.

Finally, there was no keybinding for `copy` in `vi-normal-mode`. So, I added.

There is no keybinding for `paste` in `vi-normal-mode`. But, I am not sure if it makes sense to have one. Please, tell me if it does make sense.

### DIFF
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -88,6 +88,7 @@ and to index the top of the page.")
        "C-u C-M-j" 'follow-hint-nosave-buffer
        "C-x C-w" 'copy-hint-url
        "C-c" 'copy
+       "C-v" 'paste
        "button9" 'history-forwards
        "button8" 'history-backwards
        "C-+" 'zoom-page
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -166,6 +166,7 @@ and to index the top of the page.")
       (list
        "H" 'history-backwards
        "L" 'history-forwards
+       "y" 'copy
        "M-h" 'history-backwards-query
        "M-l" 'history-forwards-query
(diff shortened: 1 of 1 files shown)


Ambrevar:
Good catch, thanks!

Ambrevar:
VI bind `p` to pasting, I'll add it.

Ambrevar:
Nit: about your last commit message, it's not a "command" that was created, but a "binding".

pdelfino:
> VI bind `p` to pasting, I'll add it.

Ok, nice. But, how do you actually use it?

If I am vi-normal-mode and if I click in a comment box, the mode on Nyxt automatically changes to `vi-insert-mode`.
How do you use *paste* and [while on ](https://github.com/atlas-engineer/nyxt/commit/ff66c1bf8e7c7a214f7f1e1f14e6d3a6177fc125#diff-3bebeabe95d0dc07857cb16665158a1732d074f0202ad9918b1157bef0665f97)*vi-normal-mode*?
Could you give me an example so that I can learn?

Thanks!

> Nit: about your last commit message, it's not a "command" that was created, but a "binding".

Roger that. Sorry for the mistake.

Ambrevar:
Press escape to go back to normal mode.

PR #1682, opened 2021-07-28
move renderer/renderer specific files to folder

jmercouris:
they add noise in the main source tree of Nyxt, separating them helps
in providing a clear demaracation line


Ambrevar:
What about removing "renderer-" from the base names?
For instance, we would habe `source/renderers/gi-gtk.lisp`.
Having it twice is redundant I believe.

jmercouris:
I almost did that, but then changed my mind. I will do it.

Issue #1581, opened 2021-07-01
manual: Custom bindings example should not be bound in vi-insert


Ambrevar:
Reported here: https://old.reddit.com/r/Nyxt/comments/oa69oe/whats_preventing_you_from_using_nyxt_as_your/h3nhe4g/

PR #1686, opened 2021-07-29
Extend mode keymap scheme

Ambrevar:
At last, we can now easily customize mode bindings!

This has been a long term complaint :/ 

Please test and let me know if you like how I've extented `define-scheme`.

Fixes https://github.com/atlas-engineer/nyxt/issues/1581.


jmercouris:
A bit of a clunky extension to the macro, but I can't think of anything better without requiring an argument.

jmercouris:
Thank you!

Issue #1214, opened 2021-03-18
When many windows of Nyxt are open, the history of any given window is that of the Main window


mariari:
**Describe the bug**

When one creates a new window in nyxt, the newely created window is considered the main window (the window in which any url open will hop to). However when one tries to browse the history or go to a previous entry in the history, they are instead brought to the main windows history

EDIT :: it seems this behavior is unrelated to main/sub window, but rather what was the last window which changed the buffer URL.

**Precise recipe to reproduce the issue**

Take any buffer with a history on the main window, spawn a new window with `make-window`, go to the window which spawned the window, and view the history with navigate backwards, you'll see it's blank.

Note this persists even if you kill the window, as the window doesn't yet realize it's the main window again, this is simply fixed by switching the buffer on the main window. 

**Information**
- OS name+version:
    - Arch Linux
- Graphics card and driver:
- Desktop environment / Window manager name+version:
    - Xmonad 0.15
- How you installed Nyxt (Guix pack, package manager, build from source):
    - `make all` from master
- Information from command copy-system-information:

  If you can't run copy-system-information, please provide the following:
  - Nyxt version (from =M-x nyxt-version= or =nyxt --version=):
      - Version 2-pre-release-5-925-g5db56c8f 
  - Lisp implementation/version (if built from source):
      - SBCL 2.1.1.1 
  - Kernel name+version:
  - WebKitGTK+ | QtWebEngine version:

**Output when started from a shell**

Ambrevar:
I can reproduce.
In the original window, if I switch buffer back and forth, the history
shows up properly again.

@aartaka Any idea?

I'll give it a closer look later.

aartaka:
> I can reproduce. In the original window, if I switch buffer back and forth, the history shows up properly again. @aartaka Any idea? I'll give it a closer look later.

No idea yet. I forgot about multi-window setup when drafting GHT :(

Ambrevar:
It's probably a missing `current-buffer` let-binding somewhere.

chrisboeg:
@Ambrevar I can take a look at this tomorrow and see if I can come up with a patch.

Ambrevar:
That'd be fantastic! :)

chrisboeg:
[...]

Issue #1677, opened 2021-07-27
Replace Global History tree with Per-buffer History (prior model)


jmercouris:
Global History is a very cool idea in concept. Unfortunately, I don't think that the implementation is something that we've managed to nail down yet. For this reason, I suggest we move to the prior model.

I have a couple of problems that keep occurring:

1. When I press "history-backwards" after making a new buffer I am often surprised with a URL that doesn't exist in the current set of buffers. I don't know from where this URL is coming from, but as a user it feels almost random.
2. The `default.lisp` history file has extremely long lines, is very complex, and is hard to reason about. I cannot really look at it and understand it as I could with prior models.
3. The concepts of 'owned' and 'disowned' branches on the history tree are complex and difficult to understand.
4. Frequently, when I press "history-backwards" the previous page fails to load correctly from cache and say "Network error, page cannot be reached", I have to manually refresh the page.

Even though the history has had quite some months to mature, we still have not ironed out all of the bugs. For example, today as I was working through #1649 with Pedro, I realized the immense difficulty in working with the API, and with testing it. Together we also verified: https://discourse.atlas.engineer/t/why-having-a-history-disowned-and-a-global-history-source/229. Perhaps it is not a bug, and owned history cannot be deleted, this is however very surprising to the user, and was not a limitation of our tree based history.

Benefits of previous history model:
1. simple to program, simple to debug, simple to test
2. lightweight/smaller history file
3. behavior that more users can understand/conceptualize
4. we know that it works

What do you guys think? I know it was a lot of work to develop global history tree, but I beleive it would be a better decision for us in the long run to go back to per-buffer trees.

@pdelfino @aartaka @Ambrevar

Ambrevar:
> 1. When I press "history-backwards" after making a new buffer I am often surprised with a URL that doesn't exist in the current set of buffers. I don't know from where this URL is coming from, but as a user it feels almost random.

history-backwards has a bug, I just haven't found the time to get down
to it.

This is not related to the GHT _model_ though.

> 2. The `default.lisp` history file has extremely long lines, is very
> complex, and is hard to reason about. I cannot really look at it and
> understand it as I could with prior models.

Why do you want to look at the file itself?  The prompt buffer does a
much better job at presenting the history.

Even without GHT, default.lisp quickly becomes overwhelming.

> 3. The concepts of 'owned' and 'disowned' branches on the history tree
> are complex and difficult to understand.

Some things are complex but worth working on.  We are making a Lisp
[...]

Issue #1370, opened 2021-05-06
Fix history reparenting issue


Ambrevar:
Recipe:
- Go to https://example.org
- Go to https://en.wikipedia.org/wiki/Main_Page
- `history-backwards`
- Go to https://nyxt.atlas.engineer
- `history-backwards`

Brings us to wikipedia while we should be on example.org.

Ambrevar:
Fixed with https://github.com/atlas-engineer/nyxt/commit/a3ae7a80b0de5e054e08264e45916369bead7ab5?

Ambrevar:
Yes.

PR #1692, opened 2021-07-29
Htree without current owner

Ambrevar:
Lo and behold, the GHT is now much more reliable!

The theory is super simple: always pass the owner (the buffer ID) to all `htree` calls.

In practice, it induces a lot of changes.  Tests pass, but there might be a few broken calls here or there.
Please test!

Should fix https://github.com/atlas-engineer/nyxt/issues/1214,
https://github.com/atlas-engineer/nyxt/issues/1370 and address the biggest concern of https://github.com/atlas-engineer/nyxt/issues/1677.

Questions:

- This is _not backward compatible_.  The older history file still loads though, so the transition should be smooth for most people who don't use `htree` directly in their config.
Can we somehow preserve backward-compatibility even for the `htree` API?  It seems not possible, but please prove me wrong.
If we can't, shell we merge this as part of the 2-series still?

- Instead of passing `history` and `owner-id` as argument for `htree:forward`, etc., what about just passing an `owner` object`?  Then we could write

```lisp
(let ((owner (htree:owner history owner-id)))
  (htree:forward owner))
```

It'd be more convenient when chaining many htree calls, but it's less convenient for single htree calls.  Turns out that we never chain many htree calls, so might not be worth it.

Thoughts?

- What can of data should be use as owner IDs?  Currently, we use the buffer ID, but what about using the buffer object itself?  Maybe a bad idea, because if we don't clean up the history properly we may be holding reference back to dead buffers.  Thoughts?

Should we stick to buffer IDs, could we enforce typing?  Probably, like we do with `keymap` bindings.  I'd need to add this option then.


jmercouris:
Only bug I found was `buffer-history-tree`, and `history-tree`, unfortunately I found these bugs after merging... since you are quite a bit more familiar, could you fix these? I took a look, but was not sure.

Ambrevar:
Will look into it.

Other point I want to investigate since you reported an issue with this
patch when _not_ restoring buffers: we should reset the owners to '()
when loading the history and not restoring the buffers.

This bug has been around since we've merged the GHT.

Ambrevar:
Fixed history-tree with 87c828ae92489b70508b74bebe43b11474672859.

jmercouris:
Thank you Pierre!

PR #1696, opened 2021-07-30
Correct typo on application-mode activation from vi-insert-mode

whirm:
Found while following @Ambrevar's tip on issue #1687.


whirm:
Note that I have no idea of what I'm doing (tm). First time using common-lisp :)

Ambrevar:
Good catch, thanks!

PR #1700, opened 2021-07-31
Work In Progress: feature save input data

pdelfino:
Hi guys,

This is not ready yet. John suggested for me to do the following task:

```
Create a command that gets all the input data in HTML
Save it as a local lisp file
Later, we will re-insert the data on-demand
```

My implementation works to save a list of lists as a lisp file with the HTML input data *on my pc*. There are some things to fix and improve:

1 - Not sure if this parenscript approach is good. It looks ugly, but works. Suggestions?

2 - Since parenscript only returns primitive values (array not included), I have inserted markers and then used them to split the string. Would you do it differently?

3 - I need to *generalize* the place to save the file. The implementation uses a file path on my PC. Not sure what would be a good inspiration. Maybe the prompt buffer asking the user where to save it?


jmercouris:
I would leave the current parenscript interface you have for now, and work on the 'repopulating of the input fields' first.

pdelfino:
@jmercouris  answered those questions in a live meeting yesterday. I will synthesize it here if any co-worker or contributor reads this (maybe my future self too).

> 1 - Not sure if this parenscript approach is good. It looks ugly, but works. Suggestions?

It is ok. But there is a new way to solve tasks like this in which the DOM is parsed with `document-model`. Also, avoid `(if (not...))` and use `unless` in situations like this one.
 
> 2 - Since parenscript only returns primitive values (array not included), I have inserted markers and then used them to split the string. Would you do it differently?

This approach is ok. But there could be a more mature one. First, a JSON object could be built as a string. Then, a tested library in CL could be used to parse this JSON string in meaningful components. It would be faster and more reliable than my approach.

> 3 - I need to _generalize_ the place to save the file. The implementation uses a file path on my PC. Not sure what would be a good inspiration. Maybe the prompt buffer asking the user where to save it?

This will be done later. First, iterate the current solution to (i) read the local file with the data and (ii) re-insert the data on the HTML input fields automatically.

pdelfino:
Hi guys,

I ended up squashing too many commits. It was a bad move because the changes were made in 4 different files and the project has the practice of splitting changes into different files as different commits. Instead of fixing the history after an interactive rebase, it was faster to just create a new PR. 

Thus, I am closing #1700 in favor of #1769.

Issue #1683, opened 2021-07-28
Utilize Lparallel


jmercouris:
If we utilize lparallel (particularly in prompter), I think we'll see great performance increases by the usage of paralell mapcar etc.

https://lparallel.org/cognates/

@Ambrevar

Ambrevar:
lparallel does not have the SELECT statement like Calispel does.
(It also has an uncommon definition of "channels".)

If what you have in mind is to call object-properties in parallel, in
would only defer the issue: you'll divide the rendering time by the
number of cores you have.  Not much of an improvement in my opinion (at
the cost of huge CPU spikes :p).

Or did you have something else in mind?

Optimization ideas for the prompt buffer (it's been on the road map
since the beginning, just need to find time to work on it):

- Lazy evaluation of object-properties
- Caching / memoization of object-properties.

Combined, you would essentially reduce the object-properties computation
to practically nothing.

jmercouris:
I have no desire to replace calispel, I just want to use some of lparallel processing facilities to increase the performance of some of our mapcar operations in the codebase.

Ambrevar:
We could implement pmap for Calispel if we want to save a dependency,
it's trivial.

That said, I believe it's not a very promising strategy in the short
term.  I'd like to focus on the aforementioned ideas, they seem much
more promising! :)

jmercouris:
I think the cognates are very useful and could result in a huge speedup in many situations. The trend these days is largely towards parallelization of workloads. Many people have 16, 32 core processors. Utilizing lparallel cognate allow us to easily take advantage of those extra cores. There is no reason that something like mapcar needs to be single threaded. I believe the only reason it is is because concurrency was not included in the original CL spec.

Ambrevar:
It needs to be used sparingly.  Replacing all mapcar with pmap won't do
when the cost of firing up threads outweighs the processing of the
list.

For short lists (probably all of them) and lightweight operations, pmap
may be slower.  Benchmarking is necessary to avoid premature optimization.

Most operations performed by Nyxt are very lightweight.

jmercouris:
It doesn't need to fire up threads, it maintains some worker threads in the kernel. I'm not suggesting replacing all instances!

jmercouris:
It reuses them :-D

Ambrevar:
> It doesn't need to fire up threads, it maintains some worker threads
> in the kernel.

Fair enough.  Then I'm not sure what happens when you run multiple pmap
[...]

PR #1701, opened 2021-08-02
Parallelize

jmercouris:
As a follow up to: https://github.com/atlas-engineer/nyxt/issues/1683

```
(defparameter modes (mapcar #'mode-name (modes (current-buffer))))

;; NON PARALLEL VERSION
(time (prog1 "1" (list-commands :mode-symbols modes)))
Evaluation took:
  0.012 seconds of real time
  0.012358 seconds of total run time (0.012345 user, 0.000013 system)
  100.00% CPU
  34,664,147 processor cycles
  0 bytes consed

;; PARALLEL VERSION
(time (prog1 "1" (list-commands :mode-symbols modes)))
Evaluation took:
  0.005 seconds of real time
  0.052746 seconds of total run time (0.052746 user, 0.000000 system)
  1060.00% CPU
  14,406,970 processor cycles
  0 bytes consed
```

as you can see, over twice as fast, with just this one operation. There are of course dozens of places where we can utilize the cognates to dramatically improve performance. There are many operations that we do on lists and sets that we could parallelize for very significant performance increases.

### DIFF
--- a/source/concurrency.lisp
+++ b/source/concurrency.lisp
@@ -6,7 +6,8 @@
 (defun initialize-lparallel-kernel (&key (worker-count (sera:count-cpus)))
   "Initialize the lparallel kernel with WORKER-COUNT, if not supplied set it to
 the amount of CPU cores.."
-  (setf lparallel:*kernel* (lparallel:make-kernel worker-count)))
+  (unless lpara:*kernel*
+    (setf lpara:*kernel* (lpara:make-kernel worker-count))))
 
 (export-always 'with-protect)
 (defmacro with-protect ((format-string &rest args) &body body)
--- a/source/concurrency.lisp
+++ b/source/concurrency.lisp
@@ -3,6 +3,9 @@
 
 (in-package :nyxt)
 
+;; Initialize the lparallel kernel
+(setf lparallel:*kernel* (lparallel:make-kernel kernel-worker-count))
+
 (export-always 'with-protect)
 (defmacro with-protect ((format-string &rest args) &body body)
   "Run body with muffled condition when `*run-from-repl-p*' is nil, run normally otherwise.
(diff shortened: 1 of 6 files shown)


Ambrevar:
It processes lists faster for functional operations, but 0.012s is not enough, we would need to narrow down significant bottlenecks.

jmercouris:
.012s is a lot in terms of delay (1/10 of a second is very noticable). The important things are not the absolute value, but the magnitude and processor utilization. Gone are the days of dual core processors. We are sometimes with 8-16, even 64 cores in our machines.

I believe this demonstration shows that there is a large potential speed-up through the employment of lparallel cognates (where applicable)

You can also look at some the benchmarks here: https://lparallel.org/benchmarks/

The differences these days are even more exaggerated than what appears in these benchmarks.

jmercouris:
@Ambrevar OK to merge?

source/concurrency.lisp
```diff
@@ -3,6 +3,11 @@
 
 (in-package :nyxt)
 
+(defun initialize-lparallel-kernel (&key (worker-count (sera:count-cpus)))
+  "Initialize the lparallel kernel with WORKER-COUNT, if not supplied set it to
+the amount of CPU cores.."
+  (setf lparallel:*kernel* (lparallel:make-kernel worker-count)))
```
Ambrevar, line 9: You can use `lpara` here.

Ambrevar: Shouldn't we wrap this in un `(unless lparallel:*kernel ...)` to avoid defining it multiple times when started from the REPL?

Ambrevar:
OK to merge, see my other comments.

source/concurrency.lisp
jmercouris: Indeed.

PR #1703, opened 2021-08-03
repeat-mode: fix `repeat-every` docstring

pdelfino:
Currently, `repeat-every`'s docstring has some words in upper-case format which diverges from Nyxt's pattern. Since the tutorial (manual) uses `command-docstring-first-sentence` this is going to be ugly on tutorial. Thus, I am making the words lower case.

### DIFF
--- a/source/repeat-mode.lisp
+++ b/source/repeat-mode.lisp
@@ -42,7 +42,7 @@ Function taking a `repeat-mode' instance.")
   (nyxt/process-mode::initialize mode))
 
 (define-command-global repeat-every (&optional seconds function)
-  "Repeat a FUNCTION every SECONDS (prompts if SECONDS and/or FUNCTION are not provided)."
+  "Repeat a function every seconds (prompts if seconds and/or function are not provided)."
   (let ((seconds (or seconds
                      (ignore-errors
                       (parse-integer


aartaka:
W-w-wait a second, where does it diverge from the conventions (I am conscious that I've been somewhat lost while tinkering with WebKit stuff, but I remember no documentation format change)?

Writing function arguments as uppercase in function docstring is a solid Common Lisp convention almost everyone agrees on. We try (tried?) to stick to that too.

jmercouris:
I am with @aartaka I prefer the uppercase and used to use it in the docstrings myself. Pierre however prefers the characters " ` " and " ' ".

aartaka:
> I am with @aartaka I prefer the uppercase and used to use it in the docstrings myself. Pierre however prefers the characters " ` " and " ' ".

I'm not sure about backtick-quote for function arguments. Backtick-qoute notation is for symbols outside the function. E.g., if `buffer` is a function argument, you document it as `BUFFER`, while if it's a variable/function somewhere else, you document it as ``` `buffer'```. 

This notation comes from Emacs — it automatically xrefs all the backtick-quoted tokens and allows jumping to their definitions.

pdelfino:
> W-w-wait a second, where does it diverge from the conventions (I am conscious that I've been somewhat lost while tinkering with WebKit stuff, but I remember no documentation format change)?

Oh, sorry, tovarish Artyom. Actually, there was no documentation format change for the project.
 
> Writing function arguments as uppercase in function docstring is a solid Common Lisp convention almost everyone agrees on. We try (tried?) to stick to that too.

Oh, I did not know about this! I have basically used Common Lisp on exercises from books and on Nyxt.  This docstring (from `repeat-every`) was the first I saw those capital letters. I have been using this [style guide](https://lisp-lang.org/style-guide/). Do you recommend another one?

pdelfino:
> I am with @aartaka I prefer the uppercase and used to use it in the docstrings myself. Pierre however prefers the characters " ` " and " ' ".

I guess I prefer the " ` " and " ' ". Uppercase stuff reminds me when my mom is typing to stress something :laughing:. But I do not have a strong opinion on this.

The only thing I am worried is about the tutorial. The command `command-docstring-first-sentence` works as the name suggests. There are multiple docstrings being shown on the tutorial right now using this. There is no upper-cased word in the whole tutorial. These would be the firsts in an explanation for a single command. It would look strange, like careless copied-and-paste content.

But, we do not need to change the docstring as this PR suggests now. The `command-docstring-first-sentence` could be wrapped in a function converting the input into sentence case format. Then, the docstring would look as it is and the manual would be consistent.

aartaka:
> Oh, sorry, tovarish Artyom. Actually, there was no documentation format change for the project.

No problem :) It's good to bring confusing things up — all of us learn something this way.

> > Writing function arguments as uppercase in function docstring is a solid Common Lisp convention almost everyone agrees on. We try (tried?) to stick to that too.
> 
> Oh, I did not know about this! I have basically used Common Lisp on exercises from books and on Nyxt. This docstring (from `repeat-every`) was the first I saw using those capital letters. I have been using this [style guide](https://lisp-lang.org/style-guide/). Do you recommend another one?

I'm mainly sticking to [Google CL Style Guide](https://google.github.io/styleguide/lispguide.xml#Document_everything), as it's quite sane in documenting common practice. Just checked it: they recommend putting _any_ symbol, including function arguments, in uppercase when writing documentation. This is what lots of libraries do (see `cl-ppcre`, Alexandria, Serapeum docs, for example.)

Given that all of us use Emacs while working on Nyxt, using xref-d backtick notation sounds better for symbols that are not function arguments.

Regarding style guides: lisp-lang.org style guide is severely incomplete. Thus, you'll be better off following one of the bigger style guides: the Google one or Norvig and Pitman one (John is a huge follower of it — and for a reason: the guide is really good. He can share it with you.)

Ambrevar:
> I'm not sure about backtick-quote for function arguments. Backtick-qoute notation is for symbols outside the function. E.g., if `buffer` is a function argument, you document it as `BUFFER`, while if it's a variable/function somewhere else, you document it as ``` `buffer'```. 

This is correct and this is the convention we've followed since the
beginning.

aartaka:
So, documentation conventions clarified? Closing this?

jmercouris:
Yes.

pdelfino:
The root of the problem was uppercased words on the manual. I thought changing the docstring would be a good idea. But it was not. I was not aware of the convention. Sorry for the inconvenience.

Thus, I did this new PR now #1725. It changes the macro creating a possibility to deal with situations like this one.

Issue #1702, opened 2021-08-02
Possible bug: `watch-mode` is not prompting me about units of time


pdelfino:
**Describe the bug**
`watch-mode` is supposed to prompt  the user about units of time. This would be used to reload the buffer in specific time intervals. Unfortunately, I am not being prompted about time.

**Precise recipe to reproduce the issue**
- be on commit 865e5d65...;
- start nyxt;
- load a web page (I used [this one](https://en.wikipedia.org/wiki/Kiteboarding));
- enable watch-mode;
- see that the prompt buffer does **not** appear asking for the time interval in which the reload should happen.

For website-specific issues:
Can you reproduce this issue with Epiphany / GNOME Web (https://wiki.gnome.org/Apps/Web)?
Not relevant

**Information**
- OS name+version: 
Ubuntu 20.04 LTS

- Graphics card and driver:
product: Iris Plus Graphics 640
configuration: driver=i915 latency=0

- Desktop environment / Window manager name+version:
Name: GNOME Shell 3.36.7
Window manager's "showing the desktop" mode: OFF

- How you installed Nyxt (Guix pack, package manager, build from source):
Source

- Information from command copy-system-information:

- Information from `show-system-information`:
System information

Nyxt version: 2.1.1-370-g865e5d65
Renderer version: GI-GTK
Operating system kernel: Linux 5.8.0-63-generic
Lisp implementation: SBCL 2.0.1.debian (Dynamic space size: 1073741824)
Features: (:WEBKIT2 :WEBKIT2-2.32 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
[...]

PR #1710, opened 2021-08-04
watch-mode: update docstring with 5 min info.

pdelfino:
See the discussion here, [please](https://github.com/atlas-engineer/nyxt/issues/1702#issuecomment-891372832).

This [article](https://nyxt.atlas.engineer/article/watch-mode.org) from a few months ago nudged me. The article describes what `watch-buffer` does now. After I pressed `watch-mode`, I was expecting the prompt-buffer, since nothing appeared I thought it was a bug.

The docstring from `watch-mode` could be better. There is the 5 minutes information in a comment below the docstring. This PR only puts the comment info in the docstring info and deletes the comment.

### DIFF
--- a/source/watch-mode.lisp
+++ b/source/watch-mode.lisp
@@ -34,8 +34,7 @@
                       times to-seconds-multipliers))))
 
 (define-mode watch-mode (nyxt/repeat-mode:repeat-mode)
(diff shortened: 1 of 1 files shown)


jmercouris:
Thank you!

PR #1716, opened 2021-08-07
Persistently save bookmarks in nosave-buffer.

shamazmazum:
Hi! It took me some time to figure out that bookmarks are not worked as expected in `nosave-buffer`s. This is a commit message:

Typically, a browser is expected to save bookmarks when explicitly
told so (e.g. via bookmark-current-url command), nosave-buffer or
not. This commit makes an exception for bookmarks which are now saved
in the global cache rather than in the buffer's local cache and also
are saved to the disk.

### DIFF
--- a/source/data-storage.lisp
+++ b/source/data-storage.lisp
@@ -245,6 +245,13 @@ place of PATH `basename'.
            (setf fullname (str:concat fullname ".lisp")))
          fullname)))))
 
+(export-always 'always-save-p)
+(defgeneric always-save-p (path)
+  (:documentation "Should we always save data to this path (even if we
+are in nosave-buffer)?")
+  (:method ((path data-path)) nil)
+  (:method ((path bookmarks-data-path)) t))
+
 (export-always 'expand-data-path)
 (defmethod expand-data-path ((profile data-profile) (path data-path))
   "Return finalized path.
@@ -288,7 +295,8 @@ Define a method for your `data-path' type to make it restorable."))
 
 (defmethod store ((profile nosave-data-profile) (path data-path) &key &allow-other-keys)
   "This method guarantees PATH will not be persisted to disk in NOSAVE-DATA-PROFILE."
-  nil)
+  (when (always-save-p path)
+    (call-next-method)))
 
 (defmethod store :around ((profile data-profile) (path async-data-path) &key &allow-other-keys)
   (labels ((worker ()
@@ -382,11 +390,15 @@ means the store operations are systematically delayed."))
 
 (defmethod get-user-data ((profile nosave-data-profile) (path data-path))
   "Look up the buffer-local data in case of `nosave-data-profile'."
-  (%get-user-data profile path (user-data-cache profile)))
+  (if (always-save-p path)
+      (call-next-method)
+      (%get-user-data profile path (user-data-cache profile))))
 
 (defmethod set-user-data ((profile nosave-data-profile) (path data-path) value)
-  (setf (data (%get-user-data profile path (user-data-cache profile)))
-        value))
+  (if (always-save-p path)
+      (call-next-method)
+      (setf (data (%get-user-data profile path (user-data-cache profile)))
+            value)))
 
 (export-always 'get-data)               ; TODO: Unexport?
 ;; TODO: Better name? Isn't it too wide?


shamazmazum:
Force-pushed to retrigger stalled CI.

jmercouris:
This is a good question though, /should/ we? Do we want ANYTHING to be savable in a no-save buffer? I believe it is desirable behavior that bookmarks do not work in a no-save buffer.

jmercouris:
@aartaka how do you feel about this?

shamazmazum:
> I believe it is desirable behavior that bookmarks do not work in a no-save buffer.

At least I was confused believing the bookmark was actually saved (and `list-bookmarks` confirmed it, that is I saw the bookmark in the list). If you do not want to save bookmarks, you should empathize on that.

I look at Firefox as a reference. In Firefox you can configure it not to save history/cookies, but if you press "add bookmark" it will always be added. It's like a rule: "Cookies and history are not explicitly added by user, so we do not save them. Bookmarks are saved by user explicitly, so we do save them".

If I am incorrect, can I add this feature as a mode?

jmercouris:
I understand what you are saying, but it breaks some of my mental models about the properties of `no-save` buffers. I disagree with it as a change.

You could implement it as a mode, what is your idea exactly?

jmercouris:
Perhaps the commands that save bookmarks etc should be disabled in `no-save` buffers so that this confusion does not occur, OR they prompt the user with something like "You are in a no-save buffer, are you sure that you want to save a bookmark?".

You may view saving the bookmark as a command that the user initiates, but this is not a guarantee. It is just as possible that they have a mode that automatically bookmarks every page they visit, now there is a data leak in no-save buffers. This data leak is probably not intentional. This is of course problematic...!

shamazmazum:
> You may view saving the bookmark as a command that the user initiates, but this is not a guarantee. It is just as possible that they have a mode that automatically bookmarks every page they visit, now there is a data leak in no-save buffers. This data leak is probably not intentional. This is of course problematic...!

This has some reason. My idea is that perhaph I can write some mode like `always-save-bookmarks-mode` which installs methods for `store` and `get-user-data` (this is possible with `add-method`, isn't it?) on activation. Not sure if it's possible, I never added/removed methods in runtime :)

BTW, check #1717. nosave-buffers already can leak some data to the drive :)

jmercouris:
Indeed, that is a problem! Such a mode could exist. Yes. I'm not sure I am convinced enough to add it to the mainline of Nyxt, but you are welcome to make an extension with this mode, and we will list it!

aartaka:
We discussed the behavior while implementing the nosave buffers. I remember that we agreed that downloads should go to disk if the user has initiated them. It could well work for bookmarks too.

I'm on the side of making it an extension to not add cornercases to the justifyably strict model we have.

shamazmazum:
It seems I am more happy with the following lines in `init.lisp`:

~~~~
(defmethod expand-data-path ((profile nyxt:data-profile) (path nyxt:cookies-data-path))
  nil)

(defmethod store ((profile nyxt:data-profile) (path nyxt:history-data-path) &key &allow-other-keys)
  nil)
~~~~

They add more Firefox-like experience. No special buffers, history and cookies are just not saved. That's what I need.

A small question off-topic: strangely enough, cookies are still per-buffer (e.g. if I log in to Github in one buffer, the other buffer will show github as if I am not logged in). I thought it's a feature of `nosave-buffer`s which have their own data cache. Is this assumption wrong?

aartaka:
> It seems I am more happy with the following lines in `init.lisp`:
> 
> ```
> (defmethod expand-data-path ((profile nyxt:data-profile) (path nyxt:cookies-data-path))
>   nil)
> 
> (defmethod store ((profile nyxt:data-profile) (path nyxt:history-data-path) &key &allow-other-keys)
>   nil)
> ```
> 
> They add more Firefox-like experience. No special buffers, history and cookies are just not saved. That's what I need.

I've forgotten our data-paths are that flexible xD

> A small question off-topic: strangely enough, cookies are still per-buffer (e.g. if I log in to Github in one buffer, the other buffer will show github as if I am not logged in). I thought it's a feature of `nosave-buffer`s which have their own data cache. Is this assumption wrong?

@Ambrevar has been doing some work on that recently, so maybe he's in a better position to respond there. What I remember is that:
- WebKit used to be... unpredictable in regards to cookie storage.
- We have per-buffer cookie managers and thus cookies are mostly per-buffer. They should be shared, though.

Ambrevar:
What I did some weeks ago was to "fix" internal buffer data manager
path.  I haven't touched the cookie paths as far as I remember, but I'll
investigate on my side too.

PR #1721, opened 2021-08-10
tutorial: add a section about commands that reduce bandwidth.

pdelfino:
Just like the recent patches with sections glueing similar command (`copy and paste`, `automation`), this section clusters 3 commands that contribute to reduce the bandwidth.


jmercouris:
Thank you!

PR #1723, opened 2021-08-11
gtk.lisp: update prompt-buffer on mouse button events

tiberious726:
In the X11 tradition, mouse-2 pastes from the XA_PRIMARY clipboard, so
the prompt-buffer should update accordingly.

The prompt buffer will update regardless of which mouse button is
pressed, as we don't have a good way to decode if the user has
remapped their mouse buttons here, or if they have assigned custom
actions to other buttons

### DIFF
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -508,6 +508,13 @@ See `gtk-browser's `modifier-translator' slot."
           (character character))
       (setf (gtk:gtk-entry-text (key-string-buffer window)) ""))))
 
+(defun update-prompt (buffer)
+  (run-thread "Prompt updater"
+    ;; Rebind prompt-buffer to ensure the watcher does not mix up the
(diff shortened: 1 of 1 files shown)


tiberious726:
I wasn't sure what the "watcher" referred to by this comment was:
;; Rebind prompt-buffer to ensure the watcher does not mix up the
;; different prompt-buffers.
So I left the logic in place, but I'm not sure the rebinding in the other thread is still necessary.

jmercouris:
I see, thank you for this patch, I will review this and get back to you. There is actually a lot I would like to change about the way input is handled for 3.0.0.

Ambrevar:
Looks good to me.
@jmercouris Anything blocking this?
The whole architecture of button events could be improved, but this patch seems OK to merge in the mean time.

jmercouris:
OK, feel free to merge it then.

PR #1724, opened 2021-08-11
Feature: `reduce-bandwidth-mode`

pdelfino:
This mode enables three other modes:  `no-images`, `no-javascript`, and `no-webgl`. All of them contribute to reduce the bandwidth. Disabling `reduce-bandwidth-mode` also disables all three modes at once.


jmercouris:
Nice, thanks a lot :-)

PR #1725, opened 2021-08-11
Fix sentence case docstring

pdelfino:
This solves the problem discussed [here](https://github.com/atlas-engineer/nyxt/pull/1703#issuecomment-892187627).

### DIFF
--- a/source/input.lisp
+++ b/source/input.lisp
@@ -16,13 +16,16 @@
                                                                 '())))
                            ")"))))
 
-(defmacro command-docstring-first-sentence (fn)
+(defmacro command-docstring-first-sentence (fn &key (sentence-case-p nil))
   "Print FN first docstring sentence in HTML."
   `(if (fboundp ,fn)
        (spinneret:with-html
          (:span
           (sera:ensure-suffix
-           (or (first (ppcre:split "\\.\\s" (documentation ,fn 'function)))
+           (or (if ,sentence-case-p
+                   (str:sentence-case
+                     (first (ppcre:split "\\.\\s" (documentation ,fn 'function))))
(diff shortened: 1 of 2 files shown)


source/input.lisp
```diff
@@ -16,13 +16,16 @@
                                                                 '())))
                            ")"))))
 
-(defmacro command-docstring-first-sentence (fn)
+(defmacro command-docstring-first-sentence (fn &key (sentence-case-p :nil))
```
jmercouris, line 19: why are you using the symbol `:nil` rather than `nil`? is this intentional?

pdelfino: I will fix it. It is because I was reading the code base. This nudged me:

```lisp
(defmacro with-current-panel ((buffer-var title &key (side :left))
                              &body body)

```

But `left` is not  a reserved word as `nil` is.

source/tutorial.lisp
```diff
@@ -322,7 +322,8 @@ automate the reading experience:")
     (:li (command-markup 'nyxt/watch-mode:watch-mode) ": "
          (command-docstring-first-sentence 'nyxt/watch-mode:watch-mode))
     (:li (command-markup 'nyxt/repeat-mode:repeat-every) ": "
-         (command-docstring-first-sentence 'nyxt/repeat-mode:repeat-every))))
+         (command-docstring-first-sentence 'nyxt/repeat-mode:repeat-every
+   :sentence-case-p t))))
```
jmercouris, line 326: this indentation feels incorrect

pdelfino: Indeed. My bad. Fixed it.

PR #1727, opened 2021-08-12
gtk.lisp: Enable sandboxing

tiberious726:
Enable sandboxing for buffers that have an ID as well as those that
don't by making both use the web-context method


tiberious726:
I eliminated the conditional logic, as make-data-manager seems to handle ephemerality just fine

tiberious726:
The original `web-context` method seemed to be about reusing the web context for buffers without an ID. I couldn't get nyxt to actually produce such a buffer.

If this logic is eliminated, the web-context slot on the browser class should also go, and if it is kept (perhaps with a method named shared-web-context), I think the slot should also be named something more descriptive (or given some documentation)

jmercouris:
I agree, thanks for investigating so thoroughly.

PR #1732, opened 2021-08-14
gtk.lisp: set sandboxing in web-context's initialize-instance method

tiberious726:
Make sure this is taken care of by using a real constructor instead of
an ersatz one

### DIFF
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -75,10 +75,13 @@ data-manager will store the data separately for each buffer.")
   (:accessor-name-transformer (class*:make-name-transformer name)))
 (define-user-class buffer (gtk-buffer))
 
-(defmethod make-web-context ((manager webkit:webkit-website-data-manager))
-  (let ((context (make-instance 'webkit:webkit-web-context :website-data-manager manager)))
-    (webkit:webkit-web-context-set-sandbox-enabled context t)
-    context))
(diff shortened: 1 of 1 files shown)


tiberious726:
Also test for actually having the feature in cl-webkit. tho I don't know if its absence should be a silent condition

jmercouris:
Fantastic, thank you!

jmercouris:
It's absence can be a silent condition, if anything we should just peg a version of cl Webkit to Nyxt rather than using feature flags

source/renderers/gtk.lisp
```diff
@@ -75,10 +75,13 @@ data-manager will store the data separately for each buffer.")
   (:accessor-name-transformer (class*:make-name-transformer name)))
 (define-user-class buffer (gtk-buffer))
 
-(defmethod make-web-context ((manager webkit:webkit-website-data-manager))
-  (let ((context (make-instance 'webkit:webkit-web-context :website-data-manager manager)))
-    (webkit:webkit-web-context-set-sandbox-enabled context t)
-    context))
+(defclass webkit-web-context (webkit:webkit-web-context) ()
+  (:metaclass gobject:gobject-class))
```
Ambrevar, line 79: @tiberious726 Out of curiosity, why did you set the metaclass to `gobject-class`?

PR #1734, opened 2021-08-15
tts-mode

chrisboeg:
Initial implementation of tts-mode or text-to-speak mode.

### DIFF
--- a/source/tts-mode.lisp
+++ b/source/tts-mode.lisp
@@ -89,10 +89,8 @@ passed as the argument."
 (define-command start-speak ()
   "Start text-to-speak the content of the current buffer
 matching the selector."
-  (let ((mode (find-submode (current-buffer) 'tts-mode)))
-    (process-document mode)))
+  (process-document (current-mode 'tts-mode)))
 
 (define-command stop-speak ()
   "Stop running text-to-speak process if there is one."
-  (let ((mode (find-submode (current-buffer) 'tts-mode)))
-    (if-process-then-terminate mode)))
+    (if-process-then-terminate (current-mode 'tts-mode)))
--- a/source/tts-mode.lisp
+++ b/source/tts-mode.lisp
@@ -38,11 +38,6 @@ Example:
     (lambda (mode)
       (if-process-then-terminate mode)))))
 
-(defmethod on-signal-load-finished ((mode tts-mode) url)
-  (declare (ignore url)))
-  ;; TODO: document-mode not ready here???
-  ;; (process-document mode))
-
 (defmethod process-document ((mode tts-mode))
   "Fetch the text in buffer that matches `selector` and send it off
 to get *spoken*."
--- a/source/tts-mode.lisp
+++ b/source/tts-mode.lisp
@@ -24,18 +24,19 @@ Example:
    ((nyxt/tts-mode:executable \"espeak\")
     (nyxt/tts-mode:selector \"h1, h2, h3, h4, p\")))
 "
-  ((executable nil)
+  ((executable nil
+               :type (or string null)
+               :documentation "The executable command to run.")
    ;; TODO: If you specify multiple tags, like h1, h2, h3, p, they are selected in that
    ;; order, not in the order they appear on the page. This is different than how it's
    ;; done with e.g. JS document.querySelectorAll which returns them in the order they
    ;; appear, which is IMO the correct way ... can you configure clss to do this?
-   (selector "p")
+   (selector "p"
+             :type string
+             :documentation "css selector that describes what elements text to speak.")
    (destructor
     (lambda (mode)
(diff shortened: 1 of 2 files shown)


source/tts-mode.lisp
```diff
@@ -0,0 +1,116 @@
...
+
+\(define-configuration nyxt/tts-mode:tts-mode
+   ((nyxt/tts-mode:executable \"espeak\")
+    (nyxt/tts-mode:selector \"h1, h2, h3, h4, p\")))
+"
+  ((executable nil)
+   ;; TODO: If you specify multiple tags, like h1, h2, h3, p, they are selected in that
+   ;; order, not in the order they appear on the page. This is different than how it's
+   ;; done with e.g. JS document.querySelectorAll which returns them in the order they
+   ;; appear, which is IMO the correct way ... can you configure clss to do this?
```
chrisboeg, line 31: As described here, @jmercouris, I haven't figured out how to get clss to return the elements in the order they appear on the page. This is pretty important to get working.

source/tts-mode.lisp
```diff
@@ -0,0 +1,116 @@
...
+    (lambda (mode)
+      (initialize mode))))
+  :documentation
+  "For the mode to work, you need to set the executable slot
+to an executable thats available in the environment path.")
+
+(defmethod on-signal-load-finished ((mode tts-mode) url)
+  (declare (ignore url)))
+  ;; TODO: document-mode not ready here???
+  ;; (process-document mode))
```
chrisboeg, line 46: I believe we talked about the mode should automatically read up the content. I tried to do it as the content loads like this, but at this point, the document-model is not computed yet. Any ideas?

source/tts-mode.lisp
```diff
@@ -0,0 +1,114 @@
...
+(defmethod sanitize-text ((mode tts-mode) text)
+  "Remove anything non- alpha, numeric, and space."
+  (remove-if
+   (complement
+    (lambda (s)
+      (or
+       (string= #\Space s)
+       (alpha-char-p s)
+       (not (null (digit-char-p s))))))
+   text))
```
chrisboeg, line 103: This is probably not exactly what we want.

source/tts-mode.lisp
jmercouris: That is important. I've looked through the source myself and I can't see a way beyond iterating through the plump:dom manually. For now, let's avoid headings and just do paragraphs, and then it /should/ be in order. Alternatively we could just iterate through the plump:dom printing all text, that however seems silly.

source/tts-mode.lisp
jmercouris: It should be that as soon as you type `(document-model buffer)` it will be computed. I think there might be some confusion here though, because I think it should actually be bound to a keybinding. So, don't worry about handling it on load.

source/tts-mode.lisp
jmercouris: Indeed. I think you could do `(str:remove-punctuation "xyz")` instead, it accomplishes the same thing for now.

jmercouris: Perhaps in the future we can make a lib to CFFI bindings.

source/tts-mode.lisp
```diff
@@ -0,0 +1,114 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/tts-mode
+    (:use :common-lisp :nyxt)
+  (:documentation "Mode for text-to-speak a documents content."))
+
+(in-package :nyxt/tts-mode)
+
+(defvar executable-process-info nil)
```
jmercouris, line 10: I'm curious, why not make this a slot in tts-mode like `executable`?

source/tts-mode.lisp
```diff
@@ -0,0 +1,114 @@
...
+      (or
+       (string= #\Space s)
+       (alpha-char-p s)
+       (not (null (digit-char-p s))))))
+   text))
+           
+(define-command start-speak ()
+  "Start text-to-speak the content of the current buffer
+matching the selector."
+  (let ((mode (find-submode (current-buffer) 'tts-mode)))
```
jmercouris, line 108: you can use `current-mode` instead, it is a little simpler :-)

chrisboeg:
@jmercouris: I've applied the changes we discussed. Only thing is: are we sure about `str:remove-punctuation`. It removes all punctuation, even commas and dots. It doesn't seem to me that `espeak` uses that to "understand" the text, but maybe some tts-software does?

jmercouris:
I don't know, if it turns out to be a problem for other programs, we can use a different regex by utilizing cl-ppcre directly :-).

Thank you for this beautiful work!

PR #1738, opened 2021-08-16
libraries/history-tree: remove line since we already started 3.0.

pdelfino:
Since we already started working on version 3.0.0, this line can be removed.


jmercouris:
Thank you!

Issue #1730, opened 2021-08-13
No suggestions/completions with default search engine 


franburstall:
**Describe the bug**

I hit `M-l` or `M-x set-url-new-buffer`  and type "spherical".

What I expect: a list of suggestions from the search engine that refine "spherical".

What I get: no suggestions.  See screenshot.

![1628884808](https://user-images.githubusercontent.com/5726790/129413126-53d52299-a240-49fc-abc6-47a9f6bdc5e4.png)

On the other hand, if I type instead "ddg spherical" to declare the (same) search engine, I get the list of suggestions I expect:

![1628884848](https://user-images.githubusercontent.com/5726790/129413374-cdae255a-7cac-4152-811f-eefe946b1d2c.png)

Is this a bug or a feature?  If the latter, how do I get the second behaviour without redundantly declaring the search engine?

**Information**
- OS name+version: Debian-10
- Graphics card and driver:
- Desktop environment / Window manager name+version: EXWM
- How you installed Nyxt (Guix pack, package manager, build from source): Source
- Information from `show-system-information`:

Nyxt version: 2.1.1-507-g1d57a5e7
Renderer version: GI-GTK
Operating system kernel: Linux 4.19.0-17-amd64
Lisp implementation: SBCL 2.1.1 (Dynamic space size: 1073741824)
Features: (:SLYNK :WEBKIT2 :WEBKIT2-2.32 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
 :GLIB-2-30 :GLIB-2-32 :GLIB-2-34 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42
[...]

PR #1741, opened 2021-08-17
buffer: Allow non-prefixed completion for default search engine.

aartaka:
This addresses #1730 by adding a buffer-based toggle for search auto-completion even if there's no engine prefix.

Tends to be slow, but such are Dexador requests, we can do nothing about it.

Things to discuss: Can we optimize it? Should it be the default behavior?

Closes: #1730.
Closes: #1885.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1325,7 +1325,8 @@ Finally, if nothing else, set the `engine' to the `default-search-engine'."))
               (list (make-instance 'new-url-query
                                    :query       input
                                    :check-dns-p check-dns-p)))
-            (sera:and-let* ((buffer (current-buffer))
+            (sera:and-let* ((completion engine-completion-p)
+                            (buffer (current-buffer))
                             (always (search-always-auto-complete-p buffer))
                             (engine (default-search-engine))
                             (completion (completion-function engine))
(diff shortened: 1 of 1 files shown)


jmercouris:
Part of me says it should only be the default beahvior if we can I N S T A N T L Y return the results, or DWIM even when there is delay in showing the results in the prompt buffer. How to do this, I don't know.

jmercouris:
Well, I do /know/ of a way, I don't know of a good way.

Ambrevar:
Maybe we can play with `prompt-buffer`'s `notification-delay` so that the user input appears instantaneously.

If it does not work, we might have to introduce a new timer.

Ambrevar:
@aartaka Now that we compute the search engine completions on the second pass, can we merge this?

aartaka:
Let me rebase and test. Should be fine to merge after I fix merge conflicts.

jmercouris:
@aartaka Any update?

aartaka:
> @aartaka Any update?

Sorry, had a lot of other stuff :) I will get back to it in the first week of December, hopefully!

aartaka:
I've rebased over master and made auto-completion to work by default. It's much more responsive now -- user input is displayed right away, while suggestions are loaded with a small lag. Fine to merge, I believe.

aartaka:
While I was at it, I've done #1885. Search completions can now return two-element lists (would alists fit better there? I'm not sure.) instead of strings, and the second element of those will be recognised as a URL for the completion. Try `wiki` with this feature and you'll see no redirections!

DuckDuckGo seems to return no direct URLs in its suggestions, thus it stays the same as before.

Ambrevar:
Excellent, feel free to merge!

PR #1746, opened 2021-08-22
visual-mode: Fix toggle-mark.

hendursaga:
ParenScript's is-collapsed does not return "true" but T or nil, which made it so you could not deactivate the mark by toggling twice, etc.


jmercouris:
I THANK THEE SUMERIAN GOD OF JUSTICE!

PR #1747, opened 2021-08-23
Refactor copy placeholder

pdelfino:
I am improving the aesthetic of code that I wrote a few weeks/days ago. In the case of `copy-placeholder`, it will make the source code shorter and more reliable, since the auxiliary function will be local to the main function.


jmercouris:
A wonderful improvement, thank you :-)

PR #1748, opened 2021-08-23
Documentation use `command-docstring-first-sentence` on tutorial

pdelfino:
The macro `command-docstring-first-sentence`  was already being used in the newer sections in tutorial. Some old sections were not using it, though. There were ad-hoc explanations about the commands. Almost always, the description would be really similar to the first sentence in the commands docstring. I just inserted the `command-docstring-first-sentence` so that things are more dynamic.

As I was doing this, on a few rare occasions the ad-hoc explanation on tutorial was actually better (IMHO) than the docstring. Thus, I inserted `command-docstring-first-sentence` **and** changed the docstring to have the better description version.

I also noticed a few rare mistakes that were fixed.

I was unsure on how to treat this considering commit frequency. One commit for each insertion of  `command-docstring-first-sentence` would be too much for master, I guess. So, I did 1 commit for each section changed. Enough atomic, I guess.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1021,7 +1021,7 @@ proceeding."
 
 (-> switch-buffer (&key (:id string) (:current-is-last-p boolean)) *)
 (define-command switch-buffer (&key id (current-is-last-p nil))
-  "Switch the active buffer in the current window.
+  "Switch buffer using fuzzy completion to quickly find whatever buffer you are looking for.
 Buffers are ordered by last access.
 With CURRENT-IS-LAST-P, the current buffer is listed last so as to list the
 second latest buffer first."
--- a/source/download-mode.lisp
+++ b/source/download-mode.lisp
@@ -169,5 +169,5 @@ download."
   (download (current-buffer) (url (current-buffer))))
 
 (define-command download-open-file ()
-  "Open a downloaded file."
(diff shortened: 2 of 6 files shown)


jmercouris:
This pattern:

```
(:li (command-markup 'switch-buffer) ": " (command-docstring-first-sentence 'switch-buffer))
```

is repeating itself in the code endlessly, command-markup and then command-docstring separated by a colon. Perhaps we could use a macro for this? Something like:

(list-commands '(switch-buffer delete-buffer ...)) and it would generate all of the forms?

jmercouris:
Thank you!

pdelfino:
> This pattern:
> 
> ```
> (:li (command-markup 'switch-buffer) ": " (command-docstring-first-sentence 'switch-buffer))
> ```
> 
> is repeating itself in the code endlessly, command-markup and then command-docstring separated by a colon. Perhaps we could use a macro for this? Something like:
> 
> (list-commands '(switch-buffer delete-buffer ...)) and it would generate all of the forms?

Makes total sense. I wished I had realized this before doing it manually :sweat_smile:

I will work on that macro soon and submit a PR. Hopefully, today :)

jmercouris:
Sometimes one must do it manually to realize said macro could exist :-)

PR #1749, opened 2021-08-23
Remove an entry from *Bookmarks* buffer when the user clicks "Delete".

shamazmazum:
This makes bookmark buffer more interactive.


jmercouris:
Cool, thank you! I believe onclick can optional send an argument that is the element clicked. In that case you would not need to create/store the IDs.

shamazmazum:
That was my first acquaintance with JS, don't judge too much :)

jmercouris:
11 no worries :-)

PR #1757, opened 2021-08-29
gtk.lisp: Web context sharing and separation

tiberious726:
Prior to this commit, nyxt would create a new web-context for each
buffer, but back each non-ephemeral web-context to one of two
locations on disk, ("data-manager" for external buffers, and
"WebKitCache" for internal buffers (as the webkit-data-manager wasn't
being set, so would default)). Since they were backed to the same
location on disk, some session-sharing was possible, but in an
unreliable manner. Furthermore, while the data-directory was set to be
inside of xdg-cache-home the cache-directory wasn't being set, so some
caches were being stored inside of the data-manager-data-path and
others on the system-wide webkit default (hsts-cache-directory and
offline-application-cache-directory).

This patch causes nyxt, by default, to re-use in-memory web-contexts
based on if the buffer is internal or external and whether or not it
is meant to be ephemeral, determined by the make-web-view
function. This function can now also return a web-view in a custom
web-context.

Data-path names are generated based on the name of the context (and
cookies are now stored separately per context). The data-manager's
data-directory is stored inside of xdg-data-home and it's
cache-directory is stored inside of xdg-cache-home.

As buffers are now assigned to a web-context, it doesn't make sense to
store defaults controlling web-contexts in buffer classes. The
default-cookie policy was moved to a setting in the browser and a
per-context cookie data-path is generated

This sets the groundwork to allow user-facing browsing "sessions",
akin to Firefox's "Multi Account Container"s, where users can assign
groups of buffers to share a browsing context.

### DIFF
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -29,11 +29,25 @@ want to change the behaviour of modifiers, for instance swap 'control' and
     (delete nil (mapcar (lambda (mod) (getf plist mod)) modifier-state))))
 
 \(define-configuration browser
-  ((modifier-translator #'my-translate-modifiers)))"))
+  ((modifier-translator #'my-translate-modifiers)))")
+   (web-contexts (make-hash-table :test 'equal)
+                 :documentation "Persistent `nyxt:webkit-web-contexts'
+Keyed by strings as they must be backed to unique folders")
+   (ephemeral-web-contexts (make-hash-table :test 'equal)
+                           :documentation "Ephemeral `nyxt:webkit-web-contexts'"))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
 (define-user-class browser (gtk-browser))
+
+(defmethod get-context ((browser gtk-browser) name &key ephemeral-p)
+  (if ephemeral-p
+      (or (gethash name (ephemeral-web-contexts browser))
+          (setf (gethash name (ephemeral-web-contexts browser))
+                (make-context name :ephemeral-p t)))
+      (or (gethash name (web-contexts browser))
+          (setf (gethash name (web-contexts browser))
+                (make-context name)))))
 
 (define-class gtk-window ()
   ((gtk-object)
@@ -64,12 +78,7 @@ want to change the behaviour of modifiers, for instance swap 'control' and
     :documentation "Store all GObject signal handler IDs so that we can disconnect the signal handler when the object is finalised.
 See https://developer.gnome.org/gobject/stable/gobject-Signals.html#signal-memory-management.")
    (gtk-proxy-url (quri:uri ""))
-   (proxy-ignored-hosts '())
-   (data-manager-path (make-instance 'data-manager-data-path)
-                      :documentation "Directory in which the WebKitGTK
-data-manager will store the data separately for each buffer.")
-   (gtk-extensions-path (make-instance 'gtk-extensions-data-path)
-                        :documentation "Directory to store the WebKit-specific extensions in."))
+   (proxy-ignored-hosts '()))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
@@ -82,6 +91,24 @@ data-manager will store the data separately for each buffer.")
   #+webkit2-sandboxing
   (webkit:webkit-web-context-set-sandbox-enabled web-context t)
   web-context)
+
+(defmethod data-directory ((web-context webkit-web-context))
+  "Directly returns the CFFI object's `base-data-directory'"
+  (slot-value (slot-value web-context 'webkit::website-data-manager) 'webkit::base-data-directory))
+
+(defmethod cache-directory ((web-context webkit-web-context))
+  "Directly returns the CFFI object's `base-cache-directory'"
+  (slot-value (slot-value web-context 'webkit::website-data-manager) 'webkit::base-cache-directory))
+
+(defclass webkit-website-data-manager (webkit:webkit-website-data-manager) ()
+  (:metaclass gobject:gobject-class))
+
+;; create-gobject-from-class in gobject.base.lisp directly reads the initargs to make-instance, so this cannot
+;; be used to enforce ephermerality (or anything else that needs to happen in the cffi constructor)
+(defmethod initialize-instance :after ((data-manager webkit-website-data-manager) &key)
+  #+webkit2-tracking
+  (webkit:webkit-website-data-manager-set-itp-enabled data-manager t)
+  data-manager)
 
 (defvar gtk-running-p nil
   "Non-nil if the GTK main loop is running.
@@ -182,14 +209,36 @@ not return."
     (gtk:leave-gtk-main)))
 
 (define-class data-manager-data-path (data-path)
-  ((dirname (uiop:xdg-cache-home +data-root+ "data-manager"))
-   (ref :initform "data-manager"))
+  ((dirname (uiop:xdg-data-home +data-root+ "web-context"))
+   (ref :initform "web-context"))
   (:export-class-name-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
 
 (defmethod expand-data-path ((profile nosave-data-profile) (path data-manager-data-path))
   "We shouldn't store any `data-manager' data for `nosave-data-profile'."
   nil)
+
+(defun data-manager-data-path-for-context (name)
+  (let ((ref (format nil "~a-web-context" name)))
+    (make-instance 'data-manager-data-path
+                   :ref ref
+                   :dirname (uiop:xdg-data-home +data-root+ ref))))
+
+(define-class data-manager-cache-path (data-path)
+  ((dirname (uiop:xdg-cache-home +data-root+ "web-context"))
+   (ref :initform "web-context"))
+  (:export-class-name-p t)
+  (:accessor-name-transformer (class*:make-name-transformer name)))
+
+(defmethod expand-data-path ((profile nosave-data-profile) (path data-manager-cache-path))
+  "We shouldn't store any `data-manager-cache' data for `nosave-data-profile'."
+  nil)
+
+(defun data-manager-cache-path-for-context (name)
+  (let ((ref (format nil "~a-web-context" name)))
+    (make-instance 'data-manager-cache-path
+                   :ref ref
+                   :dirname (uiop:xdg-cache-home +data-root+ ref))))
 
 (define-class gtk-extensions-data-path (data-path)
   ((dirname (uiop:xdg-config-home +data-root+ "extensions"))
@@ -200,6 +249,12 @@ not return."
 (defmethod expand-data-path ((profile nosave-data-profile) (path gtk-extensions-data-path))
   "We shouldn't enable (possibly) user-identifying extensions for `nosave-data-profile'."
   nil)
+
+(defun gtk-extensions-data-path-for-context (name)
+  (let ((ref (format nil "~a-gtk-extensions" name)))
+    (make-instance 'data-manager-cache-path
+                   :ref ref
+                   :dirname (uiop:xdg-config-home +data-root+ ref))))
 
 (define-class gtk-download (download)
   ((gtk-object)
@@ -214,15 +269,33 @@ not return."
                                                   (uiop:xdg-data-home +data-root+ "gtk-extensions")
                                                   (dirname path)))))
 
+(defun cookies-data-path-for-context (name)
+  (make-instance 'cookies-data-path
+                   :ref (format nil "~a-web-context-cookies" name)
+                   :dirname (uiop:xdg-data-home +data-root+
+                                                 (format nil "~a-web-context" name))))
 
-(defun make-web-view (&key context-buffer)
+(defun make-web-view (&key buffer context-name ephemeral-p)
   "Return a web view instance.
-When passed a web buffer, create a buffer-local web context.
-Such contexts are not needed for internal buffers."
-  (if context-buffer
-      (make-instance 'webkit:webkit-web-view
-                     :web-context (make-context context-buffer))
-      (make-instance 'webkit:webkit-web-view)))
+
+If passed a context-name, a `nyxt:webkit-web-context' with that name is used for
+the `webkit:webkit-web-view'.  Otherwise, if :buffer is an internal-buffer or is
+not set, the browser's \"internal\" `nyxt:webkit-web-context' is
+used. Otherwise (such as an external web buffer), the \"default\"
+webkit-web-context is used,
+
+If ephemeral-p is set, the buffer is a nosave-buffer, or the current
+data-protife is a nosave-data-profile, then an ephemeral context is used, with
+the same naming rules as above."
+  (let ((internal-p (or (not buffer)
+                        (internal-buffer-p buffer)))
+        (ephemeral-p (or ephemeral-p
+                         (typep (current-data-profile) 'nosave-data-profile)
+                         (typep buffer 'nosave-buffer))))
+    (make-instance 'webkit:webkit-web-view
+                   :web-context (get-context *browser* (or context-name
+                                                           (if internal-p "internal" "default"))
+                                             :ephemeral-p ephemeral-p))))
 
 (defun make-decide-policy-handler (buffer)
   (lambda (web-view response-policy-decision policy-decision-type-response)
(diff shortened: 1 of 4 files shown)


tiberious726:
Note that this doesn't migrate user's old cookies (or other data-manager data) to the new default context 
(moving xdg-data-home/nyxt/cookies.txt to xdg-data-home/nyxt/default-web-context/cookies.txt should preserve the old cookies, if that's desired)

jmercouris:
Thank you very much for this detailed investigation and fix. I will try to go over it on Monday :-)

source/renderers/gtk.lisp
```diff
@@ -29,12 +29,26 @@ want to change the behaviour of modifiers, for instance swap 'control' and
     (delete nil (mapcar (lambda (mod) (getf plist mod)) modifier-state))))
 
 \(define-configuration browser
-  ((modifier-translator #'my-translate-modifiers)))"))
+  ((modifier-translator #'my-translate-modifiers)))")
+   (web-contexts (make-hash-table :test 'equal)
```
jmercouris, line 33: Why two separate hash tables?

source/renderers/gtk.lisp
```diff
@@ -201,6 +250,12 @@ not return."
   "We shouldn't enable (possibly) user-identifying extensions for `nosave-data-profile'."
   nil)
 
+(defun gtk-extensions-data-path-for-context (name)
```
jmercouris, line 253: by removing the slot and only including the function, it is now no longer possible for users to customize, or am I missing something?

jmercouris: s/users to customize/ users to easily customize

jmercouris:
I agree with most of the changes, that said, I would also like @Ambrevar's opinion :-)

thank you once again!

source/renderers/gtk.lisp
tiberious726: At least for WebKit (and I'm pretty sure for WebEngine too) contexts, ephemeral or not is the only fundamental division in web-contexts. Since there is just one division, keeping ephemeral and persistent web-contexts in two entirely separate containers seemed cleaner and less error-prone than doubly-keying a hash-table or relying on some keying convention.

I had initially tried to put all the contexts in a set and lean on the type system ( https://github.com/atlas-engineer/nyxt/pull/1757/files#diff-aa8247094d0aaa4eee5dda75fff637da81274cebd72dda88a004e381388e5d3dR106 ) to get ephemeral ones when needed. But I don't think it's possible to mess with the construction parameters of a class with the metaclass of gobject:gobject-class in a subclass (or there is a particular trick to specializing make-instance instead of initialize-instance that I haven't figured out yet). I didn't want to rely on the type system to tell me if a buffer is ephemeral or not without enforcing that the information the type system has is actually correct (trying to code as defensively as possible, as it's a security/privacy issue).

A nice effect of doing it this way is that keying the contexts by their name-string enforces that different contexts actually persist their data in different directories (and all a user has to do use a particular sub-folder of their nyxt data directory is give make-web-context that string)

I expect these containers to remain small enough that an alist would probably technically be a better data structure for them, but I didn't see those used elsewhere in the codebase.

source/renderers/gtk.lisp
tiberious726: Correct. This is true of all of the data-paths used for a web-context. They can still be "easily" overridden with the ref command-line argument, tho you do have to use the generated name (eg default-web-context, internal-web-context, etc).  Obviously, removing this means of customization isn't ideal, nor what I want to do overall, but I was trying to focus this patch on fixing the current default behavior and sort out the user-facing interface later (as it would be largely new functionality).

Leaving these slots in the buffer class would only serve to confuse users, unless we did something nuts, like assigning buffers to a context based on matching all of the individual data-paths (and generating the context-names if there isn't a match).  I also think its asking for trouble to have these slots writable but having absolutely no impact on the actual web-context after creation (you can currently rebind the data-path on any buffer, but if it already has a context that won't do a thing, there is also currently no validation that the slot data-paths actually represent what the web-context is using). The nasty functions here: https://github.com/atlas-engineer/nyxt/pull/1757/files#diff-aa8247094d0aaa4eee5dda75fff637da81274cebd72dda88a004e381388e5d3dR95 are a start to making sure our state is consistent with the CFFI objects.

If we want to keep this _easily_ customizable from lisp in the meantime, we could do the same thing I did with the default-cookie-policy, and move the slots (with different names, something like default-base-data-manager-data-path-suffix and default-base-data-manager-data-path) up to the browser class. Since we are generating the directories based on the context name, we'd have to change up the functions to incorporate the what the user customizes (the suffix). This gets pretty messy pretty quickly, and I think its better to sort it out properly. This is also a user-facing API change (tho I am already doing it with the cookie policy).

I believe the right way to solve this is to introduce a user-facing class, called something along the lines of browsing-context or browsing-session (both potentially confusing, I can see why Firefox called this "Accounts", but that name is more confusing/even less descriptive to users) outside of the gtk render, and let users apply their customization there. This would also be what they refer to for `make-buffer-in-session`, or however we wind up implementing it.

tiberious726:
Absolutely! Happy to help! I haven't had positive feelings about a browser since conkeror

Another thing to check is to make sure none of the documentation still says each buffer gets its own context. Which looks like it was true at some point in the past (and intentionally changed so users could share a session across buffers (I can't actually find the link right now, but I recall you saying something about logins)), _technically_ true before this patch, even tho the contexts shared the same data through disk. and no longer true in any sense (tho once the user-facing session stuff is in place, this could easily be a mode (tho picking where to persist the contexts if not ephemeral and when to reuse them on later start-ups are complex questions)). I thought I had seen another docstring that claimed as much while making these changes, but don't seem to be able to find it anymore

tiberious726:
Realized we could still get ephemeral contexts typed (and verified to be correct), by signaling an error condition if we somehow produce an incorrect one (not as clean as ensuring they are always constructed properly in the first place, but still better than nothing).

Note that these conditions are thrown by the gtk thread, which, at least for me, causes nyxt to exit by default (not in line with nyxt-condition documentation).

I reused the same condition for web-contexts and data-managers---don't want this to be too specific to WebKit.

I also made those aforementioned nasty functions less nasty, which could easily be backported if we don't want to go with typing the ephmeral contexts/data-managers

tiberious726:
Ephemeral web-views are now typed and checked in the same way---should make it easy for nosave-buffers and so on to make sure they actually are ephemeral

Ambrevar:
Fascinating contributions, @tiberious726 !
It's also a massive one so I need more time and a closer look before I can comment on this.  Will come back to it as soon as possible!

source/renderers/gtk.lisp
```diff
@@ -29,12 +29,26 @@ want to change the behaviour of modifiers, for instance swap 'control' and
...
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
 (define-user-class browser (gtk-browser))
 
+(defmethod get-context ((browser gtk-browser) name &key ephemeral-p)
+  (if ephemeral-p
+      (or (gethash name (ephemeral-web-contexts browser))
+          (setf (gethash name (ephemeral-web-contexts browser))
+                (make-context name :ephemeral-p t)))
```
Ambrevar, line 47: Isn't there an existing idiom for this "maybe set hash entry"?  Maybe in Alexandria or Serapeum?

source/renderers/gtk.lisp
```diff
@@ -83,6 +92,40 @@ data-manager will store the data separately for each buffer.")
...
+
+(defmethod initialize-instance :after ((web-context webkit-web-context-ephemeral) &key)
+  (unless (webkit:webkit-web-context-is-ephemeral web-context)
+    (error 'nyxt-web-context-condition :context web-context
+                                       :message "Web Contexts of class webkit-web-context-ephemeral must be ephemeral")))
+
+(defclass webkit-website-data-manager (webkit:webkit-website-data-manager) ()
+  (:metaclass gobject:gobject-class))
+
+;; create-gobject-from-class in gobject.base.lisp directly reads the initargs to make-instance, so this cannot
```
Ambrevar, line 114: Use quotes to refer to symbol names, here "`create-gobject-from-class'".
Also end all sentences with a period.

Ambrevar: What is "this" in "this cannot be used..."?

source/renderers/gtk.lisp
```diff
@@ -182,15 +225,37 @@ not return."
...
+   (ref :initform "web-context"))
   (:export-class-name-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
 
 (defmethod expand-data-path ((profile nosave-data-profile) (path data-manager-data-path))
   "We shouldn't store any `data-manager' data for `nosave-data-profile'."
   nil)
 
+(defun data-manager-data-path-for-context (name)
+  (let ((ref (format nil "~a-web-context" name)))
```
Ambrevar, line 238: Maybe `str:concat` instead of `format`?

source/renderers/gtk.lisp
```diff
@@ -182,15 +225,37 @@ not return."
...
+    (make-instance 'data-manager-data-path
+                   :ref ref
+                   :dirname (uiop:xdg-data-home +data-root+ ref))))
+
+(define-class data-manager-cache-path (data-path)
+  ((dirname (uiop:xdg-cache-home +data-root+ "web-context"))
+   (ref :initform "web-context"))
+  (:export-class-name-p t)
+  (:accessor-name-transformer (class*:make-name-transformer name)))
```
Ambrevar, line 247: Maybe add documentation to make explicit what's stored in this cache?

source/renderers/gtk.lisp
Ambrevar: You can also customize the data-paths in the init.lisp by overriding the expand path methods, right?

source/renderers/gtk.lisp
```diff
@@ -201,6 +266,12 @@ not return."
   "We shouldn't enable (possibly) user-identifying extensions for `nosave-data-profile'."
   nil)
 
+(defun gtk-extensions-data-path-for-context (name)
+  (let ((ref (format nil "~a-gtk-extensions" name)))
```
Ambrevar, line 270: Same: `format` -> `str:concat`?

source/renderers/gtk.lisp
```diff
@@ -214,15 +285,43 @@ not return."
                                                   (uiop:xdg-data-home +data-root+ "gtk-extensions")
                                                   (dirname path)))))
 
+(defun cookies-data-path-for-context (name)
+  (make-instance 'cookies-data-path
+                   :ref (format nil "~a-web-context-cookies" name)
+                   :dirname (uiop:xdg-data-home +data-root+
+                                                 (format nil "~a-web-context" name))))
```
Ambrevar, line 292: Indentation is off. 
`format` -> `str:concat`.

source/renderers/gtk.lisp
```diff
@@ -214,15 +285,43 @@ not return."
...
-  (if context-buffer
-      (make-instance 'webkit:webkit-web-view
-                     :web-context (make-context context-buffer))
-      (make-instance 'webkit:webkit-web-view)))
+
+If passed a context-name, a `nyxt:webkit-web-context' with that name is used for
+the `webkit:webkit-web-view'.  Otherwise, if :buffer is an internal-buffer or is
+not set, the browser's \"internal\" `nyxt:webkit-web-context' is
+used. Otherwise (such as an external web buffer), the \"default\"
+webkit-web-context is used,
```
Ambrevar, line 309: `,` -> `.` at the end.

source/renderers/gtk.lisp
```diff
@@ -214,15 +285,43 @@ not return."
...
   "Return a web view instance.
-When passed a web buffer, create a buffer-local web context.
-Such contexts are not needed for internal buffers."
-  (if context-buffer
-      (make-instance 'webkit:webkit-web-view
-                     :web-context (make-context context-buffer))
-      (make-instance 'webkit:webkit-web-view)))
+
+If passed a context-name, a `nyxt:webkit-web-context' with that name is used for
+the `webkit:webkit-web-view'.  Otherwise, if :buffer is an internal-buffer or is
```
Ambrevar, line 306: Remove `Otherwise` here since there is another `Otherwise` afterwards, and this clause has an `if`.
Also double-spacing is inconsistent, but we don't have a strong policy.

source/renderers/gtk.lisp
```diff
@@ -601,40 +700,45 @@ See `gtk-browser's `modifier-translator' slot."
...
-(defun make-context (&optional buffer)
-  ;; This is to ensure that paths are not expanded when we make
-  ;; contexts for `nosave-buffer's.
-  (with-current-buffer buffer
-    (let* ((manager (make-data-manager buffer))
-           (context (make-instance 'webkit-web-context :website-data-manager manager))
-           (cookie-manager (webkit:webkit-web-context-get-cookie-manager context))
-           (extensions-path (expand-path (gtk-extensions-path buffer))))
-      (when extensions-path
+(defun construct-web-context (name &key ephemeral-p)
```
Ambrevar, line 703: We don't use "construct" anywhere in our code base I believe.
We prefer the more succint "make".
But then we would have "make-web-context" and "make-context", which is ambiguous.
Can we come up with a better name?

Ambrevar (commented):
Great stuff, but I need a bit more time still to review it all and test it.

Question: If I got it right, we can no longer configure the cookie path individually, is that right?  I guess it's fine and it makes more sense to have it together with the other persistent data.  Is this your intention?

source/renderers/gtk.lisp
```diff
@@ -182,15 +225,37 @@ not return."
...
+
+(defmethod expand-data-path ((profile nosave-data-profile) (path data-manager-cache-path))
+  "We shouldn't store any `data-manager-cache' data for `nosave-data-profile'."
+  nil)
+
+(defun data-manager-cache-path-for-context (name)
+  (let ((ref (format nil "~a-web-context" name)))
+    (make-instance 'data-manager-cache-path
+                   :ref ref
+                   :dirname (uiop:xdg-cache-home +data-root+ ref))))
```
Ambrevar, line 257: Same here, you probably want to specialize `expand-data-path`.

source/renderers/gtk.lisp
```diff
@@ -214,15 +285,43 @@ not return."
...
+  (let ((internal-p (or (not buffer)
+                        (internal-buffer-p buffer)))
+        (ephemeral-p (or ephemeral-p
+                         (typep (current-data-profile) 'nosave-data-profile)
+                         (typep buffer 'nosave-buffer))))
+    (make-instance (if ephemeral-p
+                       'webkit-web-view-ephemeral
+                       'webkit:webkit-web-view)
+                   :web-context (get-context *browser* (or context-name
+                                                           (if internal-p "internal" "default"))
```
Ambrevar, line 323: These "internal" and "default" string are prone to typos.  Ideally, we should type these.
The most immediate fix is to use constant, e.g. `+internal-context+`.

source/renderers/gtk.lisp
```diff
@@ -601,40 +700,45 @@ See `gtk-browser's `modifier-translator' slot."
...
-         :webkit-cookie-persistent-storage-text)
-        (set-cookie-policy cookie-manager (default-cookie-policy buffer)))
-      context)))
+         cookies-data-path
+         :webkit-cookie-persistent-storage-text))
+      (set-cookie-policy cookie-manager (default-cookie-policy *browser*)))
+    context))
+
+(defun internal-context-p (name)
+  (equal name "internal"))
```
Ambrevar, line 741: Same as above, the "internal" string is brittle.  For the lack of better typing, a constant would be more robust, then this functions could be replaced by `(equal +internal-context+ name)`.

Ambrevar:
I've tested this is and works well for me so far.

Beside the minor nits I've commented above, my main concern is about the ability to map buffer to specific context.
Previous, we could have one context per buffer (even though it was a bit broken).  We will be able to keep this feature with this change?  In other words, will we still be able isolate buffers?

`make-web-view` accepts a context name, so I suppose this is the function we can leverage to make isolated buffers or, as you mentioned, "Firefox multi-account containers".  Is this correct?

source/renderers/gtk.lisp
jmercouris: Or you could use a method with equal specialization!

Ambrevar:
I'll rebase, test and merge in the coming days.  Let me know if there is anything blocking this merge.

Ambrevar:
Closing in favour of #1987.

PR #1759, opened 2021-08-30
Documentation: add a section about the four spelling check commands

pdelfino:
The app **Grammarly** also suggested inserting the article *the* in one of the docstring, so I did.

### DIFF
--- a/source/tutorial.lisp
+++ b/source/tutorial.lisp
@@ -213,6 +213,14 @@ buffer.")
                                 nyxt/web-mode:go-previous
                                 nyxt/web-mode:go-up
                                 nyxt/web-mode:go-to-homepage)))
+   (:h3 "Spelling check")
+   (:p "Several commands are provided to spell check words. The default is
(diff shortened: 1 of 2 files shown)


source/tutorial.lisp
```diff
@@ -213,6 +213,13 @@ buffer.")
                                 nyxt/web-mode:go-previous
                                 nyxt/web-mode:go-up
                                 nyxt/web-mode:go-to-homepage)))
+   (:h3 "Spelling check")
+   (:p "Several commands are provided to spell check words in English:")
```
jmercouris, line 217: Can be any language, see the browser slot `spell-check-language`.

pdelfino: That's correct. My bad. English is just the default.

 Actually, something occurred to me. The slots and the possible tinkering with them **deserve their own section**  in the manual with a few illustrative (non-exhaustive) examples showing the role they have in Nyxt's UX (architecture also) and how to change them.

PR #1763, opened 2021-08-30
gtk.lisp: create outside-renderer-thread

jmercouris:
To run javascript and return its value successfully it has to be run on a
thread outside of the main thread. Why? Because of this code:

```
(defmethod ffi-buffer-evaluate-javascript ((buffer gtk-buffer) javascript &optional world-name)
  (within-renderer-thread
   (lambda (&optional channel)
     (webkit2:webkit-web-view-evaluate-javascript
      (gtk-object buffer)
      javascript
      (if channel
          (lambda (result jsc-result)
            (declare (ignore jsc-result))
            (calispel:! channel result))
          (lambda (result jsc-result)
            (declare (ignore jsc-result))
            result))
      (lambda (condition)
        (javascript-error-handler condition)
        ;; Notify the listener that we are done.
        (when channel
          (calispel:! channel nil)))
      world-name))))
```

note the line that says `if channel`. Looking at the definition for `within-renderer-thread` we see the problem:

```
(defun within-renderer-thread (thunk)
  "If the current thread is the renderer thread, execute THUNK with `funcall'.
Otherwise run the THUNK on the renderer thread by passing it a channel and wait on the channel's result."
  (if (renderer-thread-p)
      (funcall thunk)
      (let ((channel (make-channel 1)))
        (within-gtk-thread
          (funcall thunk channel))
        (calispel:? channel))))
```

a channel is only created and blocking IFF we are running on a non-renderer thread. If we try to make a channel regardless of whether we are on a renderer thread we find ourselves in a starvation state.

Consider:

Thread 1 says "Hello, I am running some code `within-renderer-thread`". It is running our `webkit:webkit-web-view-evaluate-javascript` function. This function accepts a callback. This callback also runs on the renderer thread. Therefore in order for Thread 1 to complete its `within-renderer-thread` call, it must have the callback update the channel which is also waiting for the same `renderer` thread.

It is for this reason that `on-signal-load-finished` is now executed outside of the renderer thread. Otherwise it would not be able to return any values from JS.

### DIFF
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -109,7 +109,7 @@ See https://github.com/atlas-engineer/nyxt/issues/740")
   (within-gtk-thread
     (funcall thunk)))
 
-(defun %within-renderer-thread (thunk)
+(defun within-renderer-thread (thunk)
   "If the current thread is the renderer thread, execute THUNK with `funcall'.
 Otherwise run the THUNK on the renderer thread by passing it a channel and wait on the channel's result."
   (if (renderer-thread-p)
@@ -119,8 +119,12 @@ Otherwise run the THUNK on the renderer thread by passing it a channel and wait 
           (funcall thunk channel))
         (calispel:? channel))))
 
-(defun %within-renderer-thread-async (thunk)
-  "Same as `%within-renderer-thread' but THUNK is not blocking and does
+(defun outside-renderer-thread (thunk)
+  "Execute a thunk within a thread that is not going to be the renderer thread."
+  (bt:make-thread (lambda () (funcall thunk))))
+
+(defun within-renderer-thread-async (thunk)
+  "Same as `within-renderer-thread' but THUNK is not blocking and does
 not return."
   (if (renderer-thread-p)
       (funcall thunk)
@@ -260,7 +264,7 @@ The BODY is wrapped with `with-protect'."
        (push handler-id (handler-ids ,object)))))
 
 (defmethod initialize-instance :after ((buffer status-buffer) &key)
-  (%within-renderer-thread-async
+  (within-renderer-thread-async
    (lambda ()
      (with-slots (gtk-object) buffer
        (setf gtk-object (make-web-view))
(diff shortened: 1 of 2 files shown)


source/renderers/gtk.lisp
```diff
@@ -119,8 +119,12 @@ Otherwise run the THUNK on the renderer thread by passing it a channel and wait
           (funcall thunk channel))
         (calispel:? channel))))
 
-(defun %within-renderer-thread-async (thunk)
-  "Same as `%within-renderer-thread' but THUNK is not blocking and does
+(defun outside-renderer-thread (thunk)
```
aartaka, line 122: But wait, isn't that just a simpler and less safe version of `run-thread`? Why introduce a new function? Yes, that's a piece of documentation, but a comment in where it's immediately necessary serves as a better documentation.

jmercouris: Technically the same, but the intent is different. I will think about it.

Ambrevar:
> Thread 1 says "Hello, I am running some code within-renderer-thread". 

Then I guess "Thread 1" is the GTK thread, right?

Ambrevar:
Seems legit to me, but I'd like to tinker with it a little bit, maybe we can come up with something more robust.

We should also extend the documentation, this is tricky to understand as it is.

jmercouris:
@Ambrevar Merge?

source/renderers/gtk.lisp
jmercouris: I have thought, and I must say, I disagree. The naming/intent here is important enough. it could wrap run-thread for now, but I would not like to replace calls to this with run-thread.

Ambrevar:
This patch works but has one design issue as I understand it: it calls `outside-renderer-thread` locally.

I thought about it a little bit more.  Last time we discussed, we mentioned that we could rather run all signal bodies in a new thread.  Then we supposed it wouldn't be ideal because the user might assume to be on the GTK thread.

This last sentence, however, might be mistaken.
If I'm not wrong, we should _never_ need to assume that we are on the GTK thread.  Indeed, this is precisely what `%within-renderer-thread` is for and this helper is used at the appropriate spots in gtk.lisp (e.g. `ffi-buffer-evaluate-javascript`).  So at any point any time calling `ffi-buffer-evaluate-javascript` should work, regardless of the thread we are currently in.

I propose the following design: As much as possible of Nyxt code, and in particular all of Nyxt extension points, should be run from a thread that's not the GTK thread.  This would make the model much simpler and fix the issue from the original post.

@jmercouris: Could you try it out and send a separate pull request so that we can compare?

jmercouris:
OK, will do!

Ambrevar:
@jmercouris Have you had time to try my suggestion?
If you want me to do it, just let me know :)

Issue #1570, opened 2021-06-30
Generate cheatsheets


Ambrevar:
We can already print any page to PDF, so we can print the `describe-bindings` buffer.  But it's not space-optimized.

Can we make it so that it fits into the smallest number of pages?  Maybe print the bindings in a 2-column table and shrink the font?

pdelfino:
PR #1764 was merged. Thus, I guess this issue could be closed. I do not have the "right" to do so, though. I can only close my own issues.

Could you do the trick @jmercouris?

It would be nice to check with Pierre if the implementation really met his original. But, if it did not, we can always re-open it and fixed it.

PR #1764, opened 2021-08-30
help: create a command to print the bindings using the space wisely via a tweak in CSS.

pdelfino:
This PR **tries** to solve issue #1570. I created a new command: `print-bindings-cheatsheet`. The pdf output is attached [here](https://github.com/atlas-engineer/nyxt/files/7079949/output.pdf).

Honestly, I am **not sure** my approach is a good one.
Limitations:
 - it is a cheatsheet tailor-made to keybindings. Seems like something too specific for a command;
 - it re-writes the string wit CSS code entirely inside the `setf`. I guess I can just append the new stuff to the string, this would save some lines of code;
 - the output achieves the goal of reducing paper use, but is it too small to read?
 
 Alternative:
 - I will submit another PR with an alternative approach. This one uses CSS to shrink the font and change the display of the tables. The other will focus on **HTML** and change the `describe-bindings` code, building a different table with spinneret.

### DIFF
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -383,6 +383,17 @@ CLASS is a class symbol."
                                       (:td keyspec)
                                       (:td (string-downcase bound-value)))))))))))
 
+(define-command print-bindings-cheatsheet ()
+  "Print the buffer with the list of all known bindings for the current buffer
+optimizing the use of space."
+  (nyxt::html-set-style 
+   (cl-css:css '((h3 :font-size "10px" :font-family Helvetica Neue Helvetica
+                     :font-weight 500)
+                 (tr :font-size "7px")
+                 (div :display inline-block)))
+   (nyxt:describe-bindings))
+  (print-buffer))
+
 (defun tls-help (buffer url)
   "This function is invoked upon TLS certificate errors to give users
 help on how to proceed."


source/help.lisp
```diff
@@ -383,6 +383,33 @@ CLASS is a class symbol."
                                       (:td keyspec)
                                       (:td (string-downcase bound-value)))))))))))
 
+(define-command print-bindings-cheatsheet ()
+  "Print the buffer with the list of all known bindings for the current buffer
+optimizing the use of space."
+  (nyxt:describe-bindings)
+  (let ((original-style (style (current-buffer))))
+    (setf (style (current-buffer))
+          "body { margin-left: 20px; margin-top: 20px; }
```
jmercouris, line 392: Use cl-css instead of raw CSS string

source/help.lisp
```diff
@@ -383,6 +383,33 @@ CLASS is a class symbol."
                                       (:td keyspec)
                                       (:td (string-downcase bound-value)))))))))))
 
+(define-command print-bindings-cheatsheet ()
+  "Print the buffer with the list of all known bindings for the current buffer
+optimizing the use of space."
+  (nyxt:describe-bindings)
```
jmercouris, line 389: describe-bindings returns a buffer, this is the buffer you can apply your style to

jmercouris:
```
(let ((my-buffer (nyxt:describe-bindings)))
  (html-set-style (cl-css: xyz) my-buffer))
```

pdelfino:
Hi guys,

I think this is good to go, now :wink:

jmercouris:
Thank you!

PR #1770, opened 2021-09-03
configuration: Warn instead of ignore errors in load-after-system

hendursaga:


source/configuration.lisp
```diff
@@ -283,11 +283,16 @@ Initialization file use case:
...
-            (ql:quickload system :silent t)
-            #-quicklisp
-            (asdf:load-system system))))
+           (handler-case
+               (progn
+                 #+quicklisp
+                 (ql:quickload system :silent t)
+                 #-quicklisp
+                 (asdf:load-system system))
+             (error (c)
```
Ambrevar, line 292: I you don't want to use `c`, don't pass anything (e.g. `(error ()`).
But it could be useful to report what's going on in the `echo-warning` using the `~a` format parameter.

hendursaga: I'm new to Git(Hub), you'd want to squash these two commits then merge, correct?

Ambrevar: Yes, but no worries, this small commit does not hurt either.

Issue #1644, opened 2021-07-20
Feature request: `no-procrastinate-mode` create a mode that blocks access for distractive websites


pdelfino:
This mode would be the first step towards a mode similar to a firefox extension called [Leechblock](https://addons.mozilla.org/en-US/firefox/addon/leechblock-ng/). 

Basically, the user would define a list of "cursed" websites. Access to these websites would be blocked if `noprocrastinate-mode` is active.

I was thinking of using `force-https-mode` as a good reference and pitched the idea to @jmercouris.  He encouraged the implementation, mentioned `certificate-exception-mode` as a possible reference, and said this idea would be, actually, an **instance** of `blocker-mode`.

In the future, this feature could be improved so that the user could define **conditions** for `noprocrastinate-mode`. For instance, allowing access to "cursed" websites (such as hacker news) only after 6 PM and for only 30 minutes, instead of simply blocking it.

This feature leverages productivity, which is the old focus of Nyxt's tagline but is still part of its value proposition.

As a side note, Paul Graham has [a great essay](http://www.paulgraham.com/addiction.html) written more than 10 years ago showing why the problem this feature address is relevant.

pdelfino:
I am closing this issue since the feature was implemented 4b4f7af7cd2bded331ede347436af4833fd8a9ea

PR #1771, opened 2021-09-03
Feature `no-procrastinate-mode`

pdelfino:
This is a **working in progress** for issue #1644. There are still some doubts and additions needed. 

The doubts were written as comments inside the problematic part.


jmercouris:
Hi Pedro, would you like to consider revisiting this?

pdelfino:
Hi guys,

There are three things to solve here:
1 - if the user has never added a host to avoid procrastination, when they enable `no-procrastinate-mode` it throws an error. The initialization needs to be fixed. It will work fine if a `.local/share/nyxt/no-procrastinate.lisp` already exists with proper data.

2 - not sure if there should be a panel buffer on this.

3 - while rebasing with master and then doing an interactive rebase, I am affraid I messed too much with history. There are few commits ids, but the associated diffs are not representative. If it not possible/easy to fix, I will just close this PR and submit a new one with a better history.

I will talk tomorrow with @jmercouris about these 3 points. But feel free to start a review.

source/no-procrastinate.lisp
```diff
@@ -0,0 +1,281 @@
...
+          (format t "~s" (first (tags entry)))
+          (dolist (tag (rest (tags entry)))
+            (write-string " ")
+            (write tag))
+          (write-string ")"))
+        (write-string ")")))))
+
+(defmethod nfiles:serialize ((profile nyxt-profile) (file no-procrastinate-hosts-file) stream &key)
+  (let ((content
+          (sort (nfiles:content file)
```
jmercouris, line 258: What is the purpose of the sort here? If it is not necessary, I would remove it because it adds computation, and may confuse users. They may expect that the no procrastinate hosts _must_ be sorted.

source/urls.lisp
```diff
@@ -52,6 +52,11 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
          (or (ignore-errors (ffi-display-url url))
              url))))
 
+(defun render-url-without-path (url)
+  "Return decoded URL without path, if existent."
```
jmercouris, line 56: The name is a bit confusing. What does it mean without a path? I would change the name to `render-host-&-scheme`. Because, that is exactly what is happening here.

nyxt.asd
```diff
@@ -24,7 +24,7 @@
 (defvar *submodules-jobs* (or (uiop:getenv "NYXT_SUBMODULES_JOBS")
                               4)
   "Number of parallel 'git clone' jobs to fetch the Git submodules.
-A naive benchmark on a 16 Mpbs bandwidth gives us
+A naive benchmark on a 16Mpbs bandwidth gives us
```
jmercouris, line 27: I don't believe this change belongs in this PR.

source/buffer.lisp
```diff
@@ -350,7 +355,8 @@ The file where the system will create/save the input data.")
    (auto-mode-rules-file
     (make-instance 'auto-mode-rules-file)
     :type auto-mode-rules-file
-    :documentation "The file where the auto-mode rules are saved.")
+    :documentation "
```
jmercouris, line 358: I believe you should undo this change.

source/mode/no-procrastinate.lisp
```diff
@@ -0,0 +1,78 @@
...
+             :outline          "none"
+             :text-align       "left")))
+   (nyxt/blocker-mode:hostlists
+    (list (nyxt/blocker-mode:make-hostlist
+           :hosts (mapcar #'(lambda (y) (hostname y))
+                          (nfiles:content (no-procrastinate-hosts-file (current-buffer)))))
+          nyxt/blocker-mode:*default-hostlist*))))
+
+(defun group-no-procrastinate-hosts (buffer)
+  (let ((no-procrastinate-hosts-table (make-hash-table :test #'equalp)))
```
jmercouris, line 32: I would see if it is possible to collapse the let's into a single let. I'm not sure that it makes sense to do things this way.

source/mode/no-procrastinate.lisp
```diff
@@ -0,0 +1,78 @@
...
+             :margin           "6px"
+             :width            "100%"
+             :border           "none"
+             :outline          "none"
+             :text-align       "left")))
+   (nyxt/blocker-mode:hostlists
+    (list (nyxt/blocker-mode:make-hostlist
+           :hosts (mapcar #'(lambda (y) (hostname y))
+                          (nfiles:content (no-procrastinate-hosts-file (current-buffer)))))
+          nyxt/blocker-mode:*default-hostlist*))))
```
jmercouris, line 29: I don't believe it makes sense ot include the *default-hostlist*. Are there a set of default websites that could be considered timewasters? I guess you could put reddit, ycombinator, youtube, etc.

source/no-procrastinate.lisp
```diff
@@ -0,0 +1,281 @@
...
+  ((nfiles:base-path #p"no-procrastinate-hosts")
+   (nfiles:name "no-procrastinate-hosts"))
+  (:export-class-name-p t)
+  (:accessor-name-transformer (class*:make-name-transformer name)))
+
+(define-class no-procrastinate-entry ()
+  ((url (quri:uri ""))
+   (hostname "")
+   (title "")
+   (annotation "")
```
jmercouris, line 28: I would remove this slot since it is unused. Also remove all references to it in the codebase.

source/no-procrastinate.lisp
```diff
@@ -0,0 +1,281 @@
...
+    ;; so if we used `delete-duplicates' here it would have modified the last
+    ;; list.
+    (let ((tags (sort (remove-duplicates
+                       (apply #'append
+                              (mapcar #'tags no-procrastinate-hosts-list))
+                       :test #'string-equal)
+                      #'string-lessp)))
+      tags)))
+
+(defun last-word (s)
```
jmercouris, line 95: This exists also in `bookmark.lisp`. I would pull this function into a top-level utility file. Or user the `last-word` defined in the `nyxt/bookmark` package.

source/no-procrastinate.lisp
```diff
@@ -0,0 +1,281 @@
...
+page(s) in the active buffer."
+  (prompt
+   :prompt "Avoid procrastination on HOSTS from buffer(s)"
+   :sources (make-instance 'user-buffer-source
+                           :multi-selection-p t
+                           :actions (list (make-mapped-command no-procrastinate-current-host)))))
+
+
+(define-command no-procrastinate-host (&key url)
+  "Allow the user to avoid procrastination on a HOST via minibuffer input."
```
jmercouris, line 172: prompt buffer*

source/no-procrastinate.lisp
```diff
@@ -0,0 +1,281 @@
...
+  (let* ((url (or url
+                  (ignore-errors
+                   (quri:uri
+                    (first
+                     (prompt
```
jmercouris, line 177: you can use prompt1 instead of `(first (prompt`

PR #1773, opened 2021-09-05
manual: insert section to illustrate slot tinkering.

pdelfino:


source/manual.lisp
```diff
@@ -46,6 +46,25 @@ class and slot documentation for the individual details.")
...
+    (:ul
+     (:li "Execute command " (command-markup 'describe-slot))
+     (:li "Type " (:code 'hints-alphabet))
+     (:li "Select " (:code "hints-alphabet") " (" (:code "user-web-mode") " class option).")
+     (:li "Press the button " (:code "Configure"))
+     (:li "Insert the string \"asfdghjkl\""))
+    (:p "This will make link-hinting more comfortable for this user. As
+non-exhaustive examples, a similar approach of customization can be applied to
+other slots such as " (:code "spell-check-language") ", which will expand the
+spelling-check feature to other languages besides English. Or changing slot "
```
jmercouris, line 65: I believe two examples should suffice :-)

pdelfino: Done!

pdelfino:
I think it is good to go!

Besides removing one of the examples, I made the HTML list an ordered one, instead of unordered.

jmercouris:
Thank you!

PR #1777, opened 2021-09-08
Improve annotations.lisp

pdelfino:
Hi this patch improves annotations on the following points:

- inserts the data from the web pages' title;
- changes the message on prompt-buffer from generic **Input** to more specific **Note**;
- Rename commands `show-annotations-for-current-buffer` to `show-annotations-for-current-url`.


source/annotate.lisp
```diff
@@ -10,7 +10,8 @@
    (tags
     '()
     :type list-of-strings)
-   (date (local-time:now)))
+   (date (local-time:now))
+   (page-title ""))
```
jmercouris, line 14: I believe the page-title should be part of the url-annotation, consider things that are annotated which do not have titles, such as local documents.

pdelfino: Correct, I am fixing it.

There is already a problem with this when the users makes an annotation  in a buffer such as the Help buffer

There is a title, a snippet, and an annotation. But there is no URL, which results in:

```txt
Snippet: Full documentation about Nyxt, how it works and how to configure it.

URL:

Title: *Help*

Annotation: snippet highlighted and annotated in the initial buffer when starting Nyxt; let's see what happens.

Tags: test
```

Not sure if this is a big problem, though.

pdelfino: The example above was taken in my branch development.

But the same problem happens on master. Hilightinh something on the starter page or help buffer generates this URL problem:


```txt
Snippet: List all bindings for the current buffer.

URL:

Annotation: testeBufferURL

Tags: test
```

PR #1780, opened 2021-09-09
buffer: Fix document-model :around method when DOM elements are NIL.

Ambrevar:
Previously reopening a dead buffer would raise a condition, this fixes it.

@jmercouris Any idea why `truncate` was used here?

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -600,8 +600,12 @@ Delete it with `ffi-buffer-delete'."
          (add-nyxt-identifiers child))
        nyxt-identifier-counter)
      (setf nyxt-identifier-counter (add-nyxt-identifiers (ps:chain document body)))))
-  (setf (document-model buffer)
-        (nyxt/dom::named-json-parse (nyxt/dom::get-document-body-json))))
+  (alex:when-let ((body-json (nyxt/dom::get-document-body-json)))
+    (setf (document-model buffer)
+          (nyxt/dom::named-json-parse body-json))))
+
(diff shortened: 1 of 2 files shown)


aartaka:
> @jmercouris Any idea why `truncate` was used here?

JS returns fractional values and `cl-webkit` rules for handling those can occasionally result in floats. Thus, we need to ensure integers on the Lisp side.

Ambrevar:
OK, I've updated the patch to keep the `truncate`.
OK to merge?

aartaka:
Yes, looks alright.

Ambrevar:
No, this is not the right fix.  reopen-buffer is broken, so is any buffer-related prompt buffer when a buffer is loading.
I'm working on a proper fix.

Ambrevar:
I've just pushed a fix for the above 2 issues.

That said, `reopen-buffer` not opens a buffer that's stuck in an infinite load.
Any idea what's going on?

@jmercouris ?

Ambrevar:
I've just fixed reopen-buffer.  Turns out that we didn't free `gtk-object`!!!
This is a critical fix.  Can you people test for a while and see if it triggers new issues on your end?

Ambrevar:
I've been using this with success for the last couple of days.  Let me know if anything is wrong on your end.
I'll forward-port this to master.

jmercouris:
In general I am not a huge fan of dead buffers, I find them to be a confusing interface element. All of that said, this patch appears correct to me.

Ambrevar:
Would you suggest another design to implement "reopen-buffer"?

I agree that we currently don't have a good abstraction for this use case.

Issue #1784, opened 2021-09-13
Link hint slow-down investigation


Ambrevar:
It seems that the slow down is still (!) `url :around` (in `dom.lisp`), more precisely this chunk:

```lisp
      (if (valid-url-p result)
          url
          (merge-url* url (url (current-buffer))))
```

1. If we just put `(merge-url* url (url (current-buffer)))` instead, it seems to work fine since WebKitGTK understands relative URL.  @aartaka why did you deem necessary to check if the URL is valid?  Is there another case?

2. We should make `valid-url-p` faster.  Currently it checks the DNS to see if a hostname exists.  Should we have two versions, or maybe an option to skip the domain name resolution for cases where we don't need it?

3. `(defmethod prompter:object-attributes ((a nyxt/dom:a-element))` calls `url` and then renders it.  Basically, this instantiates a `quri:uri` object from a string, only to re-render it right after.  This wastes precious CPU cycles, so we can skip the `url` call and directly return the `dom:a-element` href.

Thoughts?  @jmercouris @aartaka.

aartaka:
>     1. If we just put `(merge-url* url (url (current-buffer)))` instead, it seems to work fine since WebKitGTK understands relative URL.  @aartaka why did you deem necessary to check if the URL is valid?  Is there another case?

Well, I thought about opening things in a new buffer. If we open a relative URL in a new buffer, it will error out that URL is not right. Thus we need to complete it. However, `merge-url*` should do this check automatically, so it's really unnecessary.

>     2. We should make `valid-url-p` faster.  Currently it checks the DNS to see if a hostname exists.  Should we have two versions, or maybe an option to skip the domain name resolution for cases where we don't need it?

Yes we should! Sometimes what we need to check is the URI syntax correctness, not the existence of a location.

>     3. `(defmethod prompter:object-attributes ((a nyxt/dom:a-element))` calls `url` and then renders it.  Basically, this instantiates a `quri:uri` object from a string, only to re-render it right after.  This wastes precious CPU cycles, so we can skip the `url` call and directly return the `dom:a-element` href.

Agreed.

Ambrevar:
Fix merged, thanks Artyom for your feedback!

PR #1788, opened 2021-09-14
Fix slow hints

Ambrevar:
Fix all points mentioned in https://github.com/atlas-engineer/nyxt/issues/1784.

@aartaka @jmercouris Let me know if it looks OK to you.
Once approved, I'll forward-port this to master.

### DIFF
--- a/source/dom.lisp
+++ b/source/dom.lisp
@@ -180,9 +180,7 @@ JSON should have the format like what `get-document-body-json' produces:
                           (when (quri:uri-fragment url) `(:fragment ,(quri:uri-fragment url)))))))
     (alex:when-let* ((result (call-next-method))
                      (url (nyxt::ensure-url result)))
-      (if (valid-url-p result)
-          url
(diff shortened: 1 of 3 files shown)


aartaka:
Looks all fine to me!

I'm curious, though: can we speed up `merge-url*` somehow? Is it possible to avoid all the branching and copying here? I feel that it could become the next bottleneck.

shamazmazum:
Not asking DNS to resolve everything you are typing is super nice!

jmercouris:
Not as fast as my solution, but it is correct :-D

Thank you for this monsieur!

jmercouris:
(whereas my solution has bugs :-D)

Ambrevar:
@aartaka: Indeed, this `valid-url-p` with all the branching is still slow, but note that it's no longer used by element hints.

Let's optimize it when the need arises and `valid-url-p` is shown to be a bottleneck again.

PR #1792, opened 2021-09-16
Fix nosave history leak

### LINKED ISSUE #1717: history-tree leaks history to hard drive
When you call `history-tree` command in `nosave-buffer`, the history is leaked to the hard drive.

1. Remove `~/.local/share/nyxt/history` (or where else your history is stored) for a clean experiment.
2. Open some page on the net in `nosave-buffer`. Follow some links
3. Look at the history by calling `history-tree`.
4. Kill history buffer.
5. Now your history is leaked to the hard drive. You can check `~/.local/share/nyxt/history/default.lisp`.

**Information**
- OS name+version: Irrelevant
- Graphics card and driver: Irrelevant
- How you installed Nyxt (Guix pack, package manager, build from source): I am on commit `af599d8`. All dependencies are installed via ultralisp.
- SBCL 2.1.7

Maybe all internal buffers must behave like `nosave-buffer` (i.e. make a `nosave-mixin` class and inherit `nosave-buffer` and `internal-buffer` from it)?

Ambrevar:
Fixes #1717.

Still seems brittle.
Both our nosave and data-path architectures are complex, maybe we could do better.  Any suggestion?

@shamazmazum Can you confirm it fixes it for you?

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -892,12 +892,14 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
   "For dummy buffers, use `ffi-buffer-delete' instead."
   (hooks:run-hook (buffer-delete-hook buffer) buffer)
   (let ((parent-window (find buffer (window-list) :key 'active-buffer)))
-    (with-data-access (history (history-path buffer))
-      (sera:and-let* ((owner (htree:owner history (id buffer)))
-                      (current (htree:current owner))
-                      (data (htree:data current)))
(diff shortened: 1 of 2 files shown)


shamazmazum:
Strange, it seems to me the history is still saved when you invoke `list-bookmarks` command and close `*Bookmarks*` buffer.

Ambrevar:
Indeed, this patch is brittle because it is not a general fix.

Don't merge then, I'll try to think of something more general.

Ambrevar:
Closing in favour of #2072.

Issue #1739, opened 2021-08-16
Going backwards results in page not loading


joshcho:
**Describe the bug**

I am not sure if this is a known issue (feel free to close otherwise, I could not find this specific one).
Going backwards in history results in a page not loading, about a quarter of the time.

**Precise recipe to reproduce the issue**
Not sure exactly the conditions, but going backwards in history (and sometimes going forwards) results in a page not loading. The page loads well if it is refreshed.

For website-specific issues:
Can you reproduce this issue with Epiphany / GNOME Web (https://wiki.gnome.org/Apps/Web)?

**Information**
- OS name+version: Arch Linux 5.13.10-arch1-1
- Graphics card and driver: NVIDIA GeForce RTX 2070 Super, NVIDIA
- Desktop environment / Window manager name+version: xmonad 0.15
- How you installed Nyxt (Guix pack, package manager, build from source): from source, but same issue persists when installed from AUR
- Information from `show-system-information`:

Nyxt version: 2.1.1-524-g66ca21c8
Renderer version: GI-GTK
Operating system kernel: Linux 5.13.10-arch1-1
Lisp implementation: SBCL 2.1.7 (Dynamic space size: 1073741824)
Features: (:WEBKIT2 :WEBKIT2-2.32 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
 :GLIB-2-30 :GLIB-2-32 :GLIB-2-34 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42
 :GLIB-2-44 :GLIB-2-46 :GLIB-2-48 :GLIB-2-50 :GLIB-2-52 :GLIB-2-54 :GLIB-2-56
[...]

PR #1793, opened 2021-09-16
Workaround spurious "load-failed" events when loading webkit history pages.

Ambrevar:
This seems to be an upstream bug.  In the meantime, temporarily disable
load-failed when loading a page from the webkit history cache.

"Fixes" #1739.

@aartaka I can reproduce this "load-failed" bug with the cl-webkit demo.
Any idea why this happens?

### DIFF
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -756,6 +756,9 @@ See `gtk-browser's `modifier-translator' slot."
             (webkit:webkit-policy-decision-ignore response-policy-decision)
             nil)))))
 
+(defvar *loading-webkit-history-p* nil
+  "Internal hack, do not use!")
+
 (define-ffi-method on-signal-load-changed ((buffer gtk-buffer) load-event)
   ;; `url' can be nil if buffer didn't have any URL associated
   ;; to the web view, e.g. the start page, or if the load failed.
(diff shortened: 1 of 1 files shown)


Ambrevar:
Updated to remove the global variable.

@jmercouris Can you confirm it works for you?

source/renderers/gtk.lisp
```diff
@@ -69,7 +69,11 @@ See https://developer.gnome.org/gobject/stable/gobject-Signals.html#signal-memor
                       :documentation "Directory in which the WebKitGTK
 data-manager will store the data separately for each buffer.")
    (gtk-extensions-path (make-instance 'gtk-extensions-data-path)
-                        :documentation "Directory to store the WebKit-specific extensions in."))
+                        :documentation "Directory to store the WebKit-specific extensions in.")
+   (loading-webkit-history-p nil
+                             :type boolean
+                             :export nil
+                             :documentation "Internal hack, do not use me!"))
```
jmercouris, line 76: Perhaps the documentation should instead read:

```
"Internal hack, do not use me! WebKitGTK may trigger "load-failed" when loading a page from the webkit-history cache.  Upstream bug?  We use this slot to know when to ignore these load failures."
```

jmercouris:
I cannot confirm it works for me, but I had a lot of trouble recreating the issue! Therefore, I can assume it is fixed by the null hypothesis!

Fantastic quality of life improvement, thank you Pierre!

Ambrevar:
@jmercouris An easy reproduction recipe is to perform a DuckDuckGo search, click on a result link, then spam `history-backwards`-`history-forwards` in cycles.

source/renderers/gtk.lisp
Ambrevar: Agreed!

PR #1797, opened 2021-09-25
Fix upload of files whose names contain wildcard characters

shamazmazum:
Hello. This patch fixes upload of files with names like `[foo]bar.baz`.

### DIFF
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -877,7 +877,7 @@ See `gtk-browser's `modifier-translator' slot."
       (gobject:g-object-ref (gobject:pointer file-chooser-request))
       (run-thread
         (let ((files (mapcar
-                      #'namestring
+                      #'uiop:native-namestring
                       (handler-case
                           (prompt :prompt (format
                                            nil "File~@[s~*~] to input"


jmercouris:
Thank you for this patch! Do you have a URL I could use to test on my system?

shamazmazum:
You can test it with https://anyconv.com/, for example. You'll need a pdf file with the name `[test].pdf`. Press "Choose file" on that site and upload your pdf. It will not work until the patch is applied.

This is how it looks with the patch:
20210928_11h52m28s_grim

You can also use wireshark to see what's going on.

Ambrevar:
This looks correct, in general we should always use `uiop:native-namestring`.

Ambrevar:
@aartaka Any particular reason you've used `namestring`?
There are a few more instance of `namestring` in gtk.lisp but also in the rest of the code base.  I suggest we turn all of those which deal with file system paths to `uiop:native-namestring`.

aartaka:
> @aartaka Any particular reason you've used `namestring`? There are a few more instance of `namestring` in gtk.lisp but also in the rest of the code base. I suggest we turn all of those which deal with file system paths to `uiop:native-namestring`.

The reason is: I didn't know of `uiop:native-namestring` until now xD

Ambrevar:
@jmercouris Please merge then, then let's update all the path-related `namestring`.

Ambrevar:
OK to merge?

PR #1804, opened 2021-09-29
Encrypted Lisp URLs

aartaka:
This does quite a set of connected things. Notable highlights are:
- `lisp:` URLs are now handled by WebKit.
- There are no more `internal-buffer`s.
- `lisp:` URLs do only work with static page creation now.
  - The use-case for on-click code evaluation will be covered by the `lisp()` JS method that I'll soon push to #1762.
  - Resource loading is not going to be there due to WebKit quirks.
- `ironclad` is a new dependency, added to ensure it's only the browser that understands the meaning of `lisp:` URLs.
- `default-new-buffer-url` now lives in `*browser*`

# Motivation
`internal-buffer`s were quite restrictive:
- They didn't allow history (more on the current situation in Things To Discuss).
- `lisp:` URLs were only openable in those 
  - and thus not embeddable anywhere.
- Our handling of internal buffer URLs (`request-resource`) and their content `with-current-html-buffer` wasn't exactly nice -- we were setting HTML statefully, injecting request and trying to avoid WebKit at all costs.
No more! Here's a WebKit-friendly and (hopefully) safe mechanism for internal page generation. It encapsulates the creation process and generates all the necessary `lisp:` URLs for us. It's `define-internal-page-command`!

# How does that work?
- The browser generates a new internal page based on the definition.
  - It creates a buffer.
  - Sets its URL to the internal HTML-producing version of the command (`%help` for `help`).
  - And loads the URL.
- The URL looks along the lines of `lisp:2C2046FB8F9EEEBADF1BCCB0F7B64CF92D40602A4BAC17FB879290240251205B39` because it's:
  - Initially a Lisp expression.
  - Encrypted to an integer via AES-128 and printed to hexadecimal string.
  - And *registered* by Nyxt.
- When the loading happens, our callback
  - Looks at whether it's an existing URL.
  - Decrypts the URL with the keys it has.
  - Evaluates the expression in it.
  - And deletes all the URLs from the page it comes from, unless this page is still open or it's a default URL.
As the consequence of that, most URLs are only used 0-1 times, and only some (usually the `default-new-buffer-url` of \*browser\*) are used more. This leaves no chance for URL collection and encryption key guessing. Even if one guesses the encryption key, we still have our internal hash-table with valid URLs that prevents any other URLs from being evaluated.

# Things to Discuss
While this patch is complete, there's one thing that I'd like to see added to it: internal pages' history. It's still not there, because
1. URLs are encrypted and I'm not sure how to store those to keep them both safe and session-independent.
2. Should we actually store those?

How does it look to you?


jmercouris:
1. The information could be stored in a serialized map that goes together with the session. By placing it in a directory within the user's control/with limited permissions, we can keep it safe. That said, a thought occurred to me: If the encrypted command -> command is registered by Nyxt anyway, perhaps we could use a random number generator?

Consider this:

```
(defparameter pad->command (make-hash-table))

(defun encrypt-url-form (form)
  (let ((random-integer (random 99999999999999999999999999999999)))
     (setf (gethash random-integer) form)
     (format "lisp://~a" random-integer)))

```

After a command is invoked from a particular Lisp URL, it could then be discarded. This of course does cause /some/ problems. For example, if we have a button to perform an action in an internal buffer, the user presses it once, it wouldn't be pressable again unless a new URL is 'encrypted' on button press.

```
(defun evaluate-lisp-url (path)
  (let ((form (gethash path))
    (setf (gethash path) nil)
    (eval form)))
```

2. Indeed, should we? I am trying to think of any way this knowledge could be exploited. That is why I suggest one time use Lisp URLs. I don't believe a malicious actor could chain them in such a way that they could gain control over a user's machine. We could control everything that is executable by virtue of 'encrypting' it in our hash table, and avoiding it being executable twice.

All of this said, I do have a problem to consider:

We have a button on a 3rd party page titled "Bookmark". This button will run some Lisp URL. This Lisp URL will be encoded as you've defined it above. What is to stop the website from automatically 'clicking' any Lisp buttons and invoking arbitrary commands on the user's machine? How do we know that the user initiated an action, and NOT the website? What is to stop the website from automatically bookmarking itself, or refreshing the page, or performing any action we embed as a Lisp URL on the page?

source/buffer.lisp
```diff
@@ -478,21 +468,21 @@ Must be one of `:always' (accept all cookies), `:never' (reject all cookies),
 
 (define-user-class editor-buffer)
 
+;; TODO: Remove now that we don't have internal buffers?
```
jmercouris, line 471: Right, if anything we should be able to make an instance of a buffer without initializing the actual WebKit view, or lazily initializing it ONLY when it is accessed.

I can imagine:

```
(defclass buffer ()
 ((view))))

(defmethod view ((buffer buffer))
  (if (slot-boundp buffer 'view)
       (slot-value buffer 'view)
       (set (slot-value buffer 'view) (function-to-make-web-view)))

```

source/renderers/gtk.lisp
```diff
@@ -633,6 +633,38 @@ See `gtk-browser's `modifier-translator' slot."
...
               context extensions-path)))))
+      (webkit:webkit-web-context-register-uri-scheme-callback
+       context "lisp"
+       (lambda (request)
+         (with-protect ("Error while processing the lisp URL: ~a" :condition)
+           (sera:and-let* ((path (webkit:webkit-uri-scheme-request-get-path request))
+                           (code (quri:url-decode (decrypt path) :lenient t))
+                           (parent-url (gethash path *lisp-urls*)))
+             ;; FIXME: This asks for a weak hash table.
+             ;; Is that worth it, though?
```
jmercouris, line 644: What would the advantage be?

source/urls.lisp
```diff
@@ -188,15 +188,43 @@ Authority is compared case-insensitively (RFC 3986)."
...
+(defvar *lisp-urls* (make-hash-table :test 'equal)
+  "A map from the encrypted Lisp forms (string) to the URL it was created on.
+Once the parent URLs is closed/navigated away from, we clear all the related lisp URLs.")
+
+(defvar %aes (ironclad:make-cipher
+              :aes
+              :mode :ecb
+              :key (ironclad:integer-to-octets (ironclad:generate-safe-prime 256))))
+
+(defun encrypt (plaintext)
```
jmercouris, line 201: The problem is encrypting "tomato" several times in a row will give you the exact same string.

jmercouris: I believe this enables an attacker to invoke remebered Lisp URLs.

source/urls.lisp
```diff
@@ -188,15 +188,43 @@ Authority is compared case-insensitively (RFC 3986)."
...
+    (ironclad:decrypt-in-place %aes msg)
+    (quri:url-decode (coerce (mapcar #'code-char (coerce msg 'list)) 'string))))
+
 (export-always 'lisp-url)
 (-> lisp-url (t &rest t) string)
-(defun lisp-url (lisp-form &rest more-lisp-forms)
-  "Generate a lisp:// URL from the given Lisp forms. This is useful for encoding
-functionality into internal-buffers."
+;; REVIEW: We mainly use only one form. Maybe get rid of the &rest here?
+;; This would also allow defining allowed keywords to transform to params.
```
jmercouris, line 214: I'm not sure I understand, could you please provide an example?

source/renderer-script.lisp
```diff
@@ -94,30 +94,76 @@ If `setf'-d to a list of two values -- set Y to `first' and X to `second' elemen
...
+                          (list rest))))))
+    (multiple-value-bind (body declarations documentation)
+        (alex:parse-body body :documentation t)
+      `(progn
+         (defun ,internal-name (,@arglist)
+           ,documentation
+           ,@declarations
+           ;; We need to ignore those to avoid warnings, as the same arglist
+           ;; is used in both internal function and a command.
+           ;; TODO: Better way to handle those?
```
jmercouris, line 149: as silly as it is, I think this is the simplest way, for now

source/buffer.lisp
jmercouris: I would say that belongs in another changeset, that would be quite a significant change. One I have wanted to make for a while... :-)

jmercouris:
Hi Artyom, just gave some short preliminary feedback. Thanks for your excellent work as always :-)

I'm excited to get this merged in!

source/buffer.lisp
aartaka: Yes, that is right. I'll remove all the buffer hierarchy-related things from this changeset in favor of #1805.

source/renderers/gtk.lisp
aartaka: Unused URLs get garbage-collected automatically, instead of deletion in the custom URL handler. This approach has some problems too, though.

source/urls.lisp
aartaka: The key changes in between sessions.

source/urls.lisp
aartaka: I mean, maybe we can allow unencrypted query parameters for encrypted URLs, like having the variable name to be a query parameter for `describe-variable` URL, e.g. `lisp:AD2473B36CFA762394?variable=%buffer`.

On a second thought, we don't need it. We usually just embed the lisp-calculated URLs into the page, so we don't need adding anything to them afterwards.

source/renderer-script.lisp
aartaka: Sure :)

aartaka:
> We have a button on a 3rd party page titled "Bookmark". This button will run some Lisp URL. This Lisp URL will be encoded as you've defined it above. What is to stop the website from automatically 'clicking' any Lisp buttons and invoking arbitrary commands on the user's machine? How do we know that the user initiated an action, and NOT the website? What is to stop the website from automatically bookmarking itself, or refreshing the page, or performing any action we embed as a Lisp URL on the page?

1 . That is not likely to work with Lisp URLs, because one needs to redirect the user to such a URL to invoke the code. Even though this is possible, it's almost impossible to redirect the user to the URL encrypted from the code the attacker actually wants to run. All in all, yes, the scenario you provide can happen, but it's extremely unlikely and the more unlikely it is that the action invoked will cause anything but annoyance. 
2. The scenario, however, is meaningful for JS if we add a `lisp()` evaluation function to the WebExtensions library (I'm already working on it). For this, there's no other way than to restrict this function to internal pages only, unfortunately.

In summary: `lisp:` URLs are harmful and HTML-producing, unless the user wants them to be otherwise, while `lisp()` JS call is potentially omnipotent, but we restrict it to internal pages only.

aartaka:
Sorry for this force-pushing -- I've accidentally corrupted my work tree (\~_\~;)

Why: this patch no longer contains the internal-buffer removal.

Still, merge this only after `web-extensions` are merged.

Issue #1807, opened 2021-10-01
Feature request: create an option to overwrite files within the built-in editor


pdelfino:
Currently, it is not possible to overwrite files within the built-in editor using the command `editor-write-file`. It is necessary to do:

```lisp
NYXT/EDITOR-MODE> (write-file-with-editor (current-buffer)  :if-exists :overwrite) 
```

Thus, I would like to suggest a *new* feature in which if the user is editing an existing file, they are going to be prompted if they  want to save the changes overwriting the file with a question such as: "overwrite file?". Depending on the answer ("yes" or "no"), the file will be overwritten.

pdelfino:
PR #1811 solves this. Thus, I am closing the issue.

PR #1811, opened 2021-10-05
editor-mode: prompt the user about overwriting file changes.

pdelfino:
Hopefully, this solves the feature request #1807 which makes the content on the [new article](https://github.com/atlas-engineer/nyxt-site/pull/14)  (private repo) ready to be published.

I removed the `(if-exists :error))` and `:if-exists if-exists`. Is it really necessary? Should I re-introduce it?


pdelfino:
@jmercouris, I have inserted the predicate suggested by you earlier today. 

The code works on the following scenarios:

1 - the user can make changes on a file that already exists. If they wanna save the changes, the prompt buffer is going to ask for a confirmation. if the answer is "yes", the changes will be overwritten;
2 - the user can make changes on a file that already exists. If they *did it accidentaly* and do **not** wanna save them, the prompt buffer is going to ask for a confirmation. if the answer is "no", the changes will **not** be overwritten;
3 - the user can open and add content on a file that **does not** exist yet. If they wanna save the new file with the content added, they just need to execute `editor-write-file` - no confirmation will be asked.

I think it is good to go :)

source/editor-mode.lisp
```diff
@@ -80,8 +80,13 @@ get/set-content (which is necessary for operation)."
 
 (define-command editor-write-file (&key (buffer (current-buffer)) (if-exists :error))
   "Write the FILE of the BUFFER to storage."
-  (write-file-with-editor buffer :if-exists if-exists)
-  (echo "File ~a written to storage." (file buffer)))
+  (if (uiop:file-exists-p (file buffer))
+      (if-confirm ("Are you sure to overwrite file?")
```
Ambrevar, line 84: Maybe just "Overwrite?".
What about showing the file name, for instance with `"Overwrite ~s?" (file buffer)`?

source/editor-mode.lisp
```diff
@@ -80,8 +80,13 @@ get/set-content (which is necessary for operation)."
...
+        (progn (write-file-with-editor buffer :if-exists :overwrite)
+               (echo "File ~a written to storage." (file buffer)))
```
Ambrevar, line 86: Use `~s` instead to automatically surround the file with double quotes.  This is useful when the file has spaces.

source/editor-mode.lisp
```diff
@@ -80,8 +80,13 @@ get/set-content (which is necessary for operation)."
...
+        (echo "File ~a not written to storage." (file buffer)))
```
Ambrevar, line 87: "written to storage" is not very idiomatic in my experience.  Simply `"File ~s saved."`?

source/editor-mode.lisp
pdelfino: I was just trying to mimic what already exists in the source code, especially the file `buffer.lisp` :

`(if-confirm ("Are you sure to delete ~a buffer~p?" count count)`

But I can change it to your suggestion. It would be more succint, indeed.

pdelfino:
Just updated the PR with @Ambrevar's suggestions. I also changed an "unrelated" file (`buffer.lisp`) to make the format of the questions more consistent across the code base.

source/editor-mode.lisp
```diff
@@ -80,8 +80,13 @@ get/set-content (which is necessary for operation)."
...
 (define-command editor-write-file (&key (buffer (current-buffer)) (if-exists :error))
   "Write the FILE of the BUFFER to storage."
-  (write-file-with-editor buffer :if-exists if-exists)
-  (echo "File ~a written to storage." (file buffer)))
+  (if (uiop:file-exists-p (file buffer))
+      (if-confirm ("Overwrite ~s?" (file buffer))
+        (progn (write-file-with-editor buffer :if-exists :overwrite)
+               (echo "File ~s saved." (file buffer)))
+        (echo "File ~s not saved." (file buffer)))
+      (progn  (write-file-with-editor buffer :if-exists if-exists)
```
jmercouris, line 88: It could be `:if-exists :error` instead, but that is purely a stylistic choice.

Issue #1805, opened 2021-10-01
Redesign buffer class hierarchy 


Ambrevar:
(I think we already had an issue about this but I can't seem to find it.  Please let me know if you do.)

Currently we have a top `buffer` class which contains way too much, and then a bunch of child classes.  Here is the approximate hierarchy:

```
- buffer
  - internal-buffer
    - prompt-buffer
    - panel-buffer 
    - editor-buffer
    - status-buffer
  - web-buffer
    - nosave-buffer
```

This design raises a few problems:
- proxy settings are missing for the prompt buffer
- https://github.com/atlas-engineer/nyxt/pull/1642: On FreeBSD the prompt buffer would take about 30s to show up because of a bug in the password manager library!

The last point is alarming: the prompt buffer should have nothing to do with it.  The prompt buffer performance issue arose because we do way too much in `buffer` and as such it becomes such a critical part of the code.  As Nyxt grows, we risk bankruptcy because of this.

Even `make-dummy-buffer` is much too expensive.

With @aartaka 's #1804 patch the buffer hierarchy is redesigned.  He thought we could merge everything into a `buffer` megaclass.  Indeed this simplifies the design (and thus avoid issue like prompt buffer proxy settings) but it does not address the bankruptcy issue.

So instead I suggest we start from a minimal `buffer` class, then use some mixin classes to still get a simple design.

```
- raw-buffer: almost no slots, not gtk-object; useful for dummy buffers.
  - proxied-buffer: proxy settings
  - buffer: gtk-object, url, title, modes, hooks, keymap, input, zoom... (nothing network-related)
    - internal-buffer
      - prompt-buffer (proxied-buffer)
      - panel-buffer (proxied-buffer)
      - ...
    - web-buffer (proxied-buffer): password manager, downloads, default-url, resource-request-hook...
      - nosave-buffer
```

Main question: should `buffer` have a web view?  The only buffers so far that don't need a web view are the dummy buffers.  In the above, I've introduced `raw-buffer` for this purpose.  I'm unsure about the naming, but introducing `raw-buffer` and keeping `buffer` has the benefit of introducing the least changes.

Ambrevar:
@jmercouris @pdelfino What do you think?

pdelfino:
This is a delicate and critical issue. I feel uneducated about this debate to be honest. 

[...]

PR #1815, opened 2021-10-11
Nyxt URLs

### LINKED ISSUE #1091: Set URLs for internal buffers
In commit 17ecdbe34063af11a31e0cdf65f0134fe176706b
@aartaka implemented URL support for internal buffers.
While it was reverted from the pull request to make it lighter, I think it should be re-applied once the Global History Tree is merged.

### LINKED ISSUE #1094: Block and redirect internal-buffer web queries properly

### LINKED ISSUE #1109: Internal buffers should have a history

### LINKED ISSUE #1728: [Feature Request] Single Help buffer instead of new help buffer each time
Using decsribe- functions will open a new help buffer every time, wouldn't it be better to collectively control the help buffer with a single one, where it can be history navigated.

aartaka:
This is a second shot at #1804. Less ambitious, more focused.

# What's New
- `nyxt:` URL scheme. 
  - Unencrypted.
  - Splits an evaluatable form into URL pieces e.g. `(list-buffers :cluster t)` turns into `nyxt:list-buffers?cluster=t&linear-view-p=t` for better parseability and nicer look in the address area.
    - The path (`list-buffers`) can only be one of registered command names (defined with `define-internal-page-command`).
    - The query parameters (`?cluster=t&linear-view-p=t`) can only be keyword of this command.
      - This keyword requirement severely restricts us in what args can we pass to the function, but ensures that argyments are easy to map to param names and helps in making `interactive` use easier.
      - Adding to that, param values can only be lisp constants. It's not really convenient, but it prevents something like `?cluster=(progn (hack-the-browser) t)&linear-view-p=t`.
- There's no more `with-current-html-buffer` -- it's replaced by `define-internal-page-command`.
- There are changes that hint at #1805 necessity: `style` of internal buffers is moved to `buffer` and `default-new-buffer-url` is moved to `browser`.
  - I'll duplicate that in #1805: it seems that `styleable-buffer` mix-in could come in handy.
  - Default new buffer URL semantically belongs to `browser`, because it should exist before any buffer can possibly exist.
- `lisp-url` is still there, merely as a deprecated placeholder to not cause compilation warnings. To be removed once #1762 is merged and JS->Lisp calls are supported.

# Things to Discuss
- Should we actually turn neat s-expressions into URLs? It feels like a proper way to analyse and transfer those, but there can be things I haven't considered in the transition. I need your input there :)

# How to Test
Compile Nyxt with this patch set and run it. Should work fine... except for side-effectful `lisp-url`s -- those are broken, and will stay this way until #1762 is merged with a better way of interactive Lisp evaluation.

What do you think of it?

Closes #1094
Closes #1728
Closes #1109
Closes #1091


aartaka:
Important note: I wanted it to be `about:` scheme instead of `nyxt:`. However, WebKit seems to forbid `about:` rebinding -- no internal page was loading when I changed `nyxt:` to `about:` :/

Ambrevar:
*Important questions*:

1. It does not seems that you guard against arbitrary function calls, do you?
`about-url` and `parse-about-url` seem to accept any function as first argument.  Shouldn't they accept only what's in the map?
2. If I get it right, the goal of `about-url` and `parse-about-url` is to allow for some form of argument passing.
   But can they be called from web buffers then?  If so, I believe it's still too dangerous.  If not, then what's the benefit over `lisp-url`?

aartaka:
> So how is `about:` supposed to be used then? I wonder! :p

Well, there's `about:blank` and that's all there is for `about:`  scheme, it seems.

aartaka:
All done!

aartaka:
>     2. If I get it right, the goal of `about-url` and `parse-about-url` is to allow for some form of argument passing.
>        But can they be called from web buffers then?  If so, I believe it's still too dangerous.  If not, then what's the benefit over `lisp-url`?

Yes, they can be called from web buffers. And constant checking is exactly the precaution to guard us from arbitrary code execution. Other than that, none of our commands `eval` their arguments. We should be safe there.

And, as I've already said, one can define their own custom commands/schemes on top of Nyxt, but that we can do nothing about -- we're merely providing a tool there.

aartaka:
> Is there a benefit in opening internal pages inside web buffers?
>
> If we do open internal pages inside web-buffers, then there is no point in keeping internal-buffers.

Indeed, that's why I was pushing for internal-buffer removal xD

> Don't we still want to be able to evaluate `lisp:` URLs or Lisp-via-JS though? I thought that was on your roadmap. Or would we register `lisp` URLs as local and that would be enough?

Lisp-via-JS uses a totally different mechanism (see https://github.com/atlas-engineer/nyxt/pull/1762/commits/5e6710bb2c75c5b57a2c10f3ef827471972518a3 for the explanation), not tied to custom schemes at all -- it runs Lisp callbacks tied to JS functions. Custom shemes don't influence it (well, it runs exclusively on `nyxt:` scheme pages, but that's because I restricted it this way. The restriction is arbitrary there.)

> About locality: If we register the URL as local, then it means that `nyxt` links on external webpages cannot be clicked? If so, aren't we defeating the whole point of having `nyxt` URLs?

Yes, that is true. What we effectively end up with is something like Firefox's `about:` URL. I wish it was a bit more liberal, but security concerns (thank you for bringing those up!) can only be mitigated if we restrict injecting `nyxt:` URLs into webpages.

aartaka:
Given that this patch changes the startup sequence, it breaks global history test. More so: now that even internal buffers have history, we cannot start the browser without history mutation. I'm not that sure about how to proceed there. Any suggestion?

Ambrevar:
It's a long standing issue, really: restoring URLs should not trigger
history save.
Off the top of my head, it's an easy (and important fix).  If you can't
figure it out, I'll give it a shot next week.

Ambrevar:
On second thought, there may be another issue: starting Nyxt for the first
time would persist a history with the "*Help*" buffer.  Should we do
that?
Maybe it's OK.  If not, then maybe we skip saving to history for some
URLs like nyxt://(help).

Ambrevar:
Didn't you sjut say that it's a problem if we cannot start without
history mutation?

aartaka:
TODO before merging:
- Fix startup function to comply with global history test, or
- rewrite the global history test somehow.

Ambrevar:
Wait, you said

> now that even internal buffers have history, we cannot start the browser without history mutation.

then

> There should be a better way to ensure history integrity than ignoring URLs.

Why do you think history integrety could be compromised?

aartaka:
> * New JS->Lisp API which would only work in `nyxt` URLs.

A small change of plans: I am disillusioned about JS->Lisp API now, as this binds us to a specific renderer and is quite questionable in terms of security. I'm pretty sure that new `webkit_web_view_set_cors_allowlist` from WebKit 2.34.1 will allow us making requests to `lisp:` URLs thereby making them as usable from JS as JS->Lisp APIs, while still being mostly renderer-agnostic (CORS allowlisting are more widespread across renderers, and [QtWebEngine has it](https://doc.qt.io/qt-5/qwebengineurlscheme.html#Flag-enum)). The benefit of JS (Fetch API) requests is that we don't need to load `lisp:` URLs into the main web-view anymore.

I'll try to test `webkit_web_view_set_cors_allowlist` ASAP. If it works as I described (it should), I'll modify this patch and we can merge it even before #1762.

aartaka:
What I wrote in the previous message, as a table
|            |old lisp: URLs           |requestable lisp: URLs                                   |JS API                                                     |
|------------|-------------------------|---------------------------------------------------------|-----------------------------------------------------------|
|Requirements|Custom URLs              |Custom URLs + CORS allowlisting                          |WebKit: A separate .so library modifying JS environments   |
|Usage       |Lisp-only                |Lisp (buffer-load) +or JS (fetch)                         |JS-only                                                    |

aartaka:
# What's new:
- `lisp:` URLs refurbished and requestable from JavaScript.
  - `nyxt/ps:send-lisp-url` as a Parenscript helper for our buttons and other actionables.
  - Requests are processed in WebKitGTK callback and return empty string for unauthorized sources.
  - Authorized sources are: 
    - Buffer with `nyxt:` URL open.
    - Status-buffer.
    - Panel-buffer.
  - All our action-provoking buttons now are actually buttons. The code for those is a new pattern:
  ``` lisp
  (:button :class "button"
           :onclick (ps:ps (nyxt/ps:send-lisp-url `(some-lisp-code)))
           :title "some title"
           contents)
  ```

I'll be glad if you can find some time to test this and tell me how the new URL processing works.
  
# Known breakages
- Given the class change, it breaks styling xD
  - I'll fix it ASAP.
- Panel buffers are half broken too. 
  - Will fix.
- Internal pages have no titles. It should be fixed when we add `title` HTMl tag to every internal page.
- Status buffer cannot use `nyxt:` URLs as it's `"decide-policy"` does not load anything into the buffer.
  - I plan to make an internal page command for status buffer contents and allow requests to `lisp:` URLs in the policy handler of status buffer. It will be much more consistent this way.

# A note for future interface design:
 
This patch changes quite some things, and internal pages are created in a totally different way with it, compared to the current way. Here's a short list of things to remember about:
- If you want to have a command that creates/opens an internal page, use `define-internal-page-command`. It will manage most of the buffer creation/substitution and page loading.
  - The body of this command should return the valid HTML.
- If you want to reference another internal page from the page you generate, use `nyxt-url`.
- If you want to run arbitrary code without loading anything, invoke `nyxt/ps:send-lisp-url` in JS environment.
  - See snippet above for the most frequent use-case: binding Lisp action to JS event.
  
The question of buffer hierarchy persists, though. Internal buffers are not necessary anymore, and their restrictive role is not managed by `nyxt:` URLs. Shall we redesign buffer hierarchy in this patch?

aartaka:
Ah, yes. History is still broken, as CI hints.

Ambrevar:
I have so many questions! :)

- Last commit (https://github.com/atlas-engineer/nyxt/pull/1815/commits/ebd54ac9ed9087dbe487021b3e0b0ec242e1695f) you reverted the calls to CORS allowlists, why?   How can it work without it?  If we don't use them, then what does this patch do now?

- I'm now confused about the difference between `lisp:` and `nyxt:` URLs.  Can you summarize it and explain how it changed over time?

- Why move `default-new-buffer-url`? You might have mentioned it but I forgot :p

Ambrevar:
> The question of buffer hierarchy persists, though. Internal buffers are not necessary anymore, and their restrictive role is not managed by nyxt: URLs. Shall we redesign buffer hierarchy in this patch?

Did you mean "is _now_ managed"?

As I understand, internal buffers are now `nyxt:` URL buffers, is that right?  If so, it seems reasonable to me to delete the `internal-buffer` class.

I remember discussing it a few weeks back and there were more points of caution, but I forgot which ones.  Do you remember?  What would be the drawback of removing internal buffers?  Any security concern?

source/urls.lisp
```diff
@@ -188,16 +188,86 @@ Authority is compared case-insensitively (RFC 3986)."
...
        (apply #'str:concat "lisp://"
               (mapcar (alex:compose #'quri:url-encode #'write-to-string)
                       (cons lisp-form more-lisp-forms)))))
 
+(export-always 'parse-nyxt-url)
+(-> parse-nyxt-url ((or string quri:uri)) (values symbol list &optional))
+(defun parse-nyxt-url (url)
+  "Return (1) the name of the function and (2) the arguments to it, parsed from about: URL.
+
+Errors if some of the params are not constants. Thanks to this,
```
Ambrevar, line 247: "Error out"

source/urls.lisp
```diff
@@ -188,16 +188,86 @@ Authority is compared case-insensitively (RFC 3986)."
...
+`parse-nyxt-url' can be repeatedly called on the same about: URL, with the
```
Ambrevar, line 248: Did you mean "nyxt: URL"?

source/annotate.lisp
```diff
@@ -112,20 +112,21 @@
...
 
-(defun render-annotations (annotations)
-  (with-current-html-buffer (buffer "*Annotations*" 'base-mode)
-    (spinneret:with-html-string
-      (:style (style buffer))
-      (:h1 "Annotations")
-      (loop for annotation in annotations
-            collect (:div (:raw (render annotation))
-                          (:hr))))))
+(define-internal-page-command render-annotations (&key annotations)
```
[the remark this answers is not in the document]
Ambrevar: OK, got it.

aartaka:
> * Last commit ([ebd54ac](https://github.com/atlas-engineer/nyxt/commit/ebd54ac9ed9087dbe487021b3e0b0ec242e1695f)) you reverted the calls to CORS allowlists, why?   How can it work without it?  If we don't use them, then what does this patch do now?

It appears that `webkit:webkit-security-manager-register-uri-scheme-as-cors-enabled` allows doing CORS request to the scheme even without allowlist. I was surprised about it, but it seems to work just fine.

> * I'm now confused about the difference between `lisp:` and `nyxt:` URLs.  Can you summarize it and explain how it changed over time?

Sure! 
- `nyxt:` URLs are essentially a reincarnation of internal buffers, as you said in your latter message. They are responsible for the layout of the Nyxt internal pages. 
- `lisp:` URLs are only for code execution on pages with `nyxt:` URLs. One can also try to load those, but that is not an intended use.

So, it didn't change much between the situation with internal buffers and `lisp:` URLs from before. Two most significant changes are:
- Internal pages have history, can be referenced and in general are first-class citizens of WebKitGTK -- all due to custom `nyxt:` URL.
- `lisp:` URLs are now requestable and can be invoked from JavaScript (but, actually, any other code that happens to be on a page with `nyxt:` URL). This is due to the same custom URL mechanism being used, with `webkit:webkit-security-manager-register-uri-scheme-as-cors-enabled` on top to allow Cross-Origin requests.

> * Why move `default-new-buffer-url`? You might have mentioned it but I forgot :p

It was in this patch from the very start :) The way I seemed to frame it in the initial PR message is:

> Default new buffer URL semantically belongs to browser, because it should exist before any buffer can possibly exist.

And it does make sense :)

aartaka:
> Did you mean "is _now_ managed"?

Yes, sorry for this typo (-_-;)

> As I understand, internal buffers are now `nyxt:` URL buffers, is that right? If so, it seems reasonable to me to delete the `internal-buffer` class.

Yes, that's a fair assesment

> I remember discussing it a few weeks back and there were more points of caution, but I forgot which ones. Do you remember? What would be the drawback of removing internal buffers? Any security concern?

I cannot remember, but the biggest thing definitely was security. Now that the mechanism is safe, we can safely remove `internal-buffer`s, I'm sure :)

Ambrevar:
OK, looks good to me to merge after the CI passes.

I suggest we remove internal buffers in a subsequent pull request.

aartaka:
> OK, looks good to me to merge after the CI passes.

I've tried to fix the startup sequence, but, apparently, the way we imitate Nyxt startup in the "Global history" test is not compatible with this patch, as it leaves out too much details that got changed. The current test relies on the absence of history on startup, at least.

I'm not sure how to proceed there, as I'm afraid of rewriting this monumental test xD

Ambrevar:
I'll have a look as soon as possible.

Ambrevar:
Wait, the CI says this passes.  Are the tests OK in the end?

aartaka:
Let me re-run the checks. It was last ran three weeks ago, so I'm not sure anymore. But it will be good news if it passes :D

aartaka:
CI seems to pass just fine 0_o

I'll merge tomorrow then.

Ambrevar:
Double-check locally to be sure.  Any idea what was wrong?  Do we still
have the old logs?  (Maybe in the old GitHub actions.)

aartaka:
Tests seem fixed---they run fine both on my machine and in the CI 0_o

aartaka:
All the bugs are fixed, tests pass, all the buffers have history and consistent styles. I believe it's fine to merge.

Ambrevar:
Woohooo, it's on master now!  Congrats for the behemoth work, Artyom!

Issue #1786, opened 2021-09-13
Feature request: symmetric to `switch-buffer-domain` create a variaton of `set-input-data-from-saved` 


pdelfino:
Nyxt offers multiple ways of swithing buffers. One of them `switch-buffer-domain` which limit the options of change to the domain on the current buffer.

Last week, @jmercouris suggested creating a new command `set-input-data-from-saved-domain`. It would be exactly like `set-input-data-from-saved`, except that it would limit the options to the ones under the same domain as the current buffer.

pdelfino:
I will send a patch about this soon.

pdelfino:
Closed with [this](https://github.com/atlas-engineer/nyxt/commit/982522f1e3289f37a08b46faceab1576d4ac05c0#diff-3ad9b39e0ff392243db77d11210f2ff3339b79df854ca63fbc29f683da471bd9) commit.

PR #1839, opened 2021-10-30
record-input-field: create new command set-input-data-from-saved-domain.

pdelfino:
Hopefully, this solves issue #1786.

Gif with three tests:
test-domain-set-input-data

### DIFF
--- a/source/record-input-field.lisp
+++ b/source/record-input-field.lisp
@@ -101,6 +101,13 @@
   ((prompter:name "Inputs")
    (prompter:constructor (get-data (inputs-path (current-buffer))))))
 
+(define-class filtered-domain-input-data-source (prompter:source)
+  ((prompter:name "Inputs")
+   (prompter:constructor (remove-if-not #'(lambda (input-entry)
+                                            (equal (quri:uri-domain (quri:uri (url input-entry)))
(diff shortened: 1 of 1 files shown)


source/record-input-field.lisp
```diff
@@ -101,6 +101,13 @@
   ((prompter:name "Inputs")
    (prompter:constructor (get-data (inputs-path (current-buffer))))))
 
+(define-class filtered-domain-input-data-source (prompter:source)
+  ((prompter:name "Inputs")
+   (prompter:constructor (remove-if-not #'(lambda (input-entry)
```
Ambrevar, line 106: Maybe just `input` instead of `input-entry`?

source/record-input-field.lisp
```diff
@@ -101,6 +101,13 @@
...
+                                            (equal (quri:uri-domain (quri:uri (url input-entry)))
```
Ambrevar, line 107: This asymmetry highlights that something is wrong: the `url` function should always return a `quri:uri`, otherwise we have inconsistent types.

Easy fix: in `save-input-data`, don't call `render-url`.

Ambrevar:
Looks good to me, but could you fix the type of the `url`?

jmercouris:
This has been merged.

Issue #1843, opened 2021-11-02
Redirects are mistakenly not stored into history


Ambrevar:
Recipe:

- `set-url` to `wiki tomato`
- Run `history-all-query` and notice that the current URL is missing.

@aartaka This is due to https://github.com/atlas-engineer/nyxt/commit/99c8f81a5164a11709ea7eb820b841f9c946f29d.  Indeed, `:webkit-load-redirected` may be sent for the final URL (I don't know why, either a server thing or a WebKit bug).

That said, I don't understand your patch.  We only store to history on `load-finished`, that is, with the final URL, so why would we need to blacklist redirected URLs?

aartaka:
Oh wait! I've got it wrong! The `:webkit-load-redirected` URL is the final result of redirection, while the URL to not store in history is the one that came before the redirection. It's more complex than I thought. I suggest we revert 99c8f81 and approach redirections in a smarter way, altering history after the fact of redirection.

aartaka:
> That said, I don't understand your patch. We only store to history on `load-finished`, that is, with the final URL, so why would we need to blacklist redirected URLs?

Do we? We also do it in `nyxt:on-signal-notify-uri`/`notify::uri`, where the filtering is most needed, as notified URLs often contain redirection URLs and other junk, as contrasted to those in `load-finished` signal.

aartaka:
So, what I suggest is:
- When we get a `:webkit-load-redirected` load status, we look at most recent URL in the buffer history.
- We delete this most recent URL.
- Westore to history the URL that came along with `:webkit-load-redirected` signal, as this is most likely the final result of redirection.

While this breaks the immutability of `htree`, we need to do the change after we actually added a redirection URL to history. The information on it being a redirection URL is only available later, when `:webkit-load-redirected` happens, thus the delay.

Ambrevar:
> Do we? We also do it in nyxt:on-signal-notify-uri/notify::uri, where the filtering is most needed, as notified URLs often contain redirection URLs and other junk, as contrasted to those in load-finished signal.

Good point!

PR #1844, opened 2021-11-02
Fix redirections skipping

aartaka:
This is a fix to my oversight: `:webkit-load-redirected` is actually emitted when the redirection is already resolved. Thus the URL that comes with `:webkit-load-redirected` shouldn't be ignored -- quite the opposite, it should be stored to history, overwriting the URL that it was redirected from (basically the parent URL). That's what this patch does.

Does it make sense?

Closes: #1843.

### DIFF
--- a/source/web-mode.lisp
+++ b/source/web-mode.lisp
@@ -655,7 +655,18 @@ ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
   nil)
 
 (defmethod nyxt:on-signal-load-redirected ((mode web-mode) url)
-  (push (render-url url) (history-blocklist mode)))
+  (with-history-access (history (buffer mode))
+    (sera:and-let* ((owner (htree:owner history (id (buffer mode))))
+                    (node (htree:current owner))
+                    (data (htree:data node)))
+      ;; We need to rewrite the previous URL with the data of the URL we get
+      ;; after redirection, thus this entry rewriting.
+      ;;
+      ;; A good alternative would be to delete the previous node and add the new
+      ;; one to history instead, but `htree' is immutable in this regard.
+      (setf (slot-value data 'nyxt:url) url
+            (title data) (title (buffer mode))
+            (nyxt::last-access data) (local-time:now)))))
 
 (defmethod nyxt:on-signal-load-finished ((mode web-mode) url)
   (add-url-to-history url (buffer mode) mode)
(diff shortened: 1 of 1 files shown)


Ambrevar:
I believe this should work, but rewriting the history is of course a bit dodgy.

What about this instead: when following a URL that's going to be redirected `WEBKIT_LOAD_STARTED` is emitted in `load-changed`.  Then I suggest to add a condition in `notify-uri` to test if `load-status` is `started` and if so, _do not store the history entry_.  This way, we would never store the wrong URL.

Do you think that would work?

aartaka:
Yes, it does make sense. See 65174d736927d12711128638c6cb84fa7dc3ec10 for what I've gone with.

To summarize, there are only two places and cases we store history in:
- In `notify::uri`, when the `load-status` is `:finished` (e.g., GitHub "Issues" and "Pull requests" buttons).
- In `on-signal-load-changed`, when the loading is finished.

I decided to not modify history on redirection, given that any chain of redirections ends with a `WEBKIT_LOAD_FINISHED`. @Ambrevar, you've said it's not the case and suggested to try `wiki tomato` search. It did end with `WEBKIT_LOAD_FINISHED` for me, though (tracked with SLY stickers in `on-load-changed`). Can you please try again, with and without this patch?

Ambrevar:
Also don't call `ld` directly, it's too low level, prefer `gcc`.  Better: use a defvar and set it to something like `(or (uiop:getenv "CC") "gcc")`.

Ambrevar:
Excellent, works perfectly for me so far!

Ambrevar:
> @Ambrevar, you've said it's not the case and suggested to try wiki tomato search. It did end with WEBKIT_LOAD_FINISHED for me,

I don't think I said this, misunderstanding?

In #1843 I meant that `load-redirected` is sent twice, but I didn't say that `load-finished` was not sent, in fact it is.

Ambrevar:
Feel free to merge if there is no other issue.

aartaka:
> > @Ambrevar, you've said it's not the case and suggested to try wiki tomato search. It did end with WEBKIT_LOAD_FINISHED for me,
> 
> I don't think I said this, misunderstanding?
> 
> In #1843 I meant that `load-redirected` is sent twice, but I didn't say that `load-finished` was not sent, in fact it is.

I reread your comment. I did misunderstand it, yes :) Anyway, I'm merging this, as it still seems to fix the problem.

PR #1850, opened 2021-11-04
Mode move

jmercouris:
this is the first step to decoupling the modes with the core completely!


jmercouris:
I forgot, this was already an issue we had talked about and agreed upon. I will just merge this in. My changes to the modes themselves will be in a separate pull request!

PR #1852, opened 2021-11-05
source/mode: remove mode suffix from modes

jmercouris:
it is duplicative to both have them in a mode folder, AND have the
mode suffix


jmercouris:
Additionally, rename `application-mode` to `passthrough-mode` that is what most people seem to know it as.

Ambrevar:
I'm confused: you've already pushed https://github.com/atlas-engineer/nyxt/commit/779749f3f23fafdfc34aae7d73fa639315161a5d.
Was it a mistake?

We should avoid too many renames, it makes it hard to follow through the Git history later on.

Otherwise I agree with this change :)

PR #1856, opened 2021-11-06
An option not to resolve DNS names when opening an URL

shamazmazum:
Here is a commit which makes use of SKIP-DOMAIN-VALIDATION in VALID-URL-P. When domain validation is skipped you do not have to wait a bit after you typed your web address in a prompt buffer. I thought this was done before, but apparently it wasn't.

Here is the rules how URL queries are distinguished from search queries:

* If you type a query with a scheme and this looks like an URL when it is an URL (e.g. `http://localhost` is considered to be URL)
* If you type something like `localhost` when it is considered to be a search query even if it is a host name.
* If you type something with dot which appears to be an URL when it is considered to be URL (e.g. `google.com`)
* Everything else is a search query.

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -73,10 +73,10 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
   (iolib/sockets:lookup-hostname name))
 
 (export-always 'valid-url-p)
-(defun valid-url-p (url &key skip-domain-validation)
+(defun valid-url-p (url &key (check-dns-p t))
   "Return non-nil when URL is a valid URL.
-With SKIP-DOMAIN-VALIDATION, the domain name existence is not verified.
-Domain name validation may take significant time since it looks up the DNS."
+The domain name existence is verified only if CHECK-DNS-P is T. Domain
+name validation may take significant time since it looks up the DNS."
   ;; List of URI schemes: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
   ;; Last updated 2020-08-26.
   (let* ((nyxt-schemes '("lisp" "javascript"))
@@ -111,7 +111,7 @@ Domain name validation may take significant time since it looks up the DNS."
                 ;; A valid URL may have an empty domain, e.g. http://192.168.1.1.
                 (quri:uri-host url)
                 (or
-                 skip-domain-validation
+                 (not check-dns-p)
                  ;; Onion links or not resolved via DNS, just accept them.
                  (string= (quri:uri-tld url) "onion")
                  ;; "http://algo" has the "algo" hostname but it's probably invalid
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1198,14 +1198,16 @@ If prefixing with 'https://' results in a valid URL, set `query' to this result
 on instantiation.
 Finally, if nothing else, set the `engine' to the `default-search-engine'."))
 
-(defmethod initialize-instance :after ((query new-url-query) &key)
+(defmethod initialize-instance :after ((query new-url-query)
+                                       &key check-dns-p &allow-other-keys)
   ;; Trim whitespace, in particular to detect URL properly.
   (setf (query query) (str:trim (query query)))
   (cond
     ((engine query)
      ;; First check engine: if set, no need to change anything.
      nil)
-    ((valid-url-p (query query))
+    ((valid-url-p (query query)
+                  :check-dns-p check-dns-p)
      ;; Valid URLs should be passed forward.
      nil)
     ;; Rest is for invalid URLs:
@@ -1216,7 +1218,9 @@ Finally, if nothing else, set the `engine' to the `default-search-engine'."))
             (uiop:native-namestring
              (uiop:ensure-absolute-pathname
               (query query) *default-pathname-defaults*)))))
-    ((valid-url-p (str:concat "https://" (query query)))
+    ((and check-dns-p
+          (valid-url-p (str:concat "https://" (query query))
+                       :check-dns-p check-dns-p))
      (setf (query query)
            (str:concat "https://" (query query))))
     (t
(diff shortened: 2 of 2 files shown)


source/urls.lisp
```diff
@@ -72,11 +72,24 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
   #-sbcl
   (iolib/sockets:lookup-hostname name))
 
+(defparameter *skip-domain-validation* nil
+  "If this parameter is set to T nyxt skips checks of domain name
+existence by default.")
+(export-always '*skip-domain-validation*)
```
Ambrevar, line 78: We avoid globals in Nyxt.

Either it's buffer-local, or window-local, or if it's truly general then we store it in the `browser` class.
In this case, I'd make this option buffer-local.

source/urls.lisp
```diff
@@ -72,11 +72,24 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
   #-sbcl
   (iolib/sockets:lookup-hostname name))
 
+(defparameter *skip-domain-validation* nil
+  "If this parameter is set to T nyxt skips checks of domain name
```
Ambrevar, line 76: "When non-nil, skip domain name checks when resolving URL queries."

source/urls.lisp
```diff
@@ -72,11 +72,24 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+existence by default.")
+(export-always '*skip-domain-validation*)
+
+(define-command toggle-domain-validation ()
+  "Toggle checks of domain name existence when opening an URL."
```
Ambrevar, line 81: "a URL"

source/urls.lisp
```diff
@@ -72,11 +72,24 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+  (echo "Domain validation ~:[enabled~;disabled~]"
+        (setq *skip-domain-validation*
```
Ambrevar, line 83: `setf` is more general, `setq` is more of an Emacs-ism.

Ambrevar:
Overall, it's OK but I wonder if it's the right solution.

What are we trying to solve here?  The `set-url` prompt buffer latency on input?

If so, I suggest the following instead: proceed in 2 passes,
- pass 1 does not check the domain validity
- pass 2 does.

This way, pass 1 should display instantly and pass 2 updates the content of the prompt buffer with more accurate results (more often than not, it should not change anything).

Thoughts?

shamazmazum:
> What are we trying to solve here? The set-url prompt buffer latency on input?

Yes. This is really annoying if you type fast :)

> This way, pass 1 should display instantly and pass 2 updates the content of the prompt buffer with more accurate results

This sound like the best solution, but I don't know how to implement it ;) As I understand, a new instance of `new-url-query` is created by `input->queries` as you type (seems like one instance for one typed character). `initialize-instance :after` for that class sets a URL or a search engine for the instance.

As I understand you want `initialize-instance` to do only basic checks (skipping DNS check) and then modify content of the prompter source (??? I don't know how to express it correctly) when asynchronous DNS check is finished. Any ideas and/or examples how to implement this asynchronous check?

Ambrevar:
I'm away at the moment, so I can't answer you immediately.
For now, all I can tell is that the prompter library supports asynchronous suggestion computation, so this could help.
For instance, we could use the filter-preprocessor to create suggestions without DNS validation, and the filter-postprocessor to update these suggestions with DNS validation this time.

I'm guessing that it should work but I'm unsure whether the display notifications are set at the right spot; if not, the prompter library will need a (trivial) update.

Ambrevar:
Another easier workaround would be to use https://github.com/lu4nx/cl-tld to validate the domains from a public database.

shamazmazum:
OK, I used filter-postprocessor. Things are now much faster. Minor style changes:

* `(declare (ignorable ...)) ` was replaced by `(declare (ignore ...))` in lambdas.
*  `skip-domain-validation` was replaced by shorter `skip-dns-check`.

What do you think?

Ambrevar:
`skip-dns-check` is indeed better.  What about the even shorted `check-dns-p`? (Which reverses the logic of course.)

About the `(declare (ignore ...))`, these are ignored indeed, but do you have a specific rationale in mind?

Looks great otherwise!  I'll test and merge as soon as possible!

Ambrevar:
I've tested cl-tld: it's indeed a fast way to check that the domain is a valid one, it just won't find hostnames (like `localhost`).  It needs an update though.

Conclusion: with an updated cl-tld and by adding support for some well known hostnames (like `localhost`) + for IP addresses, `valid-url-p` can do most of its job without any expensive system call.

shamazmazum:
> `skip-dns-check` is indeed better. What about the even shorted `check-dns-p`?

Done.

> About the `(declare (ignore ...))`... but do you have a specific rationale in mind?

This way you are sure they are ignored just by looking at declaration:)

> with an updated cl-tld and by adding support for some well known hostnames (like localhost)

What is about less known hostnames? `http://foobar/page.html` is also a valid url.
Check my patch, I do not sense any delay when I type an address.

Ambrevar:
> What is about less known hostnames? http://foobar/page.html is also a valid url.

`foobar` is a local hostname, isn't it?
Then it will be covered by the lookup-hostname function in the postprocess.

My point with cl-tld is that we can validate well-known domains in the first pass already. thus giving rather accurate results instantly, with perfect results coming later from the postprocess.

shamazmazum:
> `foobar` is a local hostname, isn't it?
Then it will be covered by the lookup-hostname function in the postprocess.

I think I am confused. Do you or do you not want to replace DNS lookup with `cl-tld` completely? If not, is there a reason to do such sophisticated checks before checking the host with DNS?

Ambrevar:
This patch is completely fine as it is, I am merely discussing future
improvements with cl-tld.

In the future, I'd like to perform a "quick and imprecise domain check"
with cl-tld followed by a full DNS check (as you do in this patch).

Ambrevar:
Don't worry about anything, I'll test this one last time and merge!

shamazmazum:
Thanks!

Ambrevar:
Merged with https://github.com/atlas-engineer/nyxt/commit/485f04228f53311eaf46f5644c541121ec6347a9.
See also https://github.com/atlas-engineer/nyxt/commit/1c8a86c76bcaa1e4b2d41717dba31938ce0235c1 for an additional optimization that follows the same logic, but with search engine completion this time.

Thanks a lot @shamazmazum for this idea, it's a tremendous improvement to the overall user experience!

Issue #1840, opened 2021-10-30
Nyxt window crashes after loading default history and changing to a web buffer


fade:
**Describe the bug**
When starting nyxt built from either 2.2.2 tag, or the HEAD of the master branch, it crashes after selecting the default history file and switching to a web buffer. 
**Precise recipe to reproduce the issue**
Start Nyxt, select default history, switch to web buffer. <crash>
This still happens when nyxt is started without any configuration.
prior to switching to the web buffer, the minibuffer shows: 
Warning: Errorr on GTK thread: The value :INVALID-CODE-OBJECT-AT-PC is not of type FUNCTION
**Information**
- OS name+version: XeroLinux (Arch, rolling) 
- Linux kernel: 5.14.14-arch1-1 #1 SMP PREEMPT Wed, 20 Oct 2021 21:35:18 +0000 x86_64
- Graphics card and driver: NVIDIA Quadro K1000M/Intel 3rd Gen Core GC running noveau/intel on XORG/X11
- Desktop environment / Window manager name+version: StumpWM 20.11-91-gf867f9e Compiled On Fri Oct 29 2021 22:26:29
- How you installed Nyxt (Guix pack, package manager, build from source): Built Nyxt from source, both 2.2.2 tag and HEAD of master, although I see that as of right now, these things are the same commit.
- Information from `show-system-information`:

Nyxt version: 2.1.1-713-g26a49cb0
Renderer version: GI-GTK
Operating system kernel: Linux 5.14.14-arch1-1
Lisp implementation: SBCL 2.1.9 (Dynamic space size: 12884901888)
Features: (:WEBKIT2 :WEBKIT2-2.34 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
[...]

PR #1858, opened 2021-11-07
Add WebKit-reliant editing commands.

### LINKED ISSUE #810: Pasting text does not mark the tag as "edited"
Maybe pages ungrey buttons only when some text input is non-empty.
Example: This GitHub new issue page.  If the title is empty, "Submit new issue" remains greyed out.

The problem:
- Copy some text to the clipboard.
- Open a new issue on GitHub.
- Paste the text into the title line.

Witness that "Submit new issue" remains greyed out.  Inserting any character into the title field ungreys the button.

### LINKED ISSUE #1504: Pasting a URL is sent as a query string to duckduckgo
Running nyxt on ubuntu 20.04 LTS. Whenever I copy a URL (for e.g. https://www.braveclojure.com/core-functions-in-depth/) into the system clipboard to paste into nyxt (by using `set-url` and then paste), the url is sent as a query string to duckduckgo.com. How do I directly navigate to the URL without being routed through ddg?

### LINKED ISSUE #1565: Allow external editor in any HTML editable element
See https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/isContentEditable.

Looks like we need to use JavaScript to check whether an element is editable or not.

### LINKED ISSUE #1823: Utilise GActions for copying/pasting
Gtk provides us with multiplatform and mostly reliable sets of action (e.g. in the context menus). We can probably utilize those. I haven't yet gotten to understand how to construct those, but it should work somehow. Here's what I tried:

``` lisp
(gtk:gtk-action-activate
          (webkit:webkit-context-menu-item-get-action
           (webkit:webkit-context-menu-item-new-from-stock-action
            (cffi:make-pointer (cffi:foreign-enum-value 'webkit::webkit-context-menu-action 
                                                        :webkit-context-menu-action-copy)))))
```

This didn't work, but I'm pretty sure it's close to what we need.

And here're the pieces of reference I used:
[https://webkitgtk.org/reference/webkit2gtk/stable/WebKitContextMenuItem.html#webkit-context-menu-item-get-gaction](WebKitContextMenuItem)
[https://docs.gtk.org/gtk3/class.Action.html](GtkAction)
GAction: [reference](https://developer-old.gnome.org/gio/stable/GAction.html), [tutorial](https://wiki.gnome.org/HowDoI/GAction)

Anyone with better Gtk knowledge?

aartaka:
This makes our editing commands (`copy`, `paste` etc.) to rely on what WebKit provides, as this is usually a bit more robust than our JS snippets.

Closes #1823
Closes #810 
Closes #1565 
Closes #1504


Ambrevar:
> @jmercouris, these bindings were in cl-webkit since forever. Were they used in Nyxt at some point? Why using JS instead?

I can give you an answer: initially we did put a lot of effort to be renderer-agnostic, so whatever could be done in pure JS would be preferred to renderer bindings to help with portability.

I think it's fine to have both: GTK-specialized version in parallel to pure JS implementation so that the QtWebEngine renderer can fall back to it, in case it's not available as bindings.

Ambrevar:
> I cannot test this patch due to WebKit 2.34.1 update

Does it crash Nyxt as reported in https://github.com/atlas-engineer/nyxt/issues/1840?

aartaka:
> > I cannot test this patch due to WebKit 2.34.1 update
> 
> Does it crash Nyxt as reported in #1840?

No, it was me not adding `cl-webkit` 3.1.0 to `build-script/nyxt.scm` and thus suffering from the broken binary not having all the `cl-webkit` API changes we did :)

aartaka:
Using this patch to write this comment, and it's extremely smooth, compared to our JS implementations! More so: pasting invokes all the necessary JS events now, and e.g. Github buttons light up when you paste, so #810 should be fixed by this patch :D

Ambrevar:
Neat!

jmercouris:
Why does the Guix version fail the CI? I presume because the inputs have a too old cl-webkit?

Ambrevar:
Indeed, @aartaka can you update the nyxt.scm?
Or simply rebase on master since you have updated it there already I believe.

aartaka:
This particular patch needed the most recent version of cl-webkit because of the mistyped export that it fixed. I've updated the submodule. It should work fine now.

aartaka:
There are now not-yet-implemented methods for the editing commands we don't yet have JS implementations for. They simply `echo-warning` that the editing command is not implemented. 

If that's enough (given that we have only one port now, it should be, and it will be of help if we are to make a new port), then we can merge.

All the key bindings (except `C-x` as Cut in CUA, as `C-x` is occupied as a command prefix already; and the Select All binding for VI, as it doesn't exist there) are there too, along with the `web-mode` commands for all the editing commands.

aartaka:
Merging then?

aartaka:
I've added JS-based cutting and select-all. It's complete and safe to merge now.

aartaka:
I couldn't hold myself from adding the `contentEditable` check and thus this patch also covers #1565 xD

source/browser.lisp
```diff
@@ -753,3 +753,51 @@ sometimes yields the wrong result."
 (define-ffi-generic ffi-set-preferred-languages (buffer value))
 (define-ffi-generic ffi-focused-p (buffer))
 (define-ffi-generic ffi-set-tracking-prevention (buffer value))
+(define-ffi-generic ffi-buffer-copy (buffer)
+  (:method ((buffer buffer))
+    ;; TODO: Abstract this to a sibling of `with-current-buffer'?
```
Ambrevar, line 758: Indeed, why don't you usee `with-current-buffer` here?

source/browser.lisp
```diff
@@ -753,3 +753,51 @@ sometimes yields the wrong result."
...
+(define-ffi-generic ffi-buffer-cut (buffer)
+  (:method ((buffer buffer))
+    (let ((switching-buffers-p
+            (not (eq buffer (current-buffer)))))
+      (when switching-buffers-p
+        (switch-buffer :id (id buffer)))
+      (let ((input (%cut)))
+        (when input
+          (copy-to-clipboard input)
+          (echo "Cut the text : ~s" input)))
```
Ambrevar, line 786: For consistency with the copy message, I'd write "Text cut: ".

source/browser.lisp
```diff
@@ -753,3 +753,51 @@ sometimes yields the wrong result."
...
+    (let ((switching-buffers-p
+            (not (eq buffer (current-buffer)))))
+      (when switching-buffers-p
+        (switch-buffer :id (id buffer)))
+      (%select-all)
+      (when switching-buffers-p
+        (switch-buffer-last)))))
+(define-ffi-generic ffi-buffer-undo (buffer)
+  (:method ((buffer buffer))
+    (echo-warning "Undoing the editingis not yet implemented for this port.")))
```
Ambrevar, line 800: Typo, maybe go with "Undoing edits is...".
I don't think we still use the "port" term, I'd use "renderer" instead.

source/browser.lisp
```diff
@@ -753,3 +753,51 @@ sometimes yields the wrong result."
...
+(define-ffi-generic ffi-buffer-redo (buffer)
+  (:method ((buffer buffer))
+    (echo-warning "Redoing the editingis not yet implemented for this port.")))
```
Ambrevar, line 803: Same.

source/mode/web.lisp
```diff
@@ -57,7 +57,7 @@ and to index the top of the page.")
     (define-scheme "web"
       scheme:cua
       (list
-       "C-z" 'nyxt/passthrough-mode:passthrough-mode
+       "C-M-Z" 'nyxt/passthrough-mode:passthrough-mode
```
Ambrevar, line 60: Could you update the changelog to signal that this particular keybinding has changed?

source/mode/web.lisp
```diff
@@ -77,6 +77,10 @@ and to index the top of the page.")
        "C-x C-w" 'copy-hint-url
        "C-c" 'copy
        "C-v" 'paste
+       ;; FIXME: C-x is a command prefix, thus no cutting in CUA :/
```
Ambrevar, line 80: Is it?  It only seems to be used for `copy-hint-url`, which is easy to rebind.  

While we are at it, I suggest we fix the copying commands.  `copy-url` is bound to `M-c l`, which is pretty bad.  Why `l` ?!?

Prefixing copy commands with `M-c` makes sense, so we could have `M-c u` for URLs and `M-c h` for hints.

source/mode/web.lisp
```diff
@@ -522,7 +533,7 @@ Otherwise go forward to the only child."
   ;; is present the clipboard and need not be retrieved from the GTK thread.
   ;; TODO: Do we still need to flush now that we have multiple threads?
   ;; (trivial-clipboard:text (trivial-clipboard:text))
```
Ambrevar, line 535: I think the above comment should be moved to the defgeneric, nah?

source/mode/web.lisp
```diff
@@ -559,6 +568,22 @@ Otherwise go forward to the only child."
           (progn (copy-to-clipboard current-value)
                  (echo "Placeholder copied."))))))
 
+(define-command cut (&optional (buffer (current-buffer)))
+  "Cut the selection in BUFFER."
```
Ambrevar, line 572: Maybe change the docstring to make it clear that it's about the "selected text" (see `copy` docstring).

source/renderer/gtk.lisp
```diff
@@ -1397,3 +1397,57 @@ As a second value, return the current buffer index starting from 0."
    (webkit:webkit-web-context-website-data-manager
     (webkit:webkit-web-view-web-context (gtk-object buffer)))
    t))
+
+(define-ffi-method ffi-buffer-copy ((gtk-buffer gtk-buffer))
+  (webkit:webkit-web-view-can-execute-editing-command*
```
Ambrevar, line 1402: Is there a use case for the non-starred command?  Otherwise, I suggest we do as with other wrappers like `webkit-cookie-manager-get-accept-policy-finish`, where we don't export the direct binding, and `webkit:webkit-web-view-can-execute-editing-command` is the wrapper.

source/renderer/gtk.lisp
```diff
@@ -1397,3 +1397,57 @@ As a second value, return the current buffer index starting from 0."
...
+       (webkit::webkit-web-view-execute-editing-command
+        (gtk-object gtk-buffer) webkit2:+webkit-editing-command-copy+)))
+   (lambda (e) (echo-warning "Cannot copy: ~a" e))))
+
+(define-ffi-method ffi-buffer-paste ((gtk-buffer gtk-buffer))
+  (webkit:webkit-web-view-can-execute-editing-command*
+   (gtk-object gtk-buffer) webkit2:+webkit-editing-command-paste+
+   (lambda (can-execute?)
+     (when can-execute?
+       (webkit::webkit-web-view-execute-editing-command
```
Ambrevar, line 1415: Why do double-colons?  The export is now fixed.

source/browser.lisp
aartaka: I cannot remember what made me implement it this way, but it looks insane 0_o

source/mode/web.lisp
aartaka: I will rebind `copy-hint-url` (and `open-file-with-external-editor`, as it uses `C-x` too). I'll leave everything else for #1870.

source/mode/web.lisp
aartaka: But wait, we no longer have the "Upcoming" section in out changelog. Should I create a new one or pospone it until we collect changes for 2.2.3/2.3/3.0?

aartaka:
OK, `cl-webkit` is updated with more consistent bindings, changelog I'll leave for later as it's uncertain what keys are going to change to and where to document it. Other points are addressed. Merging then?

Ambrevar:
Merging!

aartaka:
No, it's not the time to merge yet: it seems that binding `C-x` for cutting creates conflicts. Running Nyxt from this branch, I cannot use `C-x C-f` in Emacs mode, as CUA interferes with it.

Before merging, then, we need to make sure (in `libraries/keymap`?) that child keymap using `C-x` as a prefix combination has priority over the non-prefix usage from the parent keymap. @Ambrevar, any idea how to do that?

aartaka:
Update: ALL the `C-k`-prefixed combinations stopped working 0_o

aartaka:
Merging this without the `C-x` binding. We can add it later, when keymap resolves it in a better way.

jmercouris:
Thanks so much Artyom. High quality work!

Ambrevar:
> Before merging, then, we need to make sure (in `libraries/keymap`?) that child keymap using `C-x` as a prefix combination has priority over the non-prefix usage from the parent keymap.

Is it a good idea though?
I would say yes, it makes sense that the child keymap has priority over
its parent.

That said, finding a non-conflicting binding is also a good idea! :)

aartaka:
> That said, finding a non-conflicting binding is also a good idea! :)

Cutting is terribly traditional in CUA with `C-x`. I don't believe we can find any alternative for such a basic thing.

Ambrevar:
I meant the other way around, remap the `C-x`-prefixed keys ;)

aartaka:
> I meant the other way around, remap the `C-x`-prefixed keys ;)

Oh, that's an option too :D

Issue #1857, opened 2021-11-07
Feature idea: Having a FAQ page for users and developers?


guojing0:
Hi,

In the process of building and installing Nyxt, I found [the developer README](https://github.com/atlas-engineer/nyxt/blob/master/documents/README.org) to be really useful. The [issues](https://github.com/atlas-engineer/nyxt/issues) also contain many helpful hints and solutions.

However, I found out that similar questions and answers have appeared several times, such as

1. Installing `webkit2-gtk(-devel)` instead of `webkit-gtk3` 
2. Put `(ql:quickload :cffi)
(pushnew "/opt/local/lib/" cffi:*foreign-library-directories* :test #'equal)` inside `.sbclrc` file

and so on...

So I was wondering that, to help future users and developers save time from searching the issues, maybe we could set up a FAQ page under either the README or developer README? So that before a new user/developer asks another issue, he could have the chance to test out possible solutions that were carried out in the past?

Example: [MacPorts FAQ](https://trac.macports.org/wiki/FAQ)

Ambrevar:
Should the developer's readme just be extended to include these frequently occurring bumps?

guojing0:
Yes, it's another approach.

Issue #684, opened 2020-04-20
Next Hangs/Never Opens on MacOS Catalina


heywoodlh:
**Describe the bug**
I can't seem to run Next after building from Source or installing via MacPorts.

I have run into basically all of the issues and solutions outlined here:
https://github.com/atlas-engineer/next/issues/657

I am able to compile the application and have moved the bundle to /Applications.

However, when I attempt to run Next from the CLI or launching the application a new Next window is never opened. I can see the PID but no window ever appears.

**Precise recipe to reproduce the issue**

I just followed the [Standard Developer Installation](https://github.com/atlas-engineer/next/tree/master/documents#standard-developer-installation) instructions. In order for it to actually compile, I installed `webkit2-gtk-devel` using `port`:

`sudo port install webkit2-gtk-devel`

**Information**
- OS: MacOS Catalina 10.15.4
- Graphics card and driver: Radeon Pro 560 4 GB + Intel HD Graphics 630 1536 MB
- Next version: 1.5.0-345b9606
- How you installed Next: build from source
- Lisp implementation/version: SBCL/Quicklisp Installation (followed [Standard Developer Installation](https://github.com/atlas-engineer/next/tree/master/documents#standard-developer-installation) instructions)

**Output when started from a shell**

```bash
❯ /Applications/Next.app/Contents/MacOS/next                                                                                                                                                                                     16s 934ms 
objc[10642]: Class GNotificationCenterDelegate is implemented in both /usr/local/Cellar/glib/2.64.2/lib/libgio-2.0.0.dylib (0xfa3e2e0) and /opt/local/lib/libgio-2.0.0.dylib (0x14c06210). One of the two will be used. Which one is undefined.
WARNING: cl-cffi-gtk: GType GdkX11DeviceXI2 is not known to GObject
WARNING:
   Declared GType name 'GdkX11DeviceXI2' for class 'GDK-X11-DEVICE-XI2' does not match actual GType name 'GStrv'
WARNING: cl-cffi-gtk: GType GdkX11DeviceManagerCore is not known to GObject
WARNING:
   Declared GType name 'GdkX11DeviceManagerCore' for class 'GDK-X11-DEVICE-MANAGER-CORE' does not match actual GType name 'GClosure'
WARNING: cl-cffi-gtk: GType GdkX11DeviceManagerXI2 is not known to GObject
WARNING:
[...]

Issue #657, opened 2020-04-16
Developer installation fails on macOS


thoelze1:
I am following the standard developer installation instructions in the developer readme. My OS is macOS catalina and and my package manager is macports. I have installed the following packages through macports:

```
$ port installed requested
The following ports are currently installed:
  gtk3 @3.24.18_0+x11 (active)
  qt5 @5.14.1_0 (active)
  qt5-qtwebengine @5.14.1_1 (active)
  sbcl @2.0.3_0+fancy (active)
  webkit-gtk3 @2.4.11_3+video (active)
  xorg-server @1.20.6_0 (active)
```
I fetched cl-webkit into ~/common-lisp/cl-webkit. I also fetched cl-webengine into ~/common-lisp/cl-webengine and followed the installation instructions and the installation succeeded. Finally, I installed libfixposix according to its installation instructions and it succeeded.

When I fetched next into ~/common-lisp/next and run `make all`, the compilation fails and says, "Make sure 'xclip' and latest cl-webkit are available on your system." But according to the output from `make all`, it looks like libwebkit is the problem:

```
; Loading "next/gtk-application"
[package download-manager]........................
[package hooks]...................................
[package keymap]..................................
[package scheme]..................................
..................................................
..................................................
[package cl-webkit2]..............................
[package jscore]..Unhandled LOAD-FOREIGN-LIBRARY-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                                  {1000560083}>:
  Unable to load any of the alternatives:
   ("libwebkit2gtk-4.0.37.39.3.dylib" "libwebkit2gtk-4.0.37.dylib"
    "libwebkit2gtk-4.0.dylib")

Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {1000560083}>
0: (SB-DEBUG::DEBUGGER-DISABLED-HOOK #<LOAD-FOREIGN-LIBRARY-ERROR "Unable to load any of the alternatives:~%   ~S" {10019D4443}> #<unused argument> :QUIT T)
1: (SB-DEBUG::RUN-HOOK SB-EXT:*INVOKE-DEBUGGER-HOOK* #<LOAD-FOREIGN-LIBRARY-ERROR "Unable to load any of the alternatives:~%   ~S" {10019D4443}>)
2: (INVOKE-DEBUGGER #<LOAD-FOREIGN-LIBRARY-ERROR "Unable to load any of the alternatives:~%   ~S" {10019D4443}>)
[...]

PR #1859, opened 2021-11-07
FAQ section added in the developer README

guojing0:
A simple FAQ section is now added in the developmer README per @Ambrevar's suggestion in https://github.com/atlas-engineer/nyxt/issues/1857. We can extend it as we go.

### DIFF
--- a/documents/README.org
+++ b/documents/README.org
@@ -33,7 +33,6 @@
   - [[#interacting-with-a-compiled-version-of-nyxt][Interacting with a compiled version of Nyxt]]
     - [[#slime][SLIME]]
     - [[#sly][SLY]]
-- [[#faq][FAQ]]
 - [[#help--community][Help & Community]]
   - [[#issue-tracker][Issue tracker]]
   - [[#learning-common-lisp][Learning Common Lisp]]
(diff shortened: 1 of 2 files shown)


documents/README.org
```diff
@@ -385,6 +386,22 @@ Follow the following steps:
...
+developers in the past:
+
+1. If you install dependencies via MacPorts and your Lisp is SBCL, please try
+   putting the following into your =~/.sbclrc=:
+   #+begin_src lisp
+     (ql:quickload :cffi)
+     (pushnew "/opt/local/lib" cffi:*foreign-library-directories* :test #'equal)
+   #+end_src
+
+2. If you are on a Mac, you may want to install "webkit2-gtk(-devel)" instead
```
Ambrevar, line 401: I suppose that the closing parenthesis in `webkit2-gtk(-devel)` is a typo :)

Ambrevar: The rest of the doc uses `macOS` instead of Mac, I suggest you use this too for consistency.

Ambrevar:
What about adding these 2 suggestions directly into the `macOS specific dependencies` section?

Note that the previous section `GNU/Linux, FreeBSD GTK dependencies` has a similar tip for `.sbclrc`, so it makes sense to mirror it.

documents/README.org
guojing0: > I suppose that the closing parenthesis in `webkit2-gtk(-devel)` is a typo :)

MacPorts provide both `webkit2-gtk` and `webkit2-gtk-devel` and John mentioned the latter one in relevant issues (e.g., #657 and #684). So I am unsure which one I should type.

> The rest of the doc uses `macOS` instead of Mac, I suggest you use this too for consistency.

Of course, will do.

Ambrevar:
Excellent, thanks!

Issue #1303, opened 2021-04-16
Leverage cl-https-everywhere for force-https-mode


Ambrevar:
Not all URLs should be rewritten to HTTPS, and cl-https-everywhere can help us with that:
https://github.com/ruricolist/cl-https-everywhere

hendursaga:
Isn't the EFF retiring HTTPS Everywhere? We might want to look into DuckDuckGo's Smarter Encryption extension, or [at least its list](https://github.com/duckduckgo/smarter-encryption#just-want-the-list).

Ambrevar:
Right, I found the announcement: https://www.eff.org/deeplinks/2021/04/https-everywhere-now-uses-duckduckgos-smarter-encryption

hendursaga:
Would the list being licensed CC BY-NC-SA 4.0 be a problem? Perhaps by not bundling but allowing the option to download it??

hendursaga:
Obviously we'd want to have an option, perhaps even not the default, since the [compressed file](https://staticcdn.duckduckgo.com/https/smarter_encryption_latest.tgz) is 1/4 gigabyte..

hendursaga:
So the file literally is just a bazillion hostnames, one line each. I was thinking some sort of succinct data structure, like they use for dictionary lookup? Or do we just want to throw it into an SQLite database?

Ambrevar:
If it's this big, SQLite might be a good option.

hendursaga:
What SQL / SQLite library should we use? [cl-sqlite](https://cl-sqlite.common-lisp.dev/), perhaps?

hendursaga:
And perhaps some query generator library like [sxql](https://github.com/fukamachi/sxql)??

aadcg:
I think this is  the topic of an extension.

PR #1861, opened 2021-11-08
Use cl-https-everywhere to fix the two issues in comments

guojing0:
This PR intends to implement the suggestion of leveraging [cl-https-everywhere](https://github.com/ruricolist/cl-https-everywhere) for force-https-mode from #1303, so that HTTP-only websites can remain so and certain websites do not time out.

However, I do have one question: Since cl-https-everywhere is not available via Quicklisp (like [cl-webkit](https://github.com/joachifm/cl-webkit)), should I also edit `.gitmodules`, `nyxt.asd`, `nyxt.scm`, `README.org`, `SOURCE.org`, and `global.lisp`, accordingly? (Building from source is successful after loading `cl-https-everywhere` into REPL.)

I am not sure about this, so these files were not edited.


source/mode/force-https.lisp
```diff
@@ -39,30 +39,35 @@ help on how to proceed."
...
-       (setf (quri:uri-scheme url) "https"
-             (quri:uri-port url) (quri.port:scheme-default-port "https")
-             (url request-data) url)
+       ;; Use cl-https-everywhere to update its scheme status
+       (multiple-value-bind (rewrited-uri is-https rewrite-p)
+           (https-everywhere:rewrite-uri url)
+         (let ((rewrited-scheme (if is-https "https" "http")))
+           (setf (quri:uri-scheme url) rewrited-scheme
+                 (quri:uri-port url) (quri.port:scheme-default-port rewrited-scheme)
+                 (url request-data) (url rewrited-uri))))
```
aartaka, line 53: The first two lines are not necessary here, as the last one rewrites the url of the request anyway.

source/mode/force-https.lisp
```diff
@@ -39,30 +39,35 @@ help on how to proceed."
...
 
-Example:
+   To permanently bypass the \"Unacceptable TLS Certificate\" error:
 
-\(define-configuration web-buffer
-  ((default-modes (append '(force-https-mode) %slot-default%))))"
+   \(setf nyxt/certificate-exception-mode:*default-certificate-exceptions*
+          '(\"your.unacceptable.cert.website\"))
+
+   Example:
```
aartaka, line 67: Maybe something like "To use force-https-mode add this to your config file:"?

Not sure, though. Maybe it's understandable enough.

source/mode/force-https.lisp
```diff
@@ -39,30 +39,35 @@ help on how to proceed."
...
 
 (define-mode force-https-mode ()
   "Impose HTTPS on every queried URL.
-Use at your own risk -- it can break websites whose certificates are not known
-and websites that still don't have HTTPS version (shame on them!).
 
-To permanently bypass the \"Unacceptable TLS Certificate\" error:
-\(setf nyxt/certificate-exception-mode:*default-certificate-exceptions*
-       '(\"your.unacceptable.cert.website\"))
+   Use at your own risk -- it can break websites whose certificates are not
```
aartaka, line 59: Is that additional indentation? Why so?

Remember, it renders to e.g. `describe-function` pages and thus will be broken if you try to indent it in such a way.

source/mode/force-https.lisp
```diff
@@ -39,30 +39,35 @@ help on how to proceed."
        (https->http-loop-help (buffer request-data) url)
```
aartaka, line 39: @Ambrevar, do we need to detect HTTP->HTTPS->HTTP loops if we use `cl-https-everywhere`?

aartaka (commented):
Thanks for the pull request, it does make sense :)

Regarding the library: yes, adding it to submodules and everywhere else will be of great help!

source/mode/force-https.lisp
guojing0: My bad, I will revert the indentations.

source/mode/force-https.lisp
Ambrevar: I don't know, I've never used cl-https-everywhere, but if the library handles for us, then great!  Let's test it!

source/mode/force-https.lisp
```diff
@@ -39,21 +39,22 @@ help on how to proceed."
...
        ;; `previous-url' as well.
        (setf (previous-url mode) (quri:copy-uri url))
        (log:info "HTTPS enforced on '~a'" (render-url url))
-       ;; FIXME: http-only websites are displayed as "https://foo.bar"
-       ;; FIXME: some websites (e.g., go.com) simply time-out
-       (setf (quri:uri-scheme url) "https"
-             (quri:uri-port url) (quri.port:scheme-default-port "https")
-             (url request-data) url)
+       ;; Use cl-https-everywhere to update its scheme status
+       (multiple-value-bind (rewrited-uri is-https rewrite-p)
```
Ambrevar, line 48: Typo: `rewrited` -> `rewritten`.

source/mode/force-https.lisp
guojing0: Спасибо.

guojing0:
I am not certain about how to change `nyxt.scm` and `changelog.lisp`, so they remain unchanged.

Ambrevar:
For `nyxt.scm`, don't bother, cl-https-everywhere will have to be included in Guix first.  I can do it.

Same for the changelog, I can do it if you are in doubt.

nyxt.asd
```diff
@@ -334,7 +334,8 @@ A naive benchmark on a 16 Mpbs bandwidth gives us
 (defsystem "nyxt/gtk"
   :depends-on (nyxt
                cl-cffi-gtk
-               cl-webkit2)
+               cl-webkit2
+               cl-https-everywhere)
```
Ambrevar, line 338: Any reason you've put it in this system?  I believe it belongs to `nyxt` directly.  This should fix this patch as currently it does not load properly and does not pass the CI.

documents/SOURCES.org
```diff
@@ -50,6 +50,7 @@ which external dependencies we are directly relying on.
 Renderer-specific dependencies:
 - cl-cffi-gtk
 - cl-gobject-introspection
+- cl-https-everywhere
```
Ambrevar, line 53: Why do you list it as a renderer-specific dependency?

documents/README.org
```diff
@@ -119,6 +119,7 @@ Otherwise clone each repository mentioned in =.gitmodules= into
 #+begin_src sh
   mkdir -p ~/common-lisp
   git clone https://github.com/joachifm/cl-webkit ~/common-lisp/cl-webkit
+  git clone https://github.com/ruricolist/cl-https-everywhere.git ~/common-lisp/cl-https-everywhere
```
Ambrevar, line 122: Don't bother with this line, this part of the readme has to be rewritten.

source/mode/force-https.lisp
```diff
@@ -39,27 +39,28 @@ help on how to proceed."
...
-and websites that still don't have HTTPS version (shame on them!).
+Use at your own risk -- it can break websites whose certificates are not
+known, and websites that still don't have HTTPS version (shame on them!).
 
 To permanently bypass the \"Unacceptable TLS Certificate\" error:
 \(setf nyxt/certificate-exception-mode:*default-certificate-exceptions*
        '(\"your.unacceptable.cert.website\"))
 
-Example:
+Example: To use `force-https-mode', add this to your config file:
```
Ambrevar, line 63: Remove `Example :` then.

nyxt.asd
guojing0: Since `https-everywhere` is not available in Quicklisp as `cl-webkit`, so I used `cl-webkit` as search keyword and placed `https-everywhere` under it. I will fix it then.

documents/README.org
guojing0: I could not find a good place to place that, so I left it there.

nyxt.asd
guojing0: It seems that it still does not pass the CI:
```
caught error #<QUICKLISP-CLIENT:SYSTEM-NOT-FOUND {1001F3E973}>
System "cl-https-everywhere" not found
```

Is the error on my side due to changing something, or that it cannot load "cl-https-everywhere" correctly?

Ambrevar: The problem is how you added the Git submodule.  You can't just do it by editing `.gitmodules` by hand, you _must_ use the `git submodule` shell command.  Typically I would add it like this:

```sh
git submodule add https://github.com/ruricolist/cl-https-everywhere _build/cl-https-everywhere
```

For the Guix part of the CI, don't bother about it, I'll add the cl-https-everywhere package soon.

Ambrevar: I've added cl-https-everywhere to guix, the Guix CI should be pass now.

source/mode/force-https.lisp
Ambrevar: Any example that comes to mind?

nyxt.asd
guojing0: Interesting, it seems that both jobs (Ubuntu and Guix) cannot find cl-https-everywhere correctly.

source/mode/force-https.lisp
guojing0: Would this example [0] help? It does not support HTTPS.

[0]: http://uniqueoldyounglight.neverssl.com/online

Ambrevar: It does support HTTPS for me :p

nyxt.asd
Ambrevar: According to https://github.com/atlas-engineer/nyxt/runs/4213690610?check_suite_focus=true#step:8:35, you did.  Something else is wrong...

source/global.lisp
```diff
@@ -53,6 +53,7 @@ execution of Nyxt.")
 (alex:define-constant +nyxt-critical-dependencies+
   '(:cl-cffi-gtk
     :cl-gobject-introspection
+    :cl-https-everywhere
```
Ambrevar, line 56: I would not classify it as a critical dep :)

guojing0: Same mistake as I mentioned above as following the appearances of `cl-webkit`.

I have removed this line but somehow the Ubuntu action still does not pass. (It's really late here I will go to sleep, but I will check any replies in the morning.)

Ambrevar:
So I tried running this on Guix, and cl-https-everywhere seems to have issues loading (some `vernacular` bugs?).
This will need some further investigation.

Ambrevar:
I've reported the issue you just mentioned here: https://github.com/ruricolist/cl-https-everywhere/issues/1

hendursaga:
@guojing0 with my PR https://github.com/atlas-engineer/nyxt/pull/2083 the dependency-adding steps have been documented now

jmercouris:
Closing this pull request for now as it is a bit out of date. If it is to be updated, please open a new pull request!

Ambrevar:
As explained in #1303, we should probably look into DuckDuckGo smarter encryption.

PR #1863, opened 2021-11-08
Buffer refactor

jmercouris:
Cleaning up buffer.lisp :-)

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -14,6 +14,7 @@
     :documentation "Unique identifier for a buffer.
 Dead buffers or placeholder buffers (i.e. those not associated with a web view)
 have an empty ID.")
+   ;; TODO: Or maybe a dead-buffer should just be a buffer history?
    (data-profile
     (make-instance (or (find-data-profile (getf *options* :data-profile))
                        'default-data-profile))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -923,7 +923,7 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
     (mapc (alex:rcurry #'make-mode buffer) extra-modes)
     (when dead-buffer                   ; TODO: URL should be already set.  Useless?
       (setf (url buffer) (url dead-buffer)))
-    (buffers-set (id buffer) buffer)
+    (buffer-set *browser* (id buffer) buffer)
     (unless no-history-p
       ;; When we create buffer, current one can override the
       ;; data-profile of the created buffer. This is dangerous,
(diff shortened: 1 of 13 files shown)


jmercouris:
There is a lot more to do.

source/buffer.lisp
```diff
@@ -9,15 +9,16 @@
...
 
 (define-class buffer ()
-  ((id ""
-       :documentation "Unique identifier for a buffer.
+  ((id
+    ""
+    :documentation "Unique identifier for a buffer.
 Dead buffers or placeholder buffers (i.e. those not associated with a web view)
 have an empty ID.")
-   ;; TODO: Or maybe a dead-buffer should just be a buffer history?
```
Ambrevar, line 16: Do you have an answer to this?

Ambrevar:
The macro re-indentation is off.  Make sure to load the system before indenting, otherwise SLIME does not know if the `car` of a list is a function or a macro and thus indents it like a function.

source/buffer.lisp
jmercouris: Yes, I'm going to completely redo "dead buffers" , the web view will be a lazily loaded slot

Ambrevar: This might be slightly different from what the TODO comment suggests:
- you are talking about _how to instantiate the web view_ (dead buffer or not);
- that comment talks about how to represent the concept of dead buffers.

Thus it seems to me that the comment still needs reflecting upon :)

Ambrevar:
About `buffers-get`, `buffers-set` and `buffers-delete`, we can replace the hash-table with a `cl-custom-hash-table`.  See history tree for an example.  In this case, I guess we could just use the following hash function:

```lisp
(defun custom-hash-function (entry)
  (print-status)
  (sxhash entry))
```

The drawback is that the status area is refreshed even when just _getting_ an entry.

Alternatively, we might be able to leverage https://alex-gutev.github.io/generic-cl/#generic_hash_tables or cl-containers.

source/buffer.lisp
jmercouris: I am suggesting dead buffers will cease to exist as a concept. It doesn't matter to Nyxt whether a buffer is dead or alive, it will be transparent to all the code that uses a buffer.

Ambrevar: I think we are not on the same page :) 
What you are talking about is the implementation detail of how to manage the views of buffers, dead or alive.
What I'm talking about is how we "mark" dead buffers.
In other terms, when the user deletes a buffer, what happens? (I'm not talking about the web view here.)

jmercouris: What happens when a buffer is deleted? 

+ It is removed from the `buffers-map` object of the browser class slot `buffers`. 
+ Its web-view is deleted
+ Its destroy method is invoked

Should we wish to restore it, we only have to re add it to the `buffers-map` and recreate the web view.

Ambrevar: So _what_ do we re-add?  It means it must be stored somewhere.
This is what the comment is about: where do we store this "dead" buffer object.
An idea was to leverage the global history.

Since this is unrelated to this pull request, I suggest we restore the comment.

jmercouris: I was planning on handling it in this pull request. This is just a draft actually, I just wanted to give you a sneak peek as I was working on things! I can do it in another pull request as well though.

Ambrevar: I think we are not on the same page :) 
What you are talking about is the implementation detail of how to manage the views of buffers, dead or alive.
What I'm talking about is how we "mark" dead buffers.
In other terms, when the user deletes a buffer, what happens? (I'm not talking about the web view here.)

jmercouris:
As discussed with Pierre: merging this for now and leaving greater buffer refactor work for a greater/different pull request.

jmercouris:
Actually, I will close this pull request for now and will probably cherry pick some of the changes I've made.

PR #1868, opened 2021-11-14
Set User Agent to a more generic one in reduce-tracking-mode.

aartaka:
This adds User Agent customization to `reduce-tracking-mode`. As a fun exercise, I was able to open Jitsi when setting `nyxt/reduce-tracking-mode:preferred-user-agent` to `"Mozilla/5.0 (X11; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0"` (Firefox on Linux).

# What's new:
- `old-user-agent` and `preferred-user-agent` slots in `reduce-tracking-mode`.
  - It uses them to store the old User Agent and set new, a bit more generic User Agent, respectively.
- `ffi-buffer-user-agent`'s second argument is optional now.
  - When it's there, `ffi-buffer-user-agent` behaves like a setter.
  - When it's not there, `ffi-buffer-user-agent` behaves like getter.

# Why Safari?
Well, MacOS is pretty close to GNU/Linux and BSD, while Safari uses the same rendering engine. Thus, the website breakage should be minimal, while we will still be able to disappear in the Safari users crowd (however small it is).

# How to Test 
- Run Nyxt with this patch.
- Enable `reduce-tracking-mode`.
- Navigate to https://techblog.willshouse.com/2012/01/03/most-common-user-agents.
- It should identify you as a MacOS Safari User.

How does it look?

### DIFF
--- a/source/mode/reduce-tracking.lisp
+++ b/source/mode/reduce-tracking.lisp
@@ -8,7 +8,7 @@
 
 (define-mode reduce-tracking-mode ()
   "Set specific settings in the web view in order to mitigate fingerprinting,
-(how third-party trackers attempt to indentify you.
+(how third-party trackers attempt to indentify you).
 
 Fingerprinting can be tested with https://panopticlick.eff.org/."
   ((preferred-languages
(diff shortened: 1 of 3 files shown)


source/mode/reduce-tracking.lisp
```diff
@@ -8,16 +8,30 @@
...
     :type list-of-strings
     :documentation "The list of languages that will be sent as part of the
 Accept-Language HTTP header.")
+   (preferred-user-agent
+    ;; Check https://techblog.willshouse.com/2012/01/03/most-common-user-agents
+    ;; occasionally and refresh when necessary.
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.0 Safari/605.1.15"
+    :type string
+    :documentation "The user agent to set when enabling `reduce-tracking-mode'.
+It's Safari on MacOS by default, because this way we break less websites while
```
Ambrevar, line 25: `macOS` maybe?

source/mode/reduce-tracking.lisp
```diff
@@ -8,16 +8,30 @@
...
+still being less noticeable in the crowd.")
+   (old-user-agent
+    nil
+    :type (or null string)
```
Ambrevar, line 29: Maybe `:export nil` since it's an implementation detail?

Ambrevar:
Looks good to me!

source/mode/reduce-tracking.lisp
edgar-vincent: Also s/less/fewer :)

aartaka:
All points addressed, merging.

PR #1869, opened 2021-11-15
Fix blocker-mode

Ambrevar:
This should in particular fix the hang when blocker-mode updates the URL list.


Ambrevar:
In particular, look at commit https://github.com/atlas-engineer/nyxt/pull/1869/commits/97a2e6603959ff09ad87302761334078abd319ea.

Needs testing!

@jmercouris @aartaka ?

Ambrevar:
Feedback?  OK to merge?

source/mode/blocker.lisp
```diff
@@ -44,32 +44,37 @@ See `*default-hostlist*' for an example."))
...
-              (log:error "Could not persist hostlist ~s: ~a" path c))))
-        hosts))))
-
-(defvar update-lock (bt:make-lock))
-
-(defmethod update-hostlist-with-lock ((hostlist hostlist))
+                url-string)
+      (handler-case
+          (let ((hosts (dex:get url-string)))
+            (when (and hosts path)
```
aartaka, line 57: When here ensures we don't do anything on a failed load. Maybe have `handler-case` to only wrap `dex:get` instead of the whole function body then? Or have just one big `handler-case` to catch all types of errors?

Rationale: aesthetics. Several nested `handler-case` look surprisingly unpretty.

source/mode/blocker.lisp
```diff
@@ -44,32 +44,37 @@ See `*default-hostlist*' for an example."))
...
+              (load-hostlist hosts mode)
+              (handler-case
+                  (alex:write-string-into-file hosts (ensure-parent-exists path)
```
aartaka, line 60: Hmmmm, this and the whole locking business looks a lot like `data-path` handling. Maybe use `data-path`'s `store` and `restore`?

aartaka (commented):
Looks good, except for some suspicious similarity between `data-path` and what we have for hostlist fetching.

It's not critical, though.

source/mode/blocker.lisp
Ambrevar: My rationale was that `dex:get` may raise unexpected conditions, same for file writes.  The benefit of having 2 handler-case is that we can report different messages depending on where it failed, regardless of the condition.

source/mode/blocker.lisp
Ambrevar: Absolutely, I was coming to the same conclusion.  blocker-mode was written long before `data-path` was designed.
Not just that, but the rest of the code is very rusty compared to our new conventions and habits.

source/mode/blocker.lisp
aartaka: The conditions that `dex:get` raises are still a restricted set, so we most probably can enumerate those by names. Same for `alex:write-string-into-file`. But yes, it's quite some work, so having two generic `handler-case`s may be optimal, after all.

aartaka:
I tried to compile and run the binary of this patch rebased over `master` (to include other fixes).

What I get when typing `github.com` in the `set-url` prompt and hitting RET:
```
<WARN> [13:44:07] Warning: Error on separate thread: There is no applicable method for the generic function
                            #<STANDARD-GENERIC-FUNCTION NYXT/BLOCKER-MODE::LOAD-HOSTLIST (2)>
                          when called with arguments
                            (#<NYXT/BLOCKER-MODE:USER-BLOCKER-MODE {100B7C4783}>
                             #<NYXT/BLOCKER-MODE:HOSTLIST {1002D11503}>).
See also:
```
And new buffers don't get created.

Ambrevar:
I'm on it.

source/mode/blocker.lisp
```diff
@@ -135,13 +141,17 @@ Example:
...
+    (make-hash-table :test 'equal)
+    :documentation "The set of domain names to block.")
    (destructor
     (lambda (mode)
       (when (web-buffer-p (buffer mode))
         (hooks:remove-hook (request-resource-hook (buffer mode))
                            'request-resource-block))))
    (constructor
     (lambda (mode)
+      (mapc (alex:curry #'load-hostlist mode) (hostlists mode))
```
aartaka, line 154: Shouldn't it be `alex:rcurry` instead? The error I got was because the order of arguments was swapped.

aartaka:
Having the `alex:rcurry`, it runs just fine!

More so, I've recently got hangs back, and this seems to fix them.

I'll test on my second machine and see if it fixes hangs there too.

aartaka:
Fixes hangs on both of my machines! No other problems noticed :)

Ambrevar:
Fantastic, thanks for testing!

I'll push a rewrite soon, should rid us of the global lock and simplify
the logic.

source/mode/blocker.lisp
aartaka: Regarding `dex:get` errors: Given that `dexador` mostly wraps around `usocket`, we will be safe by using `usocket:socket-error` as the generic error type for downloads.

`alex:write-string-into-file` relies on built-in CL file-handling, and thus `file-error` is the generic error we can catch.

Given those two, we can have one `handler-case` around the whole function body, with just two handlers -- `usocket:socket-error` and `file-error` ones.

Ambrevar:
@aartaka I've just pushed a rewrite which turns `hostlist` into a data-path and uses the data-path API.

It fixes a few more bugs.

Could you test again and confirm it works for you?

aartaka:
Compilation gives the warning:
```
../git/nyxt/source/mode/blocker.lisp:21:5:
  warning: Constant NIL conflicts with its asserted type (CONS STRING T).
```

This is because `nil` is used for `hosts` typed as `(cons string *)` and thus failing the type check. A type of `list-of-strings`, `cons`, or `(or nil (cons string *))` should fix that.

source/mode/blocker.lisp
Ambrevar: I'm now using the data-path API, so `write-string-into-file` is gone, so no more nested `handler-case`!

aartaka:
Fixing this type warning lets me run Nyxt and it's all smooth after that. I tried deleting the hostlists file in runtime or starting Nyxt without the file, and it's gracefully restoring the file without me even noticing it.

Ambrevar:
Type fixed.
Thanks for testing!
I'll merge as soon as the CI passes ;)

source/mode/blocker.lisp
```diff
@@ -44,75 +48,96 @@ See `*default-hostlist*' for an example."))
...
+  "Load BLOCKER-MODE's hostlists into `blocked-hosts' in the background."
+  (dolist (hostlist (hostlists blocker-mode))
+    (when force-update-p
+      ;; TODO: Move and export `copy-object' to a separate library.
+      (setf hostlist (prompter::copy-object hostlist))
+      (setf (force-update-p hostlist) t))
+    (calispel:! (worker-channel blocker-mode) hostlist)))
+
+(defun worker (mode)
+  (do () (nil)
```
aartaka, line 132: What a strange pattern 0_o I'd use `(loop while t ...)` for readability, but this is fine :)

source/mode/blocker.lisp
```diff
@@ -4,34 +4,38 @@
...
   (:documentation "Block resource queries for listed hosts."))
 (in-package :nyxt/blocker-mode)
 (use-nyxt-package-nicknames)
 
 ;; TODO: Add convenient interface to block hosts depending on the current URL.
 
-(defclass hostlist-data-path (data-path) ())
-
-(define-class hostlist ()
+(define-class hostlist (data-path)
```
aartaka, line 14: Ha, that's a smart inheritance!

source/mode/blocker.lisp
```diff
@@ -4,34 +4,38 @@
...
-    :documentation "Where to find the list locally.
-If nil, the list won't be persisted.
-If path is relative, it will be set to (xdg-data-home path).")
    (hosts
     '()
-    :documentation "The list of domain names.")
+    :type (or (cons string *) null)
+    :documentation "List of hosts to ignore.
+This is useful to reference hosts manually instead of via `url'.")
+   (content
```
aartaka, line 25: I mean, this `content` hints that maybe we're still not using the `data-path` to the full. Maybe we should rather store the lispy parsed version of hostlists of the user machine, using the `.txt` file to only initially load it? This way we need no `content` and getting hostlists will be as simple as `with-data-unsafe`.

source/mode/blocker.lisp
```diff
@@ -44,75 +48,96 @@ See `*default-hostlist*' for an example."))
...
+                    (when file
+                      (uiop:read-file-string file)))))
+        (when data
+          (nyxt::%set-data path data)))
+    (error (c)
+      (echo-warning "Failed to load hostlist from ~s: ~a"
+                    (expand-path path) c))))
+
+(-> fetch (hostlist) string)
+(defun fetch (hostlist)
```
aartaka, line 68: Maybe integrate this into `restore`? See my comment about the `content` slot of the `hostlist` above.

source/mode/blocker.lisp
```diff
@@ -135,13 +160,34 @@ Example:
...
     (lambda (mode)
+      (bt:destroy-thread (thread mode))
       (when (web-buffer-p (buffer mode))
         (hooks:remove-hook (request-resource-hook (buffer mode))
                            'request-resource-block))))
    (constructor
     (lambda (mode)
+      (setf (worker-channel mode) (nyxt::make-channel))
+      (setf (thread mode)
+            (run-thread "async-data-path worker"
```
aartaka: Wait, I remember we did have some async path processing abstractions for history data, for example. Isn't it `async-data-path` that should manage code like that?

aartaka (commented):
This looks alright, although I'm still unsure whether we use `data-path` fully.

Ambrevar:
Good point about `fetch` belonging to `restore` and using `async-data-path` instead.  I'll work on it!

aartaka:
Sure :) Thanks for this patch, Pierre!

Issue #1681, opened 2021-07-28
2.2 test period


Ambrevar:
I've updated the https://github.com/atlas-engineer/nyxt/tree/2-series branch.

We can target next week for the 2.2 release.
Everyone is welcome to try out the `2-series` branch and report outstanding bugs.

@jmercouris @aartaka Use it as your default Nyxt if you can :)

Ambrevar:
@aartaka wants to include 33212652f596bf58d77fd798f661afebd528faa7.

As of 51e4a0fc870e65c664804a6f99e409d20e044e5a, `master` has the following significant changes:

- b665471bb7829f5e0f4092340a037dbfa82b43c2: history: Fix dead-lock possibility in restore-history-buffers.
- c5f363f05f80e0be07ac1d4f3bdb16a4e5fd35e6: Add support for GTK clipboard

OK to update `2-series` then?

jmercouris:
Ok!

Ambrevar:
Master contains important fixes (notably for the GHT) but it's not ready yet, so I suggest we wait a little bit.

Ambrevar:
Back to business after a long break :)

I've just updated the changelog on the 2-series branch.
@jmercouris @pdelfino @aartaka Can you confirm it looks good for you?

We can release 2.2 tomorrow if that's OK.

There are a few more bugs that'd be nice to fix.  Say, #1783, #1792, #1757, #1678, #1794, #1755, #1743.

Nothing critical though and I'm personally OK with releasing 2.2.0 now and fixing the rest for 2.2.1.  Thoughts?

aartaka:
Yes, looks alright to me! We've been holding this release off for too long :)

jmercouris:
I also confirmed earlier today with Pierre! :-O

Ambrevar:
Another thing I'd like to do for the remaining 2.2.* releases is backport all the backward compatible changes on master.

Ambrevar:
Released!

Ambrevar:
Also please test 2-series for at least a few minutes :)

Ambrevar:
There are still many more bug fixes for the 2.2 series! ;)
  See https://github.com/atlas-engineer/nyxt/issues/1681#issuecomment-940834785.

Ambrevar:
I'd like to release 2.2.3 tomorrow after merging the fixes on master and https://github.com/atlas-engineer/nyxt/pull/1869.

OK with everyone?
[...]

PR #1872, opened 2021-11-16
Handle onbeforeunload prompts.

aartaka:
This enables the "Do you want to leave this page"-like prompts for the buffers with the unfinished input or otherwise `onbeforeunload`-invoking code. This way, one has much less chances to lose their input/progress when doing something in Nyxt.

@Ambrevar, this it the buffer killing prompt I've been talking about in #1681.

# Caveats
- What used to be `buffer-delete` is now effectively two functions: `buffer-delete` and `buffer-hide`, the latter of which should be called from `ffi-buffer-delete`. This is not a perfect design, but I had no idea of how to do it better, given the overall WebKit brittleness there -- it does not provide us with async processing, callbacks, anything. You simply call `webkit_web_view_try_close` and process it elsewhere.
- `webkit_web_view_try_close`. It's there as a binding in `cl-webkit`, and this patch requires the lates `cl-webkit`. On a related note: is it alright when I bump up the `cl-webkit` version every week? :) I feel like I shouldn't, but we need to depend on some version of `cl-webkit` in Nyxt, thus the version bumping.

# Hot To Tests
Open one of the examples on https://www.w3schools.com/jsref/event_onbeforeunload.asp and try to close the buffer/click any link.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -913,19 +913,25 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
 (defun buffer-delete (buffer)
   "For dummy buffers, use `ffi-buffer-delete' instead."
   (hooks:run-hook (buffer-delete-hook buffer) buffer)
+  (with-data-access (history (history-path buffer))
+    (sera:and-let* ((owner (htree:owner history (id buffer)))
+                    (current (htree:current owner))
+                    (data (htree:data current)))
+      (setf (nyxt::scroll-position data) (nyxt:document-scroll-position buffer))))
+  (ffi-buffer-delete buffer))
+
+(defun buffer-hide (buffer)
+  "Stop showing the buffer in Nyxt.
+Should be called from/instead of `ffi-buffer-delete' when the renderer view
+associated to the buffer is already killed."
   (let ((parent-window (find buffer (window-list) :key 'active-buffer)))
     (with-data-access (history (history-path buffer))
-      (sera:and-let* ((owner (htree:owner history (id buffer)))
-                      (current (htree:current owner))
-                      (data (htree:data current)))
-        (setf (nyxt::scroll-position data) (nyxt:document-scroll-position buffer)))
       (htree:delete-owner history (id buffer)))
     (when parent-window
       (let ((replacement-buffer (or (first (get-inactive-buffers))
                                     (make-buffer :load-url-p nil
                                                  :url (quri:uri "about:blank")))))
         (window-set-buffer parent-window replacement-buffer)))
-    (ffi-buffer-delete buffer)
     (buffers-delete (id buffer))
     ;; (setf (id buffer) "") ; TODO: Reset ID?
     (add-to-recent-buffers buffer)))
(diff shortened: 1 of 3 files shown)


Ambrevar:
About cl-webkit version bump: it does not harm anyone, so as long as we are consistent, it's all good in my opinion!

source/buffer.lisp
```diff
@@ -913,19 +913,20 @@ If DEAD-BUFFER is a dead buffer, recreate its web view and give it a new ID."
...
+(defun buffer-hide (buffer)
+  "Stop showing the buffer in Nyxt.
+Should be called from/instead of `ffi-buffer-delete' when the renderer view
+associated to the buffer is already killed."
   (let ((parent-window (find buffer (window-list) :key 'active-buffer)))
     (with-data-access (history (history-path buffer))
-      (sera:and-let* ((owner (htree:owner history (id buffer)))
-                      (current (htree:current owner))
-                      (data (htree:data current)))
-        (setf (nyxt::scroll-position data) (nyxt:document-scroll-position buffer)))
```
Ambrevar, line 921: Why remove this?

source/renderer/gtk.lisp
```diff
@@ -1073,14 +1085,10 @@ See `gtk-browser's `modifier-translator' slot."
...
 
-(define-ffi-method ffi-buffer-delete ((buffer gtk-buffer))
-  (mapc (lambda (handler-id)
-          (gobject:g-signal-handler-disconnect (gtk-object buffer)
-                                               handler-id))
-        (handler-ids buffer))
-  (when (slot-value buffer 'gtk-object) ; Not all buffers have their own web view, e.g. prompt buffers.
-    (gtk:gtk-widget-destroy (gtk-object buffer))
-    (setf (gtk-object buffer) nil)))
+(defmethod ffi-buffer-delete ((buffer gtk-buffer))
```
Ambrevar, line 1088: Did you intend to make it a defmethod instead of a define-ffi-method?  Looks like `webkit-web-view-try-close` should be run from the GTK thread, no?

source/renderer/gtk.lisp
```diff
@@ -1028,7 +1033,14 @@ See `gtk-browser's `modifier-translator' slot."
   (connect-signal buffer "web-process-crashed" nil (web-view)
     (echo-warning "Web process crashed for buffer ~a" (id buffer))
     (log:debug "Web process crashed for web view ~a" web-view)
-    (delete-buffer :id (id buffer)))
+                  (delete-buffer :id (id buffer)))
```
Ambrevar, line 1036: Indentation off?

Ambrevar:
The `buffer-hide` introduction looks alright to me.  Can't think of a better design, but we can overhaul the buffer API for 3.0.

source/buffer.lisp
aartaka: It was throwing errors because of `gtk-object` no longer existing for `buffer`. I removed it in haste, but now I think that it can just fit into `buffer-delete`.

source/renderer/gtk.lisp
aartaka: Looking at a recent fix for `ffi-display-uri`, I tend to avoid `define-ffi-method` now. `webkit-web-view-try-close` does not need to be running on the GTK thread, no -- the patch works fine without it.

Ambrevar: Excellent!

Ambrevar:
Can you update the cl-webkit package and/or the nyxt.scm so that the CI passes?
Then I'll merge.

Ambrevar:
I mean, feel free to merge.

aartaka:
Rebased over master to resolve merge conflicts, updated cl-webkit for Guix and as a submodule. Merging.

PR #1879, opened 2021-11-18
Add TAGS file to .gitignore

hendursaga:
The TAGS file is created when you use tools like `ctags`/`etags`, which is used for navigating a codebase.


jmercouris:
Thanks! You can also add it to your global gitignore if you use such a feature; `~/.config/git/ignore` :-)

PR #1881, opened 2021-11-18
Replace hu.dwim.defclass-star with class* where needed

hendursaga:
Just a little tidy-up!

### DIFF
--- a/libraries/history-tree/history-tree.lisp
+++ b/libraries/history-tree/history-tree.lisp
@@ -46,7 +46,7 @@ goes owner-less, we can still consult the last time it was accessed.")
    (nodes '()
           :type list
           :documentation "The list of nodes that access an entry."))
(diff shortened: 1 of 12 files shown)


Ambrevar:
Thanks!

It does not pass the CI though, it seems that `class*` is unknown to `history-tree`.

hendursaga:
Sorry about that! This is confusing - it worked for me, but maybe that was because for whatever reason, class* was already loaded in my browser instance as a nickname or something?? I suppose you could discard this if you want.

Ambrevar:
After some digging, I found out that we had some issues with our test CI.  I've fixed them.

Ambrevar:
Merged with 9f4e26f46a3d6873eca0ab3923063e441e7bd9c3.
Thanks!

PR #1889, opened 2021-11-24
Enable color themes.

aartaka:
This adds an ability to style Nyxt with color themes that are short, nice, and consistent across all of the Nyxt interface.

Colors are generated at object initialization time with the help of `themed-css` macro.

# Caveats
- I do not like the implementation of `themed-css` at all, but that's the only was I could make sure that theme colors are inserted into the otherwise constant expression at the _object initialization time_, not compile time. Any macrology advice will be appreciated here!
- Some colors may be slightly off, as I've been rearranging things a lot to make the theme have a minimal number of colors while still being pretty. Please, test in all the use-cases you have for Nyxt.
- `diff-mode` is not altered, as its red and green colors are not exactly themable. Or are they? Maybe we need to have a list of generic colors in a theme then. Something like `%red%`, `%yellow%` etc. What do you think?
- The way `themed-css` is built causes some warnings about variabled being undefined. I haven't yet found a way to mitigate this. How does one ensure symbols don't escape the scope of the macro?

# Testing
- Add a theme configuration to your initfile. Two options I suggest are:
  - Laconia (because I want [my theme](https://github.com/aartaka/laconia-theme) to be everywhere and easy to setup)
``` lisp 
(define-configuration browser
  ((theme (make-instance
           'theme
           :dark-p t
           :background-color "black"
           :text-color "white"
           :accent-color "#CD5C5C"
           :primary-color "#556B2F"
           :secondary-color "lightgray"
           :tertiary-color "gray"
           :quaternary-color "dimgray"))))
```
  - A slightly simplified Drakula (built thanks to the effort of people in https://github.com/dracula/nyxt)
``` lisp
(define-configuration browser
  ((theme (make-instance
           'theme
           :dark-p t
           :background-color "#282a36"
           :text-color "#f8f8f2"
           :accent-color "#44475a"
           :primary-color "#6272a4"
           :secondary-color "#44475a"
           :tertiary-color "#ff5555"
           :quaternary-color "#6272a4"))))
```
- Run Nyxt and enjoy a whole new style (o^^)o


aartaka:
Okay, there's no more warnings about unbound variables.

Ambrevar:
> diff-mode is not altered, as its red and green colors are not exactly themable. Or are they? Maybe we need to have a list of generic colors in a theme then. Something like %red%, %yellow% etc. What do you think?

Why not have `diff-added` and `diff-removed` properties instead, like Emacs does?

I don't think we should have `%red%`, etc. since making "red" another color is semantically twisted :)

source/theme.lisp
```diff
@@ -0,0 +1,142 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt)
```
Ambrevar, line 4: Would it make sense to move it to a separate package?  A separate library?

source/theme.lisp
```diff
@@ -0,0 +1,142 @@
...
+
+(define-class theme ()
+  ((dark-p
+    nil
+    :documentation "Whether the theme is dark.
+Dark themes override `nyxt/style-mode:dark-mode' colors too.")
+   (background-color
+    "white"
+    :type string
+    :documentation "The background color of the theme")
```
Ambrevar, line 14: Missing trailing `.`.  Same below.

source/theme.lisp
```diff
@@ -0,0 +1,142 @@
...
+Should contrast with the `background-color'.")
+   (quaternary-color
+    "gainsboro"
+    :type string
+    :documentation "The quaternary color.
+Should strongly contrast with the `text-color'.")
+   (accent-color
+    "#37a8e4"
+    :type string
+    :documentation "The color of the accented elements that need attention.
```
Ambrevar, line 43: Maybe `emphasized` or `highlighted` elements instead of `accented`.

source/theme.lisp
```diff
@@ -0,0 +1,142 @@
...
+            :border-style \"solid\"
+            :border-width \"1px\"
+            :border-color %tertiary%)
+           (p
+            :color (%if-dark %accent% %secondary%)
+            :background-color %text%))"
+  `(let ((%theme% ,theme))
+     (declare (ignorable %theme%)
+              (special %theme%))
+     (macrolet ((,(intern "%IF-DARK") (DARK light)
```
Ambrevar, line 126: Why is `DARK` capitalized?

source/theme.lisp
```diff
@@ -0,0 +1,142 @@
...
+Internal variables, what you usually want is `theme' slot of `browser'.")
+
+(defun substitute-special-theme-symbols (rule)
+  "A helper function to rebuilt the themed-css expressions so that everything
+except special theme variables is quoted."
+  (typecase rule
+    (symbol
+     (if (str:s-member (list "%background%" "%text%" "%primary%" "%secondary%"
+                             "%tertiary%" "%quaternary%" "%accent%" "%font%")
+                       (str:downcase (symbol-name rule)))
```
Ambrevar, line 65: Why not just test for symbols? 
If the problem is with the package prefix, then maybe this calls for keywords instead.

Ambrevar:
The macros are quite difficult to process indeed.  At this point, I wonder if we could not simplify the design.

What about this: move the functions to a separate package, then define the special variables as `defparameters` there.  This would you won't need to substitute anything, you can just process regular symbols.

The special theme symbols would then look like this: `theme:background` (or maybe `theme:*background*`, but I'd rather leave out the ear-muffs for conciseness and accessibility).

Bonus: The user would get completion! :)

Do you think that's doable?

aartaka:
> Why not have `diff-added` and `diff-removed` properties instead, like Emacs does?

I mean, I want to have as little variables as it's only possible, that's why I'm thinking about generic colors instead of domain-specific ones. This way, we have much more control over interface style with little code, while all the places that the divergence from the style is needed in are covered by the plain `cl-css:css`.

Emacs is a bad example, I guess, as it manages style complexity through named style inheritance, while we have a quirky yet powerful CSS for it :)

> I don't think we should have `%red%`, etc. since making "red" another color is semantically twisted :)

It's twisted, but one may actually want it. Think color-blind-friendly color profiles. Configuring those with a single list of generic colors will be much easier than with a bunch of disjoint mode-specific colors :)

aartaka:
> What about this: move the functions to a separate package, then define the special variables as `defparameters` there. This would you won't need to substitute anything, you can just process regular symbols.

While the macro-magic will still be there, it will significantly simplify it. Sounds good!

> The special theme symbols would then look like this: `theme:background` (or maybe `theme:*background*`, but I'd rather leave out the ear-muffs for conciseness and accessibility).

Indeed, that sounds like a pretty clean design.

> Bonus: The user would get completion! :)

Yes, and `describe-any`!

> Do you think that's doable?

It definitely is. I will follow this course of action. Should it be a `libraries/theme/` then?

source/theme.lisp
aartaka: Oooops, I guess that was me abusing `upcase-dwim` :)

source/theme.lisp
aartaka: The prefix was the problem, yes. Your suggestion with `theme:background` etc. should resolve it.

Ambrevar:
A libraries sound good if you think there is no dependency on Nyxt.

Maybe name it "themer", to make the distinction with the actual themes
users may write.

source/theme.lisp
jmercouris: Agreed. Then one would also avoid all of the `%special-symbols%`.

source/theme.lisp
jmercouris: :-)

jmercouris:
I concur with Pierre's suggestions.

Another note: I think the name `theme` is fine. The user should write `my-theme` if they want a personalized theme!

Ambrevar:
Misunderstanding: the library is not a theme in itself, which is why I
believe that semantically we should not call it theme.

aartaka:
Rebased over master and extracted theming into a separate library.

I've gone with `theme` because no matter which theme user sets in `browser` configuration (see examples in testing snippets), `theme:background` and other variables will be correctly bound to the new theme.

Please test and report any style inconsistencies!

Ambrevar:
Fantastic, I'll test soon!

Ambrevar:
Just tested, works great!
Looks good to me beside the final nits I've reported.  Feel free to merge then!

libraries/theme/theme.lisp
```diff
@@ -0,0 +1,146 @@
...
+  (:export-accessor-names-p t)
+  (:export-predicate-name-p t)
+  (:accessor-name-transformer (class*:make-name-transformer name)))
+
+(defun regular-css-rule-p (rule)
+  (loop for i below (length (rest rule))
+        for elem in (rest rule)
+        when (and (evenp i) (not (keywordp elem)))
+          do (return nil)
+        finally (return t)))
```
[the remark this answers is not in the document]
aartaka: Hmmmmmmm. Yes, it fits perfectly. Should I copy it or reference `prompter` then?

aartaka:
@jmercouris, any opinion on styling with this patch? I've changed some colors here and there.

jmercouris:
I must say, some of the color changes are double plus not good! However, that shouldn't block merging this, I will tweak whatever I can :-)

libraries/theme/tests/test.lisp
```diff
@@ -0,0 +1,59 @@
...
+
+(prove:subtest "Inline function execution"
+  (prove:is  (theme:themed-css *theme*
+               (body
+                :color (concatenate 'string theme:accent " !important")
+                :background-color theme:primary))
+             "body { color: magenta !important; background-color: yellow; }
+"))
+
+(prove:subtest "Inline macro/special form invokation"
```
jmercouris, line 50: I'm afraid you've been reading too much of my text, invocation is with a "C" in english :-D

I accidentally write it with a "K" from german!

libraries/theme/theme.lisp
```diff
@@ -0,0 +1,146 @@
...
+    :type string
+    :documentation "The main non-text/interface color.
+Should preferably contrast both `background-color' and `text-color'.")
+   (secondary-color
+    "lightgray"
+    :type string
+    :documentation "The secondary interface color.
+Should contrast with the `text-color'.")
+   (tertiary-color
+    "darkgray"
```
jmercouris, line 57: A common mistake :-D!

https://stackoverflow.com/questions/3811973/why-is-darkgray-lighter-than-gray

jmercouris:
As soon as tests pass again... I will merge.

jmercouris:
The colors are not perfect, but they suffice for now.

jmercouris:
Thank you Artyom!

libraries/theme/theme.lisp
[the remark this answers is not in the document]
Ambrevar: @jmercouris Please take care of not merging too early, especially when there are unanswered points.  I suggest the following practice: when a pull request is not ours, we should always ask if it's OK to merge.  What do you think?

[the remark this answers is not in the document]
Ambrevar: @aartaka You can't reference `prompter` because that would introduce a hard dependency where it makes little sense.

For now I suggest to copy it and add a "TODO" comment referencing the other copy (on both sides), mentioning that we should extract it.

It would fit a utility library.  We also have `copy-object` which I find quite useful.  I'll ask Serapeum if they are OK to merge them.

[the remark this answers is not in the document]
Ambrevar: Reported here: https://github.com/ruricolist/serapeum/issues/99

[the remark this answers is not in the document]
aartaka: > @aartaka You can't reference `prompter` because that would introduce a hard dependency where it makes little sense.

Yes, that's what I thought too.

> For now I suggest to copy it and add a "TODO" comment referencing the other copy (on both sides), mentioning that we should extract it.

Yes, I did exactly that, except for the TODO bit :)

> It would fit a utility library. We also have `copy-object` which I find quite useful. I'll ask Serapeum if they are OK to merge them.

Cool, thank you Pierre!

libraries/theme/tests/test.lisp
aartaka: In my case it's a confusion between invoke/invocation. We're in the grey area of language use here :D

PR #1884, opened 2021-11-20
Improve spell check

guojing0:
This PR intends to improve parts of the suggestions mentioned in #251.

The user could "ask" Nyxt to know which languages it supports and spell check the sentences as he types.

I intend to improve it with selecting paragraphs, but I am not sure would the prompt buffer be a good place to display suggestions, since it may also allow user input?

PR #1898, opened 2021-11-30
Improve spell-check.lisp

guojing0:
This PR intends to deprecate my earlier PR #1884 in that this has clearer history of commits.


Ambrevar:
Thank you!

PR #1900, opened 2021-12-01
Fix prefix keymap

### LINKED ISSUE #1871: libraries/keymap: Allow rebindings non-prefix keys to be prefix keys in child keymaps
In #1858, I discovered that if you bind `C-x` in CUA keymap, you can no longer use any `C-x`-prefixed command in Emacs/VI keymap. I believe that this is not the behavior one can expect there: child keymaps, when using parent non-prefix keys as prefix ones, should have a priority. 

In other words, even it `C-x` is defined in CUA, it should never work without a subsequent key (like Emacs' `C-x C-f`, for example) if Emacs mode is on.

Ambrevar:
Should fix #1871.

@aartaka Can you confirm it works as expected for you?  Make sure to test various key bindings :)

To do:

- [x] Write a test!

### DIFF
--- a/libraries/keymap/keymap.lisp
+++ b/libraries/keymap/keymap.lisp
@@ -447,10 +447,18 @@ VISITED is used to detect cycles."
   (when keys
     (let ((hit (fset:@ (entries keymap) (first keys))))
       (when hit
-        (if (and (keymap-p hit)
-                 (rest keys))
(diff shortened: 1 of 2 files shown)


Ambrevar:
Merged with 77554a8754d0b1ef22426e85582b71c9f18ac804.

PR #1902, opened 2021-12-01
prompt-buffer: Fix resume-prompt to properly handle result / cancel.

### LINKED ISSUE #1873: Resumed prompt is not cancellable/closeable
Steps to reproduce:
- Run Nyxt, use prompt buffer command.
  - Example: do `set-url github.com RET`.
- Run `resume-prompt`.
- Choose the previously invoked `set-url` prompt.
- In the prompt opened by `resume-prompt`, try pressing `ESC`, `C-g` (if in Emacs mode) etc. It won't close the prompt.
- Try choosing any option out of the resumed prompts -- actions will run, but prompt will not close.

Nyxt commit 6283b8c88c978abac145c0d83cd31af9b2ff9116, installed via `nyxt.scm` Guix definition.

Ambrevar:
While we are at it, we use a channel in `prompt' to avoid a race condition in
which the (undisplayed) prompt buffer would return and
parent-thread-prompt-buffer would not be set, leading to an error.

Should fix #1873.  @aartaka You confirm?

### DIFF
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -170,8 +170,9 @@ To access the suggestion instead, see `prompter:selected-suggestion'."
       (prompter:selected-suggestion prompt-buffer)
     (values (when suggestion (prompter:value suggestion)) source)))
 
-(defun show-prompt-buffer (prompt-buffer &key height)
-  "Show the last active prompt-buffer, if any."
+(defun show-prompt-buffer (prompt-buffer &key height wait-p)
+  "Show the last active prompt-buffer, if any.
+With WAIT-P, block and return PROMPT-BUFFER results."
   ;; TODO: Add method that returns if there is only 1 source with no filter.
   (when prompt-buffer
     (push prompt-buffer (active-prompt-buffers (window prompt-buffer)))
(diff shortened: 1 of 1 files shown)


Ambrevar:
Let me know if you think 1e006ec2b389252b3fbf0d3bd777a099c75a8722 is a good idea.

aartaka:
> Should fix #1873. @aartaka You confirm?

Yes, it does fix it!

aartaka:
> Let me know if you think [1e006ec](https://github.com/atlas-engineer/nyxt/commit/1e006ec2b389252b3fbf0d3bd777a099c75a8722) is a good idea.

I'm indecisive. Was `wait-on-prompt-buffer` always used in pair with `show-prompt-buffer`? Can there be a use-case where we need to wait for an otherwise already show one?

Ambrevar:
When we hide a prompt buffer, we show the next one without waiting,
because the caller is already waiting.

That said, it's a bit convoluted, but I'm not sure how we can clarify
this design.

aartaka:
> When we hide a prompt buffer, we show the next one without waiting, because the caller is already waiting. That said, it's a bit convoluted, but I'm not sure how we can clarify this design.

Maybe expand on that in the docstring?

Ambrevar:
To clarify, wait-on-prompt-buffer was introduced in the first commit of
this pull request.

Ambrevar:
Merged with 37ff4ef7ec2e5d9e2e5fb899fc0b1b0d32753e98.
I found a better abstraction I think.

PR #1917, opened 2021-12-07
record-input-field: move back to source. Not a mode, just a command.

pdelfino:
Hi guys,

There is not a mode in the file `record-input-field.lisp`. It is just a command. Thus, I think it should be on source folder, but not on the mode directory. Did I miss something?

If my understanding is correct, here is the PR fixing the place of the file. Thanks!


Ambrevar:
Good catch.  @jmercouris Mistake?

jmercouris:
Not a mistake, I believe this should be part of web-mode and the extraction of modes to separate packages/clean-up for 3.0.0 is not yet complete.

pdelfino:
Ok. Thanks for making things clear, John! Thus, I am closing the PR.

Issue #1911, opened 2021-12-06
Some questions on `fuzzy-match` and relevant functions


guojing0:
I was thinking about writing functions/macros to help make write documentation and (user) manual easier, and then I found a TODO in [fuzzy.lisp], which was about reverse fuzzy matching. I started to read [prompter/filter.lisp] and [prompter/prompter-source.lisp], and had the following questions:

1. _score-suggestion-string_ function in Line 43 of filter.lisp: Currently we use Damerau-Levensthein for scoring. The comment and TODO suggests the Jaccard metric or fzf's algorithm (I found out it used some modified version of Smith-Waterman algorithm). Its use cases seem to be only in annotate.lisp and bookmark.lisp. Would the input usually be very large, if not, maybe we could use the Jaccard metric to obtain better results?

The Smith-Waterman algorithm seems to be [implemented in Common Lisp].

2. _score-threshold_ variable in Line 48 of filter.lisp: It seems to be a good threshold to filter unnecessary elements, but currently it's not used anywhere.

3. _substring-norm_ function in Line 6 of filter.lisp: After reading the code and some testing, I think it may be a good idea to remove duplicates from `substrings`, so that each element shares equal weight?

[fuzzy.lisp]: https://github.com/atlas-engineer/nyxt/blob/master/libraries/prompter/tests/fuzzy.lisp

[prompter/filter.lisp]: https://github.com/atlas-engineer/nyxt/blob/master/libraries/prompter/filter.lisp

[prompter/prompter-source.lisp]: https://github.com/atlas-engineer/nyxt/blob/master/libraries/prompter/prompter-source.lisp

[implemented in Common Lisp]: https://common-lisp.net/project/clarity/dictionary/string-alignment.html

Ambrevar:
1. The annotations and bookmarks can indeed come in large numbers, say more than 10000 elements.  In my experience, performance was an issue, which is why I reverted to Levenshtein.  More on that in my next comment.

2. Indeed, a comment says we should "learn" a good value.  To do this we would need relevant samples and tests.  How do we get such samples?  Not easy, because we may not want to share our bookmarks for privacy reasons :)  Maybe there exists some good public samples?

3. Good point!  Wanna send a patch?

Ambrevar:
About performance and the general question of input matching: I've recently played with Montezuma (https://github.com/sharplispers/montezuma) for Demeter and got excellent results at doing substring matching.

[...]

PR #1920, opened 2021-12-08
Remove duplicates in `substring-norm` function

guojing0:
This PR originally intended to add tests for `submatches` function in `filter.lisp`. However, since it's not used elsewhere in the codebase, I removed the test code file.

The only change is to remove duplicates in SUBSTRINGS of `substring-norm` function in filter.lisp. It intends to reflect the comment in https://github.com/atlas-engineer/nyxt/issues/1911.


libraries/prompter/filter.lisp
```diff
@@ -62,19 +62,20 @@ A higher score means the suggestion-string comes first."
...
-(defun submatches (suggestion source input)      ; TODO: Add tests!
+(defun submatches (suggestion source input) ; Function is not used; TODO some tests?
   "Return SUGGESTION untouched if all INPUT strings are contained in it.
 
 This is suitable as a prompter `filter'.
 It probably makes little sense to use it together with the
 `delete-inexact-matches' preprocessor."
-  (let ((terms (delete-duplicates (str:split " " input :omit-nulls t)
-                                  :test #'equal)))
+  (let ((terms (remove-duplicates (str:split-omit-nulls " " input)
```
Ambrevar, line 71: Why did you switch from `delete-duplicates` to `remove-duplicates`?   The latter copies a list, and since `str:split` creates one, this results in useless cons-ing.

libraries/prompter/filter.lisp
```diff
@@ -62,19 +62,20 @@ A higher score means the suggestion-string comes first."
   suggestion)
 
 (export-always 'submatches)
-(defun submatches (suggestion source input)      ; TODO: Add tests!
+(defun submatches (suggestion source input) ; Function is not used; TODO some tests?
```
Ambrevar, line 65: No need to add "Function is not used", this functions is meant to be used by the library users (note that it's exported).

libraries/prompter/filter.lisp
```diff
@@ -10,30 +10,30 @@ The norm is closer to 1 if
 - substrings length are closer to the length of STRING.
 
 Only substrings of SUBSTRING-LENGTH characters or more are considered."
-  ;; TODO: Remove duplicates in SUBSTRINGS?  Repeats could mean we insist more on it.
```
Ambrevar, line 13: So do you think insisting on a term is useless?  How do you answer this TODO note?

libraries/prompter/filter.lisp
guojing0: I didn't know the latter copies a list. Restored it.

libraries/prompter/filter.lisp
guojing0: `substring-norm` is only used in `score-suggestion-string` function, and I think maybe `norm-damerau-levenshtein` should have more weight?

Ambrevar: The weight between the two norms should be adjusted by the factor in from of them in `score-suggestion-string` (currently 1.0 for both).

In all cases, it would be nice to produce test data for our norms to measure relative gains / losses between two implementations.

guojing0: I am trying to produce some data for use cases of `substring-norm`.

As far as I can see, `substring-norm` is used in `score-suggestion-string`, which is then used in `fuzzy-match`, which is mainly used in `annotate.lisp` and `bookmark.lisp`. I was wondering that could you give me a concrete example of usual use cases, so that I could test it accordingly?

guojing0:
@Ambrevar After some considerations, I think it may be a good idea not to remove duplicates, so that "insisting more" could work.

jmercouris:
@Ambrevar @guojing0 

The next steps are unclear to me. These changes, if I understand them are mostly syntactical. Is there a meaningful change besides the amount of consing as a result of remove vs delete-duplicates?

Is this PR therefore void, or is there more to be done?

Ambrevar:
Seems like it is.  @guojing0 You did format the code a little and some fixes are valid.  Is there anything else you'd like to include in this patch?

guojing0:
@Ambrevar I just added some tests for `submatches` function according to your suggestion. That should be all.

libraries/prompter/filter.lisp
```diff
@@ -13,39 +13,38 @@ Only substrings of SUBSTRING-LENGTH characters or more are considered."
...
+                                              (length string))))
                                      (+ position-factor length-factor)))))
                           long-substrings))
            (length long-substrings))
         0)))
 
 (defun score-suggestion-string (input suggestion-string)
   "Return a SUGGESTION's score for INPUT.
-A higher score means the suggestion-string comes first."
+A higher score means the SUGGESTION-STRING comes first."
```
Ambrevar, line 40: This change is OK.

libraries/prompter/tests/submatches.lisp
```diff
@@ -0,0 +1,28 @@
...
+
+(in-package :prompter/tests)
+
+(prove:plan nil)
+
+(prove:subtest "Submatches Test"
+  (flet ((submatch (input list)
+           (let ((source (make-instance 'prompter:raw-source)))
+             (setf (prompter::current-input-downcase-p source)
+                   (str:downcasep input))
```
Ambrevar, line 12: Why settings this?

libraries/prompter/tests/submatches.lisp
```diff
@@ -0,0 +1,28 @@
...
+             (mapcar (lambda (suggestion)
+                       (prompter:submatches suggestion source input))
+                     (prompter::ensure-suggestions-list source list)))))
+
+    (prove:is (count nil (submatch "cat" '("cstheory" "category" "candidate")))
+              2)
```
Ambrevar, line 18: `count` is too imprecise here.  Instead, match against the exact result, for instance here `'(nil "category" nil)`.

Ambrevar:
It looks like most of the changes in filter.lisp are only a matter of taste.  Since they are debatable and add noise to the Git history, I suggest to revert them except where I commented that it's OK.  Does that make sense?

guojing0:
@Ambrevar So keep only that line and `submatches.lisp` file?

Ambrevar:
> @Ambrevar So keep only that line and `submatches.lisp` file?

I'd say so, unless you think something else is worth a merge.

guojing0:
See https://github.com/atlas-engineer/nyxt/pull/1931

Issue #1743, opened 2021-08-21
How to unbind a key?


franburstall:
**Describe the bug**

I have the GTK emacs key-theme enabled in my desktop environent and this means that `C-a` and `C-e` should move the cursor to the end or beginning of the line in textboxes.  This works fine in nyxt for `C-a` but not for `C-e` because there is a binding of `C-e '` to `edit-with-external-editor`.  This turns `C-e` into a prefix key.  A work-around is to hit `C-e` twice but this is a little irritating.

Folllowing this [discussion](https://discourse.atlas.engineer/t/how-to-unset-a-key-binding/115), I tried binding `C-e '` to `nothing` in the buffer class's `override-map` slot but this still leaves `C-e` as a prefix key.  

So how can I unbind `C-e '` so that `C-e` is passed to the browser unmolested by nyxt?

Failing that, can I respectfully suggest that `C-e '` is a rather unfortunate choice for an emacs keyscheme and something else be chosen instead?

**Information**
- OS name+version: Debian-10
- Graphics card and driver:
- Desktop environment / Window manager name+version: EXWM
- How you installed Nyxt (Guix pack, package manager, build from source): source
- Information from `show-system-information`:
Nyxt version: 2.1.1-526-g256a1b07
Renderer version: GI-GTK
Operating system kernel: Linux 4.19.0-17-amd64
Lisp implementation: SBCL 2.1.1 (Dynamic space size: 1073741824)
Features: (:SLYNK :WEBKIT2 :WEBKIT2-2.32 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
[...]

PR #1922, opened 2021-12-08
Forward to renderer

Ambrevar:
Should answer #1743.  What do you people think?


jmercouris:
A reasonable addition! I like it!

PR #1924, opened 2021-12-08
buffer: Compute buffer-source keywords asynchronously.

### LINKED ISSUE #1709: Switch Buffer suffers from lag
**Describe the bug**

It seems upon first invoking C-x b for a while, or if nyxt is in a more errorr prone state that the user can wait over 10 seconds for the command to launch.

I have not seen this behavior before pulling #1692 so I'm guessing the algorithm on what's it's doing is slower? But the odd thing is is that the bug where it focuses the other windows buffer first (and then swaps to it before unfocusing the prompt buffer) still stands.

I believe calling `switch-buffer` for the first time should lag, also opening more links and waiting should also trigger it? I'm unsure of when exactly it triggers

**Information**
- OS name+version:
  - Arch Linux 
- Graphics card and driver:
  - GPU: NVIDIA GeForce GTX 980, NVIDIA GeForce GTX 980 
- Desktop environment / Window manager name+version:
   - Xmonad And StumpWM 
- How you installed Nyxt (Guix pack, package manager, build from source):
  - Master `make all` 

System information
```lisp
Nyxt version: 2.1.1-446-gfd0c9bc2
Renderer version: GI-GTK
Operating system kernel: Linux 5.13.4-arch2-1
Lisp implementation: SBCL 2.1.1 (Dynamic space size: 1073741824)
Features: (:SLYNK :WEBKIT2 :WEBKIT2-2.32 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :GLIB-2-58 :GLIB :NYXT-2 :FSET-EXT-STRINGS :CUSTOM-HASH-TABLE-NATIVE :SWANK
 :PLUMP-UTF-32 :GLOBAL-VARS :DECLARE-TYPES :PARENSCRIPT :NAMED-READTABLES
 :21BIT-CHARS :CHUNGA :FLEXI-STREAMS :CLOSER-MOP :CL-PPCRE-UNICODE :CL-UNICODE
 :CL-PPCRE :BORDEAUX-THREADS :SPLIT-SEQUENCE CHIPZ-SYSTEM:GRAY-STREAMS
 :FAST-IO-SV :FAST-IO :SBCL-USES-SB-ROTATE-BYTE :ASDF-SYSTEM-CONNECTIONS
 :CL-JSON-CLOS :CL-JSON :THREAD-SUPPORT CFFI-FEATURES:FLAT-NAMESPACE
 CFFI-FEATURES:X86-64 CFFI-FEATURES:UNIX :CFFI CFFI-SYS::FLAT-NAMESPACE
 ALEXANDRIA::SEQUENCE-EMPTYP :QUICKLISP :ASDF3.3 :ASDF3.2 :ASDF3.1 :ASDF3

ASDF version: 3.3.1
Critical dependencies: (/home/katya/Documents/Workspace/Lisp/Projects/nyxt/_build/submodules/cl-cffi-gtk/gtk/cl-cffi-gtk.asd
 /home/katya/Documents/Workspace/Lisp/Projects/nyxt/_build/quicklisp-client/dists/quicklisp/software/cl-gobject-introspection-20210124-git/cl-gobject-introspection.asd
 /home/katya/Documents/Workspace/Lisp/Projects/nyxt/_build/submodules/cl-webkit/webkit2/cl-webkit2.asd)

Quicklisp dist version: 2021-06-30
Quicklisp client version: 2020-01-04
Local project directories: (/home/katya/Documents/Workspace/Lisp/Projects/nyxt/_build/submodules/
 /home/katya/Documents/Workspace/Lisp/Projects/nyxt/_build/quicklisp-client/local-projects/)
Critical dependencies(#<SYSTEM cl-cffi-gtk / cl-cffi-gtk-20201220-git / quicklisp 2021-06-30>
 #<SYSTEM cl-gobject-introspection / cl-gobject-introspection-20210124-git / quicklisp 2021-06-30>
 #<SYSTEM cl-webkit2 / cl-webkit-20210630-git / quicklisp 2021-06-30>)
```

Ambrevar:
This should remove the lag the first time the command is invoked.
This lag can be particularly long with many buffers with huge content.

Should fix #1709.

To do:

- Update the suggestions asynchronously.  The current implementation is already supposed to do it, but there seems to be a bug.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -663,10 +663,11 @@ Delete it with `ffi-buffer-delete'."
        (let ((contents (serapeum:string-join
                         (map 'list (lambda (e) (plump:text e))
                              (clss:select "p" (document-model buffer))) " ")))
-         (setf (keywords-document-model buffer)
-               (document-model buffer)
-               (slot-value buffer 'keywords)
-               (analysis:extract-keywords contents)))
+
+         (setf (keywords-document-model buffer) (document-model buffer))
+         (let ((keywords (analysis:extract-keywords contents)))
+           (unless (equal keywords '(("" . 1))) ; Same as no keywords, no need to store this so we can still check for NIL.
+             (setf  (slot-value buffer 'keywords) keywords))))
        (slot-value buffer 'keywords))))
 
 (-> proxy-adress (buffer &key (:downloads-only boolean)) *)
@@ -752,9 +753,18 @@ BUFFER's modes."
     ("Title" ,(title buffer))))
 
 (defmethod prompter:object-attributes ((buffer web-buffer))
-  `(("URL" ,(render-url (url buffer)))
-    ("Title" ,(title buffer))
-    ("Keywords" ,(format nil "~:{~a~^ ~}" (keywords buffer)))))
+  (let ((attributes `(("URL" ,(render-url (url buffer)))
+                      ("Title" ,(title buffer))
+                      ;; WARNING: `keywords' is slow, use direct access instead.
+                      ("Keywords" ,(format nil "~:{~a~^ ~}"
+                                           ;; (keywords buffer)
+                                           (slot-value buffer 'keywords))))))
+    (unless (slot-value buffer 'keywords)
+      (run-thread "buffer keywords"
+        (setf (alex:assoc-value attributes "Keywords" :test 'equal)
+              (list (format nil "~:{~a~^ ~}"
+                            (keywords buffer))))))
+    attributes))
 
 (-> make-buffer
     (&key (:title string)


source/buffer.lisp
```diff
@@ -752,9 +753,18 @@ BUFFER's modes."
...
 (defmethod prompter:object-attributes ((buffer web-buffer))
-  `(("URL" ,(render-url (url buffer)))
-    ("Title" ,(title buffer))
-    ("Keywords" ,(format nil "~:{~a~^ ~}" (keywords buffer)))))
+  (let ((attributes `(("URL" ,(render-url (url buffer)))
+                      ("Title" ,(title buffer))
+                      ;; WARNING: `keywords' is slow, use direct access instead.
+                      ("Keywords" ,(format nil "~:{~a~^ ~}"
+                                           ;; (keywords buffer)
+                                           (slot-value buffer 'keywords))))))
```
jmercouris, line 761: The whole point of (keywords buffer) was to have a lazily updatable slot value. If we are just accessing the slot-value directly... then we don't necessarily get that, no?

jmercouris:
I'm not certain this patch makes sense. I think the changes if ANY should be made in:

```
(defmethod keywords ((buffer web-buffer)) ...)
```

source/buffer.lisp
Ambrevar: The problem here is with object-attributes: they are fetched the first time and never updated afterwards.  This is what this snippet does: it forces the asynchronous update.

Ambrevar:
You may have misunderstood this patch: object-attributes are computed once, so this patch runs a thread to force-update the attributes when the keywords are ready.

This is not related to how `keywords` works, it could very well be just any expansive function.

Ambrevar:
To clarify, the benefit from this patch is to fix a performance issue (and it works! :p).

jmercouris:
Perhaps I have indeed misunderstood. Let me take a second look.

jmercouris:
I see. I understand now. I was confused by the restructuring of the keywords method. Sorry about that!

That said, I don't particularly like this fix. There are a couple of issues I predict:

1. You will always get stale data after the first iteration of `object-attributes`.

Consider this scenario:

`(slot-value buffer 'keywords)` has a value, but it is a value calculated from a PREVIOUS page. If we run through the `prompter:object-attributes` for this, you'll see that it will happily fetch and display stale keywords! Notice the comparsion is `(unless slot-value buffer 'keywords)` rather than the more expensive, and NECESSARY comparison as is done in `defmethod keywords`!

Reference source:
```
(defmethod keywords ((buffer web-buffer))
  "Calculate the keywords for a given buffer."
  (ignore-errors
   (if (not (eq (document-model buffer)
                (keywords-document-model buffer)))
       (let ((contents (serapeum:string-join
                        (map 'list (lambda (e) (plump:text e))
                             (clss:select "p" (document-model buffer))) " ")))

         (setf (keywords-document-model buffer) (document-model buffer))
         (let ((keywords (analysis:extract-keywords contents)))
           (unless (equal keywords '(("" . 1))) ; Same as no keywords, no need to store this so we can still check for NIL.
             (setf  (slot-value buffer 'keywords) keywords))))
       (slot-value buffer 'keywords))))
```

Relevant lines:
```
(if (not (eq (document-model buffer)
                (keywords-document-model buffer)))
```

These lines ensure that the keywords are updated to reflect the latest document model! `keywords-document-model` is a reference of the state of the document-model when the keywords were calculated.

Reference source:
```
(defmethod prompter:object-attributes ((buffer web-buffer))
  (let ((attributes `(("URL" ,(render-url (url buffer)))
                      ("Title" ,(title buffer))
                      ;; WARNING: `keywords' is slow, use direct access instead.
                      ("Keywords" ,(format nil "~:{~a~^ ~}"
                                           ;; (keywords buffer)
                                           (slot-value buffer 'keywords))))))
    (unless (slot-value buffer 'keywords)
      (run-thread "buffer keywords"
        (setf (alex:assoc-value attributes "Keywords" :test 'equal)
              (list (format nil "~:{~a~^ ~}"
                            (keywords buffer))))))
    attributes))
```

2. I could have sworn we were utilizing pmapcar for gathering the object-attributes, is my memory incorrect?

If my memory is correct, then all of the object attributes for all of the buffers should simultaneously be calculating the keywords. If it turns out that `(if (not (eq (document-model buffer) (keywords-document-model buffer)))` is too expensive, perhaps we can do some sort of cheaper comparison. Perhaps we can have a cache that is invalidated by any load events?

Ambrevar:
1. You are right about the staleness.  I believe that removing the `unless` (and thus always computing the keywords) should fix it, right?

2. We are using `pmapcar`, it's in `prompter:ensure-suggestions-list`.  Still too slow! :p 

It's not just the comparison that is slow (in fact, it's pretty fast), it's the keyword computation done afterwards.  Which is why the slow-down is only experienced with non-empty buffers that don't have their keywords computer yet.

To reproduce this:
- load an expansive page with lots of text
- Restart Nyxt
- Load the history
- Call `switch-buffer`.

In the end, the `keywords` defmethod is completely fine in my understanding, the performance fix here needs only be very local because of how object-attributes work.

There is a more general fix though: allow for asynchronous / non-blocking-and-lazy object-attributes.  This would require an update to the prompt-buffer library.

Thoughts?

Ambrevar:
Superseded by #1930.

PR #1925, opened 2021-12-08
documents/README: insert alert about bringing instability to stable releases.

pdelfino:
Hi guys,

I myself faced this problem recently. Thus, I thought it would be interesting to add this snippet to our developer readme file.  Curious about your thoughts.

### DIFF
--- a/documents/README.org
+++ b/documents/README.org
@@ -40,6 +40,20 @@
   - [[#irc-channel][IRC Channel]]
 :END:
 
+* Attention
+
+In case you already have a stable release version of Nyxt running in
+your environment, such as those provided on the [[https://nyxt.atlas.engineer/download][Download section]], be
+aware that having Nyxt running from source might cause problems on
+previous stable releases. In other words, a *stable* release might
+become /unstable/ after running cutting-edge Nyxt from source.
+
+For instance, along with the 2-series, changes were made on how the
+history file was being written by Nyxt. Thus, navigating with a newer
+version of Nyxt would write data in a different way, and, as a
+consequence, the file would not be readable by past stable releases,
+generating a hang bug. Other problems like this one might occur.
+
 * Developer installation with Guix
 
 First, clone the Nyxt repository.  It is most convenient to clone the


guojing0:
If the ways that old and new versions write to history files are not that different, maybe we could find a way to resolve this backward incompatibility, so that the user would not lose previous history records?

Ambrevar:
@guojing0 We've already added a backward incompatibility support for 2.2.3, but master does not have it since it's meant to be the future 3.0 which is backward incompatible in nature.

@pdelfino Did you notice a hang?  If so, it's another bug and we should fix it.

Finally, I'll implement a history backup rotation to avoid overwriting a history file that could not be read.
This way, I think we don't need to warn the user at all.  In other words, this issue of version switching should not exist.

guojing0:
> Finally, I'll implement a history backup rotation to avoid overwriting a history file that could not be read.

@Ambrevar If the implementation is not too intricate and next release not too close, maybe I could help implement this?

Ambrevar:
Sure thing, I'm targetting next week for a release, do you think that
would fit for you?

Ambrevar:
And thanks for offering help, it's more than welcome!

guojing0:
I will try to complete it before next Monday. If unsuccessful, I then will tell you. Is this OK?

Ambrevar:
Sure, thanks!

What I had in mind for history rotation is to rename the unsucessfully-loaded
file to $BASENAME-backup-$N.lisp in the same directory, where $BASENAME
is the basename of the file, and $N a number starting from 1.

We've got to detect existing backup files and increment the number
starting from the latest.
You can use the uiop:directory* function to collect the files matching a
string with a wildcard.
Sort the result, parse the last number and you should be good to go!

This is just a suggestion, counter-suggestions are welcome!

pdelfino:
OK. After the discussion above, I am closing the PR!

> @pdelfino Did you notice a hang? If so, it's another bug and we should fix it.

If my memory serves me well, yes - I recall a hang happening. Unfortunately, I cannot reproduce this anymore even in my machine.

For some unknown reason for me, I can only run Nyxt from source now. The downloaded stable release version won't even load anymore. This happened after I was trying to make some tests on the history file. Now, even after deleting the history content or history file, I can't start anymore Nyxt' stable release.

Ambrevar:
> For some unknown reason for me, I can only run Nyxt from source now. The downloaded stable release version won't even load anymore. This happened after I was trying to make some tests on
> the history file. Now, even after deleting the history content or
> history file, I can't start anymore Nyxt' stable release.

Please report this with a backtrace, start with `nyxt -IC`, etc.

pdelfino:
> For some unknown reason for me, I can only run Nyxt from source now. The downloaded stable release version won't even load anymore. This happened after I was trying to make some tests on the history file. Now, even after deleting the history content or history file, I can't start anymore Nyxt' stable release.
> Please report this with a backtrace, start with `nyxt -IC`, etc.

Using your suggestion, I found out it was a silly problem. With `nyxt -C` it **does not** work. With `nyxt -I` it **does** work. With `nyxt -IC` it **does** work. Thus... the problem is this part in my init file:

``` lisp
(define-configuration buffer
  ((override-map (let ((map (make-keymap "override-map")))
                   (define-key map
                      ;; "C-space" 'nothing
                      "C-x C-b" 'switch-buffer
                      "M-x" 'execute-command
                      "C-n" 'nyxt/web-mode:scroll-down
                      "C-p" 'nyxt/web-mode:scroll-up
                      "C-f" 'nyxt/web-mode:scroll-right
                      "C-b" 'nyxt/web-mode:scroll-left
                      "C-g" 'query-selection-in-search-engine
                      "M-s->" 'nyxt/web-mode:scroll-to-bottom
                      "M-s-<" 'nyxt/web-mode:scroll-to-top
                      "C-s" 'nyxt/web-mode:search-buffer
                      "M-w" 'nyxt/web-mode:copy
                      "C-y" 'nyxt/web-mode:paste
                      "C-Y" 'nyxt/web-mode:paste-from-clipboard-ring)))))

```

The line comment out is the one problematic. The stable release 2.0.0.0 I have won't deal with `'nothing` command.

Ambrevar:
Can you report this in a separate issue?

pdelfino:
> Can you report this in a separate issue?

[Done](https://github.com/atlas-engineer/nyxt/issues/1935) :)

guojing0:
@Ambrevar I had a meeting with John and he thinks that history backup rotation may be slightly difficult for me.

In the weekend I poked around `history.lisp` and `history-tree.lisp` and could not find a good way to do that.

Ambrevar:
No problem, I'll handle it then.

PR #1932, opened 2021-12-12
Build on FreeBSD is broken again.

shamazmazum:
FreeBSD and maybe some other BSD variants do not have GCC installed by default, the compiler name must be simply `cc` (which is a symlink to clang). `cc` is OK on most linuxes too, but, unfortunately, guix does not have `cc` symlink.

I do not know guix and have no idea how to fix it properly, but there is a dirty solution if you cannot do something better.

BTW, `libnyxt` seems to be unused. What is it for? (I suggest it's for upcoming browser extensions, but I think I can remove it for now. Is that correct?)

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -611,6 +611,7 @@ See `asdf::*immutable-systems*'."
 
 (defsystem "nyxt/web-extensions"
   :pathname "libraries/web-extensions/"
+  :depends-on (:cffi-toolchain)
   :components ((:static-file "alarms.c")
                (:static-file "bookmarks.c")
                (:static-file "browser.c")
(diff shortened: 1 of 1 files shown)


aartaka:
> FreeBSD and maybe some other BSD variants do not have GCC installed by default, the compiler name must be simply `cc` (which is a symlink to clang). `cc` is OK on most linuxes too, but, unfortunately, guix does not have `cc` symlink.

But wait, you can set `CC` environment variable to `cc`, can you not? Given that the build relies on `CC` and falls back to `gcc` only after it's not found, setting `CC` should work just fine.

> I do not know guix and have no idea how to fix it properly, but there is a dirty solution if you cannot do something better.

I guess the Guix way would be to either package `gcc` together with Nyxt or to set `CC` to a sensible compiler at the build time.

> BTW, `libnyxt` seems to be unused. What is it for? (I suggest it's for upcoming browser extensions, but I think I can remove it for now. Is that correct?)

It's not used, indeed. And you're right about the `libnyxt` -- it is the WebExtensions support library.

Yes, you can remove it for now :)

shamazmazum:
> But wait, you can set CC environment variable to cc, can you not?

Yes, I can, but should I? I just want to build nyxt as any other common lisp system: `(asdf:make :nyxt/gtk-application)` (or `(progn (ql:quickload :nyxt/gtk-application) (asdf:make :nyxt/gtk-application))`).

Here I propose a better solution: let cffi-toolchain find a proper C compiler, it's good at it. This solution must satisfy all users.

aartaka:
Wow, today I learned of `cffi-toolchain`, thanks! 

And thanks for the patch :) Merging it!

PR #1933, opened 2021-12-13
New mode for disabling search auto-completion

shamazmazum:
Here is `no-search-completion-mode` which disables suggestions from a search engine. This mode sets already existing `search-always-auto-complete-p` slot in instances of `user-buffer` class.

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -154,6 +154,7 @@ A naive benchmark on a 16 Mpbs bandwidth gives us
                (:file "mode/no-image")
                (:file "mode/no-sound")
                (:file "mode/no-script")
+               (:file "mode/no-search-completion")
                (:file "mode/no-webgl")
                (:file "mode/reduce-bandwidth")
                (:file "mode/download")
--- a/source/mode/no-search-completion.lisp
+++ b/source/mode/no-search-completion.lisp
@@ -1,0 +1,16 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package :nyxt/no-search-completion-mode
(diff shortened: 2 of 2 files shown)


Ambrevar:
@aartaka This makes me realize that the slot should be a window option, shouldn't it?

@shamazmazum Do you have a use case where you want completion when a specific buffer is focused, while having no completion on some other buffer?

shamazmazum:
> Do you have a use case where you want completion when a specific buffer is focused, while having no completion on some other buffer?

No, I just accidentely found that there is `search-always-autocomplete-p` slot in `buffer` already and thought that it's there on purpose.

shamazmazum:
If you wish, I can add an appropriate slot for `window` and rewrite this stuff as a command.

Btw, what is the purpose of `search-auto-complete-p`? It's not used anywhere.

aartaka:
@Ambrevar, yes, the slot should be a supra-buffer one!

@shamazmazum, yes, `search-auto-complete-p` is indeed unused, while it should be. Let me do a quick fix on master.

Regarding this patch: why doing a mode if you can simply set the slot to `nil` in the configuration?

aartaka:
The `search-auto-complete-p` check is there: 597d73be82501e278585686ead230564126bf62c.

shamazmazum:
> Regarding this patch: why doing a mode if you can simply set the slot to nil in the configuration?

Oh, I did not know it's possible. I should close this request then :)

aartaka:
> > Regarding this patch: why doing a mode if you can simply set the slot to nil in the configuration?
> 
> Oh, I did not know it's possible. I should close this request then :)

The snippet that does it is 
``` lisp 
(define-configuration buffer 
 ((search-always-auto-complete-p nil)))
``` 
in your init file.

shamazmazum:
Thanks!

PR #1934, opened 2021-12-13
Add `pkg-config` to macOS dependencies in developer README

guojing0:
After web-extension was added to `master` branch, compiling from source requires `pkg-config`.

The program `pkg-config` is not included in either macOS system or Apple's command line tools, so one needs to install it, like other dependencies, from third-party tools.

### DIFF
--- a/documents/README.org
+++ b/documents/README.org
@@ -134,6 +134,7 @@ run the following:
 - libfixposix
 - xclip (for clipboard support)
 - enchant (for spellchecking)
+- pkg-config (for web-extensions)
 
 - Debian-based distributions:
   #+begin_src sh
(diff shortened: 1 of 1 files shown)


aartaka:
It's not a MacOS-specific dependency, actually :)

It would be nice if you put it into the GNU/Linux and BSD dependencies.

guojing0:
@aartaka You are right. It may come with some OS, say Debian. That's why I didn't realize that it's also needed to be downloaded on Linux or BSDs.

guojing0:
Thank you, Artyom. I thought about also writing package installation instructions, but I was not sure which keywords to write, so I decided to delay that to the morning if this had not been merged. :)

Issue #1923, opened 2021-12-08
Feature request related to documentation: introduce a button in the manual to create an init-file, in case there is not one available


pdelfino:
The manual does not indicate that the `init.lisp` is **not** dynamically created.  It needs to be created by the user.

Even if the file is not created, the manual will show something like: "Nyxt configuration can be persisted in the user file `/home/pedro/.config/nyxt/init.lisp`." This leaves the impression the file was already created just like bookmarks or history which  can be especially confusing for new users. 

Talking to @jmercouris, an idea emerged: inserting in the manual a function that indicates if the file was already created or not. In case it was not created, show a button like `Create init file`. There is already a comment "(create the parent folders if necessary)" but the function to identify and the button seem better, especially for newcomers.

I will send a PR about it soon.

guojing0:
Hi, after seeing that you haven't sent a PR yet, so I was working on this.

During the process, I realize something: Instead of letting the user press the button, would it be OK to just let Nyxt automatically create `init.lisp`, if said file had not existed yet.

@jmercouris What do you think of this option?

jmercouris:
I think we should leave it in control of the user to create this file. I don't think the user would appreciate new files in `~/.config/` without explicit consent (this is of different than `~/.local/share/` (a directory in which we do automatically create files)).

jmercouris:
Jing pointed out that `auto-conf.lisp` is automatically created AND located in `~/.config`. This is of course a problem if the above comment makes nay sense. It is of my opinion therefore that we either prompt the user before the creation of the `auto-config.lisp` file OR move its location to `~/.local/share`.

pdelfino:
I am closing this since @guojing0's work was squashed and merged by Pierre as https://github.com/atlas-engineer/nyxt/commit/1b57e8c71f8351f3e75041d562a04b246479cd3f. 

Thanks!

PR #1939, opened 2021-12-16
Insert "Create init file" button in the manual

guojing0:
This PR intends to resolve the issue #1923.

The real change happens **from line 34 to line 50.** I use Lispy for strucutural editing, so when I press "Return" key to enter a new line, it seems to indent everywhere else in the source file. I could not find a way to pass around that.

My idea is to dynamically create the button based on whether `init.lisp` is already created: If so, then do nothing. Otherwise, create the real button.

### DIFF
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -41,7 +41,8 @@ below to create said file, if it's not created yet.")
                    :href (lisp-url `(echo "Init file exists"))
                    "Init file exists")
                (:a :class "button"
-                   :href (lisp-url `(with-open-file (_ ,init-file-path
+                   :href (lisp-url `(ensure-directories-exist ,init-file-path)
+                                   `(with-open-file (_ ,init-file-path
                                                        :direction :output
                                                        :if-exists nil
                                                        :if-does-not-exist :create)
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -36,13 +36,13 @@ similar programming language.")
          ", and the parent folders if necessary. You can also press the button
 below to create said file, if it's not created yet.")
      (let ((init-file-path (expand-path *init-file-path*)))
+       (ensure-directories-exist init-file-path)
        (:p (if (uiop:file-exists-p init-file-path)
                (:a :class "button"
                    :href (lisp-url `(echo "Init file exists"))
                    "Init file exists")
                (:a :class "button"
-                   :href (lisp-url `(ensure-directories-exist ,init-file-path)
-                                   `(with-open-file (_ ,init-file-path
+                   :href (lisp-url `(with-open-file (_ ,init-file-path
                                                        :direction :output
                                                        :if-exists nil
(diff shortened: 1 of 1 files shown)


Ambrevar:
> so when I press "Return" key to enter a new line, it seems to indent everywhere else in the source file. I could not find a way to pass around that.

I also use Lispy.  To force-insert a newline without indenting, I used `C-q C-j`, probably followed by `tab`.  I don't know if there is a better way, but that's good enough for me.

source/manual.lisp
```diff
@@ -11,85 +11,98 @@
...
-    (:p "Example:")
-    (:pre (:code "
+     (:p "The user needs to manually create the Nyxt configuration file "
+         (:code (expand-path *init-file-path*))
+         ", and the parent folders if necessary. You can also press the button
+below to create said file, if it's not created yet.")
+     (let ((init-file-path (expand-path *init-file-path*)))
+       (:p (if (uiop:file-exists-p init-file-path)
+               (:a :class "button"
+                   :href (lisp-url `(echo "Init file exists"))
```
Ambrevar, line 41: Maybe "Init file already exists.".

guojing0: Done

Ambrevar:
Can you fix this patch to not alter the indentation of the rest of the file?

guojing0:
Done.

jmercouris:
So close! Unfortunately it does not work if the directory `~/.config/nyxt` does not exist! :-O

jmercouris:
Or any other folders in the path for that matter!

guojing0:
I believe this issue is fixed now.

guojing0:
I think said behavior may depend on specific implementation... When I test `with-open-file` with CCL on Mac, it would automatically generate the directories if they are not created.

source/manual.lisp
```diff
@@ -31,9 +31,24 @@ existing object instances.")
...
+                   "Init file exists")
+               (:a :class "button"
+                   :href (lisp-url `(ensure-directories-exist ,init-file-path)
```
aartaka, line 44: Why not just have `ensure-directories-exits` as a separate form and generate HTML implying it succeeded? I.e.,
``` lisp
(let ((init-file-path (expand-path *init-file-path*)))
  (ensure-directories-exist)
  (:p ...))
```

Or, in case it confuses Spinneret, 
``` lisp
(let* ((init-file-path (expand-path *init-file-path*))
       (ensure (ensure-directories-exist init-file-path)))
  (:p ...))
```

In either case, you save some (duplicated!) lines and reduce branching.

guojing0: Done. In current version, I think we can still give the user an option whether or not to create the init file.

Ambrevar:
Squashed and merged as 1b57e8c71f8351f3e75041d562a04b246479cd3f.
Thanks a lot!

Issue #1557, opened 2021-06-25
A command to duplicate buffer OR make related buffer?


aartaka:
Firefox has such a thing as "Duplicate Tab" action, though I'm not sure how it's implemented. We can make our `make-duplicate-buffer` to copy a current-buffer (I've had several moments I needed this). 

What can be even better is to implement a `make-related-buffer` command that'd rely on [renderer-level resource sharing](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#webkit-web-view-new-with-related-view) to lower the resource consumption. 

Tighter buffer interrelations can come in handy in history too.

How's that looking?

Ambrevar:
Good idea.

For the GHT, I suppose that a duplicate buffer would share the same
parent.

About the WebKit related view, I think we ought to abstract this well
enough and ensure it's an impmlementation detail.  Not all renderers
might be able to do this.

Ambrevar:
In particular, it could be nice to duplicate the modes.

A related feature would be to "open link in duplicate buffer": only too often I browse a buffer with special modes, but when I want to open a link in a new buffer I need to set the modes again.

aartaka:
That I can relate to, yes!

shaunsingh:
I believe this can be closed now `duplicate-buffer` and `duplicate-buffer-with-current-modes`

jmercouris:
You're absolutely correct, good catch :-)

PR #1941, opened 2021-12-17
Commands to duplicate current buffer

guojing0:
This PR intends to help solve #1557.

The two functions are pretty similar, the only difference is line 904:
`:modes (mapcar #'mode-name (modes buffer))`

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -896,14 +896,6 @@ See `make-buffer' for a description of the arguments."
   (declare (ignorable title modes url))
   (apply #'make-buffer (append (list :buffer-class 'user-background-buffer :no-history-p t) args)))
 
-(define-command duplicate-modes (&key (url (quri:uri "")) parent-buffer)
-  "Duplicate current modes in a new buffer."
-  (let ((buffer (make-buffer :url url
-                             :modes (mapcar #'mode-name (modes (current-buffer)))
-                             :parent-buffer parent-buffer)))
-    (set-current-buffer buffer)
-    buffer))
-
 (define-command duplicate-buffer-with-current-modes (&key (modes nil) parent-buffer)
   "Duplicate current buffer in a new buffer with current modes as well."
   (let* ((curr-buffer (current-buffer))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -898,10 +898,11 @@ See `make-buffer' for a description of the arguments."
 
 (define-command duplicate-buffer-with-current-modes (&key parent-buffer)
   "Duplicate current buffer in a new buffer with current modes as well."
-  (let ((buffer (current-buffer)))
+  (let* ((buffer (current-buffer))
+         (modes (modes buffer)))
     (make-buffer :title (title buffer)
                  :url (url buffer)
-                 :modes (mapcar #'mode-name (modes buffer))
+                 :modes (mapcar #'mode-name modes)
                  :parent-buffer parent-buffer)
     (set-current-buffer buffer)
     buffer))
@@ -912,6 +913,14 @@ See `make-buffer' for a description of the arguments."
     (make-buffer :title (title buffer)
                  :url (url buffer)
                  :parent-buffer parent-buffer)
+    (set-current-buffer buffer)
+    buffer))
+
+(define-command duplicate-modes (&key (url (quri:uri "")) parent-buffer)
+  "Duplicate current modes in a new buffer."
+  (let ((buffer (make-buffer :url url
+                             :modes (mapcar #'mode-name (current-buffer))
+                             :parent-buffer parent-buffer)))
     (set-current-buffer buffer)
     buffer))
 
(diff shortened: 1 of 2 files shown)


Ambrevar:
Looks good to me!

Could you also add a function to follow a hint with the current modes activated, as mentioned in #1557 ?

guojing0:
Do you mean some kind of prompt like the one in `delete-all-buffers`?

aartaka:
What I miss in this patch is use of WebKitGTK-provided [webkit_web_view_new_with_related_view](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#webkit-web-view-new-with-related-view), as that would allow sharing memory/caches/whatever between the related buffers. That's also the reason for the similarity of buffer duplication and creating related buffer in this patch -- related buffers are WebKitGTK-inspired concept, and are quite vague unless we're dealing with WebKitGTK.

Ambrevar:
> Do you mean some kind of prompt like the one in `delete-all-buffers`?

No, I meant something like follow-hint-new-buffer.

guojing0:
@Ambrevar Done. Please see `follow-hint-with-current-modes-new-buffer` in `element-hint.lisp` for this functionality.

@aartaka I have not had much experience with WebKitGTK, so I was thinking that we provide the commands in this PR to users first. Then I will improve them upon `webkit_web_view_new_with_related_view` in the future.

aartaka:
> @aartaka I have not had much experience with WebKitGTK, so I was thinking that we provide the commands in this PR to users first. Then I will improve them upon `webkit_web_view_new_with_related_view` in the future.

Yes, sounds sane. If you'd need help with WebKitGTK things, feel free to ping me!

source/buffer.lisp
```diff
@@ -896,6 +896,33 @@ See `make-buffer' for a description of the arguments."
...
+
+(define-command duplicate-buffer-with-current-modes (&key parent-buffer)
+  "Duplicate current buffer in a new buffer with current modes as well."
+  (let* ((buffer (current-buffer)))
+    (make-buffer :title (title buffer)
+                 :url (url buffer)
+                 :modes (mapcar #'mode-name (modes buffer))
+                 :parent-buffer parent-buffer)
+    (set-current-buffer buffer)
+    buffer))
```
Ambrevar, line 915: I'm confused, what's the difference between `duplicate-modes` and `duplicate-buffer-with-current-modes`?

Other than that, the rest is perfect!

guojing0: `duplicate-modes` takes `url` as an argument, whereas `duplicate-buffer-with-current-modes` duplicates current buffer.

Ambrevar: Oh I see.  To avoid code duplication, you could have one command simply call the other one with the right parameters.  What do you think?

source/buffer.lisp
```diff
@@ -896,6 +896,33 @@ See `make-buffer' for a description of the arguments."
...
+
+(define-command duplicate-buffer (&key parent-buffer)
+  "Duplicate current buffer in a new buffer."
+  (let ((buffer (current-buffer)))
+    (make-buffer :title (title buffer)
+                 :url (url buffer)
+                 :parent-buffer parent-buffer)
+    (set-current-buffer buffer)
```
Ambrevar, line 923: Why call `set-current-buffer` on the already current buffer?

source/buffer.lisp
guojing0: I actually thought about avoiding code duplication, but could not find a good way to resolve this, since setting `modes` to `'(web-mode base-mode)` seems like "hard coding".

Pleas see the new code...

source/buffer.lisp
Ambrevar: @guojing0 You might have missed this question :)

Ambrevar: Oh, I see, you've updated the code now...

source/buffer.lisp
Ambrevar: Sorry,  I am still a bit confused about the use cases.  The name `duplicate-modes` is  a bit confusing I think.  The difference seems to be mostly that it opens a buffer with a blank URL by default.

Why not have a single command with both `buffer` and `url` as parameters? Then depending on which is specified, use the current buffer URL or not.

guojing0: 1. Yes, you are right: The only difference is opening a new buffer with a blank URL, by default.
2. Do you mean something like `enable-mode` command: Prompting which buffer and which mode(s) to enable? Or we could just remove `duplicate-modes`.

Ambrevar: I think removing `duplicate-modes` would be enough for now.

With @jmercouris we were discussing of generalizing the concept of "prompting for buffers", because many commands have a version for the current buffer + a version that prompts for buffer.

So I suggest leave this out for now, we will think of something more general.

About which modes to enable, well, if you get to choose then it's not really a duplicate :)  So I think it's also fine to leave this out.

guojing0: @Ambrevar Done.

Ambrevar:
Can you fix the compilation warning (parent-buffer unused) ?

guojing0:
See https://github.com/atlas-engineer/nyxt/pull/1989

PR #1951, opened 2021-12-20
Improve fuzzy match, remove `mk-string-metrics' dependency

BlueFlo0d:


Ambrevar:
Looks good, thanks a lot for this precious contribution!

Could you provide a bit more background, like where the algorithm comes from, maybe a link to a reference implementation?

Also we need some kind of measure here to justify the replacement.  Could you provide some test data that shows how this flex match provides "humanly more relevant" results?

BlueFlo0d:
Thanks for the comments!

> Could you provide a bit more background, like where the algorithm comes from, maybe a link to a reference implementation?
I invented it myself ;) or at least developed it independently.

> Also we need some kind of measure here to justify the replacement. Could you provide some test data that shows how this flex match provides "humanly more relevant" results?
This is indeed very good suggestion and lead me to improve some parameters.
Here's some result:

```
| Input      | Flex                    | Old                |
|------------+-------------------------+--------------------|
| sw         | switch-buffer           | switch-buffer-last |
| swbuf      | switch-buffer           | make-buffer        |
| swbufd     | switch-buffer-domain    | make-buffer        |
| de         | describe-all            | make-buffer        |
| def        | base-mode (*)           | base-mode          |
| defu       | describe-function       | load-system        |
| lr         | lisp-repl               | lisp-repl          |
| lb         | toggle-toolbars (*)     | toggle-toolbars    |
| lbu        | list-buffers            | about              |
| del        | delete-windows          | delete-buffer      |
| delb       | delete-buffer           | darken             |
| delbo      | delete-bookmark         | darken             |
| liscon (+) | about                   | about              |
| nyin       | nyxt-init-time          | download-mode      |
| emmo       | emacs-mode              | zoom-images        |
| remhand    | enable-hook-handler (-) | darken             |

*: There appear to be quirks in filtering algorithm (unrelated to scoring),
which in those case gives only a few (1-3) bad candidates. I may investigate this
in the future.

+: It does search in docstring!

-: It appears that there's a
mistake in docstring: enable-hook-handler says "Remove handler(s) from
a hook" and got picked up. disable-hook-handler is the second place.
```

aartaka:
> mistake in docstring: enable-hook-handler says "Remove handler(s) from
> a hook" and got picked up. disable-hook-handler is the second place.

Thanks for flagging this! Should be fixed by 51c8e16064789702678baaa992e389db62201680.

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
+    (flet ((word-end ()
+             ;; bonus for word ending early (shorter words at the beginning wins)
+             (let ((next (position-if (lambda (c) (member c word-separator :test 'eq))
+                                      suggestion-string :start i)))
+               (unless next (setq next (length suggestion-string)))
+               (incf score (/ 1.0 (+ 3.0 next)))
+               (setq i 0))))
+      ;; flex match, with higher weight for the beginning
+      (loop for c across input do
+        (if (member c word-separator :test 'eq)
```
aartaka, line 28: `:test` doesn't make any difference for chars. Both `eq` and `eql` behave the same. You can safely remove `:test` there.

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
-     (* 1.0 (substring-norm (str:split " " input) suggestion-string))))
+  (let ((i 0)
+        (score 0.0)
+        (lastchar #\ ))
+    (declare (optimize (speed 3) (safety 0))
+             (type single-float score) (type fixnum i) (type character lastchar)
+             (type (simple-array character) input suggestion-string))
+    (flet ((word-end ()
+             ;; bonus for word ending early (shorter words at the beginning wins)
+             (let ((next (position-if (lambda (c) (member c word-separator :test 'eq))
```
aartaka, line 21: `:test` is not necessary, see other comment.

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
-  ;; Damerau-Levensthein but it's much slower.
-  ;; TODO: Check out fzf for a possibly good scoring algorithm.
-  (+ (* 1.0 (mk-string-metrics:norm-damerau-levenshtein suggestion-string input))
-     (* 1.0 (substring-norm (str:split " " input) suggestion-string))))
+  (let ((i 0)
+        (score 0.0)
+        (lastchar #\ ))
+    (declare (optimize (speed 3) (safety 0))
+             (type single-float score) (type fixnum i) (type character lastchar)
+             (type (simple-array character) input suggestion-string))
```
aartaka, line 18: `(simple-array character)` = `string`? It's more readable this way.

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
+             (let ((next (position-if (lambda (c) (member c word-separator :test 'eq))
+                                      suggestion-string :start i)))
+               (unless next (setq next (length suggestion-string)))
+               (incf score (/ 1.0 (+ 3.0 next)))
+               (setq i 0))))
+      ;; flex match, with higher weight for the beginning
+      (loop for c across input do
+        (if (member c word-separator :test 'eq)
+            (word-end)
+            (let ((next (position c suggestion-string :start i :test 'eq)))
```
aartaka, line 30: `:test` is not necessary.

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
+              (if next
+                  (progn
+                    ;; bonus for continuous match
+                    (when (and (> next 0)
+                               (eq (aref suggestion-string (1- next)) lastchar))
+                      (incf score 0.3))
+                    (incf score (/ 1.0 (+ 3.0 next)))
+                    (setq i next lastchar c))
```
aartaka, line 38: Putting pairs of `setf` elements on separate lines is more readable. Also, we tend to use `setf` everywhere for consistency :)

I see that `lastchar` and `i` are only used inside the loop. What about putting those into a `with` loop clause?

aartaka (commented):
Thank you (again) for suggesting this!

All in all, it looks cool, although a little cryptic :D 

There are some style/redundancy suggestions that I have, it'll be cool if you fix those :)

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
+        (score 0.0)
+        (lastchar #\ ))
+    (declare (optimize (speed 3) (safety 0))
+             (type single-float score) (type fixnum i) (type character lastchar)
+             (type (simple-array character) input suggestion-string))
+    (flet ((word-end ()
+             ;; bonus for word ending early (shorter words at the beginning wins)
+             (let ((next (position-if (lambda (c) (member c word-separator :test 'eq))
+                                      suggestion-string :start i)))
+               (unless next (setq next (length suggestion-string)))
```
Ambrevar, line 23: To avoid unnecessary `setf` calls, you could let-bind `next` directly to 

```
(or (position-if ... 
     (length suggestion-string))
```

It's also shorter.

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
+               (incf score (/ 1.0 (+ 3.0 next)))
```
Ambrevar, line 24: Why +3 ?
I recommend you let-bind the value to a meaningful name, and if the name alone is not documenting enough, add a docstring / comment.

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
+               (setq i 0))))
+      ;; flex match, with higher weight for the beginning
+      (loop for c across input do
+        (if (member c word-separator :test 'eq)
+            (word-end)
```
Ambrevar, line 29: If the "then" clause is flet, then why not make the "else" clause flet as well?

libraries/prompter/filter.lisp
```diff
@@ -3,47 +3,42 @@
...
+            (let ((next (position c suggestion-string :start i :test 'eq)))
+              (if next
+                  (progn
+                    ;; bonus for continuous match
+                    (when (and (> next 0)
+                               (eq (aref suggestion-string (1- next)) lastchar))
+                      (incf score 0.3))
```
Ambrevar, line 36: Same for the magic number 0.3: can you bind it and document it?

Ambrevar:
We would need a benchmark test as well to ensure that we don't go slower than before, because this is critical to the prompt buffer performance.

libraries/prompter/filter.lisp
BlueFlo0d: `string` also includes `base-string`? Some other part of Nyxt source seems to account for this, but maybe it's just because `mk-string-metrics` need `(simple-array character)` instead of the potentially larger `string` type?

BlueFlo0d:
Regarding benchmark, I just tried a single input (which is totally unscientific but maybe not so unscientific because the difference is 1.5 order of magnitude):

Flex:
```
(time (loop for i below 100 do
  (prompter::score-suggestion-string-1 "sw" "switch-buffer")))

Evaluation took:
  0.000 seconds of real time
  0.000017 seconds of total run time (0.000017 user, 0.000000 system)
  100.00% CPU
  54,391 processor cycles
  0 bytes consed
```

Old:
```
(time (loop for i below 100 do
  (prompter::score-suggestion-string-1 "sw" "switch-buffer")))

Evaluation took:
  0.000 seconds of real time
  0.000452 seconds of total run time (0.000444 user, 0.000008 system)
  100.00% CPU
  1,400,981 processor cycles
  179,968 bytes consed
```

libraries/prompter/filter.lisp
BlueFlo0d: It's used a little bit outside the loop (the last `(word-end)`), and if it's moved into `finally`, the `flet` is still outside the loop.

aartaka: Ah, sorry, didn't notice it :sweat_smile:

libraries/prompter/filter.lisp
aartaka: Every `base-string` is `string`, yes. We don't need the exact typing of `base-string` here, especially now that we don't rely on `mk-string-metrics`.

Ambrevar:
Looking really good at this point!

For the benchmark, could you add it to a test file (see `tests/benchmark`) and use a benchmark library?
We use trivial-benchmark for now, feel free to suggest something else if you find something better.

I recommend you test more inputs, in particular inputs with about 10 words or bigger length.  In the past we had a performance issue and the algorithm had a O(n!) cost in the number of words... (or was it letters?)
`sw` is too short to be significant.

BlueFlo0d:
Added a random benchmark. My results:

Old
```
real-time:        1.4551153    [1.449324, 1.459504]    (median 1.455777, deviation 0.0030146528, total 14.551153)
 run-time:         1.4561883    [1.45037, 1.460512]     (median 1.456908, deviation 0.0030444318, total 14.561883)
 user-run-time:    1.4527795    [1.448307, 1.456279]    (median 1.4527, deviation 0.0026130534, total 14.527795)
 system-run-time:  0.0034135    [0.00153, 0.006623]     (median 0.002963, deviation 0.00139568, total 0.034135)
 gc-run-time:      6.8003       [3.838, 9.394]          (median 5.508, deviation 2.3869004, total 68.003)
 bytes-consed:     3.2042403e8  [320414240, 320436416]  (median 320423968, deviation 5063.544, total 3204240368)
```

Flex
```
real-time:        0.1040545    [0.103233, 0.105354]    (median 0.103564, deviation 8.248357e-4, total 1.040545)
 run-time:         0.1041051    [0.103262, 0.105431]    (median 0.103653, deviation 8.207629e-4, total 1.041051)
 user-run-time:    0.1036938    [0.103193, 0.105137]    (median 0.103585, deviation 5.515888e-4, total 1.036938)
 system-run-time:  4.118e-4     [5.4e-5, 0.001298]      (median 7.1e-5, deviation 4.6692244e-4, total 0.004118)
 gc-run-time:      0            [0, 0]                  (median 0, deviation 0.0, total 0)
 bytes-consed:     3275.2       [0, 16384]              (median 0, deviation 6550.401, total 32752)
```

Also when playing with it, I found that I can make my function works with `string` (rather than `(simple-array character)`), and we could clean up the conversions elsewhere in Nyxt a bit. Do you think it's worth it?

aartaka:
> Added a random benchmark. My results:
> 
> Old
> 
> ```
> real-time:        1.4551153    [1.449324, 1.459504]    (median 1.455777, deviation 0.0030146528, total 14.551153)
>  run-time:         1.4561883    [1.45037, 1.460512]     (median 1.456908, deviation 0.0030444318, total 14.561883)
>  user-run-time:    1.4527795    [1.448307, 1.456279]    (median 1.4527, deviation 0.0026130534, total 14.527795)
>  system-run-time:  0.0034135    [0.00153, 0.006623]     (median 0.002963, deviation 0.00139568, total 0.034135)
>  gc-run-time:      6.8003       [3.838, 9.394]          (median 5.508, deviation 2.3869004, total 68.003)
>  bytes-consed:     3.2042403e8  [320414240, 320436416]  (median 320423968, deviation 5063.544, total 3204240368)
> ```
> 
> Flex
> 
> ```
> real-time:        0.1040545    [0.103233, 0.105354]    (median 0.103564, deviation 8.248357e-4, total 1.040545)
>  run-time:         0.1041051    [0.103262, 0.105431]    (median 0.103653, deviation 8.207629e-4, total 1.041051)
>  user-run-time:    0.1036938    [0.103193, 0.105137]    (median 0.103585, deviation 5.515888e-4, total 1.036938)
>  system-run-time:  4.118e-4     [5.4e-5, 0.001298]      (median 7.1e-5, deviation 4.6692244e-4, total 0.004118)
>  gc-run-time:      0            [0, 0]                  (median 0, deviation 0.0, total 0)
>  bytes-consed:     3275.2       [0, 16384]              (median 0, deviation 6550.401, total 32752)
> ```

Cool!

> Also when playing with it, I found that I can make my function works with `string` (rather than `(simple-array character)`), and we could clean up the conversions elsewhere in Nyxt a bit. Do you think it's worth it?

Let's leave it for later, this patch is good enough to be merged already :)

If you'd like to clean it all up, though, I'm not going to stop you :D

aartaka:
CI fails with the MK-STRING-METRICS dependency of `prompter`. Can your remove the dependence? We'll merge after CI passes.

aartaka:
Hmmmm, that is interesting. Some test are failing, some with quite unintuitive results. Would you mind running tests on your machine and fixing those that fail?

BlueFlo0d:
I think those results are because of race condition... I run those test manually and they output expected result. It's probably the old algorithm runs more slowly so it doesn't trigger those race condition, but the Flex algorithm is too fast. Let me investigate.

BlueFlo0d:
Fixed, also I see that `prompter/tests` uses a snapshot version of `*suggestions*` rather than `nyxt::*command-list*` to do all tests. Should I do the same for `tests/benchmark/prompter`? I don't know if I should let benchmark system depends on `nyxt/prompter/tests` though.

aartaka:
Cool! Merging then?

aartaka:
Merged as 568f7647309f0f45f023d9775d965462811a5dc5. There was a merge commit on your branch, so I cherry-picked everything except for it. I'll clutter git history much less if you rebase instead of merging :)

In any way, thanks for this change! I'm already rebuilding Nyxt to have this awesome fuzzy matching :D

PR #1952, opened 2021-12-20
Add `remove-command'

BlueFlo0d:
This is useful for a long running image, for which it's important for all operations to be undoable.


Ambrevar:
Nice addition indeed!
Can you squash the commits?

What about naming it `delete-command` considering that it has a stateful behaviour, in the line of `delete-duplicates` vs. `remove-duplicates`?

BlueFlo0d:
All fixed, I suppose :)

Ambrevar:
Merged with 4dc8c052dc55801c267148ef0fde1d11b395b284 with an edited commit message which update the command name ;)

Thanks for this useful contribution!

PR #1953, opened 2021-12-20
window.lisp: Add `delete-all-panel-buffers` command

guojing0:
I found it really helpful when implementing panel buffer-related functionality. It should be helpful to users when we have more kinds of panels in the future.


aartaka:
But what about multiple selection in `delete-panel-buffer`? You can simply mark all the buffers (it's just one keybinding) and delete all of those :)

Ambrevar:
Agreed with @aartaka, in my opinion what we have is already good enough.  @guojing0 Is there a specific reason why you'd like to have a dedicated command?

PR #1954, opened 2021-12-20
Add `show-headings-panel`

guojing0:
Add a new panel `show-headings-panel` so that users could select which headings they would want to visit.


source/mode/jump-heading.lisp
```diff
@@ -72,3 +72,23 @@ of buffers."
                              :name (format nil "Headings: ~a" (title buffer))
                              :buffer buffer)))))
 
+(nyxt::define-panel headings (panel-buffer)
+  (flet ((buffer-markup (heading)
+           "Create the presentation for a buffer."
+           (spinneret:with-html
+             (:p (:a :class "button"
+                     :href (lisp-url `(switch-buffer :id ,(id (buffer heading)))
+                                     `(scroll-page-to-heading heading))
```
aartaka, line 81: `heading` should have a comma in front of it so that it doesn't break the action.

However, even with this, it shouldn't work, as `heading` is an object, not a literal. I suggest you use `(scroll-to-element :key ,(get-nyxt-id (element heading)))` instead.

source/mode/jump-heading.lisp
```diff
@@ -72,3 +72,23 @@ of buffers."
...
+                     (:span :title (title heading)
+                            :class "title" (title heading)))))))
+    (spinneret:with-html-string (:style (style panel-buffer))
+      (:style (cl-css:css
```
aartaka, line 85: Make sure to use `theme:themed-css` and 
```
:background-color theme:secondary
:color theme:background
```
inside the button rule. This way, user theming will apply to this button.

source/mode/jump-heading.lisp
```diff
@@ -72,3 +72,23 @@ of buffers."
...
+               '((".button"
+                  :white-space "nowrap"
+                  :overflow-x "hidden"
+                  :display "block"
+                  :text-overflow "ellipsis"))))
+      (:body
+       (:h1 "Headings")
+       (loop for heading in (get-headings)
```
aartaka, line 93: This `for ... in ... collect` is synonymic to `(mapcar #'buffer-markup (get-heading))` and I'm not sure it's adding any significant information. Why `loop` over `mapcar` here?

aartaka: However, if you prefer `loop` over everything, like @jmercouris does, that's fine too -- `loop` is indeed a powerful abstraction :)

aartaka (commented):
Cool stuff! What would be the procedure to test that this works as intended?

source/mode/jump-heading.lisp
guojing0: Thank you for the suggestion. I think when testing within the virtual machine I had the comma.

When playing with `show-buffers-panel` it also displayed same kind of error, so I was unsure whether it's my problem or what not.

source/mode/jump-heading.lisp
guojing0: I copied and pasted from another file, in which it used `loop` and I did not pay much attention to it. I do like `loop`, but you are right, here `mapcar` is better.

guojing0:
@aartaka I tested it on the documentation and some websites.

aartaka:
I've tested this patch a while ago, and, it seems, using `lisp:` URLs in panel buffers is not exactly possible. I'll test again, but I suggest to wait for #1815 so that the mechanism is consistent with the revamped system there.

guojing0:
@aartaka Interesting, I will. I think there is `show-buffers-panel` command that does something similiar.

guojing0:
See https://github.com/atlas-engineer/nyxt/pull/1990/

PR #1959, opened 2021-12-23
Fix some build failures related to *features*

BlueFlo0d:
- Don't call webkit:webkit-web-view-is-muted without webkit2-mute
feature.
- Some CL implementations gives several OS features (e.g. I
have both :darwin and :bsd), account for that.


aartaka:
Thanks for this patch! Even though `cons*` looks fishy to me, I don't see any immediate problem with it :)

PR #1961, opened 2021-12-23
mode/visual: Refactor into caret-action helper

hendursaga:
Simple code de-duplication.

### DIFF
--- a/source/mode/visual.lisp
+++ b/source/mode/visual.lisp
@@ -95,6 +95,11 @@
 (defmethod %follow-hint ((element nyxt/dom:text-element))
   (set-caret-on-start :nyxt-identifier (get-nyxt-id element)))
 
+(defmethod caret-action ((mode visual-mode))
+  (if (mark-set mode)
+      "extend"
+      "move"))
+
 (define-parenscript block-page-keypresses ()
   (setf (ps:@ window block-keypresses)
         (lambda (event)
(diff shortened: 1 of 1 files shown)


aartaka:
The way I'd refactor it: rename the `action` arg to `extend-p`  and pass `(mark-set mode)`. Does it feel better to you?

hendursaga:
@aartaka that's an idea; however, keep in mind `action` represents `alter` in the [parameters documented here](https://developer.mozilla.org/en-US/docs/Web/API/Selection/modify#parameters). Honestly, though, `alter` seems like a poor choice for an argument name, but seeing as they're not passed via ParenScript as keyword/named arguments.. what matters is whether or not :extend or :move is passed. I was thinking of refactoring further using typed enums, so I think this commit needs no further modifications.

aartaka:
> @aartaka that's an idea; however, keep in mind `action` represents `alter` in the [parameters documented here](https://developer.mozilla.org/en-US/docs/Web/API/Selection/modify#parameters). Honestly, though, `alter` seems like a poor choice for an argument name, but seeing as they're not passed via ParenScript as keyword/named arguments.. what matters is whether or not :extend or :move is passed. I was thinking of refactoring further using typed enums, so I think this commit needs no further modifications.

Does make sense, thanks for the explanation!

aartaka:
Thanks for the patch @hendursaga!

PR #1971, opened 2021-12-28
Refactor visual mode

hendursaga:
I mentioned I had more coming, did I not? :smile:

In addition to adding enum types for the various options for the parameters for the Web Selection API or whatever you'd call it, I added a few new "scales" with their respective commands. That and some refactoring.

The only thing missing is to make the bindings work for the new `forward-document`/`backward-document` commands. I've discussed some with @jmercouris, but he's on vacation and I couldn't figure out the override-map idea he mentioned.

### DIFF
--- a/source/mode/visual.lisp
+++ b/source/mode/visual.lisp
@@ -49,7 +49,9 @@
        "C-a" 'beginning-line
        "C-e" 'end-line
        "M-a" 'backward-sentence
-       "M-e" 'forward-sentence)
+       "M-e" 'forward-sentence
+       "M-}" 'forward-paragraph
+       "M-{" 'backward-paragraph)
       ;; vi keybindings only enable use of vim's plain "visual" mode for now
       scheme:vi-normal
       (list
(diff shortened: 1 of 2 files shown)


aartaka:
Looks fine and clean to me. Out of interest: why keywords?

The keybindings issue sounds like a bug, as more recent modes should be prioritized over the old ones when we dispatch kybindings. I opened #1973 for the discussion of it.

hendursaga:
@aartaka, to me, I feel seeing a keyword versus some random string signals "hey, there's something special about this, this isn't just any ol' string." Also I get auto-completion for free, and I don't think I got that before conversion.

aartaka:
> @aartaka, to me, I feel seeing a keyword versus some random string signals "hey, there's something special about this, this isn't just any ol' string." Also I get auto-completion for free, and I don't think I got that before conversion.

Yes, that's enough of a reason :) Let me merge this, and we'll fix the bug with keybindings a bit later.

aartaka:
Thanks for your contribution, @hendursaga!

PR #1975, opened 2021-12-29
Cleaned stray commit, refactored, cleaned whitespace, clarification.

opalvault:
After a lot of messing with my fork I think I got it figured out. This PR should only contain the only file and commits that I've done.

Let me know how it looks on your end.

### DIFF
--- a/documents/README.org
+++ b/documents/README.org
@@ -388,16 +388,16 @@ Prerequisites:
     
 Steps:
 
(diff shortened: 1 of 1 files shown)


aartaka:
There still was one merge commit left on your branch, I've removed it :) Please remember to rebase onto the upstream or squash merge commits into the ones with content, as they don't add anything meaningful. Merging this.

PR #1979, opened 2021-12-31
Easier-to-understand code structure of `manual.lisp`

guojing0:
The purpose of this PR is to have an easily understood code structure of `manual.lisp` file. It is achieved via the following 3 steps:

1. `tutorial.lisp` only contains `tutorial-content` function, which has been moved to `manual.lisp`. We can could remove `tutorial.lisp`, so that people would not be confused by presence of both `tutorial.lisp` and `manual.lisp`.
2. Besides having `tutorial-content` function in `manual.lisp`, we could also abstract "Configuration" part of manual into a separate `config-content` function.
3. As a result, we then could call two said functions in the main `manual-content` function at the top of `manual.lisp` file.

### DIFF
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -6,11 +6,14 @@
 (defun manual-content ()
   (str:concat
    (spinneret:with-html-string
-    (:h1 "Nyxt manual")
-    (:p "This manual first includes the tutorial, then covers the configuration
+     (:h1 "Nyxt manual")
+     (:p "This manual first includes the tutorial, then covers the configuration
 of Nyxt."))
    (tutorial-content)
-   (spinneret:with-html-string (:h2 "Configuration")
+   (config-content)))
+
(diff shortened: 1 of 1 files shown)


Ambrevar:
I actually like to have one file per "document", this makes it easier to focus on the content rather than have Lisp code scattered around.  What do you think about this point of view?

guojing0:
I agree, though I think it may be good to abstract out a function `config-content`, so that it's "more uniform" with `(tutorial-content)`.

Ambrevar:
Agreed with that last point!

guojing0:
@Ambrevar Done

Ambrevar:
Looking good!

Can you squash away the first commit?
Also if you think that it needs re-indenting, can you do it in a separate commit?
Thanks!

guojing0:
Done.

PR #1981, opened 2021-12-31
source/tutorial: update tutorial after command rename.

pdelfino:
Just updating the tutorial after `application-mode` became `passthrough-mode`.


aartaka:
Thanks!

PR #1982, opened 2021-12-31
source/global.lisp: fix small grammar mistake

pdelfino:
It is a small typo being fixed. The current phrase is:

> Add a handler can be added with:

It could be changed to "Add a handler with" or "A handler can be added with". I have picked up the latter.


aartaka:
That's a sharp eye that you have, Pedro! Thanks :)

PR #1984, opened 2022-01-03
makefile: Set SBCL dynamic-space-size to 3 GiB.

### LINKED ISSUE #1946: Increase dynamic space size for .deb and in makefile
SBCL default dynamic space size varies across installation.

I suggest we set it to 3 GiB by default, it may fix some issues like the one reported by @jmercouris in https://github.com/atlas-engineer/nyxt/pull/1940#issuecomment-996929347.

Thoughts?

Ambrevar:
Also enable core compression for Ubuntu package to allow users to compress Nyxt
from itself.

Should fix #1946.


Ambrevar:
The CI now passes.  OK to merge?

PR #1986, opened 2022-01-04
libraries/web-extensions/Makefile: Fix pkg-config search and CC fallb…

Ambrevar:
…ack.

On Guix, "CC" is "cc" but it's missing in the environment.

While we are at it, we list the .o files automatically.

This should fix the CI on master.
@aartaka What do you think?


Ambrevar:
I've also updated the license headers in all the web-extension files.
@aartaka Don't forget to include the header when creating new files ;)

aartaka:
Looks like the most elegant Makefile spells I've seen :D

I will remember adding license headers, yes!

jmercouris:
I'm telling you man, Pierre is like some sort of compilation wizard. I don't know how his beard became so long!

Issue #1640, opened 2021-07-20
Overhaul make-buffer, buffer-make and friends


Ambrevar:
We have too many functions and commands with confusing names for making buffers.

- `buffer-make`
- `make-buffer`
- `make-buffer-focus`
- `make-*-buffer`...

`buffer-make` does not seem very useful.
I suggest to `flet` it in `make-buffer`.

`make-buffer-focus` lacks some of the `make-buffer` parameters, like `load-url-p`.  Other `make-*-buffer` might lack parameters as well.

Looking at the lambda lists of `make-buffer-focus` and `make-nosave-buffer`, we can see that we have a composition issue.  For instance by default we can't bind a key to "make nosave buffer and focus".
Same with other `make-*-buffer` commands.

So what about replacing these commands with a unique `make-custom-buffer` that prompts for stuff like:
- the class
- modes
- the booleans (load-url-p, no-history-p)
?

In total, we would have
- `make-buffer`
- `make-buffer-focus` (because it's useful, but should we still?)
- `make-buffer-custom` (or `make-custom-buffer`?)

Thoughts?

Ambrevar:
Actually `buffer-make` could be moved to an `initialize-instance :after`.  We could move the keyword parameters there as well, so that we would still be able to make dummy buffers with something like

```lisp
(make-instance 'user-internal-buffer :dummy t)
```

Ambrevar:
Fixed with #2050.

Issue #1875, opened 2021-11-17
Disable / destroy modes on buffer-delete?


Ambrevar:
It seems that we don't kill modes when deleting buffer.

It's an oversight, isn't it?
Any specific reason why we wouldn't do it?

@jmercouris @pdelfino @aartaka ?

aartaka:
> It's an oversight, isn't it? Any specific reason why we wouldn't do it?

Do we re-initialize modes on buffer restoration? If no, then there's no reason to keep them.

Ambrevar:
We do re-initialize the _default modes_ only.

This is inconsistent and I'd call it a bug, so let's fix this.

Then we have a few options:

- Kill modes when deleting buffers.  This means that reopened buffers
  have their state reset.

- Keep modes when deleting buffers.  State is preserved!

- Add a mode slot option to choose if a mode should be destroyed or not.
  This allows us to fine-tuned the garbage collection behaviour.

jmercouris:
I would go with option two: "Keep modes when deleting buffers.  State is preserved". 

Why did we not originally do this? It is not an oversight. I believe it is because our history was contained within a mode. By keeping the mode alive when restoring a buffer, we could also maintain forward/back etc.

Ambrevar:
That's a good point!

However we _do_ call `initialize-modes` when reopening a buffer, so that
part at least was clearly an oversight.

aadcg:
@Ambrevar what's the status here, given that considerable time and changes have been made in the mean time?

aadcg:
See https://github.com/atlas-engineer/nyxt/issues/2209#issuecomment-2663210070.

Issue #1787, opened 2021-09-14
I get logged out of sites when re-opening Nyxt


anonimitoraf:
**Precise recipe to reproduce the issue**
* Sign in to https://www.youtube.com or https://www.github.com
* Close Nyxt
* Open Nyxt
* Expectation: Still logged in | Reality: Logged out

Is there a setting I have to enable for persistent sessions?

For website-specific issues:
Can you reproduce this issue with Epiphany / GNOME Web (https://wiki.gnome.org/Apps/Web)?
No

**Information**
- OS name+version: Manjaro 21.1.0 Pahvo
- Graphics card and driver: 
NVIDIA GeForce GTX 1060 6GB
VGA compatible controller: NVIDIA Corporation GP106 [GeForce GTX 1060 6GB] (rev a1)
	Subsystem: Micro-Star International Co., Ltd. [MSI] Device 8c94
	Kernel driver in use: nvidia
	Kernel modules: nouveau, nvidia_drm, nvidia
- Desktop environment / Window manager name+version: i3 version 4.19.1 (2021-02-01) © 2009 Michael Stapelberg and contributors
- How you installed Nyxt (Guix pack, package manager, build from source): From source
- Information from `show-system-information`:
```
Nyxt version: 2.1.1-624-g935278d3
Renderer version: GI-GTK
Operating system kernel: Linux 5.12.19-1-MANJARO
Lisp implementation: SBCL 2.1.7 (Dynamic space size: 1073741824)
Features: (:WEBKIT2 :WEBKIT2-2.32 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
[...]

PR #1987, opened 2022-01-04
gtk.lisp: Web context sharing and separation (REBASE)

Ambrevar:
This is a follow up to #1757.
It's rebased on `2-series` and should address all suggestions  from the previous pull request.

This will hopefully fix the sudden log-out issues such as in #1787.

EDIT:  Fixes:
- #1733 
- #1640
- (Maybe) #837
- (Maybe) #797 

EDIT 2: To fix afterwards:
- #1875 (initialize-modes)

### DIFF
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -31,10 +31,12 @@ want to change the behaviour of modifiers, for instance swap 'control' and
 \(define-configuration browser
   ((modifier-translator #'my-translate-modifiers)))")
    (web-contexts (make-hash-table :test 'equal)
-                 :documentation "Persistent `nyxt:webkit-web-contexts'
-Keyed by strings as they must be backed to unique folders")
+                 :documentation "Persistent `nyxt::webkit-web-context's
+Keyed by strings as they must be backed to unique folders
+See also the `ephemeral-web-contexts' slot.")
    (ephemeral-web-contexts (make-hash-table :test 'equal)
-                           :documentation "Ephemeral `nyxt:webkit-web-contexts'"))
+                           :documentation "Ephemeral `nyxt::webkit-web-context's.
+See also the `web-contexts' slot."))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
@@ -116,8 +118,9 @@ failures."))
 (defclass webkit-website-data-manager (webkit:webkit-website-data-manager) ()
   (:metaclass gobject:gobject-class))
 
-;; create-gobject-from-class in gobject.base.lisp directly reads the initargs to make-instance, so this cannot
-;; be used to enforce ephermerality (or anything else that needs to happen in the cffi constructor)
+;; `gobject::create-gobject-from-class' directly reads the initargs to
+;; `make-instance', so this cannot be used to enforce ephemerality (or anything
+;; else that needs to happen in the CFFI constructor).
 (defmethod initialize-instance :after ((data-manager webkit-website-data-manager) &key)
   #+webkit2-tracking
   (webkit:webkit-website-data-manager-set-itp-enabled data-manager t)
--- a/source/renderers/gtk.lisp
+++ b/source/renderers/gtk.lisp
@@ -29,11 +29,25 @@ want to change the behaviour of modifiers, for instance swap 'control' and
     (delete nil (mapcar (lambda (mod) (getf plist mod)) modifier-state))))
 
 \(define-configuration browser
-  ((modifier-translator #'my-translate-modifiers)))"))
+  ((modifier-translator #'my-translate-modifiers)))")
+   (web-contexts (make-hash-table :test 'equal)
+                 :documentation "Persistent `nyxt:webkit-web-contexts'
+Keyed by strings as they must be backed to unique folders")
+   (ephemeral-web-contexts (make-hash-table :test 'equal)
+                           :documentation "Ephemeral `nyxt:webkit-web-contexts'"))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
 (define-user-class browser (gtk-browser))
+
+(defmethod get-context ((browser gtk-browser) name &key ephemeral-p)
+  (if ephemeral-p
+      (or (gethash name (ephemeral-web-contexts browser))
+          (setf (gethash name (ephemeral-web-contexts browser))
+                (make-context name :ephemeral-p t)))
+      (or (gethash name (web-contexts browser))
+          (setf (gethash name (web-contexts browser))
+                (make-context name)))))
 
 (define-class gtk-window ()
   ((gtk-object)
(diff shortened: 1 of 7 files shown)


Ambrevar:
Another logout issue is mentioned in #1625.

Ambrevar:
@aartaka @jmercouris Could you please proof-read this before we merge?

Ambrevar:
I suggest the following plan to make sure this gets tested:  merge on 2-series and master, then wait a week before we release 2.2.4.

Cool?

aartaka:
> I suggest the following plan to make sure this gets tested: merge on 2-series and master, then wait a week before we release 2.2.4.

That's a strong move xD

Ambrevar:
What do you mean?

source/renderers/gtk.lisp
```diff
@@ -29,12 +29,29 @@ want to change the behaviour of modifiers, for instance swap 'control' and
...
+See also the `ephemeral-web-contexts' slot.")
+   (ephemeral-web-contexts (make-hash-table :test 'equal)
+                           :documentation "Ephemeral `nyxt::webkit-web-context's.
+See also the `web-contexts' slot."))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
 (define-user-class browser (gtk-browser))
 
+(alex:define-constant +internal+ "internal" :test 'equal)
```
aartaka, line 45: I fear no man, but CL constants make me shiver |_・)

source/renderers/gtk.lisp
```diff
@@ -208,21 +268,58 @@ not return."
   "We shouldn't enable (possibly) user-identifying extensions for `nosave-data-profile'."
   nil)
 
+(defmethod expand-data-path ((profile data-profile) (path gtk-extensions-data-path))
+  "Return finalized path for gtk-extension directory."
+  (expand-default-path
+   path
+   :root (uiop:strcat (uiop:native-namestring (dirname path))
```
aartaka, line 275: Today I learned of `uiop:strcat`. Neat function!

source/renderers/gtk.lisp
```diff
@@ -616,38 +713,43 @@ See `gtk-browser's `modifier-translator' slot."
...
+                 (make-instance 'webkit-web-context
+                                :website-data-manager
+                                (make-instance 'webkit-website-data-manager
+                                               :base-data-directory (expand-path data-manager-data-path)
+                                               :base-cache-directory (expand-path data-manager-cache-path))))))
+         (gtk-extensions-path (expand-path (make-instance 'gtk-extensions-data-path :context-name name)))
+         (cookie-manager (webkit:webkit-web-context-get-cookie-manager context)))
+    (unless (or ephemeral-p
+                (internal-context-p name))
+      (unless (uiop:emptyp gtk-extensions-path)
```
aartaka, line 735: Another requirement for the refactoring of the `data-storage` library: make a macro controlling the expansion of data paths. Something like:
```
(with-paths-allowed (path1 path2 ...)
  ... Code requiring paths ...)
```

aartaka (commented):
I have particular feedback except some collateral insights there :D

This change is complex, despite the documentation. How about we -- I'm tempted to try doing that myself, but I already have too much stuff on me (´＿｀。) -- try adding a couple of context-controlling commands so that there're examples of using the new API available for future us? This will also give us some understanding of how to mimic Firefox's Containers.

aartaka:
> What do you mean?

I mean, putting it on master will leave us no choise but to debug it, ahah. Quite a good strategy it is :)

source/renderers/gtk.lisp
Ambrevar: Alexandria is here for you!  Alternatively we could use keywords (and deftype something like `selection-action`), but then it would be a bit harder to give arbitrary names to contexts (think case and whitespaces).  What do you think?  

Should we restrict context names for the sake of better code?

source/renderers/gtk.lisp
Ambrevar: Yup! Very handy in `.asd` files!

source/renderers/gtk.lisp
Ambrevar: Noted!

Ambrevar:
I think it's a good point, I'll try adding a few commands before merging this.

It might be possible to write tests if we can control where all the data is stored.  I'll work on it.

source/renderers/gtk.lisp
aartaka: It's optimal in the current form, I think :)

Ambrevar:
So I've looked into the question of the user-facing interface.

For now we make buffers with `buffer-make` and there is no `context` argument.  So following the current API's logic we would add an argument and propagate it down the call stack.  But clearly this leaves to be desired and calls for some re-design: indeed, every new option we want to add to the buffer class calls for this kind of change (+1 arg to propagate to a bunch of functions),  which does not scale well.

With @jmercouris we had talked about redesigning `buffer-make` and `make-buffer`:

- Make it one function which just calls `make-instance 'user-buffer` with some good default arguments.
- Move the `buffer-make` code to `initialize-instance :after` (or `:around` if need be).

Then the implementation of the user facing interface for context would be trivial: we just add a `context-name` slot to `buffer` and change `make-web-view` to get the context name from the buffer slot.

With that in place, the user can create buffer with arbitrary context by calling

```lisp
(make-instance 'user-buffer :context-name "foo")
```

Does that make sense?
@aartaka @jmercouris What do you think?

If you agree, then I'll start refactoring `buffer-make` and `make-buffer` as suggested.

Ambrevar:
At this point, it's clear that this pull request does NOT belong to `2-series`.  Don't merge! :)  I'll rebase on `master` later.

aartaka:
Looks fine to me :)

Ambrevar:
Question: Should we collapse the concepts of data-profile with that of context name?
In practice, this would mean that the context name would default to that of the data profile.  This would make it easier for the user to understand: new data-profile means new web context.

Ambrevar:
See also #1875 about `initialize-modes`.  I believe we should fix it once this pull request is merged.

jmercouris:
This feels like something I've wanted to literally do for months! It isn't perfect, but a great start! Thank you Pierre!

If you remember, when we first did the GTK rewrite we just had make instance, and it was great!

Ambrevar:
> If you remember, when we first did the GTK rewrite we just had make instance, and it was great!

Sorry, I don't remember :P  But I trust you!

I can't remember the reason why we came up with buffer-make.  I suppose
we wanted to factor some code or expose an easier API.  That's a lesson
for the future!

Ambrevar:
I'll proceed then and merge on master.

Ambrevar:
This should also fix #1640.

Ambrevar:
Closing in favour of #2050.

Ambrevar:
> I can't remember the reason why we came up with buffer-make. I suppose we wanted to factor some code or expose an easier API. That's a lesson for the future!

I think I remember now: it's because code in `buffer-make` requires the view to be instantiated (e.g. modes), but this view is instantiated in the renderer, not in `buffer.lisp`.  So it's a bit tricky to get the load order right.  I guess this is why back in the days we've opted for `buffer-make`, a function that first fully instantiates the object with its view, and _then_ runs the rest of the code that depends on the view.

In hindsight, `buffer-make` is not necessary but it wasn't very obvious back then.

PR #1995, opened 2022-01-06
Fix *Bookmarks* buffer after regression introduced in 04b6284.

shamazmazum:
Hello. After 04b6284 interactive "Delete" button was broken in `*Bookmarks*` buffer (invoked by running `list-bookmarks` command). This patch fixes it


aartaka:
Thanks!

Issue #1966, opened 2021-12-27
Current make-hook/make-handler practice is too rigid?


BlueFlo0d:
It seems that currently, all those functions accept function objects instead of symbols.
Therefore there's no easy way to redefine a handler function, which makes it very difficult to develop on a long running image. Am I missing anything?

aartaka:
> It seems that currently, all those functions accept function objects instead of symbols.
> Therefore there's no easy way to redefine a handler function, which makes it very difficult to develop on a long running image. Am I missing anything?

There is a way -- simply do `hooks:remove-hook` and pass the name of the handler to it (that's the whole purpose for the named handlers). After it's removed, you can `hooks:add-hook` again, thus updating the named hook.

BlueFlo0d:
> 
> There is a way -- simply do `hooks:remove-hook` and pass the name of the handler to it (that's the whole purpose for the named handlers). After it's removed, you can `hooks:add-hook` again, thus updating the named hook.

However hooks are not named (and we can't name them because almost all hooks are buffer local). One need to iterate over all buffer to update a hook this way. IMO it's too much a hassle comparing to just add one needed level of indirection. What do you think?

In general, I'm not sure if `serapeum:hooks` is the right thing for our use cases. We have very few use cases for anonymous function handlers (I counted 4), and all of them just need a little bit of rewrite to turn into named function handlers by turning closure variables into buffer-local-variable, and by doing so we make them much more transparent. In the end we just need something as simple as
```
(defclass hook () ((functions) (combination))
```
Just one `combination` more than what we do in Emacs.

aartaka:
> However hooks are not named (and we can't name them because almost all hooks are buffer local). One need to iterate over all buffer to update a hook this way. IMO it's too much a hassle comparing to just add one needed level of indirection. What do you think?

Sorry, hooks are not named indeed, I've been unclear about it. I meant named handlers — those are easily removable. While your point on buffer iteration is valid, destructively modifying buffer-local hooks is not an intended use-case. They are meant to be `define-configuration`-ed, thus allowing both universality (all the instances of a class get a reasonably similar value) and ability to revert to the older value (due to our user-classes system).
 
> In general, I'm not sure if `serapeum:hooks` is the right thing for our use cases.

It's fun that you put it this way, because `serapeum/contrib/hooks` originated and was made exactly for Nyxt :D

[...]

PR #2001, opened 2022-01-07
Improving hooks facility

BlueFlo0d:
This implements proposal 1,2,3' and 5 in #1966 

Slight backward compatibility breakage: We now use handler-alist (with elements (HANDLER . ENABLE-P)) instead of two separate list handlers and disabled-handlers. Readers of handlers and disabled-handlers are simulated, but not writers.

Also I'm not really familiar with git forge and seems that this PR doesn't include the commits in the `serapeum` submodule (which should point to `https://github.com/BlueFlo0d/serapeum/commit/3e75120a7f30173ca1d1058f71953ff3d5831b26`). Help wanted!


Ambrevar:
You can't update serapeum's code from Nyxt if this is what you were trying to do.  If you need a Serapeum update, it should happen at https://github.com/ruricolist/serapeum.

Besides, I don't think you should remove the `make-handler-*` layer: this one was designed on purpose for type checking, which gives us a pretty good guarantee that handlers won't break hook execution regardless of the `composition`.

BlueFlo0d:
> You can't update serapeum's code from Nyxt if this is what you were trying to do. If you need a Serapeum update, it should happen at https://github.com/ruricolist/serapeum.

I talked to `jmercouris` on IRC and he suggests updating our submodule first before asking for merging it into upstream. There is indeed some incompatible change and `serapeum` might want it interface to stay very stable.

> Besides, I don't think you should remove the make-handler-* layer: this one was designed on purpose for type checking, which gives us a pretty good guarantee that handlers won't break hook execution regardless of the composition.

I don't intend to remove the check, it should be done automatically by `make-hook-*` and `add-hook` for symbols. I took a second look and it seems that `add-hook` might not be working properly right now, let me check.

BlueFlo0d:
Turns out that SBCL's type check won't propagate past `symbol-function`. It makes sense -- this is exactly what we want so that functions are redefinable, but SBCL is too cautious to say anything about a function which can be redefined. I need more time to figure this out...

BlueFlo0d:
Ha, I think I come up a way to hack it. I could call `compile` explicitly on a "probe" form to check function type at any time!

There's a remaining question: what we do if type doesn't match? Currently we just get a style warning, which I think is not great. The wrong function still gets into the hook and compromise its functionality.

Overall I think there are some nice improvement over error handling that could be built on our current facility. For example, we can disable a handler if it triggers any error, potentially implemented as a restart. But that's probably another PR.

Also the way I currently run Nyxt (from `sbcl` command line) always prints a `The current thread is not at the foreground` when encountering errors and don't really give me a break loop. Am I doing something wrong?

Ambrevar:
Ha, indeed calling `compile` on a form could be a great trick!

To clarify, the whole reasoning behind the twisted architecture of serapeum/hooks was precisely because I didn't know how to check the type of the function behind a function symbol!

So if we can finally do it more or less cleanly, then we can remove much of the need for the macro and the macro-time function definitions.  I'm guessing that the `make-handler-*` functions could go as well.

Could you provide a proof-of-concept of this "probe form"?

About your last comment: honestly I never run from the `sbcl` command line, so I wouldn't know.
I run Nyxt either from the binary on which I can connect with SLY, or directly from SLY.
Which editor / REPL do you use otherwise?

BlueFlo0d:
> To clarify, the whole reasoning behind the twisted architecture of serapeum/hooks was precisely because I didn't know how to check the type of the function behind a function symbol!

Yep, I could try to clean it up.

> So if we can finally do it more or less cleanly, then we can remove much of the need for the macro and the macro-time function definitions. I'm guessing that the `make-handler-*` functions could go as well.
> 
> Could you provide a proof-of-concept of this "probe form"?

Sure!

```
(defun ftypep-magice (function ftype)
  (handler-case
      (let ((*error-output* (make-broadcast-stream))) ;; silence!
        (compile nil `(lambda () (let ((f ,function)) (declare (type ,ftype f))
                                   f)))
        t)
    (style-warning () nil)))

(defun c1 (x) x)
(defun c2 (x) 2)
(defun c3 (x) "3")
(ftypep-magice #'c1 '(function (string) string)) ;; => T
(ftypep-magice #'c2 '(function (string) string)) ;; => NIL
(ftypep-magice #'c3 '(function (string) string)) ;; => T
```

> 
> About your last comment: honestly I never run from the `sbcl` command line, so I wouldn't know. I run Nyxt either from the binary on which I can connect with SLY, or directly from SLY. Which editor / REPL do you use otherwise?

I'll try connecting with SLY. I can't run it directly from SLY because somehow GUI doesn't show up, so currently I'm doing extremely low-tech development -- well, I do high-tech development in SLY without GUI and then test from commandline with GUI but without SLY. I'll give GUI + connect from SLY a shot.

Ambrevar:
Fantastic!!!

This is awesome, I recommend you present your proposal to the Serapeum
maintainer.

This is going to be backward incompatible for sure, but

considering hooks are a contrib, we could very well have a "v2", or
simply another contrib, like "typed-hooks".

If you'd rather not bother with Serapeum, no worries, you can work on
your fork in the meantime and when it's ready I'll present it to
ruricolist.

About your REPL issues: can you open an issue about it and report your
system information?

BlueFlo0d:
> To clarify, the whole reasoning behind the twisted architecture of serapeum/hooks was precisely because I didn't know how to check the type of the function behind a function symbol!

I'm now working on this. I'm not sure how far you're thinking about going. Do we still need to generate handler-class and hook-class?

BlueFlo0d:
> I'm now working on this. I'm not sure how far you're thinking about going. Do we still need to generate handler-class and hook-class?

I ended up still having such generated classes, but replace `make-handler` and `make-handler-*` by an `initialize-instance :after` method of `handler`, thus doing check under the hood all the time.

Another question about a minor detail: any reason to use a separate `make-hook-*` rather than an `initialize-instance` method?

Ambrevar:
>  I ended up still having such generated classes,

Why?  I'd really like to get rid of them, generating code is in general ugly and it seems to me that we can do without it.

But I can't see the code change than you did in Serapeum.  Could you share your work?  Maybe for Serapeum and update the Nyxt submodule to use it.

BlueFlo0d:
> Why? I'd really like to get rid of them, generating code is in general ugly and it seems to me that we can do without it.

So that `define-hook-type` macro defines a hook class and a handler class, it looks quite intuitive. 

Without this we would likely manually setting a `handler-type` slot for hooks and handlers, which is fine but looks a bit like we're simulating some class (as in category of objects) without using CLOS. But it may has a few benefits too, like expected type specifier is now explicitly in the `hook` objects and can be inspected more directly. Generated hook/handler will hide the specifier in the code of some generic function.

> But I can't see the code change than you did in Serapeum. Could you share your work? Maybe for Serapeum and update the Nyxt submodule to use it.

I thought I shared it? git forge is really confusing... the submodule in my nyxt fork points to https://github.com/ruricolist/serapeum/tree/6ff51d2cac2a649bb0d89956b8e022f8a99d0a97, is this visible?

My changes also contains quite a few improvements of error handling. In fact I think good error handling will alleviate need of type checking to some extent, because malfunctioning handlers won't make the hook non functional and can be fixed on the spot using a `disable-handler` restart.

Ambrevar:
In my understanding, if we can do type checking with `compile` then we don't need a handler class, we can just use `function` or `function-symbol` objects.

Then do define a hook, you'd just define the class and instantiate it.  Does that make sense?

About the repo: what you need to point at is https://github.com/BlueFlo0d/serapeum/, not the ruricolist version for which you don't have write access I suppose.

On second thought, I suggest you _don't_ fork the Serapeum repository.  Instead, create a new one and copy the hook code there.  This way, Nyxt will depend on your hook code but still depend on upstream Serapeum, it will be easier to maintain.

If later your hook system gets accepted in Serapeum then we will just have to remove the submodule corresponding to your repo.

About error handling: support for restart is welcome, but remember that non-interactive support must be there as well, which is why type checking must remain I believe.

Ambrevar:
> Maybe we still have a handler class, but never auto generate any?

Agreed!

BlueFlo0d:
> Another question about a minor detail: any reason to use a separate make-hook-* rather than an initialize-instance method?

This question still remains. Thoughts?

BlueFlo0d:
FYI, my latest version is now at https://github.com/BlueFlo0d/hooks. I'm not sure what I should do to Nyxt repo though, we have to override an subtree inside a submodule?

jmercouris:
Great! If you like I can take your `hooks.lisp` file and create a complete ASDF loadable system, so it can be utilized. I can put it on our GitHub, so that we can merge this work in!

BlueFlo0d:
> Great! If you like I can take your `hooks.lisp` file and create a complete ASDF loadable system, so it can be utilized. I can put it on our GitHub, so that we can merge this work in!

I get it now! I'm doing it myself right now. What's a sensible name? Does simply `hooks` seem prone to collision?

jmercouris:
Hooks seems a fine name, I don't see any collisions in QL.

Ambrevar:
There is cl-hooks, so I would use another name.

"typed-hooks" is an obvious candidate, but feel free to get more
creative!

> Another question about a minor detail: any reason to use a separate make-hook-* rather than an initialize-instance method?

I'm not sure anymore, I'd need to have a closer look at the original
code.  If you think that `make-instance' is good enough as a user facing
API, then we are all set.

Off the top of my head, I'd say the main reason we would want to have
`make-hook` is if the hook initialization arguments must be special somehow.

BlueFlo0d:
> There is cl-hooks, so I would use another name. "typed-hooks" is an obvious candidate, but feel free to get more creative!

I have the bad habit of disliking long names (`call/cc` much better than `call-with-current-continuation`!), so I changed to `nhooks`, in the same vein as Emacs' `nadvice`.

Also @jmercouris feel free to clone the repository!

> I'm not sure anymore, I'd need to have a closer look at the original code. If you think that `make-instance' is good enough as a user facing API, then we are all set. Off the top of my head, I'd say the main reason we would want to have `make-hook` is if the hook initialization arguments must be special somehow.

We used to treat `:handlers` argument specially because to support writing `:handler (list #'named-function)` we have to map over the list and wrap function objects with handlers. But now we can just write `:handler '(named-function)` and nothing need to be done.

Ambrevar:
> I have the bad habit of disliking long names (`call/cc` much better than `call-with-current-continuation`!), so I changed to `nhooks`, in the same vein as Emacs' `nadvice`.

"nhooks" is fine.  Note that Common Lisp now supports "package local
nickname" (see how we use "alex" and "sera"), so you can always name
your package with a long name, optionally provide an official nickname,
then leave the users to define their own nicknames.

The benefit of long names is that it reduces the chances of collision.

jmercouris:
Et voila: https://github.com/atlas-engineer/nhooks

I've added everyone as maintainers.

jmercouris:
@Ambrevar could you please give this a look and merge if you think it is ready?

Ambrevar:
`nhooks` must be added to Guix.
There are some other required changes like adding a `readme`.  I'll work on it.

Ambrevar:
You also need to update build-scripts/nyxt.scm to add `cl-nhooks`.  Let me know if you need help.

Finally, this pull request contains some (spurious?) alterations to the submodules.  Can you remove them?

jmercouris:
@Ambrevar I believe this is ready to go?

Ambrevar:
Merged!  Thanks a lot, @BlueFlo0d, this is a very valuable contribution!

PR #2002, opened 2022-01-07
WIP: Quick extension installation

aartaka:
This is two simple commands to install Nyxt extensions and WeExtensions. First one works just fine, second one doesn't work at all xD I'm putting it here to remember finishing it and to have your feedback on how it should work in the end.

# Things to discuss
- We used to have git-related commands. Should we resurrect those? Maybe have some way to conveniently call shell commands? UIOP is fine so far, though.
- Should we check the pages one opens in Nyxt and suggest to install the Nyxt extension if one is on the page that matches the extension one?

### DIFF
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -755,3 +755,29 @@ Requirements:
                                    (nyxt-init-file ,name))))
           (uiop:subprocess-error ()
             (echo-warning "Cannot install extension: failed to clone git repository"))))))
+
+(define-command install-web-extension (&optional (url (url (current-buffer))))
+  "Install the Nyxt extension from the URL.
+
+Requirements:
+- Only addons.mozilla.org is allowed for extension installation.
+- The page with the extension should be loaded in the current buffer.
+- ? should be in the PATH.
+- Repository name and ASDF system names should be the same."
+  (when (str:s-member (quri:uri-host url) "addons.mozilla.org")
+    (let ((download-url (plump:get-attribute
+                         (elt (clss:select ".InstallButtonWrapper-download-link"
+                                (document-model (current-buffer)))
+                              0)
+                         "href")))
+      (hooks:add-hook (after-download-hook *browser*)
+                      (make-handler-download
+                       (lambda (download)
+                         (when (string= download-url (url download))
+                           ;; TODO: What is the most accessible archive
+                           ;; extractor supporting xpi archives?
+                           (uiop:run-program '())
+                           (hooks:remove-hook (after-download-hook *browser*)
+                                              'download-web-extension-handler)))
+                       'download-web-extension-handler))
+      (download (current-buffer) download-url))))
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -757,7 +757,7 @@ Requirements:
             (echo-warning "Cannot install extension: failed to clone git repository"))))))
 
 (define-command install-web-extension (&optional (url (url (current-buffer))))
-  "Install the Nyxt extension from the URL.
+  "Install the Firefox WebExtensions-compliant addon from the URL.
 
 Requirements:
 - Only addons.mozilla.org is allowed for extension installation.
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -764,7 +764,7 @@ Requirements:
 - The page with the extension should be loaded in the current buffer.
 - ? should be in the PATH.
 - Repository name and ASDF system names should be the same."
-  (when (str:s-member (quri:uri-host url) "addons.mozilla.org")
+  (when (string= (quri:uri-host url) "addons.mozilla.org")
     (let ((download-url (plump:get-attribute
                          (elt (clss:select ".InstallButtonWrapper-download-link"
                                 (document-model (current-buffer)))
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -725,3 +725,33 @@ System information is also saved into the clipboard."
                              (:p (:pre (documentation (fn command) t)))
                              (:pre :class "nyxt-source" (:code (let ((*print-case* :downcase))
                                                                  (write-to-string (sexp command)))))))))))
+
+(define-command install-lisp-extension (&optional (url (url (current-buffer))))
+  "Install the Nyxt extension from the git repository at URL.
+
+Requirements:
+- git should be in the PATH.
+- Repository name and ASDF system names should be the same."
+  (let* ((name (alex:lastcar (str:split "/" (quri:uri-path url))))
+         (install-path (str:concat (expand-path *extensions-path*) name))
+         (config-path (make-instance 'init-data-path
+                                     :dirname (dirname *init-file-path*)
+                                     :basename name)))
+    (if (uiop:directory-exists-p install-path)
+        (echo "This extension is already installed at ~s" install-path)
+        (handler-case
+            (progn
+              (uiop:run-program
+               `("git" "clone" "--recurse-submodules" ,(render-url url) ,install-path))
+              (alex:write-string-into-file
+               (format nil "(in-package :nyxt-user)
+
+;; Put your configuration for ~a here." name)
+               (expand-path config-path)
+               :if-does-not-exist :create
+               :if-exists nil)
+              (nyxt::append-configuration
+               `(load-after-system ,(intern (str:upcase name) :keyword)
+                                   (nyxt-init-file ,name))))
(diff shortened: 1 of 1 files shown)


source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
                              (:p (:pre (documentation (fn command) t)))
                              (:pre :class "nyxt-source" (:code (let ((*print-case* :downcase))
                                                                  (write-to-string (sexp command)))))))))))
+
+(define-command install-lisp-extension (&optional (url (url (current-buffer))))
+  "Install the Nyxt extension from the git repository at URL.
```
Ambrevar, line 730: `git` -> Git.

Should it be "Nyxt extension" or "Lisp extension"?  I suggest "Lisp extension for Nyxt".

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+
+Requirements:
+- git should be in the PATH.
```
Ambrevar, line 733: "in the PATH' is too geeky.
"Git should be installed."

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+- Repository name and ASDF system names should be the same."
```
Ambrevar, line 734: Why?  This is unfortunately not true for many repos.  Can't we just add a keyword arg to specify the system name?  Better: parse the .asd!

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+  (let* ((name (alex:lastcar (str:split "/" (quri:uri-path url))))
+         (install-path (str:concat (expand-path *extensions-path*) name))
+         (config-path (make-instance 'init-data-path
+                                     :dirname (dirname *init-file-path*)
+                                     :basename name)))
+    (if (uiop:directory-exists-p install-path)
+        (echo "This extension is already installed at ~s" install-path)
```
Ambrevar, line 741: May not be installed, maybe it's just an empty dir.  Rather "Target directory ~s already exists, not installing."

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+
+;; Put your configuration for ~a here." name)
+               (expand-path config-path)
+               :if-does-not-exist :create
+               :if-exists nil)
+              (nyxt::append-configuration
+               `(load-after-system ,(intern (str:upcase name) :keyword)
+                                   (nyxt-init-file (str:concat ,name ".lisp")))))
+          (uiop:subprocess-error ()
+            (echo-warning "Cannot install extension: failed to clone git repository"))))))
```
Ambrevar, line 757: "Git"

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+         (install-path (str:concat (expand-path *extensions-path*) name))
+         (config-path (make-instance 'init-data-path
+                                     :dirname (dirname *init-file-path*)
+                                     :basename name)))
+    (if (uiop:directory-exists-p install-path)
+        (echo "This extension is already installed at ~s" install-path)
+        (handler-case
+            (progn
+              (uiop:run-program
+               `("git" "clone" "--recurse-submodules" ,(render-url url) ,install-path))
```
Ambrevar, line 745: References to external executables should always be `defvar`ed so that package maintainers can patch the paths (e.g. in Nix and Guix).  See `*xdg-open-program*` for instance.

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+               :if-does-not-exist :create
+               :if-exists nil)
+              (nyxt::append-configuration
+               `(load-after-system ,(intern (str:upcase name) :keyword)
+                                   (nyxt-init-file (str:concat ,name ".lisp")))))
+          (uiop:subprocess-error ()
+            (echo-warning "Cannot install extension: failed to clone git repository"))))))
+
+(define-command install-web-extension (&optional (url (url (current-buffer))))
+  "Install the Firefox WebExtensions-compliant addon from the URL.
```
Ambrevar, line 760: "add-on"

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+
+Requirements:
+- Only addons.mozilla.org is allowed for extension installation.
+- The page with the extension should be loaded in the current buffer.
+- ? should be in the PATH.
```
Ambrevar, line 765: What?

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+- Repository name and ASDF system names should be the same."
```
Ambrevar, line 766: Same as above.

source/help.lisp
```diff
@@ -725,3 +725,59 @@ System information is also saved into the clipboard."
...
+                                (document-model (current-buffer)))
+                              0)
+                         "href")))
+      (hooks:add-hook (after-download-hook *browser*)
+                      (make-handler-download
+                       (lambda (download)
+                         (when (string= download-url (url download))
+                           ;; TODO: What is the most accessible archive
+                           ;; extractor supporting xpi archives?
+                           (uiop:run-program '())
```
Ambrevar, line 779: Apparently they are just renamed ZIP files, so you can probably use whatever CL zip library to extract them.

Ambrevar: Guix has at least these two, I've never tested them:
- http://method-combination.net/lisp/chipz/
- https://www.xach.com/lisp/salza2/

source/help.lisp
aartaka: I thought about going down that path, and even started implementing it. But then, all kinds of stuff end up in .asd file, how do you know that a certain `defsystem` is the one you're looking for?

Restricting the kind of installable extensions is much easier than inferring what others' code means, but we can try doing that, yes.

source/help.lisp
aartaka: Indeed. The hostname of their extension catalogue is confusing :D

source/help.lisp
aartaka: I haven't yet written it down. See the comment on proper archiving utility below.

source/help.lisp
aartaka: Wow, and `chipz` is even a dependency on Nyxt!

source/help.lisp
Ambrevar: Modern ASDF files are meant to have only one main system and then subsystem.  The systems are named "MAINSYSTEM/SUBSYSTEM".

What you can do is load the asd (e.g. with `asdf:load-asd`), then to list the systems of the file you can do this (is there better?):

```lisp
(delete-if
 (lambda (system)
   (uiop:pathname-equal
    "/path/to/serapeum.asd"
    (asdf:system-source-file
     (asdf:find-system system))))
 (asdf:registered-systems))

; => ("serapeum" "serapeum/tests" "serapeum/docs")
```

Then get the primary system with

```lisp
(first (delete-if (complement #'asdf:primary-system-p) *))
```

aartaka: Wow, there's a lot of forethought in ASDF :D

BlueFlo0d:
It sounds to me like we need a proper package manager a la ELPA?

aartaka:
> It sounds to me like we need a proper package manager a la ELPA?

Yes, we do. There even was an issue about that, although I cannot find it now. This package manager will most probably depend on Quicklisp, CLPM, or something else, because there's no reason to not reuse good work done by others :)

This PR is a temporary solution to ease things a little. It's unclear when we'll have proper package manager, after all :upside_down_face:

Ambrevar:
Here is the issue about CLPM: https://github.com/atlas-engineer/nyxt/issues/1065

aartaka:
Closing this until we have a better way (= a way that doesn't involve unreliable scripts) to install extensions.

Ambrevar:
Hmmm, in my opinion it was better than nothing.  We could still merge this for 3.0, and wait for 4.0 to have a full-blown package manager.  Thoughts?

aartaka:
Hmmmmmmm. I'd say yes to merging the Lisp-extensions installation part of this, and thinking some more on the WebExtensions-related one, as it's too barebones and our implementation may change at any moment.

Ambrevar:
OK!

PR #2003, opened 2022-01-07
Fix "can't open-code test" SBCL warnings

Ambrevar:
This should finally rid us of the warnings in the line of

```log
source/buffer.lisp:856:1:
  note:
    can't open-code test of unknown type BROWSER
```

that started occurring with SBCL 2.1.9 at least.
You can trigger these warnings by force-compiling Nyxt or simply by changing branch (between `master` and `2-series` for instance).

I like the new `:module` in the .asd: it structures the list and make us less prone to break the load order by introducing new files and re-ordering files.

The only "fix" that I'm a bit skeptical about is the class forward declaration in f97d8fc612f059295b87de2106a47088bb6edb22.  What do you think?

### DIFF
--- a/libraries/history-tree/history-tree.lisp
+++ b/libraries/history-tree/history-tree.lisp
@@ -142,24 +142,6 @@ If the node has no owner, return Epoch."
                      (alex:hash-table-values (bindings node))))
       (local-time:unix-to-timestamp 0)))
 
-(export-always 'data-last-access)
-(declaim (ftype (function (history-tree t) local-time:timestamp) data-last-access))
-(defun data-last-access (history data)
-  "Return data last access across all its nodes, regardless of the owner.
-Return Epoch if DATA is not found or if entry has no timestamp."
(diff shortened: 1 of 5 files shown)


Ambrevar:
Actually the forward-declaration of class triggers a warning on CCL, so that's not the proper fix.  Any better idea?

aartaka:
Hmmmm, what about exporting a symbol in advance, so that it exists and has a type declared for it?

Ambrevar:
Doesn't work because the type is still unknown.

With no better idea, I suggest to merge this without the last commit.

aartaka:
> Doesn't work because the type is still unknown.

Oh, okay :)

> With no better idea, I suggest to merge this without the last commit.

Yes, sounds sane.

aartaka:
Thanks, Pierre, these were annoying :)

Ambrevar:
2 of them remain because we need to either forward declare the browser,
or untangle our window / buffer / browser definitions.

Issue #2009, opened 2022-01-10
Allow toggling status-bar and message-area


aartaka:
Commands toggling status bar and message area could come in handy for those that want to maximize screen real estate. They can help our internal use too: right now, when Nyxt enters fullscreen mode, status bar and message area are still there, which is not nice for a fullscreen mode. Thus, we should ideally toggle those off on fullscreen and toggle them back on on unfullscreen.

The signatures for these commands could be 
``` lisp
(define-command-global toggle-status-bar
    (&key (window (current-window)) (enable-p nil enable-provided)
     (height (height (status-buffer window)))))

(define-command-global toggle-message-area
    (&key (window (current-window)) (enable-p nil enable-provided) height))
```

aartaka:
By status bar i mean what is `status-buffer` in our code and modeline in common sense — the place where URL, modes etc. are displayed.

By message area I mean the blank area below status bar where messages are echoed.

Both of those are below the prompt buffer when it appears, yes.

I'm not sure they should re-appear when prompt buffer is invoked, if they were disabled prior to that. Why would one need those while in prompt-buffer?

guojing0:
To give an example, when one uses the spell checker, the result would `echo` on the message area.

aartaka:
Yes, that is tricky. But then, if one disables those, they'd probably understand what they're doing. It works much the same way as Emacs users doing `(scroll-bar-mode -1)` to disable a (relatively useful) scrollbar to get some more screen space.

guojing0:
Well, I am in the group thinking that "people sometimes may do things but later do not realize what they have done". In other words, I usually perfer presenting the least confusing option(s) to users, advanced or not. I will think about this when implementing said features.

aartaka:
I mean, we're not going to force this on anyone, as disabling interface elements leads to information loss. The only case where I'd suggest actually disabling status bar and message area is the one I mentioned in the original post -- fullscreening page/video.

In all the other cases, it's user's conscious choice to disable/enable those.

guojing0:
Oh, my bad. I must have missed this part ("fullscreening page/video").

aartaka:
> EDIT: One usually does not interact much with Emacs sccroll bar, in that it only tells the user limited information, i.e., where they are in the buffer.

[...]

PR #2012, opened 2022-01-11
Separate commands for toggling status and message buffers

guojing0:
This is to resolve Issue https://github.com/atlas-engineer/nyxt/issues/2009 by introducing `toggle-status-buffer` and `toggle-message-buffer` commands.


aartaka:
CI breaks because removed `(un)present-current-window` is used elsewhere :)

A question to test the use-case: what if I want to make one command call to disable the status buffer? Even if it's disabled. What should I do then? That's why I suggested adding `enable-p` keyword argument to the `toggle-status-bar` in #2009: one may want to disable things unconditionally of thier current state.

source/window.lisp
```diff
@@ -237,20 +237,19 @@ mapped to query parameters."
...
+(define-command toggle-status-buffer (&optional (window (current-window)))
+  "Toggle the visibility of the status buffer."
+  (if (zerop (ffi-window-get-status-buffer-height window))
+      (ffi-window-set-status-buffer-height window (height (status-buffer window)))
+      (ffi-window-set-status-buffer-height window 0)))
+
+(define-command toggle-message-buffer (&optional (window (current-window)))
+  "Toggle the visibility of the message buffer."
+  (if (zerop (ffi-window-get-message-buffer-height window))
+      (ffi-window-set-message-buffer-height window (height (message-buffer-height window)))
```
aartaka, line 254: What does `(height (message-buffer-height window))` returns?

aartaka:
To rephrase my chaotic questions, here's a list of distinct API actions one should have with status bar and message area:
- Enable status bar.
- Disable status bar.
- Toggle status bar.
- Enable message area.
- Disable message area.
- Toggle message area.
- Enable both.
- Disable both.
- Toggle both.

The most ergonomic way that I can think of is three commands with keyword arguments from #2009, but I may be mistaken.

guojing0:
I vaguely remember that we may have a macro that automatically generates `enable-*`, `disable-*`, and `toggle-*` commands?

aartaka:
> I vaguely remember that we may have a macro that automatically generates `enable-*`, `disable-*`, and `toggle-*` commands?

I don't remember of such a macro, but we don't need it, actually: we only need some three commands and several helper functions.

aartaka:
The way I'd do it is close to what I suggested in #2009 :)

guojing0:
> what if I want to make one command call to disable the status buffer? Even if it's disabled already.

Sorry I do not follow. Why would one do that?

aartaka:
> > what if I want to make one command call to disable the status buffer? Even if it's disabled already.
> 
> Sorry I do not follow. Why would one do that?

This might be useful when writing configuration snippets and to replace `(un)present-current-window`.
- Existing use-case: we need to unconditionally hide both aread when fullscreening a video and unconditionally show those when unfullscreening.
- Future use-case: one might want to have a minimum distraction mode where it unconditionally disables everything but the web page. Even if other mode have already disabled e.g. status buffer, it should still keep it disabled, instead of toggling.

> I agree with `enable-p`. To understand the desired behavior of `enable-p`, if it's set to `T`, then that area is hidden?

If set to `T`, then show; if set to `NIL`, then hide; if not provided, toggle. Actually, it might be better to name it `show-p`. What do you think?

guojing0:
Thank you for the explanation, Artyom. It's really useful to understand better.

However, given current implementation, I do not think we need that many APIs. Say, if an user only wants to permanently disable status buffer, then he can just add something like `(toggle-status-buffer :show-p nil)` in his config file.

source/window.lisp
```diff
@@ -239,18 +239,57 @@ mapped to query parameters."
...
+(defun disable-status-buffer (&optional (window (current-window)))
+  (ffi-window-set-status-buffer-height window 0))
+
+(defun enable-message-buffer (&optional (window (current-window)))
+  (ffi-window-set-message-buffer-height window (message-buffer-height window)))
+
+(defun disable-message-buffer (&optional (window (current-window)))
   (ffi-window-set-message-buffer-height window 0))
 
+(define-command toggle-status-buffer (&optional (window (current-window))
```
aartaka, line 257: SBCL complains about `&optional` and `&key` arguments being present in the same arglist. Mixing the two confuses both human readers and compilers. 

In this particular case, `&key` would work for both args.

source/window.lisp
```diff
@@ -239,18 +239,57 @@ mapped to query parameters."
...
+                                      &key (show-p nil show-provided-p))
+  "Toggle the visibility of the status buffer by default.
```
aartaka, line 259: The "by default" is not necessary here, given that you explain `show-p` later.

source/window.lisp
```diff
@@ -239,18 +239,57 @@ mapped to query parameters."
...
+
+If SHOW-P is provided:
+- If SHOW-P is T, then `status-buffer' is always enabled;
```
aartaka, line 262: "If SHOW-P is non-nil", maybe?

Then the second clause becomes "If SHOW-P is nil...".

source/window.lisp
```diff
@@ -239,18 +239,57 @@ mapped to query parameters."
...
+- Otherwise, it is always disabled."
+  (if show-provided-p
```
aartaka, line 264: These nested `if` don't look right. How about a `cond`? Something like:
``` lisp
(cond
  ((and show-provided-p show-p)
   (enable-status-buffer window))
  ((and (not show-provided-p)
        (zerop (ffi-window-get-status-buffer-height window)))
   (enable-status-buffer window))
  (t (disable-status-buffer window)))
```

source/window.lisp
```diff
@@ -239,18 +239,57 @@ mapped to query parameters."
...
+      (if show-p
+          (enable-status-buffer window)
+          (disable-status-buffer window))
+      (if (zerop (ffi-window-get-status-buffer-height window))
+          (enable-status-buffer window)
+          (disable-status-buffer window))))
+
+(define-command toggle-message-buffer (&optional (window (current-window))
+                                       &key (show-p nil show-provided-p))
+  "Toggle the visibility of the message buffer by default.
```
aartaka, line 274: Same as above.

aartaka (changes requested):
Looks good and comprehensive enough! There are small things that I'd like to see resolved before we merge, though :)

aartaka:
> However, given current implementation, I do not think we need that many APIs, unless I misunderstood your intention: Say, if an user only wants to permanently disable status buffer, then he can just add something like `(toggle-status-buffer :show-p nil)` in his config file.

That's exactly what I meant -- adding command behavior altering keyword arguments qualifies as API expansion :)

guojing0:
Many thanks. In this case, do you think `(un)present-current-window` can be removed? Or, you have some suggestions for a better name, as someone whose second language is English, when I first saw this it's quite confusing.

source/window.lisp
guojing0: From the language perspective, "non-nil" is more correct, but would it be more clear if we "only allow" `T` and `NIL`?

aartaka:
> Many thanks. In this case, do you think `(un)present-current-window` can be removed? Or, you have some suggestions for a better name, as someone whose second language is English, when I first saw this it's quite confusing.

I was confused too :D 

Yes, I believe it can be removed. Just use the matching `toggle-message-buffer` and `toggle-status-buffer` instead in `"fullscreen"` and `"unfullscreen"` signals in `renderer/gtk.lisp`

PR #2013, opened 2022-01-11
libraries/web-extensions: List objects automatically in Makefile.

Ambrevar:


Ambrevar:
Actually, what if you moved `-fPIC -shared` to the LDFLAGS, and then remove the `libnyxt.so` rule, Make should handle the rest automatically via the default rules, nah?

aartaka:
> Actually, what if you moved `-fPIC -shared` to the LDFLAGS, and then remove the `libnyxt.so` rule, Make should handle the rest automatically via the default rules, nah?

I don't see an implicit rule for .so files in [Implicit Rules Catalogue](https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html#Catalogue-of-Rules)...

libraries/web-extensions/Makefile
```diff
@@ -10,7 +10,7 @@ CC := ${shell command -v ${CC} 2> /dev/null || echo gcc}
 CFLAGS += ${shell pkg-config ${PKG_MODULES} --cflags} -fPIC
 LDLIBS += ${shell pkg-config ${PKG_MODULES} --libs}
 
-objects = nyxt.o globals.o extevent.o browser.o tabs.o management.o runtime.o extension.o storage.o alarms.o bookmarks.o browser_action.o commands.o history.o notifications.o permissions.o web_navigation.o web_request.o
+objects = $(patsubst %.c,%.o,$(wildcard *.c))
```
aartaka, line 13: The rest of the Makefile uses `${}` syntax, using `$()` is inconsistent :)

aartaka:
Yes, it does work!

aartaka:
I mean that this `$(patsubst %.c,%.o,$(wildcard *.c))` spell works.

libraries/web-extensions/Makefile
aartaka, line 13: Should `objects` be capitalized? What are the conventions for Makefile variable naming, actually?

Ambrevar: Sorry about that,  but actually this makefile deviates from the syntax of nyxt/Makefile :)

Ambrevar: I'm unsure, but I believe uppercase is for overridable environment variables, so keeping this lowercase is the right thing to do as far as I can tell.

Ambrevar:
> I don't see an implicit rule for .so files in [Implicit Rules Catalogue](https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html#Catalogue-of-Rules)... 

```sh
$ make -p | grep '.so'
make: *** No targets specified and no makefile found.  Stop.
# This is free software: you are free to change and redistribute it.
...
.LIBPATTERNS = lib%.so lib%.a
```

There is .LIBPATTERNS, but I don't know how it's used, so I guess it
won't work.

OK to merge then?

libraries/web-extensions/Makefile
aartaka: Yes, that's sane :)

aartaka:
OK to merge!

libraries/web-extensions/Makefile
aartaka: I'll change it after we merge then xD

PR #2015, opened 2022-01-13
Fix to not remove highlighting of search terms

taiju:
Hi,

I found what I thought was a bug and fixed it.

# Steps to Reproduce

1. Open https://nyxt.atlas.engineer/ in Nyxt.
2. Execute the `search-buffer` command to search for "nyxt".
3. Execute `visual-mode` command.

# Expected behavior

Able to use visual-mode, including search terms.

# Actual behavior

Search terms are removed when switching to visual-mode.

If we use visual mode after searching in a document, you may want to select a string of search terms, so it would be inconvenient if it is removed.

Therefore, I changed the CSS selector expression so that search term highlights are not targeted for removal by `remove-element-hints`.

Nyxt is a great browser and I really like it. Thanks again.

### DIFF
--- a/source/mode/element-hint.lisp
+++ b/source/mode/element-hint.lisp
@@ -80,7 +80,7 @@
 (define-parenscript remove-element-hints ()
   (defun hints-remove-all ()
     "Removes all the elements"
-    (ps:dolist (element (nyxt/ps:qsa document ".nyxt-hint"))
+    (ps:dolist (element (nyxt/ps:qsa document ":not(.nyxt-search-node) > .nyxt-hint"))
       (ps:chain element (remove))))
   (hints-remove-all))
 


aartaka:
You're right in saying that invoking `visual-mode` (or any other mode) should not remove the text. However, the fix would cause problems for our search facility, making it unable to clean up all the traces of its presence on the page. 

I suggest you open an issue and we discuss it there, as the fix is likely to end up being much less trivial than this PR :)

taiju:
I see. So this fix causes a problem.
I added an issue.  #2020

aartaka:
> I added an issue. #2020

Thanks you!

aartaka:
On a second thought, your fix is fine, as it separates search hints and element hints, which is a desireable thing for your use-case. I don't see any problem with it now that you expanded on it in #2020. Merging this until we find a way to clearly delineate search and element hinting.

Thanks for bringing this up!

Issue #2016, opened 2022-01-13
lisp-repl command does not work properly


shamazmazum:
Hi! My nyxt version is #<QL-DIST:RELEASE atlas-engineer-nyxt-20220113114424 / ultralisp 20220113114500>.

`lisp-repl` fails to switch to a new buffer with REPL. That's because `ffi-buffer-load` tries to access `status` slot of a buffer, but `internal-buffer` does not have this slot. Moving this slot to `buffer` just causes more problems.

I checked that `make-internal-buffer` is only used in three places: `source/mode/os-package-manager.lisp`, `source/mode/diff.lisp` and `source/mode/repl.lisp`. Are internal buffers deprecated? What's the alternative? `define-internal-page-command`, maybe? I don't know how to fix this properly. Again, sorry, if this is a known bug :)

aartaka:
I can reproduce, and yes, internal buffer should not be there! We need to rewrite `lisp-repl` using the new `define-internal-page-command`, yes.

shamazmazum:
Sent a PR. BTW, is it a good idea to export `define-internal-page-command`? It seems to be used in many places outside of `nyxt` package.

aartaka:
> Sent a PR.

Wow, that was fast ＼（＾▽＾）／

> BTW, is it a good idea to export `define-internal-page-command`? It seems to be used in many places outside of `nyxt` package.

Yes, exporting it is the right move, I'm sure :)

PR #2017, opened 2022-01-13
Fix repl-mode.

shamazmazum:
Do not use MAKE-INTERNAL-BUFFER, use DEFINE-INTERNAL-PAGE-COMMAND
instead (see #2016).

Repl-mode seems to start correctly again, but I removed `base-mode` from a new buffer (because you can choose only one mode in `define-internal-page-command`, as it seems). I cannot see the difference. Is this acceptable?

### DIFF
--- a/source/mode/repl.lisp
+++ b/source/mode/repl.lisp
@@ -54,11 +54,7 @@
           :documentation "The CSS applied to a REPL when it is set-up.")
    (evaluation-history (list)
                        :documentation "A list of pairs of (INPUT RESULTS).
-INPUT is a string and RESULTS is a list of Lisp values.")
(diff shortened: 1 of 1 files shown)


aartaka:
Thanks for this prompt fix! I'll review it tomorrow :)

> Repl-mode seems to start correctly again, but I removed `base-mode` from a new buffer (because you can choose only one mode in `define-internal-page-command`, as it seems). I cannot see the difference. Is this acceptable?

`base-mode` is enabled by default for all the buffers Nyxt creates, removing it changes nothing in this case.

aartaka:
Thanks for the fix, @shamazmazum!

PR #2022, opened 2022-01-16
mode/auto,urls: Add match-port URL designator predicate

hendursaga:
Real simple patch, the only thing that might need changing would be `#'eq` but it appears to work just fine.

### DIFF
--- a/source/mode/auto.lisp
+++ b/source/mode/auto.lisp
@@ -485,6 +485,8 @@ Auto-mode is re-enabled once the page is reloaded."
 ;; - match-regex works for any address that matches a given regex. You can add these manually,
 ;;   but remember: with great regex comes great responsibility!
 ;;   Example: (match-regex \"https://github\\.com/.*/.*\") will activate only in repos on GitHub.
+;; - match-port matches the port number(s) only.
+;;   Example: (match-port 8000 8080) will work for e.g. localhost:8000, 127.0.0.1:8080, etc.
 ;; - String format matches the exact URL and nothing else
 ;;   Example: \"https://lispcookbook.github.io/cl-cookbook/pattern_matching.html\"
 ;;            will work on the Pattern Matching article of CL Cookbook, and nowhere else.
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -333,6 +333,15 @@ return a boolean.  It defines an equivalence relation induced by EQ-FN-LIST.
         (some (alex:curry #'string= (quri:uri-domain (url url-designator)))
               (cons domain other-domains)))))
 
+(-> match-port (integer &rest integer) (function (quri:uri) boolean))
+(export-always 'match-port)
+(defun match-port (port &rest other-ports)
+  "Return a predicate for URL designators matching one of PORT or OTHER-PORTS."
+  #'(lambda (url-designator)
+      (when url-designator
+        (some (alex:curry #'eq (quri:uri-port (url url-designator)))
+              (cons port other-ports)))))
+
 (-> match-file-extension (string &rest string) (function (quri:uri) boolean))
 (export-always 'match-file-extension)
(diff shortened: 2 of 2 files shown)


source/mode/auto.lisp
```diff
@@ -485,6 +485,8 @@ Auto-mode is re-enabled once the page is reloaded."
 ;; - match-regex works for any address that matches a given regex. You can add these manually,
 ;;   but remember: with great regex comes great responsibility!
 ;;   Example: (match-regex \"https://github\\.com/.*/.*\") will activate only in repos on GitHub.
+;; - match-port matches the port number(s) only.
+;;   Example: (match-port 8000 8080) will work for e.g. localhost:8000, 127.0.0.1:8080, etc.
```
aartaka, line 489: But then it's not only about localhost, right? Maybe include an example for a wider web, like `fsf.org:8000`?

"e.g." should be surrounded by commas, as it's synonymous to "for example", AFAIK :)

source/urls.lisp
```diff
@@ -333,6 +333,15 @@ return a boolean.  It defines an equivalence relation induced by EQ-FN-LIST.
         (some (alex:curry #'string= (quri:uri-domain (url url-designator)))
               (cons domain other-domains)))))
 
+(-> match-port (integer &rest integer) (function (quri:uri) boolean))
+(export-always 'match-port)
+(defun match-port (port &rest other-ports)
+  "Return a predicate for URL designators matching one of PORT or OTHER-PORTS."
```
aartaka, line 339: Looking at this docstring (and docstrings of other `match-*` functions, they are not that helpful, aren't they? I mean, `predicate` and `matching` do say something, but we might do a better documentation job here. How about rephrasing the `match-port` docstring to something along the lines of

```
Return a predicate returning non-nil for URL designator matching one of PORT or OTHER-PORTS, and nil otherwise.

Can be used in `auto-mode' rules.
```

source/mode/auto.lisp
```diff
@@ -485,6 +485,8 @@ Auto-mode is re-enabled once the page is reloaded."
...
 ;; - String format matches the exact URL and nothing else
 ;;   Example: \"https://lispcookbook.github.io/cl-cookbook/pattern_matching.html\"
 ;;            will work on the Pattern Matching article of CL Cookbook, and nowhere else.
```
aartaka, line 492: Having our IRC discussion, I now think that we should mention the Turing-completeness of `auto-mode` rules in the file comment. But that can wait, I guess :)

aartaka (commented):
I have some documentation suggestions, even though it's perfectly fine to leave it as it is :)

source/mode/auto.lisp
Ambrevar: Actually I recommend we try to stay clear from academic lingo and instead recommend the use of full English words.  So prefer "for example" or "for instance" to "e.g.".

source/urls.lisp
hendursaga: I was only ~~stealing~~ copying from the other functions' docs! :stuck_out_tongue: 

That does bring up a point, though, that any suggestions/changes should also apply to the others. In that case, perhaps it should be done in a separate PR or something? I was thinking of doing a little refactoring, after all..

And yes, come to think of it, I think one example should be something non-local, I just couldn't think of any off the top of my head when I wrote that. I do prefer common abbreviations like, e.g. "e.g.," "i.e.," etc. I prefer the British style of being lenient concerning commas after the "e.g."

Can we merge after I push and squash a few minor changes?

aartaka: Sure it does apply to us, we can rephrase things if you don't want to. 

Yes, this PR is fine to merge.

aartaka:
@hendursaga, thanks for your work, merging it :)

PR #2025, opened 2022-01-17
Make thread name mandatory in `run-thread'.

Ambrevar:
@jmercouris @pdelfino @aartaka Do you think it's a good idea?

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -498,24 +498,17 @@ raised condition."
                                args)))
             `(warn ,format-string ,@new-args))))))
 
-(defmacro run-thread (&body body)   ; TODO: Copied from Nyxt.  Move to serapeum?
+(defmacro run-thread (name &body body)   ; TODO: Copied from Nyxt.  Move to serapeum?
   "Run body in a new protected new thread.
 This is a \"safe\" wrapper around `bt:make-thread'."
-  ;; We parse the body instead of using a macro argument for backward compatibility.
-  (let* ((name (if (stringp (first body))
-                   (first body)
-                   "anonymous thread"))
-         (body (if (stringp (first body))
-                   (rest body)
-                   body)))
-    `(bt:make-thread
-      (lambda ()
-        ,(if *debug-on-error*
-             `(progn
-                ,@body)
-             `(with-protect ("Error on separate prompter thread: ~a" :condition)
-                ,@body)))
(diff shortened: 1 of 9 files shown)


libraries/prompter/prompter-source.lisp
```diff
@@ -498,16 +498,12 @@ raised condition."
...
   "Run body in a new protected new thread.
 This is a \"safe\" wrapper around `bt:make-thread'."
-  ;; We parse the body instead of using a macro argument for backward compatibility.
-  (let* ((name (if (stringp (first body))
-                   (first body)
-                   "anonymous thread"))
-         (body (if (stringp (first body))
-                   (rest body)
-                   body)))
+  (let ((body (if (stringp (first body))
```
aartaka, line 504: But why? We already have `name`, then everything in `body` is, well... body of the function?

Am I missing something?

source/browser.lisp
```diff
@@ -422,6 +422,7 @@ Return the download object matching the download."
                          (download-manager:filename download))
                    (push download-render (downloads *browser*))
                    (run-thread
+                     "Download watcher"
```
aartaka, line 425: What's the convention for thread names, actually? I see that we append `nyxt` in front of those, so it should be lowercase, but other than that, it's a total chaos xD

source/concurrency.lisp
```diff
@@ -70,17 +70,13 @@ This is a blocking operation."
...
 This supersedes `bt:make-thread' in Nyxt.  Don't use the latter unless you know
 what you are doing!"
-  ;; We parse the body instead of using a macro argument for backward compatibility.
-  (let* ((name (if (stringp (first body))
-                   (first body)
-                   "anonymous thread"))
-         (body (if (stringp (first body))
-                   (rest body)
-                   body)))
+  (let ((body (if (stringp (first body))
```
aartaka, line 77: Same as in prompter-source, why parse body if `name` is already parsed for us?

aartaka (commented):
I am confused by the `body` parsing and am not sure in understanding out thread naming conventions, but it's perfect otherwise :)

source/browser.lisp
Ambrevar: Hmmm, OK, we can put it all lower case.

Ambrevar:
I believe I've addressed all your points!

aartaka:
Yes, I see nothing problematic about this PR now :D

PR #2031, opened 2022-01-18
Lisp urls cleanup

### LINKED ISSUE #2027: Lisp URL related issues
I've noticed a few issues with Lisp URLs and  since the fix is probably common to all of them, let me list them here:

1. Looks like `lisp-url` should no longer be used.  Its only call site is `send-lisp-url`.  Move it there?

2. Actually `download.lisp` still uses `lisp-url` and as such all buttons are broken.

3. If I got it right, this means that `user-interface:button` can be removed.

4. The following snippet is faulty in `gtk.lisp`:

```lisp
;; Objects and other complex structures make cl-json choke.
(unless (or (typep result 'standard-object)
                                       (and (typep result 'cons)
                                            (some #'listp result)))
                             (cl-json:encode-json-to-string result))
```

It does not take provisions for non-cons sequences, nor for standard-objects nor structs in these sequences.  Something as simple as `(list (quri:quri "atlas.engineer"))` fails here.

Proposed solution:

```lisp
;; `sb-pcl:slot-object` has these 3 subclasses.
(defun slot-object-p (result)  (typep result '(or standard-object structure-object condition)))

; ...

(unless (or (slot-object-p result)
                                       (and (typep result 'sequence)
                                            (some #'slot-object-p  result)))
                             (cl-json:encode-json-to-string result))
```

@aartaka What do you think?

EDIT: Typos.

Ambrevar:
Fixes #2027.

@aartaka What do you think?

### DIFF
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -42,15 +42,16 @@ below to create said file, if it's not created yet.")
        (ensure-directories-exist init-file-path)
        (:p (if (uiop:file-exists-p init-file-path)
                (:a :class "button"
-                   :href (lisp-url `(echo "Init file exists"))
+                   :href (ps:ps (nyxt/ps:send-lisp-url  `(echo "Init file exists")))
                    "Init file exists")
                (:a :class "button"
-                   :href (lisp-url `(with-open-file (_ ,init-file-path
-                                                       :direction :output
-                                                       :if-exists nil
-                                                       :if-does-not-exist :create)
-                                      (echo "Init file created at ~a."
-                                            ,init-file-path)))
+                   :href (ps:ps (nyxt/ps:send-lisp-url
+                                 `(with-open-file (_ ,init-file-path
+                                                     :direction :output
+                                                     :if-exists nil
+                                                     :if-does-not-exist :create)
+                                    (echo "Init file created at ~a."
+                                          ,init-file-path))))
                    "Create init file"))))
     (:p "Example:")
     (:pre (:code "
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -42,10 +42,10 @@ below to create said file, if it's not created yet.")
        (ensure-directories-exist init-file-path)
        (:p (if (uiop:file-exists-p init-file-path)
                (:a :class "button"
-                   :href (ps:ps (nyxt/ps:send-lisp-url  `(echo "Init file exists")))
+                   :href (ps:ps (nyxt/ps:lisp-eval  `(echo "Init file exists")))
                    "Init file exists")
                (:a :class "button"
-                   :href (ps:ps (nyxt/ps:send-lisp-url
+                   :href (ps:ps (nyxt/ps:lisp-eval
                                  `(progn (ensure-file-exists ,init-file-path)
                                          (echo "Init file created at ~a."
                                                ,init-file-path))))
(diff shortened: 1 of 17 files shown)


source/renderer/gtk.lisp
```diff
@@ -748,9 +766,9 @@ See `gtk-browser's `modifier-translator' slot."
                  (log:debug "Evaluate Lisp code from internal page: ~a" code)
                  (values (let ((result (first (evaluate code))))
                            ;; Objects and other complex structures make cl-json choke.
-                           (unless (or (typep result 'standard-object)
-                                       (and (typep result 'cons)
-                                            (some #'listp result)))
+                           (when (or (scalar-p result)
```
aartaka, line 769: While this is a perfect fix, I'm still unsure about both my initial implementation and your implementation here. 

I mean, it allows encoding "scalar" values and sequences of those. But
- Why not allow nested sequences of sequences of sequences of scalars?
- Should we actually ever return sequences?
- If we rely on `cl-json`, then maybe we should just allow any value it can parse? In this case, catching `cl-json:unencodable-value-error` (as I initially did) may actually be the best solution to this problem of `cl-json` compatibility.

Again, I'm not sure about how to approach that. Some part of me wants to restrict it all to literal `atom`s, but it's too restrictive to be useful. Encoding it to the format that `cl-json` would support sounds optimal, especially given that we can override the (en|de)coding methods and allow arbitraty objects (like buffers) in the `nyxt:` URL arguments.

source/manual.lisp
```diff
@@ -42,15 +42,16 @@ below to create said file, if it's not created yet.")
...
                    "Init file exists")
                (:a :class "button"
-                   :href (lisp-url `(with-open-file (_ ,init-file-path
-                                                       :direction :output
-                                                       :if-exists nil
-                                                       :if-does-not-exist :create)
-                                      (echo "Init file created at ~a."
-                                            ,init-file-path)))
+                   :href (ps:ps (nyxt/ps:send-lisp-url
+                                 `(with-open-file (_ ,init-file-path
```
aartaka, line 49: Quicklisp has this elegant pearl:
``` lisp
(defun ensure-file-exists (pathname)
  (open pathname :direction :probe :if-does-not-exist :create))
```

source/parenscript-macro.lisp
```diff
@@ -72,7 +72,8 @@
 (export-always 'send-lisp-url)
 (defpsmacro send-lisp-url (form &optional callback)
```
aartaka, line 73: Now that `lisp-url` is part of this function, it's no more than implementation detail. What `send-lisp-url` does now is evaluating arbitrary Lisp code. How about renaming it to something like `lisp-eval` then?

aartaka (commented):
Thanks for this patch, it should fix most of problems I didn't catch with nyxt/lisp URLs!

I have small stylistic/API suggestions :)

source/renderer/gtk.lisp
Ambrevar: All fine with me.  We can also merge this for now and keep a TODO comment in there.

source/manual.lisp
Ambrevar: Nice!

Ambrevar: But the benefit of the `with-open-file` form is that on failure it won't display the message.

source/parenscript-macro.lisp
Ambrevar: OK!

Ambrevar:
Updated, tell me what you think.

aartaka:
Looks perfect, feel free to merge :)

Thanks for taking the time to fix it!

PR #2034, opened 2022-01-19
Refactor file management into a separate mode.

### LINKED ISSUE #1219: Restore file-manager-mode and open-file as an extension (for now)

aartaka:
This puts all the Nyxt-specific file management utilities under the `nyxt/file-manager-mode` package and uses the newly re-introduced `file-manager-mode` as the auxiliary mode when prompting for files.

# What's New:
- `file-manager-mode` -- a prompt-buffer mode for file search. Not much there yet.
- `C-backspace` (CUA) and `C-l` (Emacs) strip one level of nesting from the currently inputted pathname.
- An ad-hoc search for executables to open the file with: `program-source`.

# Things to Discuss
How do we discover executables portably?

Closes #1219?

### DIFF
--- a/source/mode/file-manager.lisp
+++ b/source/mode/file-manager.lisp
@@ -159,8 +159,7 @@ See `supported-media-types' of `file-mode'."
                  (let* ((program (prompt1
                                    :prompt "The program to open the selected files with"
                                    :sources (list (make-instance 'user-program-source)))))
-                   (dolist (file files)
-                     (uiop:launch-program (list program (namestring file)))))))
+                   (uiop:launch-program (cons (uiop:native-namestring program) (mapcar #'uiop:native-namestring files))))))
          (slot-value source 'prompter:actions))))
 
 #+linux
--- a/source/mode/file-manager.lisp
+++ b/source/mode/file-manager.lisp
@@ -77,7 +77,7 @@
   "Return non-nil if FILE is executable for USER name.
 When the user is unspecified, take the current one."
   (sera:true
-   ;; FIXME: Some files (like /usr/bin/[) cause IOLIB to choke on ENOENT.
+   ;; FIXME: Some files cause IOLIB to choke on ENOENT.
    (let ((permissions (ignore-errors (iolib/os:file-permissions file))))
      (or (and (string= (file-author file)
                        user)
--- a/source/mode/file-manager.lisp
+++ b/source/mode/file-manager.lisp
@@ -95,7 +95,7 @@ When the user is unspecified, take the current one."
      (remove-if
       #'uiop:hidden-pathname-p
       (mapcar #'uiop:resolve-symlinks
-              (alex:mappend (lambda (path) (uiop:directory-files path)) paths))))))
+              (alex:mappend #'uiop:directory-files paths))))))
 
 (define-class program-source (prompter:source)
   ((prompter:name "Programs")
--- a/source/mode/file-manager.lisp
+++ b/source/mode/file-manager.lisp
@@ -77,12 +77,11 @@
   "Return non-nil if FILE is executable for USER name.
 When the user is unspecified, take the current one."
   (sera:true
-   ;; FIXME: Some files cause IOLIB to choke on ENOENT.
-   (let ((permissions (ignore-errors (iolib/os:file-permissions file))))
-     (or (and (string= (file-author file)
-                       user)
+   (let* ((filename (uiop:native-namestring file))
+          (permissions (iolib/os:file-permissions filename)))
+     (or (and (string= (file-author file) user)
               (member :user-exec permissions))
-         (and (= (file-group-id file)
+         (and (= (file-group-id filename)
                  (group-id user))
               (member :group-exec permissions))
          (member :other-exec permissions)))))
--- a/source/mode/file-manager.lisp
+++ b/source/mode/file-manager.lisp
@@ -115,22 +115,26 @@ When the user is unspecified, take the current one."
   (lambda (pathname)
     (string-equal (pathname-type pathname) ext)))
 
-(defun make-file-source-preprocessor (&rest filters)
-  "Return a preprocessor that lists all files satisfying all FILTERS.
+(defun make-file-source-preprocessor ()
+  "Return a preprocessor that lists all files satisfying `extensions' and `allow-directories'.
 It's suitable for `prompter:filter-preprocessor'."
   (lambda (suggestions source input)
     (declare (ignore suggestions))
-    (let ((pathname (uiop:ensure-pathname (if (uiop:emptyp input)
-                                              *default-pathname-defaults*
-                                              input))))
+    (let* ((pathname (uiop:ensure-pathname (if (uiop:emptyp input)
+                                               *default-pathname-defaults*
+                                               input)))
+           (directory (uiop:pathname-directory-pathname pathname)))
       (prompter:filter-exact-matches
        (prompter:ensure-suggestions-list
         source
         (sera:filter
-         (apply #'alex:conjoin (or filters (list #'identity)))
-         (directory-elements (if (uiop:directory-pathname-p pathname)
-                                 pathname
-                                 (uiop:pathname-directory-pathname pathname)))))
+         (lambda (path)
+           (or (and (uiop:directory-pathname-p path)
+                    (allow-directories source))
+               (and (uiop:file-pathname-p path)
+                    (or (null (extensions source))
+                        (str:s-member (extensions source) (pathname-type path))))))
+         (directory-elements directory)))
        source
        input))))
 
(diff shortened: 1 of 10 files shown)


Ambrevar:
> How do we discover executables portably?

You mean `sera:resolve-executable`?

source/mode/file-manager.lisp
```diff
@@ -107,7 +152,14 @@ See `supported-media-types' of `file-mode'."
...
                               :new-buffer-p t
-                              :supported-p (supported-media-or-directory file source))))))
+                              :supported-p (supported-media-or-directory file source)))))
+               (make-command open-with* (files)
+                 "Open the file with selected program."
+                 (let* ((program (prompt1
+                                   :prompt "The program to open the selected files with"
+                                   :sources (list (make-instance 'user-program-source)))))
+                   (dolist (file files)
+                     (uiop:launch-program (list program (namestring file)))))))
```
Ambrevar, line 162: In my opinion a more familiar behaviour is to open the files with the same program.  So remove `dolist` and call `uiop:launch-program` once over `files`.

Ambrevar:
Oh, I understand, you want to _list_ executables_ in `$PATH`, right?

This is what Emacs Helm does:

```elisp
(defun helm-external-commands-list-1 (&optional sort)
  "Returns a list of all external commands the user can execute.
If `helm-external-commands-list' is non-nil it will return its
contents.  Else it calculates all external commands and sets
`helm-external-commands-list'."
  (helm-aif helm-external-commands-list
      it
    (setq helm-external-commands-list
          (cl-loop
                for dir in (split-string (getenv "PATH") path-separator)
                when (and (file-exists-p dir) (file-accessible-directory-p dir))
                for lsdir = (cl-loop for i in (directory-files dir t)
                                  for bn = (file-name-nondirectory i)
                                  when (and (not (member bn completions))
                                            (not (file-directory-p i))
                                            (file-executable-p i))
                                  collect bn)
                append lsdir into completions
                finally return
                (if sort (sort completions 'string-lessp) completions)))))
```

source/mode/file-manager.lisp
```diff
@@ -26,6 +53,23 @@
                             (append files included-directories)
                             files))))
 
+(export-always 'recursive-directory-elements)
+(defun executables ()
+  (let ((paths (serapeum::$path))) ; FIXME: It's not an exported API.
+    ;; FIXME: Check for executability? A lot of junk without it.
+    ;; StumpWM uses non-portable (sb-unix:unix-access filename sb-unix:x_ok).
+    (alex:mappend (lambda (path) (uiop:directory-files path)) paths)))
```
Ambrevar, line 61: I recommend you trim hidden files.  There is `uiop:hidden-pathname-p`, but that won't work on Windows.  OK for now though.

Ambrevar:
> This puts all the Nyxt-specific file management utilities under the nyxt/file-manager-mode 

This sounds like you've moved code, but you haven't, right?  Misunderstanding on my end?

source/mode/file-manager.lisp
aartaka: > > This puts all the Nyxt-specific file management utilities under the nyxt/file-manager-mode
> 
> This sounds like you've moved code, but you haven't, right? Misunderstanding on my end?

I did -- from source/file-manager.lisp to source/mode/file-manager.lisp.

This causes CI failures, I'll fix those tomorrow :)

aartaka:
> > This puts all the Nyxt-specific file management utilities under the nyxt/file-manager-mode
> 
> This sounds like you've moved code, but you haven't, right? Misunderstanding on my end?

I did -- from source/file-manager.lisp to source/mode/file-manager.lisp.

This causes CI failures, I'll fix those tomorrow :)

source/mode/file-manager.lisp
aartaka: Ooooops, wrong thread :p 

Yes, will do!

Ambrevar:
This is what we use in the code base:

```lisp
(member :user-exec (iolib/os:file-permissions "path/to/file"))
```

Ambrevar:
Probably a good time to add the `executable-p` function!

Ambrevar:
This function was much more difficult to implement than I expected.  Here is a draft.  It depends on iolib/os with SBCL.  WIthout SBCL, we also need Osicat.  It does _not_ work on Windows.

```lisp
(defun current-user ()
  (uiop:getenv "USER"))

(defun current-group-id ()
  (alex:assoc-value (osicat:user-info (current-user)) :group-id))

(defun file-group-id (file)
  #-sbcl
  (osicat-posix:stat-gid (osicat-posix:lstat file))
  #+sbcl
  (sb-posix:stat-gid (sb-posix:lstat file)))

(defun executable-p (file)
  (or (and (string= (file-author file)
                    (current-user))
           (member :user-exec (iolib/os:file-permissions file)))
      (and (= (file-group-id file)
              (current-group-id))
           (member :group-exec (iolib/os:file-permissions file)))))
```

Any suggestion?

aartaka:
Looks just fine, I'll incorporate it!

What does `(or trivial-types:pathname-designator)` stand for? is that a forgotten `or` or some shorthand for e.g. `(or ... null)`?

Ambrevar:
Forgotten, you can remove the `(or`.

source/mode/file-manager.lisp
```diff
@@ -26,6 +54,60 @@
...
+  #-sbcl
+  (osicat-posix:stat-gid (osicat-posix:lstat file)))
+
+(sera:-> executable-p ((or trivial-types:pathname-designator) &key (:user string)) boolean)
+(defun executable-p (file &key (user (current-user)))
+  "Return non-nil if FILE is executable for USER name.
+When the user is unspecified, take the current one."
+  (sera:true
+   ;; FIXME: Some files (like /usr/bin/[) cause IOLIB to choke on ENOENT.
+   (let ((permissions (ignore-errors (iolib/os:file-permissions file))))
```
Ambrevar, line 82: The string `"/usr/bin/["` works for me.  If you want a pathname, you must call `uiop:parse-native-namestring` beforehand.

source/mode/file-manager.lisp
```diff
@@ -26,6 +54,60 @@
...
+  #-sbcl
+  (alex:assoc-value (osicat:user-info user) :group-id))
+
+(defun file-group-id (file)
+  #+sbcl
+  (sb-posix:stat-gid (sb-posix:lstat file))
+  #-sbcl
+  (osicat-posix:stat-gid (osicat-posix:lstat file)))
+
+(sera:-> executable-p ((or trivial-types:pathname-designator) &key (:user string)) boolean)
```
Ambrevar, line 76: Remove `sera:` before `->`.

source/mode/file-manager.lisp
```diff
@@ -26,6 +54,60 @@
...
+         (member :other-exec permissions)))))
+
+(export-always 'executables)
+(defun executables ()
+  (let ((paths (str:split ":" (uiop:getenv "PATH") :omit-nulls t)))
+    (sera:filter
+     #'executable-p
+     (remove-if
+      #'uiop:hidden-pathname-p
+      (mapcar #'uiop:resolve-symlinks
```
Ambrevar, line 98: Do you really want to resolve the symlinks here?
If so, then  you should remove duplicates afterwards.

I am not sure it's a good idea because then it would list names that are not wanted.
Say the user has `foo` that links to `foobar`, the user expects to see `foo` in the list, not `foobar`.

source/mode/file-manager.lisp
```diff
@@ -26,6 +54,60 @@
...
+              (alex:mappend (lambda (path) (uiop:directory-files path)) paths))))))
```
Ambrevar, line 99: Useless lambda?

source/mode/file-manager.lisp
aartaka: Well, it misteriously failed for me, so I plugged `ignore-errors` in here just in case. I'm not sure how to debug ENOENTs here, and I'm not sure we need to :)

source/mode/file-manager.lisp
aartaka: Done!

source/mode/file-manager.lisp
aartaka: Hmmmm, that is reasonable. I've removed symlink resolution.

source/mode/file-manager.lisp
aartaka: Yes!

aartaka:
All addressed, let's see whether CI passes :D

jmercouris:
Congratulations Artyom :-)

it passed! 111

aartaka:
Merging then :)

PR #2039, opened 2022-01-21
nyxt.asd: Modularize.

Ambrevar:
With #2034 @aartaka expressed his concern with `nyxt.asd` getting harder and harder to maintain.  Indeed, because everything is loaded in a hard-coded order, it gets very difficult to understand what depends on what and thus to untangle circular dependencies.

In this pull request I'm using the `:module` feature of ASDF to structure our load order a bit.
I've removed all the `:serial t` and added explicit dependencies so this should help us understand what's going on.

It will also help us modularize Nyxt in the future.  For instance, the `Misc modes` module could be disabled altogether and Nyxt would still run.  This makes it easier to know what's required and what is not.

Thoughts?


Ambrevar:
CCL does not work here for some reason.  Any idea?

nyxt.asd
```diff
@@ -94,109 +93,128 @@ A naive benchmark on a 16 Mpbs bandwidth gives us
...
+                ((:file "bookmarklets")
+                 (:file "element-hint" :depends-on ("search-buffer"))
+                 (:file "element-frame")
+                 (:file "jump-heading")
+                 (:file "summarize")
+                 (:file "scroll")
+                 (:file "search-buffer")
+                 (:file "spell-check")
+                 (:file "zoom")))
+               (:module "Misc modes"
```
Ambrevar, line 173: Just "Modes".

jmercouris:
Public note; Pierre and I have spoken about this, and have agreed on how to proceed.

PR #2042, opened 2022-01-21
mode/input-edit: Refactor into move-n-elements

hendursaga:
This could allow `cursor-forwards`/`cursor-backwards` to become more generic in the future, as well.


aartaka:
Looks like a right move. Thanks, @hendursaga!

PR #2045, opened 2022-01-23
fix type error with NYXT_COMPRESS

Uthar:
Fixes:
```
Unhandled SIMPLE-TYPE-ERROR in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                         {100180C293}>:
  The value of SB-IMPL::COMPRESSION is "6", which is not of type (OR BOOLEAN
                                                                     (INTEGER
                                                                      -1 9)).
```


aartaka:
Thanks!

Issue #2011, opened 2022-01-11
headings panel outputs mangled characters sometimes


aartaka:
See https://github.com/atlas-engineer/nyxt/pull/1990#issuecomment-1009768618

aartaka:
It seems to be caused by how we transform Lisp strings into C strings in `cl-webkit`. Those half-decoded strings break exactly where we go outside ASCII, with characters like em dash and non-latin letters.

Is there a way to cleanly transform Lisp strings into UTF-8 arrays to pass to `cffi`, except for `slynk-backend:string-to-utf8`?

guojing0:
Do you think [flexi-streams](https://edicl.github.io/flexi-streams/), [babel](https://babel.common-lisp.dev/), and/or [trivial-utf-8](https://gitlab.common-lisp.net/trivial-utf-8/trivial-utf-8) can be helpful?

aartaka:
Yes, those can help! Will try those out in `cl-webkit` :)

PR #2049, opened 2022-01-24
Internal schemes redesign.

aartaka:
This introduces a new way to define internal schemes that is not strictly renderer-dependent. Given a supported renderer, one should be able to define their own scheme (e.g. `telegram:`) not needing to know anything about how schemes are working at the renderer level.

# What's new:
- `define-internal-scheme` macro.
- `*internal-schemes*` variable to store all the defined schemes.
- `scheme` class encapsulating all the information about the scheme and renderer-agnostic way to call it.
  - There are QtWebEngine-specific settings, in case we'd ever resurrect the port.

# Things to discuss
Should we have magic variables (`%url%` and `%buffer%`) in `define-internal-scheme`? I mean, the allow us to refer to the request url and buffer conveniently, but maybe passing a `(lambda (url buffer) ...)` in body would be better?

# How to Test
Simply run Nyxt and try clicking around manual or visit some Gopher/Gemini space pages. Should work flawlessly.


aartaka:
This also fixed #2011 now.

source/mode/small-web.lisp
```diff
@@ -184,21 +184,23 @@ Second return value should be the MIME-type of the content."))
...
-      (:style (style (buffer mode)))
-      (:style (style mode))
-      (loop for line in (cl-gopher:lines contents)
-            collect (:raw (line->html line))))))
+    (values (spinneret:with-html-string
+              (:style (style (buffer mode)))
+              (:style (style mode))
+              (loop for line in (cl-gopher:lines contents)
+                    collect (:raw (line->html line))))
+            "text/html;charset=utf8")))
```
Ambrevar, line 192: I wonder if this `chrset=utf8` wouldn't call for a more generic fix, perhaps a function or a global?

source/mode/small-web.lisp
```diff
@@ -235,6 +237,20 @@ Second return value should be the MIME-type of the content."))
   (declare (ignore mode))
   (render-binary-content line "image/png"))
 
+;; TODO: :display-isolated-p?
```
Ambrevar, line 240: Can you expand on this TODO?

source/mode/small-web.lisp
```diff
@@ -278,3 +294,67 @@ Second return value should be the MIME-type of the content."))
          (:video :src url :controls t))
         (t (:a :class "button" :href url text))))
     (:br)))
+
+;; TODO: :secure-p t?
+(define-internal-scheme ("gemini")
+  (block process-gemini-scheme
```
Ambrevar, line 300: Block/return usually indicates that the coming flow of code is going to be tangled :p  

Can't we just move the `handler-case` around the `sera:mvlet*`?

source/mode/small-web.lisp
```diff
@@ -278,3 +294,67 @@ Second return value should be the MIME-type of the content."))
...
+                         "text/html;charset=utf8"))
+               (values body meta)))
+          ((:redirect :permanent-redirect)
+           (push %url% (nyxt/small-web-mode:redirections (current-mode 'small-web)))
+           (if (< (length (nyxt/small-web-mode:redirections (current-mode 'small-web)))
+                  (nyxt/small-web-mode:allowed-redirections-count (current-mode 'small-web)))
+               (buffer-load (quri:merge-uris (quri:uri meta) (quri:uri %url%)) :buffer %buffer%)
+               (make-gemini-error-page
+                "Error"
+                (format nil "The server has caused too much (~a+) redirections.~& ~a~{ -> ~a~}"
```
Ambrevar, line 348: Typo: "too many"

source/urls.lisp
```diff
@@ -65,6 +65,106 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
                               ,@body)
                             ,args))))))))
 
+(define-class scheme ()
+  ((name (error "Scheme must have a name/scheme")
+         :documentation "Scheme/name of the internal scheme.
+E.g. \"gopher\", \"irc\".")
```
Ambrevar, line 71: Prefer "for instance" to "e.g." which is more cryptic.

source/urls.lisp
```diff
@@ -65,6 +65,106 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+   (callback
+    nil
+    :type (or null (function (url-designator buffer) t))
```
Ambrevar, line 74: The return type  could be `(values &optional (or string array) string)` or something like that.

source/urls.lisp
```diff
@@ -65,6 +65,106 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+
+Optionally returns two values:
+- The data for page contents (either as string or as a unsigned byte array).
+- The MIME type for the contents.")
+   (local-p
+    nil
+    :documentation "Local schemes are not accessible to the pages of other schemes.")
+   (local-access-p
+    nil
+    :documentation "Local access allowed schemes can load resources from other schemes.
```
Ambrevar, line 87: A bit unclear, maybe "Local-access-allowed schemes can..."

source/urls.lisp
```diff
@@ -65,6 +65,106 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+
+QtWebEngine-specific.")
+   (no-access-p
+    nil
+    :documentation "No-access schemes cannot access pages with any other scheme.
+
+Has slightly different behavior between ports.")
```
Ambrevar, line 94: What does this mean? :p

source/urls.lisp
```diff
@@ -65,6 +65,106 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+   (display-isolated-p
+    nil
+    :documentation "Display isolated schemes cannot be displayed by other schemes.
```
Ambrevar, line 97: What's the difference between "display" and "access"?

source/urls.lisp
```diff
@@ -65,6 +65,106 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+    :documentation "Whether this scheme passes all Content Security Policy checks unconditionally.
+
+QtWebEngine-specific.")
+   (cors-enabled-p
+    nil
+    :documentation "Whether other pages can do requests to the resources with this scheme."))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer (class*:make-name-transformer name))
+  (:documentation "Representation of Nyxt-specific internal scheme."))
```
Ambrevar, line 126: Typo "schemes".

Ambrevar: The webkit / webengine difference call for a tree, no?

Or user classes, and have the renderer combine their webkit-schemes / webengine-scheme class to the user-scheme class, like we do with `buffer`, `gtk-buffer` and `user-buffer`.

source/urls.lisp
```diff
@@ -65,6 +65,106 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+
+(defvar *internal-schemes*
+  (make-hash-table :test 'equal)
```
Ambrevar, line 129: Pro-tip: This is equivalent to the handy `(sera:dict)` ;)

source/urls.lisp
```diff
@@ -233,7 +333,7 @@ Example:
           (error "There's no nyxt:~a page defined" (param-name function-name))))))
 
 (defun internal-url-p (url)
-  (str:s-member (list "nyxt" "gopher") (quri:uri-scheme (url url))))
+  (string= "nyxt" (quri:uri-scheme (url url))))
```
Ambrevar, line 336: Why was gopher internal in the first place?

source/urls.lisp
```diff
@@ -260,6 +360,63 @@ guarantee of the same result."
...
+           (first result))
+          (t (error "Cannot display evaluation result")))))))
+
+(defun sequence-p (object)
+  "Return true if OBJECT is a sequence that's not a string."
+  (typep object '(and sequence (not string))))
+
+(defun scalar-p (object)
+  ;; See `sb-int::compound-object-p'.
+  "Return true if OBJECT is a of one of the following types:
```
Ambrevar, line 385: Typo: "is a of one"

source/urls.lisp
```diff
@@ -260,6 +360,63 @@ guarantee of the same result."
...
+          ((arrayp (first result))
+           (first result))
+          (t (error "Cannot display evaluation result")))))))
+
+(defun sequence-p (object)
+  "Return true if OBJECT is a sequence that's not a string."
+  (typep object '(and sequence (not string))))
+
+(defun scalar-p (object)
+  ;; See `sb-int::compound-object-p'.
```
Ambrevar, line 384: This comment is no longer relevant.
Instead, we can add the suggestion to list the direct subclasses of the `T` type.

Ambrevar:
Excellent, this looks much, much better!  Thanks a lot for this!

About your question, unlike `define-configuration` this is not going to be used many times, I suspect it would be used once for an extension that adds support for a protocol.  Then a lambda is cleaner than the `%url%` and `%buffer%` symbol generation.

source/urls.lisp
aartaka: I mean, QtWebEngine has a slightly different explanation of this type of schemes than WebKitGTK has. Not sure whether it's actually necessary to mention :)

aartaka:
All addressed, including explicit lambda usage and `block`/`return-from` issue.

What I'm still uncertain about is the issue with renderer-specific slots. I'm not sure that we need to remove them, as people might need to write renderer-independent code relying on those. But then, this "renderer-independent" code won't work for one of the renderers...

I'm hesitant here. The best solution seems to be:
- Have basic classes (`local`, `secure`, `cors-enabled`) in the `scheme`.
- Move renderer-specifics to `gtk-scheme` (and `qt-scheme`, if we ever get to it).
- Embrace the renderer-dpendence that the person writing scheme bindings will rely on anyway :)

aartaka:
Done!

source/mode/small-web.lisp
aartaka: A more generic fix is in `cl-webkit` already -- it implies `text/html;charset=utf8` by default. What I did there is a temporary fix, until we release `cl-webkit`-3.6.0.

source/mode/small-web.lisp
aartaka: Embedding Gopher pages via iframes should not be possible. No one does it anyway. So, making it `display-isolated-p`, we're solidifying this behavior and putting a reasonable restriction on our implementation of Gopher.

source/urls.lisp
aartaka: Actually, callbacks can return anything and `cl-webkit` will accept it, treating as an error. Maybe we should leave it without typing, after all?

source/mode/small-web.lisp
Ambrevar: OK, could you add this explanation to the comment?

source/urls.lisp
Ambrevar: OK.

source/mode/small-web.lisp
aartaka: Done!

source/urls.lisp
```diff
@@ -260,6 +324,69 @@ guarantee of the same result."
...
+               (first result))
+              (t (error "Cannot display evaluation result")))))))
+  :local-p t)
+
+(defun sequence-p (object)
+  "Return true if OBJECT is a sequence that's not a string."
+  (typep object '(and sequence (not string))))
+
+(defun scalar-p (object)
+  ;; REVIEW: List direct T subclasses instead?
```
Ambrevar, line 351: No, we need to manually filter them, because they vary across compilers and there is no automatic way to infer which one is serializable and which one is not.

Ambrevar:
> I'm hesitant here. The best solution seems to be:
>
>    Have basic classes (local, secure, cors-enabled) in the scheme.
>    Move renderer-specifics to gtk-scheme (and qt-scheme, if we ever get to it).
>    Embrace the renderer-dpendence that the person writing scheme bindings will rely on anyway :)

What does the first suggestion mean?
OK with the 2nd one.  What's the drawback actually, what makes you hesitate?

aartaka:
> > I'm hesitant here. The best solution seems to be:
> > Have basic classes (local, secure, cors-enabled) in the scheme.
> > Move renderer-specifics to gtk-scheme (and qt-scheme, if we ever get to it).
> > Embrace the renderer-dpendence that the person writing scheme bindings will rely on anyway :)
> 
> What does the first suggestion mean? 

I mean, having slots (like `local-p` etc.) in `scheme` to show the type of this scheme to WebKit.

> OK with the 2nd one. What's the drawback actually, what makes you hesitate?

I wanted to have it not rely on renderer-specific code in any way. That's why I was not sure about moving WebKitGTK-specifics to `gtk-scheme`. Now it feels like the right choice.

aartaka:
Good to merge? :)

Ambrevar:
Wait, I'd like to give this another test.  Will do as soon as possible.

Ambrevar:
Clicking on the `Manual` button from the help page gets me this:

```log
 0: (PATHNAME NIL) [external]
 1: (ENSURE-DIRECTORIES-EXIST NIL :VERBOSE NIL :MODE 511)
 2: (NYXT::CONFIG-CONTENT)
 3: (NYXT::MANUAL-CONTENT)
 4: ((FLET "<BODY>24" :IN "/home/ambrevar/common-lisp/nyxt/source/help.lisp"))
 5: ((LABELS SPINNERET::PRINT-BLOCK-TAG :IN "/home/ambrevar/common-lisp/spinneret/functions.lisp") #<unavailable argument> #<unavailable argument> #<unavailable argument> #<unavailable argument> #<unavail..
 6: (SPINNERET.TAG::BODY #<unavailable argument> #<unavailable argument> #<unavailable argument> #<unavailable argument>)
 7: ((LAMBDA NIL :IN "/home/ambrevar/common-lisp/nyxt/source/help.lisp"))
 8: ((LAMBDA (NYXT:URL NYXT:BUFFER) :IN "/home/ambrevar/common-lisp/nyxt/source/urls.lisp") "nyxt:manual" #<unused argument>)
 9: ((LAMBDA NIL :IN CL-WEBKIT2::URI-SCHEME-PROCESSED))
10: ((LAMBDA NIL :IN BORDEAUX-THREADS::BINDING-DEFAULT-SPECIALS))
```

Ambrevar:
False alarm, there is a bug on master when started with `no-init`.  Will fix.

Ambrevar:
OK to merge!

PR #2051, opened 2022-01-24
help.lisp: Amends dashboard command when no bookmarks exist

efimerspan:
If one invokes `dashboard` and doesn't have any bookmarks saved in their `bookmarks-path`, the command will fail because the `:raw` Spinneret rule expects an array. I've tried to work around this issue by introducing an additional check.


aartaka:
Yes, that does make sense, thank you!

PR #2052, opened 2022-01-24
Update bug_report.md

pdelfino:
This will create an automatic label of `bug`on bug reports.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Bug report
 title: ''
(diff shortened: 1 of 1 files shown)


aartaka:
Some bugs aren't exactly bugs, merely an unexpected behavior or undocumented API. That's why I'm not sure about this automatic label assignment :)

Ambrevar:
Indeed, in my experience a good chunk of bug reports are not bugs, they could be already fixed, or a misunderstanding of how the interface works (in which case we should fix the interface to be more accessible, but it's not really a `bug`).

pdelfino:
OK. It makes sense. So, I am closing the PR...

PR #2054, opened 2022-01-25
Headless mode (#1354) rebase

### LINKED ISSUE #1168: Make Nyxt configurable as headless

aartaka:
This is a fresh rebase of #1354. Not much changed -- a good sign of `prompter` having a good and approachable API!

# Things to DIscuss

- Should `prompt-buffer-channel` be a hook, actually?

# Things to do
- [ ] Test it with some non-trivial scripting.
- [ ]  `nyxt-ready-hook` to start interactive commands from.
- [ ] Maybe utility macros to ease scripting? (see Selenium)

Closes #1168.

### DIFF
--- a/source/global.lisp
+++ b/source/global.lisp
@@ -13,6 +13,10 @@
 This is useful when the browser is run from a REPL so that quitting does not
 close the connection.")
 
+(defvar *headless-p* nil
+  "If non-nil, don't display anything.
+This is convenient for testing purposes or to drive Nyxt programmatically.")
+
 (export-always '*browser*)
 (defvar *browser* nil
   "The entry-point object to a complete instance of Nyxt.
--- a/source/window.lisp
+++ b/source/window.lisp
@@ -21,6 +21,13 @@ It's a function of the window argument that returns the title as a string.")
     (list)
     :export nil
     :documentation "A list of panel buffers appearing on the window.")
+   (prompt-buffer-channel
+    (make-channel) ; TODO: Rename `prompt-buffer-ready-channel'?
+    :export nil
+    :documentation "A channel one may listen to if waiting
+for the prompt buffer to be available.
+You should not rely on the value of this channel.
+The channel is popped when a prompt buffer is hidden.")
    (key-stack
     '()
     :documentation "A stack of the key chords a user has pressed.")
@@ -34,6 +41,7 @@ It's a function of the window argument that returns the title as a string.")
     :export nil
     :type boolean
     :documentation "Whether the window is displayed in fullscreen.")
+   ;; TODO: each frame should have a status buffer, not each window
    (status-buffer
     :export nil)
    (message-buffer-height
--- a/tests/renderer-online/set-url.lisp
+++ b/tests/renderer-online/set-url.lisp
@@ -1,0 +1,46 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt/tests/renderer)
+
+(plan nil)
+
+(class-star:define-class nyxt-user::test-data-profile (nyxt:data-profile)
+  ((nyxt:name :initform "test"))
+  (:documentation "Test profile."))
+
+(defmethod nyxt:expand-data-path ((profile nyxt-user::test-data-profile) (path nyxt:data-path))
+  "Don't persist data"
+  nil)
+
+(defmacro with-prompt-buffer-test (command &body body)
+  (alexandria:with-gensyms (thread)
+    `(let ((,thread (bt:make-thread (lambda () ,command))))
+       (calispel:? (prompt-buffer-channel (current-window)))
+       ,@body
+       (return-selection)
+       (bt:join-thread ,thread))))
+
+;; TODO: Use `with-prompt-buffer-test'.
+;; (with-prompt-buffer-test (set-url)
+;;   (set-prompt-input (current-prompt-buffer) "foobar"))
+
+(subtest "set-url"
+  (setf nyxt::*headless-p* t)
+  (nyxt:start :no-init t :no-auto-config t
+              :socket "/tmp/nyxt-test.socket"
+              :data-profile "test")
+  (sleep 2)                             ; TODO: Wait properly.
+  (let ((url "http://example.org/"))
+    (let ((thread (bt:make-thread (lambda () (nyxt:set-url)))))
+      (calispel:? (nyxt::prompt-buffer-channel (nyxt:current-window)))
+      (nyxt::set-prompt-input (nyxt:current-prompt-buffer) url)
+      (prompter:all-ready-p (nyxt:current-prompt-buffer))
+      (nyxt/prompt-buffer-mode:return-selection)
+      ;; (bt:join-thread thread) ; TODO: Make sure thread always returns.
+      )
+    (sleep 1)                           ; TODO: Wait properly.
+    (prove:is (nyxt:render-url (nyxt:url (nyxt:current-buffer)))
+              url)))
+
+(finalize)
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -357,7 +357,15 @@ A naive benchmark on a 16 Mpbs bandwidth gives us
                cl-gobject-introspection
                bordeaux-threads)
   :pathname "source/"
-  :components ((:file "renderer/gi-gtk")))
+  :components ((:file "renderer/gi-gtk"))
+  :in-order-to ((test-op (test-op "nyxt/gobject/gtk/tests"))))
+
+(defsystem "nyxt/gobject/gtk/tests"
+  :depends-on (nyxt/gobject/gtk prove)
+  :components ((:file "tests/renderer-package"))
+  :perform (test-op (op c)
+                    (nyxt-run-test c "tests/renderer-offline/")
+                    (nyxt-run-test c "tests/renderer-online/" :network-needed-p t)))
 
 (defsystem "nyxt/qt"
   :depends-on (nyxt
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -185,6 +185,7 @@ See also `hide-prompt-buffer'."
   ;; TODO: Add method that returns if there is only 1 source with no filter.
   (when prompt-buffer
     (push prompt-buffer (active-prompt-buffers (window prompt-buffer)))
+    (calispel:! (prompt-buffer-channel (window prompt-buffer)) prompt-buffer)
     (prompt-render prompt-buffer)
     (run-thread "Show prompt watcher"
       (let ((prompt-buffer prompt-buffer))
@@ -202,6 +203,8 @@ See also `show-prompt-buffer'."
     ;; Note that PROMPT-BUFFER is not necessarily first in the list, e.g. a new
     ;; prompt-buffer was invoked before the old one reaches here.
     (alex:deletef (active-prompt-buffers window) prompt-buffer)
+    ;; The channel values are irrelevant, so is the element order:
+    (calispel:? (prompt-buffer-channel (window prompt-buffer)) 0)
     (if (resumable-p prompt-buffer)
         (flet ((prompter= (prompter1 prompter2)
                  (and (string= (prompter:prompt prompter1)
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -368,7 +368,8 @@ response.  The BODY is wrapped with `with-protect'."
 
        (gtk:gtk-container-add gtk-object root-box-layout)
        (setf (slot-value *browser* 'last-active-window) window)
-       (gtk:gtk-widget-show-all gtk-object)
+       (unless *headless-p*
+         (gtk:gtk-widget-show-all gtk-object))
        (connect-signal window "key_press_event" nil (widget event)
          (declare (ignore widget))
          #+darwin
@@ -1058,7 +1059,8 @@ See `gtk-browser's `modifier-translator' slot."
   (let ((old-buffer (active-buffer window)))
     (gtk:gtk-container-remove (main-buffer-container window) (gtk-object old-buffer))
     (gtk:gtk-box-pack-start (main-buffer-container window) (gtk-object buffer) :expand t :fill t)
-    (gtk:gtk-widget-show (gtk-object buffer))
+    (unless *headless-p*
+      (gtk:gtk-widget-show (gtk-object buffer)))
     (when focus
       (gtk:gtk-widget-grab-focus (gtk-object buffer)))
     (nyxt/web-extensions::tabs-on-activated old-buffer buffer)
@@ -1076,7 +1078,8 @@ See `gtk-browser's `modifier-translator' slot."
             (push buffer (panel-buffers-right window))))
   (setf (gtk:gtk-widget-size-request (gtk-object buffer))
         (list (width buffer) -1))
-  (gtk:gtk-widget-show (gtk-object buffer)))
+  (unless *headless-p*
+    (gtk:gtk-widget-show (gtk-object buffer))))
 
 (define-ffi-method ffi-window-set-panel-buffer-width ((window gtk-window) (buffer panel-buffer) width)
   "Set the width of a panel buffer."
--- a/tests/renderer-package.lisp
+++ b/tests/renderer-package.lisp
@@ -1,0 +1,8 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package nyxt/tests/renderer
+  (:use #:common-lisp)
+  (:use #:nyxt)
+  (:use #:prove)
+  (:import-from #:class-star #:define-class))


aartaka:
@Ambrevar, how does one set prompt input? `update-prompt-input` doesn't seem to work well somewhy...

See tests/renderer-online/set-url.lisp.

source/browser.lisp
```diff
@@ -324,6 +329,7 @@ prevents otherwise."))
                (startup browser urls)))))))
   ;; Set `init-time' at the end of finalize to take the complete startup time
   ;; into account.
+  (hooks:run-hook (after-startup-hook *browser*))
```
Ambrevar: Shouldn't it be after `ready-p` is set to `t`? (2 s-exps down)

Ambrevar: Sorry, I just saw that you moved it to `startup`.

Ambrevar:
Indeed, you must call `set-prompt-buffer-input`.
`update-prompt-buffer` is to update the display.  The fact that it takes an input is mostly an implementation detail.

We must fix this bad API.  I suggest documenting it (e.g. mentioning `set-prompt-buffer-input` in the docstring).
Should we unexport `update-prompt-input` and export `set-prompt-buffer-input` instead?

Ambrevar:
Thanks for this update!

It's a great start, but as you can see, it's a lot of boilerplate code just to drive a simple `set-url`!!! :p 

In my opinion, we should aim for something much simpler.
We previously discussed whether we should use hooks or channels.
Actually, I see now benefits of channels:
- no need to declare functions / lambdas, you can listen / write straight from the top level.
- No need to clean up the hook when you are done, since reading and writing are a one-time operation.
The downsides:
- non-theaded code is blocking; while adding to a hook does not block.

So what does it mean to aim for "short code"?  Let's try to reach this pseudo-code:

```lisp
(let ((url-channel (nyxt::make-channel 1))
        (url "http://example.org/"))
    (setf nyxt::*headless-p* t)
    (nhooks:add-hook
     nyxt:*after-SOME-hook*
     (make-instance
      'nhooks:handler
      :fn (lambda ()
            (nyxt::set-prompt-buffer-input (nyxt:current-prompt-buffer) url)
            (nyxt/prompt-buffer-mode:return-selection)
            (nhooks:add-hook
             (buffer-loaded-hook (current-buffer))
             (make-instance
              'nhooks:handler
              :fn (lambda (buffer)
                    (calispel:! url-channel (nyxt:render-url (nyxt:url buffer)))
                    (nhooks:remove-hook (buffer-loaded-hook buffer) 'example-org-loaded))
              :name 'example-org-loaded)))
      :name 'browser-started))
    (nyxt:start :no-init t :no-auto-config t
                :socket "/tmp/nyxt-test.socket"
                :data-profile "test")
    (prove:is (calispel:? url-channel 10) url))
```

To reach there, it would be nice if the first buffer load could be controlled somehow.

Ambrevar:
And here is the same pseudo code, using the channel paradigm:

```lisp
(let ((url-channel (nyxt::make-channel 1))
        (url "http://example.org/"))
    (setf nyxt::*headless-p* t)
    (when (calispel:? nyxt:*after-SOME-startup-channel*)
      (nyxt::set-prompt-buffer-input (nyxt:current-prompt-buffer) url)
      (nyxt/prompt-buffer-mode:return-selection)
      (when (calispel:? nyxt:*after-buffer-loaded-channel*)
        (prove:is (nyxt:url buffer) url)))
    (nyxt:start :no-init t :no-auto-config t
                :socket "/tmp/nyxt-test.socket"
                :data-profile "test"))
```

As you can see, it's much shorter and easier to write.
Now channels values are read by only one reader at a time.  What it we want to unblock all the sites that are waiting on a channel?  We need some other mechanism, maybe a form of "broadcast  channel"?  Any idea?  Any CL library for that?

Ambrevar:
Also now that internal buffers have URLs, we can set the URL to something internal, thus no network is needed and we can remove the timeouts :)

aartaka:
> Indeed, you must call `set-prompt-buffer-input`. `update-prompt-buffer` is to update the display. The fact that it takes an input is mostly an implementation detail.

Thanks for the hint!

> We must fix this bad API. I suggest documenting it (e.g. mentioning `set-prompt-buffer-input` in the docstring). Should we unexport `update-prompt-input` and export `set-prompt-buffer-input` instead?

Documenting could indeed help. I'd export both, as `update-prompt-buffer-input` is used in several places in Nyxt code, while `set-prompt-buffer-input` can come in handy for automation.

aartaka:
Regarding hooks vs. channels:

I don't like channels in the context of event handling, as those block the thread they are on, are extremely time-sensitive, and are a concept that is hard to understand at first.

Hooks are familliar, somewhat-asynchronous, and more flexible in general. The issue that you identified is merely a notation one -- hook addition is overly verbose and nested. But it's all Lisp -- we can write macros to shorten the code. I suggest at least two macros to handle hooks: `on` and `once-on`. `on` simply adds a handler to the hook, while `once-on` also deletes it after it fires (which is a frequent use-case). Examples:
``` lisp
(once-on (buffer-loaded-hook buffer) ; Hook.
    (buffer) ; Lambda list for the callback.
  (calispel:! buffer-loaded-channel t)) ; hooks:remove-hook is hidden.

(on (after-download-hook)
    (download)
  (uiop:launch-program (list "xdg-open" (nyxt::destination-path download))))
```
Obviously, JavaScript with its `await` syntax is an influence here.

Alternatively, we can introduce something similar to Haskell's `do`-notation, but based on hooks:
``` lisp
(hook-do
  (buffer) <- (buffer-loaded-hook (current-buffer)) ; BUFFER is only set when `buffer-loaded-hook' fires
  (echo "Buffer ~a loaded ~a" (id buffer) (render-url (url buffer))) ; Inside `buffer-loaded-hook'.
  (calispel:! buffer-loaded-channel t) ; Inside `buffer-loaded-hook'.
  (request-data) <- (request-resource-hook buffer) ; REQUEST-DATA is only set when the hook fires.
  (not (search "google" (render-url (url request-data))))) ; Inside `request-resource-hook` and `buffer-loaded-hook'.
```

Ambrevar:
Good points!

I'd like to clarify something about what I meant with my "broadcast"
channels: what we are really dealing with here is called "signal" in the
GTK sense (close to Unix signals, not to be confused with unrelated CL
signals).

Blocking is indeed a benefit: more often than not you don't want to
execute the following code until you got some result.

Another approach would be to use futures and promises. Are you familiar
with these?  What do you think?

aartaka:
> Good points! I'd like to clarify something about what I meant with my "broadcast" channels: what we are really dealing with here is called "signal" in the GTK sense (close to Unix signals, not to be confused with unrelated CL signals). 

Now I understand it. But still, why introduce channels as a user-facing thing, if we have hooks?

> Blocking is indeed a benefit: more often than not you don't want to execute the following code until you got some result. Another approach would be to use futures and promises. Are you familiar with these? What do you think?

The code I suggested is heavily inspired by JavaScript Promises handling, so yes, I am familliar with those. But still, out pseud-promise hooks would work just fine, given enough syntactic sugar, so why introduce yet another abstraction?

Ambrevar:
Can you explain what you mean with `buffer-loaded-channel` in your previous examples?
Is it meant for blocking?
If not, we would need to introduce a way to make blocking calls, for instance `(once-on (hook :block-p t) ...)`.

aartaka:
> Can you explain what you mean with `buffer-loaded-channel` in your previous examples?

Oh, that was confusing, I suppose. It was basically a copy-paste from the test/renderer-online/set-url.lisp test, so `buffer-loaded-channel` doesn't matter that much.

> Is it meant for blocking?

No :)

> If not, we would need to introduce a way to make blocking calls, for instance `(once-on (hook :block-p t) ...)`.

Hmmmmmm. I mean, why would we need to? All the code that needs to depend on the hook activating goes inside `once-on` and that should be enough. A typical callback ~hell~ nesting.

Ambrevar:
Fair enough :)

Can you give these macros a try then?

aartaka:
> Fair enough :) Can you give these macros a try then?

Sure!

aartaka:
`on` and `once-on` added in 4628258729c91d1da383ab4380fe5228a1add731!

Do they maybe belong to `hooks` rather than to the core of Nyxt?

aartaka:
And can we make `prompt-buffer-channel` to be a hook? Will we lose some of its features when doing so?

Ambrevar:
No loss, the prompt-buffer-channel was introduce exactly for the purpose
we are discussing, if we are going for hooks then it makes sense to turn
it into a hook!

Ambrevar:
Do you need the `sleep 1` if you set the URL to an internal 1?

Also by "set-url thread", you mean the renderer thread?

source/configuration.lisp
```diff
@@ -357,3 +357,37 @@ exists."
...
+
+See `on'."
+  (let ((handler-name (gensym "once-on-hook-handler")))
+    (alex:once-only (hook)
+      `(hooks:add-hook
+        ,hook (make-instance 'hooks:handler
+                             :fn (lambda ,(alex:ensure-list args)
+                                   (declare (ignorable ,@(alex:ensure-list args)))
+                                   (unwind-protect
+                                        ,@body
```
Ambrevar, line 391: Missing `progn` around `,@body`.

source/configuration.lisp
```diff
@@ -357,3 +357,37 @@ exists."
...
+
+(export-always 'once-on)
+(defmacro once-on (hook args &body body)
+  "Attach a handler with ARGS and BODY to the HOOK.
+
+Remove the handler after it fires the first time.
+
+See `on'."
+  (let ((handler-name (gensym "once-on-hook-handler")))
+    (alex:once-only (hook)
```
Ambrevar, line 385: Looks like the `args` should also go to `once-only`, no?

aartaka: Yes!

aartaka:
> Do you need the `sleep 1` if you set the URL to an internal 1? 

`sleep 1` is there for a different purpose: to ensure that the thread that invokes `set-url` has already invoked `set-url`. Otherwise we are at risk of querying the non-existent prompt-buffer and failing the test.

> Also by "set-url thread", you mean the renderer thread?

I mean the thread that `set-url` is invoked on. I didn't have a better name, so I've thrown this in :)

BlueFlo0d:
> Alternatively, we can introduce something similar to Haskell's do-notation, but based on hooks:
Haskell sucks (irrelevant to topic sorry). What about good'ol direct style (w.r.t. CPS).
```
(cps
 (let ((buffer (buffer-loaded-hook (current-buffer))))
   (echo "Buffer ~a loaded ~a" (id buffer) (render-url (url buffer)))
   (calispel:! buffer-loaded-channel t)
   (not (search "google" (render-url (url (request-resource-hook buffer)))))) )
```

With regard to this use case of hook, it does seem to fit nicely, but I have to think more, there's something that I feel off.

> I don't like channels in the context of event handling, as those block the thread they are on, are extremely time-sensitive, and are a concept that is hard to understand at first.

Maybe that's exactly what I'm worried about, but for channel. I think the channel behavior is more understandable as it resembles the command loop paradigm in Emacs (writing channels explicitly is confusing, but if you hide them in functions, those functions just look like normal functions that take longer time to run). The simple control flow is already broken to a certain degree in Nyxt, for goodness or badness. On the other hand, hooks+CPS transform (or call it by some other name) causes following code to run in a different thread, and the order of execution can be unpredictable (from mental model aspect) if multiple handlers are registered.

I'm not saying the Emacs command loop is the way to go, people are cursing it forever for its single-threadness and non-responsiveness, but there're other aspects of it that I like a lot. Just want to add this factor to the discussion.

aartaka:
The `prompt-buffer-ready-hook` is there instead of `prompt-buffer-channel`, macros are fixed. What's left is discussing things :)

Ambrevar:
> `sleep 1` is there for a different purpose: to ensure that the thread
> that invokes `set-url` has already invoked `set-url`. Otherwise we are
> at risk of querying the non-existent prompt-buffer and failing the
> test.

This does not feel right, we should have a hook then.

BlueFlo0d:
> Another approach would be to use futures and promises. Are you familiar with these? What do you think?

We can just use `cl-cont`.

Ambrevar:
@BlueFlo0d Can you provide an example of how we could leverage continuations for headless / prompt-buffer scripting?

Ambrevar:
I'll have a look at the `(sleep 1)` issue later.
Left to do:

- Export `set-prompt-buffer-input`.
- Document it with `update-prompt-buffer`.

BlueFlo0d:
> @BlueFlo0d Can you provide an example of how we could leverage continuations for headless / prompt-buffer scripting?

See my direct style example above.
```

(cps
 (let ((buffer (? (buffer-loaded-hook (current-buffer)))))
   (echo "Buffer ~a loaded ~a" (id buffer) (render-url (url buffer)))
   (calispel:! buffer-loaded-channel t)
   (not (search "google" (render-url (url (? (request-resource-hook buffer))))))))
```
You can call `?` just like a normal function which “looks like” waiting for a hook to run, and it can be implemented by capturing the continuation of the `?` and register as a handler to said hook.

aartaka:
As an addition, should we have `--headless` CLI argument to Nyxt to start it headless (with `*headless-p*` set to true) right away?

Ambrevar:
Should I address the remaining issues and merge this?

Ambrevar:
@BlueFlo0d Sorry, I don't understand the code completely.  What do `?` and `cps` do?

aartaka:
Yes, if you don't mind :)

source/global.lisp
```diff
@@ -23,7 +27,7 @@ It can be initialized with
 It's possible to run multiple interfaces of Nyxt at the same time.  You can
 let-bind *browser* to temporarily switch interface.")
 
-(declaim (type hooks:hook-void *after-init-hook*))
+(declaim (type hooks:hook-void *after-init-hook* *after-startup-hook*))
```
Ambrevar, line 30: I would move this to the declaration site.

source/configuration.lisp
```diff
@@ -357,3 +357,39 @@ exists."
...
   (open pathname :direction :probe :if-does-not-exist :create))
+
+(export-always 'on)
+(defmacro on (hook args &body body)
+  "Attach a handler with ARGS and BODY to the HOOK.
+
+ARGS can be
+- A symbol if there's only one argument to the callback.
+- A list of arguments.
+- An empty list, if the hook is void."
```
Ambrevar, line 368: You mean "if the hook handlers take no argument"?

source/window.lisp
```diff
@@ -34,6 +34,7 @@ It's a function of the window argument that returns the title as a string.")
     :export nil
     :type boolean
     :documentation "Whether the window is displayed in fullscreen.")
+   ;; TODO: each frame should have a status buffer, not each window
```
Ambrevar, line 37: Is this related to the pull request?

What do you mean?  We don't have frames (beside panel buffers), do we?

tests/renderer-online/set-url.lisp
```diff
@@ -0,0 +1,55 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt/tests/renderer)
+
+(plan nil)
+
+(class-star:define-class nyxt-user::test-data-profile (nyxt:data-profile)
+  ((nyxt:name :initform "test"))
+  (:documentation "Test profile."))
```
Ambrevar, line 10: Needs to be updated since master no longer has data-profiles.

Ambrevar:
Replacing `sleep 1` with `(prompter:all-ready-p prompt-buffer)` should work, no?

source/window.lisp
aartaka: This comment was there before the patch. Seems like it was deleted on master 0_o

source/configuration.lisp
aartaka: Yes, exactly :)

aartaka:
Yes, could work!

BlueFlo0d:
> @BlueFlo0d Sorry, I don't understand the code completely. What do `?` and `cps` do?

`(cps &body body)` is a macro that do CPS transform to its enclosing forms. This supports using `?` inside a `cps` form.

Behaviorally, `(? hook)` block and wait for `hook` to be called, and return some value only after that. 

There're several possibility about how to specify what value to return, maybe the most straightforward one being using the syntax `(? hook return-form)`, and evaluate `return-form` only after `hook` is called. Or maybe just plainly `(? hook)` and always return `nil`, and let user get needed values themselves. Or it could depend on the hook combination, and return whatever value passed to the hook handler. I think the last one is the most natural.

The magic CPS transform brings is that, `(? hook)` doesn't actually need to block any thread -- it simply reify its current continuation and register it as a hook handler. In the example above, the continuation of `(? (buffer-loaded-hook (current-buffer)))` is roughly speaking something like
```
 (lambda (buffer)
   (echo "Buffer ~a loaded ~a" (id buffer) (render-url (url buffer)))
   (calispel:! buffer-loaded-channel t)
   (not (search "google" (render-url (url (? (request-resource-hook buffer)))))))
```
and this will be registered as a handler of `(buffer-loaded-hook (current-buffer))`. Code outside the `cps` form can execute without blocking, and the above continuation (or, the "rest" of code inside `cps` form) will be executed in the thread running `(buffer-loaded-hook (current-buffer))` later.

Ambrevar:
Understood.

If I understood you right, you are talking about a hypothetical `cps`
macro, not cl-cont.

If you implement the continuation passing by adding the continuation as a
handler to the hook, is it different from the `on` macro that Artyom
just implemented?

BlueFlo0d:
> Understood. If I understood you right, you are talking about a hypothetical `cps` macro, not cl-cont. 

`cps` is actually basically just `cl-cont:with-continuation`, IIRC. I'm pretty sure we can implement the above described `cps` and `?` functionality within 20 lines using `cl-cont` and `nhooks`.

> If you implement the continuation passing by adding the continuation as a handler to the hook, is it different from the `on` macro that Artyom just implemented?

They're apart from each other by one CPS transform. Basically, using `on`, you have to write explicitly in continuation-passing style, while using `cps`/`cl-cont`, you write in direct style and the macros convert your code to CPS under the hood. Imagine if you have some computation that need to wait for several "events" (formally hook triggering) along the way, in CPS style you have to write `(progn A (on (...) ... B (on (...) ... C (on (...) ... D))))`, while in direct style you just need to write `(progn A (? ...) B (? ...) C (? ...) D)`. The difference become larger when dealing with more complicated procedure. Another advantage of `cps`/`cl-cont` will emerge when you consider factoring some asynchronous code into functions. In CPS style, you may write `(progn A (on (h1) ... (let ((result (f)) B)))` to arrange `(let ((result (f)) B))` to run after `h1`, however if in some case `(f)` itself need to wait for something, there's no way to delay `B` until `f` produce sensible result. There is a way without using `cps`/`cl-cont` -- you just need to in turn write `f` as a CPS function, something like `(defun f (cont) (funcall cont ...))` and then write `(progn A (on (h1) (f (lambda (result) B))))`, however it's much less a hassle to write everything normally and let `cl-cont` take care of it.

Ambrevar:
So it's mostly a syntactic benefit you are proposing.
Fair enough, but that would drag in an additional dependency for
something that seems (in my current opinion) relatively minor.

I suggest we merge `on-hook` for now and if you think your CPS style is
superior, please send a working patch to exhibit the benefits, it will
help with the discussion I'm sure :)

BlueFlo0d:
> So it's mostly a syntactic benefit you are proposing. 

I don't think this is syntactic. I think the "factoring some asynchronous code into functions" section shows it pretty clearly. Blocking computation within a dynamic extent from anywhere inside is a straightforward insertion of `?` in direct style (or, "my CPS style"), while is impossible without propagating changes through all the code within such dynamic extent in CPS form (or, "`on` style").

The well-know definition of "syntactic sugar", an synonym of "macro expressibility", defines equi-expressibility as mutually translatable via _local_ syntactic transform (See the classic paper 
_Felleisen, Matthias. 1990 "On the expressive power of programming languages."_). CPS transform is well known to be a non-local transform and continuation capturing is real expressive power. You can't sweep everything under the rug of "being syntactic", otherwise you will be equating LISP and assembly and Turing machine. CPS transform is known to be non-purely-syntactic.


> I suggest we merge `on-hook` for now and if you think your CPS style is superior, please send a working patch to exhibit the benefits, it will help with the discussion I'm sure :)

Sure, I'll try to find sometime to do it.

Ambrevar:
Merged with efa3b7da80f216a1800f9b9683d0740b8540b8d6.

Ambrevar:
Note: I found this library for events / notifications:

https://github.com/fukamachi/event-emitter

Looks very similar to @aartaka's implementation!

PR #2058, opened 2022-01-26
DIFF is now an internal page command

shamazmazum:
Hi! This is connected to #2016. `diff` command uses an old `make-internal-buffer` function to create a buffer and currently does not work. This commit defines `diff` with `make-internal-page-command-global` instead.

### DIFF
--- a/source/mode/diff.lisp
+++ b/source/mode/diff.lisp
@@ -1,70 +1,50 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-(in-package :nyxt)
+(uiop:define-package nyxt/diff-mode
+  (:use #:cl #:nyxt)
+  (:import-from #:serapeum #:export-always))
 
+(in-package :nyxt/diff-mode)
+
+(export-always 'diff-mode)
 (define-mode diff-mode ()
   "Diff mode is used to view the diffs between two buffers."
   ((rememberable-p nil)
-   (buffer (error "Please supply a buffer.")
-           :documentation "The buffer where the diff will be displayed.")
-   (old-html :documentation "String HTML representation.")
-   (new-html :documentation "String HTML representation.")
-   (diff-html :documentation "String HTML representation of the computed diff
-   between `old-html' and `new-html' slots.")
-   (diff-style (cl-css:css
-                ;; TODO: Apply theme here?
-                '((".nyxt-diff-insert"
-                   :text-decoration "none"
-                   :background-color "#bbeabb")
-                  ("ins.nyxt-diff-replace"
-                   :text-decoration "none"
-                   :background-color "#bbeabb")
-                  (".nyxt-diff-delete"
-                   :text-decoration "none"
-                   :background-color "#efcbcf")
-                  ("del.nyxt-diff-replace"
-                   :text-decoration "none"
-                   :background-color "#efcbcf")))
-               :documentation "Diff colours for its visual representation.
+   (style (theme:themed-css (theme *browser*)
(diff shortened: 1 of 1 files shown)


shamazmazum:
I never used this command before. It seems the original command created a document with completely overriden CSS (like green color for added text, red color for removed text). This is what I do in this patch. Usually diffs look very bad because of this.

Is there any way to merge this new CSS with the original style (taken from one of compared buffers)? I am just not sure if the style is always in `<head><style>` tag. Maybe webkit has means to get CSS of a document?

shamazmazum:
This is what it looks like when comparing two text files.
20220126_13h32m26s_grim

aartaka:
> I never used this command before. It seems the original command created a document with completely overriden CSS (like green color for added text, red color for removed text). This is what I do in this patch. Usually diffs look very bad because of this.

Yes, and you're doing it right :)

> Is there any way to merge this new CSS with the original style (taken from one of compared buffers)? I am just not sure if the style is always in `<head><style>` tag. Maybe webkit has means to get CSS of a document?

It's somewhat simpler: `define-internal-page-command(-global)` injects default style automatically, you only need to override some of it in the `style` of the `diff-mode`, as CSS of the mode is inserted after the buffer one, thus allowing us to override any CSS buffer sets, while preserving the rest of buffer's style.

Is that what you need there?

source/mode/diff.lisp
```diff
@@ -1,70 +1,50 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-(in-package :nyxt)
+(uiop:define-package nyxt/diff-mode
+  (:use #:cl #:nyxt)
+  (:export #:diff-mode
```
aartaka, line 6: Note on exports and imports: we usually use `serapeum:export-always` to export symbols from the package. We usually import it directly. Given this, the package definition would look something like:

``` lisp
(uiop:define-package nyxt/diff-mode
   (:use #:cl #:nyxt)
   (:import-from #:serapeum #:export-always))
```

And then there will be `export-always` scattered around in the code.

shamazmazum:
Fixed exporting.

This is what I ment about styles. Consider this page:
20220127_08h20m51s_grim

I changed it a bit and generated diff:
20220127_08h26m43s_grim

As you can see, it's ugly in its appearance. I looked at source code:
20220127_08h26m47s_grim

I think, it cannot load `static/style.css` (sure, why should it?) and that's why it's ugly. If you don't know a solution, you can merge if everything else is OK.

aartaka:
There's no way to load external resources to `nyxt:` pages, so I believe that the current state is the best we can do about preserving styles :)

Merging this. Thanks for the fix!

aadcg:
Given the drawback that @shamazmazum pointed out about non-preservation of the style (impossibility of loading external resources to `nyxt:` pages) shouldn't we revert to the old way things were done?

aartaka:
Oh, no, that won't work either, because `nyxt:` pages are local...

Maybe a separate, non-local `diff:` scheme?

aadcg:
Are you suggesting that we could have internal pages with a `diff:` scheme that would allow loading external resources? 

To make it clear, the property of "non loading external resources" comes from the `nyxt:` scheme, not from internal pages right?

aartaka:
> Are you suggesting that we could have internal pages with a `diff:` scheme that would allow loading external resources?

Yes.

> To make it clear, the property of "non loading external resources" comes from the `nyxt:` scheme, not from internal pages right?

Exactly. `nyxt:` scheme is marked as `:local-p t` (see urls.lisp), which restricts loading external content into `nyxt:` pages.

Issue #2059, opened 2022-01-26
lisp-eval broken?


Ambrevar:
No `lisp-eval` call seems to do anything on master (1d88381513cf0469583febcd6fb070841fd614eb).  @aartaka Regression?

aartaka:
You sure? "Edit user files button" works for me at ed97f71e66c7a4b3eea481d3bb7e5951fb88df76 :upside_down_face:

Ambrevar:
Commit d3f7ca64c600818c8e938ad80da9a67e8c823f2b works for me.

Ambrevar:
My bad, there is an issue when I load my init file.  I presume something to do with the modes.

Ambrevar:
Got it!  It's noscript-mode :p 

I guess we should change noscript-mode to not do anything on internal page.  What do you think?

aartaka:
> Got it! It's noscript-mode :p 

Wow, what an unexpected move :D

> I guess we should change noscript-mode to not do anything on internal page. What do you think?

Yes, that sounds reasonable :)

Ambrevar:
Will do!

PR #2060, opened 2022-01-26
No script mode for internal buffer

Ambrevar:
Fixes https://github.com/atlas-engineer/nyxt/issues/2059.
@aartaka Looks good to you?

### DIFF
--- a/source/mode/no-script.lisp
+++ b/source/mode/no-script.lisp
@@ -7,10 +7,16 @@
 (in-package :nyxt/no-script-mode)
 
 (define-mode no-script-mode ()
-  "Disable Javascript in current buffer."
+  "Disable Javascript in current buffer.
+If the current URL is an internal one (for instance 'nyxt://...'), JavaScript is enabled.
+It's automatically disabled when loading a non-internal URL."
   ((destructor
     (lambda (mode)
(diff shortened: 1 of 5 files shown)


source/mode/no-script.lisp
```diff
@@ -7,10 +7,18 @@
...
   ((destructor
     (lambda (mode)
       (ffi-buffer-enable-javascript-markup (buffer mode) t)))
    (constructor
     (lambda (mode)
-      (ffi-buffer-enable-javascript-markup (buffer mode) nil)))))
+      (declare (ignore mode))
+      (echo "Reload the buffer for no-script-mode to take effect.")))))
+
+(defmethod nyxt:on-signal-load-started ((mode no-script-mode) url)
```
aartaka, line 21: That's suspicious. Maybe add a default auto-mode rule instead of all this code?

Ambrevar: Do we have "default auto-mode rules"?

I'm no sold on this idea yet.  We need to clarify what it would entail.

"All this code" is rather short in the end.

aartaka:
`on-signal-load-started` is useful. Even if we don't merge this, it'd be nice to cherry-pick :)

aartaka:
Looks good now that #2061 is resolved :)

Issue #2004, opened 2022-01-08
Default auto-mode rules?


aartaka:
There are often URLs that would only work well if some modes are enabled. Think .onion URLs — unless one is using system-wide proxy, `proxy-mode` is a must.

Another example is #1994: all `gopher:` URLs should have `gopher-mode` enabled for them.

I believe it could be convenient if we provide some `auto-mode` rules by default. Obviously, they will be there in the `auto-mode-rules.lisp` file so that one can disable those.

Does that make sense?

guojing0:
Do you mean that if Nyxt detects `.onion` URLs, then `proxy-mode` is enabled before loading these URLs? Similarly, if the URL starts with `gopher:` then `gopher-mode` is enabled before loading said page?

aartaka:
Yes, that's what I mean. Good starting point for .onion URLs could be [this post on the forum](https://discourse.atlas.engineer/t/tor-v3-onion-address-auto-mode-rule/395).

hendursaga:
For onion addresses, a good idea might be to specialize the proxy-mode class, with just the Tor proxy configured. I would also add support for [Lokinet](https://lokinet.org/) addresses, too, with the same idea.

hendursaga:
Naturally, eventually this should include support for [I2P](https://geti2p.net/en/) and [GNUnet](https://www.gnunet.org/en/) addresses, along with whatever other "alternative" Internet protocols there are.

Ambrevar:
I found out another critical issue: start the `lisp-repl`, evaluate `*browser*`, click on the link.  You are now in the help buffer of `*browser*`, but you are still in `repl-mode`! :/ 

@aartaka Thoughts on this?

Actually if we had major modes, the issue would be fixed in this case and in all the aforementioned cases.  What do you think?

Another fix is to remember modes in history, but then we've got to make the distinction between modes that were automatically enabled and those that were enabled by the user.

aartaka:
> I found out another critical issue: start the `lisp-repl`, evaluate `*browser*`, click on the link. You are now in the help buffer of `*browser*`, but you are still in `repl-mode`! :/

Indeed, that's a quirky one 0_o

> Actually if we had major modes, the issue would be fixed in this case and in all the aforementioned cases. What do you think?

Yes, that sounds sane. I guess there was enough of precedents to finally get to major modes :D 

> Another fix is to remember modes in history, but then we've got to make the distinction between modes that were automatically enabled and those that were enabled by the user.

[...]

PR #2061, opened 2022-01-26
Default auto-mode rules.

aartaka:
This is an implementation for #2004 and response to #2060. @Ambrevar, here's how default auto-mode rules look for me as the prototype level.

# What's New
Auto mode rules, if null, will be set to the list of reasonable rules that `small-web-mode` and `no-script-mode` would need  to work properly. 

# To Do: 
Remove the code that simulates this from `small-web-mode`.

# Things to Discuss
Is it alrights that it implies empty auto-mode-rules.lisp file? Long-time users may need to move their file to re-populate it with default rules, but it will cost nothing for new users.

### DIFF
--- a/source/mode/auto.lisp
+++ b/source/mode/auto.lisp
@@ -397,7 +397,17 @@ Auto-mode is re-enabled once the page is reloaded."
     (values list &optional))
 (sera:export-always 'add-modes-to-auto-mode-rules)
 (defun add-modes-to-auto-mode-rules (test &key (append-p nil) exclude include (exact-p nil))
-  (with-data-access (rules (auto-mode-rules-path (current-buffer)))
+  (with-data-access
+      (rules (auto-mode-rules-path (current-buffer))
+       :default (list (make-instance 'auto-mode-rule
+                                     :test '(match-scheme "nyxt")
+                                     :excluded '(no-script-mode))
+                      (make-instance 'auto-mode-rule
+                                     :test '(match-scheme "gopher")
+                                     :included '(small-web-mode))
+                      (make-instance 'auto-mode-rule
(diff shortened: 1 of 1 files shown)


Ambrevar:
1. This is hard-coded, I think it should be made extensible (e.g. if someone adds a new protocol/scheme support).
2. It's useless for people who are not using auto-mode, right?
3. That is only works on new auto-mode rule file is a problem in my opinion, as it will confuse existing users, which we don't want to.  Instead, I propose the following fix: add the mode not to the file but when applying the rule: if the rule is empty, get the default rule.

With all that said, I'm a bit cold on this change.
In particular, notice how auto-mode is now tightly coupled to gemini/gopher and no-script-mode.  3 things that should be unrelated.

The benefit of the `on-signal` approach is that everything remains decoupled.  For instance, my proposal in #2060 keeps all the no-script related code to `mode/no-script.lisp`.

Ambrevar:
> Even though I don't like enable-mode etc. used by something other than auto-mode, it actually fits the situation.

Neither this nor #2060 uses `enable-mode`, what do you mean?

aartaka:
> > Even though I don't like enable-mode etc. used by something other than auto-mode, it actually fits the situation.
> 
> Neither this nor #2060 uses `enable-mode`, what do you mean?

I mean it as used in `small-web-mode`, for example. In general, toggling modes or other mode-related settings when on certain URLs doesn't feel right, but it's a necessaity, I guess.

Ambrevar:
This calls for an important question: should auto-mode be "promoted" to
a first class feature of the browser?

After all, it's convoluted enough to have "auto-mode" be a mode that can
disable itself.

On the other hand, everything-as-a-mode is also pretty neat...

PR #2063, opened 2022-01-26
Disambiguate symbol in resolve-backtick-quote-links

Ambrevar:
This addresses the issue mentioned in https://github.com/atlas-engineer/nyxt/issues/967#issuecomment-1022143645.

## To test:

Run `describe-variable`, find `*after-init-hook*`, then click on `*global*`.
The description of `*global*` should open automatically.

Then run `describe-command`, find `describe-command`, click on `execute-command`.  This time the multiple choices should be prompted.

## To do:

1. Currently the prompt buffer displays for a fraction of a second on single match.  This is normal, since the effect of `auto-return-p` is done after the first source update, which happens after display.

It's a  bit of a hack.  To fix it, we could hide the display until the first update is done.  But then if the update takes a while to complete, the user would get no feedback.  Maybe it's OK to leave it?

2. For now we only list Nyxt-related symbols.  What about symbols from other libraries?  If, say, `sera:single` is in a docstring, it won't be clickable then, which is too bad :p  Maybe we could add an extra parameter to all `describe-*` commands to specify which package to include, where `nil` would mean "all", and the default would be `(nyxt nyxt-user)`.

@aartaka What do you think?

### DIFF
--- a/libraries/prompter/prompter.lisp
+++ b/libraries/prompter/prompter.lisp
@@ -81,6 +81,12 @@ It's called after the sources are cleaned up.
 
 Note that the function is executed *before* performing an action.")
 
+     (auto-return-p nil
+                    :type boolean
+                    :documentation
+                    "Automatically return and run default action when the
+suggestions are narrowed down to just one item.")
+
      (history (make-history)
               :type (or containers:ring-buffer-reverse null)
               :documentation
@@ -166,6 +172,9 @@ compution is not finished.")))
       (setf (slot-value prompter 'input) text)
       (update-sources prompter text)
       (select-first prompter)))
+  (when (and (auto-return-p prompter)
+             (sera:single (all-suggestions prompter)))
+    (return-selection prompter))
   text)
 
 (export-always 'destroy)
--- a/libraries/prompter/prompter.lisp
+++ b/libraries/prompter/prompter.lisp
@@ -489,7 +489,8 @@ suggestions."
   "Return the list of the suggestions in the prompter."
   (alex:mappend #'suggestions (sources prompter)))
 
-(defun default-action (prompter)
+(export-always 'default-action)
+(defmethod default-action ((prompter prompter))
   (first (actions prompter)))
 
 (export-always 'resume)
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -122,7 +122,14 @@ When INPUT does not have a unique match, prompt for the list of exact matches."
                   (symbol (let ((*package* (symbol-package parent-symbol)))
                             (read-from-string name)))
                   (url (when symbol
-                         (ps:ps (nyxt/ps:lisp-eval `(nyxt::describe-any ,(princ-to-string symbol)))))))
+                         (let ((old-indent ps:*indent-num-spaces*)
+                               (old-pring-pretty ps:*ps-print-pretty*))
+                           ;; Set those for all of Nyxt?
+                           (setf ps:*indent-num-spaces* 0
+                                 ps:*ps-print-pretty* nil)
+                           (ps:ps (nyxt/ps:lisp-eval `(nyxt::describe-any ,(princ-to-string symbol))))
+                           (setf ps:*indent-num-spaces* old-indent
+                                 ps:*ps-print-pretty* old-print-pretty)))))
              (spinneret:with-html-string
                (if url
                    (:a :href (javascript-url url)
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -104,20 +104,12 @@ If INPUT, narrow down to exact matches of it."
            (let* ((name (subseq target-string (1+ match-start) (1- match-end)))
                   (symbol (let ((*package* (symbol-package parent-symbol)))
                             (read-from-string name)))
-                  (definition (swank-backend:find-definitions symbol))
-                  (type (caaar definition))
-                  (url (if type
-                           (cond
-                             ((eq type 'defvar)
-                              (nyxt-url 'describe-variable :variable symbol))
-                             ((eq type 'defclass)
-                              (nyxt-url 'describe-class :class symbol))
-                             ((member type '(defun defmethod defgeneric))
-                              (nyxt-url 'describe-function :function symbol))
-                             (t nil)))))
+                  (url (when symbol
+                         (ps:ps (nyxt/ps:lisp-eval `(nyxt::describe-any ,(princ-to-string symbol)))))))
              (spinneret:with-html-string
                (if url
-                   (:a :href url (:code name))
+                   (:button :class "button" ; TODO: Can we make it look slimmer?  Maybe more like a link?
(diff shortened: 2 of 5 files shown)


aartaka:
1. Not sure about this. Will need to look at the code. Tomorrow :)

2. Yes, having universal documentation for everything sounds better than Nyxt-only documentation! [But I'm biased here.](https://discourse.atlas.engineer/t/nyxt-as-a-universal-documentation-viewer/278)

Ambrevar:
3\. Another question: is it possible to have links that call a JS with `lisp-eval`?  I tried, it didn't work.  I had to create a button for it to work, but it looks rather jarring as part of the docstring.

Ambrevar:
1\. I may have a better solution: in `describe-any` we could instantiate all the buffer sources _without_ running prompt-buffer, then count the number of suggestions across all these sources that match the input.  If it's one, don't bring up the prompt buffer, otherwise bring it up.

How does that sound?

aartaka:
> 1. I may have a better solution: in `describe-any` we could instantiate all the buffer sources _without_ running prompt-buffer, then count the number of suggestions across all these sources that match the input. If it's one, don't bring up the prompt buffer, otherwise bring it up.
> 
> How does that sound?

Yes! Sounds perfect!

aartaka:
3. I don't understand. Can you expand on what you're trying to do? What's the code that you tried running?

Ambrevar:
I tried this:

```lisp
(:a :href url (:code name))
```

Then clicking on the `execute-command` link in the `describe-command` description opens a new buffer with

```txt
Page could not be loaded.

URL: (function () { var request = fetch('lisp://%28DESCRIBE-ANY%20%22EXECUTE-COMMAND%22%29'); if (null) { __PS_MV_REG = []; return request.then(function (response) { return response.ok ? response.json() : null; }).then(function (data) { return null(data); }); }; })();
```

aartaka:
> I tried this:
> 
> ```lisp
> (:a :href url (:code name))
> ```
> 
> Then clicking on the `execute-command` link in the `describe-command` description opens a new buffer with
> 
> ```
> Page could not be loaded.
> 
> URL: (function () { var request = fetch('lisp://%28DESCRIBE-ANY%20%22EXECUTE-COMMAND%22%29'); if (null) { __PS_MV_REG = []; return request.then(function (response) { return response.ok ? response.json() : null; }).then(function (data) { return null(data); }); }; })();
> ```

Indeed, inline JavaScript code is not a link. To run it, you can prepend `javascript:` scheme before it :)

libraries/prompter/prompter.lisp
```diff
@@ -81,6 +81,12 @@ It's called after the sources are cleaned up.
 
 Note that the function is executed *before* performing an action.")
 
+     (auto-return-p nil
```
aartaka, line 84: Wow, I can use it for Vimium-like element hints!

source/help.lisp
```diff
@@ -32,26 +32,37 @@
...
-                                 :actions (list (make-command describe-class* (class)
-                                                  (describe-class :class class))))
-                  (make-instance 'slot-source
-                                 :actions (list (make-command describe-slot** (slot)
-                                                  (describe-slot :class (class-sym slot)
-                                                                 :name (name slot))))))))
+(define-command describe-any (&optional input)
+  "Inspect anything and show it in a help buffer.
+If INPUT, narrow down to exact matches of it."
+  (let ((preprocessor (if input
```
aartaka, line 38: Why it's `if input` there, but `(not (uiop:emptyp input))` below?

source/help.lisp
```diff
@@ -94,27 +105,20 @@
...
-                           (cond
-                             ((eq type 'defvar)
-                              (nyxt-url 'describe-variable :variable symbol))
-                             ((eq type 'defclass)
-                              (nyxt-url 'describe-class :class symbol))
-                             ((member type '(defun defmethod defgeneric))
-                              (nyxt-url 'describe-function :function symbol))
-                             (t nil)))))
+                  (url (when symbol
+                         (ps:ps (nyxt/ps:lisp-eval `(nyxt::describe-any ,(princ-to-string symbol)))))))
```
aartaka, line 109: A proper URL could be made with
``` lisp
(str:concat
 "javascript:" 
 (quri:url-encode
  (ps:ps (nyxt/ps:lisp-eval `(nyxt::describe-any ,(princ-to-string symbol))))))
```

libraries/prompter/prompter.lisp
Ambrevar: I was thinking of reverting this.  Should I keep it then?

source/help.lisp
Ambrevar: If `input` is not empty, then we set the preprocessor filter to match exactly _and_ we set the prompter to return automatically.

Does that make sense now?

Ambrevar:
> Indeed, inline JavaScript code is not a link. To run it, you can prepend `javascript:`
> scheme before it :)

Facepalm :P

source/help.lisp
aartaka: I mean, `auto-return-p` and `prompter:filter-exact-match` should go together, as far as I understant. But then, when `input` is `""`, the filter will be `prompter:filter-exact-match`, and `auto-return-p` will be false, which is somewhat unexpected.

Ambrevar:
I've followed your suggestion and it works in 6fd7a20f0ce73ddea5c5c8e095810723692503d1.  Sadly the formatting is off, the link is too large and has prepended spaces.  Any idea?

libraries/prompter/prompter.lisp
aartaka: It depends on what you came up with instead of `auto-return-p` :)

I mean, Vimium-like hints would still work without this, but it woulnd't be generalizeable. Having some prompter API to automatically select good candidatess would be nice.

Ambrevar:
See the latest commit for a for of point 1.

I couldn't get the `ps` formatting trick to work.  Can you check it out, maybe push to this branch if you have a fix?  (Don't force push).

aartaka:
Does f8e3e183f4ffd846c7483a24548009801da8ffbe do the trick?

Ambrevar:
Nope :(

Ambrevar:
Oh, I know what's wrong!  We are now wrapping the transformed docstring into a `<pre>` to keep line breaks, but since the generated HTML has line breaks in the link, it's what creates this ugliness.

I can think of a few options:

- Don't use `pre` and insert `<br>` on line breaks.
- Tweak spinneret to generate one-liners when possible.

Thoughts?

Ambrevar:
Fix pushed!

OK to merge now?

aartaka:
OK to merge!

Ambrevar:
Merged with ee5e60b6e0df556ce5af7ddb04f5640bd465e753.

aartaka:
Thanks for this, Pierre!

PR #2068, opened 2022-01-28
Nyxt-native debug on error

aartaka:
This adds an ability to catch and process errors in Nyxt itself. Right now it's implemented as a prompt, but alternative designs are welcome (see Things to Discuss below).

# What's new
- `*debug-on-error*` variable to decide on whether to show debigging prompt in Nyxt.
  - `toggle-debug-on-error` command to toggle this variable.
- When `*debug-on-error*` is set and an error is caught, `with-protect` will pop us a prompt showing available restarts and the backtrace, freezing the thread until on of restarts is chosen.

# Things to discuss
- Should we catch conditions too? May be an overkill.
- Should we make the prompt resumable? Right now closing the prompt invokes `abort` restart, which may be too brutal of an action.
- Prompt may be a bad design, especially regarding the backtrace. It's not copyable too.
  - Maybe generate internal pages with restart buttons and backtraces (as discussed with @Ambrevar)? This will require storing all the pending conditions somewhere and manipulating those via nyxt/lisp URLs somehow.

# How to Test
- Run Nyxt with this patch.
- Execute `toggle-debug-on-error` command.
- Provoke some error.
  - For example, open `lisp-repl` and type `(error "some error text")` in it.
- Prompt should appear with a list of restarts and a backtrace.
- Choose the restart you like and enjoy the error being handled :)

### DIFF
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -629,6 +629,22 @@ evaluate in order."
             until (eq object :eof)
             collect (funcall (lambda () (eval object)))))))
 
+(defmethod prompter:object-attributes ((restart restart))
+  `(("Name" ,(restart-name restart))))
+
+(define-class restarts-source (prompter:source)
+  ((prompter:name "Restarts"))
+  (:export-class-name-p t)
+  (:export-predicate-name-p t))
+
+(defun debug-prompt (condition)
+  (handler-case
+      (prompt1 :prompt (format nil "~a" condition)
+        :sources (list (make-instance 'restarts-source
+                                      :constructor (compute-restarts condition))))
+    (nyxt-prompt-buffer-canceled () 'abort)
+    (error () 'abort)))
+
 (defun error-buffer (&optional (title "Unknown error") (text ""))
   (sera:lret* ((error-buffer (make-instance 'user-web-buffer)))
     (with-current-buffer error-buffer
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -631,14 +631,29 @@ evaluate in order."
 
 (define-class restarts-source (prompter:source)
   ((prompter:name "Restarts"))
-  (:export-class-name-p t)
-  (:export-predicate-name-p t))
+  (:export-class-name-p t))
 
+(define-class backtrace-source (prompter:source)
+  ((prompter:name "Backtrace"))
+  (:export-class-name-p t))
+
+;; TODO: Make it resumable somehow. Channels?
 (defun debug-prompt (condition)
   (handler-case
       (prompt1 :prompt (format nil "~a" condition)
         :sources (list (make-instance 'restarts-source
-                                      :constructor (compute-restarts condition))))
+                                      :constructor (compute-restarts condition))
+                       ;; FIXME: Maybe use internal pages instead of prompt-buffer?
+                       (make-instance 'backtrace-source
+                                      :constructor (serapeum:filter
+                                                    (lambda (line)
+                                                      (digit-char-p (aref line 0)))
+                                                    (str:split
+                                                     #\newline
+                                                     (with-output-to-string (s)
+                                                       (uiop:print-backtrace
+                                                        :stream s))
+                                                     :omit-nulls t)))))
     (nyxt-prompt-buffer-canceled () 'abort)
     (error () 'abort)))
 
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -645,6 +645,14 @@ evaluate in order."
     (nyxt-prompt-buffer-canceled () 'abort)
     (error () 'abort)))
 
+(define-command-global toggle-debug-on-error (&key (value nil value-provided-p))
+  "Toggle Nyxt-native debugging.
+
+See `*debug-on-error*'."
+  (if value-provided-p
+      (setf *debug-on-error* value)
+      (setf *debug-on-error* (not *debug-on-error*))))
+
 (defun error-buffer (&optional (title "Unknown error") (text ""))
   (sera:lret* ((error-buffer (make-instance 'user-web-buffer)))
     (with-current-buffer error-buffer
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -629,9 +629,6 @@ evaluate in order."
             until (eq object :eof)
             collect (funcall (lambda () (eval object)))))))
 
-(defmethod prompter:object-attributes ((restart restart))
-  `(("Name" ,(restart-name restart))))
-
 (define-class restarts-source (prompter:source)
   ((prompter:name "Restarts"))
   (:export-class-name-p t)
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -629,6 +629,9 @@ evaluate in order."
             until (eq object :eof)
             collect (funcall (lambda () (eval object)))))))
 
+(defmethod prompter:object-attributes ((restart restart))
+  `(("Name" ,(restart-name restart))))
+
 (define-class restarts-source (prompter:source)
   ((prompter:name "Restarts"))
   (:export-class-name-p t))
--- a/source/concurrency.lisp
+++ b/source/concurrency.lisp
@@ -16,31 +16,50 @@ When the condition is muffled, a warning is reported to the user as per
 FORMAT-STRING and ARGS.
 As a special case, the first `:condition' keyword in ARGS is replaced with the
 raised condition."
-  (alex:with-gensyms (c sub-c)
-    `(if *run-from-repl-p*
-         (handler-case (progn ,@body)
-           (nyxt-prompt-buffer-canceled ()
-             (log:debug "Prompt buffer interrupted")))
-         (ignore-errors
-          (handler-bind
-              ((error
-                 (lambda (,c)
-                   (declare (ignorable ,c))
-                   ,(let ((condition-index (position :condition args)))
-                      (flet ((new-args (condition condition-index &optional escaped-p)
-                               (if condition-index
-                                   (append (subseq args 0 condition-index)
-                                           (list (if escaped-p
-                                                     `(plump:encode-entities (princ-to-string ,condition))
-                                                     `,condition))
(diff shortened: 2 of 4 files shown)


source/help.lisp
```diff
@@ -629,6 +629,45 @@ evaluate in order."
...
+
+(define-class restarts-source (prompter:source)
+  ((prompter:name "Restarts"))
+  (:export-class-name-p t))
+
+(define-class backtrace-source (prompter:source)
+  ((prompter:name "Backtrace"))
+  (:export-class-name-p t))
+
+;; TODO: Make it resumable somehow. Channels?
```
Ambrevar, line 643: Does it make sense though?  Because then the continuation is lost.  Common Lisp does not have first class continuations and I don't know if `cl-cont` would cut it here.

source/help.lisp
```diff
@@ -629,6 +629,45 @@ evaluate in order."
...
+      (prompt1 :prompt (format nil "~a" condition)
+        :sources (list (make-instance 'restarts-source
+                                      :constructor (compute-restarts condition))
+                       ;; FIXME: Maybe use internal pages instead of prompt-buffer?
+                       (make-instance 'backtrace-source
+                                      :constructor (serapeum:filter
+                                                    (lambda (line)
+                                                      (digit-char-p (aref line 0)))
+                                                    (str:split
+                                                     #\newline
```
Ambrevar, line 655: This is `sera:lines`.

source/help.lisp
```diff
@@ -629,6 +629,45 @@ evaluate in order."
...
+    (nyxt-prompt-buffer-canceled () 'abort)
+    (error () 'abort)))
+
+(define-command-global toggle-debug-on-error (&key (value nil value-provided-p))
+  "Toggle Nyxt-native debugging.
+
+See `*debug-on-error*'."
+  (if value-provided-p
+      (setf *debug-on-error* value)
+      (setf *debug-on-error* (not *debug-on-error*))))
```
Ambrevar, line 669: Move the condition inside the value of the `setf`.

Ambrevar:
> Should we catch conditions too? May be an overkill.

In my experience, this breaks Nyxt in particular with channels / concurrency.  I don't know exactly why and I'm not sure whether `T` is meant to be caught generally speaking.

> Prompt may be a bad design, especially regarding the backtrace. It's not copyable too. 

We could at least make sure everything is copyable in the prompt.

Ambrevar:
Amazing, Artyom!

I'll test as soon as possible.  In the meantime, have you considered using these two standard configuration knobs? `*debugger-hook*` `*break-on-signals*`

`*debugger-hook` is used by Swank / Slynk to hook into the Lisp process.
Its value is

```lisp
(defun slynk-debugger-hook (condition hook)
  "Debugger function for binding *DEBUGGER-HOOK*."
  (declare (ignore hook))
  (handler-case
      (call-with-debugger-hook #'slynk-debugger-hook
                               (lambda () (invoke-sly-debugger condition)))
    (invoke-default-debugger ()
      (invoke-default-debugger condition))))
```

This way we would not have to touch `with-protect` and we would catch conditions everywhere, not just in Nyxt.  Thoughts?

jmercouris:
Does not appear to work for me. I verified the value of the variable. Not sure why it is not showing anything upon `(error "tomato")`.

aartaka:
@jmercouris, indeed, we catch errors in our Lisp REPL, so debugger is not invoked there. What you can try is providing an error-provoking code in the prompt of command like
```
(define-command-global eval-expression ()
  "Prompt for the expression and evaluate it, echoing result to the `message-area'."
  (let ((expression-string
          ;; Read an arbitrary expression. No error checking, though.
          (first (prompt :prompt "Expression to evaluate"
                         :sources (list (make-instance 'prompter:raw-source))))))
    ;; Message the evaluation result to the message-area down below.
    (echo "~S" (eval (read-from-string expression-string)))))
```

@Ambrevar, `*debugger-hook*` sounds good, but when I tried testing it, it wasn't firing for non-main threads, so I've sticked to `*debugger-hook*`-compatible handler... in `with-protect` :D 

`*break-on-signals*` catches too much stuff, even signals we already suppressed. It's not much help for us, I guess.

aartaka:
A new design for this patch: we now create internal pages with error information. This allows us to not resolve conditions until the user wants it, and show all the information in the format easy to copy and interact with.

aartaka:
Debugger is now set globally, and the debugging buffers are now closed once one of restarts in invoked. I believe this is complete (well, once I fix the CI).

Obviously, we can make Nyxt-native debugging even better:
- Provide debugging prompts in Lisp REPL, injecting `debug->html` into it.
- Introduce `debugger-mode` with `a` bound to `abort` restart and `c` bound to `continue` restart.
- Introduce the global storage of introspectable objects, so that both debugger and `help-mode` can benefit from it, alloving to introspect an arbitrary number of values.
  - What this means for our debugger: we can introspect objects in backtraces!

aartaka:
CI passes! Merging?

Ambrevar:
About `*debugger-hook*`, this is how Slynk binds it in SBCL:

```lisp
(defimplementation call-with-debugger-hook (hook fun)
  (let ((*debugger-hook* hook)
        #+#.(slynk-sbcl::sbcl-with-new-stepper-p)
        (sb-ext:*stepper-hook*
         (lambda (condition)
           (typecase condition
             (sb-ext:step-form-condition
              (let ((sb-debug:*stack-top-hint* (sb-di::find-stepped-frame)))
                (sb-impl::invoke-debugger condition)))))))
    (handler-bind (#+#.(slynk-sbcl::sbcl-with-new-stepper-p)
                   (sb-ext:step-condition #'sb-impl::invoke-stepper))
      (call-with-break-hook hook fun))))
```

So there might be some more work to do to use it properly.

source/help.lisp
```diff
@@ -629,6 +629,78 @@ evaluate in order."
...
+      (invoke-restart-interactively (calispel:? channel)))))
+
+(defun debug->html (condition id &optional restarts buffer delete-buffer-p)
+  "Produce HTML code for the CONDITION with RESTARTS."
+  (spinneret:with-html-string
+    (:h* (symbol-name (type-of condition)))
+    (:pre (format nil "~a" condition))
+    (:section
+     (loop for restart in restarts
+           for i from 0 by 1
```
Ambrevar, line 668: `by 1` is redundant, isn't it?

source/help.lisp
```diff
@@ -629,6 +629,78 @@ evaluate in order."
             until (eq object :eof)
             collect (funcall (lambda () (eval object)))))))
 
+(defvar *debug-conditions* (make-hash-table)
+  "A hash-table from numeric condition ID to the (CONDITION RESTARTS CHANNEL) lists.
+
+- CONDITION is the signaled condition.
+- RESTARTS is the list of restarts computed for this CONDITION.
+- CHANNEL is the channel to send the user-chosen restart at.")
```
Ambrevar, line 637: This asks for a class! :)

source/help.lisp
```diff
@@ -629,6 +629,78 @@ evaluate in order."
...
+    (:pre (format nil "~a" condition))
+    (:section
+     (loop for restart in restarts
+           for i from 0 by 1
+           collect (:button :class "button"
+                            :onclick (ps:ps (nyxt/ps:lisp-eval
+                                             `(progn
+                                                (calispel:!
+                                                 (third (gethash ,id *debug-conditions*))
+                                                 (nth ,i (second (gethash ,id *debug-conditions*))))
```
Ambrevar, line 674: `third` and `second` make the code hard to read here.  With a  class everything would be crystal clear!

Also let-bind the result of `gethash` to avoid calling it twice.

Ambrevar (commented):
A few more nits and we are good to go!

aartaka:
> About `*debugger-hook*`, this is how Slynk binds it in SBCL:
> 
> ```lisp
> (defimplementation call-with-debugger-hook (hook fun)
>   (let ((*debugger-hook* hook)
>         #+#.(slynk-sbcl::sbcl-with-new-stepper-p)
>         (sb-ext:*stepper-hook*
>          (lambda (condition)
>            (typecase condition
>              (sb-ext:step-form-condition
>               (let ((sb-debug:*stack-top-hint* (sb-di::find-stepped-frame)))
>                 (sb-impl::invoke-debugger condition)))))))
>     (handler-bind (#+#.(slynk-sbcl::sbcl-with-new-stepper-p)
>                    (sb-ext:step-condition #'sb-impl::invoke-stepper))
>       (call-with-break-hook hook fun))))
> ```
> 
> So there might be some more work to do to use it properly.

Yes, that's a proper SBCL solution there. Notice, though, that all the stepper-related code is for the purpose of stepping into functions, which is not immediately necessary for a simple debugger like ours. Setting a `*debugger-hook*` and calling a `swank-backend:install-debugger-globally` (or just the latter, at that matter) should set all the basic debugging handlers for SBCL (`sb-ext:*invoke-debugger-hook*`), CCL (`ccl:*break-hook*` and `ccl:*select-interactive-process-hook*`), or whatever implementation there is, in a portable manner.

aartaka:
Other than that, this patch seems complete :)

source/help.lisp
```diff
@@ -629,12 +629,28 @@ evaluate in order."
...
 
-(defvar *debug-conditions* (make-hash-table)
-  "A hash-table from numeric condition ID to the (CONDITION RESTARTS CHANNEL) lists.
+(define-class condition-handler ()
+  ((condition-itself
+    (error "condition-handler should always wrap a condition.")
+    :type condition
+    :documentation "The condition itself.")
+   (restarts
+    nil
```
Ambrevar, line 638: Nit: I like to write it `'()` instead to emphasize that it's a list and not a boolean.

Ambrevar:
Feel free to merge then!

source/help.lisp
aartaka: Done :)

PR #2070, opened 2022-01-29
Lisp scheme: send `null` instead of no response at all?

shamazmazum:
Hi! Working on bookmarks I spotted a minor bug. When you set log level to `:debug` and try to remove a bookmark using `*Bookmarks*` buffer you'll see this in the error output stream:

~~~~
<DEBUG> [18:30:40] Error when evaluating lisp URL: The value
		                                     NIL
                                                   is not of type
                                                     ARRAY
~~~~

This patch removes these errors by sending `null` response when a result of call to lisp cannot be encoded as a JSON object.

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -397,10 +397,11 @@ guarantee of the same result."
                         ;; supports, then we can override the encoding and
                         ;; decoding methods and allow arbitrary objects (like
                         ;; buffers) in the nyxt:// URL arguments..
-                        (when (or (scalar-p result)
-                                  (and (sequence-p result)
-                                       (every #'scalar-p result)))
-                          (cl-json:encode-json-to-string result)))
+                        (cl-json:encode-json-to-string
+                         (when (or (scalar-p result)
+                                   (and (sequence-p result)
+                                        (every #'scalar-p result)))
+                           result)))
                       "application/json"))
             (values "undefined" "application/json;charset=utf8"))))
   :cors-enabled-p t


shamazmazum:
Maybe you may wish to replace `null` with something else like `"unrepresentable"` because `NIL` may be a valid value returned by a lisp function.

source/urls.lisp
```diff
@@ -397,10 +397,11 @@ guarantee of the same result."
                         ;; supports, then we can override the encoding and
                         ;; decoding methods and allow arbitrary objects (like
                         ;; buffers) in the nyxt:// URL arguments..
-                        (when (or (scalar-p result)
-                                  (and (sequence-p result)
-                                       (every #'scalar-p result)))
-                          (cl-json:encode-json-to-string result)))
+                        (cl-json:encode-json-to-string
+                         (if (or (scalar-p result)
```
Ambrevar, line 401: Should be `when`.

Ambrevar:
Makes sense to me, but I'll leave it to @aartaka to confirm! :)

source/urls.lisp
shamazmazum: Ok, replaced with `when`.

aartaka:
Yes, `null` should be fine, given that it's falsy and is a literal.

An alternative would be to return `undefined` for consistency with the clause when the evaluation is not allowed, but having different values can be useful down the road.

Merging this :)

PR #2071, opened 2022-01-30
Group bookmarks according to tags.

shamazmazum:
Hi. I decides to improve `list-bookmarks` command a bit. Bookmarks are now grouped by tags in this manner:
20220130_09h10m50s_grim

When you click on a button, a list of bookmarks under this tags expands:
20220130_09h10m56s_grim

I think it can help a lot in organizing bookmarks.

### DIFF
--- a/source/mode/bookmark.lisp
+++ b/source/mode/bookmark.lisp
@@ -1,0 +1,72 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(uiop:define-package nyxt/bookmark-mode
+  (:use #:cl #:nyxt)
+  (:import-from #:serapeum #:export-always)
+  (:documentation "Bookmark buffer mode and command"))
+(in-package :nyxt/bookmark-mode)
+
+(export-always 'bookmark-mode)
+(define-mode bookmark-mode ()
+  "Mode for the bookmarks buffer."
+  ((rememberable-p nil)
+   (style (theme:themed-css (theme *browser*)
+            ("summary"
+             :background-color theme:secondary
+             :color            theme:background
+             :font-size        "14px"
+             :padding          "16px"
+             :margin           "6px"
+             :width            "100%"
+             :border           "none"
+             :outline          "none"
+             :text-align       "left")))))
+
+(defun group-bookmarks (buffer)
+  (let ((bookmarks-table (make-hash-table :test #'equalp)))
+    (with-data-unsafe (bookmarks (bookmarks-path buffer))
+      (dolist (bookmark bookmarks)
+        (let ((tags (tags bookmark)))
+          (if tags
+              (dolist (tag tags)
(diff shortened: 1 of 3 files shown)


Ambrevar:
Nice!

I'm OK with merging this now, but it's also a very tempting opportunity to go even further!
We could have an HTML table with clickable buttons to sort, filter, etc.

What do you think?

shamazmazum:
I can try to add a button which sorts bookmarks either in alphabetical order or by date. I hope it's not very hard (I know a little about web development). If I fail to do so, I'll let you know :)

Ambrevar:
Thanks!
Don't sweat too much on it, this is by all means a great improvement
already, we can merge it as it is if anything else fails.

source/mode/bookmark.lisp
```diff
@@ -0,0 +1,87 @@
...
+  (let ((bookmarks (group-bookmarks bookmarks-buffer)))
+    (spinneret:with-html-string
+      (:style (style (find-mode bookmarks-buffer 'bookmark-mode)))
+      (:h1 "Bookmarks")
+      (:body
+       (if (zerop (hash-table-count bookmarks))
+           (format nil "No bookmarks in ~s." (expand-path (bookmarks-path bookmarks-buffer)))
+           (maphash
+            (lambda (tag bookmarks)
+              (:button :class "collapsible"
```
aartaka, line 56: Maybe use `<details>` tag instead?

source/mode/bookmark.lisp
```diff
@@ -0,0 +1,87 @@
...
+              (push bookmark (gethash tags bookmarks-table nil))))))
+    bookmarks-table))
+
+(export-always 'list-bookmarks)
+(define-internal-page-command-global list-bookmarks ()
+    (bookmarks-buffer "*Bookmarks*" 'bookmark-mode)
+  "List all bookmarks in a new buffer."
+  (let ((bookmarks (group-bookmarks bookmarks-buffer)))
+    (spinneret:with-html-string
+      (:style (style (find-mode bookmarks-buffer 'bookmark-mode)))
```
aartaka, line 49: I wonder if we should do that by default in `define-internal-page-command` :thinking:

shamazmazum:
@aartaka Changed javascript to `<details>`. Never heard about this tag before.

If you don't mind, can you merge it please? I wanted to do sorting, but I don't have time for this task right now. Maybe later.

20220203_19h32m11s_grim

source/mode/bookmark.lisp
shamazmazum: Do you mean `define-internal-page-command` should be used instead of `define-internal-page-command-global`? I cannot see any example how the first can be used.

aartaka: No, I meant the family of commands, including both global and non-global ones. Sorry if it was confusing :)

Let's leave it the way it is now and abstract it if more cases of mode-required styling emerge.

aartaka:
Sure! Thanks for this, @shamazmazum!

Issue #2020, opened 2022-01-14
Search terms are deleted in visual-mode after searching in a page.


taiju:
Search terms are deleted in visual-mode after searching in a page. From #2015 

**Steps to Reproduce**

1. Open https://nyxt.atlas.engineer/ in Nyxt.
2. Execute the search-buffer command to search for "nyxt".
3. Execute visual-mode command.

**Expected behavior**

Able to use visual-mode, including search terms.

**Actual behavior**

Search terms are removed when switching to visual-mode.

**Information**
- OS name+version: GNU Guix
- Graphics card and driver: AMD Renoir
- Desktop environment / Window manager name+version: GNOME 40.4
- How you installed Nyxt (Guix pack, package manager, build from source): guix install
- Information from `show-system-information`: 

```
Nyxt version: 2.2.3
Renderer version: GI-GTK
Operating system kernel: Linux 5.15.13
Lisp implementation: SBCL 2.1.9 (Dynamic space size: 3221225472)
Features: (:WEBKIT2 :WEBKIT2-2.34 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
 :GLIB-2-30 :GLIB-2-32 :GLIB-2-34 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42
 :GLIB-2-44 :GLIB-2-46 :GLIB-2-48 :GLIB-2-50 :GLIB-2-52 :GLIB-2-54 :GLIB-2-56
 :GLIB-2-58 :GLIB :NYXT-2 :FSET-EXT-STRINGS :CUSTOM-HASH-TABLE-NATIVE :SWANK
[...]

PR #2073, opened 2022-01-31
Make it possible to choose whether to keep or remove search hints.

taiju:
Hi,

I've added an implementation that clears search hints when the prompt is closed, based on the exchange in #2020.
This implementation may already be unnecessary, but I would appreciate your feedback.

Thanks,

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -72,7 +72,7 @@ Auto-completions come from the default search engine.")
    (keep-search-hints-p
     t
     :type boolean
-    :documentation "Whether to keep search hints when the search prompt is closed.")
+    :documentation "Whether to keep search hints when the search prompt for the `search-buffer' command is closed.")
    (search-engines
     (list (make-instance 'search-engine
                          :shortcut "wiki"
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -72,7 +72,7 @@ Auto-completions come from the default search engine.")
    (keep-search-hints-p
     t
     :type boolean
-    :documentation "Whether to keep search hints when the prompter is closed.")
+    :documentation "Whether to keep search hints when the search prompt is closed.")
    (search-engines
     (list (make-instance 'search-engine
                          :shortcut "wiki"
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -69,6 +69,10 @@ after the mode-specific hook.")
     :type boolean
     :documentation "Whether auto-completion works even for non-prefixed search.
 Auto-completions come from the default search engine.")
+   (keep-search-hints-p
+    t
+    :type boolean
+    :documentation "Whether to keep search hints when the prompter is closed.")
    (search-engines
     (list (make-instance 'search-engine
                          :shortcut "wiki"
--- a/source/mode/search-buffer.lisp
+++ b/source/mode/search-buffer.lisp
@@ -187,8 +187,8 @@
                             :case-sensitive-p case-sensitive-p
                             :actions (list (lambda (search-match)
                                              (unless (keep-search-hints-p (current-buffer))
-                                               (remove-search-hints)
-                                               search-match)))))))
+                                               (remove-search-hints))
+                                             search-match))))))
 
 (define-command search-buffers (&key case-sensitive-p)
   "Search multiple buffers."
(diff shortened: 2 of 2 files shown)


source/mode/search-buffer.lisp
```diff
@@ -164,6 +164,8 @@
                                      (prompt-buffer-selection-highlight-hint :scroll t)))
    (prompter:destructor (lambda (prompter source)
                           (declare (ignore prompter source))
+                          (unless (slot-value (current-buffer) 'keep-search-hints-p)
```
Ambrevar, line 167: Replace the `slot-value` call with `(keep-search-hints-p (current-buffer))`.

source/mode/search-buffer.lisp
```diff
@@ -182,7 +184,11 @@
    :prompt "Search text"
    :sources (list
              (make-instance 'user-search-buffer-source
-                            :case-sensitive-p case-sensitive-p))))
+                            :case-sensitive-p case-sensitive-p
+                            :actions (list (lambda (search-match)
+                                             (unless (slot-value (current-buffer) 'keep-search-hints-p)
```
Ambrevar, line 189: Same.

Ambrevar:
@aartaka What do you think?

source/mode/search-buffer.lisp
taiju: Thank you for your advice.
I replaced it.
322270cbc6cf697a06213a61e42cc092b605dfce

source/mode/search-buffer.lisp
taiju: Same.
322270cbc6cf697a06213a61e42cc092b605dfce

source/mode/search-buffer.lisp
```diff
@@ -182,7 +184,11 @@
    :prompt "Search text"
    :sources (list
              (make-instance 'user-search-buffer-source
-                            :case-sensitive-p case-sensitive-p))))
+                            :case-sensitive-p case-sensitive-p
+                            :actions (list (lambda (search-match)
+                                             (unless (keep-search-hints-p (current-buffer))
+                                               (remove-search-hints)
+                                               search-match)))))))
```
aartaka, line 191: Maybe return `search-match` unconditionally? If `keep-search-hints-p` is non-`nil`, this will return `nil` instead of the found search match.

source/buffer.lisp
```diff
@@ -69,6 +69,10 @@ after the mode-specific hook.")
     :type boolean
     :documentation "Whether auto-completion works even for non-prefixed search.
 Auto-completions come from the default search engine.")
+   (keep-search-hints-p
+    t
+    :type boolean
+    :documentation "Whether to keep search hints when the prompter is closed.")
```
aartaka, line 75: Maybe call it "search prompt"? `prompter` is the name of the underlying library, it can cause confusion for those reading the documentation :)

aartaka, line 75: Another thing: maybe mention `search-buffer` in the docstring so that it's understandable where this is used.

And mention this slot in the documentation of search commands, so that this toggle is discoverable enough :)

aartaka (commented):
It's doing exactly what's necessary. Thanks, @taiju! I've got some minor comments, mostly regarding documentation. Let's discuss those and we can merge :)

source/buffer.lisp
taiju: I modified it.
e7937cf849e272d4654fc3667aefc6ae33c04ac9

source/mode/search-buffer.lisp
taiju: Sorry, that was just an error.
I have corrected it.
88140ef278b0323b4e97f76c272baa1658fb1e19

source/buffer.lisp
taiju: Thank you. I got it.
I'm out of time, so it may be tomorrow.

taiju: Fixed. However, I am not good at English, so if there is a problem with the fix, I would appreciate it if you could correct it.
d78cdd774d653b19e136d0ce2fb6e31f476a9d1a
a112ab8d8fc3de09e083554f23205808e1577077

aartaka:
Looks just fine, thank you @taiju!

taiju:
@aartaka Thanks for your review and merge!
I learned a lot from it!

Issue #1302, opened 2021-04-16
Auto refresh list-buffers buffer on changes


Ambrevar:
A draft was submitted in #1035.

jmercouris:
Why would we want auto refresh? I don't agree with this premise. Everything should only occur on-demand, in my opinion. Maybe I am too influenced by the behavior of emacs, but that is what I think.

Ambrevar:
I think presenting the user with an outdated state of things is confusing.
Emacs does not shine with its accessibility and I believe Nyxt should do better here.

jmercouris:
Maybe. I'll need more time with this. Perhaps a practical implementation will change my mind.

aadcg:
> Maybe I am too influenced by the behavior of emacs, but that is what I think.

What happens when you kill a buffer from `list-buffers` or `ibuffer` in Emacs?

jmercouris:
when I kill it from this menu, something happens, when it is killed from elsewhere, nothing happens. It would be no issue to wrap each delete button on `list-buffers` with `(progn (delete-buffer xyz) (list-buffers))`. I think Pierre is suggesting something different than what your original PR sought to achieve!

Ambrevar:
I'm suggesting that the `list-buffers` buffer watches on `(buffers
*browser*)` and automatically refreshes on every change.

aadcg:
I see. No strong opinion. But we all agree that deleting buffers from `list-buffers` should update it.

PhilHudson:
OT: How to 'fix' ibuffer:

    (add-hook 'ibuffer-mode-hook #'ibuffer-auto-mode)

Useful now.

PR #2077, opened 2022-02-03
To delete a buffer, remove it from the list.

taiju:
The behavior when deleting buffers with list-buffers was bothering me, so I improved it.
However, after I finished writing this patch, I noticed the following issue.

#1302

If it does not match the discussion in this issue, I will withdraw it.


taiju:
There was a problem with this implementation, so the pull request is withdrawn.

Issue #2078, opened 2022-02-04
The callback specified in lisp-eval does not work.


taiju:
**Describe the bug**

The callback specified in lisp-eval does not work.

This is probably because the response is being rejected by CORS.

**Precise recipe to reproduce the issue**

- Run the list-buffers command
- Run the open-inspector command
- Delete a buffer with the `✕` button in the list-buffers window.

The fetch function of lisp-eval called in this procedure will result in an error.
Therefore, Promise cannot be chained and the function specified in the callback will not be called.

A screenshot is attached.

![Screenshot from 2022-02-04 17-09-20](https://user-images.githubusercontent.com/95721/152498533-0acb0f5b-c294-4d29-a574-4ae3603bb4f9.png)

**Information**
*OS*: GNU Guix
*Graphics*: AMD® Renoir
*Window Manager*: GNOME 40.4
*Installation*: Build from source

```
Nyxt version: 2.1.1-1552-gaa777272
Renderer version: GI-GTK
Operating system kernel: Linux 5.15.18
Lisp implementation: SBCL 2.1.9 (Dynamic space size: 3221225472)
Features: (:WEBKIT2 :WEBKIT2-2.34 :WEBKIT2-CORS-ALLOWLIST :WEBKIT2-PASTE-PLAINTEXT
 :WEBKIT2-TRACKING :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA
 :WEBKIT2-SANDBOXING :GTK-3-22 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14
 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20
 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4
 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF :GLIB-2-30 :GLIB-2-32 :GLIB-2-34
 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42 :GLIB-2-44 :GLIB-2-46 :GLIB-2-48
 :GLIB-2-50 :GLIB-2-52 :GLIB-2-54 :GLIB-2-56 :GLIB-2-58 :GLIB :NYXT-UNSTABLE
[...]

PR #2079, opened 2022-02-04
Add `no-cors' option to the fetch function.

taiju:
Hi,

When I wrote the code to pass a callback to the lisp-eval function, I found that it did not work, so I added the following issue.

#2078

I wrote a patch to avoid this problem by changing the request mode of the fetch function used internally to `no-cors`.

This is an option that should be avoided as much as possible, but for this application, I think it is safe to specify it.

The best solution is to give the `Access-Control-Allow-Origin` header if the internal page can give a header to the response.
However, I could not find a way to do that.

Thank you,

### DIFF
--- a/source/parenscript-macro.lisp
+++ b/source/parenscript-macro.lisp
@@ -73,7 +73,8 @@
 (defpsmacro lisp-eval (form &optional callback)
   "Request the lisp: URL and invoke callback when there's a successful result."
   `(let ((request (fetch (lisp (str:concat
-                                "lisp://" (quri:url-encode (write-to-string ,form)))))))
+                                "lisp://" (quri:url-encode (write-to-string ,form))))
+                         (create :mode "no-cors"))))
      (when ,callback
        (chain request
               (then (lambda (response)


aartaka:
There should be a way to do that with WebKit-provided primitives, without altering the `fetch` options...

taiju:
I see. That area might be a little difficult for me...

taiju:
Could it be controlled by the way the internal-scheme is defined?
I just haven't been able to get it to work yet.

aartaka:
> Could it be controlled by the way the internal-scheme is defined?
> I just haven't been able to get it to work yet.

Yes, it could. But wait, does it make the Lisp-side action?

aartaka:
If the Lisp-side action happens, then `nyxt/ps:lisp-eval` does the main thing it's intended for -- Lisp code evaluation. If it doesn't run the callback afterwards, then it's strange, but a minor annoyance. We can run any code after invoking `nyxt/ps:lisp-eval` and independently of its result. It will be unsynced with the `lisp:` URL response, yet should work fine in most cases.

aartaka:
Or do buttons do nothing too?

taiju:
Yes, as you said, it is usually not a problem.
This can be a problem if we want to wait for the process to be completed.
For example, the following code.

```diff
diff --git a/source/mode/buffer-listing.lisp b/source/mode/buffer-listing.lisp
index 178af062..77b96aa4 100644
--- a/source/mode/buffer-listing.lisp
+++ b/source/mode/buffer-listing.lisp
@@ -42,7 +42,8 @@ With LINEAR-VIEW-P, list buffers linearly instead."
                (spinneret:with-html
                  (:p (:button :class "button"
                               :onclick (ps:ps (nyxt/ps:lisp-eval
-                                               `(nyxt::delete-buffer :id ,(id buffer)))) "✕")
+                                               `(nyxt::delete-buffer :id ,(id buffer))
+					       (lambda () (ps:chain location (reload))))) "✕")
                      (:button :class "button"
                               :onclick (ps:ps (nyxt/ps:lisp-eval
                                                `(nyxt::switch-buffer :id ,(id buffer)))) "→")
```

This is the code to reload the screen after deleting a buffer in the buffer list screen.
If we do not use a callback, the reload process will be executed before the buffer is deleted.
As a result, even if the buffer has been deleted, it will still be visible in the buffer list.

aartaka:
OK, after some thought, this seems to be a reasonable fix. Sorry to keep you waiting for so long, @taiju!

taiju:
I would have liked to have looked into other ways of doing things a little more, but I couldn't find the time to move on.
Thank you for the merge!

Issue #1876, opened 2021-11-17
Leverage cl-tld to validate domains


Ambrevar:
In particular in `input->queries` and `valid-url-p` to get more procise URL / domain validation without expensive system calls.

https://github.com/lu4nx/cl-tld

guojing0:
This seems interesting; I will look into it.

Do you mean "to get more **precise** URL ..."?

Ambrevar:
Indeed! :)

hendursaga:
If/when this gets finished, it might be a good idea to request adding Nyxt to the [Public Suffix projects' list](https://publicsuffix.org/learn/).

hendursaga:
I'm not sure how to assign myself, but I'm working on this right now. @Ambrevar when you say "without expensive system calls" are you meaning to search for and remove the `check-dns-p` logic, or at least set it to `NIL` by default, in arguments?

Ambrevar:
Something like that, yes.

`check-dns-p` would still be useful because it would check for valid
domains, while cl-tld only allows to check for valid tlds, which is more
limited.

So we would proceed like we do now:

- 1st pass: syntax check using cl-tld.
  This yields a bunch of prompt buffer suggestions, possibly with false positives.
- 2nd pass: DNS checks to filter out the false positives.

PR #2081, opened 2022-02-06
Add cl-tld dependency, use it in valid-tld-p

hendursaga:
This partially solves #1876, there might be other places in the code, such as `input->queries` where `valid-tld-p` would be good to use.


hendursaga:
I'm afraid I didn't manually run the tests on my machine. It appears perhaps we might have to tweak a few of them?

source/urls.lisp
```diff
@@ -153,6 +153,12 @@ For keyword arguments' meaning, see `scheme' slot documentation."
     (iolib/sockets:lookup-hostname name)
     (t () nil)))
 
+(export-always 'valid-tld-p)
+(defun valid-tld-p (name)
+  "Return NIL if hostname NAME does not include a valid TLD as determined by the
+Public Suffix list, T otherwise."
+  (and (cl-tld:get-tld name) T))
```
Ambrevar, line 160: You can replace this with `sera:true`.

source/urls.lisp
```diff
@@ -153,6 +153,12 @@ For keyword arguments' meaning, see `scheme' slot documentation."
     (iolib/sockets:lookup-hostname name)
     (t () nil)))
 
+(export-always 'valid-tld-p)
+(defun valid-tld-p (name)
+  "Return NIL if hostname NAME does not include a valid TLD as determined by the
```
Ambrevar, line 158: Replace NAME with HOSTNAME in docstring and in arg.

Ambrevar:
The tests fail because of this:

```txt
 Parse URL
    ✓ full URL 

    ✓ URL without protocol (92006ms)

    ✓ search engine 

    ✓ search engine with special characters 

    ✓ default search engine 

    ✓ wiki search engine (4000ms)

    ✓ local file 

    × empty domain
Raised an error not is a domain (expected: #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/?q=foo>)
    × same domain and TLD
Raised an error not is a domain (expected: #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/?q=algo>)
    × localhost
Raised an error not is a domain (expected: #<QURI.URI.HTTP:URI-HTTP http://localhost:8080>)
    × ignore wildcards
Raised an error not is a domain (expected: #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/?q=*spurious*>)
    ✓ about:blank 

    ✓ valid syntax but unknown scheme 

 URL processing
    × Invalid URL (empty host)
Raised an error not is a domain (expected: NIL)
; 
; compilation unit aborted
;   caught 1 fatal ERROR condition
    × Invalid URL (TLD == host)
Raised an error not is a domain (expected: NIL)
...
```

My guess is that you are not catching errors in `valid-tld-p`.

You also need to add cl-tld to `build-scripts/nyxt.scm`.

jmercouris:
@Ambrevar does `cl-tld` exist in Guix?

hendursaga:
@jmercouris https://guix.gnu.org/en/packages/cl-tld-0.1-2.f5014da/ seems to suggest so..? Perhaps the test is just stalling for network reasons?

hendursaga:
https://git.savannah.gnu.org/cgit/guix.git/commit/?id=342f3de7428759ae98b4a5f6e45bdadf49d0bd85 confirms it, there's an alias from `cl-tld` to `sbcl-cl-tld` and, SURPRISE, Pierre was the one who committed :)

PR #2093, opened 2022-02-10
Quick Lisp extension installation.

aartaka:
This is a cleaned-up version of #2002. Only Lisp extension are supported now, WebExtensions installation dropped for implementation simplicity. @Ambrevar, your review from #2002 is addressed.

Thoughts?

### DIFF
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -834,3 +834,52 @@ System information is also saved into the clipboard."
                              (:p (:pre (documentation (fn command) t)))
                              (:pre :class "nyxt-source" (:code (let ((*print-case* :downcase))
                                                                  (write-to-string (sexp command)))))))))))
+
+(defvar *git* "git")
+
+(define-command install-lisp-extension (&optional (url (url (current-buffer))))
+  "Install the Lisp extension for Nyxt from the Git repository at URL.
+
+Requirements:
+- git should be installed."
+  (let* ((name (alex:lastcar (str:split "/" (quri:uri-path url))))
+         (install-path (str:concat (expand-path *extensions-path*) name))
+         (config-path (make-instance 'init-data-path
+                                     :dirname (dirname *init-file-path*)
+                                     :basename name)))
+    (if (uiop:directory-exists-p install-path)
+        (echo "Target directory ~s already exists, not installing." install-path)
+        (handler-case
+            (progn
+              (uiop:run-program
+               (list *git* "clone" "--recurse-submodules" (render-url url) install-path))
+              (let* ((asd (first
+                           (sort (sera:filter (lambda (file) (string= "asd" (pathname-type file)))
+                                              (uiop:directory-files install-path))
+                                 #'< :key #'length)))
+                     (load-asd (asdf:load-asd asd))
+                     (primary-system (first
+                                      (delete-if
+                                       (complement
+                                        (alex:conjoin
+                                         #'asdf:primary-system-p
+                                         (lambda (system)
+                                           (uiop:pathname-equal asd (asdf:system-source-file
+                                                                     (asdf:find-system system))))))
+                                       (asdf:registered-systems)))))
+                (declare (ignore load-asd))
+                (if primary-system
+                    (progn
+                      (alex:write-string-into-file
+                       (format nil "(in-package :nyxt-user)
+
+;; Put your configuration for ~a here.
+" primary-system)
+                       (expand-path config-path)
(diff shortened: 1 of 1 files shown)


source/help.lisp
```diff
@@ -834,3 +834,52 @@ System information is also saved into the clipboard."
...
+(define-command install-lisp-extension (&optional (url (url (current-buffer))))
+  "Install the Lisp extension for Nyxt from the Git repository at URL.
+
+Requirements:
+- git should be installed."
+  (let* ((name (alex:lastcar (str:split "/" (quri:uri-path url))))
+         (install-path (str:concat (expand-path *extensions-path*) name))
+         (config-path (make-instance 'init-data-path
+                                     :dirname (dirname *init-file-path*)
+                                     :basename name)))
```
Ambrevar, line 849: Should be `auto-config.lisp` instead, no?

source/help.lisp
```diff
@@ -834,3 +834,52 @@ System information is also saved into the clipboard."
...
+                                        (alex:conjoin
+                                         #'asdf::primary-system-p
+                                         (lambda (system)
+                                           (uiop:pathname-equal asd (asdf:system-source-file
+                                                                     (asdf:find-system system))))))
+                                       (asdf:registered-systems)))))
+                (declare (ignore load-asd))
+                (if primary-system
+                    (progn
+                      (alex:write-string-into-file
```
Ambrevar, line 873: If auto-config.lisp, then no need for this manual write, see `append-configuration`.

source/help.lisp
```diff
@@ -834,3 +834,52 @@ System information is also saved into the clipboard."
...
+                       (format nil "(in-package :nyxt-user)
+
+;; Put your configuration for ~a here.
+" primary-system)
+                       (expand-path config-path)
+                       :if-does-not-exist :create
+                       :if-exists nil)
+                      (nyxt::append-configuration
+                       `(load-after-system ,primary-system ,(expand-path config-path))))
+                    (echo-warning "Cannot install extension: primary ASDF system is not found."))))
```
Ambrevar, line 883: Nit: you can remove the "is".

Ambrevar: I would add an argument to the command to optianally ask for the primary system name, this way the user can still install the extension.  On error, we could list all the systems found in the file, this way the user can recall the command, this time specifying the right system name.

source/help.lisp
aartaka: No :p 

This is the name of the extenison-specific config file. For instance, `nx-fruit` would have `nx-fruit.lisp`.

source/help.lisp
aartaka: But auto-config.lisp doesn't have all the extension systems installed, does it? That's the whole reason we have `load-after-system` -- we split config into files dependent on loading a certain system.

aartaka:
All addressed, although I'm not fully satisfied with how it behaves. Right now it does nothing if checkout directory does already exist. Should we overwrite instead?

Ambrevar:
When the directory exists, I think we should prompt the user:

- overwrite
- ignore
- update existing (e.g. with `git -C /path/to/repo pull`)

Ambrevar:
The CI fails with 

```log

; wrote /home/runner/.cache/common-lisp/sbcl-2.2.0-linux-x64/home/runner/work/nyxt/nyxt/source/global-tmp8WGMHWCD.fasl
; compilation finished in 0:00:00.012
; 
; compilation unit aborted
;   caught 1 fatal ERROR condition
;   caught 1 WARNING condition
caught error #<SIMPLE-ERROR "~@<~S is an already defined constant whose value ~
                              ~S is not equal to the provided initial value ~S ~
                              under ~S.~:@>" {1005610403}>
;   caught 60 STYLE-WARNING conditions
NYXT:+VERSION+ is an already defined constant whose value "2.2.4" is not equal
;   printed 3887 notes
to the provided initial value "2b0e2b3" under #<FUNCTION EQUAL>.
Error: Process completed with exit code 17.
```

Any idea why?

aartaka:
No, I'm terribly puzzled by this 0_o Maybe some compile-time magic?

aartaka:
Closing this for later.

Ambrevar:
Do you have a copy of this somewhere?  Or some doc?  We should not forget about it :)

aartaka:
Nope, I only have the code on this branch. I'll not delete it, for one. I'll open an issue, so that it's more visible :)

PR #2095, opened 2022-02-11
source/bookmark.lisp: fix bug to get multiple buffers.

pdelfino:
This seems to fix PR #2094. Now, Nyxt stores the webpages from multiple buffers that were selected.

The UI is still a little bit uncomfortable for the user since the prompt-buffer asks in sequence which tags to attribute without a clear indication of which webpage are the tags being bind to.


aartaka:
Oh, sorry, haven't seen this. But you're right with the fix, thanks! 

I'll close this, as `master` has the same fix (sorry again, I was unaware of the PR **<(_ _)>**).

PR #2096, opened 2022-02-11
Introduce paren auto-completion and editing keybindings to Lisp REPL.

aartaka:
This minor patch adds editing command and `auto-paren`-like behavior to our REPL.

# What's new
- `paren` and `closing-paren` commands of `repl-mode` that (respectively) insert "()" on opening paren input and insert as much closing parens as it is necessary to close the expression before the cursor.
- `input-edit-mode` commands for editing.

# Things To Discuss
- `input-edit-mode` is becoming a kind of a meta-mode, being used in `prompt-buffer-mode` and `repl-mode` and likely to be used elsewhere. Should we do something about it? Maybe make it a mixin for any mode to inherit and not care about editing keybindings that are likely to stay the same?
- [x] `paren` feels wrong. Should we auto-insert the closing paren on the opening paren input? What are the restrictions there? Maybe remove it and leave `closing-paren` only? Should be complete too.
- [x] Maybe introduce a square bracket syntax (`]`) to auto-close all the open parens (for instance, `(some ((nested (expres (s (ion)))))` becomes `(some ((nested (expres (s (ion]`), as some ancient lisps did? Feels like a cool feature, and easy to do -- simply check all the closing parens and transform square ones into round ones before evaluation.
- [x] Bug: both `prompt-buffer-mode` and `repl-mode` have VI movement bindings, but not deletion bindings, while Emacs bindings have those in both modes.
- [x] Add a changelog entry?
  - But then, shoud we add all the keybinding changes, like `prompt-buffer-mode` ones, too?

# Testing

Simply run Nyxt, open `lisp-repl` and try editing. Feels much smoother than it used to be!


Ambrevar:
- `input-edit-mode`: Agreed on the mixin!
- `paren`: In my experience, I only press `)` when I want to _move_ the cursor after the closing paren.  Insertion is always done when pressing `(`. 
- The `]` is not standard Common Lisp, so I'm a bit unsure here.  We don't want to alienate our users :)
- The bug: Is there something to discuss?
- Changelog: the more, the better I suppose.  On the other hand, it's easy for the user to list the bindings dynamically.

Note: I use `lispy`, do you as well?  I find it to be one of the best implementations for Emacs.

aartaka:
We have tab-completion of symbols!

Ambrevar:
Woohoo!  I'll test later this week!

aartaka:
It's ready for review, I believe!

source/mode/repl.lisp
```diff
@@ -15,9 +15,29 @@
...
+       "M-d" 'nyxt/input-edit-mode:delete-forwards-word)
+      scheme:vi-normal
+      (list
+       ;; TODO: deleting chars/words
+       "l" 'nyxt/input-edit-mode:cursor-forwards
+       "h" 'nyxt/input-edit-mode:cursor-backwards
+       "w" 'nyxt/input-edit-mode:cursor-forwards-word
+       "b" 'nyxt/input-edit-mode:cursor-backwards-word
+       "x" 'nyxt/input-edit-mode:delete-forwards
+       ;; VI has no short keybinding for delete-backwards-word, hasn't it?
```
Ambrevar, line 39: Yes, `d b`.

source/mode/repl.lisp
```diff
@@ -99,6 +143,43 @@ INPUT is a string and RESULTS is a list of Lisp values.")))
...
+    (setf (input repl) (str:insert "()" cursor input)
+          (cursor repl) (1+ cursor))))
+
+(define-command closing-paren (&optional (repl (current-mode 'repl)))
+  "Automatically closes all the open parens before the cursor."
+  (let* ((input (input repl))
+         (parens-to-complete (- (count #\( input)
+                                (count #\) input))))
+    (setf (input repl) (str:concat input (make-string parens-to-complete :initial-element #\)))
+          (cursor repl) (+ parens-to-complete (length input)))))
```
Ambrevar, line 160: If the parens are balanced, then pressing `)` in Lispy brings me to the closing paren of the current s-exp.

Example
```
(foo (ba|r ) )
```
Pressing `)` brings the cursor `|` to
```
(foo (bar )| )
```

Ambrevar: Currently one can delete parentheses, leaving the input unbalanced.  Would be great to have balanced deleting commands, e.g. backspace on a closing parenthesis would delete the whole s-exp.

Of course this can also wait for another patch :)

source/mode/repl.lisp
```diff
@@ -99,6 +143,43 @@ INPUT is a string and RESULTS is a list of Lisp values.")))
...
+
+(define-command tab-complete-symbol (&optional (repl (current-mode 'repl)))
+  "Complete the current symbol and insert the completion into the REPL prompt."
+  (let* ((input (input repl))
+         (cursor (cursor repl))
+         (paren-or-space (find-if (lambda (c) (member c '(#\( #\) #\ ))) input
+                                  :start (1- cursor) :from-end t))
+         (paren-or-space (if paren-or-space (1+ paren-or-space) 0))
```
Ambrevar, line 168: `(1+ paren-or-space)` raises a condition.
Indeed, `paren-or-space` is a character, not a position.  Mistake?

Ambrevar:
Great stuff!

There seems to be a package issue in the input.  Running `lisp-implementation-version` returns 

```
"The function NYXT-USER::LIST-IMPLEMENTATION-VERSION is undefined."
```

in the REPL.

source/mode/repl.lisp
aartaka: Done!

aartaka:
`lisp-implementation-version` works just fine on my side 0_o

aartaka:
Other than that, it's all addressed!

Ambrevar:
> `lisp-implementation-version` works just fine on my side 0_o

Weird, works for me now as well.

source/mode/repl.lisp
```diff
@@ -99,6 +143,46 @@ INPUT is a string and RESULTS is a list of Lisp values.")))
...
+      (setf (input repl) (str:concat input (make-string parens-to-complete :initial-element #\)))))
+    (alexandria:when-let ((cursor (ignore-errors (1+ (position #\) (input repl) :start cursor)))))
+      (setf (cursor repl) cursor))))
+
+(define-command tab-complete-symbol (&optional (repl (current-mode 'repl)))
+  "Complete the current symbol and insert the completion into the REPL prompt."
+  (let* ((input (input repl))
+         (cursor (cursor repl))
+         (paren-or-space (position-if (lambda (c) (member c '(#\( #\) #\ ))) input
+                                      :start (1- cursor) :from-end t))
```
Ambrevar, line 170: Just like me, you mistook `:start` for the starting position of the search ;)

`:start` and `:end` designates the bounds of the input on which to operate, so here you want to replace `:start` with `:end`.  I know, it's confusing...

source/mode/repl.lisp
```diff
@@ -99,6 +143,46 @@ INPUT is a string and RESULTS is a list of Lisp values.")))
...
+    (when (plusp parens-to-complete)
+      (setf (input repl) (str:concat input (make-string parens-to-complete :initial-element #\)))))
+    (alexandria:when-let ((cursor (ignore-errors (1+ (position #\) (input repl) :start cursor)))))
+      (setf (cursor repl) cursor))))
+
+(define-command tab-complete-symbol (&optional (repl (current-mode 'repl)))
+  "Complete the current symbol and insert the completion into the REPL prompt."
+  (let* ((input (input repl))
+         (cursor (cursor repl))
+         (paren-or-space (position-if (lambda (c) (member c '(#\( #\) #\ ))) input
```
Ambrevar, line 169: `#\ ` can be hard to read and break some editors, prefer `#\space`.

Also the code is a bit hard to understand for me if you name this `paren-or-space`.  I would name it something like `previous-delimiter-position`.

PR #2099, opened 2022-02-12
Universal introspection

aartaka:
This adds the ability to introspect several objects at the same time, with an arbitrary nesting level, and with the pretty HTML-powered printing.

# What's New
- `value->html` is revamped to meaningfully print:
  - plists and alists as HTML dictionary lists (`<dl>` tag).
  - Lists and other sequences as HTML `<ul>` lists.
  - hash-tables, structure and class instances as dictionary lists.
  - Everything else as whatever `print` produces.
- `help-mode` has a table of all the inspected values, allowing inspection of up to infinity values :D 
  - As a matter of detail, it's a weak calss-allocated hash-table from strings to objects.
  - `inspected-values` (the name of the table) supercedes the `inspected-value` slot for good.
- `describe-value` now accepts an ID of the value to introspect, and it implies presence of `help-mode` to fetch the value from.
  - It's not much of a problem, as `describe-value` itself enables `help-mode` on all the pages it creates.
  - Passing arbitrary values into `describe-value` through `nyxt:` URLs would have been a problem, and this restriction on IDs fixes it.

# Things to Discuss
- Maybe use another notation (table?) for instances? It's too easy to confuse alists, plists, and objects.
  - Maybe that's for good, though -- they were intended for the same purpose of associative data storage.
- We need to somehow display type of the value/slot, so that similar notations don't trick one into seeing similar objects.
  - Tooltips?
  - Different notations for everything, or at least objects (see above)?
- Maybe restrict the depth of list printing? Right now nested lists are printed until they end, and that can produce really long structures.
  - It's much more convenient than what SLIME/SLY inspectors do with every list being a separate object that you need to introspect separately.

# Things To Do:
- Add object introspection links to:
   - [x] Lisp REPL.
   - [x] `describe-*` functions. 
   - [x] Manual?
- Test on different quirky values, like circular/dotted lists and CFFI structures.

# How To Test
Run `describe-variable` and choose `*browser*` or anything you like. Try clicking through slot values. Try different variables. Enjoy :D


aartaka:
Manual and `describe-*` functions already use `describe-value` wherever it makes sense, so no need to do anything here.

REPL is... more complicated. It requires `source/mode/repl.lisp` to be loaded after `source/help.lisp` which requires quite a restructuring of nyxt.asd. I'm not sure how to approach it. Maybe move introspected values into `browser`? Maybe move all the introspection code into configuration.lisp (`value->html` uses `help-mode`, though)?

Other than this dependency ordering and ambiguous markup (see Things To Discuss), it's review-ready :)

aartaka:
Should we print functions too? I mean, would be insane seeing a function object dissected by `value->html`, but do we want to break this abstraction?

aartaka:
Alright, REPL is ispection-enabled now due to all the inspection-related code being abstracted into inspector.lisp file loaded alongside, e.g., configuration.lisp.

The input on Things To Discuss will be much appreciated, still!

Ambrevar:
> Maybe restrict the depth of list printing? 

By the way, do you handle cycles?  Potential infinite loops around the corner ;)

Ambrevar:
> Should we print functions too? I mean, would be insane seeing a function object dissected by `value->html`,

How do you intent to dissect the function?  With `disassemble`?  If so, I think disassembly should be done using a special button because it's verbose and most certainly reserved to very specific cases.

Ambrevar:
> * Maybe use another notation (table?) for instances? It's too easy to confuse alists, plists, and objects.

Actually I suggest using tables for everything.  The header row/column could be used to differentiate plists from alists, etc.

The idea behind tables is that then we can inspect _multiple_ sequences / structures in one table!  You could have customizable sorting predicates on the keys (by default we could detect the type), add a filter input, etc.

A while ago I had the idea to implement this for SLY but I never finished the prototype.  However I found it very useful!  I believe it would make for the best (free) Common Lisp inspector out there! :)

Does that make sense?

aartaka:
> > * Maybe use another notation (table?) for instances? It's too easy to confuse alists, plists, and objects.
> 
> Actually I suggest using tables for everything. The header row/column could be used to differentiate plists from alists, etc.

This sounds like too much. Tables are good for structured data (arrays, alists, plists, hash-tables, classes), but arbitrary trees of conses are likely to break those. That's why I think it's better to leave at least non-associative lists as `<ul>` elements.

> The idea behind tables is that then we can inspect _multiple_ sequences / structures in one table! You could have customizable sorting predicates on the keys (by default we could detect the type), add a filter input, etc.

I don't exactly understand. What do you mean bu multiple sequences? Multiple return values? Multi-dimentional sequences?

> A while ago I had the idea to implement this for SLY but I never finished the prototype. However I found it very useful! I believe it would make for the best (free) Common Lisp inspector out there! :)

What restricts ourselves from being simply the best inspector, not just among free ones? ;)

aartaka:
Okay, I've changed the format to use tables for 
- hash-tables.
- plists.
- alists.
- 1- and 2D arrays.

I'm hesitant about objects. For those, dictionary lists feel a bit more pleasant visually, especially given that seeing lots of nested tables is scary, and this will be the case with objects (for instance, `*browser*` has at least three hash-tables in it!) How about we leave it as dictionary lists for classes and structures?

Test and see for yourself -- classes are much better looking as they are now! I don't want to repeat mistakes of ASP.NET forcing tables on everything xD

aartaka:
> By the way, do you handle cycles? Potential infinite loops around the corner ;)

Wow, I thought that circularity should be detected by `trivial-types:proper-list-p`, but it isn't! 06651ee91b4c52473336443947825882f8f231bf should fix this, relying on `alex:circular-list-p` and `alex:circular-tree-p`.

Ambrevar:
trivial-types are indeed... a bit too trivial!  `property-list` doesn't do what you think it does, etc.  So in general I would not rely on it :p

aartaka:
Functions are now inspectable, redirecting to `describe-functions`! Hopefully we can find a way to inspect lambdas too :)

Ambrevar:
> This sounds like too much. Tables are good for structured data (arrays, alists, plists, hash-tables, classes), but arbitrary trees of conses are likely to break those.

Misunderstanding: I meant to use tables for everything you listed, so arrays, etc.  Of course we won't use it for trees, graphs, etc. :)

> I don't exactly understand. What do you mean bu multiple sequences? Multiple return values? Multi-dimentional sequences?

No no, nothing like that, I mean when inspecting multiple objects at the same time, say a list of sequences.
There is no interface for it in SLY, but we could do it in Nyxt, for instance by "add object to inspector" while other objects are already being inspected, which would essentially add a row to the table.

Example:
- I inspect FILE1, it shows up in the inspector.
- I add the inspection of FILE2 to the  existing session, then I end up with this:

```org
| Symbol | name | directory      | type | path                   |
|--------+------+----------------+------+------------------------|
| FILE1  | foo  | /home/doe/bar/ | lisp | /home/doe/bar/foo.lisp |
| FILE2  | sbcl | /usr/bin       |      | /usr/bin/sbcl          |

```

Does that make sense?

Ambrevar:
> Hopefully we can find a way to inspect lambdas too

You can use `function-lambda-expression`, but it does not give results on all compilers.  With SBCL we are fine though.

aartaka:
> > Hopefully we can find a way to inspect lambdas too
> 
> You can use `function-lambda-expression`, but it does not give results on all compilers. With SBCL we are fine though.

It's part of the standard, so no harm in using it :)

Ambrevar:
Just tested, great job!

My main complaint for now is that the hash table "key row" is hard to read, because it's black on dark gray.  Could you change it?

source/inspector.lisp
```diff
@@ -0,0 +1,197 @@
...
+(export-always 'inspected-value)
+(defmethod inspected-value (id)
+  (gethash id *inspected-values*))
+
+(defmethod (setf inspected-value) (new-value id)
+  (setf (gethash id *inspected-values*) new-value))
+
+(defun escaped-literal-print (value)
+  (spinneret:with-html-string
+    (:code (:raw (spinneret::escape-string (format nil "~s" value))))))
```
Ambrevar, line 37: The `(format nil "~s")` is `write-to-string`, no?

source/inspector.lisp
```diff
@@ -0,0 +1,197 @@
...
+  (:method ((value null) &optional nested-p)
+    (declare (ignore nested-p))
+    (escaped-literal-print value))
+  (:method ((value string) &optional nested-p)
+    (declare (ignore nested-p))
+    (escaped-literal-print value))
+  (:documentation "Produce HTML showing the structure of the VALUE.
+If it's NESTED-P, compress the output.
+
+Redefine this method if you want to have a different markup for Lisp values in
```
Ambrevar, line 68: To be picky: "Specialize this generic function"

source/inspector.lisp
```diff
@@ -0,0 +1,197 @@
...
+    (declare (ignore nested-p))
+    (escaped-literal-print value))
+  (:method ((value null) &optional nested-p)
+    (declare (ignore nested-p))
+    (escaped-literal-print value))
+  (:method ((value string) &optional nested-p)
+    (declare (ignore nested-p))
+    (escaped-literal-print value))
+  (:documentation "Produce HTML showing the structure of the VALUE.
+If it's NESTED-P, compress the output.
```
Ambrevar, line 66: I don't really understand why you chose "nested-p" here, can you expand?

source/inspector.lisp
aartaka: Not necessarily. `write-to-string` is based on `write` and thus relies on control variables (`*print-base*`, `*print-case*`, `*print-circle*`, `*print-escape*`, `*print-gensym*`, `*print-length*` etc.), while `format` is a bit more granular (although some variables still influence it, like `*print-case*`).

source/inspector.lisp
aartaka: Done in a71ec4c651589488d565b3498a0b00842b5e4dc8!

source/inspector.lisp
aartaka: Basically, it used to mean that the printing is for a value inside some other printed value (e.g., slot value, list element, hash-table value), which might need some compaction.

But naming `compact-p` makes more sense in the current state. Done in 12d7a4516b338c6ff048bf427476a4ee5c78c1a1 :)

aartaka:
> My main complaint for now is that the hash table "key row" is hard to read, because it's black on dark gray. Could you change it?

Wow, it's not supposed to be like that. `theme:primary-color` should contrast both `theme:text-color` and `theme:background-color` (black and white respectively for the default theme).

aartaka:
@jmercouris, seems like we should fix some colors :)

source/inspector.lisp
aartaka: Or is it the other way around? :thinking:

aartaka: Got it, `format nil "~s"` is equivalent to `prin1-to-string`, as both are binding `*print-escape*` to t and leave everything else configurable!

jmercouris:
> @jmercouris, seems like we should fix some colors :)

Yes. I've seen the issue you've opened up about it. You are right, the contrast is all wrong!

aartaka:
I've changed the color of the hash keys to `theme:background-color` and adjusted the `theme` colors for a better contrast (although it doesn't feel perfect yet). Merging then?

Ambrevar:
OK for me!

PR #2103, opened 2022-02-14
Fix #2101

### LINKED ISSUE #2101: File upload prompt is broken.
When I try to upload a file I do not see any files in the prompter and get I lot of warnings in the terminal from where nyxt is started:

~~~~
Attributes of #P"/home/vasily/.wolf/" should be a non-dotted alist instead of (("Name"
   Attributes of #P"/home/vasily/.wolf/" should be a non-dotted alist instead of (("Name"
                                                                                          NIL)
                                                                                   NIL)
                                                                                   NIL)
"
                                                                                          NIL)
WARNING:
                                                                                         ("Ext
                                                                                         ("Extension"/

                                                                                        ("Extension"/                es/" shoul  be a non-dotted alist instead of (("Name"
                                                                                        ("Extension"/                es/" shoul  be a non-dotted alist instead of (("Name" 
~~~~

The bug is introduced in 4f639c50b489820235a0ace438c10aaf84a58fa5 and can be fixed by using the old method for `promter:object-attributes`:

~~~~
(defmethod prompter:object-attributes ((path pathname))
  `(("Path" ,(uiop:native-namestring path))))

#+nil
(defmethod prompter:object-attributes ((path pathname))
  `(("Name" ,(pathname-name path))
    ("Extension" ,(or (nfiles:pathname-type* path) ""))
    ("Directory" ,(uiop:native-namestring (nfiles:parent path)))))
~~~~

I can send a patch unless you have better solution.

shamazmazum:
1) Print top-level directories in the "Name" column of prompter buffer.
   This allows fixes autocompletion for files residing not in the home
   directory.
2) Fix PROMPTER:ATTRIBUTES-DEFAULT for file suggestions. We need to
   return the whole namestring intact, not just "Name" part of
   suggestion's attributes.

### DIFF
--- a/source/mode/file-manager.lisp
+++ b/source/mode/file-manager.lisp
@@ -107,7 +107,10 @@ When the user is unspecified, take the current one."
 (define-user-class program-source)
 
 (defmethod prompter:object-attributes ((path pathname))
-  `(("Name" ,(or (pathname-name path) ""))
+  `(("Path" ,(uiop:native-namestring path))
+    ("Name" ,(if (uiop:directory-pathname-p path)
+                 (enough-namestring path (nfiles:parent path))
+                 (pathname-name path)))
     ("Extension" ,(or (nfiles:pathname-type* path) ""))
     ("Directory" ,(uiop:native-namestring (nfiles:parent path)))))
 
@@ -141,6 +144,9 @@ It's suitable for `prompter:filter-preprocessor'."
 
 (define-class file-source (prompter:source)
   ((prompter:name "Files")
+   (prompter:active-attributes-keys
+    '("Name" "Extension" "Directory")
+    :accessor nil)
    (prompter:filter-preprocessor (make-file-source-preprocessor))
    (prompter:multi-selection-p t)
    (open-file-in-new-buffer-p t :documentation "If nil, don't open files and directories in a new buffer.")


source/mode/file-manager.lisp
```diff
@@ -107,7 +107,9 @@ When the user is unspecified, take the current one."
 (define-user-class program-source)
 
 (defmethod prompter:object-attributes ((path pathname))
-  `(("Name" ,(or (pathname-name path) ""))
+  `(("Name" ,(if (uiop:directory-pathname-p path)
+                 (enough-namestring path (nfiles:parent path))
+                 (pathname-name path)))
```
Ambrevar, line 112: Isn't it the same as the simpler `(alex:last-elt (pathname-directory (uiop:ensure-directory-pathname path)))`

source/mode/file-manager.lisp
```diff
@@ -139,10 +141,29 @@ It's suitable for `prompter:filter-preprocessor'."
...
+(defun make-file-suggestion (value &optional source input)
+  (declare (ignore source input))
+    (make-instance 'file-suggestion
+                   :value value
+                   :attributes (prompter:object-attributes value)))
+
+(defmethod prompter:attributes-default ((file-suggestion file-suggestion))
+  "Return file pathname as is, without any modifications we made for
+drawing nice table in the prompter buffer."
+  (namestring (prompter:value file-suggestion)))
```
Ambrevar, line 160: So this returns the full path, right?

Ambrevar:
For 2., instead you of creating a custom suggestion maker (which is not the intended use I believe), you could create a specialized object-attributes that only prints the `namestring`, right?

Also what about this other approach: add another object attribute with the `namestring`, and just print this one attribute by default.  What do you think?

source/mode/file-manager.lisp
shamazmazum: Right

source/mode/file-manager.lisp
shamazmazum: Yes, I think so.

shamazmazum:
> you could create a specialized object-attributes that only prints the namestring, right?

Do you mean a method for `prompter:object-attributes`? We had this before 4f639c5:

~~~~
(defmethod prompter:object-attributes ((path pathname))
   ;; TODO: Add dirname, basename, extension.
   ;; It will be useful when we have per-attribute filtering.
   `(("Path" ,(uiop:native-namestring path))))
~~~~

If you press tab in prompter buffer, `attributes-default` will be used to set the input:

~~~~
(define-command-prompt insert-selection (prompt-buffer)
	"Insert current selection default property in the prompt buffer input."
  (alex:when-let ((selection (prompter:attributes-default
	                      (prompter:selected-suggestion prompt-buffer))))
    (nyxt::set-prompt-buffer-input selection)))
~~~~

`attributes-default` return value of the first attribute:

~~~~
(export-always 'attributes-default)
(defmethod attributes-default ((suggestion suggestion))
  "Return SUGGESTION default attribute value."
  (second (first (attributes suggestion))))
~~~~

So I have two options, as I think. 

1) Return `Path` as the first attribute. But why we need `Name`, `Extension` and all others if we have just `Path`?
2) Write a custom `suggestion` class.

I did the second.

aartaka:
There's a third option: dump all the attributes (`Name`, `Extension`, `Directory`, and `Path`) into the same method, and filter those with `prompter:active-attributes-keys` slot of the `user-file-source`. Or am I missing something?

aartaka:
I guess I've just rephrased what @Ambrevar said :D

shamazmazum:
I just missed `active-attributes-keys`. But I cannot use `prompter:active-attributes-keys` and `(:accessor-name-transformer (class*:make-name-transformer name))` in the same `define-class`:

~~~~
attempt to add the method
  #<SB-MOP:STANDARD-READER-METHOD COMMON-LISP:NIL, slot:PROMPTER:ACTIVE-ATTRIBUTES-KEYS, (#<STANDARD-CLASS NYXT/FILE-MANAGER-MODE::FILE-SOURCE {102A23BAA3}>) {102A2401E3}>
to the generic function
  #<STANDARD-GENERIC-FUNCTION PROMPTER:ACTIVE-ATTRIBUTES-KEYS (1)>;
 but the method and generic function differ in whether they accept
&REST or &KEY arguments.
   [Condition of type SB-INT:SIMPLE-PROGRAM-ERROR]
~~~~

Is it intentional? If it is so, I don't see how I can use `prompter:active-attributes-keys` in `file-source`. If it's not 

~~~~
(defmethod active-attributes-keys ((source source) &key &allow-other-keys)
  "Return active attributes keys.
If the `active-attributes' slot is NIL, return all attributes keys."
  (or (slot-value source 'active-attributes-keys)
      (attributes-keys source)))
~~~~

should be replaced with something like this:

~~~~
(defmethod active-attributes-keys :around ((source source))
  "Return active attributes keys.
If the `active-attributes' slot is NIL, return all attributes keys."
  (or (call-next-method) (attributes-keys source)))
~~~~

shamazmazum:
Oh my, that super macro `define-class` has `:accessor nil` option for a slot. Thank god.

Never used it. `defclass` suffices my needs. I hope this can be merged

source/mode/file-manager.lisp
shamazmazum: Second thought.

No, this is not the same. My version makes difference between a directory `foo/` and a file `foo`. Yours represents them simply as `foo`.

Ambrevar:
Looks good to me!  @aartaka ?

aartaka:
Yes, looks perfect!

Ambrevar:
@shamazmazum Could you just reword the commit message so that the first line is in the following format:

```
TARGET-FILE: What and why the change is made.
```

shamazmazum:
Done

jmercouris:
Thank you :-)

Issue #2043, opened 2022-01-21
Create a Nyxt theme with a style inspired by the visual identity of the Nyxt Academy slides


pdelfino:
It would be interesting to use it in some videos. And it would be fun :)

pdelfino:
The designer who did the visual identity suggested these material for us. 
![3](https://user-images.githubusercontent.com/7698207/150614405-b106b8c0-3da5-427e-8e12-c971f8a6fbbf.jpeg)
![2](https://user-images.githubusercontent.com/7698207/150614407-a257b4cb-518d-4a82-a5b8-6eba665aab0e.jpeg)
![WhatsApp Image 2022-01-21 at 20 37 44](https://user-images.githubusercontent.com/7698207/150614409-2a604ce3-f0b7-4484-98f4-c95931c672eb.jpeg)

She uses [this website ](https://coolors.co/a45c30-ffffff-000000-247ba0-3e78b2) to generate combinations.

aartaka:
I like the second combination -- this shade of yellow would look insanely cook in element hints!

aartaka:
I've played a bit this Coolors, and here's a combinations that could work:

![nyxt-academy](https://user-images.githubusercontent.com/57838654/150635081-43afb4b3-2020-4e45-ba7e-afb18203f2f8.png)

Or, as a configuration snippet that you could try out in your config file:

```
(define-configuration browser
  ((theme (make-instance
           'theme:theme
           :dark-p t
           :background-color "black"
           :text-color "white"
           :accent-color "#FCBA04"
           :primary-color "#A45C30"
           :secondary-color "#DB9665"
           :tertiary-color "#AD693E"
           :quaternary-color "#7D3509"))))
```

It has some Greek amphora vibe to it, hasn't it?

![E5965B75](https://user-images.githubusercontent.com/57838654/150634995-fd9d24ed-2f76-4b3a-a188-c86115a6b84b.png)

aartaka:
What I don't like about the theme that I suggested is:
- The yellow is a bit too bright, maybe use some lighter shade?
- Primary and tertiary colors are too close. This could cause confusion.

aartaka:
@pdelfino, @jmercouris, any feedback?

pdelfino:
> @pdelfino, @jmercouris, any feedback?

I think it is great! Good job, Artyom.

I showed this to Nivea (the designer) and she liked it :slightly_smiling_face:

 For me, it is ready to go :)

[...]

Issue #2115, opened 2022-02-17
Bring more contrast into the default color scheme


aartaka:
# Analysis

Right now, based on the results of [color contrast checking](https://webaim.org/resources/contrastchecker/) our default theme, one can find _a lot_ of not-contrasting-enough colors. This poses comprehensibility problems [that we encounter already](https://github.com/atlas-engineer/nyxt/pull/2099#issuecomment-1041457822). Constrast values:
```
text+primary = AAA
background+primary = fail
text+secondary = AA
background+tertiary = AA, but only for large text and graphics
text+quaternary = AAA

accent+text = fail
accent+background = AAA
accent+primary = AA, but only for large text and graphics
accent+secondary = fail
accent+tertiary = fail
accent+quaternary = fail
```

# Plan
The contrast of the text/background with primary/secondary/tertiary/quaternary colors we should fix, as 
- We commit to it in the `theme` documentation.
- It's much easier on the eyes this way.

Accent color's I'm not that certain about fixing. It's implied to stand out hue-wise, so contrast issues should be mostly ruled our given its brightness and hue. But there should be some other precautions for e.g., color-blind people and monochrome screens. What Trello (for instance) does in this case is [using colors+geometric patterns](https://wearecolorblind.com/examples/trello-colorblind-friendly-mode/).

# Suggestions:
Make the primary color lighter. I suggest using `rgb(100,100,100)` as it has a good contrast with backgroud color and a bearable contrast with the text color. We mostly use primary with background color (think buttons and structural elements, so should be no harm leaning more to constrasting background first.

Make tertiary color a bit darker. A value of `rgb(115, 115, 115)` works well here. However, it's too close to secondary color. How about moving secondary color up a bit to something like `rgb(150, 150, 150)`? This swaps the colors, so maybe we should just change the value/documentation of `theme` slots in this case?

Am I over-engineering here?

EDIT: Add link to Trello example, clarify.

Ambrevar:
Sounds good to me!

aartaka:
#2099 should've fixed it.

PR #2116, opened 2022-02-18
Introducing dark and light themes. Also, make it possible to change them on common settings

pdelfino:
This is a draft for issue #2043.

@aartaka suggested:

> @pdelfino, would you be interested is implementing theme settings, as we discussed via email (+dark-theme+ and +light-theme+ variables, Common Settings page with color samples and ability to change the theme). If not, I can do it sometime this week :)

I am not sure about the "color samples" on the Common Settings. Artyom, could you give me a hint on how to implement that?

Also, apart from the "color sample", is the rest of the PR fully implementing your suggestion?

Thanks!

### DIFF
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -431,53 +431,81 @@ CLASS is a class symbol."
   "Configure a set of frequently used settings."
   (let ((spinneret:*html-style* :tree))
     (spinneret:with-html-string
-      (:h1 "Common Settings")
-      (:p "Set the values for frequently configured settings. "
-          "Changes only apply to newly created buffers.")
-      (:h2 "Keybinding style")
-      (:p (:button :class "button"
-                   :onclick (ps:ps (nyxt/ps:lisp-eval
-                                    `(progn (nyxt::configure-slot
-                                             'default-modes 'buffer
-                                             :value '%slot-default%)
-                                            (nyxt/emacs-mode:emacs-mode :activate nil)
-                                            (nyxt/vi-mode:vi-normal-mode :activate nil))))
-                   "Use default (CUA)"))
-      (:p (:button :class "button"
-                   :onclick (ps:ps (nyxt/ps:lisp-eval
-                                    `(progn (nyxt::configure-slot
-                                             'default-modes 'buffer
-                                             :value '(append '(emacs-mode) %slot-default%))
-                                            (nyxt/emacs-mode:emacs-mode :activate t)
-                                            (nyxt/vi-mode:vi-normal-mode :activate nil))))
-                   "Use Emacs"))
-      (:p (:button :class "button"
-                   :onclick (ps:ps (nyxt/ps:lisp-eval
-                                    `(progn (nyxt::configure-slot
-                                             'default-modes 'buffer
-                                             :value '(append '(vi-normal-mode) %slot-default%))
-                                            (nyxt/emacs-mode:emacs-mode :activate nil)
-                                            (nyxt/vi-mode:vi-normal-mode :activate t))))
-                   "Use vi"))
-      (:h2 "Default new buffer URL")
-      (:button :class "button"
-               :onclick (ps:ps (nyxt/ps:lisp-eval
-                                `(nyxt::configure-slot 'default-new-buffer-url 'browser :type 'STRING)))
-               "Set default new buffer URL")
-      (:h2 "Default zoom ratio")
-      (:button :class "button"
-               :onclick (ps:ps (nyxt/ps:lisp-eval
-                                `(nyxt::configure-slot 'current-zoom-ratio 'buffer)))
-               "Set default zoom ratio")
-      (:h2 "Disable compositing")
-      (:p "On some systems, compositing can cause issues with rendering. If you
+      (theme:with-theme (nyxt:theme nyxt:*browser*)
+        (:h1 "Common Settings")
+        (:p "Set the values for frequently configured settings. "
+            "Changes only apply to newly created buffers.")
+        (:h2 "Keybinding style")
+        (:p (:button :class "button"
+                     :onclick (ps:ps (nyxt/ps:lisp-eval
+                                      `(progn (nyxt::configure-slot
+                                               'default-modes 'buffer
+                                               :value '%slot-default%)
+                                              (nyxt/emacs-mode:emacs-mode :activate nil)
+                                              (nyxt/vi-mode:vi-normal-mode :activate nil))))
+                     "Use default (CUA)"))
+        (:p (:button :class "button"
+                     :onclick (ps:ps (nyxt/ps:lisp-eval
+                                      `(progn (nyxt::configure-slot
+                                               'default-modes 'buffer
+                                               :value '(append '(emacs-mode) %slot-default%))
+                                              (nyxt/emacs-mode:emacs-mode :activate t)
+                                              (nyxt/vi-mode:vi-normal-mode :activate nil))))
+                     "Use Emacs"))
+        (:p (:button :class "button"
+                     :onclick (ps:ps (nyxt/ps:lisp-eval
+                                      `(progn (nyxt::configure-slot
+                                               'default-modes 'buffer
+                                               :value '(append '(vi-normal-mode) %slot-default%))
+                                              (nyxt/emacs-mode:emacs-mode :activate nil)
+                                              (nyxt/vi-mode:vi-normal-mode :activate t))))
+                     "Use vi"))
+        (flet ((generate-colors (theme-symbol text)
+                 (spinneret:with-html-string
+                     (:p (:button :class "button"
+                                  :style (format nil "background-color: ~a; color: ~a"
+                                                 (theme:primary-color (symbol-value theme-symbol))
+                                                 (theme:background-color (symbol-value theme-symbol)))
+                                  :onclick (ps:ps (nyxt/ps:lisp-eval
+                                                   `(configure-slot 'theme 'browser
+                                                                    :value ',theme-symbol)))
+                                  text))
+                   (:p "Colors:")
+                   (:dl
+                    (loop for (name color text-color) in '(("Text" theme:text-color theme:background-color)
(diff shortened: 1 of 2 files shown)


pdelfino:
In addition, I think you *did not* suggest this, Artyom... But since `(configure-slot` works permanents (since it changes `auto-config.lisp`) and requires re-start, I thought it would be cool to add some commands that change theme instantaneously and just for the current session.

0 - not sure if this is a good idea at all. Curious about your thoughts.
1 - not sure if the `user-interface.lisp` is the best place to put the commands
2 - it seems to work, except for the lower part of the U.I. (status area, icons, message area...). I do not fully understand the side effect behavior here. I will try to study it tomorrow with a fresh mind.

source/user-interface.lisp
```diff
@@ -23,3 +23,13 @@
    (ps:ps (let ((element (ps:chain document (get-element-by-id (ps:lisp (user-interface:id button))))))
             (setf (ps:chain element text-content) (ps:lisp (user-interface:text button)))
             (setf (ps:chain element onclick) (ps:lisp (user-interface:action button)))))))
+
+(define-command set-dark-theme ()
+  "Set the THEME of BROWSER to be dark for this session. This does not change
+future sessions of Nyxt. To do that, see the Common Settings"
+  (Setf (Theme *Browser*) Theme::+Dark-Theme+))
```
aartaka, line 30: Why this capitalization? 0_o

source/user-interface.lisp
```diff
@@ -23,3 +23,13 @@
    (ps:ps (let ((element (ps:chain document (get-element-by-id (ps:lisp (user-interface:id button))))))
             (setf (ps:chain element text-content) (ps:lisp (user-interface:text button)))
             (setf (ps:chain element onclick) (ps:lisp (user-interface:action button)))))))
+
+(define-command set-dark-theme ()
+  "Set the THEME of BROWSER to be dark for this session. This does not change
```
aartaka, line 28: Not sure you need to capitalise `browser` here -- it's not one of the arguments. Same for `theme`. You can use the backtick-quote notation, though :)

aartaka, line 28: Make sure the first line of the docstring is strictly one sentence.

source/user-interface.lisp
```diff
@@ -23,3 +23,13 @@
...
+future sessions of Nyxt. To do that, see the Common Settings"
+  (Setf (Theme *Browser*) Theme::+Dark-Theme+))
+
+(define-command set-light-theme ()
+  "Set the THEME of BROWSER to be light for this session. This does not change
```
aartaka, line 33: Same. `theme` and `browser` are not to be capitalized and there should only be one sentence on the first docstring line.

aartaka (commented):
All good, except for some stylistic moments! It's exactly as I imagined it looking!

**Regarding color snippets:** I was thinking of something along the lines of 
```
(spinneret:with-html-string
  (theme:with-theme (nyxt:theme nyxt:*browser*)
    (:span :style (format nil "background-color: ~a" theme:primary)
           theme:primary)))
```

This is just one color, so you'd need to repeat this generation for all of those, make sure no color is blending with the background (maybe a contrasting border?), and ensure the text is contrasting enough.

**Regarding commands**: I'm not sure we need those. What's the use-case? If one wants to enable dark theme for just one website/page, they'll use `dark-mode`. If it's the other way around -- they'll use `invert-color`. Maybe these commands don't belong to the patch. Mind opening an issue about this after we merge?

**Regarding `setf` not working as intended**: That's how it's expected to behave, as buffer/status-buffer/message-buffer style is set at the initialization time. If we change the theme at runtime, those are not updated automatically. To avoid that, you need to set the `style` of every buffer to the newly-generated one, which is not exactly elegant. Not sure it belongs to this patch.

aartaka:
Also, I've forgotten to check the color contrast (see #2115) of this theme when we were thinking of it. Here's the results I've got, they are not perfect:

```
text+primary = AA
background+primary = AA for graphics and large text only
text+secondary = fail
background+tertiary = AA
text+quaternary = AAA

accent+text = fail
accent+background = AA
accent+primary = fail
accent+secondary = fail
accent+tertiary = fail
accent+quaternary = AA
```

Accent color is fine, but `background+primary` (probably the most frequent interface color combination we have) and `text+secondary` we should fix, I guess. @pdelfino, are you interested in experimenting with the theme to make it more contrasting, or should I do that?

pdelfino:
> Accent color is fine, but `background+primary` (probably the most frequent interface color combination we have) and `text+secondary` we should fix, I guess. @pdelfino, are you interested in experimenting with the theme to make it more contrasting, or should I do that?

I will try that on Monday! Thanks for all the feedback :)

aartaka:
The simple fix would be to swap the `primary` and the `tertiary` colors -- current `tertiary` color has a nice contrast with both `text` and `background` colors.

source/user-interface.lisp
pdelfino: An accident with the command `upcase-region` :blush: . I will fix it!

pdelfino:
> Regarding commands

Commands already removed. 

> The simple fix would be to swap the `primary` and the `tertiary` colors -- current `tertiary` color has a nice contrast with both `text` and `background` colors.

Contrast is fixed! :+1: 

> Regarding color snippets

I tried an implementation that generates these results:

Screenshot from 2022-02-24 19-21-39
Screenshot from 2022-02-24 19-21-56

Since the results look kind of ugly,  I am afraid that I might have misunderstood something. Not sure if this is what you had in mind. Did I miss something?

aartaka:
What I though about is rather something along the lines of
``` diff
modified   source/help.lisp
@@ -460,19 +460,33 @@ CLASS is a class symbol."
                                               (nyxt/emacs-mode:emacs-mode :activate nil)
                                               (nyxt/vi-mode:vi-normal-mode :activate t))))
                      "Use vi"))
-        (:h2 "Theme style")
-        (:p (:button :class "button"
-                     :onclick (ps:ps (nyxt/ps:lisp-eval
-                                      `(configure-slot 'theme 'browser
-                                                       :value 'theme::+light-theme+)))
-                     (:span :style (format nil "background-color: ~a" (theme:primary-color theme::+light-theme+))
-                            "Use default (Light theme)")))
-        (:p (:button :class "button"
-                     :onclick (ps:ps (nyxt/ps:lisp-eval
-                                      `(configure-slot 'theme 'browser
-                                                       :value 'theme::+dark-theme+)))
-                     (:span :style (format nil "background-color: ~a" (theme:primary-color theme::+dark-theme+))
-                            "Use Dark theme")))
+        (flet ((generate-colors (theme-symbol text)
+                 (spinneret:with-html-string
+                   (:p (:button :class "button"
+                                :style (format nil "background-color: ~a; color: ~a"
+                                               (theme:primary-color (symbol-value theme-symbol))
+                                               (theme:background-color (symbol-value theme-symbol)))
+                                :onclick (ps:ps (nyxt/ps:lisp-eval
+                                                 `(configure-slot 'theme 'browser
+                                                                  :value ',theme-symbol)))
+                                text))
+                   (:p "Colors:")
+                   (:dl
+                    (loop for (name color text-color) in '(("Text" theme:text-color theme:background-color)
+                                                           ("Accent" theme:accent-color theme:background-color)
+                                                           ("Primary" theme:primary-color theme:background-color)
+                                                           ("Secondary" theme:secondary-color theme:background-color)
+                                                           ("Tertiary" theme:tertiary-color theme:text-color)
+                                                           ("Quaternary" theme:quaternary-color theme:text-color)
+                                                           ("Background" theme:background-color theme:text-color))
+                          collect (:dt name ": ")
+                          collect (:dd (:span :style (format nil "background-color: ~a; color: ~a; border-radius: 0.2em"
+                                                             (slot-value (symbol-value theme-symbol) color)
+                                                             (slot-value (symbol-value theme-symbol) text-color))
+                                              (slot-value (symbol-value theme-symbol) color))))))))
+          (:h2 "Theme style")
+          (:raw (generate-colors 'theme::+light-theme+ "Use default (Light theme)"))
+          (:raw (generate-colors 'theme::+dark-theme+ "Use Dark theme")))
         (:h2 "Default new buffer URL")
         (:button :class "button"
                  :onclick (ps:ps (nyxt/ps:lisp-eval
```

This way, all the colors of the theme are visible and one is aware of what they choose, not only by primary color, but by the full list :)

Alternatively, define a `value->html` method for `theme:theme` and list colors beautifully here! This way, people inspecting themes can see the colors too :D

aartaka:
And remember: themes provide their own sets of colors, sou you better set both `background-color` and `color` with the ones theme provides. Otherwise colors might get diluted by the current theme :)

pdelfino:
Thanks, @aartaka! Your suggestion was totally out of sight for me. I think this is good to go now. What do you think?

source/help.lisp
```diff
@@ -431,53 +431,81 @@ CLASS is a class symbol."
...
-                                `(nyxt::configure-slot 'default-new-buffer-url 'browser :type 'STRING)))
-               "Set default new buffer URL")
-      (:h2 "Default zoom ratio")
-      (:button :class "button"
-               :onclick (ps:ps (nyxt/ps:lisp-eval
-                                `(nyxt::configure-slot 'current-zoom-ratio 'buffer)))
-               "Set default zoom ratio")
-      (:h2 "Disable compositing")
-      (:p "On some systems, compositing can cause issues with rendering. If you
+      (theme:with-theme (nyxt:theme nyxt:*browser*)
```
aartaka, line 434: Do you use it anywhere, actually? `theme:with-theme` binds `theme:text`, `theme:accent`  etc. so you should only use it if you have to generate CSS with colors injected.

aartaka:
Just tested it and it's a wonderful theme! Thanks for taking the time to implement it, Pedro! Merging right now :D

pdelfino:
> Just tested it and it's a wonderful theme! Thanks for taking the time to implement it, Pedro! Merging right now :D

Thank you, Artyom. It was mainly your contribution :)
I learned a bit about themes due to you. Thanks!

PR #2121, opened 2022-02-23
Remove dummy buffer after setting first buffer of window.

### LINKED ISSUE #2119: Extraneous, dangling empty buffer on startup
Recipe:

- start
- don't restore the history (press `esc`)

You land on the help page, but actually there is a second buffer without name.  It shouldn't be there.

Ambrevar:
Should fix #2119.

I believe the dangling dummy buffer issue was introduced around c247e8282f62b78b4c058a62059b31428521800f.

@aartaka What do you think?


aartaka:
Looks good!

jmercouris:
Much nicer, thank you.

PR #2123, opened 2022-02-24
mode/tts: Cleanup docs, get mode working again.

hendursaga:
Note https://github.com/Shinmera/CLSS/issues/14 is still a limitation/bug, it might need to be replaced with some ParenScript.

### DIFF
--- a/source/mode/tts.lisp
+++ b/source/mode/tts.lisp
@@ -2,19 +2,19 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (uiop:define-package :nyxt/tts-mode
-    (:use :common-lisp :nyxt)
-  (:documentation "Mode for text-to-speak a documents content."))
+  (:use :common-lisp :nyxt)
+  (:documentation "Mode for interacting with Text-to-Speech (TTS) software."))
 
 (in-package :nyxt/tts-mode)
 
 (define-mode tts-mode ()
-  "A mode for text-to-speak a documents content.
-By default, no executable is configured, so for the mode to work,
-you need to configure an executable than can take a string of text.
-Something like `espeak` should work out of the box.
-Also by default, the content that will be send to the executable
-is the text-content of p-tags on the page. This can be configured
-by changing the `selector`-slot.
+  "A mode for interacting with Text-to-Speech (TTS) software.
+
+By default, no executable is configured, so for the mode to work, you need to
+configure an executable that can take a string of text. Something like `espeak`
+or `espeak-ng` should work out of the box. The content that will be sent to the
+executable is the text content of paragraph tags on the page. This can be
+configured by changing the `selector` slot.
 
 Example:
 
@@ -25,13 +25,10 @@ Example:
   ((executable nil
                :type (or string null)
                :documentation "The executable command to run.")
-   ;; TODO: If you specify multiple tags, like h1, h2, h3, p, they are selected in that
-   ;; order, not in the order they appear on the page. This is different than how it's
-   ;; done with e.g. JS document.querySelectorAll which returns them in the order they
-   ;; appear, which is IMO the correct way ... can you configure clss to do this?
+   ;; TODO: work around limitation https://github.com/Shinmera/CLSS/issues/14
    (selector "p"
-             :type string
-             :documentation "css selector that describes what elements text to speak.")
+       :type string
+       :documentation "CSS selector that describes which elements' text to speak.")
    (executable-process-info nil
                             :type (or uiop/launch-program::process-info null)
                             :documentation "Holds the process-info object of the running process")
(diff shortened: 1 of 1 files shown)


Ambrevar:
Thanks for the cleanup!
This mode needs more love indeed ;)

To clarify, what exactly did you fix?

aartaka:
No need to use Parenscript, you can `sort` the `clss`-returned results by their Nyxt ID! See https://github.com/atlas-engineer/nyxt/blob/61ef14ea25679e47f0b2d343a663e8970bba49d5/source/mode/jump-heading.lisp#L34-L43

hendursaga:
> To clarify, what exactly did you fix?

@Ambrevar Reading over my one commit again, it appears I just linted/re-formatted and re-wrote the documentation. I think there was some issue with my config that I fixed on that end that made it work again, actually. You can re-name the commit/PR if you want.

> No need to use Parenscript

@aartaka Great tip! should I try and submit another patch to cover that, or you, or something else?

aartaka:
> @aartaka Great tip! should I try and submit another patch to cover that, or you, or something else?

You're welcome to, if you're still interested :)

hendursaga:
Anyone have any input on what the default CSS selector should be? I was thinking at least include heading tags.

aartaka:
> Anyone have any input on what the default CSS selector should be? I was thinking at least include heading tags.

I'm not sure. Ideally, we should have some smart text cleanup, like what [Readability.js](https://github.com/mozilla/readability) does.

hendursaga:
I believe this can be merged now, unfortunately `if-process-then-terminate` doesn't work at all on my Linux system, I suspect it has something to do with `uiop:terminate-process`.

source/mode/tts.lisp
```diff
@@ -2,36 +2,33 @@
...
-   ;; TODO: If you specify multiple tags, like h1, h2, h3, p, they are selected in that
-   ;; order, not in the order they appear on the page. This is different than how it's
-   ;; done with e.g. JS document.querySelectorAll which returns them in the order they
-   ;; appear, which is IMO the correct way ... can you configure clss to do this?
-   (selector "p"
-             :type string
-             :documentation "css selector that describes what elements text to speak.")
+   (selector "p, h1, h2, h3, h4, h5, h6"
+       :type string
+       :documentation "CSS selector that describes which elements' text to speak.")
```
Ambrevar, line 31: Indentation is off.

source/mode/tts.lisp
```diff
@@ -40,57 +37,55 @@ Example:
       (if-process-then-terminate mode)))))
 
 (defmethod process-document ((mode tts-mode))
-  "Fetch the text in buffer that matches `selector` and send it off
-to get *spoken*."
+  "Fetch the text in buffer that matches `selector` and send it off to the TTS."
```
Ambrevar, line 40: Should be 
```
`selector'
```
if you mean the mode slot.

Ambrevar:
About the `uiop:terminate-process` issue: seems to run well on my machine.
Can you try the following:

```lisp
(let ((p-info (uiop:launch-program "sleep 1 && echo should-not-display" :output t))) (uiop:terminate-process p-info) (uiop:process-alive-p p-info))
```

It should return NIL and not output anything.

source/mode/tts.lisp
hendursaga: Yeah, I noticed that too, my Emacs for whatever crazy reason wouldn't format it properly. There a way to format it manually and tell Emacs to not override it? This is after even connecting Sly to Nyxt.

aartaka: There are sly-specific indentation variables, but in this case it's easier to just move the initform (`"p, h1..."`) to a separate line :)

hendursaga:
> It should return NIL and not output anything.

I'm afraid I get an error, reproduced below:

```
Incorrect I/O specifier T for :OUTPUT
   [Condition of type UIOP/UTILITY:PARAMETER-ERROR]
```

Do I need to provide more info on my system, or what?

aartaka:
Try without `:output t`, should behave the same yet without erroring :)

hendursaga:
> Try without `:output t`, should behave the same yet without erroring :)

@aartaka Already did that, except it doesn't return NIL, just T. There is no output, though.

aartaka:
> > Try without `:output t`, should behave the same yet without erroring :)
> 
> @aartaka Already did that, except it doesn't return NIL, just T. There is no output, though.

I mean, that's exactly what @Ambrevar described — it shouldn't output anything when terminating the process before `echo should-not-print`.

Or am I missing something?

Ambrevar:
> > It should return NIL and not output anything.
> 
> I'm afraid I get an error, reproduced below:
> 
> ```
> Incorrect I/O specifier T for :OUTPUT
>    [Condition of type UIOP/UTILITY:PARAMETER-ERROR]
> ```
> 
> Do I need to provide more info on my system, or what?

While this message is indeed unrelated to the ability to terminate a process, it should not happen.  It was also reported here:
https://github.com/atlas-engineer/nyxt/issues/2102#issuecomment-1047924054

Which REPL are you running?
Which SBCL version, and which asdf version?  (`(asdf:asdf-version)`)

hendursaga:
> Which REPL are you running?

Emacs' Sly, both recent versions. Should I be using Nyxt's REPL?

> Which SBCL version, and which asdf version?

2.2.1 and 3.3.1, respectively.

Ambrevar:
Your ASDF is outdated, try with ASDF 3.3.5.

    git clone -b release https://gitlab.common-lisp.net/asdf/asdf.git ~/common-lisp/asdf

hendursaga:
> Your ASDF is outdated, try with ASDF 3.3.5.

I cloned it to `~/.local/share/common-lisp/source/` which I prefer. `(asdf:asdf-version)` shows 3.3.5 inside SBCL, but even after re-compiling Nyxt, from the Nyxt REPL/remote Sly it's still showing the old version. This is after trying `(require "asdf")`, `(ql:quickload :asdf)` etc. How might I "force reload" or load it properly in the first place?

Ambrevar:
How are you compiling Nyxt? Are you using Guix?

By default, Nyxt disables all the default ASDF registries (this is to
prevent unexpected conflicts).

So if you want to build using some local libraries, you need to
re-enable them with `reset-asdf-registries`.

To list your current asdf registries, inspect the value of asdf:*default-source-registries*.

hendursaga:
@Ambrevar No, I'm compiling manually. I haven't touched Guix in awhile.

`reset-asdf-registries` works, but this still doesn't show how I can reload ASDF, because it seems by the time the very first line of configuration is loaded, ASDF is already loaded. The first few lines of my config are now

```common-lisp
(reset-asdf-registries)

(require "asdf")

#-quicklisp
(let ((quicklisp-init (merge-pathnames "Repos/quicklisp/setup.lisp"
                                       (user-homedir-pathname))))
  (when (probe-file quicklisp-init)
    (load quicklisp-init)))

(in-package #:nyxt-user)
```

Ambrevar:
Hmmm... You are building with the makefile, right?

If so, try deleting the following s-exp from it:

    (setf asdf:*default-source-registries* (list (quote asdf/source-registry:environment-source-registry)))

then build again, with a modern ASDF that can be found in the default
ASDF registry.

hendursaga:
@Ambrevar I'm afraid it still doesn't change anything. The makefile calls SBCL with `--no-userinit` and when I run SBCL with that flag and then `(require "asdf")` the version is still the one SBCL comes with. My SBCL config merely requires ASDF and loads Quicklisp. There is no system config.

jmercouris:
Is there a reason we should not merge this PR? I'm having trouble connecting the issues with UIOP mentally to this PR :-D

Ambrevar:
I think we can merge it, it's just that the original author does not
seem to be able to test it :/  So we need to confirm with them if they
are OK merging it as it is :)

hendursaga:
@Ambrevar @jmercouris Oh, sorry, yeah, I was waiting for someone else to mention "let's just merge this" - you can go ahead

Ambrevar:
Done, thanks!

@hendursaga About your issue, feel free to open a new issue about it.  By the way, could you reproduce with ASDF 3.3.5?

PR #2124, opened 2022-02-25
Remove file:// scheme use

### LINKED ISSUE #1454: Consistent file representation
Currently, we have several file representations (this shot us in the foot in #1451):
- `file://` URLs (e.g., in `list-downloads`).
- namestrings.
- pathnames (e.g., in `file-source`).
- `data-path`-s (for persistence).

Maybe we should stick to at most two (URLs and `data-path`? pathnames and `data-path`?).

Alternatively, define `file-designator` (analogy: `url-designator`) and define `path` on it (analogy: `url`) to reliably get the path as a string.

Ambrevar:
Closes #1454.

The rationale is that we don't need `file:` except for interacting with the renderer.

After this change, `cl:pathnames` and `nfiles` are used consistently.  Should be way less confusing.

### DIFF
--- a/source/mode/file-manager.lisp
+++ b/source/mode/file-manager.lisp
@@ -238,31 +238,29 @@ NEW-BUFFER-P defines whether the file/directory is opened in a new buffer.
 SUPPORTED-P says whether the file can be opened by Nyxt.
 
 Can be used as a `open-file-function'."
-  (let ((file-url (if (valid-url-p filename)
-                      (quri:uri filename)
-                      (quri::make-uri-file :path filename))))
-    (handler-case
-        (if supported-p
+  (handler-case
+      (if supported-p
+          (let ((file-url (quri::make-uri-file :path filename)))
             (if new-buffer-p
                 (make-buffer-focus :url file-url)
-                (buffer-load file-url))
-            (let ((process (uiop:launch-program
-                            #+linux
-                            (list *xdg-open-program* (quri:render-uri file-url))
-                            #+darwin
-                            (list "open" (uiop:native-namestring filename))
-                            :error-output :stream)))
-              (nyxt:echo "Opening ~s with ~s." filename *xdg-open-program*)
-              (run-thread "file opener"
-                (let ((status (uiop:wait-process process)))
-                  (unless (= 0 status)
-                    (echo-warning "When opening file ~s with ~s : ~a"
-                                  filename
-                                  *xdg-open-program*
-                                  (alex:read-stream-content-into-string
-                                   (uiop:process-info-error-output process))))))))
-      ;; We can probably signal something and display a notification.
-      (error (c) (log:error "Opening ~a: ~a~&" filename c)))))
+                (buffer-load file-url)))
+          (let ((process (uiop:launch-program
+                          #+linux
+                          (list *xdg-open-program* (quri:render-uri filename))
+                          #+darwin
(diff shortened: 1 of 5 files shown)


source/mode/file-manager.lisp
```diff
@@ -238,31 +238,29 @@ NEW-BUFFER-P defines whether the file/directory is opened in a new buffer.
...
-                                  filename
-                                  *xdg-open-program*
-                                  (alex:read-stream-content-into-string
-                                   (uiop:process-info-error-output process))))))))
-      ;; We can probably signal something and display a notification.
-      (error (c) (log:error "Opening ~a: ~a~&" filename c)))))
+                (buffer-load file-url)))
+          (let ((process (uiop:launch-program
+                          #+linux
+                          (list *xdg-open-program* (quri:render-uri filename))
```
aartaka, line 249: Why `quri:render-url` there?

source/mode/file-manager.lisp
```diff
@@ -238,31 +238,29 @@ NEW-BUFFER-P defines whether the file/directory is opened in a new buffer.
...
+                          #+darwin
+                          (list "open" (uiop:native-namestring filename))
+                          :error-output :stream)))
+            (nyxt:echo "Opening ~s with ~s." filename *xdg-open-program*)
+            (run-thread "file opener"
+              (let ((status (uiop:wait-process process)))
+                (unless (= 0 status)
```
aartaka, line 256: While we're at it, maybe handle `uiop/run-program:subprocess-error` instead?

aartaka:
Looks good, except for some confusions that I've commented on :D

source/mode/file-manager.lisp
jmercouris: Because other programs not recognize the filename otherwise. It must be serialized.

jmercouris:
Thank you PIerre.

source/mode/file-manager.lisp
aartaka: I'm confused. What is `filename` at this point? It should be a string, shouldn't it? `quri:render-uri` gets a `quri:uri`. This should cause a type error, unless I lack some part of the context.

source/mode/file-manager.lisp
aartaka: Fair point :)

source/mode/file-manager.lisp
aartaka: [Yes, it should be a string or a pathname.](https://github.com/atlas-engineer/nyxt/blob/8b9c21a5409f5f2517ef64aa41954b5a41b4e6d5/source/mode/file-manager.lisp#L174)

aartaka: Fixed in 41b4a83f192a49ff00ec95e99e84b10fdcebb76c.

jmercouris: Perhaps I was confused, but looking at the darwin directive below leads me to believe it is a path otherwise `(uiop:native-namestring filename)` should /not/ work! Hence why it seemed logical to me that it should be serialized as it is done.

jmercouris: You are correct Artyom!

Ambrevar: To clarify: yes, it was a typo, I meant uiop:native-namestring indeed!

PR #2126, opened 2022-02-25
source/bookmark.lisp: remove unnecessary sort function.

pdelfino:
During a code review on no-[procrastinate](https://github.com/atlas-engineer/nyxt/pull/1771)-mode with @jmercouris, he brought my attention to the fact that this `sort` is unnecessary (and computationally expensive).

Did I miss something? Or should this PR remove it?

### DIFF
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -274,9 +274,7 @@ rest in background buffers."
         (write-string ")")))))
 
 (defmethod nfiles:serialize ((profile nyxt-profile) (file bookmarks-file) stream &key)
-  (let ((content
-          (sort (nfiles:content file)
-                #'url< :key #'url)))
+  (let ((content (nfiles:content file)))
     (write-string "(" stream)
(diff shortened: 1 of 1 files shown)


Ambrevar:
It is not unnecessary, in fact it does something very useful: make the output file predictable, and thus versionable.  Otherwise every time the bookmarks are persisted the file may change, which would lead your version control system (say git) to generate big diffs even if there was a single change.

Computationally, it's fast enough because:
- Bookmarks are serialized only on user demand (not in a tight loop).
- Bookmarks are in the order of less than 10000.  Sorting 10000 items with SBCL on my machine takes a fraction of a second.

Does that make sense?

pdelfino:
> Does that make sense?

It makes total sense, Pierre. I was indeed missing somethings. Thanks for the explanation!

Ambrevar:
You're welcome.  I've added a comment in ef78a4416f72b514af55adb90b6f5a27c34100a1.

PR #2128, opened 2022-03-04
source/mode/bookmark.lisp: remove unnecessary let expression.

pdelfino:
As pointed out by @jmercouris in a similar code, this nested let construction seems to be unnecessary.

Did I miss something?


Ambrevar:
Good catch!

PR #2130, opened 2022-03-04
source/manual.lisp: insert blocker-mode configuration pointer.

pdelfino:
The manual includes the tutorial, but also focuses on configuration. For me, tweaking with blocker mode is a hidden gem only explicit in the docstring! A pointer should be included on the manual. Thus, this PR :wink:


Ambrevar:
I wonder if we should mention that it defaults to StevenBlack: the default is prone to change and thus the manual would get outdated.

Instead, suggesting how to see and change the value is enough in my opinion.

jmercouris:
A good point.

PR #2139, opened 2022-03-10
User scripts (Greasemonkey/Tampermonkey-like) support

aartaka:
This is a severely under-tested prototype for Greasemonkey-like script support. What it can do:
- Load renderer-independent scripts via `nyxt/web-mode:user-scripts` slot and a new `user-script` class.
- Load scripts by URL, filename and directly as code.
- Load their dependencies by URL and filename (that last one other browsers can't do).
  - [X] Need to fix relative URL resolution, though...
- Everything else Greasemonkey and WebKit scripts can do -- run at document start and end, inject into all frames of main only, run arbitrary JavaScript.

What this cannot do (yet?):
- [ ] Cache scripts locally. When there are more than five scripts, each one requested by `make-greasemonkey-script`, startup time gets to 10+ seconds. This should be fixed.
- Call [Greasemonkey APIs](https://wiki.greasespot.net/Greasemonkey_Manual:API)
  - And `@require` scripts that use those.
- Rely on Chrome [match patterns](https://wiki.greasespot.net/Metadata_Block#.40match), as those are more complicated than what WebKit allows us to do.
  - [ ] Maybe pre-process those then?
- It's severely under-tested, so if anyone has examples of good user-scripts, I'll be glad to test those! @hendursaga, @edgar-vincent?

Looks good?


source/mode/user-script.lisp
```diff
@@ -0,0 +1,91 @@
...
+              (alex:read-file-into-string (quri:uri-path script))
+              (dex:get (quri:render-uri script)))))
+       ((uiop:file-pathname-p script)
+        (alex:read-file-into-string script))
+       (t script)))))
+
+(in-package :nyxt)
+
+(export-always 'make-greasemonkey-script)
+(defun make-greasemonkey-script (greasemonkey-script)
```
Ambrevar, line 59: Maybe make it an `initialize-instance :after` instead with a mandatory argument?

You could also do both: initialize most of it in `initialize-instance :after`, then add a trivial `make-greasemonkey-script` wrapper.

source/mode/user-script.lisp
```diff
@@ -0,0 +1,91 @@
...
+  (sera:and-let* ((code (nyxt/web-mode::get-script-text greasemonkey-script))
+                  (start-position (search "// ==UserScript==" code))
+                  (end-position (search "// ==/UserScript==" code))
+                  (meta (subseq code
```
Ambrevar, line 63: Why call it `meta`?  What do you think of `body` instead?

source/mode/web.lisp
```diff
@@ -75,6 +81,22 @@ and to index the top of the page.")
                    :documentation "Defines which elements are to be hinted. The
 hints-selector syntax is that of CLSS, and broadly, that of CSS. Use it to
 define which elements are picked up by element hinting.")
+   (user-scripts
+    nil
```
Ambrevar, line 85: Style: I prefer `'()` when the type is `list`, to emphasize it's not a boolean.

source/renderer/gtk.lisp
```diff
@@ -1578,43 +1578,51 @@ local anyways, and it's better to refresh it if a load was queried."
       (webkit:webkit-user-content-manager-remove-style-sheet
        content-manager style-sheet))))
 
-(define-ffi-method ffi-buffer-add-user-script ((buffer gtk-buffer) javascript &key
-                                               world-name all-frames-p
-                                               at-document-start-p run-now-p
-                                               allow-list block-list)
+(defvar %user-scripts%
+  (make-hash-table :test 'equalp)
+  "The table mapping `nyxt/web-mode:user-script's to WebKit-specific UserScripts.")
```
Ambrevar, line 1583: Does not seem to be used.  Mistake?

Ambrevar:
Can you explain why we cannot use GreaseMonkey's API?  What would we need?

Note: the CI fails ;)

Otherwise it's looking great, it's a very welcome addition, thanks a lot for your work!

source/mode/user-script.lisp
aartaka: `user-script` can be useful outside Greasemonkey scripts, and it's a good abstraction over what renderers provide us with. That's why `make-greasemonkey-script` should be a separate function, not an `initialize-instance`.

source/mode/user-script.lisp
aartaka: They call it [Metadata Block](https://wiki.greasespot.net/Metadata_Block) on Greasemonkey wiki. It's not a body of the script, in fact. It's a header comment with magic keywords that Greasemonkey/Tampermonkey recognizes. `meta` fits this purpose, I believe -- it's a meta-data about the script that isn't it's source code.

source/mode/web.lisp
aartaka: We should have some linter bot for this type of things that slip out of my mind. Or a Pull Request checklist, at least :D

source/renderer/gtk.lisp
aartaka: Yes, it's a leftover from the initial implementation!

aartaka:
> Can you explain why we cannot use GreaseMonkey's API? What would we need?

We cannot use those, as they are JS functions doing things beyond what JS is usually allowed to do (much like what WebExtensions scripts can do). To enable those, we'd need to implement Greasemonkey APIs as yet another set of WebKitWebExtension C functions so that they are callable on the Lisp side. And I don't want to have anything beyond WebExtensions support in our C library -- it's too much of a temptation to expand ad infinitum then :p 

> Note: the CI fails ;)

Yes, and I know the reason :) Will fix ASAP!

source/mode/user-script.lisp
Ambrevar: If you export the class name, then you can also `(make-instance 'user-script ...)`.

That said, it's good  practice to have the mandatory initialization be done in `initialize-instance :after`, because then you ensure that an object cannot be accidentally initialized with `make-instance` and thus left incomplete.

source/mode/user-script.lisp
Ambrevar: OK!

source/mode/user-script.lisp
aartaka: I mean, the only mandatory slot of `user-script` is `code`, so no chance one can make an incomplete object with `make-instance`, unless deliberately providing nonsensical values :)

Ambrevar: What I meant by "incomplete" is that with initialize-instance :after, you can _guarantee_  that an instance has all its slots set for a given file.  Without it, you could very well have an object that represents a file but without any of its slots set.

aartaka:
OK, the skeleton for user-scripts support is ready. It's slow (as scripts are loaded anew at every startup), it's working with half the scripts (as most of them are not exacty portable and spec-conforming), and it still doesn't have Greasemonkey APIs. 

Scripts that I tested (to the partial success) it with:
```
(define-configuration nyxt/web-mode:web-mode
  ((nyxt/web-mode:user-scripts
    (mapcar
     #'make-greasemonkey-script
     (list
      "https://greasyfork.org/scripts/7543-google-search-extra-buttons/code/Google%20Search%20Extra%20Buttons.user.js"
      (quri:url-encode "https://greasyfork.org/scripts/14146-网页限制解除/code/网页限制解除.user.js")
      "https://greasyfork.org/scripts/423851-simple-youtube-age-restriction-bypass/code/Simple%20YouTube%20Age%20Restriction%20Bypass.user.js"
      "https://greasyfork.org/scripts/38182-hide-youtube-google-ad/code/Hide%20youtube%20google%20ad.user.js"
      "https://greasyfork.org/scripts/4870-maximize-video/code/Maximize%20Video.user.js"
      "https://greasyfork.org/scripts/370246-sci-hub-button/code/Sci-hub%20button.user.js")))))
```

With the only visible success being partial ad dissappearance on YouTube :D

Ambrevar:
I believe fetched code should be validated, e.g. against a checksum or a GPG signature.

I've started working on such verification support on the side of `nfiles`.  The API is there, but the verification process won't be part of the library as we can't know in advance how the user wants to verify the file.

A popular option would be to compute the SHA1 / SHA3 sums.

My benchmarks so far:

- https://github.com/massung/sha1 is much too slow.
- https://github.com/sharplispers/ironclad is pretty good both for sha1 and sha3.
- https://github.com/pmai/sha3 is about as fast a Ironclad and possibly a bit lighter.  Not that it really matters at this level.
- OpenSSL's implementation is about twice as fast as Ironclad and cl-sha3.  I've tested with `openssl dgst -sha3-512 /path/to/file`.

Thoughts?

Ambrevar:
In any case, it could be a good idea to include some checksumming capability in Nyxt.

What we could do is add a wrapper that tests whether `openssl` or `sha1sum` is found in the PATH, run them if they are, otherwise fall back to `cl-sha3` or `ironclad`.

Ambrevar:
@aartaka I've just pushed a big simplification of both blocker-mode and user-scripts thanks to the latest `nfiles`'s `remote-file` class.

Make sure you pull the latest.

Does it do the job as expected?

Note that the first time the script is fetched, it is blocking.  This is by design, so that the user does not load a page without having the script enabled at first.

I've also updated `get-script-text` to make it recursive, thus effectively loading the whole JS tree of `require`.  I haven't tested though.

aartaka:
Regarding the speed of checksums: Ironclad is already included and (as you've said) is fast enough, so I'm on the side of Ironclad. Given that checksum comparison is unlikely to be frequent, it's good enough.

Ambrevar:
I'm not sure we actually depend on Ironclad.

A quick inspection on Guix shows that dexador depends on clack/lack for
the test-suite only, and these 2 seem the reason why Ironclad is dragged
in by Dexador on Guix.

aartaka:
Ummmmmm... how does one create a proper script? Is it `(make-instance 'nyxt/web-mode:user-script :url "https://something.user.js")`, or is there something else?

Ambrevar:
That's it!  But if you want to persist your script locally, you should
provide a `:base-path "user.js"` initarg.

Ambrevar:
Full example:

```lisp
(make-instance 'user-script :base-path "extra-button42.js" :url (quri:uri "https://greasyfork.org/scripts/7543-google-search-extra-buttons/code/Google%20Search%20Extra%20Buttons.user.js"))
```

Ambrevar:
If you confirm that the new nfiles API is satisfactory, I'll release a new version and update the Guix package.

aartaka:
Is it possible to load user-script from the disk?

Ambrevar:
Yes, this is what Nfiles does.

- If you specify the URL _and_ a base-path, then if the local file
  exists, it is loaded, otherwise it downloads the URL to a local file
  and loads this file afterwards.

- If you specify just a base-path (no URL), then the local file is loaded.

Ambrevar:
I see that we've got a `nyxt/web-mode:user-user-script' which is
superclassed by gtk-user-script, but `nyxt/web-mode:user-script'
isn't...

This is a bug with user-classes...  I'm on it!

Ambrevar:
Closing in favour of #2217.

Issue #2053, opened 2022-01-24
Extract title of the web page when executing `bookmark-hint`


pdelfino:
**Is your feature request related to a problem? Please describe.**
The problem is that commands related to bookmarking usually store the title of the page. However, `bookmark-hint` does **not** behave like that. And this is one of the most mnemonic components of the page.

**Describe the solution you'd like**
I wish the command `bookmark-hint` extracted the title of the webpage being hinted _without being need to loaded it._

**Describe alternatives you've considered**
Ideally, it would be better do it without loading the page. But a quick load and then a restoration to the previous condition could do it.
Doing it fast enough, the user would not even what happened.

aartaka:
Several ways to approach this:
- Make a `background-buffer` and load the URL there.
- Make a request to the URL and see what's there in `<title>` tag.
- And yes, loading it in the current buffer, but we'd better avoid user-facing changes, however fast they can be.

guojing0:
Of the first two approaches suggested by @aartaka , which one would you think is faster?

jmercouris:
The second approach is the fastest by a large margin.

Ambrevar:
Fixed with #2143.

PR #2143, opened 2022-03-11
Feature bookmark hint get title

pdelfino:
Hopefully, this PR solves #2053.

There is a small intersection with on-going PR #1771 about the functions `render-host-&-scheme` and `extract-url-title`. Depending on which one gets merged first, I will rebase, and adapt the other.

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -52,6 +52,15 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
          (or (ignore-errors (ffi-display-url url))
              url))))
 
+(export-always 'fetch-url-title)
+(defun fetch-url-title (url)
+  "Return page's title. The URL is fetched, which could explain a possible
+bottleneck."
+  (let* ((html-source (dex:get url))
+         (html-parsed (plump:parse html-source))
+         (title (plump:text (aref (clss:select "title" html-parsed) 0))))
+    title))
+
 (defmacro defmemo (name params &body body) ; TODO: Replace with https://github.com/AccelerationNet/function-cache?
   (alex:with-gensyms (memo-table args result result?)
     `(let ((,memo-table (make-hash-table :test 'equal)))
--- a/source/bookmark.lisp
+++ b/source/bookmark.lisp
@@ -194,6 +194,7 @@ In particular, we ignore the protocol (e.g. HTTP or HTTPS does not matter)."
     (if (not (valid-url-p url))
         (echo "Invalid URL '~a'" url)
         (let* ((url (quri:uri url))
+               (title (fetch-url-title url))
                (tags (prompt
                       :prompt "Tag(s)"
                       :sources (list
@@ -202,7 +203,7 @@ In particular, we ignore the protocol (e.g. HTTP or HTTPS does not matter)."
                                                :multi-selection-p t)
                                 (make-instance 'tag-source
                                                :marks (url-bookmark-tags url))))))
-          (bookmark-add url :tags tags)))))
+          (bookmark-add url :tags tags :title title)))))
 
 (define-command delete-bookmark (&optional urls-or-bookmark-entries)
   "Delete bookmark(s) matching URLS-OR-BOOKMARK-ENTRIES.
--- a/source/mode/element-hint.lisp
+++ b/source/mode/element-hint.lisp
@@ -407,7 +407,15 @@ visible nosave active buffer."
   (query-hints "Bookmark hint"
                (lambda (result)
                  (dolist (url (mapcar #'url result))
-                   (bookmark-url :url url)))
+                   (let ((tags (prompt
+                                :prompt "Tag(s)"
+                                :sources (list
+                                          (make-instance 'prompter:word-source
+                                                         :name "New tags"
+                                                         :multi-selection-p t)
+                                          (make-instance 'nyxt::tag-source
+                                                         :marks (url-bookmark-tags url))))))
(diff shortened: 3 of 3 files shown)


source/urls.lisp
```diff
@@ -52,6 +52,20 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+(export-always 'render-host-&-scheme)
+(defun render-host-&-scheme (url)
+  "Return decoded URL without path, if existent"
+  (format nil "~a://~a" (quri:uri-scheme url)
+          (quri:uri-host url)))
+
+(export-always 'extract-url-title)
+(defun extract-url-title (url)
+  "Return page title"
+  (let* ((html-source (dex:get (render-host-&-scheme url)))
```
Ambrevar, line 64: Why loading the host of URL instead of URL?  Looks like you would get the wrong title then.

source/urls.lisp
```diff
@@ -52,6 +52,20 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
              url))))
 
+(export-always 'render-host-&-scheme)
+(defun render-host-&-scheme (url)
+  "Return decoded URL without path, if existent"
+  (format nil "~a://~a" (quri:uri-scheme url)
+          (quri:uri-host url)))
+
+(export-always 'extract-url-title)
+(defun extract-url-title (url)
```
Ambrevar, line 62: Maybe the name could be more specific.  In this case, we _fetch_ a network resource, while _extract_ implies the data is already present.  So instead I suggest `fetch-url-title`.

source/urls.lisp
```diff
@@ -52,6 +52,20 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
+  "Return page title"
```
Ambrevar, line 63: Missing trailing period.

source/urls.lisp
```diff
@@ -52,6 +52,20 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
          (or (ignore-errors (ffi-display-url url))
              url))))
 
+(export-always 'render-host-&-scheme)
+(defun render-host-&-scheme (url)
+  "Return decoded URL without path, if existent"
```
Ambrevar, line 57: Missing trailing period.

Ambrevar (commented):
Thanks for taking care of this!

source/urls.lisp
pdelfino: You are correct. Thanks! I forgot to adapt this part because the context is similar to no-procrastinate-mode, but different.

pdelfino:
Thanks for the feedback, @Ambrevar. I implemented all the things pointed by you. I think this is ready to be merged.

source/urls.lisp
```diff
@@ -52,6 +52,20 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
...
 
+(export-always 'render-host-and-scheme)
+(defun render-host-and-scheme (url)
+  "Return decoded URL without path, if existent."
+  (format nil "~a://~a" (quri:uri-scheme url)
+          (quri:uri-host url)))
+
+(export-always 'fetch-url-title)
+(defun fetch-url-title (url)
+  "Return page title."
```
Ambrevar, line 63: Maybe in a second line explain that the URL is fetched.  It's important to emphasize when network transactions occur, because they are often a bottleneck.

source/urls.lisp
```diff
@@ -52,6 +52,20 @@ If the URL contains hexadecimal-encoded characters, return their unicode counter
          (or (ignore-errors (ffi-display-url url))
              url))))
 
+(export-always 'render-host-and-scheme)
+(defun render-host-and-scheme (url)
+  "Return decoded URL without path, if existent."
+  (format nil "~a://~a" (quri:uri-scheme url)
+          (quri:uri-host url)))
```
Ambrevar, line 59: This function does not seem to be used anymore.  Remove?

source/urls.lisp
pdelfino: Indeed. It is a command that may take way more time than others to finish.

Ambrevar:
The tests fail on CCL:

```
    × URL without protocol
Raised an error There is no applicable method for the generic function:
        #<STANDARD-GENERIC-FUNCTION HISTORY-TREE:OWNERS #x30200420178F>
      when called with arguments:
        (NIL) (expected: #<QURI.URI.HTTP:URI-HTTPS https://nyxt.atlas.engineer>)
```

My guess is that it does not like a NIL default or something.  Could you have a look?  Don't hesitate if you need help!

jmercouris:
@Ambrevar do you understand why this error is happening? It doesn't make really much sense to me in this context.

pdelfino:
> @pdelfino The content is ready to be merged, but the commit history would need an update. I think 1 commit is enough. Can you squash it?

Ok. There were 5 commits (one had an awful commit message, I re-worded it). There are changes in 3 files. 

Do you mean 1 commit for all 3 changes? Or 1 commit per file that was changed?

I did the latter: 3 commits for changes in 3 different files. 

If you prefer, I can squash again to have 1 commit for changes in 3 different files.

pdelfino:
> Merged with [b579cec](https://github.com/atlas-engineer/nyxt/commit/b579cec170f4440f50fc672e3ff7e78b89d92e80), thanks for this contribution!

Thanks for the attentive code review. Learning a lot :wink:

PR #2150, opened 2022-03-18
Annotations on tutorial

pdelfino:
Considering the questions and the interest from users about the feature on annotations (especially after the [article](https://nyxt.atlas.engineer/article/annotations.org) was published), I suggest the we incorporate it to the tutorial aiming for the 3.0 release. Also, the app grammarly indicated a problem with indefinite article rule in one of the docstrings - that I tried to fix.


jmercouris:
Thanks Pedro, if you fix the typo, it should be ready I believe!

PR #2154, opened 2022-03-23
Reducing type specificity in prompter

K1D77A:
These changes should fix problems when SBCL Type checks class slots (it does not by default). 
Also added a class for headless browser which and fixed a small bug when running headless. Although headless doesn't work :( because it gets stuck in a webkit ffi call.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -240,6 +240,11 @@ prevents otherwise."))
 
 (define-user-class browser)
 
+(define-class headless-browser (browser)
+  ())
+
+(define-user-class headless-browser)
+
 (defmethod external-editor-program ((browser browser))
   (alex:ensure-list (slot-value browser 'external-editor-program)))
 
@@ -294,38 +299,49 @@ prevents otherwise."))
    browser
    (lambda ()
      (run-thread "finalization"
-       ;; Restart on init error, in case `*init-file*' broke the state.
-       ;; We only `handler-case' when there is an init file, this way we avoid
-       ;; looping indefinitely.
-       (if (or (getf *options* :no-init)
-               (not (uiop:file-exists-p (nfiles:expand *init-file*))))
-           (startup browser urls)
-           (catch 'startup-error
-             (handler-bind ((error (lambda (c)
-                                     (log:error "Startup failed (probably due to a mistake in ~s):~&~a"
-                                                (nfiles:expand *init-file*) c)
-                                     (throw 'startup-error
-                                       (if *run-from-repl-p*
-                                           (progn
-                                             (quit)
-                                             (reset-all-user-classes)
-                                             (apply #'start (append *options* (list :urls urls :no-init t))))
-                                           (restart-with-message c))))))
-               (startup browser urls)))))))
+                 ;; Restart on init error, in case `*init-file*' broke the state.
+                 ;; We only `handler-case' when there is an init file, this way we avoid
+                 ;; looping indefinitely.
+                 (if (or (getf *options* :no-init)
+                         (not (uiop:file-exists-p (nfiles:expand *init-file*))))
+                     (startup browser urls)
+                     (catch 'startup-error
+                       (handler-bind ((error (lambda (c)
+                                               (log:error "Startup failed (probably due to a mistake in ~s):~&~a"
+                                                          (nfiles:expand *init-file*) c)
+                                               (throw 'startup-error
+                                                 (if *run-from-repl-p*
+                                                     (progn
+                                                       (quit)
+                                                       (reset-all-user-classes)
+                                                       (apply #'start (append *options* (list :urls urls :no-init t))))
+                                                     (restart-with-message c))))))
+                         (startup browser urls)))))))
   ;; Set `init-time' at the end of finalize to take the complete startup time
   ;; into account.
   (setf (slot-value *browser* 'init-time)
         (local-time:timestamp-difference (local-time:now) startup-timestamp))
   (setf (slot-value *browser* 'ready-p) t))
 
+(defmethod startup ((browser headless-browser) urls)
+  "Startup the headless. Dont restore history but open URLs."
+  (open-urls (or urls (list (default-new-buffer-url browser))))
+  (hooks:run-hook *after-startup-hook*)
+  (funcall* (startup-error-reporter-function *browser*)))
+
 (defmethod startup ((browser browser) urls)
   (labels ((clear-history-owners (buffer)
              "Warning: We clear the previous owners here.
 After this, buffers from a previous session are permanently lost, they cannot be
 restored."
-             (nfiles:with-file-content (history (history-file buffer))
-               (when history
-                 (clrhash (htree:owners history)))))
+             (handler-case 
+                 (nfiles:with-file-content (history (history-file buffer))
+                   (when history
+                     (clrhash (htree:owners history))))
+               (no-applicable-method ()
+                 nil
+                 ;;hack because history file is sometimes nil
+                 )))
            (restore-session ()
              (let ((buffer (current-buffer)))
                (when (histories-list buffer)
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -410,27 +410,27 @@ Return the download object matching the download."
         (:lisp
          (alex:when-let* ((path (download-directory buffer))
                           (download-dir (nfiles:expand path)))
-           (when (eq proxy-url :auto)
-             (setf proxy-url (proxy-url buffer :downloads-only t)))
-           (let* ((download nil))
-             (with-protect ("Download error: ~a" :condition)
-               (nfiles:with-file-content (downloads path)
-                 (setf download
-                       (download-manager:resolve url
-                                                 :directory download-dir
-                                                 :cookies cookies
-                                                 :proxy proxy-url))
-                 (push download downloads)
-                 ;; Add a watcher / renderer for monitoring download
-                 (let ((download-render (make-instance 'user-download :url (render-url url))))
-                   (setf (destination-path download-render)
-                         (uiop:ensure-pathname
-                          (download-manager:filename download)))
-                   (push download-render (downloads *browser*))
-                   (run-thread
-                     "download watcher"
-                     (download-watch download-render download)))
-                 download)))))
+                         (when (eq proxy-url :auto)
+                           (setf proxy-url (proxy-url buffer :downloads-only t)))
+                         (let* ((download nil))
+                           (with-protect ("Download error: ~a" :condition)
+                             (nfiles:with-file-content (downloads path)
+                               (setf download
+                                     (download-manager:resolve url
+                                                               :directory download-dir
+                                                               :cookies cookies
+                                                               :proxy proxy-url))
+                               (push download downloads)
+                               ;; Add a watcher / renderer for monitoring download
+                               (let ((download-render (make-instance 'user-download :url (render-url url))))
+                                 (setf (destination-path download-render)
+                                       (uiop:ensure-pathname
+                                        (download-manager:filename download)))
+                                 (push download-render (downloads *browser*))
+                                 (run-thread
+                                  "download watcher"
+                                  (download-watch download-render download)))
+                               download)))))
         (:renderer
          (ffi-buffer-download buffer (render-url url))))
     (list-downloads)))
@@ -673,17 +673,17 @@ sometimes yields the wrong result."
   (let ((buffer (current-buffer)))
     (values
      (sera:and-let* ((path (nfiles:expand (standard-output-file buffer))))
-       (setf *standard-output*
-             (open path
-                   :direction :output
-                   :if-does-not-exist :create
-                   :if-exists :append)))
+                    (setf *standard-output*
+                          (open path
+                                :direction :output
+                                :if-does-not-exist :create
+                                :if-exists :append)))
      (sera:and-let* ((path (nfiles:expand (error-output-file buffer))))
-       (setf *error-output*
-             (open path
-                   :direction :output
-                   :if-does-not-exist :create
-                   :if-exists :append))))))
+                    (setf *error-output*
+                          (open path
+                                :direction :output
+                                :if-does-not-exist :create
+                                :if-exists :append))))))
 
 (defmacro define-ffi-generic (name arguments &body options)
   `(progn
@@ -749,8 +749,8 @@ sometimes yields the wrong result."
                   (ps:let ((style (ps:chain document body style)))
                     (setf (ps:@ style zoom)
                           (ps:lisp value)))))
-      (with-current-buffer buffer
-        (zoom)))))
+           (with-current-buffer buffer
+             (zoom)))))
 (define-ffi-generic ffi-buffer-get-document (buffer)
   (:method ((buffer buffer))
     (pflet ((get-html (start end)
@@ -758,11 +758,11 @@ sometimes yields the wrong result."
                                                                              (ps:lisp end))))
             (get-html-length ()
                              (ps:chain document document-element |innerHTML| length)))
-      (with-current-buffer buffer
-        (let ((slice-size 10000))
-          (reduce #'str:concat
-                  (loop for i from 0 to (truncate (get-html-length)) by slice-size
-                        collect (get-html i (+ i slice-size)))))))))
+           (with-current-buffer buffer
+             (let ((slice-size 10000))
+               (reduce #'str:concat
+                       (loop for i from 0 to (truncate (get-html-length)) by slice-size
+                             collect (get-html i (+ i slice-size)))))))))
 (define-ffi-generic ffi-generate-input-event (window event))
 (define-ffi-generic ffi-generated-input-event-p (window event))
 (define-ffi-generic ffi-within-renderer-thread (browser thunk)
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -299,24 +299,24 @@ prevents otherwise."))
    browser
    (lambda ()
      (run-thread "finalization"
-                 ;; Restart on init error, in case `*init-file*' broke the state.
-                 ;; We only `handler-case' when there is an init file, this way we avoid
-                 ;; looping indefinitely.
-                 (if (or (getf *options* :no-init)
-                         (not (uiop:file-exists-p (nfiles:expand *init-file*))))
-                     (startup browser urls)
-                     (catch 'startup-error
-                       (handler-bind ((error (lambda (c)
-                                               (log:error "Startup failed (probably due to a mistake in ~s):~&~a"
-                                                          (nfiles:expand *init-file*) c)
-                                               (throw 'startup-error
-                                                 (if *run-from-repl-p*
-                                                     (progn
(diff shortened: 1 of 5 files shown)


source/browser.lisp
```diff
@@ -294,38 +299,49 @@ prevents otherwise."))
...
-                                                (nfiles:expand *init-file*) c)
-                                     (throw 'startup-error
-                                       (if *run-from-repl-p*
-                                           (progn
-                                             (quit)
-                                             (reset-all-user-classes)
-                                             (apply #'start (append *options* (list :urls urls :no-init t))))
-                                           (restart-with-message c))))))
-               (startup browser urls)))))))
+                 ;; Restart on init error, in case `*init-file*' broke the state.
```
aartaka, line 302: The indent seems off, please make sure the whole Nyxt image is loaded so that the indent is macro-aware.

source/browser.lisp
```diff
@@ -294,38 +299,49 @@ prevents otherwise."))
...
+                                                       (apply #'start (append *options* (list :urls urls :no-init t))))
+                                                     (restart-with-message c))))))
+                         (startup browser urls)))))))
   ;; Set `init-time' at the end of finalize to take the complete startup time
   ;; into account.
   (setf (slot-value *browser* 'init-time)
         (local-time:timestamp-difference (local-time:now) startup-timestamp))
   (setf (slot-value *browser* 'ready-p) t))
 
+(defmethod startup ((browser headless-browser) urls)
```
aartaka, line 326: Why avoiding history restoration? Does it break something?

source/browser.lisp
```diff
@@ -294,38 +299,49 @@ prevents otherwise."))
...
-             (nfiles:with-file-content (history (history-file buffer))
-               (when history
-                 (clrhash (htree:owners history)))))
+             (handler-case 
+                 (nfiles:with-file-content (history (history-file buffer))
+                   (when history
+                     (clrhash (htree:owners history))))
+               (no-applicable-method ()
+                 nil
+                 ;;hack because history file is sometimes nil
```
aartaka, line 343: That should not happen. Ever. If that happens, it's a bug.

source/browser.lisp
```diff
@@ -394,27 +410,27 @@ Return the download object matching the download."
...
-                 (let ((download-render (make-instance 'user-download :url (render-url url))))
-                   (setf (destination-path download-render)
-                         (uiop:ensure-pathname
-                          (download-manager:filename download)))
-                   (push download-render (downloads *browser*))
-                   (run-thread
-                     "download watcher"
-                     (download-watch download-render download)))
-                 download)))))
+                         (when (eq proxy-url :auto)
```
aartaka, line 413: Indent is off.

source/browser.lisp
```diff
@@ -657,17 +673,17 @@ sometimes yields the wrong result."
   (let ((buffer (current-buffer)))
     (values
      (sera:and-let* ((path (nfiles:expand (standard-output-file buffer))))
-       (setf *standard-output*
-             (open path
-                   :direction :output
-                   :if-does-not-exist :create
-                   :if-exists :append)))
+                    (setf *standard-output*
```
aartaka, line 676: Indent is off.

source/browser.lisp
```diff
@@ -657,17 +673,17 @@ sometimes yields the wrong result."
...
+                                :direction :output
+                                :if-does-not-exist :create
+                                :if-exists :append)))
      (sera:and-let* ((path (nfiles:expand (error-output-file buffer))))
-       (setf *error-output*
-             (open path
-                   :direction :output
-                   :if-does-not-exist :create
-                   :if-exists :append))))))
+                    (setf *error-output*
```
aartaka, line 682: Indent is off.

source/browser.lisp
```diff
@@ -733,20 +749,20 @@ sometimes yields the wrong result."
                   (ps:let ((style (ps:chain document body style)))
                     (setf (ps:@ style zoom)
                           (ps:lisp value)))))
-      (with-current-buffer buffer
-        (zoom)))))
+           (with-current-buffer buffer
```
aartaka, line 752: Indent.

source/browser.lisp
```diff
@@ -733,20 +749,20 @@ sometimes yields the wrong result."
...
                       (ps:chain document document-element |innerHTML| (slice (ps:lisp start)
                                                                              (ps:lisp end))))
             (get-html-length ()
                              (ps:chain document document-element |innerHTML| length)))
-      (with-current-buffer buffer
-        (let ((slice-size 10000))
-          (reduce #'str:concat
-                  (loop for i from 0 to (truncate (get-html-length)) by slice-size
-                        collect (get-html i (+ i slice-size)))))))))
+           (with-current-buffer buffer
```
aartaka, line 761: Indent.

source/start.lisp
```diff
@@ -56,103 +56,103 @@ For instance, if we want to load some Slynk configuration code that lives in
   (uiop:quit))
 
 (sera:eval-always
-  (defun define-opts ()
```
aartaka, line 59: The indent is terribly off. Was there any change?

source/start.lisp
```diff
@@ -210,7 +210,7 @@ before running this command."
             (file-position stream p)))))))
 
 (cffi:defcallback handle-interrupt
-    :void ((signum :int) (siginfo :pointer) (ptr :pointer))
+  :void ((signum :int) (siginfo :pointer) (ptr :pointer))
```
aartaka, line 213: Indent.

source/start.lisp
```diff
@@ -414,7 +414,7 @@ Otherwise bind socket and return the listening thread."
        (log:info "Listening to socket ~s." socket-path)
        (uiop:delete-file-if-exists socket-path) ; Safe since socket-path is a :socket at this point.
        (run-thread "socket listener"
-         (listen-socket))))))
```
aartaka, line 417: Indent.

source/configuration.lisp
```diff
@@ -66,9 +66,12 @@ This is set globally so that extensions can be loaded even if there is no
   (or %buffer
       (alex:if-let ((w (or window (current-window))))
         (active-buffer w)
-        (when *browser*
-             (log:debug "No active window, picking last active buffer.")
-             (last-active-buffer)))))
+        (if *browser*
```
aartaka, line 69: I still doubt that this is necessary. If `*browser*` is `nil`, then either
- Nyxt hasn't started yet.
- Something is terribly broken.

source/buffer.lisp
```diff
@@ -86,7 +89,7 @@ Auto-completions come from the default search engine.")
                           #'(lambda (results)
                               (alex:when-let* ((results results)
                                                (results (decode-json results)))
-                                (mapcar #'list (second results) (fourth results))))))
```
aartaka, line 89: Indent.

source/buffer.lisp
```diff
@@ -234,64 +237,64 @@ A value of 0.95 means that the bottom 5% will be the top 5% when scrolling
 down.")
    (style
     (theme:themed-css
-        (theme *browser*)
```
aartaka, line 237: Has something changed in this snippet? I cannot see the change behind the broken intentation...

source/buffer.lisp
```diff
@@ -234,64 +237,64 @@ A value of 0.95 means that the bottom 5% will be the top 5% when scrolling
...
+      :border-style "solid"
+      :color theme:text
+      :background-color theme:background)
+     (th
+      :background-color theme:primary
+      :color theme:background
+      :text-align "left")))
    (buffer-load-hook
     (make-instance 'hook-url->url
-     :combination #'hooks:combine-composed-hook)
```
aartaka, line 294: Indentation.

source/buffer.lisp
```diff
@@ -539,41 +546,41 @@ inherited from the superclasses."))
 (define-class panel-buffer (user-web-buffer)
   ((width 250 :documentation "The width in pixels.")
    (style (theme:themed-css (theme *browser*)
-            (body
```
aartaka, line 542: Indent.

source/buffer.lisp
```diff
@@ -1039,8 +1046,8 @@ If URL is `:default', use `default-new-buffer-url'."
     (sera:and-let* ((owner (htree:owner history (id buffer)))
                     (current (htree:current owner))
                     (data (htree:data current)))
-      (setf (nyxt::scroll-position data) (nyxt:document-scroll-position buffer))
```
aartaka, line 1042: Intentation.

source/buffer.lisp
```diff
@@ -1221,7 +1228,7 @@ second latest buffer first."
                                :actions (list (make-mapped-command buffer-delete))))))
 
 (define-internal-page-command-global reduce-to-buffer (&key (delete t))
-    (reduced-buffer "*Reduced Buffers*" 'base-mode)
```
aartaka, line 1224: Indentation.

source/buffer.lisp
```diff
@@ -1441,11 +1448,11 @@ Finally, if nothing else, set the `engine' to the `default-search-engine'."))
                                 (engine (default-search-engine))
                                 (completion (completion-function engine))
                                 (all-terms (str:join " " terms)))
-                  (mapcar (alex:rcurry #'make-completion-query
```
aartaka, line 1444: Indentation.

source/buffer.lisp
```diff
@@ -1577,7 +1584,7 @@ HISTORY may be NIL for buffers without history."
            (sera:equals (id buffer))
            buffers
            :key (lambda (b) (alex:when-let ((owner (htree:owner history (id b))))
-                              (htree:creator-id owner))))
```
aartaka, line 1580: Indentation.

source/buffer.lisp
```diff
@@ -1610,41 +1617,41 @@ The tree is browsed in a depth-first fashion.
 When there is no previous buffer, go to the last one so as to cycle."
   (labels ((buffer-last-child (&optional (buffer (current-buffer)))
              (alex:if-let ((next-siblings (second (buffer-siblings buffer))))
-                          (buffer-last-child (alex:last-elt next-siblings))
```
aartaka, line 1613: Indentation.

aartaka (changes requested):
There's a lot of broken indentation, can you fix it, please? :)

Other then that, I have some questions:
- What's the purpose of `headless-browser`?
- In which exact cases does `last-access`, `*browser*`, `current-buffer`, and `current-window` are nil? It sounds like a terribly broken Lisp image...

aartaka:
The prompter fixes are fine and I'm ready to accept those right away! The rest we need to discuss...

K1D77A:
> There's a lot of broken indentation, can you fix it, please? :)
> 
> Other then that, I have some questions:
> 
>     * What's the purpose of `headless-browser`?
> 
>     * In which exact cases does `last-access`, `*browser*`, `current-buffer`, and `current-window` are nil? It sounds like a terribly broken Lisp image...
Try running in headless mode and you will see... Headless doesn't work properly and I had to make all of the changes you can see in order to figure out what part of the program is broken in order to actually find the problem. The reason that current-buffer returns nil is because the after method on initialize-instance for buffers never returns. In this case current-buffer is nil and it breaks the entire system. Signalling a condition in this instance is more helpful. 
The headless class is so specific behaviour can be specialized for the headless browser.

I'll try to fix the indenting problems.

source/browser.lisp
K1D77A: It does happen when trying to run in headless mode.

aartaka:
Note that sometimes you need to `sleep 1` for things to initialize properly. We tried to mitigate it, but it's unpredictable at times. 

Here's a script that worked for me, fetching all the repos I have available at Github:

``` lisp
;;; script.lisp
(in-package :nyxt)

(define-configuration browser
  ((session-restore-prompt :never-restore)))

(once-on nyxt:*after-startup-hook* ()
  (run-thread "load github"
    (buffer-load (quri:uri "https://github.com")))
  (on (buffer-loaded-hook (current-buffer)) buffer
    (when (string= "github.com" (quri:uri-host (url buffer)))
      ;; Note this sleep here. document-model takes some time to fill somewhy. Dunno why, it's a bug.
      (sleep 1)
      (alex:write-string-into-file
       (str:join nyxt::+newline+
                 (mapcar (alex:curry #'string-trim serapeum:whitespace)
                         (map 'list #'plump:text (clss:select "div#repos-container li.source"
                                                   (document-model buffer)))))
       #p"/home/aartaka/.config/nyxt/github-repos.txt"
       :if-exists :supersede
       :if-does-not-exist :create)
      (quit))))
```

It works when used with `nyxt --headless --verbose --init script.lisp`.

aartaka:
And yes, `--init` is super unintuitive here. It should be `--script`, but `--script` doesn't work for this case. See 2155.

K1D77A:
> Note that sometimes you need to `sleep 1` for things to initialize properly. We tried to mitigate it, but it's unpredictable at times.
> 
> Here's a script that worked for me, fetching all the repos I have available at Github:
> 
> ```lisp
> ;;; script.lisp
> (in-package :nyxt)
> 
> (define-configuration browser
>   ((session-restore-prompt :never-restore)))
> 
> (once-on nyxt:*after-startup-hook* ()
>   (run-thread "load github"
>     (buffer-load (quri:uri "https://github.com")))
>   (on (buffer-loaded-hook (current-buffer)) buffer
>     (when (string= "github.com" (quri:uri-host (url buffer)))
>       ;; Note this sleep here. document-model takes some time to fill somewhy. Dunno why, it's a bug.
>       (sleep 1)
>       (alex:write-string-into-file
>        (str:join nyxt::+newline+
>                  (mapcar (alex:curry #'string-trim serapeum:whitespace)
>                          (map 'list #'plump:text (clss:select "div#repos-container li.source"
>                                                    (document-model buffer)))))
>        #p"/home/aartaka/.config/nyxt/github-repos.txt"
>        :if-exists :supersede
>        :if-does-not-exist :create)
>       (quit))))
> ```
> 
> It works when used with `nyxt --headless --verbose --init script.lisp`.

I have been trying to load nyxt headless from the repl.

aartaka:
Yup, the same script works when started with (the exact duplicate of the CLI version) `(nyxt:start :headless t :init "/home/aartaka/.config/nyxt/script.lisp")` from the REPL.

K1D77A:
image

The document model doesn't load even for the default webpage:

image

image


image

Nothing

aartaka:
Do you close the history prompt?

K1D77A:
> Do you close the history prompt?

Can't it default to being shut for headless mode? What history do you need when running it in a programmatic way?

K1D77A:
> Do you close the history prompt?

I added 
 (define-configuration headless-browser
        ((session-restore-prompt :never-restore)))
    and 
 (define-configurationbrowser
        ((session-restore-prompt :never-restore)))
 In two different sessions and neither worked. Its always nil.

aartaka:
Do you add it to your init file?

aartaka:
It works only when ran before the `*browser*` is set.

aartaka:
> Can't it default to being shut for headless mode? What history do you need when running it in a programmatic way?

It can, but I'd rather have headless mode be the exact replica of the non-headless mode in as much behaviors as possible.

K1D77A:
image
image

after open-urls call
image

aartaka:
Do you add this configuration before `nyxt:start`?

aartaka:
I see.

K1D77A:
> 



> Do you add this configuration before `nyxt:start`?

Yes

K1D77A:
> > Can't it default to being shut for headless mode? What history do you need when running it in a programmatic way?
> 
> It can, but I'd rather have headless mode be the exact replica of the non-headless mode in as much behaviors as possible.

Personally I think headless should be oriented towards webscraping, this means much of the default functionality can be stripped out for the sake of performance.

aartaka:
Does it report something like `<INFO> [14:32:18] Finished loading "nyxt:help".`? If not, something interferes with the buffer creation.

aartaka:
> Personally I think headless should be oriented towards webscraping, this means much of the default functionality can be stripped out for the sake of performance.

This is one of the possible use cases, yes. There can be more, though. You can optimize your Nyxt this way, but that won't necessarily work for everyone.

K1D77A:
> Does it report something like `<INFO> [14:32:18] Finished loading "nyxt:help".`? If not, something interferes with the buffer creation.

Nope. Yesterday I believe I traced this problem to 
(defmethod initialize-instance :after ((buffer gtk-buffer) &key extra-modes
                                                             no-hook-p
                                       &allow-other-keys)

aartaka:
Can you try adding the `session-restore-prompt` snippet to your config? Does it change the situation?

K1D77A:
> Can you try adding the `session-restore-prompt` snippet to your config? Does it change the situation?

Okay I added them to a file then loaded that with the :init flag, but still no go on the document-model.

K1D77A:
image

image

image

aartaka:
Here's a session I just had:
```
CL-USER> (nyxt:start :headless t :no-init t)
<INFO> [15:46:48] Listening to socket #P"/run/user/1000/nyxt/nyxt.socket".
Nyxt version 3-479-g75977e76
NIL
<INFO> [15:46:48] Loading #P"/home/aartaka/.local/share/nyxt/history/default.lisp".
CL-USER> (nyxt:current-prompt-buffer)
#<NYXT:USER-PROMPT-BUFFER 1072 {100F455FB3}>
CL-USER> (nyxt/prompt-buffer-mode:return-selection)
<INFO> [15:47:11] Loading #P"/home/aartaka/.local/share/nyxt/history/default-tmpFW30LZWC.lisp".
T
<INFO> [15:47:11] Loading "about:blank".
<INFO> [15:47:11] Loading "nyxt:help".
<INFO> [15:47:11] Finished loading "about:blank".
<INFO> [15:47:11] Finished loading "nyxt:help".
CL-USER> (nyxt:quit)
<INFO> [15:47:25] Deleting socket #P"/run/user/1000/nyxt/nyxt.socket".
NIL
```

What this means is that there was a prompt buffer open that prevented initial buffers from being created. Can you check the `current-prompt-buffer` and return the selection if there is one? It could well be the case that it's still blocking the renderer somewhy.

On a related note, do you restart your Lisp image? Experimenting with what I suggested in the same image may not work, as the image already has redefined Nyxt classes and thus will behave the way initially defined, instead of what you set up.

K1D77A:
image

Yes I do restart everytime.

aartaka:
And note that I'm testing things on the upstream Nyxt, can you do that too, so that we can see where our results differ?

aartaka:
Oh wait, but why do you load `:nyxt` instead of `:nyxt/gtk`?

aartaka:
If there's no renderer, then there's nothing to load pages with :thinking:

K1D77A:
> If there's no renderer, then there's nothing to load pages with thinking

Oops. Sorry for wasting your time :*( perhaps a condition that is signalled if a renderer is not loaded when #'start is called might be worth adding?

K1D77A:
I'm going to close this. Refork and just make the changes to prompter then submit the pull.

PR #2156, opened 2022-03-23
reduce type specificity to fix debug 3 safety 3 speed 0 on sbcl

K1D77A:
I think the indenting might be incorrect... I am using emacs with sly and have 1tab = 4 spaces... If I'm missing something here about indenting please let me know.


K1D77A:
Okay I think I got it :facepalm:

aartaka:
Can you squash commits into one?

Ambrevar:
Why was this closed?

Ambrevar:
Sorry, just got it that it was merged separately.

PR #2157, opened 2022-03-23
reduce type specificity for valid type checks in prompter

K1D77A:
I couldn't squash :(


aartaka:
Thanks for this patch!

PR #2159, opened 2022-03-23
 add headless-browser. Add condition when +renderer+ isn't set 

K1D77A:
Added headless-browser class. Add condition when +renderer+ isn't set. Also cleaned up the options so no more (getf ..) 
 I dont know if the commits are correct this is the first time I've contributed to a project like this sorry :(

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -238,7 +238,12 @@ windows.
 A typical Nyxt session encompasses a single instance of this class, but nothing
 prevents otherwise."))
 
+(define-class headless-browser (browser)
+  ())
+
 (define-user-class browser)
+
+(define-user-class headless-browser)
 
 (defmethod external-editor-program ((browser browser))
   (alex:ensure-list (slot-value browser 'external-editor-program)))
@@ -412,7 +417,7 @@ Return the download object matching the download."
                           (download-manager:filename download)))
                    (push download-render (downloads *browser*))
                    (run-thread
-                     "download watcher"
+                       "download watcher"
                      (download-watch download-render download)))
                  download)))))
         (:renderer
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -238,12 +238,10 @@ windows.
 A typical Nyxt session encompasses a single instance of this class, but nothing
 prevents otherwise."))
 
-(define-class headless-browser (browser)
-  ())
 
 (define-user-class browser)
 
-(define-user-class headless-browser)
+;;;potential for a headless-browser
 
 (defmethod external-editor-program ((browser browser))
   (alex:ensure-list (slot-value browser 'external-editor-program)))
--- a/source/start.lisp
+++ b/source/start.lisp
@@ -290,8 +290,8 @@ Return the short error message and the full error message as second value."
    (make-instance 'nyxt/file-manager-mode:user-file-source
                   :extensions '("lisp")
                   :actions (list (make-command load-file* (files)
-                                               (dolist (file files)
-                                                 (load-lisp file)))))))
+                                   (dolist (file files)
+                                     (load-lisp file)))))))
 
 (define-command load-init-file (&key (init-file (nfiles:expand *init-file*)))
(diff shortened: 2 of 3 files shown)


source/browser.lisp
```diff
@@ -238,8 +238,13 @@ windows.
 A typical Nyxt session encompasses a single instance of this class, but nothing
 prevents otherwise."))
 
+(define-class headless-browser (browser)
```
aartaka, line 241: Huh, you seem to like this solution :D

However, do we really need it? Is there another way things can be done in, without introducing yet another fundamental entity?

source/start.lisp
```diff
@@ -465,9 +465,11 @@ Examples:
   (nyxt:start :urls '(\"https://nyxt.atlas.engineer\" \"https://en.wikipedia.org\")
               :verbose t
               :with-file '(\"history\" \"/tmp/nyxt/history.lisp\"))"
-          (with-output-to-string (s) (opts:describe :stream s)))
+            (with-output-to-string (s) (opts:describe :stream s)))
   ;; Extensions should be made accessible straight from the beginning,
   ;; e.g. before a script is run.
+  (unless +renderer+
```
aartaka, line 471: I'm thinking of whether the error is the right approach. Maybe rather do `log:warn`? While I cannot imagine one loading bare Nyxt and starting it, there could be use-cases for it...

source/start.lisp
```diff
@@ -476,54 +478,55 @@ Examples:
 
   ;; Options should be accessible anytime, even when run from the REPL.
   (setf *options* options)
-
```
aartaka, line 479: Ha, that's a smart one! This code is there for a really long time, and no one thought that destructuring can clarify it.

@Ambrevar, any reason it was done with `getf`? I remember that dome operations on plists require understanding whther the key is set to `nil` or missing, but that's not the case, right?

source/start.lisp
```diff
@@ -571,10 +574,13 @@ Finally, run the browser, load URL-STRINGS if any, then run
                  (echo-warning "~a." message)
                  (error-in-new-window "*Init file errors*" full-message)))))
       (load-or-eval :remote nil)
-      (setf *browser* (make-instance 'user-browser
-                                     :startup-error-reporter-function startup-error-reporter
-                                     :startup-timestamp startup-timestamp
-                                     :socket-thread thread))
+      (setf *browser* (make-instance
+                       (if *headless-p*
+                           'headless-user-browser
```
aartaka, line 579: It probably should be `user-headless-browser`. 

Anyway, I'm still hesitant about accepting the necessity of `headless-browser`...

source/start.lisp
K1D77A: Okay sounds like a plan :+1:

source/browser.lisp
K1D77A: Yes I am a big fan of CLOS. Seems to me that using a subclass of the normal class will allow the user to create specific behaviour like I had done previously.

aartaka: Yes, it's convenient, but the same thing can be done with an `:around` method (also part of CLOS) and a check for `*headless-p*`. Introduction of another entity is not strictly necessary here.

aartaka:
You're gonna hate me after this request, but can you squash these commits into one now that the changes are more or less homogenous?

Indentation and addition of blank lines (are they necessary?) belong to a separate commit (or even PR), as an exception to the rule.

K1D77A:
> You're gonna hate me after this request, but can you squash these commits into one now that the changes are more or less homogenous?
> 
> Indentation and addition of blank lines (are they necessary?) belong to a separate commit (or even PR), as an exception to the rule.

I keep trying to squash the commits but git complains. I will just make a new one tomorrow :facepalm:

aartaka:
No problem, take your time :)

source/start.lisp
Ambrevar: No reason I believe.

PR #2163, opened 2022-03-24
source/tutorial: add missing commands related to bookmarks.

pdelfino:


aartaka:
This one is clearly a positive change, no reason to open a PR for it :)

Issue #2000, opened 2022-01-06
vi modes inhibit input events on mobile


efimerspan:
**Describe the bug**
Activating `vi-insert-mode` or `vi-normal-mode` on mobile devices that make use of touch input events will cause events such scrolling or clicking on hyperlinks to not work at all, without any kind of feedback. However, disabling the left-click keybinding for `vi-mode` as defined [here](https://github.com/atlas-engineer/nyxt/blob/ceabd7feae945f07825b394cfb6378ca831de803/source/mode/vi.lisp#L36), will dismiss this issue and allow for normal behavior.

**Precise recipe to reproduce the issue**
Compile Nyxt from source in Mobian, launch the resulting executable and navigate to a `web-mode` buffer. Try scrolling via touch input without any hiccups. Activate `vi-normal-mode` through the prompt buffer and experience the issue in question.

For website-specific issues:
Can you reproduce this issue with Epiphany / GNOME Web (https://wiki.gnome.org/Apps/Web)?
No.

**Information**
- OS name+version: Linux mobian 5.14-sdm845 SMP PREEMPT Sun Jul 25 15:38:14 UTC 2021 aarch64 GNU/Linux
- Graphics card and driver: Adreno 630
- Desktop environment / Window manager name+version: mobian-phosh 0.3.1
- How you installed Nyxt (Guix pack, package manager, build from source): Build from source.
- Information from `show-system-information`:

```
Nyxt version: 2.1.1-1257-g226606f7
Renderer version: GI-GTK
Operating system kernel: Linux 5.14-sdm845
Lisp implementation: SBCL 2.1.11.debian (Dynamic space size: 3221225472)
Features: (:WEBKIT2 :WEBKIT2-2.34 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
[...]

PR #2165, opened 2022-03-27
Repeat key

### LINKED ISSUE #1601: Feature request: a universal-argument command for Nyxt, inspired by Emacs's `C-u`
I would like to have a command to loop (iterate) the execution of other commands.

Nyxt now has the command `go-next` which goes to the next page while reading [GNU manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Prefix-Command-Arguments.html) or a [page](https://github.com/atlas-engineer/nyxt/issues) in Github issues.

**First example:**
Suppose a web page is showing the pagination:
1,2,3,4,5,6,8..... 100,101,102,103.
Being on page 1, I would like to go to page 25. So, I could do:

`M-x` `universal-argument` `go-next` 25

**Second Example**:
Suppose I am reading a Twitter feed. If I scroll to the bottom, more content is loaded. Suppose I want to load a lot of content and I do not want to use the mouse. I would do:
`M-x` `universal-argument` `scroll-to-bottom` 10

----

@jmercouris is building something like a DSL/low-code for Nyxt. This is the [inspiration](https://github.com/disconcision/fructure).

The examples above could be possible applications of his creations.

### LINKED ISSUE #1756: Is it possible for commands to get a numeric argument?
This is not a bug report. It is a question.

# Table of Contents

1.  [Prologue](#org3f4ef86)
    1.  [Numeric arguments in Emacs](#org7a65101)
    2.  [Relation of `C-u` and numeric arguments](#orgaf4290c)
    3.  [More information on this topic](#org07c7e29)
2.  [The context](#org2e16dc8)
3.  [The question](#orgfb07a4e)
4.  [Additional context](#org7a92ed3)

<a id="org3f4ef86"></a>

# Prologue

<a id="org7a65101"></a>

## Numeric arguments in Emacs

In Emacs, it is possible to provide a function with a numeric argument so that the behavior of a function is changed.

Some functions consider the numeric argument as the number of times the function need to be repeated. Here are some examples:

-   When `open-line` is provided with a positive numeric argument, let's call it N, N newlines are inserted in point.
-   When `kill-line` is provided with a positive numeric argument, let's call it N, N lines below the current point are killed.

Other functions use the numeric argument as a modifier of the function. Here are some examples:

-   `buffer-menu` show a list of buffers. When the function is called with a numeric argument equal to 4, only those buffers visiting files are listed.
-   `list-directory` executes `ls` in a given directory. When the function is called with a numeric argument equal to 4, `ls -l` is executed instead.
-   `where-is` shows which keys are mapped to a given function. When the function is called with a numeric argument equal to 4, the information is inserted in the current point.
-   `goto-line` prompts for a line numbering and moves the cursor to that specific line. When the function is provided with a numeric argument, the cursor is moved to the line whose numbering is the numeric argument (e.g. C-u 30 M-g M-g moves the cursor to the 30th line)

<a id="orgaf4290c"></a>

## Relation of `C-u` and numeric arguments

`C-u` provides functions with a numeric argument equal to 4, `C-u C-u <<function call>>` with 16, `C-u C-u C-u` with 64, and so on.

<a id="org07c7e29"></a>

## More information on this topic

More information on this topic can be found in `(info "(emacs) Arguments")`. For those who don't have Emacs running, you can find that Info node in [this](https://www.gnu.org/software/emacs/manual/html_node/emacs/Arguments.html) URL.

<a id="org2e16dc8"></a>

# The context

I have created the following command for creating an Org Mode headline of the currently visited URL.

```lisp
(define-command-global my/org-current-site-create-headline-and-copy ()
  "Create a headline for the current site and insert it to the
clipboard.

The title and the URL of the current buffer is used as the title of
the headline and the LINK property of that headline, respectively.

If the current buffer doesn't have a URL, then nothing is done.

For example, calling this command in the Emacs official site will
insert the following in the clipboard

* GNU Emacs - GNU Project
:PROPERTIES:
:LINK: https://www.gnu.org/software/emacs
:END:
"
  (let ((url (render-url (url (current-buffer))))
        (title (title (current-buffer)))
        data)
    (cond ((equal url "")
           (echo-warning "The current buffer doesn't have a URL"))
          (t
           (when (equal title "")
             (setq title "Unknown title"))
           (setq data (format nil "~{~A~^~%~}~%"
                              (list
                               ;; TODO: With a N prefix argument,
                               ;; repeat the "*" character N times.
                               (str:concat "* " title)
                               ":PROPERTIES:"
                               (str:concat ":LINK: " url)
                               ":END:")))
           (echo "Copied to clipboard ~A" data)
           (copy-to-clipboard data)))))
```

As you can read in the `TODO` comment, I want to provide the command with a numeric argument, so that as many asterisks as the numeric argument are used for the created Org Mode headline.

<a id="orgfb07a4e"></a>

# The question

Do Nyxt commands currently support getting the numeric argument? If not, are there any plans on making Nyxt support this?

<a id="org7a92ed3"></a>

# Additional context

I've inspected the source code and I've noticed that `C-u` is included in the definition of mappings (see below), and this makes me think that Nyxt don't currently support providing commands with the numeric argument.

I asked on #nyxt for this, and I was told that a current workaround would be to map 1 through 9 to different calls of the command (i.e. `C-u 1 {command}` `C-u 2 {command}`, &#x2026;, `C-u 9 {command}`. While this solution might suit my needs for this scenario, I thought that some day I will want to do something more complex and it would be difficult to map all the numbers, so I decided to create this issue. In addition to that, I would rather not having to map all the possible numbers.

aartaka:
This introduces the `repeat-key` command to repeat the key-bound command an arbitrary number of times. This patch also includes:
- Conveninent bindings in Emacs (`M-1`, `M-2` etc.) and VI (mere digits) modes for pre-configured `repeat-key`.
- A major refactoring of the key-dispatch API, introducing `command-dispatcher` and `input-skip-dispatcher` slots of `window`, so that one doesn't have to re-implement `dispatch-key-event` every time they need to process the pressed key.
- `describe-key` nicely rewritten using the new API, as a proof of how much lines it saves :D  

# Things to Discuss 
- I don't particularly like `input-skip-dispatcher`.
  - Maybe only allow changing `command-dispatcher` for a one-shot use, binding it back to `dispatch-command` after _any_ key is processed? 
  -  Suggestions are welcome, as I'm a bit lost in our input handling :D
- When `input-skip-dispatcher` is ran and echoing some message (like in both `describe-key` and `repeat-key`), the echoed message is immeduately flushed somewhy. Anyone knows the reason? Is that the keypress that flushes the message area? How do I override this behavior?

Closes #1601 
Closes #1756

### DIFF
--- a/source/input.lisp
+++ b/source/input.lisp
@@ -105,6 +105,10 @@ KEYCODE-LESS-DISPLAY (KEYCODE-DISPLAY)."
 (defun dispatch-command (function)
   (run-async function))
 
+(export-always 'dispatch-input-skip)
+(defun dispatch-input-skip (keyspecs)
+  (log:debug "Skipping input event key ~s" keyspecs))
+
 (export-always 'dispatch-input-event)
 (defun dispatch-input-event (event buffer window printable-p)
   "Dispatch keys in WINDOW `key-stack'.
@@ -157,13 +161,14 @@ Return nil to forward to renderer or non-nil otherwise."
              ((or (and buffer (forward-input-events-p buffer))
                   (pointer-event-p (first (last key-stack))))
               (log:debug "Forward key ~s" (keyspecs key-stack))
+              (funcall (input-skip-dispatcher window) (keyspecs key-stack))
               (setf key-stack nil)
               nil)
 
              ((and buffer (not (forward-input-events-p buffer)))
               ;; After checking `pointer-event-p', otherwise pointer events
               ;; might not be forwarded.
-              (log:debug "Skipping input event key ~s" (keyspecs key-stack))
+              (funcall (input-skip-dispatcher window) (keyspecs key-stack))
               (setf key-stack nil)
               t)
 
--- a/source/input.lisp
+++ b/source/input.lisp
@@ -101,6 +101,10 @@ KEYCODE-LESS-DISPLAY (KEYCODE-DISPLAY)."
         (format nil "~s [~a]" no-code-specs (keymap:keys->keyspecs keys))
         (format nil "~s" no-code-specs))))
 
+(export-always 'dispatch-command)
+(defun dispatch-command (function)
+  (run-async function))
+
 (export-always 'dispatch-input-event)
 (defun dispatch-input-event (event buffer window printable-p)
   "Dispatch keys in WINDOW `key-stack'.
@@ -144,7 +148,7 @@ Return nil to forward to renderer or non-nil otherwise."
               ;; thread.
               (setf (last-key window) (first key-stack))
               (unwind-protect
-                   (run-async bound-function)
+                   (funcall (command-dispatcher window) bound-function)
                 ;; We must reset the key-stack on errors or else all subsequent
                 ;; keypresses will keep triggering the same erroring command.
                 (setf key-stack nil))
--- a/source/mode/base.lisp
+++ b/source/mode/base.lisp
@@ -3,6 +3,13 @@
 
 (in-package :nyxt)
 
+(defmacro make-repeat (times)
+  (let ((name (gensym (format nil "repeat-~a-" times))))
+    `(make-command ,name ()
+       ,(format nil "Repeat the command associated to the pressed key ~R times." times)
+       ;; FIXME: A dirty hack.
+       (funcall (read-from-string "nyxt/repeat-mode:repeat-key") :times ,times))))
+
 (define-mode base-mode ()
   "Bind general-purpose commands defined by `define-command'.
 This mode is a good candidate to be passed to `make-buffer'."
@@ -89,6 +96,15 @@ This mode is a good candidate to be passed to `make-buffer'."
        "C-d" 'list-downloads
        "M-x" 'execute-command
        "C-M-x" 'execute-extended-command
+       "M-1" (make-repeat 1)
+       "M-2" (make-repeat 2)
+       "M-3" (make-repeat 3)
+       "M-4" (make-repeat 4)
+       "M-5" (make-repeat 5)
+       "M-6" (make-repeat 6)
+       "M-7" (make-repeat 7)
+       "M-8" (make-repeat 8)
+       "M-9" (make-repeat 9)
        "C-x r j" 'set-url-from-bookmark
        "C-x r M" 'bookmark-current-url
        "C-x r m" 'bookmark-buffer-url
@@ -133,6 +149,15 @@ This mode is a good candidate to be passed to `make-buffer'."
        "C-h k" 'describe-key
        "C-h b" 'describe-bindings
        ":" 'execute-command
+       "1" (make-repeat 1)
+       "2" (make-repeat 2)
+       "3" (make-repeat 3)
+       "4" (make-repeat 4)
+       "5" (make-repeat 5)
+       "6" (make-repeat 6)
+       "7" (make-repeat 7)
+       "8" (make-repeat 8)
+       "9" (make-repeat 9)
        "W" 'make-window
        "C-w C-w" 'make-window
        "C-w q" 'delete-current-window
--- a/source/mode/repeat.lisp
+++ b/source/mode/repeat.lisp
@@ -37,6 +37,7 @@ Function taking a `repeat-mode' instance.")
    (constructor #'initialize)))
 
 (defmethod initialize ((mode repeat-mode))
+  ;; TODO: Cache prompt input now that we have prompt-buffer hooks.
   (unless (repeat-action mode)
     (let ((prompted-action
             (first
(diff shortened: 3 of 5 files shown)


source/input.lisp
```diff
@@ -153,13 +161,14 @@ Return nil to forward to renderer or non-nil otherwise."
              ((or (and buffer (forward-input-events-p buffer))
                   (pointer-event-p (first (last key-stack))))
               (log:debug "Forward key ~s" (keyspecs key-stack))
+              (funcall (input-skip-dispatcher window) (keyspecs key-stack))
```
Ambrevar, line 164: Why call the input-skip-dispatcher here?

source/mode/base.lisp
```diff
@@ -3,6 +3,13 @@
 
 (in-package :nyxt)
 
+(defmacro make-repeat (times)
+  (let ((name (gensym (format nil "repeat-~a-" times))))
+    `(make-command ,name ()
+       ,(format nil "Repeat the command associated to the pressed key ~R times." times)
+       ;; FIXME: A dirty hack.
+       (funcall (read-from-string "nyxt/repeat-mode:repeat-key") :times ,times))))
```
Ambrevar, line 11: Instead of a macro, what about making this a command and read the digit via the `last-event` slot?

source/mode/repeat.lisp
```diff
@@ -37,6 +37,7 @@ Function taking a `repeat-mode' instance.")
    (constructor #'initialize)))
 
 (defmethod initialize ((mode repeat-mode))
+  ;; TODO: Cache prompt input now that we have prompt-buffer hooks.
```
Ambrevar, line 40: Can you expand on what you mean here?

source/mode/repeat.lisp
```diff
@@ -70,4 +71,33 @@ Function taking a `repeat-mode' instance.")
...
+  (setf (command-dispatcher (current-window)) #'dispatch-command
+        (input-skip-dispatcher (current-window)) #'dispatch-input-skip))
+
+;; FIXME: This design does not allow for multi-digit TIMES. Maybe introduce some
+;; global variable, like Emacs does?
+(define-command-global repeat-key
+    (&key (times (ignore-errors
+                  (parse-integer
+                   (prompt1 :prompt "Repeat for X times"
+                     :input "4"
```
Ambrevar, line 98: Why 4?

aartaka:
> Discussion:
> 
> * What don't you like about the `input-skip-dispatcher`?  That it does only something trivial and that we could simplify it?  I think the current approach that you took is simple enough and very flexible, so fine with me.

Yes, I do dislike its triviality. But, I guess, that's still better than what we had.

> * The dismissed message is probably because of `(echo-dismiss)` at the beginning of  `dispatch-input-event`.  Indeed, Nyxt discard the echo area on keypress.  This is by design.  Can you exhibit a recipe which is problematic?

Huh, maybe we should move it into `dispatch-command` then.

The recipe is every non-default command dispatcher, be it `describe-key` or `repeat-key`. When cancelled, they echo something to the message area, but this is immediately dismissed.

source/mode/repeat.lisp
aartaka: Just a reasonable default inspired by Emacs :)

source/mode/repeat.lisp
```diff
@@ -70,4 +71,33 @@ Function taking a `repeat-mode' instance.")
                               :sources (list (make-instance 'prompter:raw-source))))))))
     (when times
       (enable-modes 'repeat-mode (current-buffer)
-                    (list :repeat-count times :repeat-action function)))))
+                    (list :repeat-count times
+                          :repeat-action #'(lambda (mode)
+                                             (declare (ignore mode))
+                                             (nyxt::run-async function)))))))
```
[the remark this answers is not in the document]
aartaka: Yes, that's a good catch! While we still waste some time on computation in between `run-async` calls, the runtime of the command can last for much longer and we might need to wait for it.

source/mode/repeat.lisp
aartaka: I mean remembering prompt inputs (or, even better -- results of prompts as selected objects!) and re-playing those too.

source/input.lisp
aartaka: Something should be called in this case too, and I was too lazy to introduce yet another dispatcher :)

source/mode/repeat.lisp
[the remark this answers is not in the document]
aartaka: Done in 44d75652299253e4da972e8b4d0131be53923340.

source/input.lisp
aartaka: Removed in 5ff2eca22879677e55201957a1c0c94793c79281.

source/mode/repeat.lisp
Ambrevar: OK, then "remember" is clearer than the misleading "cache" in my opinion.

Ambrevar:
Acknowledged, ready to merge then?

source/mode/base.lisp
Ambrevar: Friendly ping! :)

Ambrevar:
(After answering my last remaining question :p)

source/mode/base.lisp
aartaka: Yes, I'm still thinking about it. In this particular case, we need to have a bunch of similar commands with slightly differing documentation and parameter. While the parameter a command can provide, it cannot change the documentation of itself that easily. And we'll have to wrap it into a `make-command` anyway. Having a macro that does:
- Docstring generation.
- `make-command` wrapping.
- Parameter substitution.

For us is a good solution, I believe :)

aartaka:
I'm uncertain about the input-forwarding case having no dispatcher. Having a dispatcher for forwarded events could help is with, for example #2000 (although I'm not yet sure about this).

However, it's trivial to add, and we can do it after merging :)

source/mode/base.lisp
Ambrevar: Maybe a misunderstanding.  What I was suggesting is to have a unique command (with a general purpose docstring) which reads the `last-event` slot to know which number was pressed, and thus guess how many times to repeat.  Does that make sense?

aartaka: Hmmmmmm... Yes it does!

aartaka:
OK, the key lookup is done, now `repeat-key` checks the last presseed key for whether it's a digit. Not fully error-prone (if we bind something else to numbers, it might interpret it wrong), but hella strong and convenient!

Adding to that, now it's not just one-digit commands, there's now an infinite megnitute of repeat times, when you keep Meta pressed and press more digits!

What I don' particularly like about this change is `(read-from-string "nyxt/repeat-mode:repeat-key")` in source/mode/base.lisp, but well, we have to live with it, I guess :)

We can try caching prompt input in this patch too, but let's leave it for later.

aartaka:
Merge-ready!

Issue #2065, opened 2022-01-27
Add button to "go to source" (and install source code)


Ambrevar:
This is easy to implement but hard to distribute, because we need to make sure that the sources are installed along Nyxt, and that Nyxt can find them afterwards in their new location!

Also we can "go to source" for non-Nyxt code if found.

florhizome:
Since i helped tip this off...
I have been packaging a  lot of stuff with Guix the last month and I think from that i would say there is no perfect way, since distributions like Guix will continue to pop up to screw your way. other then going the extra mile in creativity ,try to provide alternatives.  
another thing that I would deem safeto say is, that you want to avoid hardcoding absolute paths yourself during build. 
Another way-to-simple thought from me is that you want to try to keep stuff together while keeping the amount of actual config files low (and to have them separated easily without having to do filtering etc)
I guess it would not be possible to (incrementally) symlink the source-path(s) into the config dir at startup, since that already means that nyxt knows where the source is? 
you can of course give an env var like EMACSLOADPATH (-> NYXTLOADPATH?) as a possible solution

PR #2050, opened 2022-01-24
gtk.lisp: Web context sharing and separation (MASTER)

Ambrevar:
This is a follow up to #1757 and #1987.
It's rebased on `master` and should address all suggestions  from the previous pull requests.

This will hopefully fix the sudden log-out issues and such.

Fixes:
- #1733 
- #1640

Maybe fixes:
- #797
- #837
- #1625 
- #1787

To do after merging:
- #1805 
- #1875 (initialize-modes)

### Things to discuss

 - Can we have multiple non-persisted contexts?
  According to Artyom, yes!
- Should non-web-buffers have a `status` now that we load them like any other buffer?
  Otherwise we must protect all `status' accesses, which is cumbersome.  See 671f4b41c7cdf73b7bc662de109b17374c0112af.
  Also a yes!

- Do you like the `make-buffer-with-context` command name?
  Remove it and instead add a prompt-buffer to `make-buffer` with the `new-query-url` source and a bunch of actions, some of which would set the context, the tag, the class (`nosave`), etc.  Then also remove `make-nosave-buffer`?

PR #2176, opened 2022-04-04
Fix gtk extensions dir + install libnyxt.so + include source

Ambrevar:
3 commits for 3 related fixes:

- This fixes the GTK extension directory (was mistakenly made context-dependent in #2050) and sets it to something system-wide.  Indeed, we have only one such library, there is no point (for now at least) in supporting multiple libs.
- Install the lib together with the source code to a system-wide dir.  This partially addresses #2065.
- Update `nyxt-source-registry` to look into the newly installed source.

Thoughts?

This raises some questions, such as:

- Relying on `asdf-user::*data-dir*` is brittle.  Anything better?
- Should we introduce a new environment variable so that packagers have access to a more explicit configuration knob?

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -18,6 +18,9 @@
 (defvar *bindir* (if (getenv "BINDIR")
                      (ensure-directory-pathname (getenv "BINDIR"))
                      (merge-pathnames* "bin/" *prefix*)))
+(defvar *dest-source-dir* (if (getenv "NYXT_SOURCE_PATH")
+                              (ensure-directory-pathname (getenv "NYXT_SOURCE_PATH"))
+                              (merge-pathnames* "nyxt/" *datadir*)))
 
 (defvar *nyxt-renderer* (or (getenv "NYXT_RENDERER")
                             "gi-gtk"))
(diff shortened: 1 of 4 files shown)


nyxt.asd
```diff
@@ -473,7 +475,28 @@ See `asdf::*immutable-systems*'."
                 (ensure-parent-exists binary-file)
                 (uiop:copy-file (system-relative-pathname c "nyxt") binary-file)
                 ;; TODO: Use iolib/os:file-permissions instead of chmod?  Too verbose?
-                (uiop:run-program (list "chmod" "+x" binary-file))))))
+                (uiop:run-program (list "chmod" "+x" binary-file)))
+              (let ((source-dir (uiop:ensure-directory-pathname (uiop:strcat *datadir* "/nyxt"))))
+                ;; TODO: Don't copy .fasls and .o?  Copy only the output of `git ls-files`, + libnyxt.so?
```
Ambrevar, line 480: Thoughts on this?

Ambrevar:
Tests welcome!

nyxt.asd
aartaka: Yes, git-visible files are likely the ones we need!

source/configuration.lisp
```diff
@@ -56,6 +69,7 @@ This is set globally so that extensions can be loaded even if there is no
 (defun nyxt-source-registry ()
   `(:source-registry
     (:tree ,(nfiles:expand *extensions-directory*))
+    (:tree ,(nfiles:expand *source-directory*))
```
aartaka, line 72: Ha, so it was that easy 0_o

nyxt.asd
Ambrevar: That said, `libnyxt.so` is not listed by Git.

Ambrevar:
Update:

- I've introduced 2 variables, one `*dest-source-dir*` in the .asd, another one `NYXT_SOURCE_PATH` (an environment variable).
- Added `git ls-files` support, falling back to copying the directories on failure.

Tested, seems to work perfectly fine.

Ambrevar:
I've tested by setting

```lisp
(setf (uiop:getenv "DESTDIR") "/tmp/nyxt-install/")
```

before loading the system.

Ambrevar:
OK to merge?

aartaka:
Had no chance to check :cold_sweat:

aartaka:
But well, if you say it works, I'll take your word for it :D

PR #2180, opened 2022-04-07
mode/file-manager.lisp: Fix URL opening on BSD systems.

shamazmazum:
Hi. BSD systems also use `xdg-open` to open something in a preferred application. Tested on FreeBSD.

This fixes file opening and compilation on BSD and MacOS.


aartaka:
Looks good to me! Thanks for this patch :)

jmercouris:
Thank you!

PR #2187, opened 2022-04-11
Progress bar fix

shamazmazum:
These two commits fix the progress bar (you can download any file to see that it's broken).

The first commit fixes a wrong method qualifier introduced recently, but the second one is rather misterious: the progress bar's width was set in px units rather than in percents. I wonder how it worked earlier :) Will be grateful if someone can explain.

### DIFF
--- a/libraries/user-interface/user-interface.lisp
+++ b/libraries/user-interface/user-interface.lisp
@@ -79,7 +79,7 @@ filled up, use a number between 0 and 100.")))
                ;; empty string to force markup to make closing :div tag
                "")))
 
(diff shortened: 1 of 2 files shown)


Ambrevar:
@jmercouris may know whay the px units.

Other than that looks good to me, thanks for the fix!

shamazmazum:
Hi! I hope this won't be forgotten :)

Ambrevar:
Cc @jmercouris

jmercouris:
Just tested, also it wasn't PX, it was percentage before too :-)

shamazmazum:
Thanks!

Issue #2185, opened 2022-04-10
Hint actions without prompt


ilysym:
**Is your feature request related to a problem? Please describe.**
Showing prompt window for every hint action is redundant and distracting in many cases, in particular for simple navigation.

**Describe the solution you'd like**
Add a flag argument to `query-hints` which will tell Nyxt to skip prompt and perform requested action once user have typed hint letters, without additional confirmation. This would allow user to define "quicker" versions of existing hint commands.

**Describe alternatives you've considered**
This proposal probably doesn't require alternatives because browser behaviour won't need to change with default configuration.

**Additional context**
In Vimium, luakit and most of similar software, hints actions are performed without additional prompt. I think this is most reasonable default behaviour for navigation actions.

When navigating many web sites, showing link URL doesn't provide any useful information, since links within the website often just a some constant URL + numeric id of a page. Moreover, a user doesn't typically care about link URL anyway and just wants to follow a link immediately.

On the other hand, being able to preview a link before following it is obviously useful in some cases. So it would be useful to have separate commands for these cases. Similar considerations are applicable for other hint actions (copy URL etc).

aadcg:
Sounds reasonable to me. What Nyxt offers that others don't (AFAIK) is the ability to choose a link by both the hint string and the text associated with the link.

Ambrevar:
Yup, we could totally add a command that shuts up the prompt, it's trivial

Any taker? :)

aadcg:
I'll take it.

aadcg:
@ambrevar, I expected that the following would exit from the prompt when there was a single item left. What am I missing?

```lisp
(prompt
 :auto-return-p t
 :history nil
 :sources
 (make-instance
  'hint-source
  :multi-selection-p t
  :constructor (lambda (source)
                 (declare (ignore source))
                 (add-element-hints :selector (if (current-mode 'web)
                                                  (hints-selector (current-mode 'web))
                                                  "a, button, input, textarea, details, select, img:not([alt=\"\"])")))))
```

aadcg:
The following seems to be true - it's not possible to use the hinting mechanism without relying on the `prompt` interface, since it's not possible to exit from `(make-instance 'hint-source ...)` with a DOM element.

aartaka:
Look at the code introduced in #2054, there is an example of interactive prompt programming. In particular, tests/renderer-online/set-url.lisp can be a good example here.

aadcg:
[...]

PR #2192, opened 2022-04-13
Improve hinting UI

aadcg:
This is my suggestion to solve #2185.  However, it doesn't work!

There are some funny business going on with the `prompt-buffer` and I didn't figure it out yet.

@Ambrevar, I think you could give me a hand here.

### DIFF
--- a/source/mode/scroll.lisp
+++ b/source/mode/scroll.lisp
@@ -36,7 +36,7 @@ The amount scrolled is determined by the buffer's `horizontal-scroll-distance'."
 (define-command scroll-page-down ()
   "Scroll down by one page height."
   (peval (ps:chain window (scroll-by 0 (* (ps:lisp (page-scroll-ratio (current-buffer)))
-                                              (ps:@ window inner-height))))))
+                                          (ps:@ window inner-height))))))
 
 (define-command scroll-page-up ()
   "Scroll up by one page height."


aadcg:
```lisp
(prompt :prompt "Test prompt"
        :auto-return-p t
        :sources (list (make-instance 'prompter:source :name "Test"
                                                       :constructor '("foo" "bar"))))
```

The above behaves as expected, i.e. after narrowing down to either "foo" or "bar", the prompt vanishes. The same isn't true when it comes to the hinting mechanism.

source/mode/element-hint.lisp
```diff
@@ -153,14 +153,17 @@
...
                          annotate-visible-only-p
+                         (auto-return-p t)
                          (selector
                              (if (current-mode 'web)
                                  (hints-selector (current-mode 'web))
                                  "a, button, input, textarea, details, select, img:not([alt=\"\"])")))
   "Prompt to choose several elements out of those matching SELECTOR, hinting them visually.
-MULTI-SELECTION-P is whether several elements can be chosen.
+MULTI-SELECTION-P defines whether several elements can be chosen.
+AUTO-RETURN-P defines whether the prompt exists automatically after narrowing
```
Ambrevar, line 165: Typo: "exits"

Ambrevar:
Works for me!

I tried on the `help` page then wrote `common` (for  common settings) and it got me there without pressing enter.

Can you produce a recipe?

source/mode/element-hint.lisp
```diff
@@ -153,14 +153,17 @@
 
 (serapeum:export-always 'query-hints)
 (defun query-hints (prompt function
-                    &key multi-selection-p
+                    &key (multi-selection-p t)
```
Ambrevar, line 156: If you enable `multi-selection` by default, it means we must pass `:multi-selection nil` where it does not make sense (e.g. `follow-hint`).  

That said, is it worth the change?

aartaka: `follow-hint` is multi-selection-enabled :)

aadcg: Right @aartaka. I made this change since multi-selection seemed to be the rule, rather than the exception.

aadcg:
> Works for me!
> 
> I tried on the `help` page then wrote `common` (for common settings) and it got me there without pressing enter.
> 
> Can you produce a recipe?

Did you start Nyxt without any config? I suspect that you have set `auto-follow-hints-p` to `t`.

source/mode/element-hint.lisp
aadcg: Good catch!

aartaka:
See this [snippet from prompter.lisp](https://github.com/atlas-engineer/nyxt/blob/7f50560c2c17284087116e042dd14666167dc161/libraries/prompter/prompter.lisp#L175): `auto-return-p` only works when there's only one suggestion left.

``` lisp
(when (and (auto-return-p prompter)
             (sera:single (all-suggestions prompter)))
    (return-selection prompter))
```

source/mode/element-hint.lisp
Ambrevar: Oops, looks like I read it wrong, my bad.

Ambrevar:
@aadcg No I started without any init file.

@aartaka This is correct, isn't it?

aartaka:
Yes, correct behavior it is! We just need to adjust the filters for the sources we want to make `auto-return-p`-able!

aartaka:
(Actually, if we user `auto-return-p` and reasonable custom filters, then we don't need `auto-follow-hints-p`...)

Ambrevar:
> We just need to adjust the filters for the sources we want to make `auto-return-p`-able!

What do you mean?  Are the filters problematic for link hinting?

aadcg:
I agree that `auto-follow-hints-p` (and thus `auto-follow-timer`) should be deprecated. 

However, I'm still confused about @Ambrevar's claim that this is working for him.

aadcg:
I think I understood my confusion. There's a `prompt` and a `prompter`. Let me try to fix this.

aadcg:
Ok, I understood why I said it doesn't work, while @Ambrevar says it does.  We're both right.

Let's look at simple things, shall we? Here's the bug I found:

```lisp
(prompt :prompt "Test prompt"
        :auto-return-p t
        :sources (list (make-instance 'prompter:source :name "Test"
                                                       :constructor '("fooooo" "barrrrr"))))
```

When prompted, typing "fo" suffices to narrow down to one suggestion. But the prompt doesn't exit at that time! You need to type an extra character. In fact, you can type any kind of bogus character that bears no connections with "fooooo", say "foj" or "fom".

This is especially annoying when it comes to narrowing down the hinting mechanism. Say I want to go to hint "PT", then I need to type "PT.", where "." is any character.

Ambrevar:
Excellent investigation!  Indeed, the auto-return-p logic was run in
parallel to the source update, which is wrong.  I've moved it to run
after the sources are done updating, see
94be3b8cdf726df0776d508e5803064722601d91.

Does it work for you now?

aadcg:
Thank you for looking into it @Ambrevar. 

There's something that worries me, tho. Commit 94be3b8c doesn't seem to descend from its parent. The diff looks fishy. What's going on here?

Ambrevar:
It's fine here, how are you looking at it?  `git`, magit, github?

aadcg:
It's indeed ok. My bad, sorry! I was trying to follow your solution.

aartaka:
> @aartaka, do we still need `prompter:follow-mode-functions` in `hint-source`?

Yes we do! One may wish to disable `auto-return-p` and have hints nicely highlighted when chosen!

Actually, I'd like to have this behavior be configurable. How about we resurrect `nyxt/web-mode:auto-follow-hints-p` and use it as the default value of `auto-return-p` in `query-hints`? This way ones wishing to have it instantaneous will have it, and ones (like me) wanting to utilize multiple selection and fuzzy matching will enjoy it setting the slot to nil.

source/mode/element-hint.lisp
```diff
@@ -125,13 +125,11 @@
     (lambda (suggestion)
       (highlight-selected-hint :element suggestion
                                :scroll nil)
-      (sera:and-let* ((auto-follow (nyxt/web-mode:auto-follow-hints-p (current-mode 'web)))
-                      (matches (string-equal
+      (sera:and-let* ((matches (string-equal
```
aartaka, line 128: But wait, is this form even necessary now that we use `auto-return-p`? It should rather be abstracted into a separate prompt filter.

Ideas for the filter:
- If hint letters (e.g. "dx" for "DX" hint) are inputted, make the hint with these letters the sole one, triggering `auto-follow-p`.
- If there's more than three letters, fall back to `prompter:delete-inexact-matches`, or even regular fuzzy matching.
- Make it a post-process filter, so that user has some time to quickly type in more than two letters.

source/mode/element-hint.lisp
```diff
@@ -153,38 +151,36 @@
 
 (serapeum:export-always 'query-hints)
 (defun query-hints (prompt function
-                    &key multi-selection-p
+                    &key (multi-selection-p t)
                          annotate-visible-only-p
+                         (auto-return-p t)
```
aartaka, line 156: Should it be `t` by default? Don't underestimate the power of defaults...

I'd say we should allow the user to explore the power of fuzzy matching by default, while allowing to configure the `auto-return-p` easily enough for those that want it.

source/mode/element-hint.lisp
```diff
@@ -346,70 +342,58 @@ FUNCTION is the action to perform on the selected elements."
...
                    (%follow-hint (first results))
                    (mapcar (alex:rcurry #'%follow-hint-new-buffer buffer)
-                           (rest results)))
-                 :multi-selection-p t)))
+                           (rest results))))))
 
 (define-command follow-hint-new-buffer ()
-  "Show a set of element hints, and open the user inputted one in a new
-buffer (not set to visible active buffer)."
+  "Like `follow-hint', but open them in new buffers (no focus)."
```
aartaka, line 354: "them" should rather be "selected hints" so that it doesn't cofuse the reader what "them" is :)

source/mode/element-hint.lisp
```diff
@@ -346,70 +342,58 @@ FUNCTION is the action to perform on the selected elements."
...
     (query-hints "Open element in new buffer"
                  (lambda (result) (mapcar (alex:rcurry #'%follow-hint-new-buffer buffer)
-                                          result))
-                 :multi-selection-p t)))
+                                          result)))))
 
 (define-command follow-hint-new-buffer-focus ()
-  "Show a set of element hints, and open the user inputted one in a new
-visible active buffer."
+  "Like `follow-hint-new-buffer', but with focus."
```
aartaka, line 361: "but with focus" needs an expansion -- it's not clear that it opens a _new_ buffer that's _immediately focused_.

source/mode/element-hint.lisp
```diff
@@ -346,70 +342,58 @@ FUNCTION is the action to perform on the selected elements."
...
                  (lambda (result)
                    (mapcar (alex:rcurry #'%follow-hint-with-current-modes-new-buffer buffer)
-                           result))
-                 :multi-selection-p t)))
+                           result)))))
 
 (define-command copy-hint-url ()
-  "Show a set of element hints, and copy the URL of the user inputted one."
+  "Prompt user for element hints and save to clipboard."
   (query-hints "Copy element URL" (lambda (result)  (%copy-hint-url (first result)))
```
aartaka, line 392: Should be `:multi-selection-p nil` here -- we cannot copy ten hints into clipboard all at once.

source/prompt-buffer.lisp
```diff
@@ -480,7 +480,8 @@ Example use:
 
 \(prompt
   :prompt \"Test prompt\"
-  :sources (list (make-instance 'prompter:source :filter #'my-suggestion-filter)))
```
aartaka, line 483: Why this change? If you want to add an example of simple prompt, then it's better as a separate example, instead of replacing the existing one. Two different examples are better than one :)

aartaka (changes requested):

source/mode/element-hint.lisp
aadcg: Honestly, I think the defaults of the hinting mechanism are too advanced and confusing. Here's how I see it: 

- no `prompt-bufer` at all, since it hides a significant part of the hints. @Ambrevar, correct me if I'm wrong, but this seemed to be far from trivial.
- no `multi-selection-p` disabled by default. It's the most general case, but not the most frequent.
- `auto-return-p` enabled by default. 

To make it crystal clear, from an UI/UX perspective: the user triggers one of `follow-hint`, `follow-hint-new-buffer` or `follow-hint-new-buffer-focus`; the buffer gets populated by the hints (`prompt-buffer` doesn`t get in the way); the user sends the input events corresponding to a single hint (usually two characters of the latin alphabet).

source/mode/element-hint.lisp
aadcg: Fair! See fde820fe.

source/mode/element-hint.lisp
aadcg: Are you suggesting that it was formerly, i.e. before the PR, wrong?

source/prompt-buffer.lisp
aadcg: First of all, the example was broken since it lacked `:name`. 

Second, an example should strive to be simple and down-to-earth. The main purpose of `prompt` is to suggest objects for selection (hence I added the `constructor`).

source/mode/element-hint.lisp
aadcg: I disagree. The reference to the new buffer comes from `follow-hint-new-buffer`. With respect to focus, your suggestion wouldn't be exhaustive since the user can select multiple hints.

The first sentence of docstrings should be succinct, even if it sacrifices some details. Those can come below.

source/mode/element-hint.lisp
Ambrevar: - The prompt-buffer is a killer feature of link hinting (that's unique to Nyxt), I say let's boast it!
- `auto-return-p` is confusing by default, I would never make it the default, only expose it as an opt-in.  Rationale: let's make Nyxt more familiar and accessible.

The UX you are proposing echoes that of other tech-user browsers like Qutebrowser.  It's fine to offer this UX, but I recommend we stick with a more visually telling interface.   Tech-savvy users can always configure Nyxt if they want to. Same argument why we have CUA bindings by default.

source/mode/element-hint.lisp
```diff
@@ -153,38 +151,36 @@
...
+                :history nil
+                :sources
+                (make-instance 'hint-source
+                               :multi-selection-p multi-selection-p
+                               :constructor (lambda (source) (declare (ignore source))
+                                              (add-element-hints :selector selector))))))
+    (funcall function result))
+  ;; remove-element-hints should go into (prompt :after-destructor) but it
+  ;; misbehaves
+  (remove-element-hints))
```
Ambrevar, line 183: Why remove `with-current-buffer`?  This was added because without it, `remove-element-hints` may work on the wrong buffer (link hints may have focused to  new buffer in the mean time).

source/mode/element-hint.lisp
```diff
@@ -346,70 +342,59 @@ FUNCTION is the action to perform on the selected elements."
           (remove-focus)))))
 
 (define-command follow-hint ()
-  "Show a set of element hints, and go to the user inputted one in the current
-buffer.
-
-Auto-follows hints by their ID, if `web-mode's `auto-follow-hints-p' is true."
+  "Prompt user for element hints and open them in the current buffer."
```
Ambrevar, line 345: You can even remove `user`.

source/mode/element-hint.lisp
```diff
@@ -346,70 +342,59 @@ FUNCTION is the action to perform on the selected elements."
...
                    (%follow-hint-new-buffer-focus (first result) buffer)
                    (mapcar (alex:rcurry #'%follow-hint-new-buffer buffer)
-                           (rest result)))
-                 :multi-selection-p t)))
+                           (rest result))))))
 
 (define-command follow-hint-nosave-buffer ()
-  "Show a set of element hints, and open the user inputted one in a new
-nosave buffer (not set to visible active buffer)."
+  "Like `follow-hint', but open the selected hints in new nosave buffers (no
```
Ambrevar, line 370: Put `nosave` between Lisp quotes so that it enables cross-referencing.

source/mode/element-hint.lisp
```diff
@@ -346,70 +342,59 @@ FUNCTION is the action to perform on the selected elements."
...
   (query-hints "Go to element in new buffer"
                (lambda (result)
                  (%follow-hint-nosave-buffer-focus (first result))
-                 (mapcar #'%follow-hint-nosave-buffer (rest result)))
-               :multi-selection-p t))
+                 (mapcar #'%follow-hint-nosave-buffer (rest result)))))
 
 (define-command follow-hint-with-current-modes-new-buffer ()
-  "Follow hint and open link in a new buffer with current modes."
+  "Prompt user for element hints and open them in a new buffer with current
```
Ambrevar, line 383: Same, remove `user`.

source/mode/element-hint.lisp
```diff
@@ -346,70 +342,59 @@ FUNCTION is the action to perform on the selected elements."
...
     (query-hints "Open element with current modes in new buffer"
                  (lambda (result)
                    (mapcar (alex:rcurry #'%follow-hint-with-current-modes-new-buffer buffer)
-                           result))
-                 :multi-selection-p t)))
+                           result)))))
 
 (define-command copy-hint-url ()
-  "Show a set of element hints, and copy the URL of the user inputted one."
+  "Prompt user for element hints and save to clipboard."
```
Ambrevar, line 392: Same, remove `user`.

source/mode/element-hint.lisp
```diff
@@ -346,70 +342,59 @@ FUNCTION is the action to perform on the selected elements."
...
   (query-hints "Copy element URL" (lambda (result)  (%copy-hint-url (first result)))
                :selector "a, img"))
 
 (define-command bookmark-hint ()
-  "Show link hints on screen, and allow the user to bookmark one."
+  "Prompt user for element hints and bookmark them."
```
Ambrevar, line 397: Same.

source/mode/element-hint.lisp
```diff
@@ -422,19 +407,17 @@ visible nosave active buffer."
                                           (make-instance 'nyxt::tag-source
                                                          :marks (url-bookmark-tags url))))))
                      (bookmark-add url :tags tags :title (fetch-url-title url)))))
-               :multi-selection-p t
                :selector "a, img"))
 
 (define-command download-hint-url ()
-  "Download the file under the URL(s) hinted by the user."
+  "Prompt user for element hints and download them."
```
Ambrevar, line 413: Same.

source/mode/element-hint.lisp
Ambrevar: Maybe clarify which buffer gets focused.

source/mode/element-hint.lisp
```diff
@@ -125,13 +125,11 @@
...
       (highlight-selected-hint :element suggestion
                                :scroll nil)
-      (sera:and-let* ((auto-follow (nyxt/web-mode:auto-follow-hints-p (current-mode 'web)))
-                      (matches (string-equal
+      (sera:and-let* ((matches (string-equal
                                 (prompter:input (current-prompt-buffer))
                                 (plump:get-attribute  suggestion "nyxt-hint")))
                       (input (prompter:input (current-prompt-buffer))))
         (run-thread "hint auto-follow thread"
-          (sleep (nyxt/web-mode:auto-follow-timer (current-mode 'web)))
```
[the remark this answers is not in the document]
aadcg: The slot was deprecated, but @aartaka mentioned it might make sense to bring it back.

source/mode/element-hint.lisp
aadcg: Ok, I'll revert it. There was a `let` nested into `let*`, does that make sense?

source/mode/element-hint.lisp
aadcg: See 410a6520.

source/mode/element-hint.lisp
aadcg: Done, see 0154ad47.

source/mode/element-hint.lisp
aadcg: See 3094d770.

aadcg:
> - The prompt-buffer is a killer feature of link hinting (that's unique to Nyxt), I say let's boast it!
> - `auto-return-p` is confusing by default, I would never make it the default, only expose it as an opt-in.  Rationale: let's make Nyxt more familiar and accessible.

It's certainly unique. But I don't see how it makes selecting one single (not _multiple_) hint easier. 
I don't follow how `auto-return-p` defaulting to `nil` makes it more familiar, given what you mentioned below.

> The UX you are proposing echoes that of other tech-user browsers like Qutebrowser.  It's fine to offer this UX, but I recommend we stick with a more visually telling interface.   Tech-savvy users can always configure Nyxt if they want to. Same argument why we have CUA bindings by default.

We have CUA because it's what most people expect. Shouldn't that apply to the hinting mechanism?

I'd be thankful if @Ambrevar, @aartaka, @jmercouris or @pdelfino would me how they use the hinting system with screenshots or a short videos/gifs.

Ambrevar:
`auto-return-p` is not used in any prompt-buffer by default, so having
it in just this one would be inconsistent behaviour.

Link-hinting as a concept is not something familiar for people coming
from popular, non-power browsers.  This is where the prompt-buffer prove
useful: to guide the users and making it obvious that they can complete
against _text_ as well.

All that said, I get your point of having a command that behaves like
link-hinting in Conkeror and friends.  Then what about having a separate
command?

source/mode/element-hint.lisp
aartaka: Before the PR it used the default of `:multi-selection-p nil`. You've changed the default to `t`. You can look at other commands utilizing `:multiple-selection-p` -- you've removed the `:multiple-selection-p t` clause from those, because it's default now. However, this function was left intact -- yet it, as the one relying on the old default, should've changed its `:multi-selection-p` to `nil`.

source/mode/element-hint.lisp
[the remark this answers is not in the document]
aartaka: I meant bringing the `auto-follow-hints-p` back, as it could make a good toggle and a sane default fom `auto-return-p`. Sleep timer would be unnecessary in this scenario, as everything would be done by `prompter`.

Ambrevar:
To come back to the original intent of the pull request / issue: the main problem is that we don't see the links under the prompt buffer.  To mitigate this, two provisions that we should implement regardless of this pull request:

- Add a prompt buffer binding to toggle enlarge/shrink prompt buffer height (e.g. toggle between 1 suggestion displayed and the default).
- Add prompt buffer bindings to scroll the buffer, so that we can display the part of the buffer that's hidden.  This should work also we are at the bottom of the page.

jmercouris:
It seems I programmed them AND put them in the source at some time 🙃

aartaka:
> It seems I programmed them AND put them in the source at some time 🙃

`git blame` says it was me in d5487cb94bb28e14a0351fc168f6b8bb8cd1e409 :P

jmercouris:
I swear I wrote those commands for Kabouik!!

source/mode/element-hint.lisp
aadcg: You're right. I misinterpreted the diff.

source/mode/element-hint.lisp
aadcg: I still don't get your suggestion regarding filters.

Please check the latest state of this PR and tell me if your concerns persist.

source/mode/element-hint.lisp
aadcg: I'm not convinced, but I will abide to the current behavior.

aadcg:
> `auto-return-p` is not used in any prompt-buffer by default, so having it in just this one would be inconsistent behaviour. 

Fair, we agree.

> Link-hinting as a concept is not something familiar for people coming from popular, non-power browsers. This is where the prompt-buffer prove useful: to guide the users and making it obvious that they can complete against _text_ as well. 

The elephant in the room is that prompt-buffer is a vertical completion system that defeats the purpose of link-hinting. Scrolling (manually or automatically) or re-centering doesn't fix it. 

I've asked above, and I reiterate: is it possible to have a prompt that doesn't eagerly hide part of a buffer? How hard would it be to support both a line-based completion UX (think `icomplete`) and vertical-based? 

> All that said, I get your point of having a command that behaves like link-hinting in Conkeror and friends. Then what about having a separate command?

I don't see the need. I think it's enough to tweak some class slots.

aadcg:
> Using the no-prompt hinting without multiple selection would be totally fine for VI people, but can be totally unpredictable to everyone else.

I disagree. First of all, it's a fallacy to think that there are VI-like and Emacs-like interfaces. At most, there are slightly different approaches to keybindings in VI and Emacs (no need to expand here, you get it), though they actually agree on their goals (both are modal editors, but achieve it differently).

Think of `avy`, an Emacs package. The UX isn't either "Emacsy" or "VImsy".

aadcg:
@aartaka I appreciate the detailed way you've described your hinting workflow. I requested since I was a bit skeptical someone was actually using the interface as you've described. 

If it works for you, great! The interface is certainly complete and, to some extend, user-friendly. Still, I find that it gets in the way.

aartaka:
> > Using the no-prompt hinting without multiple selection would be totally fine for VI people, but can be totally unpredictable to everyone else.
> 
> I disagree. First of all, it's a fallacy to think that there are VI-like and Emacs-like interfaces. At most, there are slightly different approaches to keybindings in VI and Emacs (no need to expand here, you get it), though they actually agree on their goals (both are modal editors, but achieve it differently).

In the core, yes, both VI and Emacs are modal. However, there's not only [`avy`](https://github.com/abo-abo/avy) in Emacs world, there's also [helm](https://github.com/emacs-helm/helm), which is an incremental narrowing framework with a lot of text matching and verticality. It was _the_ inspiration for our `prompter`, and is a good example of what I mean by non-VI (and non-vanilla-Emacs way, if you want me to spell it out :P) way of doing things -- it's not only modal, it's textual (not in the sense of TUI, but rather in the sense of textual media) also -- and this is crucial for such a semantically rich domain as web pages, with all the text they contain. So
- Using text _and_ hints is a good universal interface. Yes, it might get in the way. But if it's configurable -- just configure it not to (which is why I suggest resurrecting `auto-follow-hints-p`)!
- Using just hints -- a good setting for those that want it and that are not necessary concerned/used to the textual side of web.

aartaka:
> The elephant in the room is that prompt-buffer is a vertical completion system that defeats the purpose of link-hinting. Scrolling (manually or automatically) or re-centering doesn't fix it.
> 
> I've asked above, and I reiterate: is it possible to have a prompt that doesn't eagerly hide part of a buffer? How hard would it be to support both a line-based completion UX (think `icomplete`) and vertical-based?

Yes. Just do `(ffi-window-set-prompt-buffer-height (nyxt:ffi-buffer-evaluate-javascript (nyxt:current-prompt-buffer) (ps:ps (ps:chain (nyxt/ps:qs "#prompt") offset-height))))` after prompt rendering.

aadcg:
@aartaka thank you for the reply and the code snippet that sets the prompt buffer height. We all agree on the topics that really matter to the PR at hand! I brought to the table some topics that clearly don't fit the moment, sorry.

I believe we're very close to merge, so let's go through what might be missing or unclear:

- I deleted the deprecated `ANNOTATE-VISIBLE-ONLY-P` since there's a new release coming soon. Is this the correct policy?
- What to do about commit 6b722af0?

Thank you!

aadcg:
Oops... I pushed to the wrong branch! What should I do now? Thanks.

jmercouris:
Well, definitely not a force push :-D

I suggest you make a commit that reverts all the changes and submit a new pull request

jmercouris:
Never ever force push onto master!

PR #2194, opened 2022-04-14
From define-user-class to user-class metaclass

### LINKED ISSUE #1774: Possible bug: changing the default value of slot named `hints-alphabet` via "Configure" button (GUI) does not work
**Describe the bug**
Using the Nyxt's GUI to change the default value of `hints-alphabet` does change the `auto-config.lisp` file. However, it does not generate the expected behavior. Moreover, it blocks Nyxt from starting.

- start Nyxt
- execute command `describe-slot`;
- type `link-alphabet`;
- select the second option, from `user-browser` class;
- click on the `Configure` button;
- insert the string (with quotes): `"asfdghjkl"`
- see that the local file `auto-config.lisp` has changed with the following expression included:
```lisp
(DEFINE-CONFIGURATION NYXT/WEB-MODE:USER-WEB-MODE
  ((NYXT/WEB-MODE:HINTS-ALPHABET "asfdghjkl")))
```
- look at the REPL, it shows this message:
```lisp
end of file on #<SB-IMPL::STRING-INPUT-STREAM {100A66A473}>
   [Condition of type END-OF-FILE]
```
- close Nyxt
- quit Slime/Sly;
- start Slime
- load the system nyxt/gi-gtk;
- try starting Nyxt
-  an error occurs with the following message:
```lisp
define-configuration argument USER-USER-WEB-MODE is not a known class.
```

Observation: if I remove the new addition on `auto-config.lisp` restoring how it used to be and if I simply insert the same expression in my **init.lisp** it **does** work:

```lisp
(DEFINE-CONFIGURATION NYXT/WEB-MODE:USER-WEB-MODE
  ((NYXT/WEB-MODE:HINTS-ALPHABET "asfdghjkl")))
  ```

For website-specific issues:
Can you reproduce this issue with Epiphany / GNOME Web (https://wiki.gnome.org/Apps/Web)?
Not relevant.

**Information**
- OS name+version: 
Ubuntu 20.04 LTS

- Graphics card and driver:
product: Iris Plus Graphics 640
configuration: driver=i915 latency=0

- Desktop environment / Window manager name+version:
Name: GNOME Shell 3.36.7
Window manager's "showing the desktop" mode: OFF

- How you installed Nyxt (Guix pack, package manager, build from source):
Source

- Information from command `show-system-information`:

Since Nyxt is not starting, I cannot do it.

  If you can't run `show-system-information`, try `nyxt --system-information` from
  a shell.  It this still does not work, please provide the following:
 
  - Nyxt version (from the `nyxt-version` command or `nyxt --version` from a shell): 
  
 ```lisp
  NYXT> (nyxt-version)
 <INFO> [18:27:54] nyxt message.lisp (echo form-fun-6) -
  Version 2.1.1-607-g0bafa5cc
NIL 
 ```
  
 - Lisp implementation/version (if built from source):  This is SBCL 2.0.1.debian

  - Kernel name+version:  
  ```bash
$ cat /proc/version
Linux version 5.11.0-27-generic (buildd@lcy01-amd64-019) (gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, GNU ld (GNU Binutils for Ubuntu) 2.34) #29~20.04.1-Ubuntu SMP Wed Aug 11 15:58:17 UTC 2021
```

  - WebKitGTK+ | QtWebEngine version: 
  ```bash
  $ dpkg -s libgtk-3-0|grep '^Version'
Version: 3.24.20-0ubuntu1
```
It was run from source in the Slime's REPL.

### PR DESCRIPTION
**Pull request by Ambrevar**
This addresses most points in #1431.

- Functional config.
- No more `user-` prefix.
- Clear identification of a user-class, since it's an explicit metaclass.
- 100+ lines of dirty macros are gone.

Other perks:

- The customizability of a class is no longer inherited, you now specify it manually for each class.  This gives more control.

- Once a class is set to be a `user-class`, there is no going back since one cannot change the metaclass.  I haven't tried `change-class` on an instance, maybe it could work with a bit of hacking.

**To do:**

- [x] For now it compiles but does not start.  I have to investigate.
  Update: Fixed.
- [x] Documentation must be updated.
- [x] There are some code snippets that still refer to `%slot-default&`.
- *(In another pull request)* For defclass-star, merge `class*` into it.  This would introduce a lot of noise, hence why I decided to do it in a second step.

**To test:**
- [x] `append-configuration`
- [x] `password` interface configuration

**Questions:**

- In commit 7824556e48984b34732ce875aab8896b3955b87e, it seems that the logic was inverted and we didn't allow configuration for user classes.  Was it a mistake?  @jmercouris 
- In commit a77dd74f49f3d58bc888bfd60fadd4c62b37e67b I've realized that we export too much, it seems to me that the GTK slots are mostly implementation details.  While we are at it, should they be user classes then?  I would say no since most slots don't need to be configured.  TODO: Unexport more accessors if I'm right.
- ~~Look at the new `pushnew-direct-superclass` macro, it essentially sets the renderer-specific classes (e.g. `buffer` is now superclasses by `gtk-buffer`).  It does what we want, but it's not functional.  (It wasn't functional before either.) Any better way?  Mixins (https://github.com/rpav/dynamic-mixins) are a tempting option here, I just can't figure out how to fit them into the picture.  Indeed, it would be ideal if we didn't have to declare a class as a mixin at definition site, because we might not know in advance.  See the `scheme` class for example.~~  UPDATE: No longer relevant now that we have proper mixins.

Comments, thoughts, criticism?  @aadcg @jmercouris @pdelfino @aartaka 

Possibly fixes #1774 #1156 #1767.

### CODE REVIEW

**Comment by Ambrevar** [comment]
I know why it's not starting: the new hierarchy is so that `gtk-buffer` superclasses `buffer`, thus the renderer-less FFI methods are called...

The design is wrong.  We could re-introduce the mixins (`user-buffer` or a new name) like we have on master, but I'd like to find a way to avoid them so that the renderer can choose what to extend, without have the Core specify it in advance.

Tips welcome!

**Comment by Ambrevar** [comment]
OK, I think I've got it: we can create a method that invoked in our special `make-instance` for user classes, which will dispatch over which class to actually instantiate depending on the renderer.

For instance `find-renderer-class` returns `gtk-window` for the GTK renderer, and thus `make-instance 'window` would make a `gtk-window`.

**Comment by Ambrevar** [comment]
Nope, that won't work, because then `web-buffer` would still make inherit from `buffer` and the `gtk` parts would be missing...

To sum up, what we are trying to do is to "insert" the GTK component (`gtk-buffer`) between `buffer` and all of its subclasses.  The only reasonable way to do this is via a mixin, but since we are trying to get rid of explicit references to mixins at call sites...

(Could also do this via some upside down MOP hacking, but it's still unclear how and it's not going to be pretty.)

This is all too complicated.  High time we redesigned and simplified our API.  Something more CLOS-streamlined.

**Major rewrite proposal**

- Introduce a new `renderer` class, with `fallback-renderer` (pure JS), `gtk-renderer`, `gi-gtk-renderer` (inherits from `gtk-renderer`), etc.  It would essentially replace `gtk-browser`.
- Have all FFI methods specialize against the renderer.
- Rename `gtk-buffer` to `gtk-buffer-view`, have it inherit either nothing or a new `buffer-view` class.
- Add a `buffer-view` slot to `buffer` (we already have a `prompt-buffer-view` slot) and set it to `gtk-buffer-view` when using the GTK renderer.
- Rename `gtk-window` to `gtk-window-widget`, have it inherit either nothing or a new `window-widget` class.
- Add a `window-widget` slot to `window and set it to `gtk-window-widget` when using the GTK renderer.

Alternate name suggestions are welcome!

Thoughts? @aartaka @jmercouris @pdelfino @aadcg

**Comment by Ambrevar** [comment]
### Minor rewrite proposal (mixins) 

- Stick to mixins, but rename them as initially suggested in #1431:
  - `buffer` becomes `buffer-core`, then we define `buffer` to be a mixin class initially with only `buffer-core`
    -  In `gtk.lisp`, we set the mixin to be `(gtk-buffer buffer-core)`.
  - Repeat for `window`, `browser`, `download`, etc.

**Comment by aartaka** [comment]
I'll respond in a more detailed way later. For now, what about doing `(change-class buffer 'gtk-buffer)` in `make-instance :around` method?

**Comment by Ambrevar** [comment]
It won't work for the reason mentioned above:

> because then web-buffer would still make inherit from buffer and the gtk parts would be missing...

To rephrase, `(make-buffer 'web-buffer)` would instantiate a `web-buffer` which inherits from `buffer`.  Then `gtk-buffer` is missing from the graph!

**Comment by aartaka** [comment]
Okay, as a continuation of the `change-class` madness spiralling, how about defining an entirely empty `buffer-core`/`renderer-buffer` class for `buffer` to inherit from and redefining it to have GTK-related slots in renderer/gtk.lisp? This way GTK slots will propagate to all the subclasses of `buffer-core`, including `buffer` with its subclasses.

**Comment by Ambrevar** [comment]
I think your suggestion is very similar to my "minor rewrite proposal", but I see one drawback here: to switch renderer one has to redefine the `buffer-core` class which is not practical.  With my approach it's a one-liner:

```lisp
(defclass buffer (gtk-buffer buffer-core))
```

(As on master.)

aartaka:
Yes, the minor rewrite proposal I'm definitely on the side of!

aartaka:
Just to clarify, here's an annotated tree of inheritance for the minor rewrite proposal:
- `buffer-core`: has all the former `buffer` slots.
  - `buffer`: has no slots at all and relies on mix-ins. To understand what buffer has/does by looking at the source, one has to find the `buffer-core` source.
    - `web-buffer`: has additional slots on top of what `buffer` (`buffer-core` actually) provides.

With this approach, `buffer` is confusingly empty if you look at its source in the built-in inspector. With the redefinable `renderer-buffer` at the root of the hierarchy, the renderer details are abstracted away, which is a downside, yet the `buffer` is transparent and understandable. I.e.
- `renderer-buffer` has no slots until redefined in gtk.lisp. After this, it is introspectable.
  - `buffer` inherits all the slots of `renderer-buffer` and adds the sensible ones to them. It's perfectly visible in the source.
    - `web-buffer` inherits from `buffer` and has all its slots propagated.

The difference is obviously minor, but by redefining the most fundamental class (`renderer-buffer`/`buffer-core`) we make sources of the classes relying on it clearer.

And yes, a good inspector (even the one that we have already) should mitigate this problem, but... Lots of people on e.g. IRC rely on the sources, not inspector, for Nyxt hacking.

Ambrevar:
The second approach is what I originally tried in
cdfc2264a8b3d7249f1f8b88b7721279840d3cf5.  It does not work because then
the ffi- fallback methods are called instead of the renderer specialized
methods.

Ambrevar:
The "minor rewrite" is now pushed in ec910933c6eb3fe4b09b61c5edadd615cacfcc2d and Nyxt starts!  Woohoo!

aartaka:
> The second approach is what I originally tried in [cdfc226](https://github.com/atlas-engineer/nyxt/commit/cdfc2264a8b3d7249f1f8b88b7721279840d3cf5). It does not work because then the ffi- fallback methods are called instead of the renderer specialized methods.

If we specialize the fallback methods against `t`, not `buffer`, it'll work just fine -- `t` is a superclass of anything, after all :)

aartaka:
Minor rewrite is definitely a lesser evil there, but the mixing metaclasses look weird. Maybe we can still resort to my suggestion that will look even simpler than how it looks now? :)

Ambrevar:
Good point!  But then the aforementioned issue remains:

> I think your suggestion is very similar to my "minor rewrite proposal", but I see one drawback here: to switch renderer one has to redefine the `buffer-core` class which is not practical. With my approach it's a one-liner:
> 
> ```lisp
> (defclass buffer (gtk-buffer buffer-core))
> ```
> 
> (As on master.)

Suggestion: keep `buffer` as on `master` but have it inherit from `renderer-buffer`, a mixin class.
Then set the mixin to inherit from `gtk-buffer`.

Thoughts?

aartaka:
But what's the problem with redefining the `renderer-core`? I mean, we're not likely to switch renderer at runtime anyway, and it plays no role before the runtime, right?

aartaka:
`renderer-buffer`, I mean :sweat:

Ambrevar:
It's a nit since in the current state of things we are only supporting 1 renderer, but in the future we might have Qt (or CFE), in which case one might want to change renderer at runtime indeed (e.g. from the REPL, close all windows, set the new renderer, then restart).

Am I overengineering?

aartaka:
Does your minor refactoring approach allow that?

Ambrevar:
Indeed, see:

```lisp
(defclass browser (gtk-browser) ()
  (:metaclass mixin-class))
```

Ambrevar:
@aartaka Done with 600ef7cdd7727ed5fadc4115d3a66b4621cd7b81.

Ambrevar:
It's mostly ready to merge, the only thing left is the `append-configuration` and `configure-slots` bits which need a big overhaul.

I'll do it later.

Ambrevar:
In the meantime, tests and feedback welcome of course :)

aartaka:
> It's mostly ready to merge, the only thing left is the `append-configuration` and `configure-slots` bits which need a big overhaul.

A high time we redesign and enrich our (common) settings page :D

source/user-classes.lisp
```diff
@@ -0,0 +1,61 @@
...
+behaviour of some CLASS instance.
+
+This method is run after the instance has been initialized (in particular, after
+the `initialize-instance' :after method).
+
+This is meant to be used by users only.  Don't use it in public code, prefer
+`initialize-instance :after' instead."))
+
+(defmethod make-instance :around ((class user-class) &key)
+  (let ((initialized-object (call-next-method)))
```
aartaka, line 29: A small nit: use `sera:lret` :)

aartaka:
Note that renderer-specific-code often runs in `initialize-instance :after` and implies that user has already configured the class to their liking. 

In this patch, `customize-instance` runs _after_ `initialize-instance`, thus making it impossible to customize renderer behavior. Bug or feature? Should ve expose two functions — before and after initialization? Should we rather run renderer-specific initialization in `customize-instance :after` (this one I'm on the side of) or even `make-instance :after`?

source/user-classes.lisp
Ambrevar: Thanks!

Ambrevar:
Do you have an example where it could be a problem?

You are making a fair point, this is a crucial design issue.

I believe it cannot be `make-instance :after` which is executed after `(call-next-method` in `make-instance :around`.

If we use `customize-instance :after`, the issue is that for non-user classes we cannot use it, so some classes would be using `customize-instance :after` and other `initialize-instance :after`.  It's inconsistent and prone to trip us.

Possible solution: make everything a user-class and always use `customize-instance` for stuff with side effects.   The the user can customize either `customize-instance :before` and `customize-instance :after`.  Is there any drawback?  After all, if a class is exported it should be fine to make it customizable.

Ambrevar:
@aartaka I've just pushed a rewrite of configure-slot and now `append-configuration` is `auto-configure`.
I've leveraged nfiles for much of the process.

Bonus: Now the class and slot settings won't be duplicated.

aartaka:
Sounds perfect :D 

I'll check a bit later today!

source/browser.lisp
```diff
@@ -569,26 +572,27 @@ The following example does a few things:
...
-                                          (url-dispatching-handler
-                                           'emacs-file
-                                           (match-scheme \"file\")
-                                           (lambda (url)
-                                             (uiop:launch-program
-                                              `(\"emacs\" ,(quri:uri-path url)))
-                                             nil)))
-                                    :initial-value %slot-default%))))"
+\(defmethod configure-instance ((buffer buffer))
+  (setf (request-resource-hook buffer)
```
aartaka, line 576: But wait, `setf` is not necessary here, as `hooks:add-hook` does all the work of modifying the hook, right?

source/buffer.lisp
```diff
@@ -132,7 +137,7 @@ customize to their needs.
 
 Example:
 
-\(define-configuration buffer
+\(defmethod customize-instance ((buffer buffer))
   ((override-map (let ((map (make-keymap \"override-map\")))
```
aartaka, line 141: Should it be a `setf`?

source/buffer.lisp
```diff
@@ -189,11 +194,11 @@ There's no more ability to pass the results to the renderer with :FORWARD.
 
 Example:
 
-\(define-configuration buffer
-  ((request-resource-hook
-    (reduce #'hooks:add-hook
-            '(old-reddit-handler auto-proxy-handler)
-            :initial-value %slot-default%))))")
+\(defmethod configure-instance ((buffer buffer))
+  (setf (request-resource-hook buffer)
```
aartaka, line 198: Unnecessary `setf`?

source/buffer.lisp
```diff
@@ -315,10 +320,10 @@ The handlers take the buffer as argument.")
 See `password:*interfaces*' for the list of all currently registered interfaces.
 To use, say, KeepassXC, set this slot to
 
-  (make-instance 'password:user-keepassxc-interface)
+  (make-instance 'password:keepassxc-interface)
 
-Password interfaces may have user classes (that is, prefixed with 'user-' as in
-the above example), in which case you can use `define-configuration' on them.")
+Password interfaces may be user classes, in which case you can specialize the
```
aartaka, line 325: Wait, do user classes make sense? Maybe something like "Password interfaces are configurable through a `customize-instance' method"?

source/configuration.lisp
```diff
@@ -71,6 +87,109 @@ This is set globally so that extensions can be loaded even if there is no
...
+
+(defun write-init-form-slot (class-name slot-form)
+  `(setf (slot-value ,class-name ',(name slot-form)) ,(value slot-form)))
+
+(defun read-init-form-class (form)
+  "Return:
+- the class name
+- the list of forms, either `slot-form' or a raw s-exp.
+Return NIL if not a class form."
+  (alex:when-let ((class-name (when (eq (second form) 'customize-instance)
```
aartaka, line 113: Will `trivia:match` work better here?

source/configuration.lisp
```diff
@@ -71,6 +87,109 @@ This is set globally so that extensions can be loaded even if there is no
...
+                                (second (first (find-if #'consp form))))))
+    (let ((body (sera:nlet lp ((sexp form))
```
aartaka, line 115: `alexandria:parse-body`?

source/configuration.lisp
```diff
@@ -71,6 +87,109 @@ This is set globally so that extensions can be loaded even if there is no
...
+(defmethod nfiles:deserialize ((profile nyxt-profile) (file auto-init-file) raw-content &key)
+  (flet ((make-init-form (form)
+           (multiple-value-bind (name forms)
+               (read-init-form-class form)
+             (if name
+                 (make-instance 'class-form
+                                :class-name name
+                                :forms forms)
+                 form))))
+    (loop
```
aartaka, line 148: `uiop:slurp-stream-forms`?

source/configuration.lisp
```diff
@@ -71,6 +87,109 @@ This is set globally so that extensions can be loaded even if there is no
...
+      collect (make-init-form form))))
+
+(defmethod nfiles:serialize ((profile nyxt-profile) (file auto-init-file) stream &key)
+  (dolist (form (nfiles:content file))
+    (write
+     (if (class-form-p form)
+         (write-init-form-class form)
+         form)
+     :stream stream)
+    (write-string +newline+ stream)))
```
aartaka, line 160: `(terpri stream)`?

source/help.lisp
```diff
@@ -478,46 +478,25 @@ A command is a special kind of function that can be called with
...
-  (let ((path (nfiles:expand *auto-config-file*)))
-    (with-open-file (file path
-                          :direction :output
-                          :if-exists :append
-                          :if-does-not-exist :create)
-      (let ((*print-case* :downcase))
-        (log:info "Appending to ~s:~&~s" path form)
-        (format file format-directive form)))))
+  (let ((input
+          (loop while t do
```
aartaka, line 486: I wonder if that should rather be `(loop for input = (read-from-string ...) until (typep ...))`, but that can wait :)

source/manual.lisp
```diff
@@ -52,11 +52,15 @@ below to create said file, if it's not created yet.")
                    "Create init file"))))
     (:p "Example:")
     (:pre (:code "
-\(define-configuration buffer
-  ((default-modes (append '(no-script-mode) %slot-default%))))"))
+\(defmethod customize-instance ((buffer buffer))
+  ((override-map (let ((map (make-keymap \"override-map\")))
```
aartaka, line 56: `setf` instead of double paren?

source/manual.lisp
```diff
@@ -98,34 +102,35 @@ add the following to your configuration:")
     (:ul
      (:li "vi bindings:"
           (:pre (:code "
-\(define-configuration buffer
-  ((default-modes (append '(vi-normal-mode) %slot-default%))))")))
+\(defmethod customize-instance ((buffer web-buffer))
+  (nyxt/vi-mode:vi-normal-mode :buffer buffer))")))
```
aartaka, line 106: On that note, maybe we should include a snippet utilizing `enable-modes` so that one can enable multiple modes easily?

source/manual.lisp
```diff
@@ -98,34 +102,35 @@ add the following to your configuration:")
...
+          scheme:vi-normal
+          (list \"g b\" (make-command switch-buffer* ()
+                          (switch-buffer :current-is-last-p t))))))"))
     (:p "The " (:code "override-map") " is a keymap that has priority over
 all other keymaps.  By default, it has few bindings like the one
 for " (command-markup 'execute-command) ".  You can use it to set keys globally:")
     (:pre (:code "
-\(define-configuration buffer
+\(defmethod customize-instance ((buffer buffer))
   ((override-map (let ((map (make-keymap \"override-map\")))
```
aartaka, line 129: `setf`?

source/manual.lisp
```diff
@@ -265,18 +272,18 @@ can set a hook like the following in your configuration file:")
...
         " for a simpler way to achieve the same result.)")
     (:p "Or, if you want to set multiple handlers at once,")
     (:pre (:code "
-\(define-configuration web-buffer
-  ((request-resource-hook
-    (reduce #'hooks:add-hook
-            '(old-reddit-handler auto-proxy-handler)
-            :initial-value %slot-default%))))"))
+\(defmethod customize-configuration ((buffer web-buffer))
+  (setf (request-resource-hook buffer)
```
aartaka, line 283: `setf` unnecessary here, we're imperative with `hooks:add-hook` :)

source/mode/proxy.lisp
```diff
@@ -17,14 +17,15 @@ a proxy for all buffers, add it to the list of default modes.
 
 Example to use Tor as a proxy both for browsing and downloading:
 
-\(define-configuration nyxt/proxy-mode:proxy-mode
-  ((nyxt/proxy-mode:proxy (make-instance 'proxy
-                                         :url (quri:uri \"socks5://localhost:9050\")
-                                         :allowlist '(\"localhost\" \"localhost:8080\")
-                                         :proxied-downloads-p t))))
+\(defmethod customize-instance ((mode nyxt/proxy-mode:proxy-mode))
```
aartaka, line 20: There recently were questions about username and password proxy setting, maybe expand the documentation while we're at it?

source/mode/tts.lisp
```diff
@@ -19,10 +19,9 @@ can be configured by changing the `selector` slot.
 
 Example:
 
-\(define-configuration nyxt/tts-mode:tts-mode
-   ((nyxt/tts-mode:executable \"espeak\")
-    (nyxt/tts-mode:selector \"p, h1, h2, h3, h4, h5, h6\")))
-"
+\(defmethod customize-instance ((mode nyxt/tts-mode:tts-mode))
+   (setf (nyxt/tts-mode:executable mode) \"espeak\")
```
aartaka, line 23: Collapse into one `setf` form?

source/mode/vi.lisp
```diff
@@ -14,8 +14,8 @@ in your configuration file.
 
 Example:
 
-\(define-configuration buffer
-  ((default-modes (append '(vi-normal-mode) %slot-default%))))
+\(defmethod customize-instance ((buffer buffer))
+  (nyxt/emacs-mode:emacs-mode :buffer buffer))
```
aartaka, line 18: Why Emacs mode?

source/password.lisp
```diff
@@ -4,19 +4,19 @@
 (in-package :nyxt)
 
 (defun make-password-interface-user-classes ()
-  "Define user classes so that users may apply define-configuration
-macro to change slot values."
-  (loop for interface in password:*interfaces* do
-           (eval `(define-user-class ,(intern (symbol-name interface)
-                                              (package-name (symbol-package interface)))))))
+  "Define classes with the `user-class' metaclass so that users may use `customize-instance'."
```
aartaka, line 7: Oh wow, we can do it this way :D

aartaka (commented):
All cool, my notes are mostly stylistic :)

source/configuration.lisp
Ambrevar: All I could come up is 

```lisp
(match form
                     ((or (list 'defmethod 'customize-instance _ (cons (list _ class-name) _) _)
                          (list 'defmethod 'customize-instance (cons (list _ class-name) _) _))
                      class-name))
```

This is overly verbose because I don't know how to put a "maybe" around the method qualifier (the `_` after `customize-instance`).

aartaka: Here's the pattern that worked for me:
``` lisp
(list* 'defmethod 'customize-instance (or (list* _ (list* (cons (list _ class-name) _) _)) (list* (cons (list _ class-name) _) _)))
```

But do we actually care about `:after` methods as configurable pieces of code? Does any of our auto-configuration code generate those?

source/mode/proxy.lisp
Ambrevar: Which question again?

source/manual.lisp
Ambrevar: Yup, but I'd like to wait for the mode refactoring because I'm not completely sure we are going to keep `enable-modes`.

source/mode/proxy.lisp
aartaka: [This, for example](https://discourse.atlas.engineer/t/problems-with-proxy-mode/448).

source/manual.lisp
aartaka: No problem :)

source/mode/tts.lisp
Ambrevar: Maybe don't encourage this in the doc to not confuse newcomers?

source/mode/vi.lisp
Ambrevar: Ooops!

Ambrevar:
Thanks for your thorough review.  Most points are addressed.

source/configuration.lisp
Ambrevar: Not very readable... :D 

Do we care?  Not yet, but we might in the future, so better take provisions right away.

source/mode/proxy.lisp
Ambrevar: Thanks, but what do you think should be expanded?

source/mode/tts.lisp
aartaka: But it's a totally valid CL :)

I don't insist, though!

source/mode/proxy.lisp
aartaka: I suggest adding an example with authorized proxy. I.e. the one with some `:url (quri:uri "http://username:pass@localhost:8080"`.

aartaka:
> Thanks for your thorough review. Most points are addressed.

Cool. Looks perfectly mergeable to me then :)

Ambrevar:
Let's merge tomorrow if nobody disagrees in the mean time.

Ambrevar:
Wait, we still need to address your concern from https://github.com/atlas-engineer/nyxt/pull/2194#issuecomment-1100949977.  We may indeed have to add another hook.

aartaka:
> Wait, we still need to address your concern from [#2194 (comment)](https://github.com/atlas-engineer/nyxt/pull/2194#issuecomment-1100949977). We may indeed have to add another hook.

Why not just use `customize-instance :after`?

Ambrevar:
Where?

One plan is to:

- replace all `initialize-instance :after` with `customize-instance`,
- then leave it to the user to use either `customize-instance :before` or
  `customize-instance :after`.

Works and gives good flexibility.  Drawbacks:

- Our codebase would now use a non-standard CLOS API.
- Having all initialization done in one place does not separate slot
  setting from side effects.  We may lack some flexibility.

Regarding the last point, maybe an example is necessery:  say `browser`
sets some slots in its `customize-instance` method _and_
it initializes some FFI stuff like the GTK contexts (which are not
reversible).

For the user, it's then impossible to hook between the two, that is, to
run code that has access to the finalize slot values but sits before the
side effects are run.

I know of no programming paradigm to handle this cleanly.  We could
still draw a conventional line and use 2 methods:

- Set the final slots values in `initialize-instance :after`.
- Process the (possibly irreversible) side effects in
  `customize-instance :after`.

Thus `customize-instance` would sit in-between.  But we still lack a way
to run code _after_ `customize-instance :after`.  Need for another hook?

aartaka:
> We could still draw a conventional line and use 2 methods: 
> - Set the final slots values in `initialize-instance :after`. 
> - Process the (possibly irreversible) side effects in `customize-instance :after`. Thus `customize-instance` would sit in-between. 

Yes, that's exactly what I've been suggesting when I've been suggesting `costomize-instance :after`, thanks, for mentioning the final slot value calculation necessity.

>But we still lack a way to run code _after_ `customize-instance :after`. Need for another hook?

But do we really need it? Some renderer-specific configuration dependent on side effects? Should we even allow renderer-specific configuration?

If we still need it, we can use `customize-instance :around` -- it wraps around `:before`, next, and `:after` methods. I was unable to find the description of his behavior in the spec, but SBCL does it this way, and I have a vague recollection that it should be this way. See an example
``` lisp
(defmethod hello () (format t "hello~%"))
(defmethod hello :around () (format t "around before call-next-method~%") (call-next-method) (format t "around after call-next-method~%"))
(defmethod hello :after () (format t "after hello~%"))
(hello)
;; around before call-next-method
;; hello
;; after hello
;; around after call-next-method
```

aartaka:
Oh hell, CLOS is so thought out...

Ambrevar:
Done, we now use `customize-instance :after` for side effects.

OK to merge?

aartaka:
OK to merge! Thanks for this elegant refactoring, Pierre!

aartaka:
Now I need to rewrite my whole dozen of configuration files to a new style (⊙﹏⊙)

Ambrevar:
You're welcome :D

aartaka:
How about using `progn` method conbination for `customize-instance`? This way one can define as many methods as they want to, and all of those will be invoked :D

Ambrevar:
Aha!  Very good point!  I've mentioned it in #1431 so we don't forget.  I'll finish the class stuff once #2200 is merged.

Issue #1541, opened 2021-06-21
Paginate the manual, and any web page! (a.k.a. Emacs' info-mode)


Ambrevar:
The manual is currently one long page, it can be a bit overwhelming to read.

Same with any long page, by the way.

It would be nice to have an equivalent of Emacs' Info mode that basically displays a virtual paging / sectioning of the content.

edgar-vincent:
What if we kept the long page (I happen to like long pages, as it makes searching easier :) and added an outline/TOC in an [info-buffer](https://github.com/atlas-engineer/nyxt/pull/1391)?

Ambrevar:
We can do both, make it an option, everything is possible! :)

guojing0:
I like this idea and think it could really benefit reading long pages!

So if I understand correctly, say we want to paginate `h2` tags (of course we could give the users the ability to change level of "h") of [Nyxt documentation](https://nyxt.atlas.engineer/documentation), then we would have each "page/buffer" for "Core concepts", "Basic controls", "The Nyxt help system", and so on?

Ambrevar:
Absolutely!

Ambrevar:
Wait, why did you close this @aadcg ?  I believe we don't have pagination for help-mode, do we?

aadcg:
Oh, right! Sorry.

aadcg:
We should revisit this issue as we have been accumulating a bunch of hacks whose goal is achieve what is describes in the issue. Not a high priority. And now we have @lansingthomas!

lansingthomas:
> It would be nice to have an equivalent of Emacs' Info mode that basically displays a virtual paging / sectioning of the content.

Would someone please screenshot emacs info mode and post in this thread.

lansingthomas:
Thanks!

So like `jump-to-heading` but with a persistent title/path. Cool interface.

Names like: "info mode" might be confusing for users in our ecosystem so we might want to try out some different names.

Could we achieve a similar effect by adding a persistent title/path to `jump-to-heading`?
I guess the difference is that `jump-to-heading` keeps the prompt buffer open, which takes up space in a bad way.


### some thoughts
- overwhelm in our manual is a real issue.
   - The proposed tool communicates position in the hierarchy system, that part would an improvement.
   - I like how the Nyxt Manual is a big long doc. Using a Nyxt tool to manage it feels interactive and builds familiarity with Nyxt for navigating texts.
[...]

Issue #1851, opened 2021-11-04
Web mode as a meta mode


jmercouris:
I noticed that many of our modes are based upon web-mode. However, web mode is a very large mode! I am thinking of the following, turn web-mode into a meta mode.

Toggling web-mode will actually toggle a set of modes (jump heading mode, search buffer mode, etc).

For example, we'll add a jump-heading mode that web-mode will toggle. What do you think? A logical way to decompose web-mode?

@Ambrevar

jmercouris:
Definition: a meta mode is a mode that toggles modes.

Ambrevar:
Why not, but first a question: is there a use case where one would want or need `jump-heading-mode` without `search-buffer-mode`?

In other words, does this decomposition make sense in terms of "mode units", from the user perspective?

Alternative suggestion: we could further split web-mode.lisp into multiple files, one for history, one for bindings, etc.

jmercouris:
Perhaps it does not make sense... in any case, I want a way to make `web-mode` smaller and more focused. We can still bind things to disabled modes. Perhaps web-mode should just be a common set of bindings for `jump-heading` mode, etc.?

aadcg:
I bumped into this from another perspective, while working on #2195.

I drew a silly graph from the way we structure our modules. The arrow "A -> B" defines "A depends on B". The 2 arrows I stroke out (emanating from "Help") are those I find illogical. I replaced those 2 with one connecting "Help" to "web-mode". 

"web-mode" is actually "web-mode commands".

![2022-04-22_10:13:24](https://user-images.githubusercontent.com/45483512/164676309-d7d9890d-3aed-4f1c-b0ab-e029741371d1.png)

Ambrevar:
Yes, `help`-related code needs to be reorganized some bits.

Note that I did reorganize the asd modules in an ad-hoc fashion: it's better than before but  by no means fully "logical", it's just good enough as long as we have this illogical mutual dependencies in the code base (web-mode, help, etc.).  

In other words: no need to focus on the .asd, we should focus on the modes vs. the core instead.

aadcg:
I see your point, and it does make sense.

I'm afraid I'm not qualified enough to give much advice, so I'll perhaps wait until a consensus is reached. 

[...]

PR #2195, opened 2022-04-15
Improve the doc interface

aadcg:
This is actual trivial, since it suffices to add some keybindings.

But I'm having a hard time doing it since we need to rethink the modules of the Nyxt system.

Not sure how to proceed yet. Trying to slowly make sense of everything.

Ideas welcome.

Addresses #1541.

### DIFF
--- a/source/mode/help.lisp
+++ b/source/mode/help.lisp
@@ -10,4 +10,21 @@
 
 (define-mode help-mode ()
   "Mode for displaying documentation."
-  ((rememberable-p nil)))
+  ((rememberable-p nil)
+   ;; add 'Type "?" for help' at the top of the web-buffer
+   ;; (constructor
+   ;;  (lambda (mode)
+   ;;    (...)))
+   (keymap-scheme
+    (define-scheme "help-mode"
+      scheme:cua
+      (list
+       "q" 'nyxt::delete-current-buffer
+       "n" 'nyxt/web-mode::next-heading
+       "p" 'nyxt/web-mode::previous-heading
+       "m" 'nyxt/web-mode:jump-to-heading
+       "s" 'nyxt/web-mode::search-buffer
+       "t" 'nyxt::headings-panel
+       ;; TODO local help-mode keybindings don't appear when describe-bindings
+       ;; is called.  Why?  %define-internal-page?
+       "?" 'nyxt::describe-bindings)))))


Ambrevar:
Can you elaborate on what's blocking you?  Why do we need to rethinkg the Nyxt module system?
(I assume you mean the ASDF modules.)

aadcg:
@Ambrevar, yes I was referring to the ASDF modules. I think that `help` should inherit from `web-mode`. Commands such as `jump-to-heading` are paramount for `help`, but they're defined in `web-mode`.

Even if we decided not to chance anything, there's an issue related to `describe-bindings`. It doesn't show the keymap defined in `help`. I couldn't figure this out. Any ideas?

Ambrevar:
For the module question, actually I think the matter at hand here "what
should web-mode be"?

See #1851 for a discussion on web-mode, #1805 for the one about the
buffer class hierarchy.

At this point, I think web-mode has no reason to exist.

Instead, we should use mixins:

- history-buffer: enables history (web-buffers, internal-buffers, but
  NOT status-buffer, message-buffer).
- html-buffer: enables jump-to-headings, plump parsing, etc.

About the keymap: something fishy indeed.  I'll have a closer look next week.

aadcg:
@Ambrevar, regarding the keybindings I was able to make sense of it with @aartaka. No need to look into it.

I'm making sense of our modules.

aadcg:
We have 5 files sharing the same name both at `source/` and `source/mode`:

- bookmark
- help
- message
- no-procrastinate
- prompt-buffer

Shouldn't we at least append `mode` to those existing under `source/mode` to avoid confusion? 

Especially in our `defsystem` statement, I find it hard to follow.

aadcg:
>     Shouldn't we at least append mode to those existing under
>     source/mode to avoid confusion?
>
>     Especially in our defsystem statement, I find it hard to follow.
>
> No, not necessarily. The mode/ directory is there exactly to mark those
> as modes. While you have to look at the :pathname in the defsystem, it's
> quite convenient otherwise.

It's exactly due to the :pathname detail that I bought it up.  I did
confuse me for a little while!

aadcg:
I'll continue the work started here as soon as the 3.0 API will be closer to its final form. Thanks.

aadcg:
Working on this again. 

The `headings-panel` command is broken. The problem was introduces in ef5a6f728f9c993a3a4d0f49d871440b7964ec0c. Investigating.

PR #2203, opened 2022-04-23
Fix docstrings

aadcg:


source/buffer.lisp
```diff
@@ -1649,8 +1649,8 @@ When there is no next buffer, go to the first one so as to cycle."
                    (set-current-buffer next))))
 
 (define-command switch-buffer-last ()
-  "Switch to the last showing buffer in the list of buffers.
-That is to say, the one with the most recent access time after the current buffer."
+  "Switch to the last visited buffer.
+I.e., the one with the most recent access time."
```
Ambrevar, line 1653: I recommend we refrain from this academic initials: they are not necessarily well understood by everyone. (Plus somewhat elitist.)

`I.e.` translate into "That is,".

source/mode/os-package-manager.lisp
```diff
@@ -381,7 +381,7 @@ OBJECTS can be a list of packages, a generation, etc."
     buffer))
 
 (define-command-global switch-os-generation ()
-  "Switch generation of selected profile."
+  "Switch generation of selected profiles."
```
Ambrevar, line 384: I believe you can select only 1 profile here.

source/mode/os-package-manager.lisp
```diff
@@ -397,7 +397,7 @@ OBJECTS can be a list of packages, a generation, etc."
                         profile generation)))
 
 (define-command-global delete-os-generations ()
-  "Delete generations of selected profile."
+  "Delete generations of selected profiles."
```
Ambrevar, line 400: Same.

Ambrevar:
Excellent, thanks for taking the pain to go through all this!

jmercouris:
I can fix the discrepancies pointed out since Andre will be AFK.

aadcg:
@jmercouris that would be appreciated :)

aadcg:
> Excellent, thanks for taking the pain to go through all this!

No pain at all!

PR #2213, opened 2022-04-29
Generic commands

Ambrevar:
This does a very logical thing that we should have done 3 years ago: superclass `command` with... `standard-generic-function`.  It's semantically so meaningful, can't believe we never thought of it before.

Credits to @aartaka for the idea.

In short this allows us:

- To not generate the hooks, since now we can simply `defmethod ... :before`.
- Remove 100 lines of twisted `compile` function generation.
- The previous points coincidentally fixes "go to definition".
- The code is overall some 140+ lines shorter and more expressive than before!

Drawbacks:
- We can no longer use the SBCL function typing it seems.
  Trying to manually type a generic function triggers a warning on compilation, and doing it after compilation seems to break the method calls.  Any idea?

## To do:

- [x] Update `make-command` and friends to be less verbose and match the generic function generation more.

## Questions:

- I've renamed `make-command` to `lambda-command`, etc.  Do you like the new names?  Does the rationale make sense to you?


aartaka:
> This does a very logical thing that we should have done 3 years ago: superclass `command` with... `standard-generic-function`. It's semantically so meaningful, can't believe we never thought of it before.
> 
> Credits to @aartaka for the idea.

It was a random dream/idea and look where we're now :D

> In short this allows us:
> 
> * To not generate the hooks, since now we can simply `defmethod ... :before`.
> * Remove 100 lines of twisted `compile` function generation.
> * The previous points coincidentally fixes "go to definition".

Yay!

> Drawbacks:
> 
> * We can no longer use the SBCL function typing it seems.
>   Trying to manually type a generic function triggers a warning on compilation, and doing it after compilation seems to break the method calls.  Any idea?

Indeed, generic functions are not exactly typeable. Maybe insert `check-type` or something in the body of the function?

Ambrevar:
Should be ready, I'll merge on Monday if there is nothing against it.

source/command.lisp
```diff
@@ -66,221 +46,109 @@ We need a `command' class for multiple reasons:
...
-    (define-command-preamble name arglist body
-      `(make-instance 'command :name ',name :visibility :mode
-                               :docstring ,doc
-                               :fn (lambda (,@arglist) ,@body)))))
+    `(progn
+       (export-always ',name (symbol-package ',name))
+       (prog1 (defgeneric ,name (,@(generalize-lambda-list arglist))
+                (:documentation ,doc)
+                (:method (,@arglist) ,@body)
+                (:generic-function-class command))
```
aartaka, line 141: Once #2205 is merged, we can add `hookable` as the method combination for commands, so that they have infinite number of hooks :D

Ambrevar:
Ok to merge?

aartaka:
OK to merge!

Ambrevar:
Merged with https://github.com/atlas-engineer/nyxt/commit/a7d36abd4cb4f10e596bdb16b392132f600b3f10.

aartaka:
Yay! Thank you Pierre!

PR #2222, opened 2022-05-03
mode/repl.lisp: Verbatim printing mode

shamazmazum:
This commit brings back printing with `prin1-to-string` in REPL mode which was replaced with `value->html`. Rationale behind this:
20220503_21h23m56s_grim

I cannot read the output of `*features*`. Printer says it's a property list, but I don't see semantics of a property list here. It's just a list with even number of arguments, there are no keys and values in it.

Because `value->html` was added on purpose, I think that user must have a choice between it and the previous behaviour (this choice is provided with `verbatim-p` which I introduced). Or maybe even revert aaed554.

### DIFF
--- a/source/mode/repl.lisp
+++ b/source/mode/repl.lisp
@@ -85,7 +85,13 @@ INPUT is a string and RESULTS is a list of Lisp values.")
     :type (or null integer)
     :documentation "Current element of history being edited in the REPL prompt.
 
-Scroll history with `evaluation-history-previous' and `evaluation-history-next'."))
+Scroll history with `evaluation-history-previous' and `evaluation-history-next'.")
+   (verbatim-p
+    nil
+    :type boolean
+    :documentation "If NIL results of evaluation will be
+pretty-printed with VALUE->HTML. Printed representations of most
+objects cannot be read back in this case. If T, use PRIN1-TO-STRING."))
   (:toggler-command-p nil))
 
 (defun package-short-name (package)
@@ -207,11 +213,15 @@ Scroll history with `evaluation-history-previous' and `evaluation-history-next'.
                           for (package input results) in (reverse (evaluation-history repl-mode))
                           collect (:li (:b package "> " input)
                                        (loop
-                                         for result in results
-                                         collect (:li (:raw
-                                                       (value->html
-                                                        result (or (typep result 'standard-object)
-                                                                   (typep result 'structure-object))))))))))
+                                         for result in results collect
+                                             (:li
+                                              (if (verbatim-p repl-mode)
+                                                  (prin1-to-string result)
+                                                  (:raw
+                                                   (value->html
+                                                    result
+                                                    (or (typep result 'standard-object)
+                                                        (typep result 'structure-object)))))))))))
              (:div :id "input"
                    (:span :id "prompt"
                           (format nil "~a>" (package-short-name *package*)))


aartaka:
Hmmmmmm, that feels like quite radical of a change. 

**The problem** that you encountered was indeed a valid one -- `*features*` is not recognised as it should.

**The need** is also quite reasonable -- to make its rendering better.

**The solution** that you suggest I can't agree with, as the simpler way would be to make list printing in `value->html` to not try to infer the list structure and simply print it as is.

aartaka:
05e08b62f2905e0bfe92109ed333db87a159ffa2 should fix the issue of us being overly smart about property lists :)

If you notice any other inconsitencies in how `value->html` works on a certain data -- please, report it, `value->html` would certainly use some improvement!

Other than that, should we close this PR?

shamazmazum:
Another idea was to print compact (and readable) representation of lists.

This can hardly be copy-pasted into a problem report, for example:
20220504_05h40m50s_grim

I see that `value->html` is mostly used in the REPL and in `describe-foo`. Maybe create a pretty-printer for lists then? It can print some elements on the first line with `value->html` then go to another line and so on.

shamazmazum:
I see you have your own ideas about `repl-mode` #2223. You can close this PR if it blocks you in some way. I use this mode rarely, SLIME is much better :)

jmercouris:
SLIME is better- today! But with feedback and help, we can make it better :-)

Ambrevar:
We can do better to detect plist.  @aartaka What about the following:
- even number of elements.
- Every odd element is a keyword.
- Not all even elements are keywords.

With the above, `*features*` would be detected as a list and not a plist.

aartaka:
> We can do better to detect plist. @aartaka What about the following:
> 
>     * even number of elements.
> 
>     * Every odd element is a keyword.
> 
>     * Not all even elements are keywords.
> 
> 
> With the above, `*features*` would be detected as a list and not a plist.

Hmmmm. But what if all the plist values are keywords too? 

Lists are pretty unstructured and there are too much exceptions, so I'm not sure what to do here.

Ambrevar:
Then it's a false match, but that's fine, at least it's a rarer case.

Ambrevar:
So what's left to do here?

- Handle `*print-pretty*` to display long structures more concisely, maybe with an interactive ellipsis?
- I'm still pushing my suggestion (https://github.com/atlas-engineer/nyxt/pull/2222#issuecomment-1116995299) to better handle plist display :)

aartaka:
> * Handle `*print-pretty*` to display long structures more concisely, maybe with an interactive ellipsis?

I'm not sure I understand you here. Why `*print-pretty*`? 

We do need the ellipsis, true, but I'd better make it dependent on the structure size, not the `*print-pretty*`.

> * I'm still pushing my suggestion ([mode/repl.lisp: Verbatim printing mode #2222 (comment)](https://github.com/atlas-engineer/nyxt/pull/2222#issuecomment-1116995299)) to better handle plist display :)

But wait, it's done already: d222dcbc3e4b0873bba68f01864171602703a7a8 :)

Ambrevar:
Oh, I didn't notice, great!

About `*print-pretty*`: because it allows us to control if we want to print a list "horizontally" or with each item on a new line.

There is also `*print-length*` which we can leverage to print an ellipsis.  So we could add easily accessible option to toggle between the rendering of content with `*print-length*` set to, say, 100, and NIL.

aartaka:
I see! I've opened #2243 regarding it. Closing this then?

Ambrevar:
Yup, thanks!

PR #2223, opened 2022-05-03
Redesign the REPL to be a Notebook-like environment.

aartaka:
This a draft of how our REPL could look as a Notebook-like multiple-pane enviroment. See the attached screenshot for the visuals. The features so far are:
- Evaluation of independent cells.
- Movement between cells.

repl-redesigned

# Things to Do
- CUA bindings (based on Jupyter Notebook).
- Inline debugging (will require refactoring our internal debugger a bit).
- More commands for cell management.

# Things to Discuss
- Do we actually want to have it as a default REPL? Feels unfamilliar, even though it's cool...
- What are the VI bindings that could make sense for Notebook mode?
- Should we allow moving cells?
- How should one add new cells -- only in the bottom of the screen or between any existing cells?

### DIFF
--- a/source/mode/repl.lisp
+++ b/source/mode/repl.lisp
@@ -94,7 +94,12 @@
           :documentation "The CSS applied to a REPL when it is set-up.")
    (evaluations
     (list)
-    :documentation "A list of `evaluation's representing the current state of the REPL."))
+    :documentation "A list of `evaluation's representing the current state of the REPL.")
+   (current-evaluation
+    nil
+    :type (maybe integer)
+    :reader t
+    :documentation "Index of the evaluation currently being focused."))
   (:toggler-command-p nil))
 
 (defun package-short-name (package)
@@ -128,6 +133,15 @@
     (with-current-buffer (buffer mode)
       (selection-start new-position))))
 
+(define-parenscript focus (selector)
+  (ps:chain (nyxt/ps:qs document (ps:lisp selector)) (focus)))
+
+(defmethod (setf current-evaluation) (new-index (mode repl-mode))
+  (if new-index
+      (focus (format nil ".input-buffer[data-repl-id=\"~a\"]" new-index))
+      (focus ".input-buffer[data-repl-id=\"\"]"))
+  (setf (slot-value mode 'current-evaluation) new-index))
+
 (defmethod current-cell-id ((mode repl-mode))
   (pflet ((get-id ()
            (ps:chain document active-element (get-attribute "data-repl-id"))))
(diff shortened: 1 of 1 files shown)


jmercouris:
Artyom, you just keep blowing my mind. This is so damn cool.

To some questions:

1. Do we want it to be the default REPL? No, not yet.
2. No idea
3. Moving cells should be allowed
4. Only in the bottom screen, you can add and then move it if you want.

libraries/web-extensions/nyxt.lisp
```diff
@@ -0,0 +1,17 @@
...
+(sb-alien:define-alien-callable inject-js-world-function int ()
+  (g:g-signal-connect
+   (webkit:webkit-script-world-get-default) "window-object-cleared"
+   (lambda (world page frame)
+     (declare (ignorable page))
+     (webkit::%make-jsc-function (webkit:webkit-frame-get-javascript-context-for-script-world frame world)
+                                 "hello" (cffi:callback hello) 0))))
+
+(sb-alien:define-alien-callable inject-js-world-function int ()
+  (format t "Hello!"))
```
Ambrevar, line 17: Wait, is this related to this pull request?

Ambrevar:
Wow, fantastic stuff!

Looks like a great opportunity to handle separate `nyxt` processes, nah? :D

libraries/web-extensions/nyxt.lisp
aartaka: Oh, that's interesting 0_o 

I guess that my compulsive keypresses countering Magit hang caused this to be staged 😅

aartaka:
> Wow, fantastic stuff!
> 
> Looks like a great opportunity to handle separate `nyxt` processes, nah? :D

I don't fully understand how that could work. Can you provide some code snippet for how you imagine it?

Ambrevar:
Untested:

```lisp
(let ((nyxt-program (sera:resolve-executable "nyxt")))
  (let ((process (uiop:launch-program (list nyxt-program
                                            "--headless"
                                            "--no-init"
                                            "--eval" (write-to-string SNIPPET))
                                      :output :stream
                                      :error-output :stream)))
    (run-thread "REPL printer"
      (loop for object in (read (uiop:process-info-output process) nil :eof)
            for error-object in (read (uiop:process-info-error-output process) nil :eof)
            until (or (eq object :eof)
                      (eq error-object :eof))
            do (progn
                 (WRITE-TO-CELL object)
                 (WRITE-TO-CELL-ERROR error-object))))))
```

Of course the drawback is that we have no debugger access.  To debug a remote process, we  would need something like Slynk (not all of it, but the bidirectional stream to control a Lisp process remotely).

aartaka:
But then, do we really need it in a Notebook-like environment? The cool part about Jupyter, for example, is that cells, even though separate from each other, modify the same shared image.

aartaka:
In 73595af08310bc0ae84644b4d58b79ac71b46f9b i've added automatic variable binding useful for a notebook-like workflow. While it would be better to go with a reader macro (@Ambrevar methioned a `#vX.Y` reader macro in his FOSDEM presentation), autogenerated variables are nice too :)

aadcg:
@aartaka, this is amazing and exciting!

I have strong opinions about this, since I've used Jupyter Notebooks extensively in the past. 

My take on topics you brought for discussion:
1. Absolutely, no hesitation, as soon as we're happy with it. 

Let's think about REPL interfaces. They're composed by inputs and outputs. 

The simplest interface allows the user's caret to navigate the line (not multi-lines) relative to the current input (think Linux console with GNU Readline). 

A richer interface (think Emacs' `comint`) gives the user's caret complete freedom across all inputs and outputs. A (hard) question immediately arises: should older inputs and outputs be read-only (Emacs today says no, for instance)? In this UX, while depending on the language, multi-line inputs are supported. 

Notice that in the cases above, text and colour are the only things supported. A yet richer interface allows images, videos, LaTeX, markdown and the kitchen sink. Besides, the ability to move pairs of input/output around. It seems a bad idea to let people mix and match mismatching input/output pairs. This is basically the interface of org's mode literate programming facilities and Jupyter Notebooks. Notice that it's no secret that the creator of the latter got its inspiration from the former. 

In short, there's no reason not to aim for the most advanced REPL interface.

2. I can help you with keybindings in general. Would you allow me to push (not force-push) to this branch? :)
3. Yes, pair-wise as mentioned above.
4. The user should be allowed to add cell-pair anywhere. From an implementation perspective, it makes sense to add to the bottom and move where needed.

I need to spend some time playing with Jupyter Notebooks again and I might change some of the opinions above.

Also, it's completely ok to refer to Jupyter Notebooks in ours discussions, but let's not use that word when the feature is out. We are doing our take on a modern web-based REPL (that we can come up with a name for, if needed).

aadcg:
It seems that `previous-cell` isn't behaving properly. 

What do you think about navigating output-wise and editing outputs?

aartaka:
(MREPL is already taken by SLY...)

Ambrevar:
> But then, do we really need it in a Notebook-like environment? The cool part about Jupyter, for example, is that cells, even though separate from each other, modify the same shared image.

Absolutely, but the ability to choose between both would be great!  So each input/output pairs would store the information of which Lisp process they are running, and multiple pairs may share a process.  The UI should make it obvious which process a pair is associated to.

aadcg:
@aartaka, I'm concerned with `paren` and `closing-paren`. 

Concerning the former, I understand the attempt at structural editing, but then if I delete an "(", I'd like the pair to be gone as well.

Concerning  the latter, how can I say `(print "Hello Artyom! :)")?

Do we want go about these textual editing aspects now, in the context of the REPL? Sooner or later, it will have to be abstracted and generalized somewhere else. 

I'd keep things pretty bare now, but I also acknowledge the benefits of having some amenities.

aartaka:
Oh wow, I see 0_o

aartaka:
As I expected, only implementing the full-fledged Lisp reader can save us here xD

jmercouris:
> 

By using a unique character/emoji for each process! If I remember I had written code that does that at some point, it is somewhere in our Git history. We had used it for the tree history if I recall correctly.

aartaka:
> > The UI should make it obvious which process a pair is associated to.
> 
> By using a unique character/emoji for each process! If I remember I had written code that does that at some point, it is somewhere in our Git history. We had used it for the tree history if I recall correctly.

Yes, I remember it. Was quite unsettling :D

jmercouris:
Well, there was no other easy way to get it to work for color blind people! I also thought about shapes with increasing number of sides, triangle to square to pentagon etc. but that quickly gets out of hand!

aartaka:
How about we merge this without multi-process support and inline debugging? The `master` is moving too fast, and I want to solidify our new REPL design there. The debugger and multiple process can extend it later.

The only thing left there is keybindings. @aadcg, any new insight about that?

Ambrevar:
Completely fine with merging now!

aadcg:
I agree that multi processing and debugging should come later.

@aartaka, as we've discussed, there's no such thing as "CUA jupyter notebook keybindings". Jupyter Notebook uses VI keybindings. We seem to have agreed that CUA users will do well with the mouse.

I'd be fine about merging if we leave out `paren` and `closing-paren`. Also, I wasn't able to evaluate a cell in CUA mode. Are you aware of that @aartaka?

aartaka:
Rebased over master, it works. Merging.

aartaka:
> @aartaka, as we've discussed, there's no such thing as "CUA jupyter notebook keybindings". Jupyter Notebook uses VI keybindings. We seem to have agreed that CUA users will do well with the mouse.

Yep.

> I'd be fine about merging if we leave out `paren` and `closing-paren`. 

Done!

> Also, I wasn't able to evaluate a cell in CUA mode. Are you aware of that @aartaka?

It seems to work for me in CUA. Did you press `C-return` or just a `return`?

Ambrevar:
Hooray!

PR #2224, opened 2022-05-04
Redesign buffer hierarchy.

### LINKED ISSUE #1805: Redesign buffer class hierarchy

Ambrevar:
Rationale: Make `buffer' as light as possible, so that in particular dummy
buffers are not so bloated.
Then ensure "special" buffers only have what they need.  For instance, the
message buffer should probably not have modes, network access or a download directory.

Base class hierarchy:

- buffer
  - navigable-buffer
	- focus-buffer
	- network-buffer
	- input-buffer
	- modable-buffer

Then our actually usable buffers are partly defined by their superclasses:

- web-buffer (focus-buffer network-buffer modable-buffer navigable-buffer input-buffer)

- editor-buffer (input-buffer modable-buffer navigable-buffer)

- prompt-buffer (internal-buffer network-buffer input-buffer modable-buffer navigable-buffer prompter:prompter)

Bonus: Rename list-modes to modes-string, since it's what it does.

## Questions:

- Do we still need `internal-buffer`?  For instance `editor` is an internal buffer, and if we try to set the URL to an HTTP page from this buffer, it will open a new one.  But with the `nyxt:` scheme support this should no longer be needed.  @aartaka ?
- Do you like the name `focus-buffer`?
- Should `panel-buffer` be modable?
- Remove `dummy-buffer`?  It's just a `buffer` now.
- Some functions like the `customize-instance` `:after` method for `buffer` could be specialized, e.g. against `focus-buffer` and the history snippet could be moved there.  Logic-wise, both ways work, but is there a benefit (in terms of extensibility?) in having multiple specializations?

Fixes #1805.


aartaka:
> Rationale: Make `buffer' as light as possible, so that in particular dummy buffers are not so bloated. Then ensure "special" buffers only have what they need. For instance, the message buffer should probably not have modes, network access or a download directory.
> 
> Base class hierarchy:
> 
> * buffer
>   * navigable-buffer
>   * focus-buffer
>   * network-buffer
>   * input-buffer
>   * modable-buffer
> 
> Then our actually usable buffers are partly defined by their superclasses:
> 
> * web-buffer (focus-buffer network-buffer modable-buffer navigable-buffer input-buffer)
> * editor-buffer (input-buffer modable-buffer navigable-buffer)
> * prompt-buffer (internal-buffer network-buffer input-buffer modable-buffer navigable-buffer prompter:prompter)

You mention `internal-buffer` here, but where is it in the hierarchy above?

> Bonus: Rename list-modes to modes-string, since it's what it does.
OK!
> ## Questions:
> * Do we still need `internal-buffer`?  For instance `editor` is an internal buffer, and if we try to set the URL to an HTTP page from this buffer, it will open a new one.  But with the `nyxt:` scheme support this should no longer be needed.  @aartaka ?

Actually, what would it mean -- opening an HTTP link in an editor buffer? It wouldn't make sense, I suppose. I guess we have to keep `internal-buffer`, but make it non-instantiatable.

> * Do you like the name `focus-buffer`?

`focusable-buffer`? `view-buffer`? The option with `main-buffer` is fine too...

> * Should `panel-buffer` be modable?

Yes! At least for mode-specific styling...

> * Remove `dummy-buffer`?  It's just a `buffer` now.

Yes!

> * Some functions like the `customize-instance` `:after` method for `buffer` could be specialized, e.g. against `focus-buffer` and the history snippet could be moved there.  Logic-wise, both ways work, but is there a benefit (in terms of extensibility?) in having multiple specializations?

I'm for multiple specializations so that it's easier to debug and extend!

Ambrevar:
>  You mention internal-buffer here, but where is it in the hierarchy above?

I didn't list everything.  Here is the definition:

```lisp
(define-class internal-buffer (focus-buffer) ; TODO: Remove?
  ()
  (:export-class-name-p t)
  (:export-accessor-names-p t)
  (:export-predicate-name-p t)
  (:accessor-name-transformer (class*:make-name-transformer name))
  (:metaclass user-class))
```

> Actually, what would it mean -- opening an HTTP link in an editor buffer? It wouldn't make sense, I suppose. I guess we have to keep internal-buffer, but make it non-instantiatable.

Isn't it the same as with our help buffers?  We _can_ open an HTTP URL in a help buffer, so why not in all internal buffers?

I'm not sure about making anything non-instantiable, because there are cases where we do instantiate `buffer` and `focus-buffer` to access their slot values.

> focusable-buffer? view-buffer? The option with main-buffer is fine too...

I like `focusable-buffer` the most because it's more consistent with `navigable-buffer`, etc.  Plus `focus-buffer` is ambiguous, it sounds like a verb.

>  I'm for multiple specializations so that it's easier to debug and extend!

I think it's about the same, maybe a bit worth for multiple methods because you'd have to trace the methods, whereas when you have only one you can follow the logic from a single spot.

aartaka:
> > You mention internal-buffer here, but where is it in the hierarchy above?
> 
> I didn't list everything. Here is the definition:
> 
> ```lisp
> (define-class internal-buffer (focus-buffer) ; TODO: Remove?
>   ()
>   (:export-class-name-p t)
>   (:export-accessor-names-p t)
>   (:export-predicate-name-p t)
>   (:accessor-name-transformer (class*:make-name-transformer name))
>   (:metaclass user-class))
> ```

Yes, that I've seen :D

>  > Actually, what would it mean -- opening an HTTP link in an editor buffer? It wouldn't make sense, I suppose. I guess we have to keep internal-buffer, but make it non-instantiatable.
> 
> Isn't it the same as with our help buffers? We _can_ open an HTTP URL in a help buffer, so why not in all internal buffers?

But wait, is help buffer a separate class, or is it just a `web-buffer`? To clarify:
- Help buffer is a regular `web-buffer`, which can open any URL, but just happens to open `nyxt:manual`, for example.
- Internal buffer is a buffer of `internal-buffer` class, that is not intended for any URL opening except this one URL it's assigned.

But then, if we move all the functions of `ditor-buffer` into `editor-mode` and make `editor-buffer` to be a plain `web-mode`, it would hurt no one, I suppose. If it doesn't break anything, I'm all in for `internal-buffer` removal :D

> I'm not sure about making anything non-instantiable, because there are cases where we do instantiate `buffer` and `focus-buffer` to access their slot values.

Yes, that's a good point. What I meant was rather that `internal-buffer` shouldn't be instantiatable by user action. But if we're removing it, then it doesn't make sense discussing this :)

aartaka:
To clarify: the biggest obstacle to `internal-buffer` removal is the fact that we have a `focusable` `editor-buffer` that has special behavior hardcoded into it independent of the URL. If we move all the `editor-mode` logic into `editor-mode`, then `internal-buffer` will not be necessary!

Ambrevar:
Err... Can you point out where this hardcoded logic is?  I

Ambrevar:
Never mind, got it.  I'll give it a try.

Ambrevar:
I believe we can indeed remove internal contexts then.  That said, we can keep @tiberious726 logic to allow the user to extend the contexts they want.

aartaka:
Sure thing!

Ambrevar:
OK to merge?

Ambrevar:
> What's the meaning of the internal context now that there's no internal buffers?

Actually the internal  context can still be used for "buffer-less contexts", e.g. as used by the status buffer and the message buffer.

Many things look wrong in `make-web-view`:

- This internal / non-internal looks like we should rather specialize on the buffer type.
- The `nosave-profile` exception calls for another specialization.

```lisp

(defmethod make-web-view ((buffer buffer) (profile nyxt-profile))
  "Return a web view instance."
  (make-instance 'webkit-web-view-ephemeral
                 :web-context (get-context *browser* +internal+
                                           :ephemeral-p t)))

(defmethod make-web-view ((buffer focusable-buffer) (profile nyxt-profile))
  (make-instance 'webkit:webkit-web-view
                 :web-context (get-context *browser* (context-name buffer)
                                           :ephemeral-p nil)))

(defmethod make-web-view ((buffer nosave-buffer) (profile nyxt-profile))
  (make-instance 'webkit-web-view-ephemeral
                 :web-context (get-context *browser* (context-name buffer)
                                           :ephemeral-p t)))

(defmethod make-web-view ((buffer buffer) (profile nosave-profile))
  (make-instance 'webkit-web-view-ephemeral
                 :web-context (get-context *browser* (context-name buffer)
                                           :ephemeral-p t)))
```

We can address this once this pull request is merged.

source/buffer.lisp
```diff
@@ -830,8 +859,8 @@ when `proxied-downloads-p' is true."
       (url proxy))))
 
 (defun load-failed-p (buffer)
-  "Only web-buffer loads can fail."
-  (and (web-buffer-p buffer)
+  "Only `network-buffer' loads can fail."
+  (and (network-buffer-p buffer)
```
aartaka, line 863: Maybe make it a method then?

source/buffer.lisp
```diff
@@ -1009,12 +1012,13 @@ If URL is `:default', use `default-new-buffer-url'."
 (defun buffer-delete (buffer)
```
aartaka, line 1012: Make it a method too? We might need to add a specialization for other buffer types to clean things up. Saving files in `editor-buffer`, for example.

source/history.lisp
```diff
@@ -368,7 +368,8 @@ Finally go through all the owners and update their creator."
              nil))))))
 
 (defun histories-directory (&optional (buffer (current-buffer)))
-  (nfiles:parent (nfiles:expand (history-file buffer))))
+  (when (focusable-buffer-p buffer)
```
aartaka, line 371: It feels semantically confusing -- why is it `focusable-buffer`, and not `navigable-buffer` (which a presumed to be about web navigation/browsing, not page-local movement), that has history? It also seems that `navigable-buffer` is quite empty. 
- How about merging the two? 
- Or renaming the `focussable-buffer` to `main-buffer` (as you suggested) or to `navigable-buffer`, `browsable-buffer`; and `navigable-buffer` -- to `scrollable-buffer` (the name that kind of hints at its weirdness as an entity...)?

aartaka: No, I don't like `scrollable-buffer` name myself, but `navigable-buffer` is more confusing than that.

source/mode.lisp
```diff
@@ -254,18 +254,19 @@ PACKAGES should be a list of package designators."
...
-                    (modes buffer)
-                    :key #'class-of)))
-      (when (< 1 (length results))
-        (log:warn "Found multiple matching modes: ~a" results))
-      (values (first results)
-              results))
-    ;; CCL catches the error at compile time but not all implementations do,
-    ;; hence the redundant error report here.
-    (error "Mode ~a does not exist" mode-symbol)))
+  (when (modable-buffer-p buffer)
```
aartaka, line 257: Specialize agains `modable-buffer`?

source/mode/editor.lisp
```diff
@@ -33,6 +34,24 @@ get/set-content (which is necessary for operation)."
          (insert-content (ps:ps (ps:chain document (write (ps:lisp content))))))
     (ffi-buffer-evaluate-javascript-async (buffer editor) insert-content)))
 
+(define-class editor-buffer (focusable-buffer modable-buffer navigable-buffer input-buffer)
```
aartaka, line 37: I wonder if we should do the same with e.g. `web-buffer` -- move it to `web-mode`...

No, it's probably a bad idea.

source/mode/editor.lisp
```diff
@@ -33,6 +34,24 @@ get/set-content (which is necessary for operation)."
...
+  (:metaclass user-class)
+  (:documentation "Each editor buffer matches a file. Each editor buffer
+contains an `nyxt/editor-mode:editor-mode' instance (or a subclass thereof)."))
+
+(defmethod file ((buffer editor-buffer))
+  (uiop:parse-native-namestring (quri:uri-path (url buffer))))
+
+(defmethod nyxt:default-modes :around ((buffer editor-buffer))
+  ;; REVIEW: Really remove web-mode from editor-buffer?
+  (remove 'web-mode (call-next-method)))
```
aartaka, line 53: Should we rather append `web-mode` for those using it? This way the `:around` method for `editor-buffer` will not be necessary and `define-configuration` will behave predictably.

source/web-extensions-callbacks.lisp
```diff
@@ -69,6 +69,7 @@
 
 (defun all-extensions (&key (buffers (buffer-list)))
   (loop for buffer in buffers
+        when (modable-buffer-p buffer)
         append (sera:filter #'nyxt/web-extensions::extension-p (modes buffer))))
```
aartaka, line 73: Specialize `modes` against other buffer classes too, returning `nil`? Does it make sense, though? (҂⌣̀_⌣́)

aartaka (commented):
All cool, except for some minor CLOS ideas :P

source/history.lisp
Ambrevar: I've decided for `document-buffer` and `context-buffer`.

source/mode/editor.lisp
Ambrevar: Will do when we refactor web-mode.

source/web-extensions-callbacks.lisp
Ambrevar: I'll try it!

Ambrevar:
Merged with 5d86eca3aec18db98ca667f764fd26f43e54f490.

PR #2225, opened 2022-05-05
libssl-dev added for debian-based GTK dependencies

shakatoday:
fixed comiple error

Unable to load any of the alternatives:
("libcrypto.so.1.1" "libcrypto.so.1.0.0" "libcrypto.so")


Ambrevar:
Thanks for reporting this!

PR #2226, opened 2022-05-05
[FOLLOW UP #2192] Improve hinting UI

aadcg:
Sorry again for making a terrible mistke in #2192.

----

I believe we're very close to merge, so let's go through what might be missing or unclear:

-  I deleted the deprecated ANNOTATE-VISIBLE-ONLY-P since there's a new release coming soon. Is this the correct policy?
-  @aartaka what do you suggest doing about 48317740?

### DIFF
--- a/source/mode/element-hint.lisp
+++ b/source/mode/element-hint.lisp
@@ -124,16 +124,7 @@
    (prompter:follow-mode-functions
     (lambda (suggestion)
       (highlight-selected-hint :element suggestion
-                               :scroll nil)
-      (sera:and-let* ((auto-follow (nyxt/web-mode:auto-follow-hints-p (find-submode 'web-mode)))
-                      (matches (string-equal
-                                (prompter:input (current-prompt-buffer))
-                                (plump:get-attribute suggestion "nyxt-hint")))
-                      (input (prompter:input (current-prompt-buffer))))
-        (run-thread "hint auto-follow thread"
-          (sleep (nyxt/web-mode:auto-follow-timer (find-submode 'web-mode)))
-          (when (string= input (prompter:input (current-prompt-buffer)))
-            (prompter:return-selection (current-prompt-buffer)))))))
+                               :scroll nil)))
    (prompter:actions (list 'identity
                            (lambda-command click* (elements)
                              (dolist (element (rest elements))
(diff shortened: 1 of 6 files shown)


source/mode/element-hint.lisp
```diff
@@ -153,39 +144,35 @@
...
-                                   (add-element-hints :selector selector)))
-                   :after-destructor
-                   (lambda ()
-                     (with-current-buffer buffer
-                       (remove-element-hints))))))
-      (when result
-        (funcall function result))
-      (with-current-buffer buffer
-        (remove-element-hints)))))
+  (let ((buffer (current-buffer)))
```
aartaka, line 156: Maybe make the `buffer` binding part of the `when-let` below too? Will look less nested.

source/mode/element-hint.lisp
```diff
@@ -347,70 +334,60 @@ FUNCTION is the action to perform on the selected elements."
...
     (query-hints "Open element with current modes in new buffer"
                  (lambda (result)
                    (mapcar (alex:rcurry #'%follow-hint-with-current-modes-new-buffer buffer)
-                           result))
-                 :multi-selection-p t)))
+                           result)))))
 
 (define-command copy-hint-url ()
-  "Show a set of element hints, and copy the URL of the user inputted one."
+  "Prompt for element hints and save to clipboard."
```
aartaka, line 384: Save what?

source/mode/prompt-buffer.lisp
```diff
@@ -36,6 +36,12 @@ Actions can be listed and run with `return-selection-over-action' (bound to
        "C-down" 'select-last
        "C-pagedown" 'select-next-source
        "C-pageup" 'select-previous-source
+       ;; scroll-page-{down|up}-other-buffer and scroll-other-buffer-{down|up}
+       ;; aren't bound on VI. ideas?
+       "shift-pagedown" 'scroll-page-down-other-buffer
+       "shift-pageup" 'scroll-page-up-other-buffer
+       "shift-down" 'scroll-other-buffer-down
```
aartaka, line 43: I wouldn't use `shift-down` -- it's too widespread.

Actually, we can ignore CUA bindings -- they are often too restrictive for what we're trying to do :P

source/mode/prompt-buffer.lisp
```diff
@@ -36,6 +36,12 @@ Actions can be listed and run with `return-selection-over-action' (bound to
        "C-down" 'select-last
        "C-pagedown" 'select-next-source
        "C-pageup" 'select-previous-source
+       ;; scroll-page-{down|up}-other-buffer and scroll-other-buffer-{down|up}
+       ;; aren't bound on VI. ideas?
```
aartaka, line 40: CC @Ambrevar

Ambrevar: From https://stackoverflow.com/questions/9491013/scroll-the-other-window-in-vim-split, looks like there is no official Binding.

`M-j`/`M-k` or `C-M-j`/`C-M-k` could do.

Ambrevar:
@aadcg A nit, but don't forget to name your PRs explicit and not (only) by references to something else.  For instance, you could name this one "[FOLLOW UP #2192] Improve hinting UI".

source/mode/element-hint.lisp
aadcg: Sure! See 6b39e0e7.

source/mode/prompt-buffer.lisp
aadcg: Ignore all CUA bindings relative to scrolling the buffer behind the prompt correct? I'm fine with it!

See c7c92e13.

aartaka:
Yes, looks fine to me.

aadcg:
@Ambrevar ok to merge?

PR #2229, opened 2022-05-05
Hook-based customization.

aartaka:
This is an alternative to #2205. It's much simpler and has less shady points about it (boooooring :P) -- it's just one `customize-hook` in every `user-class`.

One particular problem that it creates is that `password:keepassxc-interface` seems to not have the `customize-hook` slot. Can this happen because it's receiving `user-class` as a meta-class after actually being defined? How do we circumvent this?

How does this look? Test it and say what breaks :P


source/user-classes.lisp
```diff
@@ -3,7 +3,8 @@
 
 (in-package :nyxt)
 
-(defclass user-class (standard-class) ())
+(defclass user-class (standard-class)
+  ((customize-hook :initform (make-instance 'hooks:hook-any))))
```
Ambrevar, line 7: Here the slot is unexported.  Feature?  Maybe so: The user only needs to know about `customize-instance`, and the hook is reserved to `define-configuration`.

If you agree with this design, let's add this rationale to the doc of the slot.

source/configuration.lisp
```diff
@@ -196,6 +196,48 @@ Return NIL if not a class form."
...
+                  ,@(loop for ((slot value)) on slots-and-values
+                          when (find slot (mopu:slot-names class))
+                            collect `(setf (slot-value object (quote ,slot))
+                                           (let* ((%slot-value% (slot-value object (quote ,slot)))
+                                                  (%slot-default%
+                                                    ,(if (c2mop:class-finalized-p (find-class class))
+                                                         (getf (mopu:slot-properties class slot) :initform)
+                                                         (progn
+                                                           (echo-warning
+                                                            "%SLOT-DEFAULT% not found for ~a of ~a, falling back to %SLOT-VALUE%"
```
Ambrevar, line 230: Strange message, shouldn't it be something like "Slot default not found for slot ~a of class ~a, fall back to it's current value"?

source/configuration.lisp
```diff
@@ -196,6 +196,48 @@ Return NIL if not a class form."
...
+  "Holds the default value of the slot being configured when in `define-configuration'.")
+
+(export-always 'define-configuration)
+(defmacro define-configuration (classes (&body slots-and-values))
+  `(progn
+     ,@(loop
+         for class in (uiop:ensure-list classes)
+         for handler-name = (gensym "DEFINE-CONFIGURATION")
+         collect
+         ;; Random symbol for the method to always stay by itself when computing effective method.
```
Ambrevar, line 214: Which random symbol and which method?  I don't understand what this comment refers to.

Ambrevar:
Looks great, I like it! (Of course I'm biased :p)

I'll test later.

For the password, the user-classes are created in the `nyxt` package, not in the `password` package, so you should customize `nyxt:keepassxc-interface` instead and it should work.

Ambrevar:
To do: re-add helper to delete `customize-instance` methods?
Add helper to clear the hooks?

aartaka:
All addressed, should work just fine now!

aartaka:
Merging then?

jmercouris:
For me, it is OK! I have no comments.

aartaka:
Yay, `define-configuration` works again! :D

jmercouris:
Thank you Artyom 😄

PR #2236, opened 2022-05-07
diff: add package definition docstring

hendursaga:


aadcg:
Thank you @hendursaga!

PR #2238, opened 2022-05-09
Activitypub support (read-only so far)

aartaka:
This introduces ActivityPub support, allowing to read content from [Mastodon](https://joinmastodon.org/), [Plume](https://joinplu.me/), [Lemmy](https://join-lemmy.org/) and other federated social networks -- all in the same Nyxt-native configurable interface.

# What's New
- An internal library for JSON-LD "parsing" (rather, guessing the contents of it) and "unparsing"/serialization. See `parse-object` and `unparse-object`.
  - This should be abstracted to a separate repository at some point in time...
  - Also, there are `fetch-object` and `send-object` that allow getting/sending the objects from/to somewhere else.
- An `ap:` scheme allowing one to look at ActivityPub-enabled pages.
  - `nyxt/activitypub-mode:object->html` generic function is responsible for HTML rendering of those, and is replaceable/extensible, much like any other generic function.
- `nyxt/activitypub-mode:activitypub-mode` to enable on `ap:` pages for additional functions.
- A set of helper functions (`name*`, `url*`, `author*`, `published*` etc.) for ActivityPub page information extraction.
- `activitypub-login` to authorize Nyxt acessing profile data. 
  - It's useless for e.g. Mastodon without HTTP signatures (see Things to Discuss).

# Things to Discuss
- Mastodon and, quite likely, some other federated social networks require [HTTP signatures](https://docs.joinmastodon.org/spec/security/). For that, we need to use some cryptography library and be smart about encrypting parts of our requests. Ironclad?
  - This lack of signatures is what stops us from being able to create content on ActivityPub. Once we have this -- we also have content creation, following other Fediverse people, saving things to collections -- everything that ActivityPub allows!
- Does the interface look fine? You can check profile rendering by going to `ap:ephemeral.glitch.social/@aartaka` and looking at profiles of the ones I follow :)
  - More testing with other platforms is welcome! Peertube is not yet supported, for example, as it has lots of URLs for every specific video, and out algorithm gets confused for which one to pick.
- I'm using a newly intoruced `.card` CSS class for posted content styling. Should we abstract it to `buffer`? Somewhere else? I mean, we're getting dangerously close to having a kind of UI system, yet the one scattered all around the modes and duplicating/repeating itself terribly.
- I've added the `ap:` scheme to the list of internal schemes so that we can have `lisp:` URLs and interactive actions on ActivityPub-enabled pages. These pages, however, can contain arbitrary web content, including easily exploitable HTML+JS pages! This is dangerous ヽ(ﾟДﾟ)ﾉ

# References
- [ActivityPub standard](https://www.w3.org/TR/activitypub/).
- [ActivityStreams, and underlying JSON-LD-based data trasfer format](https://www.w3.org/TR/activitystreams-core/)
- [ActivityStreams Vocabulary, an index of ActivityStreams (and, by extension, ActivityPub) entities](https://www.w3.org/TR/activitystreams-vocabulary/).
- [Mastodon Docs](https://docs.joinmastodon.org/).

Does this look sane? Again, testing is much welcome :)


source/mode/activitypub.lisp
```diff
@@ -0,0 +1,823 @@
...
+   (instance-type
+    :mastodon
+    :type (maybe keyword)
+    :allocation :class
+    :documentation "The type of the instance -- Mastodon (implied by default), Plume, Lemmy, etc.
+Rendering may be different for different types.")
+   (profile
+    nil
+    :type (maybe actor)
+    :allocation :class
```
Ambrevar, line 73: Why all these `:allocation :class`?  Can there be only one mode instance?

source/mode/activitypub.lisp
```diff
@@ -0,0 +1,823 @@
...
+                :prompt (format nil "Your authorization token for ~a" instance)
+                :invisible-input-p t
+                :sources (list (make-instance 'prompter:raw-source)))))
+  "Log into the ActivityPub on a chosen instance under a chosen nickname."
+  (let ((instance-url (nyxt::ensure-url
+                       ;; Do we have any helper functions for this?
+                       (if (or (str:starts-with-p "https://" instance)
+                               (str:starts-with-p "http://" instance))
+                           instance
+                           (str:concat "https://" instance)))))
```
Ambrevar, line 96: Not that I know of.  `ensure-https` would be nice to have indeed.

Ambrevar: You can add a function on `urls.lisp`.

source/mode/activitypub.lisp
```diff
@@ -0,0 +1,823 @@
...
+  (:documentation "The base class for all the ActivityStreams types."))
+
+;; TODO: Wipe/clean it periodically to allow feed updates?
+(defvar *url->object* (make-hash-table :test 'equal)
+  "A memoization table from URL string to the fetched objects.")
+
+(export-always 'fetch-object)
+(defgeneric fetch-object (object)
+  (:method ((object list))
+    (lpara:pmapcar #'fetch-object object))
```
Ambrevar, line 122: Nice idiom! :)

source/mode/activitypub.lisp
```diff
@@ -0,0 +1,823 @@
...
+    :type (maybe quri:uri)
+    :allocation :class
+    :documentation "The URL of the home instance for the current account.")
+   (instance-type
+    :mastodon
+    :type (maybe keyword)
+    :allocation :class
+    :documentation "The type of the instance -- Mastodon (implied by default), Plume, Lemmy, etc.
+Rendering may be different for different types.")
+   (profile
```
Ambrevar, line 70: This `profile` slot may be confusing, because we already have `nfiles` profiles.  Can we pick a different term?

Ambrevar:
Wow, this is MASSIVE!  Your productivity keeps astounding me, Artyom! :D 

I'll review this more in depth later.  Maybe we keep this for after 3.0?

source/mode/activitypub.lisp
aartaka: It's inconvenient to log into your account anew in every buffer. Thus this allocation hack -- you log into your account in one buffer, and all the other buffers have it for free!

aartaka:
> Wow, this is MASSIVE! Your productivity keeps astounding me, Artyom! :D

Yeah, I'm surprised at myself at times too :D

> I'll review this more in depth later. Maybe we keep this for after 3.0?

We're not in hurry -- the planned deadline for this feature is autumn, as far as I remember :)

source/mode/activitypub.lisp
Ambrevar: Let's go with `account` then!

source/mode/activitypub.lisp
Ambrevar: Understood, but then it's not possible to have multiple connections, e.g. when using nosave buffers.

Instead I suggest we store multiple connections in a global variable.  For instance a hash-table indexed by profile (the Nyxt one, not the ActibityPub one) could work.

aartaka:
@Ambrevar, I cannot make much sense out of [Mactodon security documentation](https://docs.joinmastodon.org/spec/security/) and [http-signature standard](https://github.com/w3c-ccg/http-signatures), as both of those seem to omit lots of practical details. Things that I, in particular, cannot untangle, are:
- What's the signing algorithm Mastodon uses? The JSON that it sends mentions ed25519 encryption, while server code seems to rely on [OpenSS and RSA](https://github.com/mastodon/mastodon/blob/main/app/controllers/concerns/signature_verification.rb). I'm lost ｜−・;）
- How do we sign messages with _public_ keys using Ironclad? All of `ironclad:sign-message` methods specialize against private keys, and it kinda makes sense... But Mastodon docs say that _one needs to sign the message with the public key_. What the Hell (>_<) 

Not requiring you to dig into that instead of myself (I'll continue bumping my head agains the scarse Mastodon docs), but I'll appreciate it if you have any pointers on those :)

Ambrevar:
Will check it out when I'm done with #2536.

aartaka:
If the inability to post content is the outside limitation that we have to live with, then my plan is:
- Turn as much methods of `nyxt/activitypub-mode` into the ones dispatching over the instance-type, so that we can add instance-specific hacks for posting/fetching content.
- Abstract all the JSON-LD and ActivityStreams part of `nyxt/activitypub-mode` into a separate library (`nactivitypub`?)
  - Look at the JSON-LD standard and parse it a bit more rigorously.
- Polish the display for things.
- ... and then we can merge, I guess?

Ambrevar:
Agreed with all points!

The readme of the nactivitypub library (better name?) should include a disclaimer on the limitations.

source/mode/activitypub.lisp
```diff
@@ -0,0 +1,373 @@
...
+          (account mode) (ap:fetch-object (sera:lret ((url (quri:copy-uri instance-url)))
+                                            (setf (quri:uri-path url) (str:concat "/users/" nickname)))
+                                          :auth-token (auth-token mode)))))
+
+(defun http->ap (url)
+  (or (alex:when-let* ((url url)
+                       (valid (valid-url-p url))
+                       (url (quri:uri url)))
+        (str:concat "ap:" (nyxt::schemeless-url url)))
+      url))
```
Ambrevar, line 97: @aartaka What's the point of the `ap:` scheme if it's all over HTTP in the end?
Or did I misunderstand?

aartaka: It's over HTTP, but it's a different type of content (`application/json+ld`) with its own rules of display (`object->html`), that may reside on the same endpoint as regular content and is only served when certain headers (`Accept`) are present.

In other words, my Mastodon profile (https://ephemeral.glitch.social/@aartaka) may be:
- Viewed in browser, when making an
  - HTTP(S) request
  - With `Accept: text/html` header.
- But it can also be viewed in ActivityPub-supporting client, when making an
  - HTTP(S) request
  - With `Accept: application/ld+json ...` header.

Those two are totally separate and thus need a scheme separation.

Ambrevar: OK, makes sense.

aartaka:
So, shall we rebase and merge this now that it's more or less stable functionality-wise? I'd enjoy having NJSON used all over Nyxt, but don't want to interfere with this PR.

Ambrevar:
> I'd enjoy having NJSON used all over Nyxt, but don't want to interfere with this PR.

What do you mean?

As for activitypub support, maybe this belongs more to an extension than Nyxt's core?

aartaka:
Yes, an extension it could be.

Ambrevar:
Sure!

aadcg:
Now that we have a [repository for nactivitypub](https://github.com/atlas-engineer/nactivitypub), does it still make sense to keep this PR open?

aartaka:
I still have to extract this PR mode into a separate extensions :sweat:

jmercouris:
Please reopen when ready.

PR #2239, opened 2022-05-10
configuration: Fix spelling mistake.

edgar-vincent:


aartaka:
Let it be this way :) Thanks, @edgar-vincent!

PR #2240, opened 2022-05-10
web-extensions-callbacks: Fix trigger-message.

aadcg:
nyxt::handler-buffer is deprecated in favour of hooks:handler.

I don't know much about what this part of the codebase. I just noticed it was using hooks/handlers wrongly.

@aartaka, I believe you could say if everything's working as expected.


aartaka:
Yeah, the WebExtensions support code went somewhat untouched through all the recent changes xD

Issue #2244, opened 2022-05-11
Revert the documentation to the `default-configuration` examples? (and explain the customize-instance power)


aartaka:
Now that #2229 is merged, our good old `define-configuration` is back—stronger than it ever was! How about we actually return all the documentation to the pre-#2194 state, with `define-configuration` as _the_ way to configure Nyxt entities?

`customize-instance` still has a place here, but it's much less featurefull than `define-configuration` and should only be advised for those favoring the OO-style of configuration.

Ambrevar:
I disagree that it's less featureful: it's the opposite, it allows for more than just configuring slots!
`define-configuration` is arguably more convenient for slot configuration.

In my opinion, both have their place and should be documented.  I suggest we document define-configuration first, since it's easier, then proceed with `customize-instance` and more sophisticated examples.

aartaka:
Yes, this course of action makes perfect sense!

PR #2245, opened 2022-05-11
Revert docs to define-configuration examples

aartaka:
This completes #2244. While separate `customize-instance` examples should ideally be here too, I don't have use-cases that are sophisticated enough for these (」゜ロ゜)」


aartaka:
Fixed! Merging?

jmercouris:
Appears sound!

Issue #2241, opened 2022-05-10
`nhooks` and `nfiles` nicknames


aadcg:
We have a nickname for the former, `hooks`, but not for the latter.

For consistency, we could go either way (drop the `hooks` nickname or add the `files` nickname). WDYT?

jmercouris:
I would add the files nickname. I believe it offers more flexibility. It'll be less of a change should the name of the underlying libraries change for any reason.

aartaka:
+1 for `files`! 

I'm against removing `hooks` nickname, as it's there for backwards-compatibility...

Ambrevar:
But we are releasing a new major version, so we can always break it!

No strong opinion on my end.

aartaka:
> But we are releasing a new major version, so we can always break it!

I mean, we can, but why break what works fine :D The less changes one needs to make to their config, the better (and I say that as a person with ~1K lines of config -- a huge area for breakage :P)

PR #2246, opened 2022-05-11
Add files as a nickname to the nfiles library.

aadcg:
In accordance to how the nhook library is used.

Closes issue #2241.


aartaka:
Have you tried loading it? Seems like the nickname is not accessible in all of the packages `nfiles` is used in... You need to add `(use-nyxt-package-nicknames)` there then

aadcg:
I did that, but I missed the tests folders!

Please check my latest 2 commit, especially b1371efb. Notice that I had to include the `in-package` declaration otherwise `use-nyxt-package-nicknames` wouldn't be defined. Does this make sense at all? Thanks.

aadcg:
Looks good to me. Can I merge? @aartaka @Ambrevar

aartaka:
Yes, looks just fine to me! Resolve the conflicts (I can do that if you don't feel like arguing with Git) and we can merge :)

aadcg:
I didn't argue with Git much :p 

Should be ready to merge!

Recently I discovered `smerge-mode.el`, quite helpful!

PR #2250, opened 2022-05-13
Ensure prompter is destroyed after it's selection is returned.

Gavinok:
This is important for use in `query-hints` since without ensuring the prompt is
destroyed the hints will remain on the screen. This is only an issue when the
hints are used to select a input window since this does not refresh the buffer
and therefor will never remove the hints.

I tried to test this against the multiple selection but none the only ones I tried didn't work complaining about the argument number being given to a lambda. If anyone knows of a working command that supports multiple selection against I'd be happy to test it.

### DIFF
--- a/libraries/prompter/prompter.lisp
+++ b/libraries/prompter/prompter.lisp
@@ -399,10 +399,10 @@ instead."
     (setf action #'identity))
   (setf (returned-p prompter) t)
   (add-input-to-history prompter)
-  (alex:when-let ((selection-values (resolve-selection prompter)))
+  (alex:if-let ((selection-values (resolve-selection prompter)))
     (let ((action-result (funcall action selection-values)))
-      (calispel:! (result-channel prompter) action-result)))
-  (destroy prompter))
+      (calispel:! (result-channel prompter) action-result))
+    (destroy prompter)))
 
 (export-always 'toggle-follow)
 (defun toggle-follow (prompter &optional (source (selected-source prompter)))
--- a/libraries/prompter/prompter.lisp
+++ b/libraries/prompter/prompter.lisp
@@ -399,10 +399,10 @@ instead."
     (setf action #'identity))
   (setf (returned-p prompter) t)
   (add-input-to-history prompter)
-  (alex:if-let ((selection-values (resolve-selection prompter)))
+  (alex:when-let ((selection-values (resolve-selection prompter)))
     (let ((action-result (funcall action selection-values)))
-      (calispel:! (result-channel prompter) action-result))
-    (destroy prompter)))
+      (calispel:! (result-channel prompter) action-result)))
+  (destroy prompter))
 
 (export-always 'toggle-follow)
 (defun toggle-follow (prompter &optional (source (selected-source prompter)))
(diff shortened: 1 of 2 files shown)


Gavinok:
Before the hits would still be there
new-bad-nyxt


After the change they go away
new-better-nyxt-

jmercouris:
I tested with multiple selection for link hints. It appears to work for me. If you post a stack trace, perhaps we can help! It also resolves the problem you posted, about, so yay!

I see this change basically results in the destruction of the prompter even if there is or is not a resolved selection. This makes sense to me. Perhaps there was another reason we did not want to destroy the prompt buffer? Speculation: maybe it has some logic tied into prompt buffer resuming that I don't quite understand?

@Ambrevar, any feedback?

Gavinok:
Hmm I can't actually seem to reproduce that issue. Chances are this had to do with one of the links I was selecting. If I can reproduce it I will open a new issue (assuming it doesn't have to do with this).

Gavinok:
Think I found the issue. Currently with this change if you run `follow-hint` select an input and then run `resume-prompt` the prompt will not properly be restored. Shouldn't be too hard to find.

Gavinok:
@jmercouris This should resolve the issue `resume-prompt` will now function properly while removing the hints.
The trade off is the fact that these hints will still not be restored.

1. Would it be best to prevent hint commands from being resumed?
2. Should we simply create new hints?
3. Should we attempt to reproduce the exact same hints in the same buffer at the same url?

Ambrevar:
Maybe @aadcg knows better what's going on since he pushed the change.

aadcg:
There at least two topics being discussed here.

1. Should prompt buffer be resumed in this context?
2. Making sure the hints are gone either if `prompt-buffer` is canceled or returns.

```lisp
(prompt
 :prompt "test"
 :extra-modes (list (resolve-symbol :element-hint-mode :mode))
 :auto-return-p (auto-follow-hints-p (find-submode 'web-mode))
 :history nil
 :sources (make-instance 'hint-source
  :multi-selection-p nil
  :constructor (lambda (source)
                 (declare (ignore source))
                 (add-element-hints :selector (alex:if-let ((mode (find-submode 'web-mode)))
                                                (hints-selector mode)
                                                "a, button, input, textarea, details, select, img:not([alt=\"\"])"))))
 :after-destructor #'remove-element-hints)
```

Naively, I'd say that the call above should get rid of the element hints in both cases. In reality, it only gets rid of them when the prompt is cancelled (by pressing `esc`). 

What do you think @Ambrevar?

aadcg:
Also, wouldn't it be sensible that `prompt-buffer` would return `nil` rather than signal an error when cancelled?

Ambrevar:
> Also, wouldn't it be sensible that `prompt-buffer` would return `nil` rather than signal an error when cancelled?

No, because we want prompter cancellation to also cancel the command.  Semantically, it makes little sense to keep going if the user meant "get out".  This is the same behaviour in Emacs and in fact in most programs with a prompt.

aadcg:
Ok, I can make sense of that. 

Would it be a bad idea to have `:after-destructor-or-return`?

Ambrevar:
Can you specify what you want this new option for?  Example?

Gavinok:
Yes this resolves the issue. Thanks @Ambrevar

PR #2251, opened 2022-05-13
Change button CSS to better indicate what is and isn't clickable.

Gavinok:
Previously there was no hover attribute for any of the status buttons other than
the modes. This PR is a simple change which gives a consistent hover attribute
to each of the buttons on the default status buffer.

Before
bad-css

After
better-css


jmercouris:
Thank you!

Issue #2033, opened 2022-01-18
Feature Request: intelligently sorted sources


jmercouris:
Pedro pointed out to me recently that when using `set-url`, what may happen is you are given many duck duck go search suggestions. However, at the same time, you may have something in your history that perfectly matches what you've typed in.

The result is that you have to jump or arrow down to the next source to find what you are looking for. A potential solution to this problem may be a way to dynamically sort sources in a prompt buffer based on some heuristic. For example; source A is 95% confident it has a match, and source B is only 80% confident it has a match, therefore source A should go first.

We could limit this dynamic sorting to the FIRST source in the prompt buffer, OR sort all sources by confidence/the heuristic.

What do you think?
@Ambrevar

jmercouris:
@pdelfino

Ambrevar:
I've also been annoyed by this, but I'm a bit unsure about the proposed
solution: I'd rather have reliable source order than something I don't
control.

Allow me to propose something else.
Add an option to trim some sources: only display 5 suggestions for the new-url-query
source, then an ellipsis somewhere.  When the user reaches the last
suggestion or hit a special key, the rest of the source is expanded.

Does that make sense?

Another "fix" would be support for a vertical prompt buffer.

guojing0:
I also encountered this and gave it some thoughts as well: If we want to put more emphasis on the users' history, then "search query" should only be displaying when an user explicitly type `ddg stuffs`, `wiki stuffs`, or `bing stuffs`, and so on.

As for "dynamically sorting sources", I agree with Pierre: This way we could have stable behavior across all users and it may be easier for debug.

Ambrevar:
> I also encountered this and gave it some thoughts as well: If we want to put more emphasis
> on the users' history, then "search query" should only be displaying when an user
> explicitly type `ddg stuffs`, `wiki stuffs`, or `bing stuffs`, and so
> on.

This was the previous default behaviour and you can turn it back on by
setting search-always-auto-complete-p to nil.

> As for "dynamically sorting sources", I agree with Pierre: This way we could have stable
> behavior across all users and it may be easier for debug.

My point here was not only debugging but also muscle memory.  Users
quickly get used to pressing M-] to go to the next source and expect
some specific source there.  If the source order changes, this kind of
habits quickly get in your way.

pdelfino:
> Allow me to propose something else. Add an option to trim some sources: only display 5 suggestions for the new-url-query source, then an ellipsis somewhere. 

[...]

Issue #1421, opened 2021-05-18
Rethinking `execute-extended-command` and command specialization


Ambrevar:
Currently we have a combinatorial explosion between command "domain" and the action we take on it.

For instance, when "set-url" there are at least 3 possible actions: 
- in the current buffer
- in a new buffer
- in a new window.

For many other commands (e.g. `set-url-from-bookmark`), the same actions make
sense.

Defining a new command for each action is too painful and too hard to maintain.

Instead, there are a couple of things we could do:

1. Use the practice of adding an `actions` keyword argument to commands, so that
users can easily bind a key to `make-command FOO :actions MY-ACTIONS`.
I believe this is the least we can do.  It makes it very easy to fine-tune the
behaviour of every prompt-buffer-using command.

2. Extend `execute-extended-command` so that instead of listing commands and
_then_ prompting the user for arguments, it would list the command with all
_combinations_ of arguments.
For instance, if `set-url` takes a `in-new-buffer-p` optional argument, then
`execute-extended-command` would list two commands:

```
COMMAND        SPECIALIZATION
-----------------------------------------
set-url        (in-new-buffer-p nil)
set-url        (in-new-buffer-p t)
```

It's obvious for boolean arguments, but what about other types?
We could add methods that would derive the list of options!
For instance, we could introduce a "load-url-action" type and have a method that
would return a list of 3 actions: "in-current-buffer", "in-new-buffer", "in-new-window".

Thus, for a `set-url` command that takes an optional `load-url-action` argument,
`execute-extended-command` would return

```
COMMAND        SPECIALIZATION
-----------------------------------------
set-url        (load-url-action in-current-buffer)
set-url        (load-url-action in-new-buffer
set-url        (load-url-action in-new-window)
```

EDIT: Typos.

Ambrevar:
See also #1882.  Pinging @aartaka in case you didn't see this issue.

aadcg:
Is this so critical for 3.0? Looks like the perfect cocktail to shoot us in the foot. We have more pressing issues.

Ambrevar:
Agreed that this can be deferred to 4.0.

aadcg:
`execute-extended-command` was deleted in 425e728350e8aa63eaab50d6879c43f62fa21e25.



> 1. Use the practice of adding an `actions` keyword argument to commands, so that
>        users can easily bind a key to `make-command FOO :actions MY-ACTIONS`.
[...]

PR #2253, opened 2022-05-13
WIP: Split web mode

Ambrevar:
Fixes#1851.

This turns `web-mode` into `document-mode` and moves some parts to their independent modes, to make `document-mode` more focused and lightweight.

`document-mode` new definition: basic operations on structured documents, like HTML documents.  In particular, `document-mode` does not assume to be online.

Roadmap:

- document-mode.  Merge the following:

  - copy/paste/ring management
  - undo/redo
  - scroll
  - zoom
  - jump-to-heading

- spell-check-mode

- [x] history-mode

- [x] hint-mode.
  
  Subclassed by:
  - [x] search-buffer-mode
  - visual
  - element-frame (???)

Nyxt geeky modes:

- summarize-mode

- autofill-mode

- bookmarklets-mode

## To do:

- [ ] Move `focus-first-input-field` to `element-hint.lisp`?

## Questions:

- `bookmarklets-mode` is odd, but I don't know where to put it.  Shouldn't we move it to `source/bookmarklets.lisp`, in its own package (we already have `nyxt/dom` after all).

- `nyxt/element-hint-mode` is verbose, why not name it just `nyxt/hint-mode`?


Ambrevar:
Indeed, I'd like to move bookmarks as a mode too.  They are mostly independent after all.

aadcg:
@Ambrevar this is not an easy task, thank you for working on this! 

I don't much to add sadly, other than saying that I'm not convinced about renaming `element-hint-mode` to `hint-mode`.

Ambrevar:
> I'm not convinced about renaming `element-hint-mode` to `hint-mode`.

Fair enough, but could you defend this point? :)

jmercouris:
There is no reason `annotate.lisp` could not be a mode. Feel free to `modeify` it. :-)

Ambrevar:
Will do!

aadcg:
> Fair enough, but could you defend this point? :)

Sure!  We have other "hinting" concepts that specify `element-hint`,
such as `bookmark-hint` or `search-hint`.

Perhaps you're arguing that `element-hint` is verbose and that by
shortening it to `hint` we're not being less clear since it'd obvious
that `*-hint` "descend" from `hint`.

After a closer look at the codebase, we're indeed not very consistent.
For example, we have the functions `generate-hints` and
`add-element-hints`.  

So, you know what?  I agree with you!  To be clear, in the example
above, I'd rename `add-element-hints` to `add-hints`.

Is it what you had in mind?

Ambrevar:
Yes!

aadcg:
Ok, I can do this change. Or should I open an issue first to get some opinions first?

jmercouris:
No need to open an issue, I've read and I also agree.

Ambrevar:
@aadcg Don't worry, I'll do it as part of this pull request, since it request lots of moving around anyways.

Ambrevar:
This is a huge refactor, but it's mostly trivial code.
Looks at the individual commits if you want to make sense of it.

The core has been trimmed by some 600+ lines of code.  (They are not gone,
they've just been moved to `mode/`).

Peeps, can you test the following:

- Record-input-field (@pdelfino)
- Annotations (@jmercouris)

Remaining questions:

- Should the download API really be a mode?  I'd day no, at least for 3.0.
- OK with jump-to-heading merged into web-mode / document-mode?
- OK with the rename of web-mode to document-mode?  (that was the name in early
  versions of Nyxt)
- What is element-frame (please share example test) and what should it belong to?
- Where to put bookmarklets?

aadcg:
> OK with jump-to-heading merged into web-mode / document-mode?

Yes.

>  OK with the rename of web-mode to document-mode? (that was the name in early
versions of Nyxt)

Yes.

> What is element-frame (please share example test) and what should it belong to?

My understanding is that it's the engine behind `expedition-mode` and could potentially spark new ideas. I'd say it's part of `document-mode`.

Ambrevar:
Yes, that sounds about right.  `select-frame-new-buffer` is not used anywhere though.  @jmercouris ?

aadcg:
> This is a huge refactor, but it's mostly trivial code.
Look at the individual commits if you want to make sense of it.

I went through it, and it looks good to me!

jmercouris:
I went through it as well. I found the changes to download mode, then I read your comment within the pull request. I'm not exactly convinced. It is more so about the behavior of the download buffer, that is what download mode is for. Like how we have a repl-mode for the REPL buffer.

I can imagine that if we add more tweakable things to the download mode like: "automatically-hide-completed-downloads-p", it would not be renderer specific behavior!

That said, this can come at a later time, and is perhaps a premature optimization within my mind!

Thank you for this PR Pierre, it resolves many things, and does many changes I am strongly happy about! :-)

jmercouris:
>     * Annotations (@jmercouris)

Yes.

> Remaining questions:
> 
>     * Should the download API really be a mode?  I'd day no, at least for 3.0.

I still think it should be, I've provided my rationale above!

>     * OK with jump-to-heading merged into web-mode / document-mode?

I don't see why it should be so, but it does not bother me.
 
>     * OK with the rename of web-mode to document-mode?  (that was the name in early
>       versions of Nyxt)

I am happy with this.

>     * What is element-frame (please share example test) and what should it belong to?

Element frame and element hint are not exactly modes. They are more so mechanisms of selecting URLs from a page. In my head, this doesn't exactly make them modes. Now, element hint within the context of using it to jump to URLs, this /IS/ a mode in my head.

>     * Where to put bookmarklets?

I see you've placed it in document-mode, again, it could also be a mode.

jmercouris:
I can confirm that annotations are broken :-)

missing symbols, `keyword-source` for example which is defined in bookmark.lisp. Not sure in which parent file I would put `keyword-source`. Probably next to `nyxt:keywords`.

Ambrevar:
My bad, I had seen this problem but forgot to move the `keyword-source` definition later.
Fixed in 59904581f19b7dfa73b4d8860e20fd1aa9714e02.

Ambrevar:
> > ```
> > * Where to put bookmarklets?
> > ```
> 
> I see you've placed it in document-mode, again, it could also be a mode.

I didn't move it, it was always in web-mode (I assume because it only works on web pages).  It's now in document-mode only because I've renamed web-mode to document-mode.

If we take a step back, bookmarklets are just a JavaScript interface and they don't interact with structured documents generally speaking (e.g. bookmarklets would make no sense for Org documents).

Thus I agree that we should move them to a separate mode.  Will do.

Ambrevar:
> > ```
> > * OK with jump-to-heading merged into web-mode / document-mode?
> > ```
> 
> 
> I don't see why it should be so, but it does not bother me.

Because it's a form of interaction with _structured documents_.  Would `jump-to-heeading` make sense on Org documents?  Yes, so it belongs to `document-mode`.

Ambrevar:
About download-mode: I agree that we may need a mode at some point.  But currently, `download-mode` only existed to "make the buffer special" (like annotate and bookmark used to) and did nothing in particular.  In other words, it was a useless mode.

Also `mode/download.lisp` was the only file not in its own package, which is inconsistent.

We would need a "download-mode" at some point, but before we can even start thinking about it, we need to fully rewrite our download manager and rethink the API with regard to the renderers.  This is a huge endeavour, so I suggest we keep this for 4.0 or later.

Ambrevar:
In 840036e8cbbe11aed3c87d50707a012f1f879546 I've moved bookmarklets to their own mode.  

This mode is _not enabled_ by default.  This means that bookmarklets are no longer visible from `execute-command` by default, unless the user enables `bookmarklets-mode` of course.

I did this to create a clearer separation between Nyxt default commands and bookmarklets, but of course I'm happy to hear counter-arguments.

Ambrevar:
It's ready to merge!

Ambrevar:
One last question: looks at what I did in 817556b09d972a3d62663d7584004c59e2dd44e4, I'm allow for modes to modify the sources of the prompt in the `set-url` commands.

- Should we use an `append` method combination instead?  (No much benefit in my opinion)
- Should we be more general and allow modes to set the final list of sources, instead of just appending?  (Thus the specialization would have to call `call-next-method` if they want to reuse the default result.)
- Can we somehow generalize this to all commands?  Would be great, but that would also mean that all command prompts must be listed somewhere (e.g. in the `command' class slot).  Remember that a  command may have multiple prompts.  How to we declare that?  What would the syntax look like?

It'd be a great feature to have, it's just very unclear to me how to implement it.

Ambrevar:
(In any case, we can address this issue after merging this.)

Ambrevar:
The question of making prompt-buffer sources (and actions?) extensible (mode-extensible?) has come up in the past, but I can't find the exact issue right now.

Possibly related issues: #1421, #2033.

source/mode/bookmark.lisp
```diff
@@ -78,3 +128,325 @@
...
+                (string= "about:blank" (render-url url)))
+      (multiple-value-bind (entries bookmarks-without-url)
+          (sera:partition (sera:partial #'url-equal url) bookmarks :key #'url)
+        (let ((entry (if entries
+                         (first entries)
+                         (make-instance 'bookmark-entry
+                                        :url url))))
+          (unless (str:emptyp title)
+            (setf (title entry) title))
+          (setf tags (delete "" tags :test #'string=))
```
aartaka, line 178: Merge into one one `setf`.

source/mode/bookmark.lisp
```diff
@@ -78,3 +128,325 @@
...
+(define-panel-global bookmarks ()
+    (panel-buffer "*Bookmarks panel*")
+  "Shows all the bookmarks in a compact panel-buffer layout."
+  (spinneret:with-html-string
+    (:style (cl-css:css
+             '((p
+                :font-size "12px"
+                :margin "0"
+                :white-space "nowrap"
+                :overflow-x "hidden"
```
aartaka, line 232: Oh wow, I should probably utilize that in inspector.lisp!

source/mode/bookmark.lisp
```diff
@@ -78,3 +128,325 @@
...
+                   :sources (list
+                             (make-instance 'prompter:word-source
+                                            :name "New tags"
+                                            ;; On no input, suggest the empty tag which effectively acts as "no tag".
+                                            ;; Without it, we would be force to specify a tag.
+                                            :filter-postprocessor
+                                            (lambda (suggestions source input)
+                                              (declare (ignore source input))
+                                              (or suggestions
+                                                  (list "")))
```
aartaka, line 269: Allow null list of tags instead?

source/mode/bookmark.lisp
```diff
@@ -78,3 +128,325 @@
...
+
+(define-command bookmark-buffer-url ()
+  "Bookmark the currently opened page(s) in the active buffer."
+  (prompt
+   :prompt "Bookmark URL from buffer(s)"
+   :sources (make-instance 'buffer-source
+                           :multi-selection-p t
+                           :actions (list (lambda-mapped-command bookmark-current-url)))))
+
+(define-command bookmark-url (&key url)
```
aartaka, line 288: Use the prompt as the default value of `url`?

source/mode/bookmark.lisp
```diff
@@ -78,3 +128,325 @@
...
+                      :sources (make-instance 'bookmark-source
+                                              :multi-selection-p t))))
+        (delete-bookmark entries))))
+
+(define-command set-url-from-bookmark
+    (&key (actions (list (lambda-command buffer-load* (suggestion-values)
+                           "Load first selected bookmark in current buffer and the rest in new buffer(s)."
+                           (mapc (lambda (url) (make-buffer :url (url url))) (rest suggestion-values))
+                           (buffer-load (url (first suggestion-values))))
+                         (lambda-command new-buffer-load (suggestion-values)
```
aartaka, line 338: Add `copy-url*` too!

aartaka: And `delete-bookmark`, maybe!

Or, should we have `bookmark-action` akin to `url-actions`?

source/mode/bookmark.lisp
```diff
@@ -78,3 +128,325 @@
...
+                (setf (getf entry :date)
+                      (local-time:parse-timestring (getf entry :date))))
+              (apply #'make-instance 'bookmark-entry
+                     entry))
+            entries)))
+
+(define-command import-bookmarks-from-html (&key (html-file nil))
+  "Import bookmarks from an HTML-FILE."
+  (let ((html-file (or html-file
+                       (prompt1
```
aartaka, line 412: Move the prompt to the argument default value? `(or html-file (prompt1 ...))` is exactly what default values are about.

source/mode/hint.lisp
```diff
@@ -1,17 +1,87 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
-;;;; SPDX-License-Identifier: BSD-3-Clause
-
-(in-package :nyxt/web-mode)
+(uiop:define-package :nyxt/hint-mode ; TODO: Rename to `hint-mode'?
```
aartaka, line 4: Outdated TODO comment?

source/mode/no-procrastinate.lisp
```diff
@@ -83,3 +94,224 @@
...
+          (sera:partition (sera:partial #'url-equal url) no-procrastinate-hosts-list :key #'url)
+        (let ((entry (if entries
+                         (first entries)
+                         (make-instance 'no-procrastinate-entry
+                                        :url url))))
+          (unless (str:emptyp title)
+            (setf (title entry) title))
+          (unless (str:emptyp hostname)
+            (setf (hostname entry) hostname))
+          (setf tags (delete "" tags :test #'string=))
```
aartaka, line 149: I wonder if we should someday add our own preconfigured versions of typical sequence primitives with changed equality/keys. Like `memq` and `memv` in Scheme.

aartaka: `cl-str` has `str:s-member` that our `cl-readability` duplicates with some alterations...

source/mode/record-input-field.lisp
```diff
@@ -1,7 +1,25 @@
...
+(uiop:define-package :nyxt/record-input-field-mode
+  (:use :common-lisp :nyxt)
+  (:import-from #:class-star #:define-class)
+  (:import-from #:keymap #:define-key #:define-scheme)
+  (:import-from #:serapeum #:export-always #:->)
+  (:documentation "Record input fields to be refilled later."))
+(in-package :nyxt/record-input-field-mode)
+(use-nyxt-package-nicknames)
+
+(define-mode record-input-field-mode ()
```
aartaka, line 13: Now that this is a mode, we can do better, monitoring every input field and asking the user whther they want to save it. Open an issue about that?

source/mode/search-buffer.lisp
```diff
@@ -1,7 +1,35 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-(in-package :nyxt/web-mode)
+(uiop:define-package :nyxt/search-buffer-mode
```
aartaka, line 4: Simply `search-mode`? But then the current implementation requires a JS-aware buffer, so `search-buffer-mode` may be better...

source/mode/spell-check.lisp
```diff
@@ -1,7 +1,30 @@
...
+  ""
+  ((rememberable-p t)
+   (spell-check-language
+    "en_US"
+    :documentation "Spell check language used by Nyxt. For
+a list of more languages available, see `enchant:broker-list-dicts'.")
+   (keymap-scheme
+    (define-scheme "visual"
+      scheme:cua
+      (list)                            ; TODO: Add CUA bindings!
```
aartaka, line 21: This is done by right mouse button in CUA, isn't it?

source/mode/document.lisp
```diff
@@ -0,0 +1,527 @@
...
+
+;; TODO: Remove document-mode from special buffers (e.g. help).
+;; This is required because special buffers cannot be part of a history (and it breaks it).
+;; Bind C-l to set-url-new-buffer?  Wait: What if we click on a link?  url
+;; changes in special buffers should open a new one.
+;; Or else we require that all special-buffer-generating commands open a new buffer.
+
+(define-mode document-mode ()
+  "Base mode for interacting with documents."
+  ((rememberable-p nil)
```
aartaka, line 24: I wonder whether it should become rememberable now that it's basically about navigating structured documents. Could one enable it on some weird page to have navigation? Should we allow it? Do we still imply it by default?

source/mode/document.lisp
```diff
@@ -0,0 +1,527 @@
...
+                           new-path-splited
+                           :initial-value "/")))
+    (buffer-load (str:concat scheme "://" authority new-path))))
+
+
+(define-command paste (&optional (buffer (current-buffer)))
+  "Paste from clipboard into active element."
+  (ffi-buffer-paste buffer))
+
+(define-class ring-source (prompter:source)
```
aartaka, line 179: Is it actually used? We don't fill it anywhere, I suppose. Maybe make an `:after` method for `ffi-buffer-copy` and `ffi-buffer-paste` and add `trivial-clipboard:text` to the ring?

source/mode/document.lisp
```diff
@@ -0,0 +1,527 @@
...
+(defmethod nyxt:on-signal-load-committed ((mode document-mode) url)
+  (declare (ignore mode url))
+  nil)
+
+(defmethod nyxt:on-signal-load-finished ((mode document-mode) url)
+  (reset-page-zoom :buffer (buffer mode)
+                   :ratio (current-zoom-ratio (buffer mode)))
+  url)
+
+;; REVIEW: Shorten the name to e.g., `show-url-qr'? It's no longer current URL only.
```
aartaka, line 280: Yes!

aartaka (commented):
I've gotten terribly lost around the code that's moved and deleted, but it looks fine to me! I've left lots of (quite insignificant) comments, but it's already good to merge as it is!

source/buffer.lisp
```diff
@@ -380,28 +343,38 @@ Password interfaces are configurable through a `customize-instance' method.")
     :type symbol
     :documentation "Select a download engine to use,
 such as :lisp or :renderer.")
+   (download-style
```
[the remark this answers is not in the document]
jmercouris: This was exactly my first reaction as well, hence my comment!

source/buffer.lisp
```diff
@@ -1659,6 +1685,6 @@ That is, the one with the most recent access time."
   (ffi-inspector-show (current-buffer)))
 
 (export-always 'print-buffer)
-(define-command print-buffer ()
+(define-command print-buffer ()         ; TODO: Move to `document-mode'?
```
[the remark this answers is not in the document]
jmercouris: Why? All buffers are printable, are they not?

source/mode/search-buffer.lisp
Ambrevar: "search" is ambiguous, because we can search buffer objects, windows, history, etc.

source/download.lisp
```diff
@@ -105,47 +105,13 @@ appearance in the buffer when they are setf'd."
...
-
 (define-internal-page-command-global list-downloads ()
-    (buffer "*Downloads*" 'download-mode)
+    (buffer "*Downloads*" 'base-mode)
   "Display a buffer listing all downloads.
 We iterate through the browser's downloads to draw every single
 download."
   (spinneret:with-html-string
-    (:style (style (find-submode 'download-mode)))
+    (:style (style (download-style buffer)))
```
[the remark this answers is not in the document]
Ambrevar: Good point.  So download-mode would have just the style?  I guess that's fine.

source/buffer.lisp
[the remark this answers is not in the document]
Ambrevar: No, buffers are now much more abstract than they used to.  `buffer` is a dummy object that does not hold a document!  Also printing the message area or the status area makes little sense.

source/help.lisp
```diff
@@ -322,7 +322,7 @@ For generic functions, describe all the methods."
                            :prompt "Describe command"
                            :sources (make-instance 'command-source)))))
     (buffer (str:concat "*Help-" (symbol-name command) "*")
-            'nyxt/help-mode:help-mode)
+            'base-mode)
```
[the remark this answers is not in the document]
Ambrevar: We need to rework this part indeed.  Also how we uniquely identify "special buffers".  In another pull request!

source/mode/bookmark.lisp
Ambrevar: No need until we've generalized the concept.  For now, bookmark actions are only used in bookmark.lisp, so it's not going to be extended from somewhere else.

source/mode/record-input-field.lisp
Ambrevar: @pdelfino ?

source/mode/document.lisp
Ambrevar: I'm not sure either, I guess time will tell!

Ambrevar:
Thank you @aartaka I'll address all your points shortly.

source/mode/document.lisp
Ambrevar: Indeed, but our clipboard system is inconsistent: we have functions in source/clipboard, source/browser, source/renderer/gtk, source/mode/document.

Let's sort it out in another pull request.

source/mode/spell-check.lisp
Ambrevar: I guess so.  Who is feeling up to add a context menu action? :)

aartaka:
Yay! Thanks for this, Pierre!

jmercouris:
Thank you Pierre!!!

PR #2255, opened 2022-05-13
EXTENSIONS.org: Amends list based on extension split

efimerspan:
I recently split out the `nx-mapper` extension into two standalone extensions as I feel it's more cohesive to have them each be its own thing and tackle its own specific problems.

### DIFF
--- a/documents/EXTENSIONS.org
+++ b/documents/EXTENSIONS.org
@@ -26,9 +26,8 @@
   that you can use to test your Nyxt installation (i.e. can you load
   extensions).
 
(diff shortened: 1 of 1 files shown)


Ambrevar:
OK, thanks for the update!

This part of the sentence confuses me:

> is a tailor for all-things Nyxt themes 

Is this about themes?

efimerspan:
Yes, it currently allows to dynamically change Nyxt themes. I can change the wording if you' d like.

jmercouris:
I think the wording is OK :-)

PR #2263, opened 2022-05-17
Foreign interface cleanup

### LINKED ISSUE #2097: Bring more clarity and consistency into FFI methods
There are several problems with our FFI methods:

# Inconsistent getters/setters

Two methods, side-by side in browser.lisp, lines 755-757
``` lisp
(define-ffi-generic ffi-buffer-user-agent (buffer &optional value))
(define-ffi-generic ffi-buffer-set-proxy (buffer &optional proxy-url ignore-hosts))
(define-ffi-generic ffi-buffer-get-proxy (buffer))
```
`ffi-buffer-user-agent` relies on the optional value to undertand whehter to behave as getter/setter, while `ffi-buffer-(set|get)-proxy` splits getter and setter into separate methods. This is inconsistent. We need to adopt just one style for such a case of get/set methods. Possible styles are:
- Optional args (one function to write, easy to use).
- Separate getters/setters (clear and maps well onto C/C++ renderers, but a lot of boilerplate).
- `setf` methods (idiomatic and (almost) zero-cost, yet wordy).

# Naming

Names like `ffi-buffer-enable-javascript-markup` and `ffi-buffer-auto-load-image` are directly inherited from WebKitGTK [WebKitSettings](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitSettings.html), which is likely to get uncomfortable if we add another renderer. We should use uniform naming, like `ffi-buffer-javascript-markup` (without enable, simply the subject) or always use `-enable`. Matter of preference here (I prefer the former). This is also connected with getters/setters inconsistency.

# Lack of documentation

When one asks us where to look when implementing a new port, we usually point them at browser.lisp and at the list of methods therein. What we constantly forget pointing people at it is that there's almost no documentation for methods. We need to write clear documentation of what every method does, so that one trying to implement a port clearly understands what are the renderer-side analogues to what our methods do.

Thoughts?

Ambrevar:
Fixes #2097.

To do / questions:

- Define `nyxt/ffi` package and remove the `ffi-` prefix.
- Define a proxy class with URL and IGNORE-LIST slots.

Comments welcome! :)


Ambrevar:
See also https://github.com/joachifm/cl-webkit/issues/65.

aartaka:
> * Define a proxy class with URL and IGNORE-LIST slots.

What I meant was rather to expand `proxy` from browser.lisp...

jmercouris:
> Fixes #2097.
> 
> To do / questions:
> 
>     * Define `nyxt/ffi` package and remove the `ffi-` prefix.

As long as we don't `use` `nyxt/ffi` it could be OK!

>     * Define a proxy class with URL and IGNORE-LIST slots.

Yes!

jmercouris:
I went ahead and did a whole bunch of renames for you, I'm sure there are some mistakes, but maybe it will help, it was a lot of query select. If you want to start over, feel free to revert :-D

jmercouris:
You'll at least have to move the package definition, or move where `foreign-interface.lisp` is defined in the ASD.

Ambrevar:
Why? We can do like nyxt/dom, no need to move anything, or is there?

Ambrevar:
I've finished the move to the `ffi` package that John started.

However:
- The browser does not show any window anymore :p 
- It's not that pretty: it highlights how intertwine everything is.  To fix it, we would need to define the default methods of `nyxt/ffi` _after_ we are done loading the core.  Doable, but is it worth it?  Plus this would force us to split the specs in 2 parts, is it really what we want?

aartaka:
> * The browser does not show any window anymore :p

:D 

> * It's not that pretty: it highlights how intertwine everything is.  To fix it, we would need to define the default methods of `nyxt/ffi` _after_ we are done loading the core.  Doable, but is it worth it?  Plus this would force us to split the specs in 2 parts, is it really what we want?

Maybe put foreign-interface later in the ASD (no, it won't work, as `ffi-*` functions are used everywhere)? Or remove all the default methods? Or put default things in a dummy renderer of their own to preserve them? I mean, it's radical, but it is what it is -- we need every renderer to implement things their way, instead of implying, for example, that there's a JS engine and it behaves the way we want.

jmercouris:
Actually let's merge without making a new package and do it later?

Ambrevar:
Agreed!

Ambrevar:
I've reverted the nyxt/ffi package change and merged with f4b9788f2a3817c049fa2da323660ec02f9ccdc5.

PR #2265, opened 2022-05-18
WIP: build-scripts: Add report-unbound-exports.

Ambrevar:
We can leverage this code in the CI to detect exported symbols that are unbound.

This can typically happen when there is a typo in an `(export ...)` call.

## To do:

- Test.
- Call it from the CI.

## Questions:

Is there a portable way to detect if a symbol points to a type?  In CCL?
We can do it in SBCL, but it's an extension.

If there is no portable way, then we shall run this CI script only with SBCL.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -86,7 +86,7 @@ jobs:
     - name: Unbound exports
       shell: bash
       run: |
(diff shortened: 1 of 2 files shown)


Ambrevar:
Merged with 407512689eb43c39cf5d537c0588ec1e4dbc8041.

PR #2274, opened 2022-05-19
Export list-downloads to avoid breaking base-mode's bindings

Gavinok:
Currently when C-j is used an error is printed since the function list-downloads
is not exported.

### DIFF
--- a/source/mode/base.lisp
+++ b/source/mode/base.lisp
@@ -43,7 +43,7 @@ This mode is a good candidate to be passed to `make-buffer'."
        "f1 b" 'describe-bindings
        "f11" 'toggle-fullscreen
        "C-O" 'load-file
-       "C-j" 'list-downloads
+       "C-j" 'nyxt/download-mode:list-downloads
        "C-space" 'execute-command
        "C-M-space" 'execute-extended-command
        "M-space" 'resume-prompt
@@ -82,7 +82,7 @@ This mode is a good candidate to be passed to `make-buffer'."
        "C-h u s" 'universal-describe-slot
        "C-h k" 'describe-key
        "C-h b" 'describe-bindings
-       "C-d" 'list-downloads
+       "C-d" 'nyxt/download-mode:list-downloads
        "M-x" 'execute-command
        "C-M-x" 'execute-extended-command
        "M-1" (read-from-string "nyxt/repeat-mode:repeat-key")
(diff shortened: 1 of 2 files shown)


Ambrevar:
@aartaka Shouldn't define-internal-page-command-global automatically do this?

Gavinok:
@Ambrevar closest I can find is 
https://github.com/atlas-engineer/nyxt/blob/master/source/command.lisp#L158 
Which will simply export it as part of the current package in the end this ends up being

```lisp
(export-always 'list-downloads (symbol-package 'list-downloads))
```
with
```lisp
(symbol-package 'list-downloads) ; => #<PACKAGE "NYXT/DOWNLOAD-MODE">
```

The docstring also doesn't seem to indicate the intention to export this as part of the `nyxt` package by default. I could also be missing the point of your message to aartaka

Ambrevar:
I had misunderstood the problem statement.  The patch you've sent should not work, or does it?

Indeed, we don't want to export to the `nyxt` package, instead the references to the function should be updated.

In other words, `list-downloads` should be replaced by `nyxt/download-mode:list-downloads` in base.lisp.

Wanna update the patch?

Gavinok:
@Ambrevar binding should now be corrected.

Ambrevar:
Thanks, but as you can see from the checks, you got bitten by the load order :p 

Since base-mode is defined before download-mode, it does not know it yet and fails to load.

So the fix at this point would be to load base-mode after all the core mode.  But then we would need to forward-declare base-mode.

Thoughts on this, anyone?

@Gavinok  Hint: before sending a pull request, make sure to rebuild your Nyxt: if it fails or reports error, something is wrong ;)

Gavinok:
Ah sorry my mistake I'll fix that now

Gavinok:
Having trouble getting Nyxt to properly build on my system due to library conflicts so I will need to look into this and #2287 when I have time to sort that out. Otherwise feel free to supersede this PR if anyone able to build a local version has the time.

Thanks for your patience!

Ambrevar:
We can help you build Nyxt: open a new issue and describe your system and your problem as precisely as possible.

Ambrevar:
I'll fix this, it's a bit tricky indeed.

Ambrevar:
Done with 9d452bfbd19da755901454fa45f2ee0b3882a5fc.

Gavinok:
Thanks @Ambrevar didn't have time yesterday to look into building the project

PR #2275, opened 2022-05-20
Fixes for repl-mode completion and resetting of cursor

Gavinok:
Empty buffer completion works, and refresh restores cursor position.

Previously you would get an error when tab completion was called with an empty input. This patch resolves this by simply checking if the input is empty and if it is not state that no delimiter was found since no text comes before the cursor.

In addition, the cursor is now set to the input area after a refresh. Since refreshes are done every time the return key is hit this prevents the user from having to select the input field every time and expression is evaluated.

### DIFF
--- a/source/mode/repl.lisp
+++ b/source/mode/repl.lisp
@@ -156,8 +156,7 @@ Scroll history with `evaluation-history-previous' and `evaluation-history-next'.
   "Complete the current symbol and insert the completion into the REPL prompt."
   (let* ((input (input repl))
          (cursor (cursor repl))
-         (empty-input (= cursor 0))
-         (previous-delimiter (unless empty-input
+         (previous-delimiter (unless (= cursor 0)
                                (position-if (lambda (c) (member c '(#\( #\) #\space))) input
                                            :end (1- cursor) :from-end t)))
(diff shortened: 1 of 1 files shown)


source/mode/repl.lisp
```diff
@@ -156,8 +156,10 @@ Scroll history with `evaluation-history-previous' and `evaluation-history-next'.
   "Complete the current symbol and insert the completion into the REPL prompt."
   (let* ((input (input repl))
          (cursor (cursor repl))
-         (previous-delimiter (position-if (lambda (c) (member c '(#\( #\) #\space))) input
-                                          :end (1- cursor) :from-end t))
+         (empty-input (= cursor 0))
```
Ambrevar, line 159: Nit: `empty-input` is not the right term here, since the cursor may be at position 0 while the  input still contains something.

Simple fix: remove the `empty-input` let-binding and just call `(= cursor 0)` directly, it's clear enough!

Ambrevar:
Good double catch!  Can you address my nitpicky comment?  Then I'll merge this!

Gavinok:
@Ambrevar Should be good to go

Issue #2268, opened 2022-05-18
Why do some function take a buffer-id instead of buffer? | Allow non-literal data in nyxt: URLs


Ambrevar:
@aartaka I see that you've changed show-url-qrcode to take a buffer-id instead of a buffer.

Why is it so?  It makes it rather inconvenient / inconsistent to call it.

aartaka:
This is because `nyxt:` URLs only allow passing literal/constant/scalar data in them. We cannot pass whole buffers (well, we can, but there's no easy way for that yet) as URL query arguments, can we?

While this is a restriction of `nyxt:` URLs, and a grave one, I have an idea for how we can circumvent that. We need to pass object by-reference in internal page commands. To do that, we just need to check the data passed to the function for being non-constant, store it in `nyxt::*inspected-values*` (should we rather put it in debug.lisp? browser.lisp? it's too useful to stay inspector-only...) and, when accessing the values in `nyxt:` URLs somehow infer that the passed value is a reference/ID of an object in `nyxt::*inspected-values*`.

Ambrevar:
Oh yeah, that'd be much, much better! 

The current implementation forces inconsistent code, like buffer ID but also string URLs.

Can you send a patch?

aartaka:
Will do, once I have time :)

Ambrevar:
My proposal to fix this (I can send a patch if you don't have time, no worries) is to use the same strategy as cl-webkit calbacks: instead of encoding the form into the URL, just encode a reference to a unique (gensymed) callback stored in a global table.

In other words (untested):

```lisp
(defpsmacro lisp-eval (form &optional callback)
  "Request the lisp: URL and invoke callback when there's a successful result."
  (let ((request-id (gensym)))
    `(let ((request-lambda (lambda () ,@form)))
       (setf (gethash ,request-id *lisp-url-callbacks*) ,request-lambda)
       (let ((request (fetch (lisp (format nil "lisp://~a" ,request-id)) ; TODO: Maybe add purely informative string in the title, to make HTML inspection more useful?
                             (create :mode "no-cors"))))
         (when ,callback
           (chain request
                  (then (lambda (response)
                          (when (@ response ok)
                            (chain response (json)))))
                  (then ,callback)))))))


;...

(define-internal-scheme "lisp"
    (lambda (url buffer)
      (let ((url (quri:uri url)))
        ;; TODO: Replace this condition with `(network-buffer-p buffer)`?
[...]

PR #2276, opened 2022-05-20
help: Add button to remove method.

Ambrevar:
@aartaka I'm doing some inspector ID trick so that no `#<...>` value appears in the JS code.  Is this the right way to go?


aartaka:
Yes, IDs are a way until #2268 is fixed.

Ambrevar:
OK, merging with dbafbfaa83676147038c78fe530947708a2d605a then, we will fix the general issue in #2268.

PR #2277, opened 2022-05-20
Switch from strings to uninterned symbols for IDs.

### LINKED ISSUE #2098: Make buffers and nyxt-specific values have numeric IDs instead of strings
Numeric IDs have some advantages over strings:
- They are strings of integers now, which doesn't make much sense. If it's a number, let it be a number :D
- `eq` is enough to compare/search. `eql` at most. It's simpler and faster. With strings, it's `equal` at least.
- Numbers are easier to encode into `nyxt:` URLs as they have no quotes.
  - May it rather be the problem of `nyxt:` URLs not being smart enough at en/decoding?
- Other browsers (conforming to WebExtensions API) use numeric IDs for tabs/windows/etc. The reasons for this are obviously beyond this issue, but there should be some. Given that conformance to WebExtensions API is quite a goal, having numeric IDs is a minor yet vital thing, given how much these IDs are passed around.

Does that make sense?

Ambrevar:
Rationale:
- IDs are expected to be numeric.
- The initial intent with string IDs was to allow for _naming_ buffers /
windows, but:
  - We never do this as of now.
	- Should we ever feel the need, we should introduce an aptly named slot, for
	instance `name'.

This simplifies the code somewhat.

## Question

`gensym` (uninterned symbol) is the idiomatic way to manipulate IDs in Common Lisp.
But Nyxt URLs might still preferer integers instead of uninterned symbols.

@aartaka Thoughts?

## To test:

- Are history owners working properly?
- Are Nyxt URLs working properly?

Fixes #2098.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -15,10 +15,9 @@
 
 (define-class buffer (renderer-buffer)
   ((id
-    ""
-    :documentation "Unique identifier for a buffer.
-Dead buffers or placeholder buffers (i.e. those not associated with a web view)
-have an empty ID.")
+    (new-id)
+    :type symbol
+    :documentation "Unique identifier for a buffer.")
    ;; TODO: Or maybe a dead-buffer should just be a buffer history?
    (profile
     (alex:if-let ((profile-class (find-profile-class (getf *options* :profile))))
@@ -108,11 +107,9 @@ representation of HTML documents.
 Rendered URLs or the Nyxt's manual qualify as examples.  Buffers are fully
 separated from one another, so that each has its own behaviour and settings."))
 
-(defmethod initialize-instance :after ((buffer buffer) &key (browser *browser*)
+(defmethod initialize-instance :after ((buffer buffer) &key
                                        &allow-other-keys)
-  "Set buffer ID and return buffer."
-  (when browser
-    (setf (id buffer) (get-unique-identifier browser)))
+  "Dummy method to allow forwarding other key arguments."
   buffer)
 
 (defmethod finalize-buffer ((buffer buffer) &key (browser *browser*) &allow-other-keys)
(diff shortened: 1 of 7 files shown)


aartaka:
If we can print symbols for Nyxt URLs, this shouldn't cause much of a problem!

I'll review this today.

source/buffer.lisp
```diff
@@ -1043,7 +1038,6 @@ associated to the buffer is already killed."
                                                  :url (quri:uri "about:blank")))))
         (window-set-buffer parent-window replacement-buffer)))
     (buffers-delete (id buffer))
-    ;; (setf (id buffer) "") ; TODO: Reset ID?
```
aartaka, line 1046: So what do we do with deleted buffers then? SImply remove their IDs from the hash-table?

source/help.lisp
```diff
@@ -727,7 +727,7 @@ the channel, wrapped alongside the condition and its restarts."))
   (declare (ignore hook))
   (when *debug-on-error*
     (let* ((*debugger-hook* *old-debugger-hook*)
-           (id (parse-integer (get-unique-identifier *browser*)))
+           (id (parse-integer (symbol-name (new-id))))
```
aartaka, line 730: We can use symbols for conditions too, no need to parse those!

aartaka (approved):
All fine, just some blank spaces that I'm curious about :)

source/buffer.lisp
Ambrevar: Yup.

aartaka: Makes sense, yes.

source/help.lisp
Ambrevar: Done in dcf818e2ffca014873e35ff2b3e8fd2198c4afc8.

Ambrevar:
Merged with dcf818e2ffca014873e35ff2b3e8fd2198c4afc8.

aartaka:
Thanks, Pierre ＼（＠￣∇￣＠）／

PR #2279, opened 2022-05-20
Rename "init-*" to "config-*" for consistency.

### LINKED ISSUE #2074: Inconsistent `auto-init-data-path` type and `auto-config` file name.
I think the `auto-*` file should be named like the `init` file but with the `auto-` suffix.

By the way, should we seize the 3.0 opportunity to rename our config files?
The benefit is that migration would be smoother since Nyxt 3 would not load the 2.* config by default.

I suggestion full-word names:

- `settings.lisp`
- `configuration.lisp`
- `initialization.lisp`

Thoughts?

Ambrevar:
Fixes #2074.

I went with Artyom's suggestion of using the XDG-recommended "config" instead of the full "configuration".

It's shorter, explicit enough and familiar.

Do you people like it?

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -258,7 +258,7 @@ prevents otherwise.")
     (let* ((message (princ-to-string condition))
            (full-message (format nil "Startup error: ~a.~%~&Restarted Nyxt without init file ~s."
                                  message
-                                 (files:expand *init-file*)))
+                                 (files:expand *config-file*)))
            (new-command-line (append (uiop:raw-command-line-arguments)
                                      `("--no-init"
                                        "--eval"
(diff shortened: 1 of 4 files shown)


aadcg:
I don't oppose this change!

jmercouris:
Looks good, does it still check for `init.lisp` though?

jmercouris:
From my first impression, it feels like it won't do that, no? Are willing to break this for 3.0.0?

Ambrevar:
Oh right, forgot about this, good catch.

So let's clarify what we want in `resolve`:

- Run the current code.
- If it returns nothing, then fallback on `~/.config/nyxt/init.lisp`.

Correct?

Ambrevar:
Actually, shall we really do it?
Because almost everything is broken when trying to load a 2.2.4 init file (mostly because of the mode re-architecture).  So not loading `init.lisp` can be seen as a feature.  Otherwise people starting Nyxt 3 for the first time will get an error message.

Instead, what about detecting the presence of `init.lisp` and echoing a warning that it was _not_ loaded, and that the user should use `config.lisp` instead?

aadcg:
> [...] So not loading init.lisp can be seen as a feature. Otherwise
> people starting Nyxt 3 for the first time will get an error message.
>
> Instead, what about detecting the presence of init.lisp and echoing a
> warning that it was not loaded, and that the user should use config.lisp
> instead?

That's sounds like a fine idea to me.

jmercouris:
I can agree with that!

Ambrevar:
Merged with 456f781aea0e30e9fbf81e4375867294bb34daa2.

PR #2288, opened 2022-05-23
input: Catch invalid symbols in keymaps.

Ambrevar:
Should at last catch the broken bindings in the CI.

The reason we hadn't done it so far is that it "pollutes" the `keymap` library by setting the default type globally.

Don't know how to solve this probably yet, but for now the benefits outweigh the costs.  `keymap` is internal to Nyxt after all.


Ambrevar:
OK, I think I've figured it out.  Now we no longer pollute the `keymap` library, instead we define our own schemes.  The drawback is that we can't use `scheme:cua`, etc. anymore.  Unless...

Proposal:
- Rename `keymaps`s `scheme` package to `keymap/scheme` (makes sense).
- Rename the new `nyxt/keymap` to `nyxt/scheme` and add `scheme` alias.

Now we can keep the `scheme:cua`, etc.

Agreed?

Ambrevar:
Merged with 08fd206b24813a4e04b110f851e105a640ce4f27.

PR #2291, opened 2022-05-24
define-package helper

Ambrevar:
We've got some 200 lines of `define-package` boilerplate, I've figured we'd rather optimized it out.

There you go: to me it's a no brainer, but let me know if you don't like the new `define-and-set-package` syntax.


source/package.lisp
```diff
@@ -1,38 +1,66 @@
...
+                  :nyxt))
+
 (eval-when (:compile-toplevel :load-toplevel :execute)
-  (trivial-package-local-nicknames:add-package-local-nickname :alex :alexandria :nyxt)
-  (trivial-package-local-nicknames:add-package-local-nickname :sera :serapeum :nyxt)
-  (trivial-package-local-nicknames:add-package-local-nickname :lpara :lparallel :nyxt)
-  (trivial-package-local-nicknames:add-package-local-nickname :class* :hu.dwim.defclass-star :nyxt)
-  (trivial-package-local-nicknames:add-package-local-nickname :hooks :nhooks :nyxt)
-  (trivial-package-local-nicknames:add-package-local-nickname :files :nfiles :nyxt))
+  (loop :for (nickname package) in
```
jmercouris, line 29: Lol, I'm not sure this optimization is necessary :-D

source/package.lisp
```diff
@@ -1,38 +1,66 @@
...
   (:export #:use-nyxt-package-nicknames)
   (:documentation "The core package of Nyxt, the infinitely extensible browser.
 
 This package should not be modified by the users.
 
 It's recommended to use the `nyxt-user' package instead to create new functions,
 modes, commands, etc."))
+
+(in-package :nyxt)
+(defvar *imports* '((:trivia :match :multiple-value-match :lambda-match :guard)
```
jmercouris, line 18: I think it would be wise to include a documentation string for `*imports*` explaining what it does.

aadcg:
Beautiful!  I second @jmercouris and think that a docstring is missing for `*imports*`.

Ambrevar:
Note that `*imports*` is a low-level implementation detail, the programmer is supposed to use `define-and-set-package`.

Ambrevar:
Merged with a53cda883dd91338e1a169a37050e3932dc10719.

PR #2295, opened 2022-05-25
Validate make instance

### LINKED ISSUE #2262: Catch make-instance non-class symbols at compile time
A frequent class of error is when calling `make-instance` over a symbol that's not a class.  This happens due to typos, renaming or, typically, when moving class definition around.

A trick could be to shadow `make-instance` with a macro that checks for class existence at macro-expansion time.
Proof of concept:

```lisp
(uiop:define-package nyxt
   (:use :cl) 
  (:shadow #:make-instance))

(in-package :nyxt)

(defmacro make-instance (sym &rest args)
   (find-class (second sym) )
 `(apply #'cl:make-instance ,sym ,args ))

(lambda () (make-instance 'i-do-not-exist))
; in: LAMBDA ()
;     (FOO::MAKE-INSTANCE 'FOO::I-DO-NOT-EXIST)
; 
; caught ERROR:
;   during macroexpansion of (MAKE-INSTANCE 'I-DO-NOT-EXIST). Use
;   *BREAK-ON-SIGNALS* to intercept.
;   
;    There is no class named FOO::I-DO-NOT-EXIST.
; 
; compilation unit finished
;   caught 1 ERROR condition
; => #<FUNCTION (LAMBDA ()) {547D097B}>
```

But really we only need this in the CI.  So I suggest we add a compilation option that defines the package and the macro only if some environment variable is set, and set this variable from the GitHub CI script.

Thoughts?

Ambrevar:
Fixes #2262.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -78,6 +78,11 @@ jobs:
       run: |
         ros -e '(asdf:load-system :nyxt/submodules)' -e '(load "build-scripts/report-warnings.lisp")' -e '(compilation-conditions "nyxt")' -e '(compilation-conditions "nyxt/gtk")'
 
+    - name: Validate make-instance symbols
+      shell: bash
+      run: |
+        ros -e '(asdf:load-system :nyxt/submodules)' -e '(push :nyxt-debug-make-instance *features*)' -e '(handler-case (asdf:load-system :nyxt) (error (a) (format t "caught error ~s~%~a~%" a a) (uiop:quit 17)))'
+
     - name: Keymap type checking
       shell: bash
       run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -81,7 +81,7 @@ jobs:
     - name: Validate make-instance symbols
       shell: bash
       run: |
-        ros -e '(asdf:load-system :nyxt/submodules)' -e '(push :nyxt-debug-make-instance *features*)' -e '(handler-case (asdf:load-system :nyxt) (error (a) (format t "caught error ~s~%~a~%" a a) (uiop:quit 17)))'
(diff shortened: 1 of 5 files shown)


aartaka:
Looks wonderful to me!

Ambrevar:
The original macro implementation didn't work: it would break `apply 'make-instance ...` :(

So then I decided to fall back to a functional wrapper and leverage CCL typing to catch errors.  This works but not everywhere, for instance it fails to catch initform errors.

At least it found a bug in `preview-file`!  Hooray!

aadcg:
I like the idea but shouldn't such things belong to the tests folder? Aren't  we "polluting" our codebase with assertions and proofs?

Ambrevar:
The `#+nyxt-debug-make-instance` is a feature macro, so the code following it is not `cl:read` unless the keyword is in the `*features*` list.  So we are not polluting anything by default.

That said, we could define the wrapper and shadow the symbol in a separate test file.  I'll see what I can do.

aadcg:
I didn't explain myself very well with the word "polluting". I know it's clean in the sense you mentioned. But it seems to me that one should not see such assertions in the source file, since it shifts the attention to the implementation itself.

Ambrevar:
This might not be doable externally.  CL is powerful, but some day we hit the limit :p  If you can think of a way to completely remove the feature macros, please share :)

aadcg:
I won't even try, as it's completely out of my abilities right now. 

What exactly are feature macros and what distinguish them from macros? Any reference? Thanks.

Ambrevar:
Maybe you can start reading from here: http://www.lispworks.com/documentation/lw71/CLHS/Body/24_aba.htm

Ambrevar:
Merging for now, I'll try to move the test to a proper ASDF test later.

jmercouris:
I must admit, this is a most clever way to do it. I would have not considered it in a million years. Well done.

Issue #1915, opened 2021-12-06
Emacs-mode and vi-normal-mode should be commands, not modes


Ambrevar:
In https://github.com/atlas-engineer/nyxt/issues/1871#issuecomment-986839343 @mariari was confused about how emacs-mode and vi-normal-mode worked, in particular that the order of activation changed the inheritance, which is indeed counter-intuitive,

The way Nyxt bindings are looked up is by navigating a bidimensional matrix: 
- 1 dimension is the mode keymaps,
- the other one the schemes.

While it makes sense to toggle modes to also toggle their keymaps, it looks like a cognitive dissonance that we also have to toggle a mode to change scheme.
Also note that we can have both vi-normal-mode and emacs-mode enabled at the same time, while only one scheme is set, so this does not make much sense.

What we could do instead is have a command that lists the known schemes and sets the chosen one.  The current scheme would be displayed in the status area, just like we display the VI indicator today.

The only remaining question is what we do of the code that's in `emacs.lisp` and, most importantly, `vi.lisp`?  Some state must be stored, where?
We also specialize `on-signal-load-finished` against `vi-insert-mode`, so if we remove the mode, what do we specialize it against?

@jmercouris @aartaka ?

Ambrevar:
Ping! :)

aartaka:
> The only remaining question is what we do of the code that's in `emacs.lisp` and, most importantly, `vi.lisp`? Some state must be stored, where? 

`*browser*`? :D It's the most "global" space for settings that we have. Alternatively, store it in the `base-mode` as the most general mode.

>We also specialize `on-signal-load-finished` against `vi-insert-mode`, so if we remove the mode, what do we specialize it against?

Not sure about it. If we store it in `base-mode`, then we can specialize against `base-mode`. Maybe move it to the `buffer`-specialized method? Feels wrong, though. Having it isolated in it's own mode sounds like a nice solution here, but yes, it does indeed have its own deficiencies.

jmercouris:
I'm afraid I don't have a good answer to your question about storing the state.

Ambrevar:
See also #1841.

aadcg:
Seems like a good plan.

This is a minor thing, but why was the name `scheme` chosen?

Ambrevar:
For the lack of a more inspired one... Less confusing names are welcome :)

aadcg:
> X uses the terms "layout" and "variant". The benefit of "layout" compared
[...]

PR #2297, opened 2022-05-25
Only allow one keymap-scheme-mode at a time.

Ambrevar:
Add keymap-scheme-mode and ensure each instance is uniquely active for a given buffer.

Partially addresses #1915.
After that, we'll have to rename `keymap-scheme-name` to `keymap-family`.


aadcg:
Looks good to me. I can do the rename. Should I push changes here or do it after merging it?

Ambrevar:
No, let's do the rename later, in a separate pull request.

I wonder why the CI is hanging here...

PR #2299, opened 2022-05-25
mode/history: Visit page in corresponding buffer when clicking on his…

Ambrevar:
…tory tree link.

Open the `history-tree` (or `buffer-history-tree`).  Now if you click on the link, it will switch to the corresponding buffer and navigate back or forward to the corresponding history location, without corrupting the flow of the history.

Try it, it's awesome!

@aartaka Initially the history tree visualization showed links, which could be seen as more appropriate.
But maybe they are not, since we are not loading anything?
Either way, the style looks really bad.  Can we somehow make it look more like links and less like buttons?

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -64,6 +64,13 @@ various parts, such as the path of all data files.")
        :border-radius "2px"
        :padding "6px"
        :margin "2px")
+      (.link
+       :all "unset"
+       :text-decoration "underline"
+       :display "inline"
+       :color theme:primary)
+      (".link:hover"
+       :color theme:text)
       (.accent
        :background-color theme:accent)
       (|.button:hover|
--- a/source/mode/history-tree.lisp
+++ b/source/mode/history-tree.lisp
@@ -7,9 +7,6 @@
 (define-mode history-tree-mode ()
   "Mode for history-tree listing."
   ((rememberable-p nil)
-   (display-buffer-id-glyphs-p t
-                               :documentation "Whether to show unique glyphs
-matching buffer `id's along with buffer history entries.")
    (style
     (theme:themed-css (theme *browser*)
       (body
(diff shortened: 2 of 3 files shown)


aartaka:
> …tory tree link.
> 
> Open the `history-tree` (or `buffer-history-tree`). Now if you click on the link, it will switch to the corresponding buffer and navigate back or forward to the corresponding history location, without corrupting the flow of the history.
>
> Try it, it's awesome!

Whoa, I've been thinking about doing it for a long time! And now you beat me to it, Pierre :D Thanks for taking the time!

> @aartaka Initially the history tree visualization showed links, which could be seen as more appropriate. But maybe they are not, since we are not loading anything? Either way, the style looks really bad. Can we somehow make it look more like links and less like buttons?

One word: CSS! You can use [`all` property](https://developer.mozilla.org/en-US/docs/Web/CSS/all#formal_definition) to reset all the styling, or you can use a mixture of `:border`, `:display`, and `:text-decoration` to remove things one by one. It's all configurable there :)

Ambrevar:
What I'd like to do (and we probably need this elsewhere) is to "import" the style of `a` elements.  Is there a way to do this?

aartaka:
> Question: why do we have two `style` lines in `history-tree`?
> 
> ```lisp
>              (:style (style output-buffer))
>              (:style (:raw (style mode)))
> ```

This is to use styles of both the buffer and the mode. Is somthing wrong with this?

aartaka:
> What I'd like to do (and we probably need this elsewhere) is to "import" the style of `a` elements. Is there a way to do this?

Nope, you can only reproduce it :)

Ambrevar:
Mind doing it?  (Feeling lazy to play with CSS... :p)

aartaka:
Done with 8b00d3b227de1e5b8803afb4b39de7004e38fc49.

Ambrevar:
Thanks a lot!

PR #2305, opened 2022-05-26
Move compilation tests to ASDF systems

Ambrevar:
Like @aadcg astutely remarked in https://github.com/atlas-engineer/nyxt/pull/2295, we should probably not pollute our code base with reader macros if they are used just for testing.

While we are at it, it makes sense to integrate all the "compilation tests" (those that are run from the `.github/workflows/tests.yml` file) into ASDF.

The following are easy to fix:
- error out on compilation warning : now it's done _at all times_.
- unbound export check: can be moved to a regular `prove` test.
- manual generation: it's enough to call it from `nyxt/test`.

The rest is harder:
- "Validate make-instance symbols"
- "Keymap type checking"

Both of them require that we load `nyxt` once, tweak something then reload `nyxt`.  But ASDF does not support calling itself recursively from a system (i.e. you can't call `asdf:load-system`).

Is there a way to force the load of the `nyxt` system`?


aadcg:
Hmm, I'm afraid I can't help much. Thank you for working on this. 

It's ok if I can't figure it all now.

Ambrevar:
Maybe we can leverage `nyxt-asdf` now.

Ambrevar:
Side note: An alternate implement of `make-instance` validation would be to define a compiler macro like this:

```lisp
(in-package :cl)
(define-compiler-macro make-instance (&whole form &rest args)
  (find-class (second (first args)))
  form)
```

Ambrevar:
Actually it does not seem to work well with `apply`, e.g.

```lisp
(apply #'make-instance 'dummy '(:arg1 VALUE ))
```

aartaka:
> Actually it does not seem to work well with `apply`, e.g.
> 
> ```lisp
> (apply #'make-instance 'dummy '(:arg1 VALUE ))
> ```

And that's by design: compiler macros are intended to be invoked as `(make-instance ...)`, and invoking the function with compiler macro via `apply` will simply ignore the compiler macro part.

But even with this restriction, compiler macros are much better than reader-macro-polluted redefinitions.

Ambrevar:
Superseded by #2680.

PR #2306, opened 2022-05-26
Lisp nyxt urls refactor

### LINKED ISSUE #2268: Why do some function take a buffer-id instead of buffer? | Allow non-literal data in nyxt: URLs

Ambrevar:
This enables full Lisp (and closures!) in Lisp URLs.  No more need to juggle around with the inspector values, etc.

Hooray!

Left to do:
- [x] `nyxt://` URLs support.  It's a bit different.
- [ ] Remove the old schemes and apply the new ones everywhere.

@aartaka Do you like the implementation?

Fixes #2268.


source/parenscript-macro.lisp
```diff
@@ -127,3 +128,31 @@
...
+
+(export-always 'funcall-ps-callback)
+(defun funcall-ps-callback (callback-index key)
+  "KEY refers to a callback in the CALLBACK-INDEX hash-table.
+See the `lisp-eval2' scheme."
+  (unwind-protect (nyxt:funcall* (gethash key callback-index))
+    (remhash key callback-index)))
+
+(export-always 'lisp-eval2)
+(defpsmacro lisp-eval2 ((callback-index &optional title callback) &body form)
```
aartaka, line 140: Why pass `callback-index`? Any specifix use-case for extensions?

source/buffer.lisp
```diff
@@ -1283,7 +1288,9 @@ URL is then transformed by BUFFER's `buffer-load-hook'."
       (cond
         ((equal "javascript" (quri:uri-scheme url))
          (ffi-buffer-evaluate-javascript buffer (quri:url-decode (quri:uri-path url))))
-        (t (ffi-buffer-load buffer url))))
+        (t
+         (clrhash (lisp-url-callbacks buffer)) ; REVIEW: Is it the only spot where to clear the Lisp URL callbacks?
```
aartaka, line 1293: We also need to clear it at buffer deletion, I suppose?

source/parenscript-macro.lisp
```diff
@@ -127,3 +128,31 @@
...
+  "Request the lisp: URL and invoke callback when there's a successful result.
+CALLBACK-INDEX is a hash-table which must exist
+TITLE is purely informative."
+  `(let ((request (ps:lisp (let ((request-id (string (gensym ""))))
```
aartaka, line 144: Maybe use a simpler idiom with a global counter var only increased in one place/function? Or serialize gensyms to URLs directly?

source/parenscript-macro.lisp
```diff
@@ -127,3 +128,31 @@
...
+(export-always 'funcall-ps-callback)
+(defun funcall-ps-callback (callback-index key)
+  "KEY refers to a callback in the CALLBACK-INDEX hash-table.
+See the `lisp-eval2' scheme."
+  (unwind-protect (nyxt:funcall* (gethash key callback-index))
+    (remhash key callback-index)))
+
+(export-always 'lisp-eval2)
+(defpsmacro lisp-eval2 ((callback-index &optional title callback) &body form)
+  "Request the lisp: URL and invoke callback when there's a successful result.
```
aartaka, line 141: What's a "successful result"? Callback existing with such an ID?

source/buffer.lisp
Ambrevar: No, then it would be garbage collected with the rest of the buffer slots.

source/parenscript-macro.lisp
Ambrevar: How do you know where to store the callback otherwise?

Ambrevar: `callback-index` is poorly named, it's actually the hash-table.

source/parenscript-macro.lisp
Ambrevar: I don't know, I believe you are the author of this docstring! :)  See `lisp-eval`.

source/parenscript-macro.lisp
Ambrevar: I'm confused, what simpler idiom do you suggest?
I don't know how to shorten `(string (gensym ""))`.

Ambrevar:
`nyxt://` refactor done in https://github.com/atlas-engineer/nyxt/pull/2306/commits/3dedeaee2439a4936aca3a5fd0c66e848c5448a9.

See how it can fix the "command buffer ID issue" in https://github.com/atlas-engineer/nyxt/pull/2306/commits/1c9259318dc409cb184e4c339685866ebc99be47.

@aartaka What do you think?

Note: The API hasn't changed for now, but this needs more testing.

Ambrevar:
Tested, it seems to be working perfectly fine :)

pdelfino:
Hi guys,

I am having a problem with internal pages on Hermes. @jmercouris thought this PR would have solved it.

I pulled this PR to my local env and tried it. Unfortunately, It did not work. Hence, I am posting it here. Especially considering:

> Tested, it seems to be working perfectly fine :)

Simplifying the problem. Suppose you have the following:

```lisp
(define-internal-page-command discuss-internal-page (&key request-data)     
          (buffer "*Get*")
        "Dummy internal page to discuss problem"
        (spinneret:with-html-string
          (:style (cl-css:css '(("code"
                                 :display "block"
                                 :overflow-x "scroll"))))
          (:h1 (format nil "TESTÃO"))))
```
Notice the parameter definition is irrelevant to the output.

It works with different argument types being called, such as a string, an empty list (false), a true, faloat, a fraction:

```lisp
NYXT> (discuss-internal-page :request-data "http://www.paulgraham.com")
#<WEB-BUFFER 1323 {100225D2D3}>

NYXT> (discuss-internal-page :request-data t)
#<WEB-BUFFER 1324 {1002A0A303}>

NYXT> (discuss-internal-page :request-data 123)
#<WEB-BUFFER 1325 {100389A2D3}>

NYXT> (discuss-internal-page :request-data 123.00)
#<WEB-BUFFER 1326 {1004F58713}>

NYXT> (discuss-internal-page :request-data (/ 1 3))
#<WEB-BUFFER 1327 {1005578753}>
```

However, it breaks with a **non-empty** list as an argument:
```lisp
NYXT> (discuss-internal-page :request-data '("http://www.paulgraham.com"))
#<WEB-BUFFER 1327 {1005578753}>

A non-constant value passed in URL params: (http://www.paulgraham.com)
```

Is this expected? Did I miss something?

Thanks

Ambrevar:
Internal pages are like "web servers", they only accept standard URL query syntax (see https://en.wikipedia.org/wiki/Query_string).  This PR does not change this.

More importantly, there is a semantic question.  Internal pages are meant to return reproducible result.  Call it twice over the same argument, it should return the same page.  This would not be true if the argument where not a scalar like an object, a function, or a list of things.

Maybe there is a way to support _sequences of constants_ though.  @aartaka ?

Finally, maybe what you want here is not an internal `nyxt://` page but rather a `lisp://` URL.  `lisp://` URL are secure and are meant to run anything, so you could easily pass your list there.  See `lisp-eval` examples in the code base and in this PR.

aartaka:
The initial design was prohibiting any sequences except string so that it would become significantly harder to use `eval` on command args. But then, `nyxt:` URLs are not invokable by anyone but the user, so it should be safe to allow essentially a Lispy JSON — all the literal, `read`able data, inluding arrays and lists.

aartaka:
Nested sequences are safe too, granted that they and their contents are recursively constant and/or sequence-like.

source/parenscript-macro.lisp
aartaka: I mean we can hard-code the global variable name, can't we? Why this much flexibility? :)

source/parenscript-macro.lisp
aartaka: Ooops...

source/parenscript-macro.lisp
aartaka: What I meant is a typical idiom for numeric ID generation:
``` lisp
(defvar *counter* 0)
(defun new-id ()
  (incf *counter*))
```

Why generate new interned symbols if we can generate numbers in a simpler and side-effect-less way?

source/renderer-script.lisp
```diff
@@ -107,91 +107,167 @@ If `setf'-d to a list of two values -- set Y to `first' and X to `second' elemen
...
+    (cond
+      ((stringp title)
+       title)
+      ((functionp title)
+       (funcall title args))
+      (t
+       (format nil "*~a*" (string-downcase (name page)))))))
+
+;; (-> find-internal-page-buffer (internal-page-symbol) (maybe buffer))
+(defun find-internal-page-buffer (name) ; TODO: Test if CCL can catch bad calls at compile-time.
```
aartaka, line 167: `define-compiler-macro`?

source/renderer-script.lisp
```diff
@@ -107,91 +107,167 @@ If `setf'-d to a list of two values -- set Y to `first' and X to `second' elemen
...
+
+(export-always 'define-internal-page)
+(defmacro define-internal-page (name (&rest arglist)
+                                ;; TODO: Move `buffer-var' (and `title'?) to a keyword arg.
+                                (buffer-var title &key mode command-p global-p)
+                                &body body)
+  "Register a new nyxt:// URL under NAME.
+When loaded, BODY is run to populate the page content.
+
+WARNING: Don't run anything sensitive in the BODY as any third-party page can
```
aartaka, line 192: Actually, no. `nyxt://` pages are declared as local and thus cannot be linked to by any web page. We're safe here.

source/renderer-script.lisp
```diff
@@ -107,91 +107,167 @@ If `setf'-d to a list of two values -- set Y to `first' and X to `second' elemen
...
+internal page buffer.  The command takes ARGLIST parameters.
+With GLOBAL-P, make this command globally accessible."
+  (multiple-value-bind (required optional rest keyword allow-other-keys-p aux key-p )
+      (alex:parse-ordinary-lambda-list arglist)
+    (declare (ignore allow-other-keys-p key-p keyword))
+    (when (or required optional rest aux)
+      (error "Only keyword parameters are allow in an internal-page definition.")))
+  (multiple-value-bind (body declarations documentation)
+      (alex:parse-body body :documentation t)
+    (let* ((keyargs (nth-value 3 (alex:parse-ordinary-lambda-list arglist)))
```
aartaka, line 214: Why? There's `keyword` binding available in this scope...

source/renderer-script.lisp
```diff
@@ -107,91 +107,167 @@ If `setf'-d to a list of two values -- set Y to `first' and X to `second' elemen
...
 (export-always 'define-internal-page-command-global)
 (defmacro define-internal-page-command-global (name (&rest arglist)
                                                (buffer-var title &optional mode)
                                                &body body)
-  "Define a command called NAME creating an internal interface page.
-
-Define a global command to access this page.
+  "Define a global command called NAME creating an internal interface page.
 
 See `define-internal-page-command' for the explanation of arguments."
```
aartaka, line 267: Refer to `define-internal-page` instead?

aartaka (commented):
Looks just fine to me, just some more comments on `nyxt:` URLs :)

source/parenscript-macro.lisp
Ambrevar: Because it's _not_ a global variable, it's stored in the buffer slot! :)

source/parenscript-macro.lisp
Ambrevar: The reason to be of `gensym` is to generate conflict-free IDs, so why not use it?  It's perfect here.

We used to do like you suggested for buffer and window ID, but recently switched to using gensyms: much better in my opinion.  Less code, more idiomatic.

source/renderer-script.lisp
```diff
@@ -107,91 +107,167 @@ If `setf'-d to a list of two values -- set Y to `first' and X to `second' elemen
...
+    :export t
+    :documentation "The key of the internal-page in `*nyxt-url-commands*'.")
+   (dynamic-title ; Not `title' so that it does not clash with other `title' methods.
+    ""
+    :initarg :title
+    :accessor nil
+    :type (or string function)
+    :documentation "If a function, it is called with the internal page arguments
+and must return a string.")
+   (page-mode
```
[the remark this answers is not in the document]
Ambrevar: Yes, but we need to think this through.  Our general-purpose notion of "modes" does not go very well with these per-URL "major modes".  We lack a good mechanism to toggle mode on URL change.

source/renderer-script.lisp
Ambrevar: Why?

source/renderer-script.lisp
Ambrevar: Wait, what I have understood is that `nyxt://` is meant to be called from anywhere, hence its strict argument processing.

What's the benefit of `nyxt://` over `lisp://` then?

By the way, I realized the `lisp://` URLs are not local, is that intentional?  Shouldn't it be swapped?

source/renderer-script.lisp
Ambrevar: Sorry, I don't understand.

internal pages only accept keyword arguments, hence this  code.

source/parenscript-macro.lisp
aartaka: Hmmm, I see now. Our design forces us to be flexible :D

source/parenscript-macro.lisp
aartaka: Yes, makes sense.

source/renderer-script.lisp
aartaka: I mean, you can test the type of the symbol at compile-time, if you use a compiler macro.

source/renderer-script.lisp
aartaka: Indeed, `nyxt://` URLs may be too strong for what they are. They cannot be invoked from elsewhere AND they restrict their arguments to the simplest values possible. We can safely drop one of the two. I suggest dropping the latter (value restriction)—@pdelfino noted its inadequacy already.

`lisp://` URLs are not protected by WebKit, and they are CORS-ready, which makes them extemely open to any request (and exploitation). That's why we check the buffer and URL of all the `lisp://` requests in the scheme callback—to avoid exploitation. I don't see a better way here yet.

source/renderer-script.lisp
aartaka: I mean, the parsed keyword arguments are already bound to the variable some couple of lines above, aren't they?

source/renderer-script.lisp
Ambrevar: CCL can do it for us via typing.

However I never thought we could leverage compiler-macro to do this on SBCL too!  (Any typing requiring `satisfies` basically.).  Thanks for the tip!

Can we generalize this trick somehow?

source/renderer-script.lisp
Ambrevar: Why not marking `lisp://` as local and not CORS ready?

source/renderer-script.lisp
Ambrevar: Oops!  Good catch,  I wasn't reading well!

source/renderer-script.lisp
aartaka: Not sure -- compiler macros are some black sheep in the standard, they are (purposefully?) hard to use and dynamically modify...

source/renderer-script.lisp
aartaka: Nope, that would forbid making requests to `lisp://` URLs -- we can only make requests without CORS warnings if the requested scheme is CORS-enabled.

aartaka: At least I had this impression the last time I tinkered with custom schemes machinery :)

Ambrevar:
Before merging I'd like to review this myself one more time.

In the mean time, is anyone against merging this?
@aartaka ?

Ambrevar:
To do: Turn `describe-value` into a command-less internal page.

aartaka:
@Ambrevar, can you summarize in which way does this PR change the `lisp://` and `nyxt://` URL behavior?

Ambrevar:
Sure:

- `lisp://` URLs: The `lisp-eval2` macro now registers the body as a callback in the `*lisp-url-callbacks*` hash table, and the "lisp2" scheme handler calls it.  

  The `lisp://` URLs now have this syntax: `lisp://ID/TITLE`, where TITLE is optional and purely informative.

- `nyxt://` URLs: The logic is the same as before except that commands are decoupled from internal pages.  Now one can call `define-internal-page` directly, which will not produce a command.

  Since an internal page may be loaded without a command being called, now the `nyxt` scheme handler does extra things, like enabling a mode.

  Internal pages now have their own data structure (cleaner), which carries, among other, the associated mode.  This really calls for a notion like major modes, but we can tackle this another time.

aartaka:
Yes, I can make sense of it. Looks totally mergeable to me :)

Ambrevar:
Closing in favour of #2354.

PR #2311, opened 2022-05-27
WIP: Fix list bookmarks

### LINKED ISSUE #2062: memory fault on list-bookmarks
If I run `list-bookmarks` over my 3000+ bookmark file, I get

```log
Unhandled memory fault at #x7F429F5DC010.
   [Condition of type SB-SYS:MEMORY-FAULT-ERROR]

Restarts:
 0: [ABORT] abort thread (#<THREAD "Nyxt renderer thread" RUNNING {1010D28AA3}>)

Backtrace:
 0: (SB-SYS:MEMORY-FAULT-ERROR #<unused argument> #.(SB-SYS:INT-SAP #X7F429F5DC010))
 1: ("foreign function: call_into_lisp")
 2: ("foreign function: funcall2")
 3: ("foreign function: handle_memory_fault_emulation_trap")
 4: ("foreign function: #x4176CE")
 5: ((LAMBDA (&REST GIR::ARGS-IN) :IN GIR::BUILD-FUNCTION))
 6: ((FLET NYXT::MAIN-FUNC :IN NYXT:FFI-INITIALIZE))
 7: ((LAMBDA NIL :IN BORDEAUX-THREADS::BINDING-DEFAULT-SPECIALS))
 8: ((FLET SB-UNIX::BODY :IN SB-THREAD::RUN))
 9: ((FLET "WITHOUT-INTERRUPTS-BODY-10" :IN SB-THREAD::RUN))
10: ((FLET SB-UNIX::BODY :IN SB-THREAD::RUN))
11: ((FLET "WITHOUT-INTERRUPTS-BODY-3" :IN SB-THREAD::RUN))
12: (SB-THREAD::RUN)
13: ("foreign function: call_into_lisp")
14: ("foreign function: funcall1")
```

It works on a small 2-entry bookmark file.
The 3000-entry bookmark file works on 2.2.4.

@aartaka This may be due to some change in the internal-page system.  Any idea what to investigate?

Ambrevar:
Fixes #2062. 

Or at least work around the memory issue.

I don't know if anyone is willing to work on pagination / buffering in cl-webkit CFFI, otherwise I suggest we implement a fix like here.

Try it out:

- Create a bookmark file with 3000+ entries like in https://github.com/atlas-engineer/nyxt/issues/2062#issuecomment-1088568861.
- Start nyxt.
- Run `list-bookmarks`.
- Run `list-bookmarks-populate`.

I've created a separate command which would obviously be gone in the final patch.

Proposal: create a list-bookmarks-populate mode, add  an `nyxt:on-signal-load-finished` specialization which does what `list-bookmarks-populate` does.

Thoughts?

### DIFF
--- a/source/foreign-interface.lisp
+++ b/source/foreign-interface.lisp
@@ -108,6 +108,10 @@ Setf-able."))
 (define-ffi-generic ffi-buffer-load (buffer url)
   (:documentation "Load URL into BUFFER through the renderer."))
 
+(define-ffi-generic ffi-buffer-load-html (buffer html-content url)
+  (:documentation "Load HTML into BUFFER through the renderer.
+If URL is not nil, relative URLs are resolved against it."))
+
 (define-ffi-generic ffi-buffer-evaluate-javascript (buffer javascript &optional world-name)
   (:documentation "Evaluate JAVASCRIPT in the BUFFER web view.
 See also `ffi-buffer-evaluate-javascript-async'."))
--- a/source/mode/bookmark.lisp
+++ b/source/mode/bookmark.lisp
@@ -103,29 +103,64 @@ Bookmarks can be persisted to disk, see the `bookmarks-file' mode slot."
     (spinneret:with-html-string
       (:style (style (find-submode 'bookmark-mode (current-buffer)))) ; TODO: Make sure this is the right buffer
       (:h1 "Bookmarks")
-      (cond
-        ((zerop (hash-table-count bookmarks))
-         (:p (format nil "No bookmarks in ~s." (files:expand (bookmarks-file bookmarks-buffer)))))
-        (t (maphash
-            (lambda (tag bookmarks)
-              (:details
-               (:summary (or tag "Unsorted"))
-               (dolist (bookmark bookmarks)
-                 (let ((uri-host (quri:uri-host (url bookmark)))
-                       (url-href (render-url (url bookmark))))
-                   (:dl
-                    (:dt (:button :onclick (ps:ps (delbkm (ps:lisp url-href))) "✕")
-                         (serapeum:ellipsize (title bookmark) 80))
-                    (:dd (:a :href url-href uri-host))
-                    (when (tags bookmark)
-                      (:dd (format nil " (~{~a~^, ~})" (tags bookmark)))))
-                   (:hr)))))
-            bookmarks)))
+      ;; (cond
+      ;;   ((zerop (hash-table-count bookmarks))
+      ;;    (:p (format nil "No bookmarks in ~s." (files:expand (bookmarks-file bookmarks-buffer)))))
+      ;;   (t (maphash
+      ;;       (lambda (tag bookmarks)
+      ;;         (:details
+      ;;          (:summary (or tag "Unsorted"))
+      ;;          (dolist (bookmark bookmarks)
+      ;;            (let ((uri-host (quri:uri-host (url bookmark)))
+      ;;                  (url-href (render-url (url bookmark))))
+      ;;              (:dl
+      ;;               (:dt (:button :onclick (ps:ps (delbkm (ps:lisp url-href))) "✕")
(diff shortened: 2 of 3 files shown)


Ambrevar:
Maybe better than creating a mostly-useless mode: add a handler to `buffer-load-hook`.

Ambrevar:
@aartaka concern was that by loading the content manually we would lose caching, history, etc.  But actually `web-view-load-html` preserves all these features.  From the doc (https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#webkit-web-view-load-html):

> You can monitor the load operation by connecting to “load-changed” signal.

There is also `webkit:webkit-web-view-load-alternate-html`:

> When this method is called from “load-failed” signal to show an error page, then the back-forward list is maintained appropriately. For everything else this method works the same way as webkit_web_view_load_html().

This would avoid duplicating the URL in history.

At this point, I'm wondering if this mechanism could not fully replace the FFI scheme handling.  The main unanswered question is _where_ do we call `webkit:webkit-web-view-load-alternate-html` when loading a `nyxt://` URL?

aartaka:
> @aartaka concern was that by loading the content manually we would lose caching, history, etc. But actually `web-view-load-html` preserves all these features. From the doc (https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#webkit-web-view-load-html):
> 
> > You can monitor the load operation by connecting to “load-changed” signal.

Huh, that's interesting...

> There is also `webkit:webkit-web-view-load-alternate-html`:
> 
> > When this method is called from “load-failed” signal to show an error page, then the back-forward list is maintained appropriately. For everything else this method works the same way as webkit_web_view_load_html().
> 
> This would avoid duplicating the URL in history.
> 
> At this point, I'm wondering if this mechanism could not fully replace the FFI scheme handling. The main unanswered question is _where_ do we call `webkit:webkit-web-view-load-alternate-html` when loading a `nyxt://` URL?

I mean, custom schemes infrastrucutre that WebKit provides is reliable enough, until there's too much data. I'd not ditch it now.

Regarding `webkit:webkit-web-view-load-alternate-html` call site -- call it in `load-failed` -- it works for custom schemes too!

Ambrevar:
> Regarding `webkit:webkit-web-view-load-alternate-html` call site -- call it in `load-failed` -- it works for custom schemes too!

But the signal would be triggered if there is an error while loading.  So what do we initially try to load for this signal to be triggered?

aartaka:
I was so hesitant about ditching the custom URL mechanisms of WebKit, that I [fixed the string allocation problem in cl-webkit](https://github.com/joachifm/cl-webkit/commit/83a90f4ee74f3811532373deec617d5eaf5f46d1) :D

aartaka:
> > Regarding `webkit:webkit-web-view-load-alternate-html` call site -- call it in `load-failed` -- it works for custom schemes too!
> 
> But the signal would be triggered if there is an error while loading. So what do we initially try to load for this signal to be triggered?

`nil` or throw a condition, as per `cl-webkit`.

Ambrevar:
Are you saying that we can now load more than 3000+ bookmarks without triggering any error?

aartaka:
Yes! There are no limits now :D

Ambrevar:
I suppose this can be closed then, but we can at least include the new ffi-buffer-load-html endpoint as it's quite useful.

aartaka:
And `load-alternative-html` one too!

Ambrevar:
Just tested, works beautifully!  Thanks a lot, @aartaka !

I've merged the new FFI functions with cd3f99276709e14624190417c737b1b9221629b2.

jmercouris:
I was thinking `ffi-buffer-load-alternate-html` appears to be a renderer specific detail for webkit. Perhaps it would have been better to add an argument to `ffi-buffer-load-html`?

Its immaterial in the end, just an observation that I am making!

PR #2312, opened 2022-05-27
Fix keymap priority

### LINKED ISSUE #2307: Which of modes or keymap parent should have priority?
In https://github.com/atlas-engineer/nyxt/commit/07da8cd9dc6a5df9fd419f2ef99913e807d004d4 @aartaka rightfully pointed out that there our key binding resolution may sometimes lead an unexpected result, namely because a the binding in the scheme parent before looking for it in the next mode.

Bug or feature?  Shall we address this for 3.0?

See also https://github.com/atlas-engineer/nyxt/issues/1841.

Ambrevar:
Fixes #2307.

@aadcg @aartaka Can you confirm?


aadcg:
It works @Ambrevar! Thanks :)

PR #2313, opened 2022-05-27
Fix script support

### LINKED ISSUE #2256: misleading --script documentation
**Describe the bug**
As of Version 3-853-g5260eabf the documentation states 

> You can even make scripts. Here is an example foo.lisp:
```lisp
#!nyxt --script
(format t "~a~&" +version+)
```

However this is misleading since it most systems don't support the use of shebangs without a full path. I have 2 proposed solutions.

### Option1
Change the code snippet to
```
#!/usr/bin/env -S nyxt --script
(format t "~a~&" +version+)
```
This has the benefit of still being a **oneline shebang** without too much fuss however it comes at the cost of portability since the `-S` flag is **not POSIX compliant**.

### Option 2
Change the code snippet to
```
#!/bin/sh
#|
exec nyxt --script $0
|#
(format t "~a~&" +version+)
```
Inspired by the method used by the Roswell project. 

This has the benefit of (as far as I can tell) being **POSIX compliant**. However this  is **longer a one liner** and **increases the overhead** since it's also starting up a shell though idk if performance is really a priority here.  

There is always the alternative of having users give the full path themselves but that seems counter intuitive to me. Personally I prefer option 2 even if it is a little verbose.
 
Simply run the command manual from within Nyxt and navigate to the Scripting heading.

Ambrevar:
Fixes #2256.

To test: Does the test work in the CI?

### DIFF
--- a/tests/executable/scripts.lisp
+++ b/tests/executable/scripts.lisp
@@ -1,0 +1,18 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt/tests)
+(use-nyxt-package-nicknames)
+
+(plan nil)
+
+(defvar *executable* (asdf:system-relative-pathname :nyxt "nyxt"))
+
+(subtest "Scripts"
+  (dolist (script (uiop:directory-files (asdf:system-relative-pathname :nyxt "examples/scripts")))
+    (prove:is (uiop:run-program (list *executable* "--script" (uiop:native-namestring script))
+                                :error-output t
+                                :output t)
+              nil)))
+
+(finalize)
--- a/tests/executable/scripts.lisp
+++ b/tests/executable/scripts.lisp
@@ -9,6 +9,12 @@
 (defvar *executable* (asdf:system-relative-pathname :nyxt "nyxt"))
 
 (subtest "Scripts"
+  (prove:is (uiop:run-program (list *executable*
+                                    "--script" (uiop:native-namestring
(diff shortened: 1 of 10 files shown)


aartaka (approved):

tests/executable/scripts.lisp
```diff
@@ -0,0 +1,18 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt/tests)
+(use-nyxt-package-nicknames)
```
Ambrevar, line 5: Move this to tests/package.lisp.

tests/executable/scripts.lisp
```diff
@@ -0,0 +1,18 @@
...
+(plan nil)
+
+(defvar *executable* (asdf:system-relative-pathname :nyxt "nyxt"))
+
+(subtest "Scripts"
+  (dolist (script (uiop:directory-files (asdf:system-relative-pathname :nyxt "examples/scripts")))
+    (prove:is (uiop:run-program (list *executable* "--script" (uiop:native-namestring script))
+                                :error-output t
+                                :output t)
+              nil)))
```
Ambrevar, line 16: Does this raise a condition?
If not, we should make sure error-output is NIL.

Ambrevar:
OK, don't merge yet, I have do test if the CI does the right thing here.

Ambrevar:
Merged with 4b7a2c5c8d12bef0718df47bd3101279c89a6117.

jmercouris:
Thanks a lot Pierre!

Issue #2302, opened 2022-05-26
Make modes hideable on status-area


aartaka:
**Is your feature request related to a problem? Please describe.**

After #2253, there appeared a bunch of modes that supercede web-mode. Even when having one-letter glyphs, those take up too much space. How about hiding some of those altogether if one configures it?

*Describe the solution you'd like**
Add a boolean slot for whether the mode should be displayed in the status area. 

**Describe alternatives you've considered**
A hack, that mostly does the same is setting `glyph` to `""`. But this is not perfect, as there's still some spacing/padding between mode glyphs,  so it looks bad.

CC @aadcg

Ambrevar:
Agreed and actually we already have an issue related to this, see https://github.com/atlas-engineer/nyxt/issues/2218.

Ambrevar:
I mean, we can fix both together.

aadcg:
Agree!

(@aartaka supersede is the right spelling :p)

aartaka:
> Agree!
> 
> (@aartaka supersede is the right spelling :p)

Ah damn, I always get it wrong :/

aadcg:
The documentation slot says:

> A glyph used to represent this mode, if unset, it will be dynamically calculated as the first letters of the mode name.



This is not true today. Was it true before? Do we care about it? To my mind, it's not very important.

CC @jmercouris (see 99604753de265142f08b73ccddbec231a462e7fb)

aadcg:
@aartaka, the slot already exists - `visible-in-status-p`. The question is which should be hidden by default. I'll do it my way, and then we'll discuss in the PR.

Ambrevar:
Which idea and which docstring?

aadcg:
> Which idea and which docstring?

This docstring:

> A glyph used to represent this mode, if unset, it will be dynamically calculated as the first letters of the mode name.

The idea is that when glyphs aren't set explicitly, then the appropriate mode prefix is computed.

Ambrevar:
Oh sure.

jmercouris:
Yes, we have foregone this idea.

PR #2314, opened 2022-05-27
Change default visible modes on status buffer.

aadcg:
Addresses #2302.

I know this will start a flame war, so bring it on! @jmercouris @aartaka @pdelfino @Ambrevar.

For now, I'd like to sign a pack of non-aggression with @aartaka, by showing the screenshot below. @aartaka, notice the glyphs on the status buffer!


source/mode/file-manager.lisp
```diff
@@ -16,7 +16,8 @@
 
 (define-mode file-manager-mode (nyxt/prompt-buffer-mode:prompt-buffer-mode)
   "Prompt buffer mode for file choosing."
-  ((keymap-scheme
+  ((visible-in-status-p t)
```
aartaka, line 19: I'd say that all the prompt-buffer modes should be visible by default.

source/mode/certificate-exception.lisp
```diff
@@ -14,7 +14,8 @@ See the `add-domain-to-certificate-exceptions' command.")
   "Enable ignoring of certificate errors.
 This can apply to specific buffers.
 See the `add-domain-to-certificate-exceptions' command."
-  ((certificate-exceptions *default-certificate-exceptions*
+  ((visible-in-status-p t)
```
aartaka, line 17: Not sure about this one, because `certificate-exception-mode` is quite invisible in its functioning.

source/mode/repeat.lisp
```diff
@@ -7,7 +7,8 @@
 
 (define-mode repeat-mode (nyxt/process-mode:process-mode)
   "Mode to repeat a simple action/function repetitively until stopped."
-  ((rememberable-p nil)
+  ((visible-in-status-p t)
```
aartaka, line 10: Repeat mode is a service mode for mostly one-shot usage. I believe it should stay invisible.

source/mode/repl.lisp
```diff
@@ -7,7 +7,8 @@
 
 (define-mode repl-mode ()
   "Mode for interacting with the REPL."
-  ((rememberable-p nil)
+  ((visible-in-status-p t)
```
aartaka, line 10: Much like all the other page-bound modes, `repl-mode` doesn't add much information when displayed.

aartaka: But they, it's a kind of a major mode...

aartaka (approved):

source/mode/file-manager.lisp
aadcg: I actually thought about it again, and I changed my mind. Don't you think that there's no ambiguity is what the prompt-buffer does, regardless of the enabled modes?

It also clutters the prompt brutally, notice the screenshot below.

source/mode/repeat.lisp
aadcg: Oh right, I took this to be smth like `watch`.

source/mode/repl.lisp
aadcg: Indeed, that was my line of reasoning... it's a "major mode".

source/mode/file-manager.lisp
Ambrevar: `prompt-buffer` and `hint-prompt-buffer` can be hidden.  However `emacs` and `my` are important. Maybe then hide all subclasses of `prompt-buffer-mode`?

aadcg: Agreed!

aadcg: I noticed that, regardless of the value of `visible-in-status-p`, it had no effect in `prompt-buffer-mode`.

I fixed it in 073672fa3, but there's the following problem:

```
; caught ERROR:
;   
;     Package NYXT/PROMPT-BUFFER-MODE does not exist.
```

The issue is that when `source/prompt-buffer` is compiled, it's not aware of `source/mode/prompt-buffer`. Any ideas @Ambrevar @aartaka?

Ambrevar: Fixed in #2321.

aadcg: Got it.

aadcg:
It seems done.  Can I squash and merge?

Ambrevar:
Looks good to me!

aadcg:
> I disagree here. When someone makes a new mode, unless they set the value of this slot, it will be invisible. I found this behavior to be surprising. I will enable a new mode I make, and nothing will happen.

It's invisible, in the sense that it won't appear in the status buffer. Notice, however that you can't possible see all of the active modes in that tiny space! That's why we have a button (the "+") that shows all of the actives modes. 

When you say that when you enable a new mode nothing happens, I can agree that a message show be shown at the echo area.

> I don't disagree with the changes on default visibility, that I am fine with. However, I think the default value of this slot should be t.

Perhaps you have a point here. I did the other way since there are more cases when we want to hide it.

aadcg:
@Ambrevar @jmercouris do you agree with 9eec5513e?

Ambrevar:
Not sure: we used to log:info the enabling of mode but we did remove it because it was too spammy.
(Now it's a log:debug :p)

One way to mitigate this would be to only display the message if the mode is invisible.

But... Anyways, even if the mode is displayed in the status area, it may very well be off-screen because of too many modes.  So the only _good way_ to check if a mode is active is to call `toggle-modes` or the like.

So I withdraw my agreement: I think the default value does not really matter in the end.

Ambrevar:
I'm fine with `t`.  What matters is that we won't change the default for `3-series`.

aadcg:
Ok, I can squash and merge this. We don't change the dafault value, but I think we should revise the default visibility value for all modes before releasing 3.0, so that we don't change mode-specific values in the 3.- series (I'll create an issue).

PR #2315, opened 2022-05-27
Demote some core modes as non-core modes

Ambrevar:
Still trying to make the Nyxt core as micro as possible... :)


aartaka:
Maybe we should even move some modes to extensions?

aadcg:
> Maybe we should even move some modes to extensions?

I second that. Especially:

- `no-procrastinate`, that extends `blocker-mode`;
- `tts`;

Notice that marking something as an extension doesn't mean that we must take it outside of our repository. There could be extensions that we ship together with nyxt.

aadcg:
I'm ok with this change, philosophically @Ambrevar.

Let me see test it to see if it breaks something terribly obvious.

aadcg:
Seems to ok @Ambrevar.

PR #2317, opened 2022-05-29
Replace define-function with reader macros.

Ambrevar:
`define-function` is too specific and only used in 2 spots.
Reader macros are a more general solution.

This saves 10 lines of code and 1 macro :)


aartaka (approved):
Yes, looks just fine to me!

PR #2318, opened 2022-05-29
small-web: Refactor Gopher

hendursaga:
Some refactoring to prevent a few duplicate calls, one doc fix, and possibly one or two bug fixes. I'm not sure how to "select" or highlight parts of this diff, but the :href change from `when` to `if` might be incorrect, I'm not sure what the original intent was in potentially evaluating to a `nil`.

### DIFF
--- a/source/mode/small-web.lisp
+++ b/source/mode/small-web.lisp
@@ -120,8 +120,7 @@ Second return value should be the MIME-type of the content."))
 (defmethod line->html ((line cl-gopher:info-message))
   (let ((line (cl-gopher:display-string line)))
     (spinneret:with-html-string
-      (if (or (str:emptyp line)
-              (every #'sera:whitespacep line))
+      (if (str:blankp line)
           (:br)
           (:pre line)))))
 
@@ -313,9 +312,10 @@ Please, check URL correctness and try again.")))
               ((:input :sensitive-input)
                (let ((text (quri:url-encode
                             (handler-case
-                                (prompt1 :prompt meta
-                                         :sources (list (make-instance 'prompter:raw-source))
-                                         :invisible-input-p (eq status :sensitive-input))
+                                (prompt1
+                                  :prompt meta
+                                  :sources (list (make-instance 'prompter:raw-source))
+                                  :invisible-input-p (eq status :sensitive-input))
                               (nyxt::nyxt-prompt-buffer-canceled () "")))))
                  (buffer-load (str:concat url "?" text) :buffer buffer)))
               (:success
--- a/source/mode/small-web.lisp
+++ b/source/mode/small-web.lisp
@@ -41,7 +41,7 @@ loading, you'd need to override `line->html' in the following way:
         (cl-gopher:display-string line))))
 
 \(defmethod line->html ((line cl-gopher:image)) (image->link line))
-\(defmethod line->html ((line cl-gopher:gif)) (image->html line))
+\(defmethod line->html ((line cl-gopher:gif)) (image->link line))
 \(defmethod line->html ((line cl-gopher:png)) (image->link line))
 
 Gemini support is a bit more chaotic, but you can override `line->html' for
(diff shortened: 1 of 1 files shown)


source/mode/small-web.lisp
```diff
@@ -118,22 +118,24 @@ Second return value should be the MIME-type of the content."))
...
 (defmethod line->html ((line cl-gopher:info-message))
-  (spinneret:with-html-string
-    (if (or (str:emptyp (cl-gopher:display-string line))
-            (every #'sera:whitespacep (cl-gopher:display-string line)))
-        (:br)
-        (:pre (cl-gopher:display-string line)))))
+  (let ((line (cl-gopher:display-string line)))
+    (spinneret:with-html-string
+      (if (or (str:emptyp line)
+              (every #'sera:whitespacep line))
```
aartaka, line 124: Note that there's a `str:blank?` with an alias of `str:blankp`, you can utilize either of those here.

source/mode/small-web.lisp
```diff
@@ -310,8 +314,8 @@ Please, check URL correctness and try again.")))
                (let ((text (quri:url-encode
                             (handler-case
                                 (prompt1 :prompt meta
-                                  :sources (list (make-instance 'prompter:raw-source))
-                                  :invisible-input-p (eq status :sensitive-input))
+                                         :sources (list (make-instance 'prompter:raw-source))
```
aartaka, line 317: Wait, does `prompt1` indent arguments this way? The last time I used it, there was a (counter-intuitive, yet automatic) indentation like in the deleted diff...

aartaka (approved):
Thanks for this, it all looks much better now!

Regarding the `when`: the alternative branch was extremely improbably given strict format Gopher lines have, so I did't even bother with adding a sensible alternative branch. But still, yours is a much better version! (´ ▽｀).。ｏ♡

source/mode/small-web.lisp
Ambrevar: Yeah, it's a macro, the original indent is the correct one.

aartaka: @Ambrevar, now that we're at it, we can use `&rest body` in lambda list and SLY will indent things like it does for a function :D

aartaka: Oh, no, it won't (#^.^#)

hendursaga: So... revert this particular hunk? We honestly should add something to `.dir-locals.el` so that it indents properly even without a SLY connection, if that's possible..

source/mode/small-web.lisp
```diff
@@ -185,8 +189,8 @@ Second return value should be the MIME-type of the content."))
         (spinneret:*html-style* :tree))
     (values (spinneret:with-html-string
               (:style (style (current-buffer)))
-              (:style (when (find-submode 'small-web-mode)
-                        (style (find-submode 'small-web-mode))))
+              (:style (alex:when-let (submode (find-submode 'small-web-mode))
+                        (style submode)))
```
hendursaga, line 193: Question: is there a problem with having two `:style`s here? As in, does the latter override the former, or do they combine, or what? Just checking.

source/mode/small-web.lisp
hendursaga: OK, no need to revert. I can just put :prompt on its own line and the macro gets indented correctly.

Ambrevar: This is not related to `.dir-locals` but to the convention of indenting macros vs. functions.

In this case, the old indentation was that of a macro, while the new one a function.

SLY and SLIME will do the right thing when they know the symbol type (macro or function).  If they don't (e.g. the system is not loaded and thus the symbol not defined), then they assume a _function_.

source/mode/small-web.lisp
aartaka: No, those two combine. We use them to apply both a generic style (common for all buffers of a certain class) and the mode-specific style to a certain page.

aartaka:
@hendursaga, you done? I'm merging then?

hendursaga:
@aartaka yes, I'm done, go ahead :)

Issue #2164, opened 2022-03-26
PDF support


aartaka:
**Is your feature request related to a problem? Please describe.**
Lots of browsers have built-in PDF-viewer (even if it's just a barebones [mozilla/pdf.js](https://github.com/mozilla/pdf.js)), and that's a valid expectation that Nyxt, as a browser, has it too. 

**Describe the solution you'd like**
Add a Nyxt-native PDF-reader, most likely as a custom scheme (Epiphany does that) embedding pdf.js reader.

**Describe alternatives you've considered**
- Add PDF-specific hooks and expand documentation so that one can easily configure their favorite PDF reader to open once a PDF is being downloaded.
- Hope for an extension with PDF support.
  - However, extensions are quite an indirect way to do thing, and PDF support is better as a built-in feature. 

**Additional context**
Epiphany does it this way:
- Add a custom `ephy-pdf:` scheme.
- When the content type of the link is PDF, redirect to `ephy-pdf:` page.
- When on `ephy-pdf:`, load a slightly customized [pdf.js template viewer](https://github.com/mozilla/pdf.js/blob/master/web/viewer.html) with a reference to the PDF being loaded.
- Et voila! pdf.js renders the PDF and allows interacting with it conveniently.

Ambrevar:
On a related note, there is pure CL library to generate PDF: https://github.com/mbattyani/cl-pdf

hendursaga:
Stop the presses!

The [new stable WebKitGTK release](https://webkitgtk.org/2022/09/16/webkitgtk2.38.0-released.html) supports PDF documents out-of-the-box using PDF.js!

aartaka:
Whaaaaat! No way...

jmercouris:
That's incredible!

shaunsingh:
Considering gnome web on webkitgtk supports pdf's via pdf.js on webkitgtk 2.38 and onwards, I think this should be doable within nyxt now. 

What would be needed to implement pdf support in nyxt? I tried looking through the source to see how nyxt handles loading/displaying images, but I didn't find anything. 

For what its worth, the pdf.js demo works just fine in nyxt, but I think implementing vim bindings ala zathura would be nice

jmercouris:
I don't believe we'll need to do anything other than wait to package Nyxt with WebKitGTK 2.38+

aadcg:
Does #2988 change the status of this issue?

PR #2319, opened 2022-05-30
Make downloads more WebKit-ish.

### LINKED ISSUE #2237: Cannot download files (PREPROCESS-REQUEST issue)
If you go to [ubuntu.com](https://ubuntu.com) and click "Download" and then "22.04 LTS" you will get something like this after a while:
20220508_10h18m13s_grim

The bug was introduced in 83629e4

aartaka:
# Description

This makes downloads to re-use existing WebKit infrastructure, and allows to download things with the `"download-started"` WebKitWebContext signal, which makes previously inaccessible downloads possible now!

While it does not yet address #2237, it fixes bugs that were biting me for a long time (like being unable to download anything from Telegram Web).

This also rewamps the request processing, allowing for more flexibility in request redirection (notable highlight: one can now redirect the download URL and make it point to something else). Given this, we can, for example, enable #2164 and other niceties.

# Discussion

- We store `download-directory` in the buffer, but this change renders it quite useless -- most downloads go though WebKitWebContext anyway, which makes them a browser-level phenomena. How about we move the downloads directory to the `browser`?
- I've moved the `request-resource-keys` processing to `on-signal-decide-policy`, but realized that I've never seen or experienced it used. Anywhere. At all. Do we actually need it? Maybe remove it before 3.0?

# Checklist:

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -451,45 +451,6 @@ view.")
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
 
-(defun preprocess-request (request-data)
-  "Deal with REQUEST-DATA with the following rules:
-- If a binding matches KEYS in `request-resource-scheme', run the bound function.
-- If `new-window-p' is non-nil, load in new buffer.
-- If `known-type-p' is nil, download the file.
-- Otherwise let the renderer load the request."
-  (with-slots (url buffer keys) request-data
-    (let* ((keymap (scheme-keymap buffer (request-resource-scheme buffer)))
-           (bound-function (the (or symbol keymap:keymap null)
-                                (keymap:lookup-key keys keymap))))
-      (declare (type quri:uri url))
-      (cond
-        ((not (valid-scheme-p (quri:uri-scheme url)))
-         (uiop:launch-program (list *open-program* (quri:render-uri url))))
-        (bound-function
-         (log:debug "Resource request key sequence ~a" (keyspecs-with-optional-keycode keys))
-         (funcall bound-function :url url :buffer buffer)
-         nil)
-        ((new-window-p request-data)
-         (log:debug "Load URL in new buffer: ~a" (render-url url))
-         (open-urls (list url))
-         nil)
-        ((and (not (known-type-p request-data))
-              (toplevel-p request-data))
-         (log:debug "Buffer ~a initiated download of ~s." (id buffer) (render-url url))
-         (funcall (resolve-symbol :download :function)
-                  buffer url
-                  :proxy-url (proxy-url buffer :downloads-only t)
-                  :cookies "")
-         ;; TODO: WebKitGTK emits "load-failed" if we call
-         ;; webkit-policy-decision-ignore on a download requestion.
-         ;; To work around this, we set the `status' to a value other than
-         ;; `:loading'.
-         (setf (slot-value buffer 'status) :finished)
-         nil)
-        (t
-         (log:debug "Forwarding ~a for buffer ~s" (render-url url) buffer)
-         request-data)))))
-
 (export-always 'url-dispatching-handler)
 (-> url-dispatching-handler
     (symbol
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -1685,9 +1685,10 @@ local anyways, and it's better to refresh it if a load was queried."
       (setf (nyxt/download-mode:completion-percentage download)
             (* 100 (webkit:webkit-download-estimated-progress webkit-download))))
     (connect-signal download "decide-destination" nil (webkit-download suggested-file-name)
-      (alex:when-let* ((download-dir (or (download-directory
-                                          (find (webkit:webkit-download-get-web-view webkit-download)
-                                                (buffer-list) :key #'gtk-object))
+      (alex:when-let* ((download-dir (or (ignore-errors
+                                          (download-directory
+                                           (find (webkit:webkit-download-get-web-view webkit-download)
+                                                 (buffer-list) :key #'gtk-object)))
                                          (make-instance 'download-directory)))
                        (download-directory (files:expand download-dir))
                        (native-download-directory (unless (files:nil-pathname-p download-directory)
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -799,10 +799,19 @@ See `gtk-browser's `modifier-translator' slot."
       (gobject:g-signal-connect
        context "initialize-web-extensions"
        (lambda (context)
-         (with-protect ("Error in signal thread: ~a" :condition)
+         (with-protect ("Error in \"initialize-web-extensions\" signal thread: ~a" :condition)
            (webkit:webkit-web-context-set-web-extensions-directory
             context
             (uiop:native-namestring gtk-extensions-path))))))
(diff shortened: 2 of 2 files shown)


aartaka:
@Ambrevar, @jmercouris, any feedback? It preserves all the functionality, while allowing for more, so I don't see much problem with this patch :)

source/renderer/gtk.lisp
```diff
@@ -799,10 +799,19 @@ See `gtk-browser's `modifier-translator' slot."
...
             context
             (uiop:native-namestring gtk-extensions-path))))))
+    (gobject:g-signal-connect
+     context "download-started"
+     (lambda (context download)
+       (declare (ignore context))
+       (with-protect ("Error in \"download-started\" signal thread: ~a" :condition)
+         ;; FIXME: Right now we cancel the download and start a new one from
+         ;; the Lisp side. We should rather wrap the DOWNLOAD into the
+         ;; Lisp-side structure.
```
Ambrevar, line 813: Which structure?  `nyxt/download-mode:download`?  
Specify it in the comment.

source/renderer/gtk.lisp
```diff
@@ -1669,7 +1688,10 @@ local anyways, and it's better to refresh it if a load was queried."
       (setf (nyxt/download-mode:completion-percentage download)
             (* 100 (webkit:webkit-download-estimated-progress webkit-download))))
     (connect-signal download "decide-destination" nil (webkit-download suggested-file-name)
-      (alex:when-let* ((download-dir (download-directory buffer))
+      (alex:when-let* ((download-dir (or (download-directory
+                                          (find (webkit:webkit-download-get-web-view webkit-download)
+                                                (buffer-list) :key #'gtk-object))
```
Ambrevar, line 1693: Why not keep `(download-directory buffer)`?

source/browser.lisp
```diff
@@ -451,45 +451,6 @@ view.")
...
-         ;; TODO: WebKitGTK emits "load-failed" if we call
-         ;; webkit-policy-decision-ignore on a download requestion.
-         ;; To work around this, we set the `status' to a value other than
-         ;; `:loading'.
-         (setf (slot-value buffer 'status) :finished)
-         nil)
-        (t
-         (log:debug "Forwarding ~a for buffer ~s" (render-url url) buffer)
-         request-data)))))
-
```
Ambrevar, line 492: Wait, why are you removing this?  This is one of the most essential request filtering points of our browser! O.o

It handlers

- `*open-program*`
- `new-window-p`
- `request-resource-scheme`.

Ambrevar:
> * I've moved the `request-resource-keys` processing to `on-signal-decide-policy`, but realized that I've never seen or experienced it used

I assume you mean `request-resource-scheme`.  We don't call it but it's useful to that the user can customize keys like `C-button1` to open a new tab.

source/renderer/gtk.lisp
aartaka: Oh wait, it's an outdated comment, I did actually wrap it into a `nyxt/download-mode:download` （＾ｖ＾）

source/renderer/gtk.lisp
aartaka: Because there's no longer `buffer` set rigidly. Note that this function is used in `make-context` too — and `make-context` has no reference to any buffer whatsoever.

aartaka:
> I assume you mean `request-resource-scheme`. We don't call it but it's useful to that the user can customize keys like `C-button1` to open a new tab.

But then, they can do the same with regular keybindings, can't they?

Ambrevar:
They _are_ regular bindings, it's just a keymap.

It makes little sense to add these bindings to a mode.  Or maybe a `renderer-request-mode` or something?

source/renderer/gtk.lisp
aartaka: No, the buffer that initiated te download can do so in the background, so using `current-buffer` is not a way either.

Ambrevar:
It's not the same.  If you add C-button1 to document mode, then you need to:

- Capture the link at point.
- Open a new buffer with that link.

More generally, `request-resource-scheme` is here to follow the WebKitGTK API which passes along the pressed keys during a request event.  The functions bound to these keys must follow a specific signature, namely `&key url buffer`.

Does that make sense?

source/browser.lisp
Ambrevar: Oh, I see it now.
But why do you say we repeated ourselves?

aartaka:
Yes, it does. 

Anyway, this does still work as it used to, as I merely moved the code from one function to another :)

source/browser.lisp
Ambrevar: Still, shall we really remove `preprocess-request`?
One of its main roles was precisely to prevent the user from shooting themselves in the foot, and by this I'm talking exposing themselves to massive security issues.

Indeed, if you get the request handlers wrong, you can potentially allow a 3rd party website to take control of your user session.

An important example of this is

```lisp
((not (valid-scheme-p (quri:uri-scheme (url request-data))))
             (uiop:launch-program (list *open-program* (quri:render-uri (url request-data)))))
```

aartaka: But basically nothing changed -- we still ignore the request with an invelid scheme and open in with `xdg-open`, like we used to.

aartaka: > But why do you say we repeated ourselves?

Because the logic in `preprocess-request` is quite similar to the one in `on-signal-decide-policy`, and it was basically an earlier yet same idea of link filtering.

Ambrevar: > But basically nothing changed -- we still ignore the request with an invelid scheme and open in with xdg-open, like we used to.

I disagree: before the hook wouldn't run for instance on invalid schemes.  Now it always does.

aartaka: > > But basically nothing changed -- we still ignore the request with an invelid scheme and open in with xdg-open, like we used to.
> 
> I disagree: before the hook wouldn't run for instance on invalid schemes. Now it always does.

Yes. And I consider it a virtue, because previously one had to rely on us to implement the most rensible action for their scheme, now they can do the most sensible action themselves, even if we don't support the scheme. It's much more flexible, while having the same degree of security.

Ambrevar: If the user wants to add a new scheme, they register it and it won't be marked as `invalid`.  So checking for scheme validity _before_ running the hooks does not remove any customizability for custom schemes.

aartaka: Yes, but still, running hooks before we do any decision on the URL is better than doing any preemptive decisions that one does not ask for (we did that with downloads, for example, basically disallowing redirections from any unsupported content, including PDFs, to some other URL).

aartaka: So, when I wanted to work on PDF support, I wanted to redirect PDF links to `pdf:` schemed ones. But `preprocess-request` was starting download before I was able to do anything to the URL. So let's allow more hookability there.

aartaka:
I've simplified the order of events even more: now every requests goes through the `pre-request-hook` and `request-resource-hook` and then goes through all the URL checks (both the `preprocess-request` ones and `on-signal-decide-policy` native ones). This way, one can configure Nyxt to do whatever to the requests, while still having all the guarantees of `preprocess-request` :D

Ambrevar:
Fair enough :)

jmercouris:
Thank you Artyom!

PR #2320, opened 2022-05-30
Package locks

### LINKED ISSUE #2316: Lock package :nyxt and all subpackages?
Seems like a good idea to prevent accidental modification of the :nyxt package and its packages. 

Thoughts?

Note: Only supported on SBCL.

Ambrevar:
Fixes #2316.

This does a very simple thing: lock all packages.
Note that this is an SBCL extension.  Package locking exists in a few other implementations, but not in CCL.  See also https://github.com/elliottjohnson/cl-package-locks/.

This simple change allowed us to catch many errors.
It also forces us to be a bit stricter with the load order, hence the split of the `help.lisp` file.  Let me know what you think.

Finally, only 2 parts are left a bit "unclean":
- web-extensions.lisp uses `nyxt` but from a different package _and_ it sets the `nyxt:background-buffer` accessor, which violates the lock.  For now I've opted for allowing `:nyxt` package modifications from `nyxt/web-extensions`, but this is a rather general issue.
- Same thing with `download-mode:renderer-download` in GTK, but really this would not be a problem if the `renderer-download` was part of the core.

Thoughts?

Note: This PR is easier to review when looking at individual commits.


Ambrevar:
OK to merge?

source/debug.lisp
```diff
@@ -1,13 +1,13 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-(uiop:define-package :nyxt/debug
-  (:use :common-lisp :nyxt)
+(uiop:define-package :nyxt/debug ; TODO: Rename to `nyxt/develop' to avoid confusiong with `debugger'?
```
jmercouris, line 4: No, I think it is fine.

Ambrevar:
Yes, locks are very much like exports.  It does not  forbid anything, just adds extra safety belt to avoid shooting yourself in the foot.

Plus it catches lots more typo / package errors!

Ambrevar:
Merged with 705d570782e5af3d03850da598bbd33a34c90ce4.

Issue #2219, opened 2022-05-03
Prompt buffer VI formatting should instead call a mode-extensible generic


Ambrevar:
Currently we hard-code the VI formatting in the status formatter.  Instead, we should call a mode-specialized generic function, which VI-mode would then specialize.

Easy fix.

Ambrevar:
On a similar topic, `print-status` should not be called manually.  It's too specific and too brittle.

Instead, a status-buffer should register a bunch of object to watch.  So if the status displays the list of modes, then it should `watch` for changes in the modes of the current buffer.

Is there a generic way to "watch for changes" in CL?
I cannot think of anything in the language or in a library, so what about this little "hack":

```lisp
(let ((method (defmethod (setf modes) :around (value (buffer modable-buffer))
                (declare (ignorable value buffer))
                (prog1 (call-next-method)
                  (print-status)))))
  (push method (watchers STATUS-BUFFER)))
```

In the above snippet, we stored the setf-around method so that it can be removed if need be.

The drawback of this approach is that it won't work if the setf-around method already exists.

Any other idea?

Ambrevar:
Other drawback of the previous snippets: it only works on setf-able places.

aadcg:
I agree, and noticed it today too. I'm just not sure if it's worth the effort just yet, see #2324.

PR #2321, opened 2022-05-30
Generic / CLOS-based status-buffer API + untangled `vi-mode` out of it

### LINKED ISSUE #2218: nyxt::mode-status for extensions should be independent of glyphs
We need to change the logic.  This is relatively easy.

Ambrevar:
Fixes #2218, #2219.

This removes some 50 lines of code! :)

To do:

- [x] Do not display all these modes in the prompt-buffer, and don't display them as buttons.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -602,6 +602,15 @@ Return BUFFER."
   "Return the current prompt-buffer."
   (first (active-prompt-buffers (current-window))))
 
+(export-always 'focused-buffer)
+(defun focused-buffer (&optional (window (current-window)) )
+  "Return the current prompt-buffer."
+  ;; TODO: Add message-buffer when we have the slot in `window'.
+  (find-if #'ffi-focused-p
+           (list (first (active-prompt-buffers window))
+                 (active-buffer window)
+                 (status-buffer window))))
+
 (defmethod write-output-to-log ((browser browser))
   "Set the *standard-output* and *error-output* to write to a log file."
   (let ((buffer (current-buffer)))
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -604,7 +604,7 @@ Return BUFFER."
 
 (export-always 'focused-buffer)
 (defun focused-buffer (&optional (window (current-window)) )
-  "Return the current prompt-buffer."
+  "Return the currently  focused buffer."
   ;; TODO: Add message-buffer when we have the slot in `window'.
   (find-if #'ffi-focused-p
            (list (first (active-prompt-buffers window))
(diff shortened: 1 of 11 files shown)


Ambrevar:
In commit 7c8748c63b2f735405f30139396dda465780fbc5 I tried to fix a long-standing issue with VI-mode, namely that clicking on the status bar while in `vi-normal-mode` does not work.

In this commit, I forward the even to the focused buffer instead of the current-buffer.

Sadly, the status buffer is focused only _after_ the even has been sent to the renderer.  So we need to click _twice_ for it to work.

The proper fix here is to query GTK for the widget the mouse event is hovering on.  I know that GTK can do this, but I can't find the API call.

Any idea?

source/browser.lisp
```diff
@@ -602,6 +602,15 @@ Return BUFFER."
   "Return the current prompt-buffer."
   (first (active-prompt-buffers (current-window))))
 
+(export-always 'focused-buffer)
+(defun focused-buffer (&optional (window (current-window)) )
+  "Return the current prompt-buffer."
```
jmercouris, line 607: This docstring is incorrect.

jmercouris:
I'm getting a very strange rendering issue on this branch.

jmercouris:
Screenshot_20220601_031545

Is this behavior expected?

jmercouris:
I'm going to assume no, and you are showing all of the modes in the prompt buffer even when they are not exactly relevant :-)

jmercouris:
I see now your comment in another thread:

```
For now it defers to the status-buffer method, but what we could do it simply specialize the method for prompt-buffer and only display keymap-scheme-modes.

That said, we have too many modes in the prompt-buffer, we should probably remove some.
```

it makes sense. I also see you did not check the box. I had misunderstood your comment here: 

```
We can still merge this and fix the VI issue later.
```

as an implicit statement that this PR is ready. I see now that it is not!

aadcg:
@Ambrevar, I just wanted to say that the goal of this PR is very much needed and welcome!  I'm afraid I can't help with the GTK API call.

source/browser.lisp
```diff
@@ -604,7 +604,7 @@ Return BUFFER."
 
 (export-always 'focused-buffer)
 (defun focused-buffer (&optional (window (current-window)) )
-  "Return the current prompt-buffer."
+  "Return the currently  focused buffer."
```
Ambrevar, line 607: Good catch! :)

Ambrevar:
Merged with 59e3425c3ef87042af648005d201a0c6a2c95f67.

Ambrevar:
I've opened an issue about the VI / status click problem here: https://github.com/atlas-engineer/nyxt/issues/2343.

PR #2322, opened 2022-05-30
Context menu fix

### LINKED ISSUE #1524: Bug: Keepass/Password Integration, Mouse-related stuff.
I'm not really a "poweruser" doing everything with the keyboard (though I'm trying to do more), I'm playing around with emacs and I like open software so I'm very glad for the project as it is and look forward to this developing. This also means I'm maybe experiencing some stuff other's wouldn't bc I'm actually using the mouse to a bigger extend. 
Some issues with that:

1) The popup menu when rightclicking a link gives the option "open new window" but it will be a new buffer - which i was looking for but it's confusing. not that a "new window" option would be bad
2) I can actually rightclick URLs in the Powerline like normal links. I feel that's not really planned - it won't work. `Copy link location` will give me `lisp://%28SET-URL%29`  --- you guys know better than me what that means :D actually being able to retrieve the current URL with a single click would be nice though. 

One major thing i would need for Nyxt to work for me on a ground level (for simple tasks where I don't need firefox) is the password manager. I use KeepassXC, and I was very happy to read it is automatically supported. I haven't gotten it to work tho - here are my issues just trying to log into github:
 * when I'm on the login page, command "copy" password  will ask for a file in my user direcory, although the file is stored in  my nextcloud sync folder. From here I can't move there though - I can't open folders below "~/" through the selection in the promptbuffer. 
 * I managed to look up the documentation and entered the correct file location. Next try I am prompted for my masterpassword. I type it in but it won't matter. I don't get an error, just a new empty password prompt.  I guess the client isn't recognized? I will look into that,  I guess . 
* finally, quite the killer bug: Now I open the keepassxc application, to copy over the password by hand. But the password field will actually remain empty (even when i   paste via mouse i will get a  wrong password error from gh). 
* on improvement: it's an issue activating the client, i would like troubleshooting about that at least. But also there are some issues that seem to be more general, i think. Also, what does setting the "masterpassword" slog  (no doc)? In general, having documentation about the options when having booleans would be good think, they are not so intuitive in lisp.  
 
These would be the daring issues rn for me. I also have some more interface/improvement and feature kind of thoughts, i think opening new PRs would be best.

Hope it's not on configurations, I haven't updated since i installed full release. 
For the record I'm running Nyxt 2.0.0 on Manjaro Linux (KDE plasma 5.21.5.) I installed it via the Guix package manager (don't know how that affects stuff).

aartaka:
# Description

This renames the long-irritative "Open in New Window" to "Open in New Buffer", and returns the "Download Link to Disk" action that should work after #2319 is merged.

Fixes #1524.

# Discussion

This requires `cl-webkit` with https://github.com/joachifm/cl-webkit/commit/39388f55bfaf718a79cc7545cf12e62d17e4aed7, should I make a new release?

# Checklist:

### DIFF
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -1451,9 +1451,15 @@ See `finalize-buffer'."
     (let ((length (webkit:webkit-context-menu-get-n-items context-menu)))
       (dotimes (i length)
         (let ((item (webkit:webkit-context-menu-get-item-at-position context-menu i)))
-          (match (webkit:webkit-context-menu-item-get-stock-action item)
-            (:webkit-context-menu-action-download-link-to-disk
-             (webkit:webkit-context-menu-remove context-menu item))))))
+          (when (eq (webkit:webkit-context-menu-item-get-stock-action item)
+                    :webkit-context-menu-action-open-link-in-new-window)
+            (webkit:webkit-context-menu-remove context-menu item)
+            (webkit:webkit-context-menu-insert
+             context-menu
+             (webkit:webkit-context-menu-item-new-from-stock-action-with-label
+              :webkit-context-menu-action-open-link-in-new-window
+              "Open Link in New Buffer")
+             i)))))
     nil)
   (connect-signal buffer "enter-fullscreen" nil (web-view)
     (declare (ignore web-view))
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -1450,16 +1450,19 @@ See `finalize-buffer'."
     (declare (ignore web-view event hit-test-result))
     (let ((length (webkit:webkit-context-menu-get-n-items context-menu)))
       (dotimes (i length)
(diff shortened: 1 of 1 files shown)


Ambrevar:
Fantastic, thanks!

Ambrevar:
Feel free to merge.

jmercouris:
I don't think a new release will hurt, but it is also not necessary since we are using submodules. The choice is yours.

Ambrevar:
@jmercouris I think what @aartaka meant is that the cl-webkit Git submodule in the Nyxt repo must be updated when this is merged, otherwise it breaks `master`.

Ambrevar:
If you mean `build-scripts/sync-submodules-with-guix.scm`, as far as I'm aware it's been used only once, when it was created.

Since then we have only updated submodules manually.

That said, we need to re-sync everything for 3.0:
- Update the Guix CL packages.  (Serapeum, etc.)
- Then call `build-scripts/sync-submodules-with-guix.scm`.

Issue #2289, opened 2022-05-23
Rethinking the theme libray


aadcg:
I was playing around with themes today and I noticed that our default theme doesn't follow all of the recommendations related to contrast mentioned in `libraries/theme/theme.lisp`. 

I put some thought to it and  played  a bit with a nice [tool](https://reallybigshoe.co.uk/visualiser/index.html) that computes the contrasts.

Today we're here:
![2022-05-23_18:11:22](https://user-images.githubusercontent.com/45483512/169873124-4dcd5952-aa1a-4e1a-95d8-3f1d26d5bc90.png)

A suggestion:
![2022-05-23_18:10:05](https://user-images.githubusercontent.com/45483512/169873145-664adac5-ab81-444c-98dd-26eff76ef23f.png)

EDIT: Of course it'd look awful if we were to apply these colors right now, as it can be seen below. The point was to tweak the colors to meet the criteria of having overall better contrasts.

![2022-05-23_20:52:08](https://user-images.githubusercontent.com/45483512/169895549-54161147-4e64-4702-92f9-0a655a17b387.png)

![2022-05-23_20:44:18](https://user-images.githubusercontent.com/45483512/169894736-41213be6-e003-469f-94b4-2eff4703db28.png)

Reference:
[Colors with a minimum of 4.5:1 contrast ratio on both black and white](https://dev.to/finnhvman/which-colors-look-good-on-black-and-white-2pe6).

Ambrevar:
True your suggestion has better contrast, but I find the original theme more pleasing to the eye.  Just a personal preference.

aartaka:
Current color themes are both parts of Nyxt identity. While we certainly can change some tertiary or quaternary colors, the accend and primary colors are too important to shift them much. We've discussed it with @jmercouris when doing a theme library — Carribean Blue is a good an recogniseable accent color.

However, having proper color contrast is important, I acknowledge that. How about this: you design accessible theme(s) (you have good ideas on those, based on our nx-modus-themes chat) and we merge those in Nyxt, and list those together with the two default themes.

This way, we preserve Nyxt identity color scheme in a default theme, while still having an easy toggle for a more accessible variant.

aadcg:
I took some time to think about this. My idea was to port the modus themes, whose palette is in the public domain, and write an external extension. In the process, I reviewed the foundations of how we think about themes so that it's general enough for any use case.

[...]

Issue #2188, opened 2022-04-12
Low-hanging fruit: improve Nyxt stylesheets


aartaka:
Right now out CSS is quite unever between pages -- some buttons change color when hovered, some don't; some links are underlined, some aren't; some lists have default markers, some have custom markers, some have no markers at all. We need to make styling of Nyxt-specific interfaces clear and pretty. Notably:
- Make all the interactive elements hoverable, and change the element state on hover.
- Ensure proper color contrast of all the element states.
- Ensure that similar styles are abstracted into `buffer`'s `style` slot to not clutter modes with lookalike styles.
- Stick to ARIA and semantic HTML guidelines (e.g. use `alt` text for images, provide `aria-role` for atypical interfaces, add tooltips to buttons...)

Help will be much appriciated! Should not be too difficult too, given that it's mostly a simple work with CSS :)

aartaka:
More pointers: [Material Design Theming Guidelines](https://material.io/design/material-theming/implementing-your-theme.html#shape)

jmercouris:
Consider that we are pretty picky :-D but also quite open!

aadcg:
@aartaka does it still make sense to keep this one open?

aartaka:
> @aartaka does it still make sense to keep this one open?

Yes, our styling is still somewhat inconsistent and would benefit from some polish.

aadcg:
I feel that this issue is too broad and hard to extract actionable tasks. Can we re-think it? Also, we now have @lansingthomas.

jmercouris:
Agreed, too abstract.

aadcg:
@aartaka could you close it? Thanks.

aartaka:
Actually, let's make it more concrete: 
- We need use action responsiveness for interactive elements: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes#user_action_pseudo-classes
- We need location management for links and sections: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes#user_action_pseudo-classes
- And we need some input styling for our inputs: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes#location_pseudo-classes

aartaka:
I'm modifying the toplevel comment with these.

PR #2332, opened 2022-06-01
Theme refactor

aadcg:
Addresses #2289 and #2188.

# Screenshots
#### Before
old-prompt-buffer

#### After
new-prompt-buffer

#### Before
dark-old-prompt-buffer

#### After
dark-new-2-prompt-buffer

# Discussion

Whether we need more colors besides primary and secondary. Conclusion: we don't.
Handling buttons and hovering. Solution: apply an opacity of 50% to all hoverable elements.

- [ ] Can someone check how the theme looks relative to `bookmark-mode`? 

# Checklist:

### DIFF
--- a/libraries/theme/package.lisp
+++ b/libraries/theme/package.lisp
@@ -14,12 +14,13 @@
    #:+light-theme+
    #:+dark-theme+
    #:background
-   #:text
+   #:on-background
    #:primary
+   #:on-primary
    #:secondary
-   #:tertiary
-   #:quaternary
+   #:on-secondary
    #:accent
+   #:on-accent
    #:font-family
    #:with-theme
    #:themed-css))
--- a/libraries/theme/tests/test.lisp
+++ b/libraries/theme/tests/test.lisp
@@ -5,55 +5,57 @@
 
 (prove:plan nil)
 
-(defvar *theme* (make-instance
-                 'theme:theme
-                 :dark-p t
-                 :background-color "black"
-                 :text-color "white"
-                 :accent-color "magenta"
-                 :primary-color "yellow"
-                 :secondary-color "blue"
-                 :tertiary-color "green"
-                 :quaternary-color "red")
-  "An absolutely crazy theme to test things on.")
+(defvar *theme* (make-instance 'theme:theme
+                               :dark-p t
+                               :background-color "black"
+                               :on-background-color "white"
+                               :primary-color "yellow"
+                               :on-primary-color "black"
+                               :secondary-color "blue"
+                               :on-secondary-color "black"
+                               :accent-color "magenta"
+                               :on-accent-color "black")
+  "Dummy theme for testing.")
 
 (prove:subtest "Basic CSS substitution"
   (prove:is (theme:themed-css *theme*
               (a
-               :color theme:primary
-               :background-color theme:background))
-            "a { color: yellow; background-color: black; }
+               :background-color theme:background
+               :color theme:primary))
+            "a { background-color: black; color: yellow; }
 "))
 
 (prove:subtest "Multi-rule/multi-color substitution"
   (prove:is (theme:themed-css *theme*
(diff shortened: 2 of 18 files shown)


aartaka (commented):
It looks more or less bearable, I'd say. Which screens does it not work on?

The only set of things that I've noticed were:
- Button hovering is quite broken -- it should switch between the `on-primary` and some other contrasting color (seems like our idea that we only need pairs of colors is not complete -- we need triplets or a separate set of element states -- like lighter button color (`primary` -> `secondary`) when hovered, for example). 
- Why use white-on-black as the source header in prompt-buffer? You can use either of `primary` or `secondary` here, I suppose -- they are delineated by indentation in addition to color, after all :)

aartaka:
REPL looks quite unintelligible too :D

aartaka:
More findings:
- Status buffer lacks colors, how about we use black for buttons and modes, while using `primary` and `secondary` for the rest?
- `secondary` has an ill contrast with white, so using it for buttons makes them indiscernible.

aadcg:
About the triplets, I'm not quite sure yet. When we'll get to where I envision, I hope to convince you :p

I follow your concerns. Can we address this in another way? Instead of thinking about globally, let's firstly focus in one subset of the UI - the status buffer. 

When I did the changes, I did mechanically (query-replace) and blindly, since anyway it would require analyzing each and every case.

I need you help with CSS. Could you please look at commit 7e7483e52 and take a look at my comment on line 696?

aadcg:
> REPL looks quite unintelligible too :D

Solved.

> Why use white-on-black as the source header in prompt-buffer? You can use either of primary or secondary here, I suppose -- they are delineated by indentation in addition to color, after all :)

Should be good now.

The status buffer was also massaged.

Hovering is still a big issue. I'm ruminating on the subject.

aadcg:
Added screenshots to the first post in this PR.

aadcg:
@aartaka Do you agree with 0c258a220?

aadcg:
Please re-read my first post. Screenshots included.

I didn't squash the commits so that the changes can be tracked with ease. Taking a look commit-wise will help to review.

@jmercouris, you have been too quiet on this topic :thinking:

jmercouris:
I am thinking. I like too much how our theme looks, but your argument is also irrefutable about minimum contrasts and usability. I am trying to think of how to solve this dilemma. The one approach I liked so far was that which was suggested by Artyom of keeping the default theme and adding high contrast variants for those who wish/require to use them.

jmercouris:
Also we should change the glyph for the source icon, seems too many people lack this in their system font!

aadcg:
@jmercouris is there a specific thing you don't like about the new theme? I'm not sure about prompt's buffer top line having an `accent` color background for instance. I think it's possible to make the theme better (more accessible) without compromising on the aesthetics of the old way. 

>  Also we should change the glyph for the source icon, seems too many people lack this in their system font!

Any suggestion?

aadcg:
> maybe we should use LASS for styling? This way, we'd need no quirky macros for CSS theming...

It looks interesting, but do we want to do it now? Which macro are you referring to?

jmercouris:
I believe Artyom is referring to the macro that generates CSS. Lass may be good, assuming it has an escape hatch! I have not investigated it thoroughly!

aadcg:
Ready to merge from my side.

aadcg:
@jmercouris, you can push commits to this branch!

aartaka (changes requested):
As a follow-up to our last week discussion: get the bright yellow accent back to the dark theme, and we're good to go :P

aadcg:
I'm merging this for now. I have concerns related to the dark theme (too long to describe). I believe @aartaka uses it a lot, so I expect some feedback in the future.

Ambrevar:
Thanks a lot!

PR #2333, opened 2022-06-02
Fix diff mode bug

### LINKED ISSUE #2328: `diff` is broken
Investigating.

aadcg:
# Description

No luck still... Something beyond my understanding is going on here. I need help! Please read commit 31956cf52 for more details.

I also need to implement this [feature](https://github.com/atlas-engineer/nyxt/pull/2058#issuecomment-1145013470).

CC @Ambrevar @aartaka 

(Will) fix #2328.

# Checklist:


source/mode/diff.lisp
```diff
@@ -29,21 +29,28 @@ the highest standard on accessibility."))
 
 (export-always 'diff)
 (define-internal-page-command-global diff
-    (&key (old-buffer-id (id (prompt1 :prompt "Old buffer"
-                                      :sources (make-instance 'buffer-source
-                                                              :return-actions  nil))))
-          (new-buffer-id (id (prompt1 :prompt "New buffer"
-                                      :sources (make-instance 'buffer-source
-                                                              :return-actions  nil)))))
+    (&key (old-buffer (prompt1
```
aartaka, line 32: Note that you cannot pass any non-literal object into `nyxt:` URL pages -- you should only pass something that you can serialize to the URL. Symbols, for one. If you get the `id` of the chosen buffers, it should work.

aadcg: Clear.

aadcg:
@aartaka, indeed I was doing it wrong, but that was because I was trying to solve the real problem that I identified before.

Let me describe the real problem.

```lisp
(define-internal-page-command-global test
    (&key (buffer-id (id (prompt1
                           :prompt "A buffer"
                           :sources (make-instance
                                     'buffer-source
                                     :multi-selection-p nil
                                     :return-actions nil)))))
    (test-buffer "*test*" 'base-mode)
  "Test"
  (spinneret:with-html-string
    (:raw (ffi-buffer-get-document (nyxt::buffers-get buffer-id)))))
```

This doesn't work and the problem lies at the guts of internal pages (namely the macro `with-current-buffer`). I'm afraid I'm still not able to fix this...

aadcg:
Found it. The culprit is commit `e176a24636a9377a3af9b931740efb4a95a8b5a8`.

aadcg:
@Ambrevar indeed we now have part of the problem solved.

Unfortunately, as I feared, there's still a nasty memory error when running the code bode on any web buffer (NOT internal pages):

```lisp
(define-internal-page-command-global testing
    (&key (buffer-id (id (prompt1
                           :prompt "A buffer"
                           :sources (make-instance
                                     'buffer-source
                                     :multi-selection-p nil
                                     :return-actions nil)))))
    (test-buffer "*test*" 'base-mode)
  "Test"
  (spinneret:with-html-string
    (:raw (ffi-buffer-get-document (nyxt::buffers-get buffer-id)))))
```

Basically that's why `diff` doesn't still work.

aadcg:
@Ambrevar I believe #2354 wasn't enough to address the issue I mention in the comment above.

aadcg:
@Ambrevar no need for your intervention for now :) 

I know how to fix this. 

The issue related to internal pages has been fixed by you on #2398. Thanks!

aartaka:
@aadcg, mind making this ready-for-review so that I can push changes too?

aadcg:
Done @aartaka.

aartaka:
@aadcg, I've fixed relative URL resolving in 9bbce3b52ce0413873dd6af40c007a3de4f299b8, but I want to add some more API niceties, so don't rush merging :)

aartaka:
Ready for review now. Manual and changelog updated, diff-mode works for both open buffers and new URLs, spawning the background buffer to fetch the sources if necessary. I like the API and the look of `diff:` URLs. However, an additional pair of eyeballs won't hurt :)

@aadcg, how does this variation on the topic of `diff-mode` look to you?

source/manual.lisp
```diff
@@ -216,6 +216,13 @@ all hosts being blocked, execute command " (:code "describe-variable") ", choose
 (:code "nyxt/blocker-mode:url-body") " slot." " To customize host blocking, read the "
 (:nxref :class-name 'nyxt/blocker-mode:blocker-mode) " documentation."))
 
+      (:nsection :title "Diff mode"
+        (:p "Diff mode computes and displays the content differences between two URLs.")
+        (:p "First, make sure that those two buffers are open. Then run "
```
Ambrevar, line 221: "open 2 separate buffers with the URLs you want to compare"?

source/mode/diff.lisp
```diff
@@ -22,28 +22,86 @@
...
-    (spinneret:with-html-string
-      (:style (style (find-submode 'nyxt/diff-mode:diff-mode diff-buffer)))
-      (:raw
-       (html-diff:html-diff
-        old-html new-html
-        :insert-class "nyxt-diff-insert"
-        :delete-class "nyxt-diff-delete"
-        :replace-class "nyxt-diff-replace")))))
+(define-class open-urls-source (prompter:source)
+  ((prompter:name "URLs currently open")
```
Ambrevar, line 33: "Currently open URLs"

source/mode/diff.lisp
```diff
@@ -22,28 +22,86 @@
...
+           (buffer-loaded-hook buffer)
+           (make-instance 'hooks:handler
+                          :fn (lambda (buffer)
+                                (sleep 1)
+                                (calispel:! channel (ffi-buffer-get-document buffer))
+                                (hooks:remove-hook (buffer-loaded-hook buffer)
+                                                   'buffer-source-fetching)
+                                (ffi-buffer-delete buffer))
+                          :name 'buffer-source-fetching))))
+      (calispel:? channel))))
```
Ambrevar, line 79: Neat!  We will probably need this in other places in the future, it's a good building block for a scraper library.

source/mode/diff.lisp
```diff
@@ -22,28 +22,86 @@
...
+                       :test #'quri:uri= :key #'url)))
+    (ffi-buffer-get-document existing-buffer)
+    (let ((channel (nyxt::make-channel 1)))
+      (run-thread "diff-mode URL fetching"
+        (let ((buffer (make-background-buffer :url url)))
+          (hooks:add-hook
+           (buffer-loaded-hook buffer)
+           (make-instance 'hooks:handler
+                          :fn (lambda (buffer)
+                                (sleep 1)
```
Ambrevar, line 73: Do we need to sleep here?  1s seems to be too much.

source/mode/diff.lisp
```diff
@@ -22,28 +22,86 @@
...
+                                (calispel:! channel (ffi-buffer-get-document buffer))
+                                (hooks:remove-hook (buffer-loaded-hook buffer)
+                                                   'buffer-source-fetching)
```
Ambrevar, line 76: Why not use `once-on`?

source/mode/diff.lisp
```diff
@@ -22,28 +22,86 @@
...
+  (when (and old-url new-url)
+    (make-buffer-focus
+     :url (format nil "diff:~a/~a" (quri:url-encode (quri:render-uri old-url))
+                  (quri:url-encode (quri:render-uri new-url))))))
+
+(defun fetch-url-source (url)
+  (alex:if-let ((existing-buffer
+                 (find url (remove-if (lambda (buffer)
+                                        (or (url-empty-p (url buffer))
+                                            (not (quri:uri-p (url buffer)))))
```
Ambrevar, line 62: Replace with `valid-url-p`?

source/mode/diff.lisp
```diff
@@ -22,28 +22,86 @@
...
+                                               :sources diffable-url-sources))
+                             (new-url (prompt1 :prompt "New URL"
+                                               :sources diffable-url-sources)))
+  "Show the difference between the OLD-URL and NEW-URL.
+
+DIFFABLE-URL-SOURCES allow you to configure the sources URLs are selected from."
+  (when (and old-url new-url)
+    (make-buffer-focus
+     :url (format nil "diff:~a/~a" (quri:url-encode (quri:render-uri old-url))
+                  (quri:url-encode (quri:render-uri new-url))))))
```
Ambrevar, line 56: Encoding the old / new URLs as path elements work, but seems awkward.

Why not encoding them as `?old=OLD&new=NEW` query parameters?

source/mode/diff.lisp
aartaka: Indeed, that might be unnecesary...

source/mode/diff.lisp
aartaka: I'm still uncertain of how `once-on` should be used, after all. Should it be a user-only helper? Should it be propagated to the source? Should it be a convention to supercede `hooks:add-hook`? I'm not sure.

source/mode/diff.lisp
aartaka: Yes, that could work.

aartaka:
All addressed except for `once-on` discussion, as I am still uncertain about its status :)

source/mode/diff.lisp
Ambrevar: Why user-only?  Is there something you deem not "core friendly" enough about `once-on`?

aartaka: I mean, it's the same thing as `current-mode` -- it _may imply too much_ to be generally useful. But then, `once-on` is much less context-dependent than `current-mode`, so it should be safe to use, I guess. I'll rewrite this.

aartaka:
I'm too used to the fact that we tend to use more wordy yet explicit code idioms that the convenient and short ones. The fact that there are both short and explicit idioms (because of macros' power) was not obvious to me xD

aadcg:
@aartaka how did you test this? 

If I request a diff between `https://github.com/atlas-engineer/nyxt/` and `https://github.com/atlas-engineer/nyxt/tree/1.5.0`, it does wild things.

aadcg:
The `open-urls-source` is evil since if you're at URL `prefix` and want to go to `prefix-foo`, it won't let you.

aartaka:
> @aartaka how did you test this?
> 
> If I request a diff between `https://github.com/atlas-engineer/nyxt/` and `https://github.com/atlas-engineer/nyxt/tree/1.5.0`, it does wild things.

I haven't used it before it broke, so I was unsure about how good it should be :sweat_smile:

aartaka:
OK, it seems that `html-diff` produces some HTML that Plump cannot parse and serialize correctly. Several solutions I see there:
- Use one of the diffed URLs as the host for `diff:` URLs so that WebKit resolves the relative URLs against it.
- Use [WebKitURISchemeResponse](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitURISchemeResponse.html) and redirect sub-resources to the real ones, saving us time and loads.
- Somehow parse the results of `html-diff` and insert `style` and `base` tags there.
- Rewrite `html-diff` using Plump (should save lots of lines and be more compatible).

Ambrevar:
> * Use one of the diffed URLs as the host for `diff:` URLs so that WebKit resolves the relative URLs against it.

How is it related to the problem that Plump cannot parse html-diff results?

aartaka:
> > * Use one of the diffed URLs as the host for `diff:` URLs so that WebKit resolves the relative URLs against it.
> 
> How is it related to the problem that Plump cannot parse html-diff results?

It avoids parsing (or doing anything at all) with html-diff results and thus kind of solves the problem. Although, obviously, a more straightforward solution would be to either
- Make Plump a bit more tolerant to the malformed HTML html-diff produces, or
- Make html-diff produce a bit more structured HTML.

aadcg:
It seems that `diff-mode` will be left out of Nyxt for 3.0 so this should be closed.

Ambrevar:
Wait, isn't a first attempt at fixing the issue?  Even if we don't merge it for 3.0, we can keep this open until it's properly fixed, right?

aadcg:
> Wait, isn't a first attempt at fixing the issue? 

Sorry, could you rephrase?

> Even if we don't merge it for 3.0, we can keep this open until it's properly fixed, right?

You're suggesting that it could be re-integrated after 3.0 once we fix it, so it could stay open right? I agree then.

Ambrevar:
Yes!

aartaka:
@aadcg, can you test? I've tried fixing with `<base>` tag addition and it seems to more or less work.

aadcg:
How did you test it?  

It doesn't produce satisfactory results as those reported in the [old article](https://nyxt.atlas.engineer/article/diff-mode.org).

aartaka:
At this point, I'm no longer sure we can retain the functionality that `diff-mode` used to have, except maybe for simple HTML-only page...

aadcg:
That was my point all along :)

The mode is trivial. It stopped working when we deprecated internal buffers where you could say "here's the HTML, render it". Those were deprecated security reasons obviously. My only hope was that by fine-tuning the parameters WebKitGTK provides for custom schemes perhaps we could  bring it back to life.

I all in favor of forgetting about it. It was devised in completely different circumstances.  I think the concept is worth exploring in the future, it is just that we need another completely different strategy. This is post 3.0.

aadcg:
I have deleted everything dependent on `diff-mode`. 

@Ambrevar, notice that your `remembrance-mode` is marginally affected by this change. I don't know how significant this is to its functionality.

Ok to merge?

Ambrevar:
Please don't remove html-diff and the corresponding feature in remembrance mode, it's a nice feature and it does not cost anything.

source/changelog.lisp
```diff
@@ -568,7 +568,8 @@ locally and is cleared after every session.")
 Nyxt-internal-pages.")
     (:li "The rules file is now moved to auto-rules.lisp (instead of the old
 auto-mode-rules.lisp)."))
-   (:li "Startup is more robust against corrupted history files."))
+   (:li "Startup is more robust against corrupted history files.")
+   (:li (:code "diff-mode") " is deprecated."))
```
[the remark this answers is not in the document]
aadcg: Sure, agree!

source/mode/remembrance.lisp
```diff
@@ -361,38 +361,6 @@ See also `auto-cache-on-load-p' in `remembrance-mode'."
...
-                             :sources (make-instance 'buffer-source
-                                                     :return-actions #'identity))))
-        (mode (find-submode 'remembrance-mode)))
-    (alex:when-let ((doc (find-url (url buffer) mode)))
-      ;; TODO: Display in internal page.
-      (buffer-load-internal-page-focus 'view-diff
-                                       :id (id buffer)
-                                       :diff (html-diff:html-diff (page-html-content doc)
-                                                                  (buffer-html-content buffer))))))
-
```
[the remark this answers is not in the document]
aadcg: Oh... I was very inattentive, sorry. This doesn't depend on `diff-mode`, of course!

aadcg:
Thanks everyone.

PR #2346, opened 2022-06-07
Dynamic print status

Ambrevar:
# Description

In https://github.com/atlas-engineer/nyxt/issues/2219#issuecomment-1141024604 I introduced a draft of how we could avoid hard-coding the calls to `print-status`.

This is useful since it allows the user to customize what triggers a status buffer re-print.

# Discussion

1. [x] Each time a `status-buffer' is created the defmethods are redefined, which triggers warnings.  Can we somehow report warning only when a  _non-watcher_ method exists?  Yes, see a2109f9de5c8852982db53a57301096120996d86.
2. [ ] I've re-introduced the `mode` `enabled-p` accessor, but the writer should not be called externally.  But we can add a setf "watcher" only on slots with writers.  Unless... Is it possible to add an `:after` method to `(setf (slot-value ...))`?

# Checklist:


Ambrevar:
I think a2109f9de5c8852982db53a57301096120996d86 fixes the redefinition warning.

Ambrevar:
Ready to merge on my end.

Ambrevar:
Another way to add slot "watchers" could be to use the Cells library: https://github.com/kennytilton/cells

But that may be too much for what we want there.

aadcg:
I agree that going forward with the Cells library isn't a good idea today.

Ambrevar:
Merged with 324ad2d904b1a834433ee82c56c30c3964cb811d.  I don't know what to do with the `enabled-p` slot, I guess we will see when the time comes.

PR #2348, opened 2022-06-08
Full startup recovery

### LINKED ISSUE #2286: Recover from post-init startup errors
Currently, Nyxt won't start if the init file compiles properly _but_ contains some broken logic.

Example file:

```lisp
(defmethod customize-instance ((buffer buffer) &key)
  (setf (auto-mode-rules-file buffer) (make-instance 'auto-mode-rules-file
                                                     :base-path "/path/to/auto/rules")))
```

This is a big one, because it confuses users tremendously + it makes it hard to work on init files.

A possible solution would be to add some guard around browser, window and buffer instantiation.  On failure, either:

- Call `clean-configuration` and try instantiating the class again.
- Or restart the browser without init file.

Ambrevar:
# Description

Previously on a config file like

```lisp
(defmethod customize-instance ((buffer buffer) &key)
  (setf (auto-mode-rules-file buffer) (make-instance 'auto-mode-rules-file
                                                     :base-path "/path/to/auto/rules")))
```

Nyxt would hand on startup, not displaying any window.
This was due to us not catching errors in `define-ffi-method`!  Hard to spot but easy to fix!

Fixes #2286.

# Discussion

- [x] Currently the recovery error reporting is not very meaningful for the above snippet.
Can we improve the message? Done with 1d75571dc54b5863540fe2142e79b27f90692e35.

- [ ] Would be nice if we could add a startup recovery test.

# Checklist:


Ambrevar:
In 1d75571dc54b5863540fe2142e79b27f90692e35 I've introduced `*debug-on-startup*` to report the right backtrace in `define-ffi-method`.

But between this and `*run-from-repl-p*` and `*debug-on-error*`, this is starting to get a bit confusing.

Maybe we could set `*debug-on-error*` during startup instead, and in `define-ffi-method` use `handler-case` only if `*run-from-repl-p*` is set?
But then we would lose the meaningful stacktrace when setting `*debug-on-error*` for FFI methods.

Other consideration: The current approach reports the backtrace of the non-GTK thread, but not the one of the GTK thread.  So if an error occurs on the GTK thread, we only report the backtrace of the caller (which is waiting on the error-channel).  It's decent, but the backtrace could be more accurate if it were captured from the GTK thread.

Suggestion 2: Adapt `with-protect` to use another control variable (we can remove `*debug-on-startup*`) so that it takes  care itself of capturing the backtrace and restarting the browser.

Thoughts?

Ambrevar:
Simplified in 915b09074b2728fcc189d1909eda819854929440.

Ambrevar:
I've used this idiom to pass the condition from the GTK thread to the calling thread:

```lisp
(let ((current-condition nil))
  (restart-case
      (handler-bind ((condition (lambda (c) (setf current-condition c))))
        (calispel:! channel (progn ,@forms)))
    (abort-ffi-method ()
      :report "Pass condition to calling thread."
      (calispel:! error-channel current-condition))))
```

But do we really need to set a let-bound variable to pass the condition around?  Isn't there a way in the standard to do that?

Cc @aartaka @jmercouris

jmercouris:
> I've used this idiom to pass the condition from the GTK thread to the calling thread:
> 
> ```lisp
> (let ((current-condition nil))
>   (restart-case
>       (handler-bind ((condition (lambda (c) (setf current-condition c))))
>         (calispel:! channel (progn ,@forms)))
>     (abort-ffi-method ()
>       :report "Pass condition to calling thread."
>       (calispel:! error-channel current-condition))))
> ```
> 
> But do we really need to set a let-bound variable to pass the condition around? Isn't there a way in the standard to do that?
> 
> Cc @aartaka @jmercouris

No, you don't need to. You could make a new condition that wraps the old condition that is caught where you want to take it in another handler-bind.

For example:

I have a condition tomato. I have a piece of code that catches condition potato.

As soon as I get a condition tomato, I make a new potato condition, and give it my tomato. Now the flow has moved elsewhere without needing to let bind the condition.

Does that work for your needs?

Ambrevar:
It doesn't, because we have no control over _which type_ of condition is thrown.

What I'm asking should work for any condition.

aartaka:
Wow, sounds cool! Thanks for that, Pierre!

PR #2349, opened 2022-06-08
Fix spell check

hendursaga:
The mode is still not where I'd like it to be, but at least it works again as it did before!

### DIFF
--- a/source/mode/spell-check.lisp
+++ b/source/mode/spell-check.lisp
@@ -12,7 +12,7 @@
    (spell-check-language
     "en_US"
     :documentation "Spell check language used by Nyxt. For
-a list of more languages available, see `enchant:broker-list-dicts'.")
+a list of more languages available, see `spell-check-list-languages'.")
    (keymap-scheme
     (define-scheme "visual"
       scheme:cua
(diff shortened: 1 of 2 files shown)


source/mode/spell-check.lisp
```diff
@@ -24,51 +24,55 @@ a list of more languages available, see `enchant:broker-list-dicts'.")
       (list
        "z =" 'spell-check-word)))))
 
+(defmacro with-spell-check ((variable) &body body)
+  `(enchant:with-dict (,variable (spell-check-language
+                                  (find-submode 'spell-check-mode)))
+     ,@body))
+
+(defun spell-dict-check (word)
```
Ambrevar, line 32: It's a predicate then.  Suffix it with `-p` to make this clear.

Optional: Declaim the function type, e.g. `(-> spell-dict-check (string) boolean)`

Ambrevar:
This is excellent, thank you so much!

Could you address the nit I've reported, then I'll merge!

hendursaga:
I've decided to not declaim, under SBCL I ran into...

```
; (VALUES &OPTIONAL BOOLEAN &REST T).
; (VALUES &OPTIONAL BOOLEAN).
```

... and after a few attempts to fix that, I decided it wasn't worth it. I learned more about the type system though! :)

You can merge.

Ambrevar:
Thanks a lot!

PR #2354, opened 2022-06-10
Lisp nyxt urls refactor2

### LINKED ISSUE #2268: Why do some function take a buffer-id instead of buffer? | Allow non-literal data in nyxt: URLs

Ambrevar:
This is a follow up to #2306.

This turns internal pages into commands and removes even more macrology!
Bonus: since less code is quoted, it catches even more mistakes at compile time!

Left to do:

- [x] Test it.
- [x] Replace `lisp-eval` with `lisp-eval2`.
- [ ] Fix all `define-internal-page-command` that have awkward arguments.
  -  `buffer-history-tree`: need to fix ID type first.
  - `summarize-buffer`: same.
  - `diff`: Need to fix #2333.
- [ ] Answer https://github.com/atlas-engineer/nyxt/pull/2306#discussion_r884151097.
- [x] Turn describe-value into a command-less internal page.

Fixes #2268.


Ambrevar:
Mostly ready to merge, but we need to decide what to do with https://github.com/atlas-engineer/nyxt/issues/2361 first.

Ambrevar:
Merged with 5746eea3e101317779aa558b4aa361328d535700.

PR #2372, opened 2022-06-15
documents/README: Add "Branch management" secion.

Ambrevar:
We are rethinking our branching strategy.  Here is the new outline.

Any one against it?

### DIFF
--- a/documents/README.org
+++ b/documents/README.org
@@ -496,9 +496,9 @@ See https://nyxt.atlas.engineer/learn-lisp for a few recommendations.
 
 ** Contributing
 
-Nyxt is a joint effort and we need you to make it succeed!  You can find various
-small projects [[https://github.com/atlas-engineer/nyxt/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-issue][on our issue tracker]] to suit your interests and skills.  Feel
-free to contact us at any point if you need guidance.
+Nyxt is a joint effort and we need you to make it succeed!  You can find ideas
+[[https://github.com/atlas-engineer/nyxt/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-issue][on our issue tracker]] to suit your interests and skills.  Feel free to contact us
+at any point if you need guidance.
 
 When ready to start working please fork the repository, add your changes and
 open a pull request on GitHub.  Refer to the [[*Branch management][branch management section]] for more
--- a/documents/README.org
+++ b/documents/README.org
@@ -502,8 +502,8 @@ Nyxt is a joint effort and we need you to make it succeed!  You can find ideas
 at any point if you need guidance.
 
 When ready to start working please fork the repository, add your changes and
-open a pull request on GitHub.  Refer to the [[*Branch management][branch management section]] for more
-detailed information.
+open a pull request on GitHub to pass the review process.  Refer to the [[*Branch management][branch
+management section]] for more detailed information.
 
 You can contribute to Nyxt without commit access.  However, if you're a frequent
 contributor, you may request it.  Remember that with great power comes great
@@ -532,6 +532,10 @@ so please take that into account!
 
 After the changed are merged, please do not forget to delete obsolete or
 dangling branches.
+
+Note to core contributors: since you have commit access, you can push trivial
+changes directly to the target branch (skipping the review process).  The merge
+commit is required when at least one commit isn't atomic.
 
 ** IRC Channel
 
--- a/documents/README.org
+++ b/documents/README.org
@@ -494,15 +494,17 @@ See https://nyxt.atlas.engineer/learn-lisp for a few recommendations.
 
 ** Contributing
 
-To contribute, please find a task [[https://github.com/atlas-engineer/nyxt/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-issue][on our issue tracker]] or within [[file:CHANGELOG.org][CHANGELOG.org]]
-document that has a TASK label affixed.  Upon finding a task that you'd like to
-work on, ideally, ensure that it is not already being worked on.
+Nyxt is a joint effort and we need you to make it succeed!  You can find various
+small projects [[https://github.com/atlas-engineer/nyxt/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-issue][on our issue tracker]] to suit your interests and skills.  Feel
+free to contact us at any point if you need guidance.
 
-After you have found a TASK item that is available:
+When ready to start working please fork the repository, add your changes and
+open a pull request on GitHub.  Refer to the [[*Branch management][branch management section]] for more
+detailed information.
 
-- make a fork of the repository,
-- add your changes,
-- make a pull request.
+You can contribute to Nyxt without commit access.  However, if you're a frequent
+contributor, you may request it.  Remember that with great power comes great
+responsibility.
 
 ** Branch management
 
--- a/documents/README.org
+++ b/documents/README.org
@@ -202,7 +202,7 @@ Notes:
    `webkit2-gtk(-devel)` instead of `webkit-gtk3` via your favorite package
    manager. If you are on a legacy version of macOS which is not supported by
    HomeBrew, you may want to try install these via MacPorts.
-  
+
 ** Qt dependencies
 
 Qt is only required if you want to build Nyxt over the WebEngine renderer.
@@ -272,7 +272,7 @@ application bundle which you can copy into your =/Applications= folder.
   #+begin_src sh
   gmake all
   #+end_src
-  
+
 - macOS:
   #+begin_src sh
   make all
@@ -403,7 +403,7 @@ be configured in the respective configuration files.
 #+begin_src lisp
 (define-command-global start-slynk (&optional (slynk-port *swank-port*))
   "Start a Slynk server.
-   
+
 Waits for incoming connections, e.g. from SLY.
 
     Warning: This allows Nyxt to be controlled remotely, that is, to execute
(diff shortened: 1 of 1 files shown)


aartaka:
We need to change the "Rebase and merge" option for GitHub PR approval to "Create a merge commit" to enforce this behavior for such unfocused cabbages as myself...

CC @jmercouris, mind changing it?

Ambrevar:
Actually, it depends on what you want, both "rebase and merge" and "merge commit" work: if the PR is trivial and only with clean commits, "rebase and merge" is what you want.
Otherwise the latter.

Now John had some concerns with regard to mistakes.  What if we press the button while we shouldn't have?  Here we must consider the default carefully:

- If "rebase and merge" is the default, mistakes occur when the PR has unclean commits.  Result: same as the previous strategy, so no big deal.
- If "merge with commit" is the default, mistakes occur when the PR is trivial with only clean commits, _or_ when the PR is not rebased on `master`. 
  -  If the former, then we have a merge commit and it's completely fine.  
  - If the latter, then our history will have some commit in parallel between `master` and the branch of the PR.  Considering we are discourage branching off from anything but `master`, it remains quite readable.  So no big deal either here.

aartaka:
So "Create a merge commit" as the sane/safe default then? :)

aartaka:
We're slowly solidifying our contributing guidelines. A high time we add a CONTRIBUTING file as GitHub suggests :)

aadcg:
>  So "Create a merge commit" as the sane/safe default then? :)

After the detailed analysis @Ambrevar wrote, I seem to agree with you @aartaka.

aadcg:
@Ambrevar I boldly decided to rewrite the section you added and reviewed some other things. What do you think?

documents/README.org
```diff
@@ -1,5 +1,7 @@
 #+TITLE: Nyxt Developer's Manual
 
+# Please install org-make-toc so the the TOC below will be automatically
+# generated.
```
Ambrevar, line 4: Add link to https://github.com/alphapapa/org-make-toc.
New line after the comment.

documents/README.org
```diff
@@ -493,20 +496,42 @@ See https://nyxt.atlas.engineer/learn-lisp for a few recommendations.
 
 ** Contributing
 
-To contribute, please find a task [[https://github.com/atlas-engineer/nyxt/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-issue][on our issue tracker]] or within [[file:CHANGELOG.org][CHANGELOG.org]]
-document that has a TASK label affixed.  Upon finding a task that you'd like to
-work on, ideally, ensure that it is not already being worked on.
+Nyxt is a joint effort and we need you to make it succeed!  You can find various
+small projects [[https://github.com/atlas-engineer/nyxt/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-issue][on our issue tracker]] to suit your interests and skills.  Feel
```
Ambrevar, line 500: "small projects" is confusing I think.

documents/README.org
```diff
@@ -493,20 +496,42 @@ See https://nyxt.atlas.engineer/learn-lisp for a few recommendations.
...
+When ready to start working please fork the repository, add your changes and
+open a pull request on GitHub.  Refer to the [[*Branch management][branch management section]] for more
+detailed information.
 
-- make a fork of the repository,
-- add your changes,
-- make a pull request.
+You can contribute to Nyxt without commit access.  However, if you're a frequent
+contributor, you may request it.  Remember that with great power comes great
+responsibility.
```
Ambrevar, line 509: Shall we?  I do not know, there is no real benefit in my understanding, unless the main team becomes overburdened to the point they can't merge pull requests any more.

More people with commit access = more potential (mis)management issues :p

documents/README.org
```diff
@@ -493,20 +496,42 @@ See https://nyxt.atlas.engineer/learn-lisp for a few recommendations.
...
+so please take that into account!
+
+After the changed are merged, please do not forget to delete obsolete or
+dangling branches.
 
 ** IRC Channel
 
 #+html: You can find Nyxt on Libera IRC: <a href="https://kiwiirc.com/nextclient/irc.libera.chat/nyxt">#nyxt</a>
-
-# Local Variables:
```
Ambrevar, line 510: Why did you delete this?

Ambrevar: It does not override the global definition, it's a local variable.

documents/README.org
```diff
@@ -510,31 +510,25 @@ responsibility.
 
 Nyxt uses the following branches:
 
-- =master= for development.
-- =2-series= to backport commits for Nyxt 2.*.
-- And so on for all following major versions.
+- =master= for development;
+- =<feature-branches>= for very particular situations;
```
Ambrevar, line 516: What do you have in mind?  I don't think we do this, do we?

documents/README.org
```diff
@@ -510,31 +510,25 @@ responsibility.
...
+- =<2,3,...>-series= to backport commits corresponding to specific major
+  versions.
 
-Contributors with commit access may push trivial commits on =master= directly.
-Non-trivial changes should go through a review process ("pull request") through
-their own branch.
+It's recommended to branch off from the target branch and to rebase onto it
+right before merging.  Both of these practices keep the history as clear as
+possible and avoid merge conflicts.
```
Ambrevar, line 522: "and avoid merge conflicts" is not relevant here, because the same conflicts would arise when rebasing.

documents/README.org
aadcg: Fair enough! I tried to avoid the word "task", since it sounds like a dull thing to do.

aadcg: Any suggestions?

documents/README.org
aadcg: I agree that we shouldn't!  But you wrote the following:

> Contributors with commit access may push trivial commits on =master= directly.

To my mind, this document is directed to developers that aren't core contributors. We, core developers, know our privileges already :)

My suggestion is to remove any reference to commit access.

documents/README.org
aadcg: Say that globally the value of `before-save-hook` is `(delete-trailing-whitespace)`. Create a file and add a local variable that bind `before-save-hook` to `(lambda nil (message "hi"))`. Close the buffer corresponding to that file and open it again. Notice that `delete-trailing-whitespace` isn't run before saving.

documents/README.org
aadcg: Yes we do! `clear-cache` for instance, that @aartaka was working on at some point.

documents/README.org
aadcg: You comparing rebasing vs merging + reabasing? You are correct. 

But our aspiring contributors don't need to know what we were managing our git history before right?

What I tried to say here is that rebasing onto the target branch right before the merge is a good rule of thumb to avoid some possible conflicts.

documents/README.org
Ambrevar: "ideas" / "sources of inspiration"?

documents/README.org
Ambrevar: I wrote this as a welcoming guide for core developers, that's why :)

documents/README.org
Ambrevar: But this is just a draft that never ended into a pull request, isn't it?  What is particular about this situation?

documents/README.org
aadcg: Resolved with 65073ddc2.

documents/README.org
aadcg: Resolved with 81407ac80.

documents/README.org
aadcg: Emacs should treat hooks specially in the context of local variables indeed. It's ok to override some variables on files (such as `fill-column`). But, in this case, it should be possible to add to the a function to the global `before-save-hook` otherwise we're tripping users. I'll ask around.

Regarding `delete-trailing-whitespace`, it's a tangent topic. In my setup I have the following for years:
```elisp
(setq require-final-newline t)
(add-hook 'before-save-hook 'delete-trailing-whitespace)
```

I believe all us (core contributors) do that as well.  Most, if not all, of the annoying trailing whitespaces were introduced by sporadic contributors. Where do we require trailing whitespaces in our manual?

aadcg: By annoying trailing whitespaces I mean those deleted by 197f76b41.

Ambrevar: This works. `eval` is less readable than a simple assignment, but who cares for this simple file :)

About `delete-trailing-whitespace`: I used to have the same settings as yours, but it tripped me too many times.  Now I use `ws-buttler`.

aadcg: That's a good piece of advice!

documents/README.org
aadcg: That's one way of looking at it.

Think of a branch like `tests`. Or possible branches such as `test-nyxt-on-ecl`, `text-nyxt-on-another-renderer`, etc. Aren't they somewhat different?

documents/README.org
aadcg: I see! Does c898fa371 sound accurate?

documents/README.org
aadcg: What I meant is that this guide is meant primarily for sporadic contributors, while also being helpful for code developers. The reason being that core developers get acquainted with the workflows through routine. It's the sporadic contributor who struggles the most.

Anyway, I agree that a line addressing core contributors doesn't hurt.

aadcg: 411aaf6e2abfa5fe875a037b1e122fa3673fe6e9 should fix this.

documents/README.org
Ambrevar: They are exceptional cases.  Of course if we wants to drift off from the rules for good reasons, we always can.  But "you can break the rules" is self-evident, no need to mention it I believe.

documents/README.org
Ambrevar: Perfect!

Ambrevar:
Merged with 85e83b70de857521820d161e3111e13083fe8faf.

aadcg:
Thank you for merging @Ambrevar. This is perhaps a case where squashing would be recommended.

Ambrevar:
Yup, wasn't too sure, but in doubt let's not do it.

Issue #2376, opened 2022-06-15
Allmost all bookmarks are removed when you remove just one


shamazmazum:
**Precise recipe to reproduce the issue**
Collect some bookmarks (like 10-20). Try to remove one bookmark in the `*Bookmarks*` buffer (shown by `list-bookmarks` command) by pressing a cross. After a while you will get almost all bookmarks removed. Maybe you need to remove two bookmarks for this to work.

Nyxt built from f4acb71 with bundled dependencies.

shamazmazum:
Pure code inside `files:with-file-content` in `delete-bookmark` works fine. So this must be effects of `files:with-file-content` which destroy bookmarks file.

shamazmazum:
Sorry, that was a bit older commit. Now I get something like

    <WARN> [21:22:20] Warning: Error when running NIL: The function COMMON-LISP:NIL is undefined.

similar to #2377. That must be fixed first.

shamazmazum:
Oops, I was wrong, the error is in pure code :) Fixed in #2377

Ambrevar:
To clarify, is this fixed in #2377?  I've just tested and it seems to work fine on my side when deleting a dozen bookmarks.

shamazmazum:
Oh, this issue can be closed now.

PR #2377, opened 2022-06-15
mode/bookmark.lisp: Make list-bookmarks interactive again

shamazmazum:
Make an entry in `*Bookmarks*` buffer disappear after you press delete button. I already did that before but that change was removed for some reason.

### DIFF
--- a/source/mode/bookmark.lisp
+++ b/source/mode/bookmark.lisp
@@ -389,12 +389,17 @@ rest in background buffers."
 
 (defmethod files:serialize ((profile nyxt-profile) (file bookmarks-file) stream &key)
   (let ((content
-          ;; Sort the entries to make serialization reproducible.
-          ;; Particularly useful when bookmarks are under version control.
-          (sort (files:content file)
-                #'url< :key #'url)))
+         ;; Sort the entries to make serialization reproducible.
+         ;; Particularly useful when bookmarks are under version control.
+         ;;
+         ;; Need non-destructive sort here or else the cached version
+         ;; of file content may become corrupted. For example, with
+         ;; destructive SORT almost all bookmarks are removed when the
+         ;; user tries to remove just few.
+         (sera:sort-new (files:content file)
+                        #'url< :key #'url)))
     (write-string "(" stream)
-    (dolist (entry content)
+    (sera:do-each (entry content)
       (write-string +newline+ stream)
       (serialize-object entry stream))
     (format stream "~%)~%")
--- a/source/mode/bookmark.lisp
+++ b/source/mode/bookmark.lisp
@@ -96,38 +96,6 @@ Bookmarks can be persisted to disk, see the `bookmarks-file' mode slot."
             (push bookmark (gethash tags bookmarks-table nil)))))
     bookmarks-table))
 
-(export-always 'list-bookmarks)
-(define-internal-page-command-global list-bookmarks ()
-    (bookmarks-buffer "*Bookmarks*")
-  "List all bookmarks in a new buffer."
-  (let ((bookmarks (group-bookmarks bookmarks-buffer)))
-    (spinneret:with-html-string
-      (:style (style (find-submode 'bookmark-mode (current-buffer)))) ; TODO: Make sure this is the right buffer
-      (:h1 "Bookmarks")
-      (cond
-        ((zerop (hash-table-count bookmarks))
-         (:p (format nil "No bookmarks in ~s." (files:expand (bookmarks-file bookmarks-buffer)))))
-        (t (maphash
-            (lambda (tag bookmarks)
-              (:details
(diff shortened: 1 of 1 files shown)


jmercouris:
Apologies the change was removed, probably it got lost during our many architectural restructures!

jmercouris:
In my testing it doesn't appear to work; the following appears in the REPL:

```
<WARN> [21:22:20] Warning: Error when running NIL: The function COMMON-LISP:NIL is undefined.
```

jmercouris:
This is however very likely an error on our part, as I've noticed it elsewhere as well.

shamazmazum:
@jmercouris yeah, I also cannot stop download with the same error ;)

shamazmazum:
> In my testing it doesn't appear to work; the following appears in the REPL:
> 
> ```
> <WARN> [21:22:20] Warning: Error when running NIL: The function COMMON-LISP:NIL is undefined.
> ```

Aha, it seems that lisp can only be called with `nyxt/ps:lisp-eval` now. The breaking commit is 30140ab.
@Ambrevar can you explain how can I pass `url` to lisp in this code (`source/modes/bookmark.lisp`)?

~~~~
(ps:ps (defun delbkm (url)
                 (let ((section (ps:chain document active-element
				          (closest ".bookmark-entry"))))
			(ps:chain section parent-node (remove-child section)))
                 (fetch (+ "lisp://" (escape (+ "(nyxt/bookmark-mode:delete-bookmark \"" url "\")/")))
			(ps:create :mode "no-cors"))))))))
~~~~

Ambrevar:
Oh yes, this `fetch` looks horribly wrong now! :p 

So here actually we can remove the whole delbkm trick since now lisp URLs are much shorter.  If you do not give it a title, it looks like this : `lisp://1234`, with `1234` the callback ID.

But this raises another problem that I didn't realize before: JS code can no longer pass JS data to the `lisp://` URLs.

@aartaka Any idea how to solve this?  I guess we would have to re-introduce argument support.

shamazmazum:
A quick fix would be partially reverting b7dd2f1 and adding one JS function for each bookmark. Will this do?

Ambrevar:
I think so.

shamazmazum:
Done. Can you also help me with #2376 after this PR is merged? That issue is not directly related to this PR but is about bookmarks too.

aartaka:
> But this raises another problem that I didn't realize before: JS code can no longer pass JS data to the `lisp://` URLs.
>
> @aartaka Any idea how to solve this? I guess we would have to re-introduce argument support.

Indeed, that's problematic. 

The good news is: the whole _raison d'être_ of this optimization is now gone with the fix to #2062, and we can attach as much wordy `nyxt/ps:lisp-eval`s to buttons as we wish to.

The bad news: we might still need to somehow pass arguments to `lisp://` URLs in the future (for reasons unconnected to code minification) and it's a non-trivial problem. We can use the mechanism from `nyxt://` URLs (namely, keyword arguments as query parameters), for one. But then, we don't always have control over the code in `lisp-eval` to the degree that we have control over the internal pages with `nyxt:://` URLs...

shamazmazum:
I added a fix for #2376. Can you merge?

shamazmazum:
On second thought I realized that this is unclear how the last commit fixes #2376 and how removes are different from `set-difference`. Removes keep the order, naturally, but it's unclear to me why #2376 happens with `set-difference`.

shamazmazum:
@Ambrevar OK, I did my best to explain the fix :)

Ambrevar:
The CI is failing:

```
; compiling file "NYXT:SOURCE;RENDERER;GTK.LISP.NEWEST" (written 20 JUN 2022 10:55:46 AM):
Heap exhausted during garbage collection: 160 bytes available, 176 requested.
```

Da hell?!?  Any idea anyone?

shamazmazum:
I wonder if this has something to do with the patch :) I can compile with only 2 Gb heap. How many memory roswell gives to sbcl?

Ambrevar:
Could be, but then https://github.com/atlas-engineer/nyxt/actions/runs/2529165821 (which is the exact same as this branch) passed!

shamazmazum:
Played I bit more. I think you have to enhance size of the heap or you may encounter similar problems in the future. Default of 1 Gb is too damn small.

Ambrevar:
Yup, it happend on https://github.com/atlas-engineer/nyxt/pull/2393 as well.  OK, I'll increase the size then.

(To clarify, the problem is only on the CI.  When built locally, we default to 3 GiB.)

Ambrevar:
Merged with 96d0b1ff3cd7d1f105acde9b7da0b204b47b4fcf.  Thanks a lot for your contribution (and your patience!)

shamazmazum:
Thanks!

PR #2391, opened 2022-06-20
Fill user-scripts in a more structured way

### LINKED ISSUE #2388: `setf`-ed `nfiles:content` does not trigger `nfiles:deserialize`
**Describe the bug**

Opening a bug report nere, as it's not only an `nfiles`, but also a recently merged user scripts one. The essence of the problem is `nfiles`-restricted, though.

`nfiles:deserialize` is used in `source/mode/user-script.lisp` to process the user scripts into the runnable form. However, if we use `(setf nfiles:content)` on a user script, `nfiles:deserialize` never happens, thus skipping the script normalization and making it break due to `@require`s (if any) being unresolved, or other reasons.

One possible example of that is https://github.com/aartaka/nx-dark-reader/commit/fd3db311234b36c46b85a450e78e5fc0377c4b8d, where I set the contents of a `nyxt/user-script-mode:user-script` to a dynamically generated contents and then push it to the `user-script-mode` list of scripts. This should make the script immediately available, but it doesn't. The reason — `nfiles:deserialize` never ran :(

Try running `nx-dark-reader` with stickers in gtk.lisp (`ffi-buffer-add-user-script`, namely) and in mode/user-script.lisp (`nfiles:deserialize` specialization, will never fire). You'll see that, eve though the script is being created just fine, it's not being parsed properly and thus loses the critical dependency — the Dark Reader library.

**Information**
- OS name+version: GuixSD
```sh
$ guix describe
Generation 22	May 11 2022 14:49:52	(current)
  guix dc00098
    repository URL: https://git.savannah.gnu.org/git/guix.git
    branch: master
    commit: dc0009853550fc5e7ad647d7dc4b962d343d62df
  guix-gaming-games 9924ad0
    repository URL: https://gitlab.com/guix-gaming-channels/games
    branch: master
    commit: 9924ad0a66f98ea1b538761fd49521acf4f689a3
  aartaka 00fc01c
    repository URL: https://github.com/aartaka/guix-channel
    branch: master
    commit: 00fc01cb35db2f878d7635f635e809a8b9b24633
  chur 8fcb5dd
    repository URL: https://github.com/aartaka/chur-guix
    branch: master
    commit: 8fcb5ddaa15fd10d71c1e50950405bb6d48e4393
  nonguix 1de0c32
    repository URL: https://gitlab.com/nonguix/nonguix
    branch: master
    commit: 1de0c32142c54bc73af5556d5e45c77152b31f0f
```
- Graphics card and driver: Intel UHD 620, `i915`
``` sh
$ lspci -v
...
00:02.0 VGA compatible controller: Intel Corporation UHD Graphics (rev 01) (prog-if 00 [VGA controller])
	Subsystem: Lenovo Device 5089
	Flags: bus master, fast devsel, latency 0, IRQ 165
	Memory at 601c000000 (64-bit, non-prefetchable) [size=16M]
	Memory at 4000000000 (64-bit, prefetchable) [size=256M]
	I/O ports at 3000 [size=64]
	Expansion ROM at 000c0000 [virtual] [disabled] [size=128K]
	Capabilities: <access denied>
	Kernel driver in use: i915
	Kernel modules: i915
...
```
- Desktop environment / Window manager name+version: StumpWM 20.11
- How you installed Nyxt (Guix pack, package manager, build from source): `guix package -f nyxt.scm`
```
Nyxt version: 3-1342-gaa991fabc
Renderer version: GTK
Operating system kernel: Linux 5.17.2
Lisp implementation: SBCL 2.2.2 (Dynamic space size: 3145728000)
Features: (:WEBKIT2 :WEBKIT2-2.36 :WEBKIT2-CORS-ALLOWLIST :WEBKIT2-PASTE-PLAINTEXT
 :WEBKIT2-TRACKING :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA
 :WEBKIT2-SANDBOXING :GTK-3-22 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14
 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20
 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4
 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF :GLIB-2-30 :GLIB-2-32 :GLIB-2-34
 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42 :GLIB-2-44 :GLIB-2-46 :GLIB-2-48
 :GLIB-2-50 :GLIB-2-52 :GLIB-2-54 :GLIB-2-56 :GLIB-2-58 :GLIB :NYXT-UNSTABLE
 :NYXT-GAA991FABC :NYXT-3 :FSET-EXT-STRINGS :CUSTOM-HASH-TABLE-NATIVE :SWANK
 :PLUMP-UTF-32 :PARENSCRIPT :GLOBAL-VARS :DECLARE-TYPES :NAMED-READTABLES
 :LPARALLEL :21BIT-CHARS :CLOSER-MOP :CL-PPCRE-UNICODE :CL-UNICODE :CL-PPCRE
 :CHUNGA :FLEXI-STREAMS :BORDEAUX-THREADS :LPARALLEL.WITH-CLTL2
 :LPARALLEL.WITH-CAS :LPARALLEL.WITH-STEALING-SCHEDULER :SPLIT-SEQUENCE
 :FAST-IO-SV :FAST-IO :CL-JSON-CLOS :CL-JSON :SBCL-USES-SB-ROTATE-BYTE
 CHIPZ-SYSTEM:GRAY-STREAMS :THREAD-SUPPORT :SLYNK CFFI-FEATURES:FLAT-NAMESPACE
 CFFI-FEATURES:X86-64 CFFI-FEATURES:UNIX :CFFI CFFI-SYS::FLAT-NAMESPACE
 ALEXANDRIA::SEQUENCE-EMPTYP :QUICKLISP :ASDF3.3 :ASDF3.2 :ASDF3.1 :ASDF3

ASDF version: 3.3.5
Critical dependencies: (/gnu/store/iyf4hl69dnlcn9zqppjc44xb3jik33qq-cl-cffi-gtk-0.11.2-2.e9a46df/share/common-lisp/source/cl-cffi-gtk/gtk/cl-cffi-gtk.asd
 /gnu/store/am8x4lz5v9dyd7v6h1p5lw223n7x5a2c-cl-gobject-introspection-0.3-1.d0136c8/share/common-lisp/source/cl-gobject-introspection/cl-gobject-introspection.asd
 /home/aartaka/git/cl-webkit/webkit2/cl-webkit2.asd)

Quicklisp dist version: 2022-02-20
Quicklisp client version: 2021-02-13
Local project directories: (/home/aartaka/quicklisp/local-projects/)
Critical dependencies(#<SYSTEM cl-cffi-gtk / cl-cffi-gtk-20201220-git / quicklisp 2022-02-20>
 #<SYSTEM cl-gobject-introspection / cl-gobject-introspection-20210124-git / quicklisp 2022-02-20>
 #<SYSTEM cl-webkit2 / cl-webkit-20220220-git / quicklisp 2022-02-20>)

Guix version: 0

```

aartaka:
# Description

This changes the way we get user script contents and properties, based on whether there is a value to the slots initially. In particular, this avoids implying that scripts are backed by files, which is vital for dynamically generated scripts, like [nx-dark-reader](https://github.com/aartaka/nx-dark-reader). So, 
- once any slot of the script os accessed, 
  - we fill the un-set slots with the new `process-user-script` function.
  - Same happens in `nfiles:deserialize` method, if the script is backed by file/URL.
- All the slots are initially unbound, so `slot-unbound` fires for any accessor and fills the script too.

Fixes #2388.

# Discussion

This change leaves us with untyped slots and no default values for them, let alone a cryptic use of `slot-unbound`. Is there a better way to do that? Maybe introduce some "parsed-or-not" flag slot instead?

# Checklist:

Regarding the tests: I am unable to test whether this change works, because having any extension loading basically stops Nyxt GUI from appearing. For example, having
``` lisp
(load-after-system :nx-kaomoji (nyxt-config-file "kaomoji.lisp"))
```
in my config file cause the most recent (d3469bc37ce467802d5b9ddac60f948baca73ea3) Nyxt (with https://github.com/aartaka/nx-kaomoji at 909c9d88aee86f177ebaa9903772a91b041dcca2) to 
- Create a socket.
- Load the config.
- Start reading the history.
And then stop. No GUI, no logs, nothing.


source/mode/user-script.lisp
```diff
@@ -140,11 +128,8 @@ Return:
   (unless (uiop:emptyp (nfiles:url-content script))
     (alex:write-string-into-file (nfiles:url-content script) destination :if-exists :supersede)))
 
-(defmethod nfiles:deserialize ((profile nyxt-profile) (script user-script) raw-content &key)
-  "If the script is not in the UserScript format, the raw content is used as is
-and only the `code' slot is set."
-  ;; TODO: Parse the stream directly?
-  (let ((code  (alex:read-stream-content-into-string raw-content)))
+(defmethod process-user-script ((script user-script))
```
Ambrevar, line 131: Maybe `parse-content` instead?  We are not really processing here.

source/mode/user-script.lisp
```diff
@@ -45,31 +45,19 @@
...
-    :type (maybe hash-table))
-   (include
-    '()
-    :type list-of-strings)
-   (exclude
-    '()
-    :type list-of-strings)
+   requires
+   include
+   exclude
```
Ambrevar, line 56: Why remove the types though?

source/mode/user-script.lisp
```diff
@@ -184,4 +169,21 @@ and only the `code' slot is set."
...
+  (declare (ignore class))
+  (process-user-script instance)
+  (slot-value instance slot-name))
+
+(defmethod nfiles:deserialize ((profile nyxt-profile) (script user-script) raw-content &key)
+  "If the script is not in the UserScript format, the raw content is used as is
+and only the `code' slot is set."
+  (declare (ignorable profile))
+  (setf (files:content script) (alex:read-stream-content-into-string raw-content))
+  (process-user-script script))
```
Ambrevar, line 189: Problem with the return value: `deserialize` should return a value ready to be used, so here it should return `(code script)`.  You can update `process-user-script` to return that.

source/mode/user-script.lisp
aartaka: It's because types and initforms are tightly coupled. If there's a type—there should be an initform. If there's no initform—there cannot be a type :(

Ambrevar: Are you sure?  This seems to work:

```lisp
(defclass foo () ((slot1 :type number)))
(make-instance 'foo)
(setf (slot-value * 'slot1) "blah") ; <= Raises condition on CCL.
```

Tested on SBCL and CCL.

aartaka: But then, if you do a slightly more involved one...
``` lisp
(class-star:define-class foo () ((slot1 :type (maybe number))))
; Evaluation aborted on #<SIMPLE-ERROR "Missing initform." {100BE88BB3}>
```

aartaka:
I've rewritten it to be simpler and more reliable:
- The whole parsing and checking is now part of `code :around` method.
- The parsing function is now called `parse-user-script` and returns the processed code.
- `include` processing is a bit smarter now (that I just rememebred and couldn't resist fixing :P)

source/mode/user-script.lisp
Ambrevar: That's because define-class expects an initform in position 2.  Pass `:unbound`:

```lisp
(class-star:define-class foo () ((slot1 :unbound :type (maybe number))))
```

source/mode/user-script.lisp
```diff
@@ -185,3 +195,15 @@ and only the `code' slot is set."
                               (otherwise :document-end)))
            code-with-requires)))
      (setf (code script) code))))
+
+(defmethod code :around ((script user-script))
+  (or (call-next-method)
+      (setf (code script)
+            (parse-user-script script))))
```
Ambrevar, line 202: I'm pondering which one between `ACCESSOR :around` and `slot-unbound` is more idiomatic.

They get us very similar results.  The main difference I can think of is that `slot-unbound` is mostly meant for precisely setting the slot, while the `:around` method can be used for other things.

In effect, this occupies a "hooking" function of `code`...

Food for thoughts!

source/renderer/gtk.lisp
```diff
@@ -1680,7 +1680,7 @@ local anyways, and it's better to refresh it if a load was queried."
       (:metaclass mixin-class))))
 
 (define-ffi-method ffi-buffer-add-user-script ((buffer gtk-buffer) (script gtk-user-script))
-  (alex:if-let ((code (nfiles:content script)))
+  (alex:if-let ((code (nyxt/user-script-mode:code script)))
```
Ambrevar, line 1683: Why?

aartaka: That's because `code` is now more reliable than `nfiles:content`:
`code`:
- Processes `nfiles:content` if there's no code yet.
- Works for files, URLs, and string-based `nfiles:content`.
`nfiles:content`:
- Only works for URLs & files (`nfiles:deserialize`).
- Is semanticaly responsible for content fetching, not the structure modification.

source/mode/user-script.lisp
aartaka: Indeed, taking a qualified method for any accessor is a sad thing, but we do need it sometimes. I've thought of three alternatives there:
- `ACCESSOR :around`:
  - Occupies the qualified method.
  - Fires at slot fetching, but only if the accessor is used.
    - But then, the rule of thumb is to use accessors over `slot-value` for most cases, isn't it?
  - Is compatible with initforms and typing just fine.
    - Requires some special initform if we want to use this method as an initializer.
- `slot-unbound`
  - Allows fine-graining the initialization per slot, per class, or per meta-class, which is an awesome level of granularity.
  - Requires the initform to not be present.
    - I didn't know about the `:unbound` hack until now.
- `nfiles:content` hook/qualified method.
  - Not yet possible due to `nfiles:content` being a function.
  - Can be fine-grained to a class of a file, which is enough of a restriction for hooking.
  - Fits in with the file-backed deserialization idea... more or less.

Now that you mentioned the `:unbound` value for initform in `define-class`, I'm leaning back to `slot-unbound`, as it's too reliable and fine-grained to not use it.

aartaka:
Yet another turn (I hope the final one this time): I've moved to `slot-unbound` and `:unbound` initform for mandatory slots (`code`, `include`, `exclude`, `all-frames-p`, `run-at`, `world-name`). Looks pretty usable now.

source/renderer/gtk.lisp
Ambrevar: Semantically this description seems upside down: nfiles:content should call `code`, not the other way around.

`nfiles:content` should work for "virtual" script to: if they have no URL, no file, and no code, then it should return nil.  If they have `code` because `(setf (file:content ...) ...)` was called, then `nfiles:content` will appropriately return the set value.

Ambrevar:
I've been thinking about our discussion yesterday and whether `nfiles:content` should be a method.

The "classic" use case of nfiles is when we define a class, like `bookmarks-file`, which goal is to return some list of bookmark objects for instance.

Note there is a decoupling between the Nfile object and the "target" object.

But in the case of user-scripts (and `blocker-mode` hostlist), the target object is also the Nfile object!  That would mean that `nfiles:content` is supposed to return... the Nfile object itself!?!
It's a bit mind-boggling, hence the confusion that arose in this discussion.

So what happens when we `(setf (nfiles:content USER-SCRIPT) SCRIPT-STRING))`?  It works, in the sense that `(nfiles:content USER-SCRIPT)` returns `SCRIPT-STRING`.  But if the user script has no local file we are left with an undesirable side-effect: the result of  `(nfiles:content USER-SCRIPT)` no longer matches `USER-SCRIPT` semantically (different `code`, etc.).

What does `blocker-mode` do?  When the content is set, it calls its `files:write-file` specialization which saves the `files:url-content`, thus ignoring the new content value.

Now for blocker-mode, this is fine because we never set the content directly, the point of hostlists is to be loaded via a URL or a file, that's it.

What if there was no persisted file and we would just set the content in memory?  Then we are fine too, because `nfiles:content` always returns the expect value. 

So in these special cases, nfiles is useful as a caching mechanism that has the same API as for nfiles with URLs and local files.

To conclude on `user-scripts`: What if you set the content to `(parse-user-script MY-VIRTUAL-CONTENT)`?  Then the structure would not match, but that would not matter since you would call `(files:content USER-SCRIPT)` in `renderer/gtk.lisp`.

aartaka:
Yes, that's a fair assessment! 

One change may be due: mention this edge-case in the documentation, maybe?

source/mode/user-script.lisp
```diff
@@ -176,12 +175,42 @@ and only the `code' slot is set."
...
+  (declare (ignore class))
+  (parse-user-script instance)
+  (slot-value instance slot-name))
+
+(defmethod nfiles:deserialize ((profile nyxt-profile) (script user-script) raw-content &key)
+  "If the script is not in the UserScript format, the raw content is used as is
+and only the `code' slot is set."
+  (declare (ignorable profile))
+  (prog1
+      (setf (files:content script) (alex:read-stream-content-into-string raw-content))
```
Ambrevar, line 215: Maybe a misunderstanding: you should not be setting the files:content at all, as it is set by nfiles when calling `read-file`.

The methods of Nfiles are meant to be functional, it's the Nfiles internals that are in charge of setting stuff (cache, etc.).

So in this case I would set the raw-content to the `code` slot, then call `parse-user-script` to tweak the slot to include the `requires`.

Makes sense?

Ambrevar: Anything blocking?

aartaka:
I don't exactly agree with the last set of comments, @Ambrevar. While you are right that Nfiles use should be mostly functional and that we should somehow avoid setting `nfiles:content` explicitly, I don't see how that can work for this particular case.
- File/URL-bcked scripts are fine, they just fetch the code from the respective place.
- "Virtual" scirpts are worse off, though—they only have `nfiles:content` that they can rely on in parsing and proper slot setting.

So, the only reliable slot/accessor that user scripts have is `nfiles:content` and we have to adjust the behavior for one of the types:
- Either set `nfiles:content` in advance for the file/URL-backed scripts, so that `parse-user-script` does the right thing.
- Rewrite the `parse-user-script` to rely on some other slot (`code`?) so that file/URL based scripts don't need their `content` destructuvely modified.
  - But then, how do we set the `code` for virtual scripts is they only have `nfiles:content` and cannot rely on methods other than `slot-unbound`? If we somehow set `code` to a new value, `slot-unbound` would not fire for it, making the whole `slot-unbound` business meaningless. The meaningfull slots should remain unbound up until the moment of parsing!

To conclude: while `setf`-int `nfiles:content` is not exactly nice, it is more or less reasonable semantically, while being the only common thing for both file/URL-backed scripts and "virtual" ones :)

Ambrevar:
> * "Virtual" scirpts are worse off, though—they only have `nfiles:content` that they can rely on in parsing and proper slot setting.

Why?  Scripts are class instances, so you can use them with all their accessors, like the `code` accessor.
What I'm saying is that we should focus use `user-scripts` as a "direct class" and not go through the Nfiles interface.  This would work for kinds of scripts.

In the end, the only adjustment I suggest to the current patch is this:

```diff
 (defmethod parse-user-script ((script user-script))
-  (let ((code (files:content script)))
+  (let ((code (slot-value script code)))
     (or
      (sera:and-let* ((start-position (search "// ==UserScript==" code))

```

and

```diff
 and only the `code' slot is set."
   (declare (ignorable profile))
   (prog1
-      (setf (files:content script) (alex:read-stream-content-into-string raw-content))
+      (setf (code script) (alex:read-stream-content-into-string raw-content))

```

By the way, the `slot-unbound` approach should probably only call `parse-user-script` is the slot is `code`, right?

Then I'd write it like this:

```diff
 (defmethod slot-unbound (class (instance user-script) slot-name)
   (declare (ignore class))
-  (parse-user-script instance)
+  (when (eq 'slot-name 'code)
+    (parse-user-script instance))
```

Ambrevar:
> Use the `setf`-watcher for `code` to re-parse the scrips.
> 
>     * When deserializing the script, `setf` the `code`, thus forcing re-parsing.

I don't think a watcher is required here, since the `code` slot is unlikely to be changed once properly initialized.

aartaka:
@Ambrevar, I've tested it, works just fine for virtual files! Merging then?

aartaka:
Oh wait no, I need to add some tests to check whether it all works properly!

aartaka:
I've added some simple tests and manual section showcasing the user script functionality. I've also tested both file-backed script and virtual ones. Safe to merge, I believe.

Ambrevar:
Ha, got bitten by the buffer presence assumption!
See the manual how I fixed it.

aartaka:
> Ha, got bitten by the buffer presence assumption!

I don't understand (◎_◎;)

> See the manual how I fixed it.

Where?

PR #2393, opened 2022-06-21
Nyxt ASDF extension

### LINKED ISSUE #2362: Extend ASDF class for Nyxt
Currently we have quite some code in our ASD.

It's bad practice to write define globals in an .asd, because it may clash with other libraries.
But it's also bad practice to use a different package, because, well, we can do better: write an ASDF extension.

This would be useful to make it trivial for extension writers to embed our Git submodule logic in their extension.  All they'd have to do is add

```lisp
:defsystem-depends-on :nyxt-asdf
```
to their definition.

Ambrevar:
# Description

Previously we used to be very impolite towards `asdf-user` and pollute the package with functions, defvar, and worse of all, we overwrote the `perform` method :/ 

Instead, we are supposed to write an ASDF extension.  The benefit is that now all the code previously used in the .asd is available to third-party extensions, most importantly submodule fetching is now as simple as

```lisp
(defsystem "nx-MYEXT/submodules"
  :defsystem-depends-on (nyxt-asdf)
  :class :nyxt-submodule-system)
```

Fixes #2362.

# Discussion

There is one dark corner I do not quite get.

ASDF says in the manual, section "6.2 A more involved example":

> You should only use 'in-package' (and before it, a 'defpackage') when you're
> going to define new classes, functions, variables, macros, etc., in the '.asd'
> file, and want to thereby avoid name clashes.  Manuals for old versions of ASDF
> recommended use of such an idiom in '.asd' files, but as of ASDF 3, we recommend
> that you don't do that anymore, and instead define any ASDF extensions in their
> own system, on which you can then declare a dependency using
> ':defsystem-depends-on

But to use custom system classes, we must

> For such class names in the ASDF package, we recommend that the ':class' option
> be specified using a keyword symbol, such as
> 
>      :class :MY-NEW-SYSTEM-SUBCLASS
> 
>    This practice will ensure that package name conflicts are avoided.
> Otherwise, the symbol 'MY-NEW-SYSTEM-SUBCLASS' will be read into the
> current package _before_ it has been exported from the ASDF extension
> loaded by ':defsystem-depends-on', causing a name conflict in the
> current package.

So from that I understand that to use the `nyxt-system` custom class, we should be in the `nyxt-asdf` package, which goes against the first quotation.

I looked at what `prove-asdf` does and it import the class symbol to the `asdf` package (ouch!).  So instead I've opted to import it into the `asdf-user` package.  I guess that's a reasonable compromise.

# To do:

- [x] `npath` could probably be turned into a method, so that we could remove this dirty `(load "nyxt-asdf-utils")`.
- [x] `nyxt/install` could be further specialized by creating new component types.  For instance, the `lib-file` would automatically install to `/usr/lib` and make it executable, etc.
- [x] Move `nyxt-asdf` into `libraries/`?  It's a bit special, but does it matter?

# Checklist:


Ambrevar:
I've implemented support for various installation rules.  Compare:

## Old `nyxt/install`

```lisp
(defsystem "nyxt/install"
  :depends-on (alexandria
               str
               #.(format nil "nyxt/~a-application" *nyxt-renderer*)
               nyxt/version)
  :perform (compile-op
            (o c)
            (labels ((ensure-parent-exists (file)
                     (ensure-all-directories-exist
                      (list (directory-namestring file))))
                     (install-file (file dest)
                       (ensure-parent-exists dest)
                       (copy-file file dest)))
              (let ((desktop-file (merge-pathnames* "applications/nyxt.desktop" *datadir*)))
                (install-file (system-relative-pathname c "assets/nyxt.desktop")
                              desktop-file))
              (mapc (lambda (icon-size)
                      (let ((icon-file (format nil "~a/icons/hicolor/~ax~a/apps/nyxt.png"
                                               *datadir* icon-size icon-size)))
                        (install-file (system-relative-pathname
                                       c
                                       (format nil "assets/nyxt_~ax~a.png"
                                               icon-size icon-size))
                                      icon-file)))
                    '(16 32 128 256 512))
              (let ((binary-file (merge-pathnames* "nyxt" *bindir*)))
                (install-file (system-relative-pathname c "nyxt") binary-file)
                (make-executable binary-file))
              (flet ((copy-directory (source destination &key verbose-p)
                       "Copy the content (the file tree) of SOURCE to DESTINATION."
                       (when verbose-p
                         (format *error-output* "~&;; Copy ~s/* inside ~s.~%" source destination))
                       (collect-sub*directories
                        (ensure-directory-pathname source)
                        (constantly t)
                        t
                        (lambda (subdirectory)
                          (mapc (lambda (file)
                                  (unless (member (pathname-type file) '("o" "fasl") :test 'equalp)
                                    (let ((destination-file
                                            (merge-pathnames*
                                             (subpathp file (ensure-directory-pathname source))
                                             (ensure-pathname destination :truenamize t :ensure-directory t))))
                                      (install-file file destination-file))))
                                (directory-files subdirectory))))))
                (handler-case
                    (progn
                      (format *error-output* "~&;; Copying Git files to ~s.~%" *dest-source-dir*)
                      (dolist (file (mapcan (lambda (dir)
                                              (split-string
                                               (run-program `("git" "-C" ,(native-namestring
                                                                           (system-source-directory :nyxt))
                                                                    "ls-files" ,dir
                                                                    ;; Do not install this non-Lisp source:
                                                                    "--" ":!:libraries/web-extensions/*")
                                                            :output '(:string :stripped t))
                                               :separator '(#\newline #\return #\linefeed)))
                                            '("source" "libraries")))
                        (let ((dest (merge-pathnames* file *dest-source-dir*)))
                          (install-file (system-relative-pathname :nyxt file)
                                        dest))))
                  (condition (c)
                    (format *error-output* "~&;; Git copy error: ~a" c)
                    (format *error-output* "~&;; Fallback to copying whole directories instead.~%")
                    (dolist (dir '("source" "libraries"))
                      (copy-directory (system-relative-pathname :nyxt dir)
                                      (merge-pathnames* dir *dest-source-dir*)
                                      :verbose-p t))))
                (let ((libnyxt-dest (merge-pathnames* "libnyxt.so" *nyxt-libdir*)))
                  (install-file (system-relative-pathname :nyxt "libraries/web-extensions/libnyxt.so")
                                libnyxt-dest)
                  (make-executable libnyxt-dest))
                (install-file (system-source-file :nyxt) (merge-pathnames* "nyxt.asd" *dest-source-dir*))))))
```

## New `nyxt/install`

```lisp
(defsystem "nyxt/install"
  :defsystem-depends-on (nyxt-asdf)
  :class :nyxt-renderer-system
  :depends-on (alexandria
               str
               nyxt/version)
  :components ((:nyxt-desktop-file "assets/nyxt.desktop")
               (:nyxt-icon-directory "assets/")
               (:nyxt-binary-file "nyxt")
               (:nyxt-library-file "libraries/web-extensions/libnyxt.so")
               (:nyxt-source-directory "source")
               (:nyxt-source-directory "nyxt-asdf")
               (:nyxt-source-directory "libraries"
                :exclude-pattern ":!:libraries/web-extensions/*" ; Do not install this non-Lisp source.
                :exclude-types '("o" "c" "fasl"))))
```

Neat, nah?

jmercouris:
Would you like me to review?

Ambrevar:
Noooooo, dooooooon't! :D 

Just joking, of course reviews are always welcome! ;)

Ambrevar:
All done! 225554bf65116eb2b5a810cb83689f0540fb0fc8

PR #2408, opened 2022-06-23
Overhaul settings

### LINKED ISSUE #2296: Improve and extend `common-settings`
`common-settings` is for now a bit barren but we can easily see how it can be expanded to be a full-fledged configuration panel.

Now the problem is that it's very fragile and it has broken so many times in the past.  Indeed, since it's mostly made of arbitrary forms, there are barely any checks done, ever.

Besides, for now we just write to `auto-config.lisp` and we don't update the live instances.

I believe we could hit 2 birds with one stone: by updating the live instances, we might be able to perform more checking.

To sum up, each settings should be applied at 3 levels (optionally letting the user choose where):
- On the existing buffers.
- On the new buffers (`customize-instance` and `customize-hook`).
- For the next session (`auto-config`).

------

Some very useful settings:

- [ ] Ensure that user-defined (or extension-defined) keyschemes are listed.
- [ ] Dynamically preview theme styles. 
- [ ] Place "Edit user files" at the top.
- [ ] Restore session: always, ask, never.
- [ ] Make default browser, see #2439.
- [ ] Spell check language (it's a mode, should we have per-mode settings?)
- [ ] Download directory.
- [ ] Default modes.
- [ ] Proxy (also a mode)
- [ ] "set default new buffer URL" should have suggestions like "nyxt:new", "nyxt:dashboard".
- [ ] Customize search engines.
- [ ] Cookie settings.
- [ ] Intelligent Tracking Protection (it's part of `reduce-tracking-mode`, but maybe we should factor it out).
- [ ]  [Text editing keybindings](http://tiborsimko.org/emacs-bindings-in-gtk-apps.html).
- [ ] Font families 
- [ ] Removal of the data files (clear history, as one use-case for that).
- [ ] WebKit cache clearing _per category — cookies, history, cache, and a dozen others_.
- [ ] [WebKitSettings](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitSettings.html), like buffer background and default font, lots of tasty stuff there. We may need to somehow propagate those to all buffers, maybe even maintaining a general set of [WwebKitSettings](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#webkit-web-view-set-settings) to apply by default.

Ambrevar:
# Description

This does multiple things:

- Adds a new `apply-configuration` helper to apply configuration to auto-config, current-instance, new instance, all instances.
- Turn Common Settings into a form-based page, as per @aartaka suggestion.
- Connect the above two!

This frees us from having to code the settings in `nyxt/lisp-eval` forms, which are a bit brittle and would involve complex PS.  Here we are mostly in pure Lisp.

Fixes #2296

# Discussion

The value we get for themes is a symbol, e.g. `theme:+light-theme`.  So to effectively switch theme, we have to call `eval` on it, which is ugly!

Possible fix: Make a `dark-theme` subclass and call `(make-instance 'dark-theme)` to set the slot of the browser.

@aartaka ?

# To do:

- [ ] Turn rest of Misc settings into the new HTML form.
- [ ] Test that it saves to auto-config properly.
- [ ] Test that it applies to new instances.
- [ ] Test that it applies to multiple instances.
- [ ] Reset "Common Settings" URL after form has been received.

- [ ] Update `configure-slot` to use the same API.
- [ ] The `setting`s function should probably be new first- class, just like commands.

# Checklist:

- [ ] My code follows the style guidelines for Common Lisp code
- [ ] My changes generate no new warnings


Ambrevar:
@jmercouris @aadcg @aartaka Please let me know if my approach seems reasonable to you.

aartaka:
I'll be able to test in a day or two, if you don't mind :)

aadcg:
Same here!

source/configuration.lisp
```diff
@@ -284,6 +284,69 @@ Return NIL if not a class form."
...
+   (make-instance
+    'hooks:handler
+    :fn (lambda (object)
+          (declare (ignorable object))
+          ;; TODO:
+          (when fun
+            (funcall fun object))
+          (when slot
+            ;; TODO: Can we use accessor / writer?  How do you call (setf ...)
+            ;; on a dynamically-found writer?
```
aartaka, line 308: The best I was able to come up with was:
``` lisp
(let ((slot 'nyxt:id)) (eval `(function (setf ,slot))))
```

aartaka (approved):
Yes, the general approach is alright, although I believe radio buttons are terribly inferior to `<select>` menus that you can style much better than radios ;)

source/configuration.lisp
Ambrevar: Ouch! :D

Ambrevar: This seems to work:

```lisp
(closer-mop:ensure-generic-function
  (let ((slot 'my-slot))
   `(setf ,slot)))
```

Then you can funcall the result.

It only works for generic-functions, but that's what we need here.

Ambrevar:
@aartaka I cannot get `lisp-eval` callbacks to work.
First, it's currently broken because `chain` should be prefixed with `ps:` since it's no logner in the Parenscript package.
But even then, in 

```lisp
(ps:when ,callback
         (ps:chain request
                   (then (lambda (response)
                              (when (@ response ok)
                                (ps:chain response (json)))))
                   (then ,callback)))
```

The response.ok is always false.  From the JS console I see:

```log
Response

body: null

bodyUsed: false

headers: Headers {append: function, delete: function, get: function, has: function, set: function, …}

ok: false

redirected: false

status: 0

statusText: ""

type: "opaque"

url: "" 
```

Any idea?  Something wrong on cl-webkit's side?

Ambrevar:
See #2079: maybe it broke it.

Ambrevar:
I won't have much time today and this week... @aartaka @aadcg Any chance one of you would have time to finish this?  It's an easy one, just need to make the new "settings" functions with the corresponding HTML.

aartaka:
Count me in!

Ambrevar:
Thanks!

Ambrevar:
Shall we wait for this for the pre-release-1 or shall we release now then?

aartaka:
My perfectionism makes me want to bring this into the pre-release, but it's unlikely that I'll get to it today, so maybe we should not wait anymore :D

Ambrevar:
I'm running late today, I'll do the release tomorrow morning anyways.

aartaka:
Then let's agree this PR is not included in release tomorrow to not rush ourselves :D

Ambrevar:
To do: Use `closer-mop:slot-definition-writers`.

Ambrevar:
@aartaka Can you summarize your latest changes?

aartaka:
I've merely fixed some errors that I encountered while playing with this branch, and added a new interface type (`"option"`) for configuration entries.

aartaka:
Getting back to this PR, I don't like the direction it takes, in several aspects:
- The classes introduced for the purpose of configuration generation complicate reading and understanding the thing a lot—there are too much layers to have the whole picture in one's mind.
- The whole system seems to have been designed top-down, with _all the possible use-cases_ that a configuration system might have, pre-baked into it.
- There's no direct correspondence between our object system and configuration, we have to conjure a whole set of new entities just to configure a class slot. Fells like downplaying the power of CLOS/MOP/user-classes here.

Given this, I want to try another course of action:
- Start with 
  - slot-only,
  - restart-required,
  - MOP-backed,
  - auto-generated HTML form configuration.
- Expand it with runtime options and misc options (maybe even leave those as-is now, because there are always exceptions).
- And make it rather a Spinneret tag inspecting classes than classes generating HTML. 

As an external API I suggest a Spinneret tag, say `:nconfig`:
``` lisp
(:nconfig
 (browser
  restore-session-on-startup-p
  open-external-link-in-new-window-p
  (default-new-buffer-url :choices (list "nyxt:new" "nyxt:dashboard" "https://nyxt.atlas.engineer"))
  native-dialogs
  (theme :choices (list theme:+light-theme+ theme:+dark-theme+))
  external-editor-program
  style)
 (web-buffer
  smooth-scrolling
  search-auto-complete-p
  search-always-auto-complete-p
  keep-search-hints-p
  zoom-ratio-default
  (style :type string))
 (window
  message-buffer-height
  prompt-buffer-open-height
  (message-buffer-style :type string))
 (status-buffer
  height
  glyph-mode-presentation-p
  (style :type string)))
```
Inferring the types of most slots via MOP, and using the `:type` or `:choises` field when unclear or overriden, and generating HTML tags for different types, like generating `<input type=number>` for `integer`slots, `<input type=radio>` for boolean slots, and some `<select>`s for `member` types, passing raw Lisp data (`nyxt:` URLs allow tht) to `apply-configuration` instead of some pre-set values or function processors.

This way our configuration would be:
- Extensible (much like the current one, but with less entities).
- Embeddable (even for extensions).
- Inferred/inspectable (via MOP).
- Simple (a mere Spinneret macro/widget with some MOP magic underneath, instead of a set of classes designed specifically for the purpose of configuration).

Does that sound sane? (゜▼゜＊）

Ambrevar:
Yes for the Spinneret approach.  

However I disagree that it should be "restart only": this would severely limit us (and limit the user) and the API we would thus design is almost guaranteed to not be extensible enough to accommodate for a "not just restart only" enhancement in the future.

aartaka:
Fair ( ͡° ͜ʖ ͡°)

Ambrevar:
Quick summary for whoever is going to resume work on this (maybe @hgluka, maybe someone else, maybe me :p):

1. Work on a new branch :)
2. Rebase on master (should be relatively easy).
3. Replace the heuristic to find writer methods with the CLOS way to do it (e.g. `c2mop:slot-definition-writers`).
4. Make use of slot type information to guard against bad input in `apply-configuration`?

Ambrevar:
Let me clarify a bit the design behind this pull request and the discussion I had with @aartaka about leveraging spinneret tags:

The pull request has 2 parts:
- A "backend" which is centered around `apply-configuration` and which could be potentially extracted to a separate library and be used in any other CL project. (Including one that does not use HTML as a UI.)
- A "frontend" which is for now just the overhaul "Common Settings" page, and which this pull request defines "manually".

If I understood Artyom correctly, he suggested to leverage spinneret tags to write simplify the declaration of the "Common Settings" page.  Note that this won't have an impact on the "backend".  I suggest that we do this but in a later pull request.  For now it's OK to have a "manual" Common Settings page, what matters is that we have a solid backend.

Ambrevar:
Note about the commented `(setf (gethash :keystyle *settings*) ...)` and friends: I left this behind to show what I had tried to specialize some settings, but this is all rather ugly.

Instead, a CLOS approach as done in the latest commit seems much better: apply settings in the most generic fashion, only to specialize with `setting` subclasses whenever necessary.

jmercouris:
Please reopen when ready.

PR #2410, opened 2022-06-23
Migration suggestions

### LINKED ISSUE #2301: Add 2-series -> 3->series migration guide?
Considering _configuration_ is the main selling point of Nyxt, having a configuration / extensions migration guide would not be too much to ask for...

Just like the manual and the changelog, I can see how embedding it as a .lisp file would be more appropriate than an article that's prone to go quickly out-of-date (not mentioning that it wouldn't be accessible offline, etc.).

Agreed?
Thoughts?

There is lot's of stuff to add in there!

Ambrevar:
# Description

This is an attempt to display suggestions on startup error.

Fixes #2301.

# Discussion

@aartaka In commit b111257a9c86dfc8247d57b2ef7480c82f9a7187 I tried to add a spinneret tag that enables cross-references on symbols.

Unfortunately it does not work, the callbacks is never registered (the lisp-url-callbacks slot is empty).

# To do:

- [x] Actually display the suggestions.

# Checklist:


aartaka:
Hmmmmm. Spinneret may be unpredictable with tag names sometimes. 
- What if you use keyword? 
- What if you change the arglist (I rember deftag argument lists being somewhat unintuitive)? 
- What if you define this tag in the Spunneret tags package (or in the spinneret-tags.lisp)?

Ambrevar:
Thanks for the help, it worked!  See https://github.com/atlas-engineer/nyxt/pull/2410/commits/94201299a54492a3f5bc8dc787dc46354b76e5b4.

Note that I had to delay spinneret evaluation using lambdas.
Indeed, the migration guide is defined at compile time, at which time there is no buffer.

Alternatively we could also not generate the guide at compile time and place everything in functions.

Other than that, ready to merge.

jmercouris:
Just tested. It of course catches some basic migration errors.

It does seem a bit too engineered to me for what it produces. Whether it will be useful remains to be seen. I think it suffices that we have "versioned" configuration files for the user. This is of course just my opinion! In any case, from the perspective of the code, it looks fantastic. Thanks Pierre!

Ambrevar:
> It of course catches some basic migration errors.

To clarify, the "quality" of the suggestions heavily depends on how many we define in the "guide".  For now it's only very few of them, we will fill it up before 3.0.



> I think it suffices that we have "versioned" configuration files for the user.

Only auto-config is versioned (in the file name), if this was what you meant.

These migration suggestions are supposed to help with the `config.lisp` file.

> It does seem a bit too engineered to me for what it produces.

Can you explain what you mean?

jmercouris:
> > It of course catches some basic migration errors.
> 
> 
> 
> To clarify, the "quality" of the suggestions heavily depends on how many we define in the "guide".  For now it's only very few of them, we will fill it up before 3.0.
> 
> 
> 
> 
> 
> 
> 
> > I think it suffices that we have "versioned" configuration files for the user.
> 
> 
> 
> Only auto-config is versioned (in the file name), if this was what you meant.
> 
> 
> 
> These migration suggestions are supposed to help with the `config.lisp` file.
> 
> 
> 
> > It does seem a bit too engineered to me for what it produces.
> 
> 
> 
> Can you explain what you mean?

I am saying that in its current iteration, we could simply assemble the strings manually as part of a migration guide that we write. Maybe another approach would be to look at changelog.lisp and extend the macro there to add this functionality, I think it may be simpler if all of this information is contained in one place. What do you think?

Ambrevar:
I don't think that would work, because the changelog includes too much noise.  The relevance of the migration suggestions is critical, otherwise the users won't bother reading lists of items that don't help them.

Another point is that if we tie our changelog to the migration guide, this will greatly restrict us in how we write our changelogs, lest we make the migration suggestions unreadable.

With the current approach, changelog authors have a lot of freedom, it's a good thing in my opinion.

aadcg:
Looks good to me.

Ambrevar:
Merged with 1ed2ac247b82c474146a2de4467dc10f3f495625.

Issue #2227, opened 2022-05-05
Avoid `with-current-buffer` pattern when evaluating JS


aadcg:
This is follow-up to #2220.

@aartaka and I noticed there's a frequent pattern in the codebase that goes along the lines of:

```lisp
(with-current-buffer buffer-object (peval|pflet ...) ...)
```

Notice that the macros `peval` and `pflet` execute JS synchronously in the `current-buffer`. 

We suggest that these macros should support a keyword argument that would specify the buffer.

What do you think? @jmercouris @Ambrevar @pdelfino

jmercouris:
Yes, it is a good point! It seems a suitable addition for peval, but not pflet. I can't imagine how the syntax would look for pflet!

Ambrevar:
`(pflet :buffer BUFFER (BINDINGS) BODY)`?

jmercouris:
Feels wrong, but it's OK I guess :-D

aadcg:
> Feels wrong, but it's OK I guess :-D

Care to expand?

jmercouris:
It just bothers me because it breaks with the expectation of it functioning purely as a let/flet type binding.

aartaka:
> It just bothers me because it breaks with the expectation of it functioning purely as a let/flet type binding.

Well, that's a part of how it works, so we better be explicit about it being buffer-dependent than reinforce a faulty expectation :)

aartaka:
I've added the 
``` lisp
(peval :async t :buffer buffer (ps:chain document active-element value))
``` 
syntax for `peval` and 
``` lisp
(pflet ((foo :buffer b ()
             (ps:chain window (prompt "How are you?")))
        (bar :async t :buffer b2 ()
             (alert "hello")))
  (do-something-here))
```

See https://github.com/atlas-engineer/nyxt/pull/2419#issuecomment-1219403939 for the syntax reasoning.

Issue #2189, opened 2022-04-12
Gemini: allowlist SSL ceritificates


aartaka:
Currently, we rely on `phos` and underlying `cl+ssl` to check Gemini pages for certificate validity. In case some certificate is invalid or self-signed (quite a frequent thing in Geminispace, actually), it simply errors out and stops loading the page. 

Most Gemini clients allow one to allowlist the certificate for a certain domain. We should probably do the same.

aadcg:
Contributions welcome, unplanned.

I think it was a mistake to add gemini and gopher support in Nyxt. Those features belong to an external extension that is maintained by a third party.

PR #2419, opened 2022-06-27
Improve hinting UI and its user-options

### LINKED ISSUE #536: Feature request: hide/narrow down hinted links as one narrows to seach
## Brief description

I think I discussed this previously on IRC, with mixed reactions but, here goes:

When `find-hint` (or equivalent) as you type the title or url you want to go to, I'd like the hints in the buffer to also be narrowed down (see image below)

## Motivation

I find this very useful, because:

- often I do not see the link I want to for all the "hinting" (think: _the
forest obscures the trees_), 

- and also, my mind doesn't know what index the link I want to follow has before I've issued the `find-hint`, so as I hit `C-g`, my mind already knows I can start typing "rome" to narrow down to what I want and have the screen less cluttered (it also re-indexes the links to be shorter to type).

- Once I've only have 3-4 links on screen, I can just C-n/C-p (or arrow) once or twice to get to the desired link, if my mind doesn't feel like typing the exact index.

- Another _highly useful_ way to use this, that I use probably daily, is as a search feature. I.e. I know the url I want to is somewhere one the screen, and I know (part of) the title, but there are too many links to find it, e.g. something with "rome" in the example image, so just start typing, and I see exactly which links are of interest.

## Summary
If I do narrow to search, why show non-matching link hints? They are by definition uninteresting.

## Example 
_(from conkeror)_
In this example I maybe want "Roman Greece", so I instantly start typing "rom" and instantly see the different links reagrding Rome and ancient Greece, and then just smash `4`, or do 3 `C-n`

myimage

### LINKED ISSUE #538: Feature request: Hinting style could be better?
## Introduction

If I split my tiling WM in two equally sized panes, and open Next on left and [Conkeror](http://conkeror.org/) on right, and in both visit Next's github issues page, you see differences in how expressive they are, or how "good" they are at showing hinted links:

## Example
Please open below image in separate buffer/tab, to follow my reasoning below:
2020-01-27-085110_1600x900_scrot

## Thoughts

In my mind I currently prefer the right, for the following main reasons:

1. it indexes far more links, while next obscures almost half the screen with a minibuffer, that really doesn't serve much purpose. I find I only use/need Next's minibuffer, when/because it's unclear what the labels are indexing, which leads me to the next point

2. on the right, I see exactly the extent of each link (yellow field), where as in Next, there areindex-badges that just appear to be sitting way off, because I can't see which link it's referring to. For example, consider indexes `DA`, or `DC` in Next above (left), what are they? I was genuinly confused when I saw it, (this is where I started looking in the minibuffer). But on the right side version, I instantly see: "oh, there's a small link there, and it's the user name of the issue author, and the index is indeed just right of the link"

3. Further examples of above points: In Next above, it's very unclear to me e.g. what `BT`, `BU`, `BV` are indexing. Is `BV` the _Pull Request_ tab? Or _Issues_ tab? 

4. This is off topic, but I noted just now, that there's no "Filters" button (`21` in Conkeror to the right), but I assume that's something related to the underlying web engine?

I assume people will feel attached to current hint-badges, and minibuffer, but if there was some option/way to get the functionality of the right (i.e. no minibuffer for hinting, and badge also marks the full link interaction area), I'd use it.

### LINKED ISSUE #2185: Hint actions without prompt

### LINKED ISSUE #2292: Compute hints in viewport
We're very good and fast at computing all hints on pages full of elements, but it'd be less confusing and more visually appealing to compute them as the viewport changes.

aadcg:
# Description
### **After**:

Closes #2292 
Closes #2185 
Closes #538
Closes #536.

# Discussion

The algorithm for `element-in-view-port-p` is still rather primitive, and prone to errors. Find an example below of an element that is not visible, but marked as in viewport (hint marked as **"D"**):

**Question**: Should `element-in-view-port-p` be a parenscript macro? I don't think so. It's a procedure, and most likely one we'll need to specialize depending on the situation.

**Question**: Should we automatically (probably via hooks on input events) re-compute hints as the viewport changes? I'd say no. It's trivial from the user to cancel ans re-issue the command.

**Issue**: JS relative to `overlapped-p`.

# Checklist:


source/prompt-buffer.lisp
```diff
@@ -198,6 +198,13 @@ See also `show-prompt-buffer'."
           (show-prompt-buffer next-prompt-buffer))
         (setf (ffi-window-prompt-buffer-height window) 0))))
 
+(defun collapse-prompt-buffer (prompt-buffer)
+  "Collapse PROMPT-BUFFER's suggestions."
```
Ambrevar, line 202: The docstring is not very clear.  If I understand this function properly, this is meant to shrink the _height_ of the buffer to just fit the input area, right?

source/mode/hint.lisp
```diff
@@ -227,6 +230,10 @@ FUNCTION is the action to perform on the selected elements."
                              :multi-selection-p multi-selection-p
                              :constructor (lambda (source)
                                             (declare (ignore source))
+                                            (unless (use-prompt-buffer-p
+                                                     (find-submode 'hint-mode))
+                                              (nyxt::collapse-prompt-buffer
+                                               (current-prompt-buffer)))
```
Ambrevar, line 236: While this works, it seems a bit convoluted.  This feature is useful and could be generalized.

We currently set the height in `show-prompt-buffer`.  So maybe we could change this function and have `height` accept some values:
- a number (height in pixels)
- `:default` for `(prompt-buffer-open-height (window prompt-buffer))`
- `:fit-to-prompt` for (collapse-prompt-buffer prompt-buffer)`.

Then set up a `height` slot in the `prompt-buffer` class.

Thoughts?

Ambrevar:
It's a very welcomed feature, thanks!

Why did you list #2189 as a fixed issue?  Does not seem related :p  Mistake?

source/prompt-buffer.lisp
aadcg: Yes. The correct behavior, in my opinion, would be to draw it in the echo area (below `status-buffer`). But it's not a priority I think.

aadcg:
> Why did you list #2189 as a fixed issue? Does not seem related :p Mistake?

Oh, absolutely! I'll fix it!

aadcg:
Still need to address `object-attributes`.

Ambrevar:
I'm not sure what you are asking exactly.  If what you want is prompt that matches only against the hints (and not the title, etc.), then I can think of at least 2 ways to do it:
- Specialize object-attributes against the source to only display the hints.
- Set the `filter-*` slots to match only against the hints.

aadcg:
With respect to the way it behaves, I'm happy.

source/mode/hint.lisp
```diff
@@ -174,6 +177,14 @@ define which elements are picked up by element hinting.")
 (define-class hint-source (prompter:source)
   ((prompter:name "Hints")
    (prompter:selection-actions-enabled-p t)
+   (prompter:filter
+    (if (shrink-prompt-buffer-p (find-submode 'hint-mode))
+        (lambda (suggestion source input)
+          (declare (ignore source))
+          (str:starts-with-p input
+                             (cadar (prompter:attributes suggestion))
```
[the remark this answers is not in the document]
Ambrevar: I think you want to either get the default attribute (assuming the hint is the default) using `prompter:attributes-default`, or, if it's not the default, just get the alist value with `alex:assoc-value`.

aadcg:
Working on #2292.

aadcg:
Still working on #2292. I have implemented a new way to compute the hint strings. Each has length 2, which gives 676 with an alphabet of 26 characters. That's more than enough to hint elements in viewport, even with smaller alphabets. 

I still need to make plump and JS talk to each other properly.

aartaka:
2 character-capped hints don't feel right :(

aadcg:
> 2 character-capped hints don't feel right :(

Could you explain?

jmercouris:
I believe Artyom is saying that the algorithm to generate hints should support any arbitrary amount of hints. If we can support more than two characters, we could support an infinite amount of hints. This would of course be better.

aartaka:
Possible solution to the viewport calculation slowness: you can check&store the viewport belonging of an element when parsing it in `nyxt/dom:get-document-body-json`. That should mitigate most performance problems stemming from Lisp->C->JS->C->Lisp communication.

aadcg:
Almost ready. Will mark it as ready for review soon.

aadcg:
Renaming the branch on the remote closes the PR, sorry. I guess `add-barebones-hinting-ui` isn't a bad name for this branch.

aartaka:
¯\\\_(ツ)_/¯

aadcg:
Help me, I'm a parenscript potato.

Can someone please make the following work according to the comment? @aartaka @jmercouris 

```lisp
(define-parenscript overlapped-p (id)
  "Whether ELEMENT is hidden by another overlapping element."
  (ps:let* ((element (nyxt/ps:qs-nyxt-id document (ps:lisp id)))
            (rect (ps:chain element (get-bounding-client-rect)))
            (nearest-element (ps:chain document (element-from-point (ps:chain rect left)
                                                                    (ps:chain rect top)))))
    ;; element is overlapped iff the (ps:chain element parent-node) call returns
    ;; the following error:
    ;; TypeError: null is not an object (evaluating 'element.parentNode')
    (loop while (and nearest-element
                     (not (= nearest-element element)))
          do (ps:try (setf element (ps:chain element parent-node))
                     (:catch (error))))))
```

aartaka:
I'm not sure that errors are the best way to find out the overlapping of elements. Maybe rather compare the rectangles there?

source/changelog.lisp
```diff
@@ -24,94 +24,269 @@ particular revision.")
      (:li "MINOR when we add functionality in a backwards-compatible manner.")
```
aartaka, line 24: Seems like the changelog is severely broken, judging by the diff. Can you revert the changes/rebase over master?

source/mode/hint.lisp
```diff
@@ -13,22 +13,37 @@
...
+    :type boolean
+    :documentation "Whether the hinting prompt buffer is collapsed to the input line.")
+   (style
     (theme:themed-css (theme *browser*)
       (".nyxt-hint"
        :background-color theme:primary
        :color theme:on-primary
        :font-family "monospace,monospace"
-       :padding "0px 0.3em"
+       ;; Test it with ems
```
aartaka, line 26: The problem of hints overlapping with elements is not the one that's solvable by their size. I suggest reverting to the previous version or using no less than `1em`. Otherwise hints are less than the text on the page, which is quite likely to be unreadable.

source/mode/hint.lisp
```diff
@@ -13,22 +13,37 @@
...
+       :font-size "65%"
+       :padding "0.1em 0.3em"
        :border-radius "0.3em"
        :z-index #.(1- (expt 2 31))))
-    :documentation "The style of the boxes, e.g. link hints.")
-   (highlighted-box-style
+    :documentation "The style of the hint overlays.")
+   (selection-style
```
aartaka, line 32: How about actually merging these style slots into one, like on `lispy-search`?

source/mode/hint.lisp
```diff
@@ -13,22 +13,37 @@
...
       (".nyxt-hint.nyxt-highlight-hint"
        :background-color theme:accent
        :color theme:on-accent))
-    :documentation "The style of highlighted boxes, e.g. link hints.")
+    :documentation "The style of selected hints.")
+   (show-hint-scope-p
+    nil
+    :type boolean
+    :documentation "Whether the hinted element should display its visual scope.")
+   (hint-scope-style
```
aartaka, line 42: Again, maybe it's better to merge all the styles into one stylesheet/slot.

source/mode/hint.lisp
```diff
@@ -76,18 +91,24 @@ For instance, to include images:
...
        (ps:chain document head (append-child style-element))
-       (ps:chain style-element sheet (insert-rule box-style 0))
-       (ps:chain style-element sheet (insert-rule highlighted-style 1)))
+       (ps:chain style-element
+                 sheet
+                 (insert-rule (ps:lisp (style (find-submode 'hint-mode))) 0))
+       (ps:chain style-element
+                 sheet
+                 (insert-rule (ps:lisp (selection-style (find-submode 'hint-mode))) 1))
+       (when (ps:lisp (show-hint-scope-p (find-submode 'hint-mode)))
```
aartaka, line 103: I'd say insert it no matter what. One could toggle this slot on runtime, thus enabling the behavior with no style present.

Should be solved by using a unified stylesheet, too.

source/mode/hint.lisp
```diff
@@ -99,17 +120,23 @@ For instance, to include images:
...
-    (dotimes (i (ps:lisp (length nyxt-identifiers)))
-      (let ((element (nyxt/ps:qs document (+ "[nyxt-identifier=\""
-                                             (aref ids i)
-                                             "\"]")))
-            (hint (aref hints i)))
-        (when element
-          (ps:chain element (set-attribute "nyxt-hint" hint))
-          (ps:let ((hint-element (hint-create-element element hint)))
-            (ps:chain fragment (append-child hint-element))))))
+    (dolist (element (nyxt/ps:qsa document "*[nyxt-hintable]"))
```
aartaka, line 124: Isn't `*[whatever]` equivalent to `[whatever]`?

source/mode/hint.lisp
```diff
@@ -99,17 +120,23 @@ For instance, to include images:
...
+      (let* ((i 0)
+             (hint (aref hints i)))
+        (ps:chain element (set-attribute "nyxt-hint" hint))
+        (ps:chain fragment (append-child (create-hint-overlay element hint)))
+        (when (ps:lisp (show-hint-scope-p (find-submode 'hint-mode)))
+          (ps:chain element class-list (add "nyxt-element-hint")))
+        (setf i (1+ i))))
+    ;; this doesn't work because parenscript doesn't support loop properly
```
aartaka, line 132: Why is it here then?

source/mode/hint.lisp
```diff
@@ -131,25 +158,41 @@ For instance, to include images:
            (char-length (ceiling (log length (length alphabet)))))
       (loop for i below length collect (select-from-alphabet i char-length alphabet)))))
 
-(defun add-hints (&key selector)
-  (let* ((dom (document-model (current-buffer)))
-         (hintable-elements (clss:select selector dom))
+(define-parenscript set-hintable-attribute ()
+  (ps:dolist (element (nyxt/ps:qsa document
+                                   (ps:lisp (hints-selector (find-submode 'hint-mode)))))
+      (when (nyxt/ps:element-in-view-port-p element)
```
aartaka, line 164: What about drawing off-screen hints for those beyond-viewport folks like me?

source/mode/hint.lisp
```diff
@@ -131,25 +158,41 @@ For instance, to include images:
...
+        (ps:chain element (set-attribute "nyxt-hintable" "")))))
+
+(define-parenscript remove-hintable-attribute ()
+  (ps:dolist (element (nyxt/ps:qsa document "*[nyxt-hintable]"))
+    (ps:chain element (remove-attribute "nyxt-hintable"))))
+
+(defun add-hints (&key selector (buffer (current-buffer)))
+  (run-thread "add CSS stylesheet" (add-stylesheet))
+  (set-hintable-attribute)
+  ;; does it fail if the changed elements are below the threshold?
```
aartaka, line 174: It shouldn't.

But why not just call `update-document-model`?

source/mode/hint.lisp
```diff
@@ -131,25 +158,41 @@ For instance, to include images:
...
+  (ps:dolist (element (nyxt/ps:qsa document ":not(.nyxt-search-node) > .nyxt-hint"))
+    (ps:chain element (remove)))
+  (when (ps:lisp (show-hint-scope-p (find-submode 'hint-mode)))
+    (ps:dolist (element (nyxt/ps:qsa document ".nyxt-element-hint"))
+      (ps:chain element class-list (remove "nyxt-element-hint")))))
+
+(defun remove-hints (&key (buffer (current-buffer)))
+  (run-thread "remove element hints overlays" (remove-hint-elements))
+  (remove-hintable-attribute)
+  (setf (document-model buffer) (nyxt/dom::named-json-parse (nyxt/dom::get-document-body-json))))
```
aartaka, line 195: Why? It should update itself once it sees how much the page has changed.

source/mode/hint.lisp
```diff
@@ -181,20 +224,31 @@ For instance, to include images:
 (define-class hint-source (prompter:source)
   ((prompter:name "Hints")
    (prompter:selection-actions-enabled-p t)
+   (prompter:filter
+    (if (shrink-prompt-buffer-p (find-submode 'hint-mode))
```
aartaka, line 228: How is prompt-buffer shrinking related to element filtering? It feels off. How about adding yet another toggle instead? We already have `auto-follow-hints-p`...

source/mode/hint.lisp
```diff
@@ -181,20 +224,31 @@ For instance, to include images:
...
+        #'prompter:fuzzy-match))
    (prompter:filter-postprocessor
     (lambda (suggestions source input)
       (declare (ignore source))
       (multiple-value-bind (matching-hints other-hints)
           (sera:partition
            (lambda (element)
-             (str:starts-with-p input (plump:get-attribute element "nyxt-hint") :ignore-case t))
+             (str:starts-with-p input
+                                (plump:get-attribute element "nyxt-hint")
```
aartaka, line 242: Alternative: use `plump:attribute` and fit it on one line :P

source/mode/hint.lisp
```diff
@@ -181,20 +224,31 @@ For instance, to include images:
...
+                                :ignore-case t))
            suggestions
            :key #'prompter:value)
         (append matching-hints other-hints))))
    (prompter:selection-actions
-    (lambda (suggestion)
-      (highlight-selected-hint :element suggestion
-                               :scroll nil)))
+    (unless (shrink-prompt-buffer-p (find-submode 'hint-mode))
```
aartaka, line 248: Why not? Can still be useful, no?

source/mode/hint.lisp
```diff
@@ -212,31 +266,38 @@ For instance, to include images:
...
                              nil)))))
 
-(serapeum:export-always 'query-hints)
+(export-always 'query-hints)
 (defun query-hints (prompt function
                     &key (multi-selection-p t)
-                      (selector (hints-selector (find-submode 'hint-mode))))
+                    ;; Does this penalise performance greatly?
+                    ;; (selector "*[nyxt-hintable]")
+                      (selector (str:concat (str:replace-all
```
aartaka, line 274: Wow, that feels terribly wrong... Can we somehow hide `nyxt-hintable` under the rug? -- it's a magic implementation detail, after all.

source/parenscript-macro.lisp
```diff
@@ -100,18 +100,37 @@
...
 
 (export-always 'element-in-view-port-p)
 (defpsmacro element-in-view-port-p (element)
-  "Is the element in the view port?"
-  `(let* ((rect (chain ,element (get-bounding-client-rect))))
+  "Whether ELEMENT is in viewport."
+  `(let* ((rect (chain ,element (get-bounding-client-rect)))
+          (style (chain window (get-computed-style ,element))))
      (if (and (>= (chain rect top) 0)
+              (<= (chain rect top) (- (chain window inner-height) 4))
```
aartaka, line 115: Why -4?

source/prompt-buffer.lisp
```diff
@@ -163,8 +182,14 @@ See also `hide-prompt-buffer'."
     (calispel:! (prompt-buffer-channel (window prompt-buffer)) prompt-buffer)
     (prompt-render prompt-buffer)
     (setf (ffi-window-prompt-buffer-height (window prompt-buffer))
-          (or height
-              (prompt-buffer-open-height (window prompt-buffer))))
+          (cond ((eq height :default)
```
aartaka, line 185: `case`?

source/prompt-buffer.lisp
```diff
@@ -5,123 +5,142 @@
...
-                       :accessor nil
-                       :export nil
-                       :documentation "By default the prompter library creates a
-new history for each new prompt buffer.  Here we set the history to be shared globally.")
+    ((window
+      nil
+      :type (or null window)
+      :export nil
+      :documentation "The window in which the prompt buffer is showing.")
+     (height
```
aartaka, line 13: This one is useful beyond hinting, actually. How about setting it by default when `:inivisible-input-p` is T? Otherwise the password prompts are ugly... 

CC @Ambrevar?

aartaka (changes requested):
Changelog is malformed and some magic variables are not explained. I am too curious about those to leave them as is :P

source/changelog.lisp
aadcg: What's wrong with the changelog? Did you follow the commits?

source/mode/hint.lisp
aadcg: Because I need to report the bug.

source/mode/hint.lisp
aadcg: Fair enough.

source/parenscript-macro.lisp
aadcg: I'm saying that if an element is partially visible by 4px, then fine. Why exactly 4? Trial and error.

source/changelog.lisp
aadcg: Which magic variables are you referring to?

source/mode/hint.lisp
aartaka: Fair point :D

source/parenscript-macro.lisp
aartaka: That is most probably dependent on the DPI of the scree. Maybe use something like half of the element?

And put a comment there explaining that it's for partial visibility :)

source/changelog.lisp
aartaka: Magic variable of -4 for element visibility, for instance :)

aartaka: Oh, okay, forget the changelog. Now I see that you've reordered it for clarity.

source/mode/hint.lisp
aadcg: I agree. How would you solve it?

source/mode/hint.lisp
aadcg: Agree.

source/mode/hint.lisp
aadcg: Indeed! It was just a reminder that I don't want `*[foo]`, but `tag1[foo], tag2[foo], ...`.

aadcg:
I'm getting a really weird warning from SBCL after commit b8cf9d222. @aartaka do you have any idea on how to fix my macro hack? If I can't come up with smth, I'll do it in the most obvious and less elegant way:

```lisp
(export-always 'define-parenscript-async)
(defmacro define-parenscript-async (script-name args &body script-body)
  "Define parenscript function SCRIPT-NAME.
SCRIPT-BODY must be a valid parenscript and will be wrapped in (PS:PS ...).
Any Lisp expression must be wrapped in (PS:LISP ...).

The returned function sends the compiled Javascript to the current buffer webview.
The function can be passed ARGS."
  `(defun ,script-name ,args
     (ffi-buffer-evaluate-javascript-async (current-buffer)
                                           (ps:ps ,@script-body))))
```

> 
> failed AVER: (NULL SB-C::CURRENT)
> This is probably a bug in SBCL itself. (Alternatively, SBCL
> might have been corrupted by bad user code, e.g. by an undefined
> Lisp operation like (FMAKUNBOUND 'COMPILE), or by stray pointers
> from alien code or from unsafe Lisp code; or there might be a
> bug in the OS or hardware that SBCL is running on.) If it seems
> to be a bug in SBCL itself, the maintainers would like to know
> about it. Bug reports are welcome on the SBCL mailing lists,
> which you can find at <http://sbcl.sourceforge.net/>.
>    [Condition of type SB-INT:BUG]
>

aadcg:
I'll resume this on Tuesday.

aadcg:
> Help me, I'm a parenscript potato.
> 
> Can someone please make the following work according to the comment? @aartaka @jmercouris
> 
> ```lisp
> (define-parenscript overlapped-p (id)
>   "Whether ELEMENT is hidden by another overlapping element."
>   (ps:let* ((element (nyxt/ps:qs-nyxt-id document (ps:lisp id)))
>             (rect (ps:chain element (get-bounding-client-rect)))
>             (nearest-element (ps:chain document (element-from-point (ps:chain rect left)
>                                                                     (ps:chain rect top)))))
>     ;; element is overlapped iff the (ps:chain element parent-node) call returns
>     ;; the following error:
>     ;; TypeError: null is not an object (evaluating 'element.parentNode')
>     (loop while (and nearest-element
>                      (not (= nearest-element element)))
>           do (ps:try (setf element (ps:chain element parent-node))
>                      (:catch (error))))))
> ```

To answer my own question, a simply way of going around is:

```
(define-parenscript overlapped-p (id)
  "Whether ELEMENT is hidden by another overlapping element."
  (ps:let* ((element (nyxt/ps:qs-nyxt-id document (ps:lisp id)))
            (rect (ps:chain element (get-bounding-client-rect)))
            (nearest-element (ps:chain document (element-from-point (ps:chain rect left)
                                                                    (ps:chain rect top)))))
    (loop while (and nearest-element
                     element
                     (not (= nearest-element element)))
          do (setf element (ps:chain element parent-node)))
    (null element)))
```

source/mode/hint.lisp
aadcg: [Reported](https://gitlab.common-lisp.net/parenscript/parenscript/-/issues/10).

source/mode/hint.lisp
aadcg: Because `update-document-model` adds the nyxt identifiers, but those get there when the page is loaded in the first place. I just want to update the `document-model` slot.

Actually my comment about changed elements shows a misunderstanding of mine.

source/mode/hint.lisp
aadcg: It seems it doesn't, according to my tests...

source/mode/hint.lisp
aadcg: I agree that prompt-buffer shrinking is unrelated to element filtering.

But it feels odd to add another toggle. How about setting the filter according to `auto-follow-hints-p`?

source/mode/hint.lisp
aadcg: Good call.

source/mode/hint.lisp
aadcg: When the prompt-buffer is reduced to the input area there's no selection (thus no highlighting).

source/mode/hint.lisp
aadcg: Yes, it's terrible. 

Is this worth it though? How big is the performance penalty between selecting `[nyxt-hintable]` and `hints-selector` such that they have the `nyxt-hintable` attribute?

source/parenscript-macro.lisp
aadcg: > That is most probably dependent on the DPI of the screen.

I don't think so. Each element has a box that is measured in pixels, and that is absolute as soon as it gets rendered in whatever combination of hardware and output device (monitor).

> Maybe use something like half of the element?

That would fail the purpose. A partially visible element is still visible. To achieve that, we any to subtract any non zero integer to `inner-height/width`. I don't see how could that value be conditional. We can set it to 1, for the time being.

source/prompt-buffer.lisp
aadcg: Good call.

source/mode/hint.lisp
aadcg: > How about setting the filter according to auto-follow-hints-p?

My above suggestion is a bad one.

source/mode/hint.lisp
aartaka: But what's the problem of adding identifiers?

source/mode/hint.lisp
aartaka: But wait, the underlying prompter still has a selection, even though it's not shown as listing of suggestions. I'm suggesting to highlight the selected hint for the combination of `auto-return-p nil` and `shrink-prompt-buffer t`.

source/mode/hint.lisp
aartaka: Performance is not my concern here, I'm worried about the fact that one can redefine the `hint-selector` to something like `a | button | img` and it will go wrong in a terrible way. You shall not change the user-configurable selector in any way, because there might be _anything_ there.

aartaka:
@aadcg, see 1db1e33d180e02e783c82534a8c4614c4ea46ae0 for the `:buffer` and `:async` argument to `peval`.

aartaka:
Same for `pflet` in c1ce1c1f6eeca4328f4dd517ac784e5a6353de5e!

aadcg:
@aartaka, I find the implementation a bit odd. Keyword arguments aren't positional, but we force them to be. It feels wrong to me.

Also, let's not forget about #2227.

aartaka:
> @aartaka, I find the implementation a bit odd. Keyword arguments aren't positional, but we force them to be. It feels wrong to me.

Still, it's better of a pattern than any other I thought of:
- Putting async and buffer before the bindigns list in `pflet`.
- Adding another mandatory arglist (that's what StumpWM, `defclass`, and some libraries do).

While this one resembles the pattern of method qualifiers that occupy an arbitrary space after the method name and before the args. So, it's still more or less CL-ish, just a bit special :)

Also note that there's no syntax (except addditional arglist) for this particular problem that would be supported by `destructuring-bind`. We have to stretch the parsing anyway, so why not make the most comfy version of the syntax we could have?

> Also, let's not forget about #2227.

Ah, I should probably close that. Thanks for mentioning!

source/prompt-buffer.lisp
aadcg: Is it possible to set a slot based on the value of another slot?

aartaka: No. The best shot here would be to check `invisible-input-p` in `intialize-instance :after` or `customize-instance :after` or somewhere else it's reliably set.

source/mode/hint.lisp
aadcg: You're correct, but there's a catch to it. That holds as longs as filtering is set to

```lisp
(lambda (suggestion source input)
          (declare (ignore source))
          (str:starts-with-p input
                             (prompter:attributes-default suggestion)
                             :ignore-case t))
```

In other words, if the filtering only cares about the hint strings.

source/mode/hint.lisp
aadcg: I've rewritten this bit, so this doesn't matter anymore.

But I don't see how would that be the case. I was setting the selector for `add-hints` by manipulating the `hints-selector` slot, not by mutating it.

aadcg:
> Also note that there's no syntax (except addditional arglist) for this particular problem that would be supported by destructuring-bind. We have to stretch the parsing anyway, so why not make the most comfy version of the syntax we could have?

I didn't follow you here.

source/prompt-buffer.lisp
aadcg: Could you test 8f836f6d9, since I don't use `password-mode`?

aadcg:
We should merge this in my opinion.

Some things like `prompter:filter` and `prompter:selection-actions` might be a bit polemical, but I still think it's the best we can do now.

Ambrevar:
Why re-order the changelog?  Any benefit?  I prefer adding the the bottom, it creates less noise in diffs.

Re-ordering also mostly breaks the history for the whole file: is it worth it?

Ambrevar:
@aartaka requested a change.

aadcg:
> Why re-order the changelog? Any benefit? I prefer adding the the bottom, it creates less noise in diffs.

It seems odd that we need to go to the bottom of a file to add an entry to the changelog. Regarding diffs, I don't see the difference it makes. Example?

Notice that the changelog was broken with regards to the order:

```
     30:(define-version "2.0.0"
     67:(define-version "2.1.0"
    104:(define-version "2.1.1"
    116:(define-version "2.2.0"
    209:(define-version "2.2.1"
    239:(define-version "2.2.2"
    258:(define-version "2.2.3"
    291:(define-version "2.2.4"
    331:(define-version "3-pre-release-1"
    440:(define-version "3.0.0"
```

If the order of old-to-new would be respected then `3-pre-release-1` would come last.

> Re-ordering also mostly breaks the history for the whole file: is it worth it?

That is not entirely true. You can still follow the history and `git blame`, it just takes an extra step. But it's a fair point.

Still, if I revert this commit, I'll move `3-pre-release-1` to the very bottom.

aadcg:
I don't understand how can I mark as resolved the changes that @aartaka requested, without dismissing his review. Can we use less GitHub paraphernalia? It gets in the way.

aartaka:
Ahhhh.

aartaka (approved):

aartaka:
I feel like lost and not remembering anything about the recent changes in this PR, but one thing I'm sure: changelog is better off staying as it is, because files naturally grown downwards, not upwards (why tho?) :D

Ambrevar:
> If the order of old-to-new would be respected then `3-pre-release-1` would come last.

Artyom did this.  In any case, the pre-release entries can disappear from the final 3.0.0 changelog I think.

aartaka:
> > If the order of old-to-new would be respected then `3-pre-release-1` would come last.

I don't understand the logic of this statement. 3.0.0 comes after 3-pre-release-1 (chronologically), so why say that 3-pre-release-1 comes after 3.0.0?

> Artyom did this. In any case, the pre-release entries can disappear from the final 3.0.0 changelog I think.

Yes, they won't be there in 3.0.0, but it's nice to keep track of separate releases while we're doing them!

aadcg:
> I feel like lost and not remembering anything about the recent changes in this PR, but one thing I'm sure: changelog is better off staying as it is, because files naturally grown downwards, not upwards (why tho?) :D

Files don't naturally grow downwards. That's a psychological barrier we inherited from writing on paper.

I've never seem a changelog that keep old entries at the top, as it makes hardly any sense. See, for instance, [Emacs changelogs](https://git.savannah.gnu.org/cgit/emacs.git/tree/ChangeLog.1), and the change log major mode Emacs provides `(info "(emacs) Change Log")`.

aadcg:
> I don't understand the logic of this statement. 3.0.0 comes after 3-pre-release-1 (chronologically), so why say that 3-pre-release-1 comes after 3.0.0?

Correct. I should have said that it makes little sense to have a changelog section corresponding to a release tag that doesn't exist!

aartaka:
Does it not exist? 0_o

aartaka:
Are we lying to ourselves about the existence of 3-pre-release-1? I don't know what to believe anymore...

aadcg:
> Yes, they won't be there in 3.0.0, but it's nice to keep track of separate releases while we're doing them!

Agree. Any tagged release should have a corresponding changelog section, which I think we consistently do.

aadcg:
>  Are we lying to ourselves about the existence of 3-pre-release-1? I don't know what to believe anymore...

I'm not sure what you're expressing here. 

I was suggesting that commit d4cf43539 might not have been a good move, a view that @Ambrevar seems to share.

source/mode/hint.lisp
```diff
@@ -13,22 +13,29 @@
     nil
     :type boolean
     :documentation "Whether the hints are automatically followed when matching user input.")
-   (box-style
+   (shrink-prompt-buffer-p
```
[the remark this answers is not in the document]
aadcg: Thanks for pointing out!

source/mode/hint.lisp
```diff
@@ -126,34 +132,60 @@ For instance, to include images:
...
+       (loop for i below length
+             collect (select-from-alphabet i
+                                           (max (ceiling (log length (length alphabet)))
+                                                1)
+                                           alphabet))))))
+
+(define-parenscript set-hintable-attribute ()
+  (let ((elements (nyxt/ps:qsa document
+                               (ps:lisp (hints-selector (find-submode 'hint-mode))))))
+    (if (ps:lisp (compute-hints-in-view-port-p (find-submode 'hint-mode)))
```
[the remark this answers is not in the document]
aadcg: Indeed!

source/mode/hint.lisp
```diff
@@ -73,21 +83,19 @@ For instance, to include images:
...
-       (ps:chain style-element sheet (insert-rule highlighted-style 1)))
+       (setf (ps:chain style-element inner-text)
+             (ps:lisp (style (find-submode 'hint-mode)))))
      (:catch (error)))))
 
-(define-parenscript hint-elements (nyxt-identifiers hints)
-  (defun hint-create-element (original-element hint)
-    "Creates a DOM element to be used as a hint"
+(define-parenscript-async hint-elements (hints)
+  (defun create-hint-overlay (original-element hint)
```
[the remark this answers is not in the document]
aadcg: Are you sure this is good call, given that it uses [Document.createElement()](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement)?

source/mode/hint.lisp
```diff
@@ -126,34 +132,60 @@ For instance, to include images:
...
-    (let* ((alphabet (hints-alphabet (find-submode 'hint-mode)))
-           (char-length (ceiling (log length (length alphabet)))))
-      (loop for i below length collect (select-from-alphabet i char-length alphabet)))))
-
-(defun add-hints (&key selector)
-  (let* ((dom (document-model (current-buffer)))
-         (hintable-elements (clss:select selector dom))
+  (let ((alphabet (hints-alphabet (find-submode 'hint-mode))))
+    (cond
+      ((zerop length) (error "Length must be greater than 0"))
```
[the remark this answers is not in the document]
aadcg: Agree! I was looking for smth like `alex:positive-integer`, thanks for pointing it out!

source/mode/hint.lisp
```diff
@@ -126,34 +132,60 @@ For instance, to include images:
...
+      ((= 1 (length alphabet))
```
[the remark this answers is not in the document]
aadcg: Ok, why not!

source/mode/hint.lisp
```diff
@@ -45,6 +52,9 @@ define which elements are picked up by element hinting.
 For instance, to include images:
 
     a, button, input, textarea, details, select, img:not([alt=\"\"])")
+   (compute-hints-in-view-port-p
```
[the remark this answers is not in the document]
aadcg: > Is this set anywhere? How do we use it? Is it meant to be exported?

It's a user option. I'm not sure if I'm following you...

>  As for show-hint-scope-p, we may need an example here.

Indeed. I'll try to come up with a good example by means of writing, since a visual one would be ideal.

Ambrevar:
I mean it's not a very big deal, it's just changelog.  Let's adopt a workflow and stick to it :)

source/mode/hint.lisp
aartaka: You weren't mutating, but you were building something on top of it having assumptions on its contents (namely, that it's a comma-separated list of selectors). But no assumption is true about its contents, except that it's some CSS selector (which is arguable too).

source/mode/hint.lisp
```diff
@@ -99,61 +107,84 @@ For instance, to include images:
...
+             collect (select-from-alphabet i
+                                           (max (ceiling (log length (length alphabet)))
+                                                1)
+                                           alphabet))))))
+
+(define-parenscript set-hintable-attribute ()
+  (let ((elements (nyxt/ps:qsa document
+                               (ps:lisp (hints-selector (find-submode 'hint-mode))))))
+    (if (ps:lisp (compute-hints-in-view-port-p (find-submode 'hint-mode)))
+        (ps:dolist (element elements)
```
aartaka, line 151: Note that you repeat `ps:dolist` twice here. Maybe put `if` inside `ps:dolist` then?

source/mode/hint.lisp
```diff
@@ -99,61 +107,84 @@ For instance, to include images:
...
+          (ps:chain element (set-attribute "nyxt-hintable" ""))))))
+
+(define-parenscript remove-hintable-attribute ()
+  (ps:dolist (element (nyxt/ps:qsa document "[nyxt-hintable]"))
+    (ps:chain element (remove-attribute "nyxt-hintable"))))
+
+(defun add-hints (&key selector (buffer (current-buffer)))
+  (add-stylesheet)
+  (set-hintable-attribute)
+  (setf (document-model buffer) (nyxt/dom::named-json-parse (nyxt/dom::get-document-body-json)))
```
aartaka, line 164: Still, this is too resemblant of `update-document-model`. Is the performance penalty of indexing more elements in `update-document-model` that scary, actually?

source/mode/hint.lisp
```diff
@@ -99,61 +107,84 @@ For instance, to include images:
...
+  (ps:dolist (element (nyxt/ps:qsa document ":not(.nyxt-search-node) > .nyxt-hint"))
+    (ps:chain element (remove)))
+  (when (ps:lisp (show-hint-scope-p (find-submode 'hint-mode)))
+    (ps:dolist (element (nyxt/ps:qsa document ".nyxt-element-hint"))
+      (ps:chain element class-list (remove "nyxt-element-hint")))))
+
+(defun remove-hints (&key (buffer (current-buffer)))
+  (remove-hint-elements)
+  (remove-hintable-attribute)
+  (setf (document-model buffer) (nyxt/dom::named-json-parse (nyxt/dom::get-document-body-json))))
```
aartaka, line 183: Again, `update-document-model` won't hurt.

source/mode/hint.lisp
```diff
@@ -181,20 +212,29 @@ For instance, to include images:
 (define-class hint-source (prompter:source)
   ((prompter:name "Hints")
    (prompter:selection-actions-enabled-p t)
+   (prompter:filter
+    (if (fit-to-prompt-p (find-submode 'hint-mode))
```
aartaka, line 216: Maybe move the check inside the filter and check `auto-return-p` of the `(current-prompt-buffer)`? That would be quirky, but more semantically sound :thinking:

source/mode/hint.lisp
```diff
@@ -212,36 +252,37 @@ For instance, to include images:
                              (nyxt/dom:hover-element (first elements))
                              nil)))))
 
-(serapeum:export-always 'query-hints)
+(export-always 'query-hints)
 (defun query-hints (prompt function
                     &key (multi-selection-p t)
-                      (selector (hints-selector (find-submode 'hint-mode))))
+                         (selector "[nyxt-hintable]"))
```
aartaka, line 258: Why leave it as argument if it's an implementation detail now?

source/mode/hint.lisp
```diff
@@ -382,7 +423,7 @@ FUNCTION is the action to perform on the selected elements."
 (define-command follow-hint ()
   "Prompt for element hints and open them in the current buffer."
   (let ((buffer (current-buffer)))
-    (query-hints "Go to element"
+    (query-hints "Open element"
```
aartaka, line 426: But wait, `follow-hint` does not necessarily open things. Does it open buttons? Does it open `<details>` or `<select>`s? Well, the latter two, maybe, but not the buttons. "Go to element" is not exactly fitting of an action either, though. Maybe "Interact with an element" or something akin to it?

aadcg: Good call.

source/mode/hint.lisp
aadcg: I finally understand your point, and you're right.

But what I realized just now is that this is a huge problem, since the calls to `query-hints` that pass the `selector` keyword argument will not behave as intended.

source/mode/hint.lisp
aadcg: Can a point a situation where it's necessary?

source/mode/hint.lisp
aadcg: That is true... Not sure how to proceed here. [Especially after this observation](https://github.com/atlas-engineer/nyxt/pull/2419#discussion_r950281810).

aadcg: Actually it's trivial to fix this. See the latest iteration.

source/mode/hint.lisp
aadcg: If you mean that I could write it as:

```lisp
(define-parenscript set-hintable-attribute (selector)
  (let ((elements (nyxt/ps:qsa document (ps:lisp selector))))
    (ps:dolist (element elements)
      (if (ps:lisp (compute-hints-in-view-port-p (find-submode 'hint-mode)))
          (when (nyxt/ps:element-in-view-port-p element)
            (ps:chain element (set-attribute "nyxt-hintable" "")))
          (ps:chain element (set-attribute "nyxt-hintable" ""))))))
```

Yes, I could but I would be evaluating `(ps:lisp (compute-hints-in-view-port-p (find-submode 'hint-mode)))` countless time for nothing.

aadcg: I was trying to avoid some calls, but it's probably not worth it.

aartaka: But then, you're not avoiding any, because the value of `compute-hints-in-view-port-p` is baked into the JavaScript the moment you start running it...

source/mode/hint.lisp
aadcg: Is 9fb57f80d what you had in mind?

source/mode/hint.lisp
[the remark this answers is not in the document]
aadcg: Does 16258e072 look better?

source/mode/hint.lisp
aadcg: It doesn't work to call `current-prompt-buffer` in the filter. Most likely because the filter is called when the prompt buffer is open? 

I think aea167a9e will make you happy.

aadcg:
Ready to merge from my side.

source/mode/hint.lisp
aartaka: Yep, aea167a should be enough :)

aartaka:
I'd expand the changelog with the value of each addition, like
```
Add nyxt/hint-mode:fit-to-prompt-p to configure whether the prompt takes the bare minimum of the screen space or not.
Add nyxt/hint-mode:show-hint-scope-p for Konkeror-style element highlighting.
Add height so that every prompt can have a configurable height.
```

But then, `:nxref` kinda shows it to those curious enough. The question is: do we want `:nxref` in changelog? Or are we better off with the plain HTML changelog that can be shown in any browser?
- Staying with plain HTML means we can publish the changelog right away on atlas.engineer. But it makes us think more on what to put in the text, like in my example above.
- Using `:nxref` (and other niceties from Nyxt) makes changelog nicer to use inside Nyxt, but quirky to host at atlas.engineer.

Trade-offs...

To not prolong the review of this PR, I suggest using plain `:code` tags for now. I'll open an issue about this.

aadcg:
All addressed @aartaka. I just left the word "Konkeror" out of the changelog.

Ok to merge?

aartaka:
@aadcg, yes!

aadcg:
Ok! Notice that the tests failed due to [this commit](https://github.com/atlas-engineer/nyxt/commit/38906b3e960c7c531fa90f27c53ee570040bfe1b#commitcomment-81930387).

aartaka:
Nope, tests fail because of d63d692e624ccbde92333d726a785ab8a455b79d :P

And 4c868b3990771866c7dae0fea30fd6d8c84d94aa fixes it on master ;)

aadcg:
Merged via a0f9f77ae.

PR #2230, opened 2022-05-05
Recursive search

aadcg:
This is a follow-up to #1246.

Still some work of investigation ahead, but it's coming as `recursive-search-from-current-buffer` with depth set to 1 is already working. 

Updates soon.

PR #2424, opened 2022-07-01
Lispy search-buffer

aartaka:
# Description

This changes our search functionality to depend on Plump and Lisp-side calculations. 
Pros:
- More controllable and extensible (for example, for #2230 (CC @aadcg) and search-based annotations).
- Stores most data on the Lisp side, thus saves lots of JavaScript calls.
- Adds a neat set of changes to `nyxt/dom` that can be used for libraries and extensions.

Cons:
- Highlighting is less exact, as it highlights the smallest element containing the search text, and this element may not be exactly small.
- May be slower than JS-based search, but, given how fast we made our hints, this is a solvable problem.

This also includes quite some changes to the existing functions that make working with Plump DOM even easier!

# Discussion

- Is the element-level highlighting enough for text search? I feel like it adds more context, but I can foresee the opposing arguments too :)
- Should we reuse the same `nyxt/dom:get-unique-selector` in hint mode? This way, we don't need to alter tha page structure anymore. But then, it may be not unique enough...

# Checklist:

How does it look?

### DIFF
--- a/source/dom.lisp
+++ b/source/dom.lisp
@@ -219,7 +219,8 @@ Return two values:
                                       (str:starts-with-p "data-" attribute)
                                       (str:starts-with-p "nyxt-" attribute)))
                                 (alex:hash-table-keys (plump:attributes element))))
-         (classes (when raw-classes (remove-if #'str:blankp (str:split " " raw-classes))))
+         (classes (when raw-classes (mapcar #'clss:css-escape
+                                            (remove-if #'str:blankp (str:split " " raw-classes)))))
          (parents (parents element))
          (family (plump:family-elements element))
          ;; Is it guaranteed that the topmost ancestor of a node is
@@ -234,7 +235,7 @@ Return two values:
              (selreturn ()
                (return-from get-unique-selector (values selector (unique-p selector)))))
       ;; ID should be globally unique, so we check it first.
-      (when (and identifier (sera:single (clss:select (selconcat :sel "#" identifier)  root)))
+      (when (and identifier (sera:single (clss:select (selconcat :sel "#" (clss:css-escape identifier))  root)))
         (selreturn))
       ;; selconcat hack doesn't look nice here, but should work for cases of
       ;; both empty selector and ID selector.
@@ -247,8 +248,9 @@ Return two values:
               classes))
       (when attributes
         (mapc (lambda (attribute)
-                (when (unique-p (selconcat :sel "[" attribute "=\""
-                                            (plump:attribute element attribute) "\"]"))
+                (when (unique-p (selconcat :sel "[" (clss:css-escape attribute) "='"
+                                           (clss:css-escape (plump:attribute element attribute))
+                                           "']"))
                   (selreturn)))
               attributes))
       ;; Check for short and nice parent-child relations, like :only-child,
--- a/source/dom.lisp
+++ b/source/dom.lisp
@@ -293,6 +293,7 @@ Return two values:
   (defmethod plump:text :around ((node plump:nesting-node))
     (alex:ensure-gethash node text-memo-table (call-next-method))))
 
+(export-always 'find-text)
 (defmethod find-text ((text string) (element plump:nesting-node)
                       &key (test #'search))
   "Find all the matches for the TEXT in ELEMENT and its children.
--- a/source/dom.lisp
+++ b/source/dom.lisp
@@ -187,6 +187,20 @@ JSON should have the format like what `get-document-body-json' produces:
   (:documentation "Get the recursive parents of the NODE.
 The closest parent goes first, the furthest one goes last."))
 
+(export-always 'ordered-select)
+(defmethod ordered-select (selector (root plump:nesting-node))
+  "A re-implementation of `clss:select' with the goal of preserving elements order.
+
+SELECTOR is any selector acceptable to `clss', including the compiled one and
+string one."
+  (let ((matched-nodes '()))
+    (labels ((collect-if-match (element)
+               (when (clss:node-matches-p selector element)
+                 (push element matched-nodes))
+               (map nil #'collect-if-match (plump:child-elements element))))
+      (collect-if-match root)
+      (coerce (nreverse matched-nodes) 'vector))))
+
 (export-always 'get-unique-selector)
 (-> get-unique-selector (plump:element) t)
 (defmemo get-unique-selector (element)
--- a/source/dom.lisp
+++ b/source/dom.lisp
@@ -2,6 +2,8 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (nyxt:define-package :nyxt/dom
+  ;; FIXME: This is because window causes conflicts with buffer.lisp (somewhy?).
+  (:shadow #:window)
   (:documentation "Nyxt-specific DOM classes and functions operating on them.
 
 The classes are generated for every HTML element, including `h1-element',
@@ -202,6 +204,10 @@ string one."
                (map nil #'collect-if-match (plump:child-elements element))))
       (collect-if-match root)
       matched-nodes)))
+
+(export-always 'get-nyxt-id)
+(defmethod get-nyxt-id ((element plump:element))
+  (plump:get-attribute element "nyxt-identifier"))
 
 (export-always 'get-unique-selector)
 (-> get-unique-selector (plump:element) t)
--- a/source/dom.lisp
+++ b/source/dom.lisp
@@ -311,7 +311,11 @@ TEST should be a function of two arguments comparing TEXT with element's
   (flet ((matches (elem)
            (find-text text elem :test test)))
     (when (funcall test text (plump:text element))
-      (or (alex:mappend #'matches (coerce (plump:child-elements element) 'list))
+      (or (remove-if
+           ;; Remove those to not clutter the output.
+           ;; TODO: More tags/attributes?
+           (alex:curry #'clss:node-matches-p "style, script")
+           (alex:mappend #'matches (coerce (plump:child-elements element) 'list)))
           (list element)))))
 
 ;; REVIEW: Export to :nyxt? We are forced to use it with nyxt/dom: prefix.
--- a/source/dom.lisp
+++ b/source/dom.lisp
@@ -188,18 +188,20 @@ JSON should have the format like what `get-document-body-json' produces:
 The closest parent goes first, the furthest one goes last."))
 
 (export-always 'ordered-select)
-(defmethod ordered-select (selector (root plump:nesting-node))
+(defun ordered-select (selector root)
   "A re-implementation of `clss:select' with the goal of preserving elements order.
 
 SELECTOR is any selector acceptable to `clss', including the compiled one and
 string one."
-  (let ((matched-nodes '()))
+  (declare (optimize speed))
+  (let ((matched-nodes (make-array 0 :adjustable t :fill-pointer 0))
+        (selector (clss::ensure-selector selector)))
     (labels ((collect-if-match (element)
                (when (clss:node-matches-p selector element)
-                 (push element matched-nodes))
+                 (vector-push-extend element matched-nodes))
                (map nil #'collect-if-match (plump:child-elements element))))
       (collect-if-match root)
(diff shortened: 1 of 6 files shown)


source/dom.lisp
```diff
@@ -2,6 +2,8 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (nyxt:define-package :nyxt/dom
+  ;; FIXME: This is because window causes conflicts with buffer.lisp (somewhy?).
+  (:shadow #:window)
```
Ambrevar, line 6: That's because 
- this package has a symbol `window`;
- it cl:uses `nyxt`;
- it is loaded before `nyxt:window` is exported.

Does that make sense?

source/dom.lisp
```diff
@@ -170,6 +172,26 @@ JSON should have the format like what `get-document-body-json' produces:
   (:documentation "Get the recursive parents of the NODE.
 The closest parent goes first, the furthest one goes last."))
 
+(export-always 'ordered-select)
+(defun ordered-select (selector root)
+  "A re-implementation of `clss:select' with the goal of preserving elements order.
+
+SELECTOR is any selector acceptable to `clss', including the compiled one and
+string one."
+  (declare (optimize speed))
```
Ambrevar, line 181: If speed is so much of an issue here, consider adding a benchmark.
Consider typing your array, it can really help too!

source/dom.lisp
```diff
@@ -170,6 +172,26 @@ JSON should have the format like what `get-document-body-json' produces:
...
+  (let ((matched-nodes (make-array 0 :adjustable t :fill-pointer 0))
+        (selector (clss::ensure-selector selector)))
+    (labels ((collect-if-match (element)
+               (when (clss:node-matches-p selector element)
+                 (vector-push-extend element matched-nodes))
+               (map nil #'collect-if-match (plump:child-elements element))))
```
Ambrevar, line 187: Style: `mapc` instead of `map nil`?  (Debatable.)

source/dom.lisp
```diff
@@ -193,11 +215,14 @@ Return two values:
          ;; TODO: Remove other attributes, unreliable ones? For example, href
          ;; and type are reliable and are unlikely to change, while data-*
          ;; attributes are unreliable and can change any moment.
-         (attributes (remove-if (rcurry #'member '("class" "id") :test #'string=)
+         (attributes (remove-if (lambda (attribute)
+                                  (or (member attribute '("class" "id") :test #'string=)
```
Ambrevar, line 219: There is `str:s-member` if you like.

source/dom.lisp
```diff
@@ -193,11 +215,14 @@ Return two values:
...
+                                      (str:starts-with-p "data-" attribute)
```
Ambrevar, line 220: Why `data-`?  Maybe this deserves a command or a predicate.

source/dom.lisp
```diff
@@ -272,6 +303,23 @@ Return two values:
   (when (plump:has-attribute img "src")
     (quri:uri (plump:get-attribute img "src"))))
 
+(let ((text-memo-table (tg:make-weak-hash-table :weakness :key)))
+  (defmethod plump:text :around ((node plump:nesting-node))
+    (alex:ensure-gethash node text-memo-table (call-next-method))))
```
Ambrevar, line 308: Nice!

source/dom.lisp
```diff
@@ -272,6 +303,23 @@ Return two values:
...
+
+(export-always 'find-text)
+(defmethod find-text ((text string) (element plump:nesting-node)
+                      &key (test #'search))
+  "Find all the matches for the TEXT in ELEMENT and its children.
+
+TEST should be a function of two arguments comparing TEXT with element's
+`plump:text' and returning a boolean for whether there's a match."
```
Ambrevar, line 316: Typo: no "for" here.

source/mode/document.lisp
```diff
@@ -406,20 +406,19 @@ ID is a buffer `id'."
...
-                   (make-instance 'heading :inner-text (plump:text e)
-                                           :element e
-                                           :buffer buffer
-                                           :keywords (ignore-errors
-                                                      (analysis:extract-keywords
-                                                       (plump:text (plump:next-element e))))
-                                           :scroll-position (heading-scroll-position e)))
-                 (clss:select "h1, h2, h3, h4, h5, h6" (document-model buffer)))
-            #'< :key (compose #'parse-integer #'get-nyxt-id #'element)))))
+      (map 'list
```
Ambrevar, line 412: Wait, that's `mapcar`, isn't it?

source/mode/search-buffer.lisp
```diff
@@ -5,10 +5,32 @@
...
+       :background-color (str:concat theme:primary " !important")
+       :color (str:concat theme:on-primary " !important"))
+      (".nyxt-highlight-search-hint"
+       :background-color (str:concat theme:accent " !important")
+       :color (str:concat theme:on-accent " !important")))
+    :documentation "The style of the search marks, including the highlighted ones.")
+   (test-function
+    (lambda (sub string)
+      (search sub string :test #'equalp))
+    :documentation "The function to match text with the text on the page while searching.
```
Ambrevar, line 27: A bit confusing, maybe "with the body of the page"

source/mode/search-buffer.lisp
```diff
@@ -5,10 +5,32 @@
...
+
+Takes two string arguments: search input and the page element text.
+
+You can redefine it to enable regex-based search, for example:
+\(define-configuration nyxt/search-buffer-mode:search-buffer-mode
+  ((nyxt/search-buffer-mode:test-function #'cl-ppcre:scan)))")
```
Ambrevar, line 33: Nice!

source/mode/search-buffer.lisp
```diff
@@ -140,47 +67,61 @@
...
+                                   (make-instance 'search-match
+                                                  :identifier (nyxt/dom:get-unique-selector element)
+                                                  :element element
+                                                  :body (plump:text element)
+                                                  :buffer buffer))
+                                 elements)))
+    (remove-search-hints)
+    (with-current-buffer buffer
+      (add-stylesheet)
+      (hint-elements (coerce (mapcar #'identifier search-matches) 'vector))
```
Ambrevar, line 123: This `coerce` can be remove if we `(map 'vector ...)`

Ambrevar (commented):
Looking great, thanks a lot!

I'll test next week!

jmercouris:
I've tested it. Searching multiple times on a page led to a crash :-O. Additionally, highlighting the whole element is not particularly helpful! I also noticed that when searching large sections of the page would become blank! I can attach screenshots later if you like.

Ambrevar:
In any case, let's not merge this before `3-pre-release-1`.

source/mode/search-buffer.lisp
aadcg: Just would like to say that the current `search-buffer` on master also supports regex, even though we don't advertise it.

aadcg:
@aartaka exciting work! It's not there yet, but it's promising. 

In terms of the design, I'm not convinced on the idea of highlighting elements. Say we you have multiple matches on one element.

source/dom.lisp
aartaka: Yes, it does!

source/dom.lisp
aartaka: That was a WIP for CLSS contribution, I've just removed it in 684e532ba2b6197079df540ab1ecfbd84bd4efb1 in favor of submodule update.

source/dom.lisp
aartaka: `mapc` only works on lists, while `plump:child-element` returns a vector. `map nil` is the only way I was able to think of there (`loop for child across ...` is too verbose for simple mapping).

source/dom.lisp
aartaka: `data` attributes are reserved for application-specific uses. They may be mutable/unpredictable/contextual and browsers/apps should respect their opaqueness. That's why we don't even look at `data` attributes there.

source/mode/document.lisp
aartaka: Nope—`clss:select` returns a vector, while `mapcar` only works with lists, thus, to map a function over a vector and get a list as a result, our only choice is a lower level `map 'list`.

aartaka:
> In any case, let's not merge this before `3-pre-release-1`.

Sure!

aartaka:
> I've tested it. Searching multiple times on a page led to a crash :-O. Additionally, highlighting the whole element is not particularly helpful! I also noticed that when searching large sections of the page would become blank! I can attach screenshots later if you like.

Huh, that's interesting, I haven't see anything like that when testing 🤔  A reason to test more then :D

aartaka:
OK, I've contributed a more extensive character set to the CLSS, which should mitigate the biggest source of errors and crashes of lispy search that @jmercouris noticed (although I'd enjoy having more detailed reports). Search should work fine now. Things left to do:
- Highlighting both elements and text matches, as per @aadcg's feedback.
- Polish the element counting function to actually update the DOM when things change significantly. Right now it seems to stay outdated for long enough to invalidate the search.

aartaka:
Element counting is fixed on master with e9e85bf1966961f293ea64cdc8078f9c61183f29 :D

aartaka:
Note that to test this properly, you need the most recent CLSS!

aartaka:
While I want this to be merged as an improvement over opaque JS-based search, I am demotivated trying to highlight both elements and exact matches inside them, as per @aadcg's suggestion. Can you try this on some familiar pages searching for something and evaluate whether the element-based highlighting is bearable? @jmercouris, @Ambrevar?

I would commit to making exact highlighting possible... but we all know that once a feature is merged, it's increasingly less probable that incomplete yet merged parts of the feature would ever be completed :P

aartaka:
The CSS escaping problem solved on the CLSS side now,  Lispy search is fully functional. The last thing I'd add is listing the exact matching snippet in the prompt-buffer so that it's clearer from the prompt-buffer alone what the context of the word is.

aartaka:
Text truncation to show the match in the prompt-buffer is there. Will be glad if anyone tests and expresses their opinion on how usable the element-level highlighting is!

aadcg:
My first comment is that I ran into some compilation errors while `(asdf:load-system :nyxt/gtk-application)` on this branch. For example:

```
; caught ERROR:
;   
;     The symbol "UI-DISPLAY" is not external in the NDEBUG package.
;   
;       Line: 27, Column: 28, File-Position: 868
```

aadcg:
The issue reported above was my fault.  I had to pull the latest `ndebug`.

aadcg:
There seem to be issues with `prompter:object-attributes` from `search-buffer`. Recipe: go to potato on wikipedia, and search for potato. @aartaka, could you take a look?

I agree with @Ambrevar that context isn't enough.

> I am demotivated trying to highlight both elements and exact matches inside them

Are you're saying you're tired of working on this, and you'd like me to jump in?

aartaka:
> There seem to be issues with `prompter:object-attributes` from `search-buffer`. Recipe: go to potato on wikipedia, and search for potato. @aartaka, could you take a look?

Fixed with ea3dfbd72391bc318bb5899fe52e7a61ea41e3a3.

> I agree with @Ambrevar that context isn't enough.
> 
> > I am demotivated trying to highlight both elements and exact matches inside them
> 
> Are you're saying you're tired of working on this, and you'd like me to jump in?

I'll be glad if you can try exploring it, yes!

aadcg:
I'll be taking over this PR.

aadcg:
@aartaka, could you explain commit 9935a6e6f?

aartaka:
> @aartaka, could you explain commit [9935a6e](https://github.com/atlas-engineer/nyxt/commit/9935a6e6f43f9167bcbaa9e883fc41b1c5ffa621)?

It basically replaces the `<mark>` node with its first child (which is the meaningful element highlighted by `<mark>`). Note that this design is already supereded by adding styles to the actual elements to highlight in 54f5aabf76291ebb785382729e5800428fda6bb5.

aadcg:
Superseded by #2529. Commits by @aartaka preserved in the new PR.

PR #2432, opened 2022-07-07
Replace our `on' and `once' on macros with nhooks versions.

Ambrevar:
@aartaka 
Looks OK to you?


aartaka:
I mean, the essence of `on` and `once-on` is being as short as possible. Maybe `:import` those?

aartaka:
But yeah, looks fine otherwise :)

Ambrevar:
In my opinion conciseness is first and foremost in number of _words_, not letters :)  But even then, `hooks:on` is still really short, and much shorter than writing what it expands to.

Anyways, if you insist on importing, go ahead.  Also note that the user is free to import to `nyxt-user` anytime :)

PR #2437, opened 2022-07-08
mode/auto: Give user feedback when mode get disabled or enabled.

Ambrevar:
# Description

Simply echo when auto-mode does its job.  Otherwise it can be jarring when a mode gets disabled (invisibly to the user).

# Discussion

I've added the echo to buffer-load-hook.  But this seems too "local" of a hack.

Can we be more generic?
Maybe we could queue the messages and have `echo` display them all with a timer, e.g. 0.5s per message, unless there are too many.  Still not ideal if there are too many messages.

Maybe show a tooltip instead?

Or increase the height of the message buffer to display multiple messages at once?

Or... Have a notification icon in the status area that shows the number of unread warnings (as per `echo-warning`)?

# Checklist:

### DIFF
--- a/source/mode/auto.lisp
+++ b/source/mode/auto.lisp
@@ -162,6 +162,20 @@ restored.")))
                     rules)
                    #'priority :key #'test)))))
 
+(defun auto-enable-modes (modes buffers &optional args)
+  "Like `enable-modes', but with more feedback."
+  (when modes
+    (enable-modes modes buffers args)
+    (hooks:once-on (buffer-loaded-hook (current-buffer)) (buffer)
+      (echo "Auto-mode enabled modes ~{~a~^, ~}." modes))))
+
+(defun auto-disable-modes (modes buffers)
+  "Like `disable-modes', but with more feedback."
+  (when modes
+    (disable-modes modes buffers)
+    (hooks:once-on (buffer-loaded-hook (current-buffer)) (buffer)
+      (echo "Auto-mode disabled modes ~{~a~^, ~}." modes))))
+
 (-> enable-matching-modes (quri:uri buffer) *)
 (defun enable-matching-modes (url buffer)
   (let ((rule (matching-auto-mode-rule url buffer)))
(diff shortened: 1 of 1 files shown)


Ambrevar:
CC @aartaka

aartaka:
I see you like `once-on` :P 

What I'd do is rather reporting those immediately, without waiting for a hook...

Ambrevar:
> What I'd do is rather reporting those immediately, without waiting for a hook...

To clarify, the reason I did this is because if we report immediately, the user won't see anything because the message is immediately overwritten by "https:... loaded".

aartaka:
Yes, makes sense (҂⌣̀_⌣́)

Ambrevar:
@lansingthomas Would you have any suggestion on how to improve the feedback when modes get enabled / disabled on a page load?  The message area is clearly too limited for this purpose.

jmercouris:
Well, actually the fact that things get overwritten in the message area is probably a generic problem with our message area. Probably we should have a queue and each message is displayed for at least 3 seconds or something, with a configurable amount of time.

aadcg:
> the fact that things get overwritten in the message area is probably a generic problem with our message area. Probably we should have a queue and each message is displayed for at least 3 seconds or something, with a configurable amount of time.

I think it's too premature to think about this. What we have right now in place is, in my view, good enough.

I think this PR does improve things considerably since it's important to have a distinctive message   that disambiguates mode toggling by the user or by `auto-mode`.

jmercouris:
Agreed. Just because it is not the perfect solution does not mean it is not a good in-between-step!

lansingthomas:
I would love to see this get merged. I think the echo is better than nothing and there are some other priorities before the modes system gets the attention it deserves. <3

jmercouris (approved):

source/mode/auto.lisp
```diff
@@ -162,6 +162,20 @@ restored.")))
                     rules)
                    #'priority :key #'test)))))
 
+(defun auto-enable-modes (modes buffers &optional args)
```
aartaka, line 165: I feel like this belongs more to the `enable-modes`/`enable-modes*` `:after` method. Less layers for things to go wrong at.

source/mode/auto.lisp
```diff
@@ -162,6 +162,20 @@ restored.")))
...
+  "Like `enable-modes', but with more feedback."
+  (when modes
+    (enable-modes modes buffers args)
+    (hooks:once-on (buffer-loaded-hook (current-buffer)) (buffer)
```
aartaka, line 169: Maybe report this right when modes get enabled? Or is this handler here because modes message is obscured by buffer loading messages?

aartaka (commented):
Some implementation comments. The idea is still good, though.

source/mode/auto.lisp
Ambrevar: Exactly, the handler ensures the message is displayed after other mode messages.  But then it effectively makes mode messages useless since the user can't see them...

lansingthomas:
>@lansingthomas Would you have any suggestion on how to improve the feedback when modes get enabled / disabled on a page load? The message area is clearly too limited for this purpose.

Aaaaaaaw Yeeeaaah

spec for some checkboxes:
https://github.com/atlas-engineer/nyxt/issues/3134

jmercouris:
Please reopen when ready.

Issue #2351, opened 2022-06-09
Extend common-settings to include more


Ambrevar:
Some very useful settings:

- [ ] keybinding style should be a drop-down menu listing all scheme _dynamically_ (so as to include extensions).
- [ ] Same for theme styles.  Need to display the preview dynamically then.
- [ ] Place "Edit user files" at the top.
- [ ] Restore session: always, ask, never.
- [ ] Make default browser. ~~(Need to program it, see https://stackoverflow.com/questions/57667116/how-do-i-set-the-default-browser-for-xdg-open-on-centos-7-if-xdg-settings-has-no).~~
  Code-ready, see #2439.
- [ ] Spell check language (it's a mode, should we have per-mode settings?)
- [ ] download directory.
- [ ] Default modes.
- [ ] proxy (also a mode)
- [ ] "set default new buffer URL" should have suggestions like "nyxt:new", "nyxt:dashboard".
- [ ] Customize search engines.

See also #2296.

aartaka:
Sone more:
- Cookie settings.
- Intelligent Tracking Protection (it's part of `reduce-tracking-mode`, but maybe we should factor it out).
- [Text editing keybindings](http://tiborsimko.org/emacs-bindings-in-gtk-apps.html).
- Font families (should we add it to `theme`?)

aadcg:
@aartaka, it's on my list but I'm still not actively working on it!

aartaka:
> @aartaka, it's on my list but I'm still not actively working on it!

No pressure then :)

aadcg:
@Ambrevar what can we achieve for 3.0 from this list?

Ambrevar:
#2408 I believe.  Help welcome there... :p

aartaka:
I think that this is actually 3.0-critical, because the current screen is almost useless and we need a robust set of configs in the new release \~_\~

aartaka:
From https://github.com/atlas-engineer/nyxt/commit/e5c8920d2a1284ed0538ecc3738abcebcae77e51: allow configuring prompt buffer keyschemes too.

PR #2439, opened 2022-07-08
configuration: Add set-default-browser.

Ambrevar:
# Description

Add a low-level function to set the default browser.

Preparation for #2351.

# Discussion

This leverages the py-configparser library.  Does the job.
Anything else you would recommend otherwise?

# Checklist:

### DIFF
--- a/source/configuration.lisp
+++ b/source/configuration.lisp
@@ -346,6 +346,36 @@ Call this function from your initialization file to re-enable the default ASDF r
           asdf/source-registry:default-system-source-registry))
   (asdf:clear-configuration))
 
+(defun set-as-default-browser (&key (name "nyxt")
+                                 (targets
+                                  (list (uiop:xdg-config-home "mimeapps.list")
+                                        (uiop:xdg-data-home "applications/mimeapps.list"))))
+  "Return the modified MIME apps list.
+Return the persisted file as second value."
+  (let* ((target (or (first (sera:filter #'uiop:file-exists-p targets))
+                     (first targets)))
+         (config (py-configparser:read-files (py-configparser:make-config)
+                                             (list target)))
+         (desktop-file (uiop:strcat name ".desktop")))
+    (dolist (section '("Added Associations" "Default Applications"))
+      (dolist (key '("text/html"
+                     "text/gemini"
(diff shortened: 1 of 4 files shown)


source/configuration.lisp
```diff
@@ -346,6 +346,35 @@ Call this function from your initialization file to re-enable the default ASDF r
...
+                                     (uiop:xdg-data-home "applications/mimeapps.list"))))
+  "Return the modified MIME apps list.
+Return the persisted file as second value."
+  (let* ((target (or (first (sera:filter #'uiop:file-exists-p targets))
+                     (first targets)))
+         (config (py-configparser:read-files (py-configparser:make-config)
+                                             (list target)))
+         (desktop-file (uiop:strcat name ".desktop")))
+    (dolist (section '("Added Associations" "Default Applications"))
+      (dolist (key '("text/html"
```
aartaka, line 361: plus `text/gemini`!

aartaka:
Maybe rather `set-as-default-browser`? The name is confusing, I thought it's about setting some meta-class or something...

source/configuration.lisp
aartaka: No, Gopher has no dedicated text/media MIMEs. I doubt it has any concept of MIMEs at all xD

Ambrevar:
Should be ready to merge.

Ambrevar:
Actually, nope!  This is XDG specific and won't work on macOS or Windows.  We need to add a feature macro here.  Which one should it be?

jmercouris:
> Actually, nope! This is XDG specific and won't work on macOS or Windows. We need to add a feature macro here. Which one should it be?



> Actually, nope! This is XDG specific and won't work on macOS or Windows. We need to add a feature macro here. Which one should it be?

Would it not be #+linux #-darwin or something like that?

Ambrevar:
I think we need a flag that encompasses all distros that run Xorg / Waylang, so the *BSD as well, Solaris, etc. but not macOS.

Maybe `#+(and unix (not darwin)) `?

aartaka:
> Maybe `#+(and unix (not darwin)) `?

Yes, should work.

jmercouris:
Yes, this would work.

Issue #2113, opened 2022-02-17
Further improve the REPL


aartaka:
#2096 set a good ground for our REPL to finally become usable. But there's still a long way to go to make a perfect REPL:
- [ ] `M-n` and `M-p` (as in SLIME/SLY) to re-use results from the history.
- [ ] Introduce Paredit/lispy-like bindings to efficiently edit s-exps.
- [x] Use more of a `define-internal-page-command(-global)` to e.g. generate evaluation history, integrate `lisp-repl` with the internal pages infrastructure.
- [x] Allow inline debugging (via `debug->html`) of expressions in the REPL.
- [x] Inspection and pretty-printing (should be covered by #2099).
  - [x] Based on this, add more `value->html` methods so that e.g., pictures are inserted inline and pathnames turn into (preview-able) file links.
- [x] Suggested by @Ambrevar: turn the REPL into a Jupyter-like multiple-pane environment, making it possible to evaluate/debug/inspect many expressions simultaneously.
- [ ] Multiple processes support (see https://github.com/atlas-engineer/nyxt/issues/2113#issuecomment-1043130100 and https://github.com/atlas-engineer/nyxt/pull/2223#issuecomment-1117040364).
- [x] Allow switching packages independently for every pane (https://github.com/atlas-engineer/nyxt/pull/2445#discussion_r917761557).
- [x] Aborting the computation.
- [ ] Who-calls, who-references, who-uses from SLIME/SLY.
- [ ] Tracing.
- [x] Better inspection.
- [ ] Allow comment/markup cells.
- [ ] Import .lisp files into interaction-ready REPL panes.
- [ ] Export REPL to .lisp files.
- [x] Persist cell contents on change (see https://github.com/atlas-engineer/nyxt/pull/2597#issuecomment-1275784658).
- [x] Follow cells after reload (https://github.com/atlas-engineer/nyxt/pull/2597#issuecomment-1275805119).
- [ ] Automatically add new empty cells after the evaluated ones (https://github.com/atlas-engineer/nyxt/pull/2597#issuecomment-1275806092),
- [x] Make the global buttons ("Add new cell", "Delete all cells") visible no matter the scroll position.

Ambrevar:
> There are pros and cons, and I'm not sure I want to give up on the pros it gives :D

Sure, I'm not saying we should disallow running a REPL on the current
image, simply have an option for both!

> You mean evaluating every statement in a separate thread? We already
> do that, see `evaluate`

[...]

PR #2445, opened 2022-07-11
mode/repl: Enable inline debugging.

aartaka:
# Description

This adds inline debugger to our `lisp-repl`. It's quite simple in its functionality, and is mostly a drop-in replacement, but it raises some architectural questions regarding our internal debugging library.

Additionally, this includes documentation for REPL functionality, both in manual and `nyxt/repl` package.

# Discussion

A debatable change in this PR is the use of `nyxt-user` package for _all_ REPL evaluations. I don't like it, but it seems to be the best way to ensure package uniformity for REPL evaluations, except per-cell package designation (which might be there someday, but not today).

The future of our debugger.lisp library (not even deserving a dedicated package yet) that I envision is:
- We publish it as a separate repository (`ndebug`?).
- We make it a separate ASDF system dependent on `swank` and `calispel`.
  - Maybe even make two systems: `ndebug/swank` and `ndebug/slynk` so that the ones preferring Slynk (like myself) don't have to bear with Swank.
    - Maybe even drop Swank/Slynk once we understand how their function-invoking streams work.
  - Export the `make-debugger-hook` function that would:
    - Wrap the raised condition into its handler with a restart-accepting-channel and other auxiliary information.
    - Create a two-way stream to interact with the app-specific prompting mechanisms (`:query-in-callback` & `:query-out-callback` keyword args).
    - Call the UI function to display the condition and the actions to resolve it (`:ui-display-callback`).
    - Once the channel returns (should we depend on Calispel here? Maybe use some `bt:semaphore`, which seems to be a bit more lightweight), call the dedicated cleanup action (`:ui-cleanup-callback`).
    - And that's all that's required from a debugger hook. Not much, huh?

# Checklist:

### DIFF
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -449,6 +449,26 @@ instance must be non-nil.")
       (:pre (:code "nyxt --profile nosave --remote --load foo.lisp --eval '(foo)'")))
 
      (:section
+      :id "repl"
+      (:h3 "Built-in REPL (" (:nxref :package :nyxt/repl-mode) ")")
+      (:p "Nyxt has a built-in REPL, available with "
+          (:nxref :command 'nyxt/repl-mode:lisp-repl "lisp-repl command") "."
+          "The REPL can be used to try out some code snippets for automation or quickly
+make some Lisp calculations. All the packages Nyxt depends on are available in
+repl with convenient nicknames, and all the code is evaluated in "
+          (:nxref :package :nyxt-user) " package.")
+      (:p "Once the REPL is open, there's only one input cell visible. This cell, always
+present at the bottom of the screen, adds new cells to the multi-pane interface
+of Nyxt REPL. You can input " (:code "(print \"Hello, Nyxt!\")") " and invoke "
+(command-markup 'nyxt/repl-mode:evaluate-cell)
+" to evaluate the cell. A new cell will appear at the top of the buffer, with
+input area containing familiar code, with some " (:code "v332 = \"Hello, Nyxt!\"")
+" variable assignment, and with a verbatim text outputted by your code:")
+      (:pre (:code "Hello, Nyxt!"))
+      (:p "This cell-based code evaluation is the basis of the Nyxt REPL. For more features, see "
+          (:nxref :package :nyxt/repl-mode "REPL mode documenation") "."))
+
+     (:section
       :id "user-scripts"
       (:h3 "User scripts")
       (:p "User scripts are a conventional and lightweight way to run arbitrary JavaScript
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -455,13 +455,12 @@ instance must be non-nil.")
           (:nxref :command 'nyxt/repl-mode:lisp-repl "lisp-repl command") "."
           "The REPL can be used to try out some code snippets for automation or quickly
 make some Lisp calculations. All the packages Nyxt depends on are available in
-repl with convenient nicknames, and all the code is evaluated in "
+REPL with convenient nicknames, and all the code is evaluated in "
           (:nxref :package :nyxt-user) " package.")
       (:p "Once the REPL is open, there's only one input cell visible. This cell, always
 present at the bottom of the screen, adds new cells to the multi-pane interface
-of Nyxt REPL. You can input " (:code "(print \"Hello, Nyxt!\")") " and invoke "
-(command-markup 'nyxt/repl-mode:evaluate-cell)
-" to evaluate the cell. A new cell will appear at the top of the buffer, with
+of Nyxt REPL. You can type in " (:code "(print \"Hello, Nyxt!\")")
+" and press C-return to evaluate the cell. A new cell will appear at the top of the buffer, with
 input area containing familiar code, with some " (:code "v332 = \"Hello, Nyxt!\"")
 " variable assignment, and with a verbatim text outputted by your code:")
       (:pre (:code "Hello, Nyxt!"))
--- a/source/mode/repl.lisp
+++ b/source/mode/repl.lisp
@@ -50,6 +50,7 @@ Supports:
   (run-thread "repl cell evaluation"
     (let ((nyxt::*interactive-p* t)
           (*standard-output* (make-string-output-stream))
+          (*package* (find-package :nyxt-user))
           (*debugger-hook*
             (lambda (condition hook)
               (let* ((*debugger-hook* hook)
--- a/source/mode/repl.lisp
+++ b/source/mode/repl.lisp
@@ -29,7 +29,7 @@
     :documentation "The evaluation is terminated.")
    (raised-condition
     nil
-    :type (maybe condition)
+    :type (maybe nyxt::condition-handler)
     :documentation "The condition that was raised during the `input' execution."))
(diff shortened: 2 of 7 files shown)


source/mode/repl.lisp
```diff
@@ -286,20 +302,39 @@
...
+               collect (let ((restart restart)
+                             (wrapper wrapper)
+                             (evaluation evaluation))
+                         (:button :class "button"
+                                  :onclick (ps:ps (nyxt/ps:lisp-eval
+                                                   (:title "condition")
+                                                   (calispel:! (nyxt::channel wrapper) restart)
+                                                   (setf (raised-condition evaluation) nil)))
+                                  (format nil "[~d] ~a" i (restart-name restart)))))
+         ;; TODO: SLIME and SLY provide introspectable backtraces. How?
```
Ambrevar, line 333: Maybe https://shinmera.github.io/dissect/ can help here.

source/mode/repl.lisp
```diff
@@ -286,20 +302,39 @@
...
+         for name = (if (serapeum:single (results evaluation))
+                        (intern (format nil "V~d" (id evaluation)))
+                        (intern (format nil "V~d.~d" (id evaluation) sub-order)))
+         do (setf (symbol-value name) result)
+         collect (:div
+                  (format nil "~(~a~) = " name)
+                  (:raw
+                   (value->html result (or (typep result 'standard-object)
+                                           (typep result 'structure-object)))))))
+      (t (:span "Calculating...")))))
```
Ambrevar, line 349: This is starting to get quite long...  Maybe split each cond into multiple functions??

source/mode/repl.lisp
```diff
@@ -2,7 +2,18 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (nyxt:define-package :nyxt/repl-mode
-    (:documentation "Mode for programming in Common Lisp."))
+    (:documentation "Mode for programming in Common Lisp.
+
+Has a multi-cell/panel/input environment that evaluates the inputted code on `evaluate-cell'.
```
Ambrevar, line 7: "It has"

source/mode/repl.lisp
```diff
@@ -2,7 +2,18 @@
...
+
+Supports:
```
Ambrevar, line 9: Maybe "Features" instead?

source/mode/repl.lisp
```diff
@@ -2,7 +2,18 @@
...
+- Creating additional cells using the bottom cell.
+- Moving cells with `move-cell-down', `move-cell-up', and dedicated cell UI buttons.
+- Basic tab-completion of the inputted symbols.
+- Multiple evaluation results.
+- Standard output recording.
+- Binding results to the automatically-generated variables.
+- Inline debugging (similar no Nyxt-native debugging with `*debug-on-error*' on.)"))
```
Ambrevar, line 16: Typo: "similar to"

source/mode/repl.lisp
```diff
@@ -50,6 +50,7 @@ Supports:
   (run-thread "repl cell evaluation"
     (let ((nyxt::*interactive-p* t)
           (*standard-output* (make-string-output-stream))
+          (*package* (find-package :nyxt-user))
```
Ambrevar, line 53: Hmmm...  I think SLIME / SLY have a trick for this similar to this: after each evaluation, they store the `*package*` value into their own `package` variable (here we could store it in a `repl-mode` a)  and before each evaluation we set the package to that value.

source/manual.lisp
```diff
@@ -448,6 +448,26 @@ instance must be non-nil.")
...
       (:pre (:code "nyxt --profile nosave --remote --load foo.lisp --eval '(foo)'")))
 
+     (:section
+      :id "repl"
+      (:h3 "Built-in REPL (" (:nxref :package :nyxt/repl-mode) ")")
+      (:p "Nyxt has a built-in REPL, available with "
+          (:nxref :command 'nyxt/repl-mode:lisp-repl "lisp-repl command") "."
+          "The REPL can be used to try out some code snippets for automation or quickly
+make some Lisp calculations. All the packages Nyxt depends on are available in
+repl with convenient nicknames, and all the code is evaluated in "
```
Ambrevar, line 458: "REPL"

source/manual.lisp
```diff
@@ -448,6 +448,26 @@ instance must be non-nil.")
...
+          (:nxref :package :nyxt-user) " package.")
+      (:p "Once the REPL is open, there's only one input cell visible. This cell, always
+present at the bottom of the screen, adds new cells to the multi-pane interface
+of Nyxt REPL. You can input " (:code "(print \"Hello, Nyxt!\")") " and invoke "
```
Ambrevar, line 462: This sound too technical, I think the user does not need to know about the implementation details.

Ambrevar:
Agreed with all points in the discussion!  And yes, a universal `ndebug` library would be awesome!  Let's ensure it does not exist yet though...

For the forced package issue, see the suggestion remotely inspired from SLIME / SLY.

source/mode/repl.lisp
aartaka: That doesn't feel like enough for a multi-pane environment. I suggest we stick to the `nyxt-user`-only panes for now, and add the package switching later. The idea is good, though, I've just referenced your comment in #2113 :)

Ambrevar: What is a "pane"?

In any case, look at what SLY is doing because I believe it solves this exact problem.

aartaka:
> Agreed with all points in the discussion! And yes, a universal `ndebug` library would be awesome! Let's ensure it does not exist yet though...

There are:
- https://github.com/phoe/trivial-custom-debugger : A set of macros to ease setting/binding debugger.
- https://github.com/hyotang666/endaira/blob/master/debugger/debug-repl.lisp : A strong yet opinionated debugger, with no utils to steal.
- https://github.com/mtstickney/trivial-debug-console : A library following the same purpose (allowing to make debugging GUI-based), yet restricted to Windows CLI/GUI apps.

So no, nothing exactly matching, at least on GitHub, CLiki, or common-lisp.net :)

Useful bits:
- https://github.com/gwkkwg/trivial-backtrace instead of `uiop:print-backtrace` or even `dissect` :D

source/mode/repl.lisp
aartaka: Pane = cell = input area.

Yes, SLY should be a good source of inspiration there. But later :P

aartaka:
Merging then?

Ambrevar:
Wait, I'd like to give it a try, will do as soon as possible!

aartaka:
Sure, take your time :)

Ambrevar:
Just tested: it's fantastic, thanks a lot for this amazing feature!

I tested with this:

```lisp
(progn
  (log:info "here")
  (cerror "Continue me" "This is an dummy error")
  (log:info "there")
  17)
```

Feel free to merge!

aartaka:
Yes, CL condition system is a gem that enables this fancy thing :D

One thing that breaks this, though, is `abort` restart -- it prevents the debugger markup from being deleted (even though it become useless and disappears on page reload). This I'll fix after merging, though.

PR #2451, opened 2022-07-12
configuration: Have define-configuration add just one handler per slot.

Ambrevar:
# Description

Currently calling `define-configuration` multiple times over the same slot has no effect.
This patch ensures there is only one handler per slot.

# Discussion

- [ ] The `add-hook` of nhooks should probably do this handler replacement for us.
- [ ] Why are handler named uniquely?  Shouldn't they be named after the slot, actually? 

# Checklist:

EDIT: Added more discussion.

### DIFF
--- a/source/configuration.lisp
+++ b/source/configuration.lisp
@@ -260,35 +260,43 @@ Return NIL if not a class form."
   `(progn
      ,@(loop
          for class in (uiop:ensure-list classes)
-         for handler-name = (gensym "DEFINE-CONFIGURATION")
-         collect
-         `(hooks:add-hook
-           (slot-value (find-class (quote ,class)) 'nyxt::customize-hook)
-           (make-instance
-            'hooks:handler
-            :fn (lambda (object)
-                  (declare (ignorable object))
-                  ,@(loop for ((slot value)) on (first slots-and-values)
-                          when (find slot (mopu:slot-names class))
-                            collect `(setf (slot-value object (quote ,slot))
-                                           (let* ((%slot-value% (slot-value object (quote ,slot)))
-                                                  (%slot-default%
-                                                    ,(if (c2mop:class-finalized-p (find-class class))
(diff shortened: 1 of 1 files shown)


Ambrevar:
@aartaka OK with you?

aadcg:
> Why are handler named uniquely? Shouldn't they be named after the slot, actually?

I'd say they should indeed.

If there's a single handler per slot, how could the user change the value of a slot at runtime? 

I don't remember now, but don't we generate new object when using `define-configuration`?

Ambrevar:
> I don't remember now, but don't we generate new objects when using `define-configuration`?

True in the `2-series` but no longer the case since our user-classes have hooks :)  See `user-classes.lisp`.

aartaka:
No, I don't agree!

The change with `%slot-value%` opened for a new workflow: composing several handlers to change the value several times. I consider this a step towards a more intuitive model of configuration. If one puts several configuration forms for the same slot in their config -- they want all of those, together, modifying the value one by one!

aartaka:
What about this?
``` lisp
(setf slot-foo (append 17 %slot-value%))
(setf slot-foo (append 18 %slot-value%))
```

This use-case is enabled by the current implementation, and I don't want to let it go!

Ambrevar:
Oh right...

Then I suggest one change: to have one handler per slot and to name the handler after the slot: this way it makes hook introspection and editing much easier.

aartaka:
"one handler per slot"?

aartaka (approved):

Ambrevar:
You mean you agree with my last proposal, but not this pull request as it is, right?

aartaka:
Yes, I don't agree with deleting the same slot handlers, but I agree with splitting slots into their own handlers!

So the course I'd be content with is removing the `:place` and `:value` from the generated handler and not deleting the handler with matching place.

Ambrevar:
Superseded by #2480.

PR #2463, opened 2022-07-15
Simplify all prompts.

aadcg:
# Description

Cosmetic changes.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
  - [ ] and Guix dependencies.
- [ ] My code follows the style guidelines for Common Lisp code
- [ ] My changes generate no new warnings


libraries/prompter/tests/tests.lisp
```diff
@@ -235,15 +232,14 @@
 
 (prove:subtest "Raw source"
   (with-report-dangling-threads
-    (let ((prompter (prompter:make
-                     :sources (list (make-instance 'prompter:raw-source)))))
+    (let ((prompter (prompter:make :sources (make-instance 'prompter:raw-source))))
```
aartaka, line 235: Even more of an optimization: use a mere source class symbol (`'prompter:raw-source`) and `prompter` will instantiate it too!

source/mode/bookmark.lisp
```diff
@@ -256,14 +256,10 @@ In particular, we ignore the protocol (e.g. HTTP or HTTPS does not matter)."
...
 (define-command bookmark-url
-    (&key (url (ignore-errors
-                (quri:uri
-                 (first
-                  (prompt
-                   :prompt "Bookmark URL"
-                   :sources (list
-                             (make-instance 'prompter:raw-source
-                                            :name "New URL"))))))))
+    (&key (url (ignore-errors (quri:uri (prompt1
```
aartaka, line 259: I'd put `handler-case`/`handler-bind` here so that we only return `nil` on `nyxt-prompt-buffer-calcelled` and rely on debugger/muffling otherwise.

source/mode/file-manager.lisp
```diff
@@ -287,11 +286,10 @@ directory name) as parameter.
 
 `file-source' also has `supported-media-types'. You can append new types to
 it. Every type in `supported-media-types' will be opened directly in Nyxt."
-  (prompt
-   :extra-modes '(file-manager-mode)
-   :input (uiop:native-namestring default-directory)
-   :prompt "Open file"
-   :sources (list (make-instance 'open-file-source))))
+  (prompt :extra-modes '(file-manager-mode)
```
aartaka, line 289: If you add an `uiop:ensure-list` to the `customize-instance :after` of `prompt-buffer`, then you can provide this as a sole symbol too!

source/mode/os-package-manager.lisp
```diff
@@ -379,32 +372,26 @@ OBJECTS can be a list of packages, a generation, etc."
...
-                                                  :include-manager-p t))
-                    :prompt "Target profile")))
-         (generation (first
-                      (prompt
-                       :sources (list (make-instance 'os-generation-source
-                                                     :profile profile))
-                       :prompt "Switch to generation"))))
+  (let* ((profile (prompt1 :sources (make-instance 'os-profile-source
+                                                   :include-manager-p t)
+                           :prompt "Target profile"))
```
aartaka, line 377: Maybe reorder the `prompt` arguments so the `:sources` go last and don't produce this code gap :)

libraries/prompter/tests/tests.lisp
aadcg: Right! See e5e8566c.

source/mode/file-manager.lisp
aadcg: It's actually not working for me...

See c0573d55.

source/mode/os-package-manager.lisp
aadcg: In general, is makes more sense for `:prompt` to be the first keyword as well.

source/mode/file-manager.lisp
aartaka: Weird 0_o Dunno what to do here...

Ambrevar: Sorry, what's not working exactly?

aartaka: @Ambrevar, @aadcg meant that calling a prompt with `:extra-modes` causes cryptic type errors in `enable-modes`. The reason (see d20802280759c7c3c6e08d55f8be3a021972b4e7 for an explanation, fix, and an architectural question) was that `finalize-buffer` and `customize-instance :after ((prompt-buffer prompt-buffer) ...)` were processing `extra-modes` in incompatible ways, causing errors.

aadcg: Thanks for fixing it @aartaka.

aadcg:
Thanks @aartaka.

PR #2469, opened 2022-07-21
Default modes as a slot

### LINKED ISSUE #2453: Best practice to configure default modes
What's special about modes is that they can only be enabled once the buffer is ready to display, so after `customize-instance :after`.

So we can't enable them in `customize-instance` from the config file.  Try with `no-image-mode`, it will fail to enable.

So what can we do?  One option is to only instantiate the mode without enabling it, like so:

```lisp
(defmethod customize-instance ((buffer web-buffer) &key)
  (call-next-method)
  (alex:appendf (modes buffer)
                (mapcar (lambda (mode) (make-instance mode :buffer buffer))
                        '(nyxt/auto-mode:auto-mode
                          nyxt/no-image-mode:no-image-mode
                          nyxt/force-https-mode:force-https-mode
                          nyxt/no-image-mode:no-image-mode
                          nyxt/no-script-mode:no-script-mode
                          nyxt/proxy-mode:proxy-mode
                          nyxt/reduce-tracking-mode:reduce-tracking-mode))))
```

But it's not very obvious and yet quite verbose.  Can we do better?

We should at least document this.

aartaka:
# Description

This makes `default-modes` to finally become a slot and to enable to `%slot-value%`-based configuration for it, thus fixing the most cryptic yet fundamental part of Nyxt configuration mechanism -- the fact that `default-modes` can only be configured once per class (see https://github.com/aartaka/nyxt-config/issues/2, )

Fixes #2453.

# Discussion

I've not yet included the `closer-mop:slot-definition-readers` change, as all of our `define-configuration` seems to rely on `slot-value` and I'm not yet sure about what design could be equivalent in functionality. Moreover, if we set `slot-value` to the result of the custom reader for, e.g., `default-modes`, then we'd set the value of the `default-modes` to the result of `append` methods, thus invalidating these methods! Not sure how to untangle this Gordian knot...

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
  - [X] I have added new changelog.lisp entries matching the added features.
- [X] Compilation and tests:

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -441,7 +441,7 @@ the buffer (which gives us more flexibility)."))
   (:method :around ((buffer buffer))
     "Remove the duplicates from the `default-modes'."
     (remove-duplicates (call-next-method)
-                       ;; Mode at the beginning of the list have higher priorities.
+                       ;; Modes at the beginning of the list have higher priority.
                        :from-end t))
   (:method append ((buffer context-buffer))
     (list
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -19,7 +19,7 @@ This is useful when there is no current buffer.")
 
 (define-class buffer (renderer-buffer)
   ((default-modes
-    '(base-mode)
+    %default-modes
     :accessor nil
     :type list
     :documentation "The symbols of the modes to instantiate on buffer creation.
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -13,6 +13,10 @@
   ()
   (:metaclass interface-class))
 
+(defvar %default-modes '(base-mode)
+  "The default modes for unspecialized buffers.
+This is useful when there is no current buffer.")
+
 (define-class buffer (renderer-buffer)
   ((default-modes
     '(base-mode)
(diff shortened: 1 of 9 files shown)


source/buffer.lisp
```diff
@@ -416,46 +429,37 @@ the buffer (which gives us more flexibility)."))
...
+  ;; bound errors?
+  (:method append ((buffer t))
+    %default-modes)
   (:method append ((buffer buffer))
-    '())
+    (slot-value buffer 'default-modes))
+  (:method :around ((buffer buffer))
+    "Remove the duplicates from the `default-modes'."
+    (remove-duplicates (call-next-method)
+                       ;; Mode at the beginning of the list have higher priorities.
```
Ambrevar, line 444: Typos: "Modes... higher priority"

Ambrevar:
I'll think about the knot...

Ambrevar:
So what's the answer to the knot problem?

aartaka:
My answer for now is sticking to `slot-value`s and being consistent with that. Because readers and writers (which are usually wrappers around `slot-value` somehow changing it for better/more complete/proper representation), are not meant for slot initialization (which `define-configuration` kinda is about), they are meant for knowledge domain-aware slot usage, which comes later than `define-configuration`.

Ambrevar:
Makes sense!

PR #2465, opened 2022-07-15
Port tests

aadcg:
# Description

A very incomplete attempt at porting the test to the new chosen library. In the process, I'm getting acquainted with `lisp-unit2`, which will be helpful regardless. 

Will work on this on Monday too, and get as far as possible. I'm afraid you'll be on your own after that @Ambrevar.

# Discussion

# Checklist:
(to be ticked by @Ambrevar)
Everything in this checklist is required for each PR.  Please do not approve a PR that does not have all of these items.  

- [x] I have pulled from master before submitting this PR
- [x] I've added the new dependencies as:
  - [x] ASDF dependencies,
  - [x] Git submodules,
  - [x] and Guix dependencies.
- [x] My code follows the style guidelines for Common Lisp code
  - See [Norvig & Pitman's Tutorial on Good Lisp Programming Style (PDF)](https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf) and [Google Common Lisp Style Guide](https://google.github.io/styleguide/lispguide.xml).
- [x] I have performed a self-review of my own code
- [x] My code has been reviewed by at least one peer (the peer review to approve a PR counts.  The reviewer must download and test the code)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] There are no merge conflicts

PR #2479, opened 2022-07-26
conditions: Always raise nyxt-prompt-buffer-non-interactively condition in REPL

### LINKED ISSUE #2470: Polish the interactive/non-interactive distinction handling
In 92a7bf24c44124f36937f627e376fbcf62a0527c, there appeared an `*interactive-p*` flag, which caused me many countless hours of debugging with no error reports and clues to why the prompt doesn't appear and the code silently fails. The reasoning behind the `*interactive-p*`—separating the calls stemming from user interaction from those that are scripted—does make sense, but the implementation interferes with lots of workflows, thus making me put a magical `(let ((*interactive-p* t)) ...)` everywhere to make things work (see 62b7dc93f2fa5b0daf34eb5084f02cbaa15399d2, 7ba4b99c229314eecc85e89f1f8ba398257529e4, 1303731feee3dba33ad0e890bf1d9ee1267bb3ee, and possibly other commits that don't mention interactivity in the commit message).

Given the number of opaque and hard-to-debug problems that `*interactive-p*` caused us, is it actually worth retaining, especially in an incomplete state it is in now? By "incomplete state" I mean that right now it simply cuts off all the prompts, while a proper non-interactive use would allow providing some default value or code snippet to replace/return from the prompt. Like... `prompt-buffer-ready-hook`, for example. It's already there, it's perfectly hookable programmatically, and is friendly towards both interactive and non-interactive workflows. 

I don't see what `*interactive-p*` adds to Nyxt, except the necessity to put magic let-bindings everywhere. We have a way to delineate interactive (i.e. user-facing) interactions from the non-interactive (i.e. headless and script) ones. We have ways (`prompt-buffer-ready-hook`, `prompter:all-ready-p` etc.) to handle prompting, and these ways are reliable enough to rely on them in the non-interactive context.

An alternative to removing `*interactive-p*` would be to allow default values, something-instead-of-prompt computation, or any other way to meaningfully replace prompt calls killed by `*interactive-p*`. But still, I don't see how the value `*interactive-p*` brings is worth adding a ton of new code to `prompter` and `prompt-buffer`.

I may be wrong with this rant, though. If there's some part of a picture that I don't see, please explain it to me!

CC @Ambrevar, @jmercouris, @aadcg

Ambrevar:
I don't know why I decided to inherit from nyxt-prompt-buffer-non-interactively,
I probably got confused when drafting this.

Possibly fixes #2470.

# Discussion

- [x] Is there a semantic relationship between `nyxt-prompt-buffer-canceled` and `nyxt-prompt-buffer-non-interactively`?

- [x] Is this enough of a fix for #2470 ?

- [x] Does this deserve a test?

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [x] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/command.lisp
+++ b/source/command.lisp
@@ -357,7 +357,7 @@ With MODE-SYMBOLS and GLOBAL-P, include global commands."
   (with-current-buffer (current-buffer)
     (let ((*interactive-p* t))
       (handler-case (apply #'funcall command args)
-        (nyxt-prompt-buffer-canceled ()
+        (prompt-buffer-canceled ()
           (log:debug "Prompt buffer interrupted")
           nil)))))
 
--- a/source/concurrency.lisp
+++ b/source/concurrency.lisp
@@ -26,7 +26,7 @@ raised condition."
   (alex:with-gensyms (c sub-c)
     `(if (or *run-from-repl-p* *debug-on-error*)
          (handler-case (progn ,@body)
-           (nyxt-prompt-buffer-canceled ()
+           (prompt-buffer-canceled ()
              (log:debug "Prompt buffer interrupted")))
          (ignore-errors
           (handler-bind
--- a/source/conditions.lisp
+++ b/source/conditions.lisp
@@ -3,17 +3,18 @@
 
 (in-package :nyxt)
 
-(export-always 'nyxt-condition)
-(define-condition nyxt-condition (error)
-  ((message :initarg :message :accessor nyxt-condition-message))
+(export-always 'nyxt-error)
+(define-condition nyxt-error (error)
+  ((message :initarg :message :accessor message))
   (:report (lambda (c stream)
              (format stream "~a" (slot-value c 'message))))
-  (:documentation "An error internal to Nyxt. It should abort the ongoing command, but not the whole process."))
+  (:documentation "An error internal to Nyxt.
+It should abort the ongoing command, but not the whole process."))
 
-(define-condition nyxt-web-context-condition (nyxt-condition)
+(define-condition web-context-error (nyxt-error)
   ((context :initarg :context :reader context)))
 
-(define-condition nyxt-prompt-buffer-canceled (error)
+(define-condition prompt-buffer-canceled (error)
   ())
 (define-condition prompt-buffer-non-interactive (error)
   ((name :initarg :name :accessor name))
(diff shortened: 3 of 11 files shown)


Ambrevar:
@aartaka ?

Ambrevar:
I think I remember why I made it inherit from the other condition: it's because we want the condition to be caught if we are not running from the REPL.

But since we are already catching `error`, then everything is caught anyways.  I guess I was confused back then.

aartaka:
> * [ ]  Is there a semantic relationship between `nyxt-prompt-buffer-canceled` and `nyxt-prompt-buffer-non-interactively`?

No, prompt-buffer being non-interactive is not necessarily prompt cancelling.

> * [ ]  Is this enough of a fix for [Polish the interactive/non-interactive distinction handling #2470](https://github.com/atlas-engineer/nyxt/issues/2470) ?

Yes. If the condition falls through to top level reliably -- it's enough of a fix.

> * [ ]  Does this deserve a test?

Not sure how to test this, given that condition handling differs from place to place in our code.

aartaka:
Actually, why this weird name -- `-non-interactively`? Maybe use something less striking, like `-non-interactive`?

Ambrevar:
Good catch!

About the tests: let's add them once #2465 is merged.

Ambrevar:
I've updated the condition name.
Feel free to merge if you're happy!

Ambrevar:
OK to merge?

aartaka:
Wait, a last-minute thing: why do we prefix these with `nyxt-`? They are already interned to `:nyxt` and thus don't need a name prefix, as they have a package one...

Ambrevar:
Good point!  What about these conditions:

- `nyxt-condition`: it's actually used only once, as a parent of `nyxt-web-context-condition`.  We can rename it `nyxt-error`, as I think keeping the prefix makes sense here.
- `nyxt-web-context-condition`: `web-context-error` seems good here.

aartaka:
>     * `nyxt-condition`: it's actually used only once, as a parent of `nyxt-web-context-condition`.  We can rename it `nyxt-error`, as I think keeping the prefix makes sense here.

Sure, this one being prefixed I don't have anything against!

>     * `nyxt-web-context-condition`: `web-context-error` seems good here.

Yes!

Ambrevar:
OK to merge?

aartaka:
Yes!

Ambrevar:
I was just about to do it!  Thanks!

PR #2480, opened 2022-07-27
configuration: Make one handler for slot-configuration in define-conf…

Ambrevar:
# Description

This helps with debugging and live-manipulation of the hook.

Supersedes #2451 

# Discussion

Do we still need the `(c2mop:class-finalized-p (find-class class))` check?
In which case can this code run and the class not be finalized?

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [x] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/nyxt.asd
+++ b/nyxt.asd
@@ -228,6 +228,7 @@
   :targets (:package :nyxt/tests)
   :serial t
   :components ((:file "tests/package")
+               (:file "tests/offline/define-configuration")
                (:file "tests/offline/global-history")
                (:file "tests/offline/user-script-parsing")
                (:file "tests/online/urls")))
--- a/source/configuration.lisp
+++ b/source/configuration.lisp
@@ -283,38 +283,41 @@ Example to get the `blocker-mode' command to use a new default hostlists:
 
 To discover the default value of a slot or all slots of a class, use the
 `describe-slot' or `describe-class' commands respectively."
-  `(progn
-     ,@(loop
-         for class in (uiop:ensure-list classes)
-         for handler-name = (gensym "DEFINE-CONFIGURATION")
-         collect
-         `(hooks:add-hook
-           (slot-value (find-class (quote ,class)) 'nyxt::customize-hook)
-           (make-instance
-            'hooks:handler
-            :fn (lambda (object)
-                  (declare (ignorable object))
-                  ,@(loop for ((slot value)) on (first slots-and-values)
-                          when (find slot (mopu:slot-names class))
-                            collect `(setf (slot-value object (quote ,slot))
-                                           (let* ((%slot-value% (slot-value object (quote ,slot)))
-                                                  (%slot-default%
-                                                    ,(if (c2mop:class-finalized-p (find-class class))
-                                                         (getf (mopu:slot-properties class slot) :initform)
-                                                         (progn
-                                                           (echo-warning
-                                                            "Slot default not found for slot ~a of class ~a, falling back to its current value"
-                                                            slot class)
-                                                           '%slot-value%))))
-                                             (declare (ignorable %slot-value% %slot-default%))
-                                             ,value))
-                          else
-                            collect `(handler-bind ((warning #'muffle-warning))
-                                       (defmethod ,slot :around ((object ,class))
-                                         (let* ((%slot-value% (call-next-method))
-                                                (%slot-default% %slot-value%))
-                                           ,value)))))
-            :name (quote ,handler-name))))))
+  (alex:with-gensyms (handler hook)
+    `(progn
+       ,@(loop
+           for class in (uiop:ensure-list classes)
+           append (loop for ((slot value)) on (first slots-and-values)
+                        ;; TODO: Shall we really make the name unique?  Since we
+                        ;; are configuring slots, maybe not.
+                        for handler-name = (gensym (format nil "CONFIGURE-~a" slot))
+                        when (find slot (mopu:slot-names class))
+                          collect
+                        `(let ((,hook (slot-value (find-class (quote ,class)) 'nyxt::customize-hook))
+                               (,handler (make-instance
+                                          'hooks:handler
+                                          :fn (lambda (object)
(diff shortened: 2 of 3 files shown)


Ambrevar:
@aartaka Can you test?
(I'm a bit swamped these days... :) )

source/configuration.lisp
```diff
@@ -255,6 +255,29 @@ Return NIL if not a class form."
 (defvar %slot-default% nil
   "Holds the default value of the slot being configured when in `define-configuration'.")
 
+(defun make-slot-configuration-handler (handler-name class slot value)
```
aartaka, line 258: That won't work, as `value` is an arbitrary expression implying the presence of `%slot-default%` and `%slot-value%` at the very least.

Turn it into a macro and everything should work just fine :)

Ambrevar: We could make it a lambda of 2 values, `%slot-default%` and `%slot-value`, no?

aartaka: Yes, that could work.

Ambrevar:
In the end I decided to revert it to a full macro, it's shorter and simpler.

I've added tests.

OK to merge then?

nyxt.asd
```diff
@@ -228,9 +228,11 @@
   :targets (:package :nyxt/tests)
   :serial t
   :components ((:file "tests/package")
-               (:file "tests/offline/global-history")
-               (:file "tests/offline/user-script-parsing")
-               (:file "tests/online/urls")))
+               (:file "tests/offline/define-configuration")
+               ;; (:file "tests/offline/global-history")
```
aartaka, line 232: Why are those commented out?

source/configuration.lisp
```diff
@@ -283,38 +283,43 @@ Example to get the `blocker-mode' command to use a new default hostlists:
...
+                                                                      (echo-warning
+                                                                       "Slot default not found for slot ~a of class ~a, falling back to its current value"
+                                                                       slot class)
+                                                                      '%slot-value%))))
+                                                        (declare (ignorable %slot-value% %slot-default%))
+                                                        ,value)))
+                                          :place (quote ,slot)
+                                          :value (quote ,value)
+                                          :name (quote ,handler-name)))
+                               (,hook (slot-value (find-class (quote ,class)) 'nyxt::customize-hook)))
```
aartaka, line 315: Aesthetic note: maybe move the hook binding above the handler one, so that all the `let` bindings are grouped and one does not have to scan the indentation to understand whether they are still in `let`?

source/configuration.lisp
```diff
@@ -283,38 +283,43 @@ Example to get the `blocker-mode' command to use a new default hostlists:
...
+                                                               ,(if (c2mop:class-finalized-p (find-class class))
+                                                                    (getf (mopu:slot-properties class slot) :initform)
+                                                                    (progn
+                                                                      (echo-warning
+                                                                       "Slot default not found for slot ~a of class ~a, falling back to its current value"
+                                                                       slot class)
+                                                                      '%slot-value%))))
+                                                        (declare (ignorable %slot-value% %slot-default%))
+                                                        ,value)))
+                                          :place (quote ,slot)
```
aartaka, line 312: Is that necessary?

source/configuration.lisp
```diff
@@ -283,38 +283,43 @@ Example to get the `blocker-mode' command to use a new default hostlists:
...
+                                          :value (quote ,value)
```
aartaka, line 313: This is prone to breaking, as `value` is an implicit `progn`.

tests/offline/define-configuration.lisp
```diff
@@ -0,0 +1,52 @@
...
+                                                           :fn (lambda () (print 'dummy1))
+                                                           :name 'dummy1)))))
+  (nyxt:define-configuration nyxt:browser
+    ((nyxt:before-exit-hook (hooks:add-hook nyxt:%slot-value%
+                                            (make-instance 'nhooks:handler
+                                                           :fn (lambda () (print 'dummy2))
+                                                           :name 'dummy2)))))
+  (let ((browser (make-instance 'browser)))
+    (assert-eql 2
+                (length (hooks:handlers (nyxt:before-exit-hook browser))))))
```
aartaka, line 52: Here a better test could be loading some nonsense URL and altering it in two of the `request-resource-hook` handlers. But then, it's hard to track without a renderer.

aartaka (approved):

nyxt.asd
Ambrevar: Oops! :p

source/configuration.lisp
Ambrevar: Damn, another thing I forgot!

tests/offline/define-configuration.lisp
Ambrevar: Why would it be better with regard to `define-configuration`?

Ambrevar:
All addressed except the test suggestion.

tests/offline/define-configuration.lisp
aartaka: It would allow tracking the `%slot-value%`-based composed config. But I guess this we can write tests for as a later time :P

Ambrevar: This test is only interested in `%slot-value%` giving us access to.  Any hook can do.
What do you mean with "tracking the `%slot-value%`-based composed config"?

To be fair, a more proper test would be on a list value, so that we ensure that`%slot-value%` refers to the latest value each time, and not just the first value.

Ambrevar:
Merged with 65129754c8ff6b954f6fd9c726feec1e5c9bc3ef.

tests/offline/define-configuration.lisp
Ambrevar: I've merged, but feel free to open an issue about this.

aartaka:
Cool, thanks!

PR #2481, opened 2022-07-29
Lisp url args

### LINKED ISSUE #2397: Allow lisp:// URL to receive arguments again
As discussed in https://github.com/atlas-engineer/nyxt/pull/2377#issuecomment-1157741439, `lisp://` URLs no longer take arguments and this can be problematic in some cases:

- In delete-bookmark, we have to generate 1 closure per bookmark which is not ideal performance-wise.
- What if we get some data from JS, how do we communicate it to the `lisp://` URL closure?

It's easy to extend the `lisp` scheme to accept argument (for instance as URL query arguments, just like `nyxt://` URLs).

Then the question that remains is how we refer to the lambda-list.

Currently we have this syntax:

```lisp
(ps:ps (nyxt/ps:lisp-eval
        (:title "describe-any")
        (nyxt::describe-any (princ-to-string symbol))))
```

I guess we could have this instead?

```lisp
(ps:ps (nyxt/ps:lisp-eval (&key key-arg1 key-arg2)
        (:title "describe-any")
        (log:info key-arg1 key-arg2)
        (nyxt::describe-any (princ-to-string symbol))))
```

But then we need to change all the `lisp-eval` calls.

Thoughts? @aartaka

aartaka:
# Description

This adds the ability to provide arbitrary keyword arguments to `lisp:` URLs and parse those in the callbacks. In addition to that there's also:
- `lisp-url` to construct `lisp:`-schemed URLs.
- Ability to only generate a callback once, while using it several times (`:id` argument to both `lisp-eval` and `lisp-url`).

Fixes #2397.

# Discussion

- This patch requires the caller to specify arguments twice: in the `nyxt/ps:lisp-eval` ergument list, as keywords, and in the body of `lisp-eval` (and `lisp-url`), as lambda with `(&key arg1 arg2)`. This is duplicative, but I haven't found a reasonable way to generate lambda lists from the keyword lists, given that the `*package*` is almost always bound to `nyxt-user` and thus does not allow to generate the package-relevant argument names. 
  - For example: using `href` in the body of `lisp-eval` in `list-bookmarks` refers to the `nyxt/list-bookmarks-mode::href`, while `(intern (symbol-name :HREF) *package*)` in `lisp-eval` results in `nyxt-user:href`, breaking the whole thing with symbol undefined error.
- `lisp-url` generates URLs with the question mark at the end, to ease argument concatenation on the JS-side. Should we do that, though? Given that we concatenate `title` to it, having a guarantee of URL ending at either `?` or `&` is a virtue, but it doesn't look right.
- This makes syntax of `lisp:` URLs slightly backwards-incompatible, but they were not saved to history anyway, so it shouldn't break many workflows.

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:
    - Maybe add some `lisp-url` generation tests?

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -417,30 +417,108 @@ guarantee of the same result."
                 (apply (form internal-page) args)))))))
   :local-p t)
 
-(ps:defpsmacro nyxt/ps::lisp-eval ((&key (buffer '(nyxt:current-buffer)) title callback) &body form)
+(-> lisp-url (&rest t &key
+                    (:id string)
+                    (:buffer t) ;; `document-buffer', actually.
+                    (:title (maybe string))
+                    (:callback (or function symbol)))
+    string)
+(defun lisp-url (&key title (id (princ-to-string (nyxt:new-id))) (buffer (current-buffer))
+                   (callback (unless (gethash id (nyxt::lisp-url-callbacks buffer))
+                               (alex:required-argument 'callback))))
+  (unless buffer
+    (error "Cannot make a `lisp-url' without BUFFER or current-buffer."))
+  (sera:synchronized ((nyxt::lisp-url-callbacks buffer))
+    (unless (gethash id (nyxt::lisp-url-callbacks buffer))
+      (log:debug "Registering callback ~a for buffer ~a" id buffer)
+      (setf (gethash id (nyxt::lisp-url-callbacks buffer)) callback)))
+  (format nil "lisp://~a?~:[~*~;title=~a&~]" id
+          title (when title
+                  (quri:url-encode title))))
+
+
+(ps:defpsmacro nyxt/ps::lisp-call (id &key title (buffer (current-buffer)) args)
+  "Call the ID-bound function on the Lisp side.
+
+Return a JS Promise fulfilled after the code runs on the Lisp side.
+
+ID should be an identifier of an already defined `lisp-url'.
+
+The ARGS are used as a keyword arglist for the function bound to the defined URL callback."
+  `(fetch (ps:lisp (str:concat
+                    (lisp-url :id ,id :buffer ,buffer :title ,title)
+                    (quri:url-encode-params
+                     (list ,@(loop for (name value . rest) on args by #'cddr
+                                   collect `(cons (symbol->param-name ,name)
+                                                  (value->param-value ,value)))))))
+          (ps:create :mode "no-cors")))
+(export-always 'nyxt/ps::lisp-call :nyxt/ps)
+
+(export-always 'lisp-url-flet)
+(defmacro lisp-url-flet (buffer ((name (&rest args) &body binding-body) &rest other-bindings)
+                         &body body)
+  "Bind the NAMEs of bindings to both
+- the `nyxt/ps:lisp-call'-able IDs, and
+- macros expanding to the Parenscript form calling them.
+
+Bindings are only available in local scope, and are only defined for BUFFER.
+
+Example:
+\(nyxt::lisp-url-flet (nyxt:current-buffer)
+    ((hello (&key name) (nyxt:echo \"Hello, ~a!\" name)))
+  (nyxt:ffi-buffer-evaluate-javascript (nyxt:current-buffer) (hello :name \"Stranger\"))
+  ;; or equivalent
+  (nyxt:ffi-buffer-evaluate-javascript
+   (nyxt:current-buffer)
+   (ps:ps (nyxt/ps:lisp-call hello :buffer (nyxt:current-buffer) :args '(:name \"Stranger\"))))))"
+  (let ((id (princ-to-string (new-id))))
+    `(let ((,name (progn (lisp-url :id ,id
+                                   :buffer ,buffer
+                                   :callback (lambda (,@args) ,@binding-body))
+                         ,id)))
+       (macrolet ((,name (&rest args)
+                    ;; Everything that uses commas comes from the outside,
+                    ;; everything with list/cons/quote belongs to this
+                    ;; macro. Otherwise it's really hard to keep track of.
+                    (cons (quote ps:ps)
+                          (list (quote nyxt/ps:lisp-call) ,id :buffer ,buffer :args ,args))))
+         ,@(if other-bindings
+               `((lisp-url-flet ,buffer (,@other-bindings)
+                   ,@body))
+               body)))))
+
+(ps:defpsmacro nyxt/ps::lisp-eval ((&key (buffer '(nyxt:current-buffer))
+                                         title callback args)
+                                   &body body)
   "Request the lisp: URL and invoke CALLBACK when there's a successful result.
+
+TITLE is purely informative.
+
 BUFFER must be a `document-buffer'.
-TITLE is purely informative."
+
+The ARGS are used as a keyword arglist for the CALLBACK."
   ;; We define it here and not in parenscript-macro because we need
   ;; `nyxt::lisp-url-callbacks' while parenscript-macro is Nyxt-independent.
-  `(let ((url (ps:lisp (let ((request-id (string (gensym ""))))
-                         (unless ,buffer
-                           (error "Cannot `nyxt/ps:lisp-eval' without BUFFER or current-buffer."))
-                         (sera:synchronized ((nyxt::lisp-url-callbacks ,buffer))
-                           (setf (gethash request-id (nyxt::lisp-url-callbacks ,buffer))
-                                 (lambda () ,@form)))
-                         (let ((url-string (format nil "lisp://~a" request-id)))
-                           (when ,title
-                             (setf url-string (str:concat url-string "/" ,title)))
-                           url-string)))))
-     (let ((request (fetch url
-                           (ps:create :mode "no-cors"))))
-       (when ,callback
-         (chain request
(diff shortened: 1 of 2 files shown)


source/urls.lisp
```diff
@@ -399,31 +399,69 @@ guarantee of the same result."
...
+(ps:defpsmacro nyxt/ps::lisp-eval ((&rest args &key (buffer '(nyxt:current-buffer)) title id callback
+                                          &allow-other-keys)
+                                   &body body)
   "Request the lisp: URL and invoke CALLBACK when there's a successful result.
+
+ID is either caller-provided or auto-generated.
+- If it's caller-provided, this ID/CALLBACK is reusable with different
+  ARGS.
+- If it's auto-generated, it's not accessible to the caller and thus not
+  reusable.
```
Ambrevar, line 432: Can you explain why the user would want access to the ID?  Maybe with an example?

source/urls.lisp
```diff
@@ -399,31 +399,69 @@ guarantee of the same result."
...
-     (let ((request (fetch url
-                           (ps:create :mode "no-cors"))))
-       (when ,callback
-         (chain request
-                (then (lambda (response)
-                        (when (@ response ok)
-                          (chain response (json)))))
-                (then ,callback))))))
+  (let ((irregular-args (alexandria:remove-from-plist
+                         args :id :buffer :callback :title)))
```
Ambrevar, line 442: This is awkward.  What about making `args` a required argument or a keyword argument, available as a sublist?

Example: `(:buffer BUFFER1 :args '(:foo 17 :bar 18))`

source/urls.lisp
```diff
@@ -399,31 +399,69 @@ guarantee of the same result."
...
+    `(let ((url (ps:lisp (let ((request-id (or ,id (string (gensym "")))))
+                           (str:concat
+                            (lisp-url
+                             :id request-id
+                             :buffer ,buffer
+                             :callback ,(if (and (sera:single body)
+                                                 (member (first (first body)) '(lambda function)))
+                                            (first body)
+                                            `(lambda ()
+                                               ,@body))
```
Ambrevar, line 452: Style: Maybe lambda on just one line?

Ambrevar:
About the duplicated arguments, what about using arbitrary keyword argument names?
```lisp
(defun foo (&key ((:href my-href)))
   (log:info my-href))
```
Alternatively, you could generate the lambda-list, no?

source/urls.lisp
aartaka: Reusing the same callback several times. For example, you can pre-make and re-use the callback, as in
``` lisp
(defvar id (nyxt:new-id))
;; Create a callback
(nyxt/ps:lisp-eval
 (:title "delete bookmark" :id id)
 (lambda (&key href)
   (nyxt/bookmark-mode:delete-bookmark href)))
;; Use it
(nyxt/ps:lisp-eval
 (:id id :href "https://github.com/"))
(nyxt/ps:lisp-eval
 (:id id :href "https://gitlab.com/"))
```

Ambrevar: If there is no other use case, then maybe replace the ID param with a boolean, like `reuse-p`, and have it remember / reuse the same ID automatically.

Currently this `:id` param feels too low level and leaves room for too many mistakes.

aartaka: What can help there is separation of `define` and `funcall` stages for `lisp-eval`. Right now the mental distance JS world is located at makes us use this `define-and-funcall-sometime` construction that this design is inspired by :)

source/urls.lisp
Ambrevar: You can parse arbitrary arglists with `destructuring-bind`.

aartaka: > You can parse arbitrary arglists with `destructuring-bind`.

Fair commentary, yes. Still, I don't see much hurt in inlining the arguments there. Especially if we talk of separating the "definition" and the "funcall" stages of the `lisp-eval`.

Ambrevar:
Can you give a full example of how to reproduce this issue?

Internal page macros don't seem to be setting the package, do they?

source/urls.lisp
Ambrevar: Not much hurt, at least for now, but it's more future-proof.

source/urls.lisp
Ambrevar: Sorry, can you rephrase?  Maybe give an example?

aartaka:
> Can you give a full example of how to reproduce this issue?

Basically parse any keyword inside the internal page body, and you'll get a `nyxt-user`-interned symbol.

> Internal page macros don't seem to be setting the package, do they?

Internal pages don't, but start.lisp does:
https://github.com/atlas-engineer/nyxt/blob/540022904ff82915dc3006a92d8588c77e0f843c/source/start.lisp#L642

aartaka:
@Ambrevar, Ive just pushed a refactoring of our `lisp-url` handling. See ddfb8fe9df1ba6c20d12a58be36d3592314653c9 for details, but the highlight is that we now have separation of `lisp-url` definition and call, and a convenient `lisp-url-flet` macro to save us dozens of lines!

source/urls.lisp
```diff
@@ -400,31 +400,118 @@ guarantee of the same result."
                     (t (error "Cannot display evaluation result"))))))))))
   :local-p t)
 
-(ps:defpsmacro nyxt/ps::lisp-eval ((&key (buffer '(nyxt:current-buffer)) title callback) &body form)
+(export-always 'lisp-url)
```
Ambrevar, line 403: Shouldn't it be unexported if `lisp-url-flet` and `lisp-call` are the high level API points?

aartaka: Hmmmmmm... I'm not sure. I can't imagine a solid use-case for it that `lisp-url-flet` an `lisp-call` woudn't cover, but I certainly am no omniscient, so...

I'll remove it.

aartaka:
I'm still not sure about the benefits of making arguments a dedicated argument, especially given the fact that we're already in the foggy waters of Parenscript compile-time JS generation. I don't want to add any more quasi-quote madness to it (like, do you need to quote args? Do you want to provided a quoted list or an unquoted one? Do you want to pre-evaluate thing? Too much non-obvious questions there.)

Ambrevar:
Anything blocking here?

source/urls.lisp
Ambrevar: Friendly ping :)

Ambrevar:
@aartaka Have you tested the callbacks?

aartaka:
> @aartaka Have you tested the callbacks?

No, I haven't, because I'm still implying they don't work.

aartaka:
> Anything blocking here?

Nothing but deciding whether we want the `:args '(:arg1 4 :arg2 ,(whatever))` or `:arg1 4 :arg2 (whatever)` syntax :)

Ambrevar:
Let's wrap this up after 3-pre-release-2.

aartaka:
I've updated this changeset to have `:args` as the separate argument to `lisp-(call|eval)`, but haven't tested it yet.

aartaka:
It works and is ready for review!

source/urls.lisp
```diff
@@ -417,30 +417,112 @@ guarantee of the same result."
...
+      (setf (gethash id (nyxt::lisp-url-callbacks buffer)) callback)))
+  (format nil "lisp://~a?~:[~*~;title=~a&~]" id
+          title (when title
+                  (quri:url-encode title))))
+
+
+(ps:defpsmacro nyxt/ps::lisp-call (id &key title (buffer (current-buffer)) args)
+  "Call the ID-bound function on the Lisp side.
+
+Return a Promise succeeding after the code runs on the Lisp side.
```
Ambrevar, line 443: "promise fulfilled after"

Ambrevar: It's a JS promise, right?

source/urls.lisp
```diff
@@ -417,30 +417,112 @@ guarantee of the same result."
...
+    ((hello (&key name) (nyxt:echo \"Hello, ~a!\" name)))
+  (nyxt:ffi-buffer-evaluate-javascript (nyxt:current-buffer) (hello :name \"Stranger\"))
+  ;; or equivalent
+  (nyxt:ffi-buffer-evaluate-javascript
+   (nyxt:current-buffer)
+   (ps:ps (nyxt/ps:lisp-call hello :buffer (nyxt:current-buffer) :args '(:name \"Stranger\"))))))
+
+BEWARE: While `lisp-url's defined by this macro stay defined for a little while
+longer than this macro lasts, it's not recommended to use them outside the
+scope."
```
Ambrevar, line 477: Is it worth mentioning this at all?

Ambrevar (commented):
Looks good to me!

aartaka:
Fixed documentation and rebased against master. Merging?

aadcg:
I believe what you have pushed is the same set of commits.

aartaka:
Yes, now it should be two commits. Although I have to test whether I missed anything during rebase...

aadcg:
Thanks @aartaka!

PR #2490, opened 2022-08-02
Incorporate auto-mode into modable-buffer.

aartaka:
# Description

This moves all the functionality of the `auto-mode` into `modable-buffer`, which results in cleaner code and more reliable mode- and page- specific mode settings. What's new:
- `define-auto-rule` as a simple way to define arbitrary rules without touching the file with user rules.
- `auto-mode` renamed to `auto-rules` as a feature.
- Rule `test`s are now being evaluated, which means we can now put arbitrary lambdas inside the rules!
- Default rules for `gopher`, `gemini`, and all the internal pages that require it.
- `apply-all-matching-auto-rules-p` that allows applying all the matching rules instead of one (requested long ago by @hendursaga)!
- Minor fixes for `no-procrastinate-mode`, and others.
- I've reshuffled the file dependencies, and auto-rules.lisp (the renamed mode/auto.lisp file) now has access to buffer types and history, improving the typing and semantics of functions there!

What's removed:
- Numerous `enable-modes` doing the manual work.
- Lots of `rememberable-p`

Fixes #2004
Fixes #2066 
Fixes #962 

# Discussion

1. It feels suspicious that I had to remove so many `rememberable-p` from the interface-related modes. The only modes that are left with `(rememeberable-p nil)` are the real one-shot modes. Maybe rename it to `one-shot-p` then?
2. Some of the slots from `auto-mode` feel weird in `modable-buffer`. Maybe there's a better place for the slots of `auto-mode`? I was unable to think of any...
3. I am now suspicious of any `enable-modes`, including the one in `reduce-bandwidths-mode`. Should we rather inherit the mode from the modes it enables, so that it's cleaner?
4. Should there be non-configurable default modes? Should `gopher`, `gemini` and internal page modes be redefinable from the `auto-rules.lisp`?
5. Should `enable` perform the work of `make-instance` automatically, or should it always be something more up to the one defining the method, like in `repeat-mode`?

# Checklist:
Everything in this checklist is required for each PR.  Please do not approve a PR that does not have all of these items.

- [X] I have pulled from master before submitting this PR
- [x] There are no merge conflicts.
- [X] I've added the new dependencies as:
  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    git submodule add https://gitlab.common-lisp.net/nyxt/py-configparser _build/py-configparser
    ```
  - [X] and Guix dependencies.
- [X] My code follows the style guidelines for Common Lisp code. See:
[...]

PR #2536, opened 2022-09-02
Remembrance (local search)

Ambrevar:
# Description

New mode which caches the _textual content_ of visited pages, then allow for searching them, displaying them, etc.

We currently store the following:
- URL
- Title
- Textual content
- Keywords.

# Discussion

- [x] Do you like the name "remembrance"?

- [x] Any better suggestion to name the "cached pages?"

- [ ] Better name for `look-up-cache`?  Maybe be more consistent with `search-buffer`?  On the other hand, we don't want to confuse the user.

# How to test

- Enable `remembrance-mode`.
- Browse a few URLs.
- Call `recollect-visited-page` and type in a search query.  Press Alt-Return and run the `view-cached-content` action.

# To do

- [x] Add option to automatically cache all visited pages.
- [x] Garbage collect cache after some time.
- [x] Include keywords in cache?  They come for free after all.
- [x] Allow manual adding of buffers, bookmarks, history entries.  
- [x] Support for diff-ing the current page with the cached page.  What's the state of diff-mode?  @aadcg ?

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:


Ambrevar:
Question: Am I using internal pages correctly here?
Note that I'm enabling `remembrance-mode` when creating the internal page, which does not seem right.  On the other hand, we need to access the cache somehow.  Any suggestion?  @aadcg @aartaka ?

Also about style: is it possible to not specify the mode style and have `(style mode)` do the right thing by fetching the default style?

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,236 @@
...
+    (make-instance 'cache-path)
+    :type cache-path
+    :documentation "Cache path where page content is saved.
+See also `cache'.")
+   (cache
+    nil
+    :type (maybe montezuma:index)
+    :export nil
+    :documentation "`cache' is separate from `cache-path' because it is managed by `montezuma'.")
+   (update-interval
```
aartaka, line 61: The logic of update and discard intervals appears more than once already, and it's always tightly coupled with files. Maybe abstract it to a separate mixin in `nfiles` to not write expiry checks over and over?

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,236 @@
...
+                                   (lookup input (find-submode 'remembrance-mode)
+                                           :suffix-matching-p (suffix-matching-p source))))
+   (prompter:multi-selection-p t)
+   (prompter:active-attributes-keys '("URL" "Title"))
+   (suffix-matching-p
+    t
+    :type boolean
+    :documentation "Whether to match suffixes in queries.
+For instance 'nana' will match 'banana'.
+This induces a performance cost."))
```
aartaka, line 191: How noticeable of a cost?

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,236 @@
...
+  (:export-class-name-p t))
+
+(defmethod prompter:object-attributes ((doc montezuma:document) (source remembrance-source))
+  (declare (ignore source))
+  `(("URL" ,(page-url-string doc))
+    ("Title" ,(page-title doc))))
+
+(define-internal-page view-cached-page (&key url-string)
```
aartaka, line 199: Maybe rather make it a scheme and enable the mode in scheme calback? This way it's quite controllable and isolated. Especially once we merge #2490.

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,236 @@
...
+   ;; (auto-update-p
+   ;;  t
+   ;;  :type boolean
+   ;;  :documentation "Automatically refresh the cache when entry is older than `update-interval'.")
+   (discard-interval
+    #.(* 30 60 60 24)
+    :type alex:non-negative-integer
+    :documentation "Discard entry if its `last-update' is older than this.")
+   ;; TODO: `style' belongs to a separate mode, doesn't it?  Same question as with `history-tree-mode'.
+   (style (theme:themed-css (theme *browser*)
```
aartaka, line 76: I see, so `rememberance-mode` is used for both cached content pages and the regular browsing commands. I guess that's indeed too fuzzy.

But then, `history-tree-mode` is a nice and clean separated mode, no? And I believe that it works its purpose just fine.

Alternative solution for both `history-tree-mode` and `rememberance-mode`: allow locating styles _inside internal pages and schemes_, instead of their modes. How crazy does that sound?

source/mode/remembrance.lisp
Ambrevar: Indeed, but the problem is that Montezuma does all the file handling and leaves us no control over it.

Montezuma looks like a dead-end in the long run.  Let's use it for now as a proof-of-concept, but a better full-text-search system if highly needed.  Maybe time for cl-xapian?

source/mode/remembrance.lisp
Ambrevar: I had run some benchmarks for Demeter for forget to write them down.  It's a big one (like 1 or 2 orders of magnitude), that's all I remember.

source/mode/remembrance.lisp
Ambrevar: But is it worth it?  It's just to view some textual content after all.

source/mode/remembrance.lisp
Ambrevar: I like this last approach!

source/mode/remembrance.lisp
aartaka: I mean, this feels like a scheme rather than a page already. Much like `reader:` or `pdf:` schemes for certain reading experiences/interfaces.

Ambrevar:
Another critical question: should this really be a separate mode, or rather a history option?
After all, this is tightly linked to the pages we visit.

aartaka:
> Another critical question: should this really be a separate mode, or rather a history option?
> After all, this is tightly linked to the pages we visit.

I'd say keep it a mode, because:
- It's an opt-in feature.
- It stores some non-trivial state.
- The cache is an opaque structure, not sure we want to store it in history or around it.
- It's also about certain pages/interfaces, not only about history movement.

Ambrevar:
> * It's also about certain pages/interfaces, not only about history movement.

What do you mean?  Example?

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,236 @@
...
+    :type alex:non-negative-integer
+    ;; TODO: 0 to disable?
+    :documentation "Update entry if its `last-update' is older than this value.
+Set to 0 to disable.")
+   ;; (auto-update-p
+   ;;  t
+   ;;  :type boolean
+   ;;  :documentation "Automatically refresh the cache when entry is older than `update-interval'.")
+   (discard-interval
+    #.(* 30 60 60 24)
```
aadcg, line 72: What does the "#." prefix do?

aadcg:
I'm not being able to test the functionality. When I call `look-up-cache` I get:

```
  #<STANDARD-GENERIC-FUNCTION NYXT:URL (19)>
when called with arguments
  (#<MONTEZUMA:DOCUMENT last-update content title url {101666F2A3}>)
```

aadcg:
> Do you like the name "remembrance"?

I do!

> Any better suggestion to name the "cached pages?"

From the technical point of view, or from the user point of view? For the former I think we're ok. For the latter, perhaps "rememberable pages"

>   Better name for look-up-cache? Maybe be more consistent with search-buffer? On the other hand, we don't want to confuse the user.

I believe you want to say "lookup", not "look up".

source/mode/remembrance.lisp
aadcg: I agree with @aartaka here.

aadcg:
> Another critical question: should this really be a separate mode, or rather a history option?
After all, this is tightly linked to the pages we visit.

A mode, as @aartaka pointed out.  Perhaps there could be a user option that would automatically remember buffers based on a list of functions that would take a buffer and output a boolean.

Ambrevar:
@aadcg Just pushed some fixes, should be working now.

Ambrevar:
> > A mode, as @aartaka pointed out. Perhaps there could be a user option that would automatically remember buffers based on a list of functions that would take a buffer and output a boolean.
> 
> This idea of a buffer function feels like `auto-mode`/`auto-rules`...

Indeed, we want to be able to toggle the mode based on rules, like domain matching rules.

aartaka:
> @aartaka I've included an important fix for Nfile profiles (crazy I forgot to implement profile switching!). Before this, tests were not working properly because the "test" profile was not really used.

We've been using it wrong all this time 0_o

> I've also added 2 helpers, `with-headless` and `wait-on-handler`. The last one is super useful to save countless handler boilerplate snippets. It also paves the way for more linear headless test writing.
> 
> I propose we add it to nhooks, what do you think?

Yes, why not :) Seems like we have to put all the fancy recent hook macros into a separate file already...

Ambrevar:
> Yes, why not :) Seems like we have to put all the fancy recent hook macros into a separate file already...

Where?

aadcg:
I'm still not able to test it.  I started Nyxt without a config file.

aartaka:
> > Yes, why not :) Seems like we have to put all the fancy recent hook macros into a separate file already...
> 
> Where?

I mean, into a dedicated file in `nhooks` repo :)

Ambrevar:
@aadcg Should work now.

aadcg:
@Ambrevar I'm not getting the following error when enabling `remembrance-mode`:

```
When attempting to read the slot's value (slot-value), the slot
NYXT::FN is missing from the object
#<NYXT:COMMAND NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE (1)>.
```

aadcg:
Am I right that in order to search over cached files you need to firstly open them and them search?

Ambrevar:
I can't reproduce @aadcg.  Can you share a precise recipe?

> Am I right that in order to search over cached files you need to firstly open them and them search?

Just follow the original post instructions: enable the mode, load some URLs, then call `recollect-visited-page`.

Ambrevar:
New feature: the query terms are now highlighted in the content view!

Ambrevar:
It's now mostly feature complete (minus the diff support), ready for review.

I'll complete the tests tomorrow.

source/configuration.lisp
```diff
@@ -334,6 +334,10 @@ To discover the default value of a slot or all slots of a class, use the
              (log:debug "No active window, picking last active buffer.")
              (last-active-buffer)))))
 
+(defmethod initialize-instance :after ((file nyxt-file) &key)
+  (when (current-buffer)
+    (setf (files:profile file) (profile (current-buffer)))))
```
aartaka, line 339: This looks fishy—what if we're in a `nosave` buffer and want to jsut expand some throwaway path?

tests/renderer-package.lisp
```diff
@@ -18,3 +18,16 @@
...
   #p"")
+
+(defvar +test-root+ #p"/tmp/nyxt-tests/")
+
+(defmethod files:resolve ((profile nyxt-user::test-profile) (file files:file))
+  "Store all files in a temporary `+test-root+' directory."
+  (nfiles:join +test-root+  (call-next-method)))
+
+(defmacro with-headless (&body body)
+  `(let ((old-headless-p nyxt::*headless-p*))
```
aartaka, line 29: This pattern of saving something to an `old-*` variable, `setf`-ing it to a new value and then safely restoring it—it's quite frequent and useful, I'm guessing. Maybe make a macro?

nyxt.asd
```diff
@@ -319,6 +321,7 @@
   :components ((:file "tests/renderer-package")
                (:file "tests/renderer-offline/set-url")
                (:file "tests/renderer-offline/execute-command-eval")
+               (:file "tests/renderer-offline/remembrance")
```
aartaka, line 324: What about the data files? Should they be listed somewhere? As `:static-file`-s?

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,367 @@
...
+It's only cached if its `last-update' is older than `update-interval'.
+
+Cached pages older than `discard-interval' are automatically purged.
+
+Return cached page.
+Return NIL if URL is not cached, for instance if it's on
+`nyxt/history-mode:history-mode' `history-blocklist'."
+  (let* ((url (sera:lret ((copy (quri:copy-uri (url buffer))))
+                ;; We drop the fragment as it does not change the page content.
+                (setf (quri:uri-fragment copy) nil)))
```
aartaka, line 183: Note that you can provide `:fragment nil` to `quri:copy-uri`, no need for `lret` here.

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,367 @@
...
+                                             ;; REVIEW: https://github.com/edicl/cl-ppcre/issues/51
+                                             ;; Only works for whole-word matches.
+                                             :preserve-case t)))
+    (spinneret:with-html-string
+      (:style (style mode))
+      (:h1 (format nil "Cached content of ~a" url-string))
+      (:div (:pre (:raw content))))))
+
+(define-command recollect-visited-page
+    (&key (return-actions (list (lambda-command buffer-load* (suggestion-values)
```
aartaka, line 269: We need to abstract `set-url` `return-actions` too :D

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,367 @@
...
+
+(defun search-cache (cache query)
+  "Return the entries matching QUERY."
+  (mapcar (alex:curry #'montezuma:get-document cache)
+          (mapcar #'montezuma:doc
+                  (montezuma:each
+                   (montezuma:search cache
+                                     query
+                                     ;; TODO: Is there a way to remove the limit?
+                                     :num-docs 10000)
```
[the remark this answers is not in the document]
aadcg: It seems that deleting `~/.cache/nyxt/remembrance.cache` fixed it.

aadcg:
Regarding displaying cached pages, they look a bit off, since the lines are too long and it seems that I can't even scroll them horizontally.



It seems that searching them yields no results... 


Another thing that looks off is the prompt message in the following prompt:

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,367 @@
...
+
+- One set of suggestions comes from natural language processing of the query;
+
+- Another set is the suggestions that contain an exact match of the whole query,
+  as a setence.
+
+The `view-cached-content' action displays the cached content (with highlighted
+query terms), even when offline."
+  (prompt
+   :prompt "Search cache"
```
aadcg, line 301: This prompt string seems off.

Ambrevar: Why? Suggestion?

nyxt.asd
Ambrevar: Why?

tests/renderer-package.lisp
Ambrevar: Parachute has this in its fixture macro.  Anything else?

source/configuration.lisp
Ambrevar: I'm not too satisfied with this, but I must say the relationship between buffers and profiles is a tricky one.

That said, I don't get your use case.  If you are in a nosave buffer, then the `write-file` method is disabled, but path expansion still works.

Ambrevar:
> Wait, it only works on HTTP(S) pages? I wanted to use the fulltext search on my own custom schemes...

No, it works with everything where Parenscript works.  Except internal pages, well, because there is (for now) little point.

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,367 @@
...
+   ;; TODO: `style' belongs to a separate mode, doesn't it?  Same question as with `history-tree-mode'.
+   (style (theme:themed-css (theme *browser*)
+            (* :margin 0
+               :padding 0
+               :list-style "none")))))
+
+(defmethod initialize-instance :after ((mode remembrance-mode) &key)
+  (setf (cache mode)
+        (make-instance 'montezuma:index
+                       :path (files:expand (cache-path mode))
```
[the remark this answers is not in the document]
Ambrevar: I can't reproduce.  I tried after removing ~/.cache/nyxt and it works for me.

source/mode/remembrance.lisp
```diff
@@ -0,0 +1,367 @@
...
+                                (lambda-command new-buffer-load (suggestion-values)
+                                  "Load cache entries in new buffer(s)."
+                                  (mapc (lambda (page-doc) (make-buffer :url (page-url page-doc))) (rest suggestion-values))
+                                  (make-buffer-focus :url (page-url (first suggestion-values))))
+                                (lambda-command copy-url* (suggestions)
+                                  "Copy bookmark URL."
+                                  (let ((url (page-url-string (first suggestions))))
+                                    (trivial-clipboard:text url)
+                                    (echo "Copied to clipboard: ~s" url)))
+                                (lambda-command view-cached-content (suggestions)
```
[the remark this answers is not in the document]
Ambrevar: Done.

Ambrevar:
> Regarding displaying cached pages, they look a bit off, since the lines are too long and it seems that I can't even scroll them horizontally.

Can't reproduce.  There should be an horizontal scroll bar at the bottom of the "pre" frame.
If you have a touchpad, horizontal scrolling also works fine.

That said, maybe we can use a regular `<p>` tag instead?  Thoughts?

Alternative: store both the innerText and the innerHTML:
- use the innerText for searching;
- use the innerHTML for rendering.

Drawback:
- It's (much?) heavier in the database;
- In some cases it may display worse content because HTML-without-JS is not always meant to be :p

Ambrevar:
> It seems that searching them yields no results...

You found a bug in `search-buffer-mode`!  It does not work on `pre` tags...  @aartaka Any idea why?

source/mode/remembrance.lisp
Ambrevar: Done!

Ambrevar:
The HTML content is now displayed in the view!

aadcg:
While testing this I got the following error:

```
EOF

Restarts:
 0: [ABORT] abort thread (#<THREAD "Nyxt renderer signal handler" RUNNING {100A298CB3}>)

Backtrace:
 0: ((:METHOD MONTEZUMA::REFILL (MONTEZUMA::BUFFERED-INDEX-INPUT)) #<MONTEZUMA::CS-INDEX-INPUT {1008BD7463}>) [fast-method]
      Locals:
        MONTEZUMA::SELF = #<MONTEZUMA::CS-INDEX-INPUT {1008BD7463}>
 1: ((:METHOD MONTEZUMA::READ-BYTE (MONTEZUMA::BUFFERED-INDEX-INPUT)) #<MONTEZUMA::CS-INDEX-INPUT {1008BD7463}>) [fast-method]
      Locals:
        MONTEZUMA::SELF = #<MONTEZUMA::CS-INDEX-INPUT {1008BD7463}>
 2: ((:METHOD MONTEZUMA::READ-VINT (MONTEZUMA::INDEX-INPUT)) #<MONTEZUMA::CS-INDEX-INPUT {1008BD7463}>) [fast-method]
      Locals:
        MONTEZUMA::SELF = #<MONTEZUMA::CS-INDEX-INPUT {1008BD7463}>
 3: ((:METHOD MONTEZUMA::READ-TERM-BUFFER (MONTEZUMA::TERM-BUFFER T T)) #<MONTEZUMA::TERM-BUFFER field:"content" text:"한국어" {1008BD7473}> #<MONTEZUMA::CS-INDEX-INPUT {1008BD7463}> #<MONTEZUMA::FIELD-INFOS..
      Locals:
        MONTEZUMA::FIELD-INFOS = #<MONTEZUMA::FIELD-INFOS {10085C84C3}>
        MONTEZUMA::INPUT = #<MONTEZUMA::CS-INDEX-INPUT {1008BD7463}>
        MONTEZUMA::SELF = #<MONTEZUMA::TERM-BUFFER field:"content" text:"한국어" {1008BD7473}>
 4: ((:METHOD MONTEZUMA::NEXT? (MONTEZUMA::SEGMENT-TERM-ENUM)) #<MONTEZUMA::SEGMENT-TERM-ENUM {100883F243}>) [fast-method]
      Locals:
        MONTEZUMA::SELF = #<MONTEZUMA::SEGMENT-TERM-ENUM {100883F243}>
 5: ((:METHOD MONTEZUMA::SCAN-TO (MONTEZUMA::SEGMENT-TERM-ENUM T)) #<MONTEZUMA::SEGMENT-TERM-ENUM {100883F243}> #S(MONTEZUMA::TERM :FIELD "url" :TEXT "https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikip..
      Locals:
        MONTEZUMA::SELF = #<MONTEZUMA::SEGMENT-TERM-ENUM {100883F243}>
        MONTEZUMA::TERM = #S(MONTEZUMA::TERM ..)
 6: ((:METHOD MONTEZUMA::SCAN-FOR-TERM-INFO (MONTEZUMA::TERM-INFOS-READER T)) #<MONTEZUMA::TERM-INFOS-READER {10085C84E3}> #S(MONTEZUMA::TERM :FIELD "url" :TEXT "https://duckduckgo.com/l/?uddg=https%3A%2F..
      Locals:
        MONTEZUMA::E = #<MONTEZUMA::SEGMENT-TERM-ENUM {100883F243}>
        MONTEZUMA::SELF = #<MONTEZUMA::TERM-INFOS-READER {10085C84E3}>
        MONTEZUMA::TERM = #S(MONTEZUMA::TERM ..)
 7: ((:METHOD MONTEZUMA::TERM-DOC-FREQ (MONTEZUMA::SEGMENT-READER T)) #<MONTEZUMA::SEGMENT-READER "_1" (1 docs, 0 deleted docs, 6 field infos) {1008543BA3}> #S(MONTEZUMA::TERM :FIELD "url" :TEXT "https://..
      Locals:
        MONTEZUMA::SELF = #<MONTEZUMA::SEGMENT-READER "_1" (1 docs, 0 deleted docs, 6 field infos) {1008543BA3}>
        MONTEZUMA::TERM = #S(MONTEZUMA::TERM ..)
 8: ((:METHOD INITIALIZE-INSTANCE :AFTER (MONTEZUMA::TERM-WEIGHT)) #<MONTEZUMA::TERM-WEIGHT query: #<MONTEZUMA:TERM-QUERY "url":"https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpec..
      Locals:
        MONTEZUMA::QUERY = #<MONTEZUMA:TERM-QUERY "url":"https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpecial%3ASearch%3Fsearch%3Dukraine%26go%3DGo&rut=25ff3b97aa81448fb84fa293c2d8d7a712f07809f371aaaa021bd7a912e1a077"^1.0 {100A2F7593}>
        MONTEZUMA::SEARCHER = #<MONTEZUMA:INDEX-SEARCHER {1008543BC3}>
        MONTEZUMA::SELF = #<MONTEZUMA::TERM-WEIGHT query: #<MONTEZUMA:TERM-QUERY "url":"https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpecial%3ASearch%3Fsearch%3Dukraine%26go%3DGo&rut=25ff3b97aa81448fb84fa293c2d8d7a712f07809f371aaaa021bd7a912e1a077"^1.0 {100A2F7593}> {100A2FD0E3}>
 9: ((LAMBDA (SB-PCL::|.P0.| SB-PCL::|.P1.|)) #<unavailable argument> #<unavailable argument>)
      [No Locals]
10: ((:METHOD MONTEZUMA:WEIGHT (MONTEZUMA::QUERY T)) #<MONTEZUMA:BOOLEAN-QUERY with 1 clauses: #<MONTEZUMA:BOOLEAN-CLAUSE :SHOULD-OCCUR #<MONTEZUMA:TERM-QUERY "url":"https://duckduckgo.com/l/?uddg=https%3..
      Locals:
        MONTEZUMA::SEARCHER = #<MONTEZUMA:INDEX-SEARCHER {1008543BC3}>
        MONTEZUMA::SELF = #<MONTEZUMA:BOOLEAN-QUERY with 1 clauses: #<MONTEZUMA:BOOLEAN-CLAUSE ..>>
11: ((:METHOD MONTEZUMA:SEARCH (MONTEZUMA:INDEX-SEARCHER T)) #<MONTEZUMA:INDEX-SEARCHER {1008543BC3}> #<MONTEZUMA:BOOLEAN-QUERY with 1 clauses: #<MONTEZUMA:BOOLEAN-CLAUSE :SHOULD-OCCUR #<MONTEZUMA:TERM-QU..
      Locals:
        MONTEZUMA::FILTER = NIL
        MONTEZUMA::MAX-SIZE = 10000
        MONTEZUMA::NUM-DOCS = 10000
        MONTEZUMA::OPTIONS = (:NUM-DOCS 10000)
        MONTEZUMA::QUERY = #<MONTEZUMA:BOOLEAN-QUERY with 1 clauses: #<MONTEZUMA:BOOLEAN-CLAUSE ..>>
        MONTEZUMA::SELF = #<MONTEZUMA:INDEX-SEARCHER {1008543BC3}>
        MONTEZUMA::SORT = NIL
12: ((:METHOD NYXT/REMEMBRANCE-MODE::SEARCH-CACHE (NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE T)) #<NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE {10082250D3}> #<MONTEZUMA:BOOLEAN-QUERY with 1 clauses: #<MONTEZUMA:B..
      Locals:
        NYXT:MODE = #<NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE {10082250D3}>
        NYXT:QUERY = #<MONTEZUMA:BOOLEAN-QUERY with 1 clauses: #<MONTEZUMA:BOOLEAN-CLAUSE ..>>
13: (NYXT/REMEMBRANCE-MODE::FIND-URL #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpecial%3ASearch%3Fsearch%3Dukraine%26go%3DGo&rut=25ff3b97aa81448fb84f..
      Locals:
        REMEMBRANCE-MODE = #<NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE {10082250D3}>
        URL = #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpecial%3ASearch%3Fsearch%3Dukraine%26go%3DGo&rut=25ff3b97aa81448fb84fa293c2d8d7a712f07809f371aaaa021bd7a912e1a077>
14: (NYXT/REMEMBRANCE-MODE::BUFFER->CACHE #<NYXT:WEB-BUFFER 816 {1007CC8783}> #<NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE {10082250D3}>)
      Locals:
        BUFFER = #<NYXT:WEB-BUFFER 816 {1007CC8783}>
        REMEMBRANCE-MODE = #<NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE {10082250D3}>
        URL = #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpecial%3ASearch%3Fsearch%3Dukraine%26go%3DGo&rut=25ff3b97aa81448fb84fa293c2d8d7a712f07809f371aaaa021bd7a912e1a077>
15: ((:METHOD NYXT:ON-SIGNAL-LOAD-FINISHED (NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE T)) #<NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE {10082250D3}> #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/l/?uddg=https..
      Locals:
        NYXT:MODE = #<NYXT/REMEMBRANCE-MODE:REMEMBRANCE-MODE {10082250D3}>
        NYXT:URL = #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpecial%3ASearch%3Fsearch%3Dukraine%26go%3DGo&rut=25ff3b97aa81448fb84fa293c2d8d7a712f07809f371aaaa021bd7a912e1a077>
16: ((:METHOD NYXT:ON-SIGNAL-LOAD-FINISHED (NYXT:BUFFER T)) #<NYXT:WEB-BUFFER 816 {1007CC8783}> #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpecial%3AS..
      Locals:
        NYXT:BUFFER = #<NYXT:WEB-BUFFER 816 {1007CC8783}>
        NYXT:URL = #<QURI.URI.HTTP:URI-HTTPS https://duckduckgo.com/l/?uddg=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSpecial%3ASearch%3Fsearch%3Dukraine%26go%3DGo&rut=25ff3b97aa81448fb84fa293c2d8d7a712f07809f371aaaa021bd7a912e1a077>
```

source/mode/remembrance.lisp
aadcg: The prompt lists the buffers that for which `remembrance-mode` is/was active. The user inserts a string that is matched against the textual content of those buffers. Am I correct?

This is confusing for the user since as he constructs the string, the list of suggestions is composed by buffer, not search matches. Unlike, say `search-buffer`, the user doesn't understand what the matches are. 

Ideally, we should be able to show the exact match and the buffer/URL corresponding to the match (via object attributes). Is there any technical limitation that stops us?

aartaka:
> > It seems that searching them yields no results...
> 
> You found a bug in `search-buffer-mode`! It does not work on `pre` tags... @aartaka Any idea why?

Whaaaaaat? No way...

Ambrevar:
> > You found a bug in `search-buffer-mode`! It does not work on `pre` tags... @aartaka Any idea why?
> 
> Whaaaaaat? No way...

Sorry, actually I got misled: it's just that the `pre` tag background is of the same color as the search hint!

Ambrevar:
> While testing this I got the following error:

Never saw this.

Did you get it just when loading a page?

jmercouris:
I've no more feedback besides what I initially suggested yesterday. After looking again, everything looks in order. Please, feel free to merge whenever you're ready.

source/mode/remembrance.lisp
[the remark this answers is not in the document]
Ambrevar: @aartaka Could you reproduce this?

source/mode/remembrance.lisp
Ambrevar: Yes: where would you display this match in the prompt buffer?  There is no room for it.

A better option would be to add a selection action.  I'll see what I can do.

aartaka:
Wait, do you cache HTML now? Why so?

Note that hosting arbitrary HMLT inside `nyxt:` URLs is unsafe, as those have lots of privileges, like `lisp-eval` etc.

aartaka:
So either use `cl-readability` to sanitize the page to the more-or-less readable and safe state (although Mozilla themselves suggest using some other HTML sanitizer on top of their Readability.js, because security is no concern for their library), or only use `innerText` that's guaranteed to be plain!

source/mode/remembrance.lisp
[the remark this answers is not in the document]
aartaka: No, it was a one-shot error, I guess. Cannot reproduce anymore.

Quite alarming Heizenbug, though.

Ambrevar:
Ooooooh you are right xD  This is really bad :(

How can cl-readability safely guarantee that there won't be any lisp-eval in the HTML?

Alternatively, can't we just display the page in a non-trusted scheme?  Maybe that was your original suggestion...

Ambrevar:
@aartaka I've updated the command to work like `view-source`.  We should be good now.

One nit: this page has no style.   How do I set the style to that of the other nyxt:// internal pages?

aartaka:
> @aartaka I've updated the command to work like `view-source`. We should be good now.
> 
> One nit: this page has no style. How do I set the style to that of the other nyxt:// internal pages?

The problem is, likely, with the relative URL resolution. You'd need to go through all the URLs on the page and set their `href` and `src` to an absolute value. Here's what `cl-readability` has (you can shorten it in half, because `cl-readability` version a calque from the JS one and is overly concerned with minute details that are not necessary for `remembrance-mode`):
``` lisp
(defgeneric fix-relative-urls (element)
  (:method ((element t))
    (flet ((relative->absolute (uri) ;; toAbsoluteURI
             (when (and uri
                        (not (uiop:emptyp (ignore-errors (quri:render-uri (quri:uri uri)))))
                        (not (uiop:string-prefix-p "#" uri)))
               (quri:render-uri
                (quri:merge-uris (quri:uri uri)
                                 (quri:uri *document-url*))))))
      (dolist (node (qsa element "a"))
        (cond
          ;; Remove links with javascript: URIs, since they won't work
          ;; after scripts have been removed from the page.
          ((and (attr node "href")
                (uiop:string-prefix-p "javascript:" (attr node "href")))
           (if (zerop (length (child-nodes node)))
               ;; If the link only contains simple text content, it can be converted to a text node.
               (replace-child node (make-text-node (inner-text node)))
               ;; If the link has multiple children, they should all be preserved.
               (set-tag-name node "span")))
          ((attr node "href")
           (setf (attr node "href") (relative->absolute (attr node "href"))))
          (t nil)))
      (dolist (node (qsa element "img" "picture" "figure" "video" "audio" "source"))
        (when (attr node "src")
          (setf (attr node "src") (relative->absolute (attr node "src"))))
        (when (attr node "poster")
          (setf (attr node "poster") (relative->absolute (attr node "poster"))))
        ;; TODO: srcset
        ;;   if (srcset) {
        ;;   var newSrcset = srcset.replace(this.REGEXPS.srcsetUrl, function(_, p1, p2, p3) {
        ;;     return toAbsoluteURI(p1) + (p2 || "") + p3;
        ;;   });

        ;;   media.setAttribute("srcset", newSrcset);
        ;; }
        )))
  (:documentation "Converts each <a> and <img> uri in the given element to an absolute URI, ignoring #ref URIs.

Readability._fixRelativeUris()"))
```

aartaka:
Yes :)

Ambrevar:
OK, let's merge after 3-pre-release-2.

aadcg:
What's the status to this PR? Does it need to be reviewed?

Ambrevar:
@aadcg If you want to give it a review, please do :)
I'll merge very soon.

aadcg:
Feel free to merge it when you please @Ambrevar!

aartaka:
Cool :D

jmercouris:
Thank you Pierre, good persistence! :-) see what I did there? Get it?

PR #2543, opened 2022-09-05
Allow bindings commands to context menu actions.

aartaka:
# Description

This allows binding Lisp commands to context menu actions with the help of the new `ffi-add-context-menu-command`. Those actions are only available when the respecitve command is (which means always for global commands, and only when the mode is enabled—for mode-specific ones). It's quite basic, which might be a good reason it works in the first place xD

# Discussion

- [ ] Any idea for a better design? Maybe include this functionality in `define-command` somehow?
- [ ] Now that I think of it, internal schemes and `define-internal-scheme` should also have an `ffi-` prefix, shouldn't they? Or maybe we should provide the most intuitive interface possible, while doing the renderer-specific things based on internal information? In this case propagating context menu actions inside `define-*command-*` makes lots of sense...
- [X] I've updated `cl-webkit` to 3.5.4, incorporating all the recent useful changes. @Ambrevar, care to send a patch to Guix?

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/foreign-interface.lisp
+++ b/source/foreign-interface.lisp
@@ -363,3 +363,20 @@ Return the text cut."))
   (:method ((buffer t))
     (echo-warning "Redoing edits is not yet implemented for this renderer."))
   (:documentation "Redo the last undone text edit performed in BUFFER's web view."))
+
+(defvar *context-menu-commands* (make-hash-table)
+  "A hash table from command symbols to context menu labels for those.
+Once a context menu appears, those commands will be added to it as actions with
+the labels they have as hash values.")
+
+(define-ffi-generic ffi-add-context-menu-command (command &optional label)
+  (:method :around ((command command) &optional label)
+    (call-next-method command (or label (format nil "~:(~a~)" (str:remove-punctuation (symbol-name (name command)))))))
+  (:method ((command command) &optional label)
+    (setf (gethash (name command) *context-menu-commands*)
+          label))
+  (:method ((command symbol) &optional label)
+    (ffi-add-context-menu-command (symbol-function command) label))
+  (:documentation "Add COMMAND as accessible in context menus with LABEL displayed for it.
+If LABEL is not provided, it's auto-generated from the command name.
+COMMAND should be either a symbol naming an already defined command, or the command object itself."))
--- a/source/foreign-interface.lisp
+++ b/source/foreign-interface.lisp
@@ -371,7 +371,7 @@ the labels they have as hash values.")
 
 (define-ffi-generic ffi-add-context-menu-command (command &optional label)
   (:method :around ((command command) &optional label)
-    (call-next-method command (or label (format nil "~:(~a~)" (str:remove-punctuation (symbol-name (name command)))))))
+    (call-next-method command (or label (str:capitalize (str:remove-punctuation (symbol-name (name command)))))))
   (:method ((command command) &optional label)
     (setf (gethash (name command) *context-menu-commands*)
           label))
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -1598,12 +1598,12 @@ See `finalize-buffer'."
                       (context-commands (alex:hash-table-keys *context-menu-commands*))
                       (accessible-context-commands (intersection accessible-commands context-commands)))
         (dolist (command accessible-context-commands)
-          (let* ((label (gethash command *context-menu-commands*))
-                 ;; Using stock actions here, because cl-cffi-gtk has a
-                 ;; terrible API for GActions, requiring an exact type to be
-                 ;; passed and disallowing NULL as a type :/
-                 (item (webkit:webkit-context-menu-item-new-from-stock-action-with-label
-                        :webkit-context-menu-action-action-custom label)))
+          ;; Using stock actions here, because cl-cffi-gtk has a terrible API
+          ;; for GActions, requiring an exact type to be passed and disallowing
+          ;; NULL as a type :/
+          (let ((item (webkit:webkit-context-menu-item-new-from-stock-action-with-label
+                       :webkit-context-menu-action-action-custom
+                       (gethash command *context-menu-commands*))))
             (gobject:g-signal-connect
              (webkit:webkit-context-menu-item-get-g-action item) "activate"
              (lambda (action parameter)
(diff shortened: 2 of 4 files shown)


aartaka:
Just to mention, WebKit context menu items:
- Can be checkboxes or radios.
- Can be grouped.
- Can access DOM/widget they are pointing at.
- Get enabled/disabled given certain conditions (like pointing at a textarea or a video).
- Can store arbitrary data and modify it besed on item presses.

None of this is immediately available to items introduced in this PR, but then, those are full-blown Nyxt commands, so maybe we don't need any of this shiny styff...

aadcg:
@Ambrevar, @aartaka I've sent the patch to Guix (it's bug #57603).

Ambrevar:
It's already merged :)

aartaka:
So, anyone willing to review this? Shall I merge?

source/foreign-interface.lisp
```diff
@@ -363,3 +363,20 @@ Return the text cut."))
...
   (:documentation "Redo the last undone text edit performed in BUFFER's web view."))
+
+(defvar *context-menu-commands* (make-hash-table)
+  "A hash table from command symbols to context menu labels for those.
+Once a context menu appears, those commands will be added to it as actions with
+the labels they have as hash values.")
+
+(define-ffi-generic ffi-add-context-menu-command (command &optional label)
+  (:method :around ((command command) &optional label)
+    (call-next-method command (or label (format nil "~:(~a~)" (str:remove-punctuation (symbol-name (name command)))))))
```
aadcg, line 374: Is `format` worth it here? It's a bit hard to follow.

Perhaps use `str:sentence-case` and `str:remove-punctuation`?

source/renderer/gtk.lisp
```diff
@@ -1583,7 +1583,31 @@ See `finalize-buffer'."
                   (webkit:webkit-context-menu-item-new-from-stock-action-with-label
                    :webkit-context-menu-action-open-link-in-new-window
                    "Open Link in New Buffer")
-                  i)))))))
+                  i))))))
+      (webkit:webkit-context-menu-append
+       context-menu (webkit:webkit-context-menu-item-new-separator))
```
aadcg, line 1588:
```
(webkit:webkit-context-menu-append context-menu
                                   (webkit:webkit-context-menu-item-new-separator))
```

Perhaps the indentation above would be easier to follow.

source/renderer/gtk.lisp
```diff
@@ -1583,7 +1583,31 @@ See `finalize-buffer'."
...
+      (sera:and-let* ((_ (plusp (hash-table-count *context-menu-commands*)))
```
aadcg, line 1589: Perhaps `alex:positive-integer-p`? `plusp` checks for reals.

aartaka: `plusp` does check for reals, yes, but `hash-table-count` returns solid non-negative integers, so `plusp` is likely to be optimized into a non-negative-integer-only version here.

source/renderer/gtk.lisp
aartaka: I usually optimize for minimal absolute indentation, so my version appeals to myself much more tha yours which makes the next indentation level to be extremely far from the current one.

But there's no answer here, I guess ¯\\\_(ツ)_/¯

source/renderer/gtk.lisp
```diff
@@ -1583,7 +1583,31 @@ See `finalize-buffer'."
...
+                      (accessible-commands
+                       (mapcar #'name
+                               (list-commands
+                                :global-p t
+                                :mode-symbols (mapcar #'sera:class-name-of
```
aadcg, line 1594: The pattern `(mapcar #'sera:class-name-of foo` seems to appear 8 times in the codebase. Perhaps it should be abstracted?

source/renderer/gtk.lisp
```diff
@@ -1583,7 +1583,31 @@ See `finalize-buffer'."
...
+                                                      (sera:filter #'enabled-p (modes buffer))))))
+                      (context-commands (alex:hash-table-keys *context-menu-commands*))
+                      (accessible-context-commands (intersection accessible-commands context-commands)))
```
aadcg, line 1597: I'd leave this binding out and would jump straight into:

`(dolist (command  (intersection accessible-commands context-commands)) ...`

source/renderer/gtk.lisp
```diff
@@ -1583,7 +1583,31 @@ See `finalize-buffer'."
...
+        (dolist (command accessible-context-commands)
+          (let* ((label (gethash command *context-menu-commands*))
```
aadcg, line 1599: I'd leave the `label` binding out as well.

source/renderer/gtk.lisp
```diff
@@ -1583,7 +1583,31 @@ See `finalize-buffer'."
...
+                 ;; Using stock actions here, because cl-cffi-gtk has a
+                 ;; terrible API for GActions, requiring an exact type to be
+                 ;; passed and disallowing NULL as a type :/
+                 (item (webkit:webkit-context-menu-item-new-from-stock-action-with-label
+                        :webkit-context-menu-action-action-custom label)))
+            (gobject:g-signal-connect
+             (webkit:webkit-context-menu-item-get-g-action item) "activate"
+             (lambda (action parameter)
+               (declare (ignore action parameter))
+               (run-async command)))
```
aadcg, line 1609:
```lisp
(gobject:g-signal-connect (webkit:webkit-context-menu-item-get-g-action item)
                                      "activate"
                                      (lambda (action parameter)
                                        (declare (ignore action parameter))
                                        (run-async command)))
```

I find the above indentation easier to follow.

aadcg:
Very useful functionality. Regarding design:

> Maybe include this functionality in define-command somehow?

That seems clear, in theory.  If you have an idea right away, you could try it.  Otherwise, I think it's absolutely fine to open an issue and leave it for later!

source/renderer/gtk.lisp
aartaka: Maybe, but at this point we can only abstract towards APL...

aartaka:
Let's leave it for later (⌒_⌒;)

aartaka:
The indentation I'm not changing ヽ(ｏ`皿′ｏ)ﾉ

aadcg:
> Let's leave it for later (⌒_⌒;)

Totally agree. There's too much going on... :woozy_face:

aadcg:
> The indentation I'm not changing ヽ(ｏ`皿′ｏ)ﾉ

I'm only pushing for conventions, whatever the choice might be :)

PR #2546, opened 2022-09-06
libraries/user-interface: rename query-select to query-selector

hendursaga:
I found this while trying to download something, which resulted in an infinite loop of ParenScript errors.


aadcg:
Thanks @hendursaga. I made this terrible mistake!

PR #2547, opened 2022-09-06
Overpower `if-confirm`.

aartaka:
# Description

This transforms `if-confirm` (and the underlying `prompter:yes-no-source`) to be or more universal utility. In particular, `if-confirm` can enjoy custom "yes"/"no" phrases, like "grant"/"deny" or "overwrite"/"cancel".

# Discussion

This change comes at the cost of `prompt` argument to `if-confirm` becoming more restricted (it should evaluate to a string now, instead of being a fuzzy `format nil` arglist. In the best case, it adds one more set of parentheses, in the worst case it adds a `format nil`. Which is fine by me, but can be devastating for extensions writers...

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

Looks sane?


source/configuration.lisp
```diff
@@ -347,19 +347,41 @@ To discover the default value of a slot or all slots of a class, use the
          ;; TODO: Raise error instead?
          (log:warn "Expected buffer, got ~a" ,buffer))))
 
+;; TODO: Disallow canceling the prompt? Allow changing order of YES and NO so
```
Ambrevar, line 350: We can't do this by design.  But some widget systems have so called "modal windows" which do precisely this.

source/configuration.lisp
```diff
@@ -347,19 +347,41 @@ To discover the default value of a slot or all slots of a class, use the
...
+;; that one makes a conscious effort to choose a YES?
+;; TODO: Add an "always (yes|no)" and cache the result?
```
Ambrevar, line 352: What do you mean, "cache the result"?  Example?

Ambrevar:
Looks OK to me!

source/configuration.lisp
aartaka: We can do this with `handler-bind` on `prompt-buffer-canceled`, no?

source/configuration.lisp
aartaka: I mean, like remembering the permission settings for websites that asked for permissions.

Ambrevar: OK, then "default to the previous answer on next prompt"?

source/configuration.lisp
Ambrevar: The problem is _focus_.  It's possible for the user to keep using the context buffer while the prompt buffer is up.  This is by design.

If we want a "modal prompt buffer", we need to forbid focus somehow.  Or display the prompt buffer in a GTK modal window.

Or... Just use a GTK modal window (also know as "confirmation dialog") instead of the prompt buffer for `if-confirm`.

source/configuration.lisp
aartaka: But that's likely beyond `if-confirm`.

source/configuration.lisp
aartaka: What I mean is the interpretation of prompt cancellation as the "no" answer. It's not necessarily intuitive. Focusing I don't care about, because most `if-confirm` prompts are not related to context buffer actions anyway.

source/configuration.lisp
aartaka: Done :)

Issue #2537, opened 2022-09-02
Add a user-funcallable-class?


aartaka:
**Is your feature request related to a problem? Please describe.**

We have configurable `user-class`-es for almost everything in Nyxt. `command` and `internal-page` are classes too, although not `user-class`-es. We may bump into this at some point, requiring to configure some part of the internal page, for example.

It would be nice if this configuration would work by the same rules as the configuration of regular user classes, with `define-configuration` and `customize-instance` as the convenient ways to hook into commands and change their behavior.

**Describe the solution you'd like**

Introduce `user-funcallable-class` and make e.g. `command` and `internal-page` to inherit from it.

**Describe alternatives you've considered**

Well, both `command` and `internal-page` inherit from generics and are non-specialized method by default, which makes them hookable with `:before`, `:after`, and `:around` methods and `intitialize-instance` methods too. It's a lot of configuration space. For example, all the settings can be tweaked in `initialize-instance :after` of a command (by the way, should we use `cascade` method combination for those too, so that parent methods are invoked too?). But still, having the proper and `define-configuration`-friendly way to configure those would be much better and cleaner than overriding some `initialize-instance` or something.

**Additional context**

I've tried to implement it myself, adding `user-funcallable-class` (inheriting from `user-class` and `closer-mop:funcallable-standard-class`), and then trying to use it as command metaclass. This caused (reasonable) conflict between the `standard-funcallable-class standard-generic-function` and `user-funcallable-class command`. Adding yet another layer of indirection (`user-generic-function`) didn't help -- `standard-generic-function` is still there in the hierarchy, breaking this fragile construction into pieces. Do we really need `standard-generic-function` there? (҂⌣̀_⌣́)

aadcg:
> It would be nice if this configuration would work by the same rules as the configuration of regular user classes, with define-configuration and customize-instance as the convenient ways to hook into commands and change their behavior.

Seems absolutely reasonable to me. If there's no good reason to break this symmetry, then let's try to preserve it.

Can't give any valuable insight. I'm wondering: what tells apart `user-class`es and `user-funcallable-class`es, in your construction?

aartaka:
> I'm wondering: what tells apart `user-class`es and `user-funcallable-class`es, in your construction?

The fact that the latter are `funcall`-able :P 

[...]

PR #2551, opened 2022-09-07
User funcallable class

aartaka:
# Description

This is a small addition as a first step to address the #2537. `user-funcallable-class`, as implemented here, is not immediately useful, as we usually initialize commands not with `make-instance` but rather with `defgeneric` and `setf`, which destroys the purpose of user configuration there. I've tried to refactor our commands to use `make-instance` instead, but it seems that this part of the spec is too rigid for dynamic generic function creation to be easy (/ω＼)

However, my idea for the future work is to initialize commands, internal pages, and panels in a `customize-instance`-friendly manner. Need to think on it for some time, though :)

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/user-classes.lisp
+++ b/source/user-classes.lisp
@@ -46,6 +46,33 @@ Do not specialize the standard method in public code, prefer
             (sera:filter #'user-class-p (cons class (mopu:superclasses class))))
     (apply #'customize-instance initialized-object initargs)))
 
+(defclass user-funcallable-class (closer-mop:funcallable-standard-class)
+  ((customize-hook :initform (make-instance 'hooks:hook-any)
+                   :documentation "An internal hook to add customization handlers to.
+
+Reserved for `define-configuration'.
+
+Prefer `define-configuration' and `customize-instance' instead."))
+  (:documentation "Classes using this metaclass will call `customize-instance' on instantiation.
+This is useful to expose a class configuration knob to the user."))
+(export-always 'user-funcallable-class)
+
+(defmethod closer-mop:validate-superclass ((class user-funcallable-class)
+                                           (superclass closer-mop:funcallable-standard-class))
+  t)
+
+(defmethod closer-mop:validate-superclass ((superclass closer-mop:funcallable-standard-class)
+                                           (class user-funcallable-class))
+  t)
+
+(defmethod #+nyxt-debug-make-instance cl:make-instance #-nyxt-debug-make-instance make-instance
+  :around ((class user-funcallable-class) &rest initargs &key &allow-other-keys)
+  (sera:lret ((initialized-object (call-next-method)))
+    (mapcar (lambda (class)
+              (hooks:run-hook (slot-value class 'customize-hook) initialized-object))
+            (sera:filter #'user-class-p (cons class (mopu:superclasses class))))
+    (apply #'customize-instance initialized-object initargs)))
+
 (defun user-class-p (class-specifier)
   (let ((metaclass (cond
                      ((symbolp  class-specifier)
@@ -53,7 +80,8 @@ Do not specialize the standard method in public code, prefer
                      ((closer-mop:classp class-specifier)
                       class-specifier)
                      (t (class-of class-specifier)))))
-    (typep metaclass 'user-class)))
+    (or (typep metaclass 'user-class)
+        (typep metaclass 'user-funcallable-class))))
 
 (defclass interface-class (standard-class) ()
   (:documentation "An interface class exists solely for the purpose of
--- a/source/user-classes.lisp
+++ b/source/user-classes.lisp
@@ -3,8 +3,9 @@
 
 (in-package :nyxt)
 
-(defclass user-class (standard-class)
+(defclass user-mixin-class ()
   ((customize-hook :initform (make-instance 'hooks:hook-any)
+                   :accessor customize-hook
                    :documentation "An internal hook to add customization handlers to.
 
 Reserved for `define-configuration'.
@@ -12,16 +13,41 @@ Reserved for `define-configuration'.
 Prefer `define-configuration' and `customize-instance' instead."))
   (:documentation "Classes using this metaclass will call `customize-instance'
 on instantiation.
-This is useful to expose a class configuration knob to the user."))
+This is useful to expose a class configuration knob to the user.
+
+This class is also for portable configuration management. Imagine some
+implementation has structs totally separate from classes, and thus configuring
+structs may be possible, but would require adding yet another class. Say,
+user-structure-class. With user-mixin-class, that would be a one-liner:
+
+(defclass user-structure-class (impl:built-in-structure-class user-mixin-class) ())
+
+while having the customize-hook and other configuration functionality in 3+
+classes (user-class, user-funcallable-class, and user-structure-class in case of
+this imaginary implementation) would already be 3x+ code duplication..."))
+
+(defclass user-class (standard-class user-mixin-class)
+  ()
+  (:documentation "User-configurable value class.
+Can be configured using `customize-instance' and `customize-hook'."))
 (export-always 'user-class)
 
-(defmethod closer-mop:validate-superclass ((class user-class)
-                                           (superclass standard-class))
-  t)
+(defmethod closer-mop:validate-superclass ((class user-class) (superclass standard-class)) t)
+(defmethod closer-mop:validate-superclass ((superclass standard-class) (class user-class)) t)
+(defmethod closer-mop:validate-superclass ((class user-class) (superclass user-mixin-class)) t)
+(defmethod closer-mop:validate-superclass ((superclass user-mixin-class) (class user-class)) t)
 
-(defmethod closer-mop:validate-superclass ((superclass standard-class)
-                                           (class user-class))
-  t)
+(defclass user-funcallable-class (closer-mop:funcallable-standard-class user-mixin-class)
+  ()
+  (:documentation "User-configurable class that behaves like function.
+One can use `funcall' on it, thus funcallable.
+Can be configured using `customize-instance' and `customize-hook'."))
+(export-always 'user-funcallable-class)
+
+(defmethod closer-mop:validate-superclass ((class user-funcallable-class) (superclass closer-mop:funcallable-standard-class)) t)
+(defmethod closer-mop:validate-superclass ((superclass closer-mop:funcallable-standard-class) (class user-funcallable-class)) t)
+(defmethod closer-mop:validate-superclass ((class user-funcallable-class) (superclass user-mixin-class)) t)
+(defmethod closer-mop:validate-superclass ((superclass user-mixin-class) (class user-funcallable-class)) t)
 
 (export-always 'customize-instance)
 (defgeneric customize-instance (object &key &allow-other-keys)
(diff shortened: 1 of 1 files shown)


aadcg:
> we usually initialize commands not with make-instance but rather with defgeneric and setf

I might be wrong, but I think you're abusing language here.  We don't initialize command with `defgeneric`. Commands are instantiated via `initialize-instance`, which rely on the fact that for every command there is a generic.

aadcg:
> which destroys the purpose of user configuration there

I'm afraid I didn't follow you here. I thought users could still resort to `customize-instance`, but I might be missing something.

aartaka:
> > we usually initialize commands not with make-instance but rather with defgeneric and setf
> 
> I might be wrong, but I think you're abusing language here. We don't initialize command with `defgeneric`. Commands are instantiated via `initialize-instance`, which rely on the fact that for every command there is a generic.

I'm not abusing the language. We initialize commands with `defgeneric`, see L171 in command.lisp. This `defgeneric` calls `ensure-generic-function`, which, in turn, calls `make-instance`. That's why our `initialize-instance :after` methods do fire, but are not doing anything exactly useful and argument-dependent. Same for the hypothetical `customize-instance`—even if it fires, it cannot do anything, because `defgeneric` has already done everything basic, while not allowing to hook into anything non-trivial.

So, yes, we initialize commands with `make-instance`, but then it's either:
- Hidden from us in `define-configuration` using `defgeneric`, or
- Simplistic and non-global in `make-command`.

We need to somehow unify those two, so that our customization methods become useful by getting access to initialization arguments in `defgeneric` before they are applied and turned into funcallable instance.

Answering another comment: `customize-instance` purpose is to fire before any irrevocable, rendrer-specific, or finalizing action happens. Once `customize-instance` is done, we can be sure that slots are those that user wants. This requires a certain mental/code discipline, like putting `gtk-object` initialization into `customize-instance :after` instead of a mere `intitialize-instance :after`. While it's more or less clear with `user-class` objects, it's much fuzzier with `user-funcallable-class` objects, because we use those two ways of creation for those (`make-instance` and `defgeneric`), and use stateful modification _after_ both `initialize-instance` and `customize-instance` are done. In other words, our funcallable classes are not finalized even after `customize-instance` is done, which makes them unconfigurable :/

Ambrevar:
Can you share what's you've tried and that didn't work?

aartaka:
> Can you share what's you've tried and that didn't work?

I've tried doing a mere `make-instance 'command :lambda-expression ... :lambda-list ... :name ...` (where the latter two are `standard-generic-function` initargs), with and without:
- `symbol-function` setting.
- `fdefinition` setting.
- `ensure-generic-function`.
And all of those ended up failing with `change-class` being unable to change from `command` to `standard-generic-function`, which seems to somehow opaquely happen inside `defgeneric`...

aadcg:
@aartaka thanks for the detailed explanation. It all makes sense now!

aartaka:
Okay, this
``` diff
modified   source/command.lisp
@@ -165,14 +165,13 @@ Example:
   (let ((doc (or (nth-value 2 (alex:parse-body body :documentation t)) "")))
     `(progn
        (export-always ',name (symbol-package ',name))
-       ;; Warning: We use `defgeneric' instead of `make-instance' (or even
-       ;; `ensure-generic-function') so that the compiler stores source location
-       ;; information (for "go to definition" to work.
-       (sera:lret ((gf (defgeneric ,name (,@(generalize-lambda-list arglist))
-                (:documentation ,doc)
-                (:method (,@arglist) ,@body)
-                (:generic-function-class command))))
-         (setf (slot-value gf 'visibility) :mode)))))
+       (setf (fdefinition (quote ,name))
+             (make-instance 'command
+                            :lambda-list (quote ,(generalize-lambda-list arglist))
+                            :name (quote ,name)
+                            :documentation ,doc
+                            :visibility :mode
+                            :lambda-expression (quote (lambda (,@arglist) ,@body)))))))
 
 (export-always 'define-command-global)
 (defmacro define-command-global (name (&rest arglist) &body body)
```

Seems to work. At least functions are being called properly. However, the whole compilation log is littered with warnings about undefined functions, and the source of commands is no longer locatable. I'm not sure what to do there (-_-)ゞ゛

aartaka:
I guess it's just SBCL shouting on us for doing something non-obvious xD It find the functions in the end, after all...

aartaka:
Just realized this: commands are not generics, commands are methods! They have advices, they are single-instance, they are defined as regular functions (which methods without specializers, given implicit generic creation, kind of are).

If we define commands as methods, they will flawlessly integrate with `user-class` (because `standard-method` is a `standard-class`), have their `:around` methods, and can preserve the source code (because generic functions will be defined in a normal way—with `defgeneric`). I only see benefits in this approach `\( ‘з’)/`

@Ambrevar, any reason you've gone with commands-as-generics in the first place? Anything I might be missing?

This PR is complete then. Let's merge, and I'll do the commands-as-methods rewrite in another PR.

Ambrevar:
> Just realized this: commands are not generics, commands are methods! They have advices, they are single-instance, they are defined as regular functions (which methods without specializers, given implicit generic creation, kind of are).

I'm not following.

Commands _are_ generic functions.  The body of the command in `define-command` is used in the only method of the command.

To be technical, we can call generic functions, we cannot call methods.

Which advice do you mean?

`command` has multiple instances, each command indeed.  What do you mean with "single-instance"?

aartaka:
> I'm not following.
> 
> Commands _are_ generic functions. The body of the command in `define-command` is used in the only method of the command.

Yes, you're right on the technical level. On the conceptual level, though, the body of the command is part of the method, so generic function is rather an "implementation detail" there.

> To be technical, we can call generic functions, we cannot call methods.

Sure, that I'm not arguing with either :)

> Which advice do you mean?

I mean `:before`, `:after`, and `:around` methods.

> `command` has multiple instances, each command indeed. What do you mean with "single-instance"?

What? Really? 

What I mean with single instance is that there's always only one unqualified method for every command. We don't specialize it against anything, so having `command` to be a generic function is too high of a level here. It's rather the complexity/abstraction level of a method that `command` belongs to.

Ambrevar:
Can you present a Lispy draft of what you have in mind?

aartaka:
> Can you present a Lispy draft of what you have in mind?

Yes, working on it!

Ambrevar:
But no rush, this is really not a priority. We don't even have a use case for this yet...

jmercouris:
Agreed, we have other more pressing issues!

I believe we have methods as commands because at some point we discussed the notion of a relationship between modes and commands. I can't recall the specifics of the conversation, unfortunately.

Ambrevar:
Also please don't merge this before 3.0 as it would raise a whole new wave of testing...

@jmercouris I believe what you are thinking about is a different discussion from Nyxt 1.* era.  The reason why commands today are defgeneric is a recent change in Nyxt 3.*.

Ambrevar:
Can you sum up the benefits of this pull request?

aadcg:
What's the status here? I delegate the review of this PR to @Ambrevar (when he's available, no rush!).

aartaka:
I removed `user-funcallable-class` from `command`, and this should make this PR ready to merge.

source/user-classes.lisp
```diff
@@ -13,15 +14,28 @@ Prefer `define-configuration' and `customize-instance' instead."))
...
+(defmethod closer-mop:validate-superclass ((superclass standard-class) (class user-class)) t)
+(defmethod closer-mop:validate-superclass ((class user-class) (superclass user-mixin-class)) t)
+(defmethod closer-mop:validate-superclass ((superclass user-mixin-class) (class user-class)) t)
 
-(defmethod closer-mop:validate-superclass ((superclass standard-class)
-                                           (class user-class))
-  t)
+(defclass user-funcallable-class (closer-mop:funcallable-standard-class user-mixin-class)
+  ()
+  (:documentation "Use-configurable funcallable class.
```
Ambrevar, line 31: Typo: "User-configurable", no?

source/user-classes.lisp
```diff
@@ -13,15 +14,28 @@ Prefer `define-configuration' and `customize-instance' instead."))
...
+Can be configured using `customize-instance' and `customize-hook'."))
```
Ambrevar, line 32: The docstring does not really explain what it means to be "funcallable" for this class.
Maybe give a use case?

Ambrevar:
Looks OK to me, but before merging, can you sum up what this does?  What introduce `user-mixin-class` in particular?

Ambrevar:
To clarify, I read your comment in https://github.com/atlas-engineer/nyxt/pull/2551#issuecomment-1246351571, but I'm not sure why this is beneficial compared to the previous architecture.

source/user-classes.lisp
aartaka: Yes `（*/∇＼*）`

aartaka:
As it appears, there's an uncloseable gap between the world of functions and world of values in CL. So one cannot simply easily make a class that's both `closer-mop:funcallable-standard-class` (the one all functions inherit from) and `standard-class` (the one all value classes inherit from). 

So, to make both function-like things and value-like things equally configurable, we need to make at least two separate classes: `user-class` (like `standard-class`, but configurable) and `user-funcallable-class` (like `closer-mop:funcallable-standard-class`, but configurable). But then, this "but configurable" part can be expressed as a separate mixing that's neither standard class nor funcallable class, thus allowing us to close more uncloseable gaps much more easier in the future. 

Imagine some implementation has structs totally separate from classes, and thus configuring structs may be possible, but would require adding yet another class. Say, `user-structure-class`. With `user-mixin-class`, that would be a one-liner, while retaining the `customize-hook` and other configuration functionality in three (`user-class`, `user-funcallable-class`, and now `user-structure-class`) would already be 3x code duplication...

Ambrevar:
Fair enough.  But then isn't `user-mixin-class` misleading?
It's a mixin, sure, but its role is to expose a customization hook.

That said, it's unexported, right?  So no big deal.

In any case, can you include this last bit of explanation to the docstring ?

aartaka:
> Fair enough. But then isn't `user-mixin-class` misleading? It's a mixin, sure, but its role is to expose a customization hook.

Why is it? Because it seems to be a class for some `user-mixin`s? Maybe call it `user-mixin` instead then?

> In any case, can you include this last bit of explanation to the docstring ?

Yes, I did now :)

Ambrevar:
> > 
> 
> Why is it? Because it seems to be a class for some `user-mixin`s? Maybe call it `user-mixin` instead then?

I mean the name says it's a mixin, but it says nothing about customization, which is the main purpose of this class.

aartaka:
> > Why is it? Because it seems to be a class for some `user-mixin`s? Maybe call it `user-mixin` instead then?
> 
> I mean the name says it's a mixin, but it says nothing about customization, which is the main purpose of this class.

Yes, because the name is an artifact of our `user-*` naming. `user-class` is a user-configurable class, `user-file` is a user-configurable file object, etc. So `user-mixin-class`, according to this, is a mixin for other user classes, which is exactly what it is...

I mean, it won't hurt renaming all those into `customizable-*`, but it won't help much either—those names are internal and relatively consistent already.

Ambrevar:
Fair enough.  Feel free to merge!

PR #2555, opened 2022-09-08
Selection actions list

### LINKED ISSUE #2198: Leverage `selection-actions` as a list
The documentation of `selection-actions` suggests that it could be a list, but there's no mechanism to handle it as a list of multiple elements.

Also, I can't think of an example where it could be used as a list. Anyone? @Ambrevar @aartaka @jmercouris @pdelfino

aadcg:
# Description

Add the ability to choose from a list of `selection-actions`. 
Simplify `return-actions` by ensuring it's a list.

Fixes #2198 

# Discussion

Also, I want to test this API with the hinting system. I can imagine it being very helpful in that context.

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -330,6 +330,9 @@ call."))
   (alex:when-let ((action (first (marks-actions source))))
     (run-thread "Prompter mark action thread" (funcall action (marks source)))))
 
+(defmethod default-selection-action ((source prompter:source))
+  (first (slot-value source 'selection-actions)))
+
 (export-always 'object-attributes)
 (defgeneric object-attributes (object source)
   (:method ((object t) (source prompter:source))
--- a/libraries/prompter/prompter.lisp
+++ b/libraries/prompter/prompter.lisp
@@ -167,13 +167,20 @@ compution is not finished.")))
 
 (defmethod (setf selection) (value (prompter prompter))
   (setf (slot-value prompter 'selection) value)
-  (let ((source (selected-source prompter)))
-    (when (selection-actions-enabled-p source)
-      (if (< 0 (selection-actions-delay source))
+  (call-selection-action prompter))
+
+(export-always 'call-selection-action)
+(defmethod call-selection-action ((prompter prompter))
+  (sera:and-let* ((source (selected-source prompter))
+                  (_ (selection-actions-enabled-p source))
+                  (action (default-selection-action source))
+                  (suggestion (selected-suggestion prompter)))
+    (let ((delay (selection-actions-delay source)))
+      (if (plusp delay)
           (run-thread "Prompter selection action thread"
-            (sleep (selection-actions-delay source))
-            (call-selection-action prompter))
-          (call-selection-action prompter)))))
+            (sleep delay)
+            (funcall action (value suggestion)))
+          (funcall action (value suggestion))))))
 
 (export-always 'input)
 (defmethod (setf input) (text (prompter prompter))
@@ -198,12 +205,6 @@ Signal destruction by sending a value to PROMPTER's `interrupt-channel'."
   ;; TODO: Interrupt before or after destructor?
   (calispel:! (sync-interrupt-channel (sync-queue prompter)) t)
   (calispel:! (interrupt-channel prompter) t))
-
-(export-always 'call-selection-action)
-(defun call-selection-action (prompter)
-  (sera:and-let* ((action (first (selection-actions (selected-source prompter))))
-                  (suggestion (selected-suggestion prompter)))
-    (funcall action (value suggestion))))
 
 (defun select (prompter steps &key wrap-over-p)
   "Select `suggestion' by jumping STEPS forward.
--- a/libraries/prompter/prompter.lisp
+++ b/libraries/prompter/prompter.lisp
@@ -182,6 +182,11 @@ compution is not finished.")))
             (funcall action (value suggestion)))
           (funcall action (value suggestion))))))
 
+(export-always 'set-selection-action)
+(defmethod set-selection-action (value (prompter prompter))
+  (setf (selection-actions (selected-source prompter))
+        (cons value (delete value (selection-actions (selected-source prompter))))))
+
 (export-always 'input)
 (defmethod (setf input) (text (prompter prompter))
   "Update PROMPTER sources and return TEXT."
--- a/source/mode/prompt-buffer.lisp
+++ b/source/mode/prompt-buffer.lisp
@@ -44,8 +44,9 @@ default)."
        "f1 b" 'run-prompt-buffer-command
        "f1 m" 'describe-prompt-buffer
        "return" 'return-selection
-       "M-return" 'return-selection-over-action
+       "M-return" 'return-marks-action
        "C-return" 'run-selection-action
+       "C-c C-j" 'set-selection-action
        "tab" 'insert-selection
        ; TODO: This is the Emacs Helm binding.  Better?
        "C-c C-f" 'toggle-selection-actions-enabled
@@ -305,6 +306,10 @@ current unmarked selection."
   (sera:and-let* ((first-prompt-buffer (first (nyxt::active-prompt-buffers window))))
     (prompter:return-actions first-prompt-buffer)))
 
+(defun prompt-buffer-selection-actions (&optional (window (current-window)))
+  (sera:and-let* ((first-prompt-buffer (first (nyxt::active-prompt-buffers window))))
+    (prompter:selection-actions (prompter:selected-source first-prompt-buffer))))
+
 ;; TODO: Should return-actions be commands?  For now, they can be either
 ;; commands or symbols.
 (defun make-action-suggestion (action &optional source input)
@@ -323,24 +328,34 @@ current unmarked selection."
                                                  (t (documentation action 'function)))))
                                        "")))))
 
-(define-class action-source (prompter:source)
+(define-class return-action-source (prompter:source)
   ((prompter:name "List of return-actions")
    (prompter:constructor (prompt-buffer-return-actions))
    (prompter:suggestion-maker 'make-action-suggestion)))
 
-(define-command-prompt return-selection-over-action (prompt-buffer)
-  "Prompt for an action to run over PROMPT-BUFFER selection."
+(define-class selection-action-source (prompter:source)
+  ((prompter:name "List of selection-actions")
+   (prompter:constructor (prompt-buffer-selection-actions))
+   (prompter:suggestion-maker 'make-action-suggestion)))
+
+(define-command-prompt return-marks-action (prompt-buffer)
+  "Prompt for an action to run over PROMPT-BUFFER `prompter:marks'."
   (if (equal (mapcar #'type-of (prompter:sources (current-prompt-buffer)))
-             '(action-source))
+             '(return-action-source))
       (echo "Already displaying return-actions of previous prompt buffer.")
-      (let ((action (prompt1 :prompt "Action to run on selection"
-                             :sources 'action-source)))
-        (when action
-          (prompter:return-selection prompt-buffer action)))))
+      (alex:when-let ((action (prompt1 :prompt "Action to run on selection"
+                                       :sources 'return-action-source)))
+        (prompter:return-selection prompt-buffer action))))
 
 (define-command-prompt run-selection-action (prompt-buffer)
-  "Run one of `prompter:selection-actions' without closing PROMPT-BUFFER."
+  "Run `prompter::default-selection-action' without closing PROMPT-BUFFER."
   (prompter:call-selection-action prompt-buffer))
+
+(define-command-prompt set-selection-action (prompt-buffer)
+  "Set `prompter:selection-actions' without closing PROMPT-BUFFER."
+  (alex:when-let ((action (prompt1 :prompt "Set selection action"
+                                   :sources 'selection-action-source)))
+    (prompter:set-selection-action action prompt-buffer)))
 
 (define-command-prompt cancel-input (prompt-buffer) ; TODO: Rename.
   "Close the PROMPT-BUFFER without further action."
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -485,7 +485,12 @@ to open a file, save it, switch buffer or delete current buffer.")
    (:li (:nxref :command 'nyxt/document-mode:paste-from-clipboard-ring) " is now conveniently bound to "
         (:code "M-y") " in Emacs scheme of "
         (:nxref :class-name 'nyxt/document-mode:document-mode) ".")
-   (:li "Prompt-buffer now has familiar bindings for text cutting."))
+   (:li "Prompt-buffer now has familiar bindings for text cutting.")
+   (:li "Add " (:nxref :command 'nyxt/prompt-buffer-mode:set-selection-action)
+        ", bound to " (:code "C-c C-j") "by default.")
+   (:li (:code "return-selection-over-action") " renamed to "
+        (:nxref :command 'nyxt/prompt-buffer-mode:return-marks-action)
+        ".  The default keybinding is the same."))
 
   (:h3 "Programming interface")
   (:ul
--- a/source/tutorial.lisp
+++ b/source/tutorial.lisp
@@ -63,9 +63,9 @@ matching your input as you type.")
                                :modes (list (make-instance 'nyxt/prompt-buffer-mode:prompt-buffer-mode)))
                ": Validate the selected suggestion(s) or the current input if there is
 no suggestion.")
-         (:li  (command-markup 'nyxt/prompt-buffer-mode:return-selection-over-action
+         (:li  (command-markup 'nyxt/prompt-buffer-mode:return-marks-action
                                :modes (list (make-instance 'nyxt/prompt-buffer-mode:prompt-buffer-mode)))
-               ": Query the user for an action to run over the selected suggestion(s)."))
+               ": Query the user for an action to run over the marked suggestion(s)."))
         (:p " Some commands support multiple selections, for
 instance " (:code "delete-buffer") " can delete all selected buffers at once.
 When the input is changed and the suggestions are re-filtered, the selection is


libraries/prompter/prompter.lisp
```diff
@@ -470,6 +471,7 @@ After timeout has elapsed for one source, return nil."
   "Return prompter object."
   (apply #'make-instance 'prompter args))
 
+;; FIXME this seems be unreliable when there's a single source.  Confirm.
```
Ambrevar, line 474: Can you share a recipe?

Ambrevar:
You might have misunderstood the point of `C-j`: it's to run a selection action _manually_, in case follow-mode is not enabled.

So you'd need to restore `call-selection-action` since otherwise the user has no way to manually call an action.

Rest looks good to me, but I haven't tested yet.

libraries/prompter/prompter.lisp
aadcg: I was wrong. All good.

aadcg:
> You might have misunderstood the point of C-j: it's to run a selection action manually, in case follow-mode is not enabled.

For the sake of clarity, `follow-mode` is a deprecated concept and we now call it `auto-return-p`.

I don't follow.  Notice that `auto-return-p` only concerns `return-actions`, not `selection-actions`.

> So you'd need to restore call-selection-action since otherwise the user has no way to manually call an action.

Which actions are you referring to? We have `return-actions`, `selection-actions` and `marks-actions`.

Can we give an example?

aadcg:
Also, on master, running `run-selection-action` (bound to `C-j` in Emacs-mode) has no effect.

Ambrevar:
> For the sake of clarity, `follow-mode` is a deprecated concept and we now call it `auto-return-p`.
> 
> I don't follow. Notice that `auto-return-p` only concerns `return-actions`, not `selection-actions`.

That's your confusion: `follow-mode` has been renamed to `selection-actions-enabled-p`.  

Try the `jump-to-heading` command and navigate the selections to see its effect.

libraries/prompter/prompter.lisp
```diff
@@ -162,13 +162,20 @@ compution is not finished.")))
...
+                  (action (default-selection-action source))
+                  (suggestion (selected-suggestion prompter)))
+    (run-thread "Prompter selection action thread"
+      (when (plusp delay) (sleep delay))
+      (funcall action suggestion))))
+
+(export-always 'set-selection-action)
+(defmethod set-selection-action (value (prompter prompter))
+  (let ((actions (mapcar #'selection-actions (sources prompter))))
+    (setf actions (delete value actions))
```
aartaka, line 177: I wonder if there's an idiom for this somewhere...

source/browser.lisp
```diff
@@ -569,7 +570,7 @@ set of useful URLs or preparing a list to send to a someone else."
                   :sources (make-instance 'buffer-source
                                           :constructor (remove-if #'internal-url-p (buffer-list)
                                                                   :key #'url)
-                                          :return-actions '(identity)
```
aartaka, line 572: But it's okay as a quoted list of one symbol, isn't it?

source/history.lisp
```diff
@@ -122,7 +122,7 @@ then become available for deletion with `delete-history-entry'."
   (let ((buffers (or (alex:ensure-list buffer)
                      (prompt :prompt "Reset histories of buffer(s)"
                              :sources (make-instance 'buffer-source
-                                                     :return-actions '())))))
+                                                     :return-actions nil)))))
```
aartaka, line 125: Should it rather be an `#'identity` for the sake of clarity?

source/mode/prompt-buffer.lisp
```diff
@@ -44,8 +44,8 @@ default)."
        "f1 b" 'run-prompt-buffer-command
        "f1 m" 'describe-prompt-buffer
        "return" 'return-selection
-       "M-return" 'return-selection-over-action
-       "C-return" 'run-selection-action
+       "M-return" 'return-marks-action
+       "C-return" 'set-selection-action
```
aartaka, line 48: Maybe use some other binding? `C-return` is usually used to mark something like forced send or definitive approval. On GitHub, it's used to send commentaries (because mere `return` is used to insert newline), for example.

source/mode/prompt-buffer.lisp
```diff
@@ -91,7 +91,7 @@ default)."
        "C-y" 'paste
        "C-w" 'cut
        "C-h b" 'run-prompt-buffer-command
-       "C-j" 'run-selection-action)
```
aartaka, line 94: Hait, can we disable selection actions without redefinining a class? Can we do one-shot selection actions if they are disabled? I'm anticipating that those use-cases can come in handy more than once there...

source/mode/repeat.lisp
```diff
@@ -28,9 +28,9 @@
     :type (or integer null)
     :documentation "The number of times to repeat the commands for.")
    (repeat-interval
-    1
-    :type number
-    :documentation "Time interval (in seconds) to repeat `repeat-action'.")
+    1.0
+    :type float
```
aartaka, line 32: This one looks fishy. Note that we can set this slot to an integer, and this can cause type errors, because integers are disjoint from floats. Maybe use `rational` then? It's more granular than `number`, but still general enough to include both floats and integers.

aartaka (commented):
I tend to agree with @Ambrevar—we _need_ manual escape hatch in the form of `(call|run)-selection-action`.

aadcg:
> That's your confusion: `follow-mode` has been renamed to `selection-actions-enabled-p`.

You're correct!  Sorry.

Now let's go back to the main topic that you have raised:

> You might have misunderstood the point of C-j: it's to run a selection action manually, in case follow-mode is not enabled.

> So you'd need to restore call-selection-action since otherwise the user has no way to manually call an action.

`selection-actions-enabled-p` is nil by default. Some modes override it of course. 

We have the command `toggle-selection-actions-enabled` that toggles the this option. Now that `selection-actions` can be manifold, we need an extra command to select the default one. 

Does it really make sense to have a command that enables a `selection-action` to run once?

libraries/prompter/prompter.lisp
aadcg: I wondered too! Open to suggestion :)

source/browser.lisp
aadcg: It's ok indeed. No need for the list, and I'd say it's always better to pass the function instead of the function symbol.

source/history.lisp
aadcg: It's not exactly equivalent, but I don't see why not!

aadcg: See fc4f9ed1d.

source/mode/prompt-buffer.lisp
aadcg: My argument is that we have `toggle-selection-actions-enabled`, but you'll say that we loose one-shot selection actions.

source/mode/repeat.lisp
aadcg: Very good observation.

aadcg: See ddfa96ccc.

aadcg: Actually, notice that `(rationalp 0.0)` returns `nil`. So perhaps we should type it as a `number`?

aadcg: Or rather, we should type it as `real`.

aadcg: Or rather, we should type it as `alex:positive-real`.

Ambrevar:
> Now that `selection-actions` can be manifold, we need an extra command to select the default action (`set-selection-action`).

Correct!

> Does it really make sense to have a command that enables a `selection-action` to run once?

Can you give an example?

aadcg:
>  Can you give an example?

Ok, let's take this slowly. It seems that both @Ambrevar and @aartaka raised concern about the fact that I've deprecated the command `run-selection-action`. I'll explain my reasoning below.

Let's take the command `jump-to-heading`, for which `selection-actions-enabled-p` is set to `t`. 

Let me describe the behavior on master. When we issue the command interactively, it's possible to disable `selection-actions` (scrolling in this case) by issuing the command `toggle-selection-actions-enabled` while the prompt is active. Now that `selection-actions-enabled-p` is set to `nil`, the command `run-selection-action` runs the action once. Personally, I think that `toggle-selection-actions-enabled`, while different, is general enough for me not to use `run-selection-action`.

With this PR, I add the possibility to define several `selection-actions`. Obviously a new command is needed to set the action, without leaving the prompt.  Given my comment above on the generality of `toggle-selection-actions-enabled`, and the fact that a new command is needed, I thought it would be sensible to bind this new command to the key that was once bound to `run-selection-action`. 

If you consider that we really need 3 commands for cover all of this, I don't oppose. I just think it may be overkill. Perhaps there are use cases that I'm actually disregarding!

Ambrevar:
OK, makes sense now, thanks for the detailed review.

So yes, I believe we need 3 actions, because selection action may _need_ to be only manual, in which case we need a trigger key.  `jump-to-heading` only scrolls, so this is rather innocent, but imagine a selection action that, say, downloads the hint!  In this case you want to be careful and not accidentally trigger the selection action.

aartaka:
> So yes, I believe we need 3 commands, because selection action may _need_ to be only manual, in which case we need a trigger key. 

Yes!

> `jump-to-heading` only scrolls, so this is rather innocent, but imagine a selection action that, say, downloads the hint! In this case you want to be careful and not accidentally trigger the selection action.

`jump-to-heading` is not innocent—it throws you to the start of the document once you start selecting prompt suggestions! I even think that we should make it's `default-selection-action` purely manual, because we don't yet have a good way to make it smooth ヽ(ｏ`皿′ｏ)ﾉ

aadcg:
Commit f96d8ba54 introduced a major bug in the hinting interface. At least the one who did it will clean up the mess! Sorry everyone.

source/mode/prompt-buffer.lisp
aadcg: I've revived this functionality.

source/mode/prompt-buffer.lisp
aadcg: @aartaka please review the keybindings.

libraries/password-manager/password.lisp
```diff
@@ -4,11 +4,14 @@
...
-               :documentation "The program to query for password information.")
-   (sleep-timer 15
-                :documentation "The amount of time to sleep, in seconds."))
+  ((executable
+    nil
+    :type (or null string)
+    :documentation "The program to query for password information.")
+   (sleep-timer
+    15.0
+    :type alexandria:non-negative-real
```
aartaka, line 13: Why change it to float? For `alexandria:...` type? If so, this might've been an `unsigned-byte` instead.

libraries/prompter/prompter-source.lisp
```diff
@@ -209,13 +209,17 @@ recompute the match-data for instance.")
 The predicate works the same as the `sort' predicate.")
 
    (return-actions
-    '(identity)
-    :type list
+    #'identity
+    :type (or null
+              (or function function-symbol)
+              (cons (or function function-symbol) *))
```
aartaka, line 215: We do need `list-of` here, don't we...

source/command-commands.lisp
```diff
@@ -139,29 +139,25 @@ together with the arglists and documentations of the functions typed in."
...
+                                           (echo "~s" (eval (first exprs)))))))
                     (make-instance
                      'predicted-command-source
-                     :return-actions
-                     (list (lambda-command run-command* (commands)
-                             "Run the chosen command."
-                             (let ((command (first commands)))
-                               (setf (last-access command) (time:now))
-                               (run-async command))))))
+                     :return-actions (lambda-command run-command* (commands)
```
aartaka, line 156: Indentation note: I'd put the action itself on a next line here, because the huge indentation jump looks worse than the less predictable (but still linear) reading order.

source/mode.lisp
```diff
@@ -376,7 +376,7 @@ If it's a single buffer, return it directly (not as a list)."
                          :prompt "Enable mode(s) for buffer(s)"
                          :sources (make-instance 'buffer-source
                                                  :multi-selection-p t
-                                                 :return-actions '())))))
+                                                 :return-actions #'identity)))))
```
aartaka, line 379: Does `nil` no longer work? It's a nice shortcut (although quite opaque, I have to admit).

source/mode/autofill.lisp
```diff
@@ -62,9 +62,9 @@ it will be in conflict with common-lisp:fill."))
 (define-class autofill-source (prompter:source)
   ((prompter:name "Autofills")
    (prompter:constructor (autofills (find-submode 'autofill-mode)))
-   (prompter:return-actions
-    (list (lambda-command autofill* (autofills)
-            (ffi-buffer-paste (current-buffer) (funcall (first autofills)))))))
+   (prompter:return-actions (lambda-command autofill* (autofills)
```
aartaka, line 65: Here the initform is too deeply indented, which is not aesthetic enough, I believe. Maybe move it below the slot name?

aartaka (approved):
Looks just fine to me, except some aesthetic clashes :P Feel free to ignore those!

libraries/password-manager/password.lisp
aadcg: I don't follow.

libraries/prompter/prompter-source.lisp
aadcg: Agree. 9ae3fe82d

aartaka: B-b-but wait, `list-of` is not defined for `prompter`, isn't it?

source/command-commands.lisp
aadcg: Which one do you prefer?

```lisp
(make-instance 'extended-command-source
               :return-actions
               (lambda-command evaluate-lisp-expression* (exprs)
                 "Evaluate the inputted Lisp expression."
                 (run-thread "evaluator"
                   (let ((*interactive-p* t))
                     (echo "~s" (eval (first exprs)))))))
```

```lisp
(make-instance
 'extended-command-source
 :return-actions
 (lambda-command evaluate-lisp-expression* (exprs)
   "Evaluate the inputted Lisp expression."
   (run-thread "evaluator"
     (let ((*interactive-p* t))
       (echo "~s" (eval (first exprs)))))))
```

source/mode.lisp
aadcg: [This was actually a suggestion of yours, if I understood it correctly](https://github.com/atlas-engineer/nyxt/pull/2555#discussion_r969306480).

source/command-commands.lisp
aartaka: Second, for sure. Less indentation, less eye movement.

libraries/password-manager/password.lisp
aartaka: The initform was `15`, not `15.0`. Was there any particular reason for a change in type?

source/mode/autofill.lisp
aadcg: I never move to a new line (unless I go beyond 80 chars per line) to avoid separating OPERATOR from ARGS in `(OPERATOR ARGS)`. I find it easier to read.

aadcg:
With regards to testing, actions are not tested on the `prompter` library. It could be added but I think it's outside the scope of this PR.

I'll update the documentation on the manual and on the `prompter`'s README.

source/mode/autofill.lisp
aartaka: Yeah, it's all personal aesthetics :)

libraries/password-manager/password.lisp
aadcg: Oh, good point. Thanks.

libraries/prompter/prompter-source.lisp
aadcg: It is not, I got carried away again...

You were commenting that it'd be nice to have the niceties of `types.lisp` available to the libraries right?

source/command-commands.lisp
aadcg: Interesting. Let's go with the second then.

libraries/prompter/prompter-source.lisp
aartaka: Yes. Maybe as a library or a contrib to e.g. `trivial-types`.

aadcg: In other words, there could be a `ntypes` library.

aadcg:
[I've added a note on things that are left to be done.](https://github.com/atlas-engineer/nyxt/issues/1213#issuecomment-1297725577)

@Ambrevar, could you have a final look? Should be almost ready to be merged.

libraries/prompter/prompter.lisp
```diff
@@ -167,13 +166,24 @@ compution is not finished.")))
...
+(export-always 'call-selection-action)
+(defmethod call-selection-action ((prompter prompter))
+  (sera:and-let* ((source (selected-source prompter))
+                  (_ (selection-actions-enabled-p source))
+                  (delay (selection-actions-delay source))
+                  (action (default-selection-action source))
+                  (suggestion (selected-suggestion prompter)))
+    (run-thread "Prompter selection action thread"
+      (when delay (sleep delay))
+      (funcall action (value suggestion)))))
```
Ambrevar, line 180: This seems wrong: you should not start a thread if there is no need for it.

Or is there any other reason you decided to always start a thread?

libraries/prompter/prompter.lisp
```diff
@@ -439,46 +436,40 @@ TIMEOUT is deprecated."
...
-  "Return prompter object.
+  "Return `prompter' object.
 The arguments are the initargs of the `prompter' class.
 
 As a special case, the `:sources' keyword argument not only accepts `source'
-objects but also symbols.  Example:
+objects but also symbols.
 
-  (prompter:make :sources 'prompter:raw-source)"
+Example: (prompter:make :sources 'prompter:raw-source)"
```
Ambrevar, line 472: Style: Leave examples on their own lines for consistency and to help the user with selecting / copy-pasting the example.

In the future we might be able to highlight such examples by deteting lines that start with `[[:space:]]*(`.

source/mode/prompt-buffer.lisp
```diff
@@ -44,8 +44,9 @@ default)."
        "f1 b" 'run-prompt-buffer-command
        "f1 m" 'describe-prompt-buffer
        "return" 'return-selection
-       "M-return" 'return-selection-over-action
+       "shift-return" 'return-marks-action
```
Ambrevar, line 47: Why change this?

libraries/prompter/prompter.lisp
```diff
@@ -167,13 +166,24 @@ compution is not finished.")))
...
+                  (suggestion (selected-suggestion prompter)))
+    (run-thread "Prompter selection action thread"
+      (when delay (sleep delay))
+      (funcall action (value suggestion)))))
+
+(export-always 'set-selection-action)
+(defmethod set-selection-action (value (prompter prompter))
+  (setf (selection-actions (selected-source prompter))
+        (delete value (selection-actions (selected-source prompter))))
+  (push value (selection-actions (selected-source prompter))))
```
Ambrevar, line 186: It's better (and safer) to do this in one go:

```lisp
(setf ... (cons value (delete value ...)))
```

Ambrevar:
The fix for #2198 looks good and I'm happy with that, thanks!

However this pull request does too many things and I'm confused about the number of things that have changed.

- It's good practice to not include style / code refactoring together with important redesign changes.
- Same with refactoring / renaming of functions / symbols.  If it's far reaching, it's better to keep it for a separate pull request.  At the moment it's hard to have a good overview of what this pull request initially intended, namely to fix #2198.
- Don't forget to update the changelog, in particular about the addition / remove of commands and the change of keybindings.
- Why did you decide to change the `M-return` keybinding?

I'd suggest to close this pull request unmerged, and open 2 or 3 other request: 
- first one that fixes #2198 (can be quickly merged),
- then one that fixes indentation, typos, etc. (also a quick merge), 
- and finally one that refactors the code.

Thoughts?

libraries/prompter/prompter.lisp
aadcg: I believe you're addressing this question to yourself, since I have simply kept everything as it was :P

See your commit c13309a4e882f7df8ee018e3bb037ca57d977b49.

libraries/prompter/prompter.lisp
aadcg: Good point.

source/mode/prompt-buffer.lisp
aadcg: This was discussed in a call with @aartaka if I recall correctly. It seems we have not yet finished discussing keybindings...

@aartaka am I making sense? Do you remember your suggestions? I remember you had a very clear picture.

libraries/prompter/prompter.lisp
Ambrevar: Misunderstanding.  Previously the code was:

```lisp
(if (< 0 (selection-actions-delay source))
          (run-thread "Prompter selection action thread"
            (sleep (selection-actions-delay source))
            (call-selection-action prompter))
          (call-selection-action prompter))
```

Notice that `call-selection-action` is  called in the _current thread_ when there is no delay.
You've changed this logic.

source/mode/prompt-buffer.lisp
Ambrevar: Alright.  You can write a rationale then (in a commit) and then add the changelog entry.  

Or if you think you are not done discussing it, then revert the change and keep if for later.

aadcg:
You have a point, even though it seems to be too much work for little to no benefit. Let's stick to what you suggest :)

Ambrevar:
Sorry for being overly picky, but the prompter library is the most sensitive part of Nyxt, so we should try to be as clean as possible.  It will help us a lot in the future if things go wrong.

aadcg:
No need to close this PR, as I've been breaking the former one into smaller ones.

About keybindings: 

We need a new keybinding for `set-selection-action`. Does `C-c C-j` make sense? @aartaka @Ambrevar @jmercouris

source/mode/prompt-buffer.lisp
aadcg: I've reverted this change from `M-return` to `shift-return`. @aartaka please open an issue if you think our default could be improved.

Ambrevar:
`C-c C-j` is fine from a Helm user perspective, but maybe a bit arcane for everyone else.

What about `C-c return`?  Or `C-u return`?

libraries/prompter/prompter-source.lisp
```diff
@@ -330,6 +330,9 @@ call."))
   (alex:when-let ((action (first (marks-actions source))))
     (run-thread "Prompter mark action thread" (funcall action (marks source)))))
 
+(defmethod default-selection-action ((source prompter:source))
+  (first (slot-value source 'selection-actions)))
```
Ambrevar, line 334: Export?

aartaka:
> * Why did you decide to change the `M-return` keybinding?

Because `M-return` should be reserved for some "strong" action closing the prompt-buffer, analogous to `return` and `C-return` (`return` stands either for "activate this now" or "I'm done, commit this action" in CUA). If we dilute the responsibility one takes when pressing `M-return`, then the character of other actions bound to `*-return` will not be perceived as final and decisive for a given prompt.

aadcg:
> `C-c C-j` is fine from a Helm user perspective, but maybe a bit arcane for everyone else.
> 
> What about `C-c return`? Or `C-u return`?

I guess I still prefer `C-c C-j`, but I don't have any strong opinion.

libraries/prompter/prompter.lisp
aadcg: Ups, thanks. Will address now.

aadcg: See 241c26ee7.

libraries/prompter/prompter-source.lisp
aadcg: I thought the same. The reason why I refrained myself from exporting is due to `default-return-action`, which isn't exported. I don't see any reason why these shouldn't be consistent.

aadcg:
Ok to merge? @Ambrevar

aadcg:
Updated changelog, should be ready to merge. @Ambrevar.

Ambrevar:
Looks good to me!

aadcg:
Thanks everyone!

PR #2557, opened 2022-09-11
* describe: sort slot/method/package lists where sensible

### LINKED ISSUE #2552: Sorted lists in describe-* pages
Describe-* pages, especially for classes, have long lists of subclasses/slots/methods etc.

These would be way more readable/useful if they were sorted alphabetically.

This looks easy to implement and I would be happy to make a PR if you thought it would be useful.

franburstall:
# Description

Sort lists of slots/packages/methods/subclasses in `describe-*` pages where it seems sensible.

Fixes #2552

# Discussion

This is a minimal solution to the problem.  One could do more.  Example: `values->html` would probably benefit from something similar.  

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/describe.lisp
+++ b/source/describe.lisp
@@ -227,11 +227,11 @@ turned into links to their respective description page."
          (:li "Total: " (length total-symbols)))
         (:h2 "Use list:")
         (:ul
-         (dolist (use (package-use-list package))
+         (dolist (use (safe-sort (package-use-list package) :key #'package-name))
            (:li (package-markup use))))
         (:h2 "Used by list:")
         (:ul
-         (dolist (use (package-used-by-list package))
+         (dolist (use (safe-sort (package-used-by-list package) :key #'package-name))
            (:li (package-markup use))))))))
 
 (define-internal-page-command-global describe-variable
@@ -482,7 +482,7 @@ A command is a special kind of function that can be called with
               :sources (make-instance 'class-source :universal universal))))
     (buffer (str:concat "*Help-" (symbol-name class) "*") (resolve-symbol :help-mode :mode))
   "Inspect a class and show it in a help buffer."
-  (let* ((slots (class-public-slots class))
+  (let* ((slots (safe-sort (class-public-slots class)))
          (slot-descs (sera:string-join (mapcar (rcurry #'describe-slot* class) slots) ""))
          (*print-case* :downcase))
     (spinneret:with-html-string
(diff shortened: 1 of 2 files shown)


source/describe.lisp
```diff
@@ -203,6 +203,10 @@ turned into links to their respective description page."
            (:code (:raw (ppcre:regex-replace-all "`[^'\\s]+'" string #'resolve-regex)))))
         "")))
 
+(defun safe-sort (s &key (key #'string))
```
aartaka, line 206: Note that there's `sera:sort-new` that does most of the work here.

Also note that in the cases where you use `remove*` and `map*` functions, a new list is cosed up, so `sort` is safe to use on it.

franburstall: TIL about `sera:sort-new`!  I have just pushed a version of `safe-sort` that uses it.

I agree that `sort` is safe to use some of the time (but in all the cases to which `safe-sort` is applied).  In the interests of reducing verbosity, I used the one function throughout.

source/describe.lisp
```diff
@@ -205,7 +205,7 @@ turned into links to their respective description page."
 
 (defun safe-sort (s &key (key #'string))
   "Sort sequence S of objects alphabetically by KEY."
-  (sort (copy-list s) #'string-lessp :key key))
+  (coerce (sera:sort-new s #'string-lessp :key key) 'list))
```
aartaka, line 208: I'd rather `coerce` it to the `(type-of s)` and propagate the predicate function as `&key (predicate #'string-lessp)` argument, so that it's more universally applicable. Then it can be put alongside `safe-read` in utilities.lisp.

franburstall: Good idea.  Also having it in utilities.lisp makes it available to mess with `value->html` at some point.  Will implement.

source/utilities.lisp
```diff
@@ -142,6 +142,12 @@ This is useful if you do not trust the input."
     (uiop:with-safe-io-syntax (:package package)
       (read input-stream eof-error-p eof-value recursive-p))))
 
+(export-always 'safe-sort)
+(defun safe-sort (s &key (predicate #'string-lessp) (key #'string))
+  "Sort sequence S of objects by KEY using PREDICATE."
+  (coerce (sera:sort-new s predicate :key key) (type-of s)))
```
Ambrevar, line 148: Isn't this redundant?  `sera:sort-new` works on list already.
In other words, the `coerce` is not needed I believe.

source/describe.lisp
```diff
@@ -203,6 +203,7 @@ turned into links to their respective description page."
            (:code (:raw (ppcre:regex-replace-all "`[^'\\s]+'" string #'resolve-regex)))))
         "")))
 
+
```
Ambrevar, line 206: Make sure you don't introduce new lines unless you are reformating / styling the code.

franburstall: Will fix.

source/utilities.lisp
franburstall: The point is the `sera-sort-new` returns a vector and that is what we are coercing back.

Ambrevar: Oh right, I misread the output :)

Ambrevar: Then don't use Serapeum if you must return a list, because you'll be allocating a vector for nothing.  Instead just call `sort` on `copy-list`.

Ambrevar:
I believe you can squash the commits.

Note: No need to include the leading `*` in your commit messages.

source/utilities.lisp
franburstall: > Then don't use Serapeum if you must return a list, because you'll be allocating a vector for nothing. Instead just call `sort` on `copy-list`.

That is what I originally did!

However, there is some virtue to using `sera:sort-new` and `coerce`:
1. The vector is not allocated entirely in vain because it is more efficient to sort vectors.
2. Coerce guarantees we get back the same type of sequence as we started with in line with the Principle of Least Surprise.  In the case at hand, the value gets fed to `loop` sometimes so this matters.
3. The current implementation works on any kind of sequence.

That said, maybe simplest to call `sort` on `copy-seq`!

franburstall:
1. Commits squashed
2. White-space fixed
3. I went with `sort` on `copy-seq`: turns out to be 50% faster with only 10% of the consing compared with `sera:sort-new`.

jmercouris:
Looks great, thank you!

PR #2570, opened 2022-09-14
Log to file

### LINKED ISSUE #2449: Enable logging by default
Currently logging is not saved anywhere by default, which does not help with error reporting when run from the binary xD

Let's enable it, possibly saving to `.local/share/nyxt/`.

Ambrevar:
# Description

Log to file.  We use a global variable because we must start logging at the very beginning, before `*browser*` is instantiated.

Fixes #2449.

# Discussion

Should we log `*standard-output*` and `*error-output*`?

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/start.lisp
+++ b/source/start.lisp
@@ -445,7 +445,7 @@ Examples:
         (progn
           (log:config :debug)
           (format t "Arguments parsed: ~a and ~a~&" options urls))
-        (log:config :pattern "<%p> [%D{%H:%M:%S}] %m%n"))
+        (log:config :pattern *log-pattern*))
 
     ;; TODO: Deprecated, remove in 4.0.
     (when (and init (not config))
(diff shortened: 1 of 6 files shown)


source/start.lisp
```diff
@@ -516,6 +516,11 @@ Then load `*config-file*' if any.
 Instantiate `*browser*'.
 Finally, run the browser, load URL-STRINGS if any, then run
 `*after-init-hook*'."
+  (let ((log-path (files:expand *log-file*)))
+    (unless (files:nil-pathname-p log-path)
+      (uiop:delete-file-if-exists log-path) ; Otherwise `log4cl' appends.
+      ;; REVIEW: Do we want to back up the log?
```
aartaka, line 522: I'd say yes, just in case.

aartaka: But then it's better to use a separate directory for logs.

aadcg:
In the PR description it says it solves the wrong issue.

source/start.lisp
Ambrevar: This raises a bunch of questions though: what do we do about infinite disk usage?  Can we delete old files?  How?

It does not seem that log4cl has any provision for this.

Ambrevar:
Good catch André!

source/start.lisp
aartaka: Well, logs are plaintext and are intended for human viewing, so log4cl doesn't care about deleting those. If the human needs to delete the files—they'll do that manually.

Hmmmm. But then, maybe we don't need logs beyond the current session ones?

Ambrevar: Most high-level, user-facing programs don't have a rolling log.

Rolling logs make sense for daemon / servers.  For user applications, they may be overkill.

aartaka: Agreed then!

Ambrevar:
Merged with bc98ad026d90e9ff9bf42cedc6082cc1ed4b09eb.

PR #2572, opened 2022-09-14
Bulk spelling & indentation fixes

hendursaga:
Lots of small spelling fixes, removing Britishisms, and re-indenting those files while connected to a running instance.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -635,9 +635,10 @@ store them somewhere and `ffi-buffer-delete' them once done."))
   (:metaclass user-class))
 
 (define-class status-buffer (input-buffer)
-  ((window nil
-           :type (maybe window)
-           :documentation "The `window' to which the status buffer is attached.")
+  ((window
+    nil
+    :type (maybe window)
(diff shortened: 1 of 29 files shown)


source/buffer.lisp
```diff
@@ -636,8 +636,8 @@ store them somewhere and `ffi-buffer-delete' them once done."))
 
 (define-class status-buffer (input-buffer)
   ((window nil
-    :type (maybe window)
-    :documentation "The `window' to which the status buffer is attached.")
+           :type (maybe window)
+           :documentation "The `window' to which the status buffer is attached.")
```
Ambrevar, line 640: in this case we should insert a newline before the initform `nil` instead.

Ambrevar (commented):
Excellent!  Just one nit and we can merge!

source/buffer.lisp
hendursaga: Fixed! You can (optionally) squash and then merge!

Ambrevar:
Merged with ccfdbfafb6d93fe550cfec88fc603ddc1ad73c29.  Thanks a lot for this painful fix! :)

PR #2574, opened 2022-09-15
mode/hint: Remove deduplication logic.

### LINKED ISSUE #2516: Duplicate Hints
**Describe the bug**

When I run something like `download-hint-url`, an `` tag enclosed within an `<a>` tag (and presumably other combos) downloads the exact same URL twice. This makes sense since the selector for `download-hint-url` is `"a, img"`.

1. Visit https://google.com/
2. Run `download-hint-url` and select the Google Easter egg.
3. Run `list-downloads` and see for yourself.

Alternatively, visit some 'booru and try downloading any of the images there.

aadcg:
It's too expensive and sacrifices the interactive nature of the prompt buffer.

For a page with 255 hints, it takes 1.5 seconds for the deduplication to finish on a reasonable machine.

Notice that pages can have more than 2000 hints.

Fixes #2516

# Description

Fixes the issue I've reported [in this comment](https://github.com/atlas-engineer/nyxt/pull/2555#issuecomment-1246895114).

Lessons:
- @aartaka we need to be more cautions when trying to close an issue in such a hasty manner (see 6196f12a9). This fix didn't take into account the frequent case when there is no URL.
- @Ambrevar noticed that commit 6196f12a9 was short-sighted, but didn't check how much overhead it introduced in terms of responsiveness with his fix (2dbf2ab9f).
- @aadcg made two mistakes. Firstly, he didn't make an in-depth analysis of commit 2dbf2ab9f. That is, he just believed that it was properly working. Secondly, he didn't take into account that the preprocessing runs multiple times, when it should only be triggered once. 

# Discussion

Let's re-open #2516 and leave it for post-3.0. Sounds reasonable?

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/source/mode/hint.lisp
+++ b/source/mode/hint.lisp
@@ -296,8 +296,14 @@ FUNCTION is the action to perform on the selected elements."
                 :hide-suggestion-count-p (fit-to-prompt-p (find-submode 'hint-mode))
                 :sources (make-instance 'hint-source
                                         :multi-selection-p multi-selection-p
-                                        :constructor (lambda (source) (declare (ignore source))
-                                                       (add-hints :selector selector)))
+                                        :constructor
+                                        (lambda (source) (declare (ignore source))
+                                          (delete-duplicates
+                                           (add-hints :selector selector)
+                                           :test (lambda (h1 h2)
+                                                   (let ((h1 (plump:attribute h1 "href"))
+                                                         (h2 (plump:attribute h2 "href")))
+                                                     (and h1 h2 (string= h1 h2)))))))
                 :after-destructor (lambda () (with-current-buffer buffer (remove-hints))))))
     (funcall function result)))
 
--- a/source/mode/hint.lisp
+++ b/source/mode/hint.lisp
@@ -297,13 +297,19 @@ FUNCTION is the action to perform on the selected elements."
                 :sources (make-instance 'hint-source
                                         :multi-selection-p multi-selection-p
                                         :constructor
-                                        (lambda (source) (declare (ignore source))
-                                          (delete-duplicates
-                                           (add-hints :selector selector)
-                                           :test (lambda (h1 h2)
-                                                   (let ((h1 (plump:attribute h1 "href"))
-                                                         (h2 (plump:attribute h2 "href")))
-                                                     (and h1 h2 (string= h1 h2)))))))
+                                        (lambda (source)
+                                          (declare (ignore source))
+                                          (multiple-value-bind (no-url-hints url-hints)
+                                              (sera:partition #'null
+                                                              (add-hints :selector selector)
+                                                              :key (alex:rcurry #'plump:attribute "href"))
+                                            (append no-url-hints
+                                                    (delete-duplicates
+                                                     url-hints
+                                                     :test (lambda (h1 h2)
+                                                             (quri:uri-equal h1 h2))
+                                                     :key (compose (alex:rcurry #'quri:merge-uris (url buffer))
+                                                                   (alex:rcurry #'plump:attribute "href")))))))
                 :after-destructor (lambda () (with-current-buffer buffer (remove-hints))))))
     (funcall function result)))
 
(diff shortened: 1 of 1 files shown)


Ambrevar:
Actually... Why is it in the preprocessor?  It should be in the constructor, this way it's computed only once.

Ambrevar:
Maybe this could be optimized by using `quri:render-uri` as a key and `string=` as a test.

aadcg:
> Actually... Why is it in the preprocessor? It should be in the constructor, this way it's computed only once.

Because at the "constructor-time", you're dealing with `plump` objects, not with `prompter:suggestion` objects.

aadcg:
Let me try something.

Ambrevar:
> Because at "constructor-time", you're dealing with `plump` objects, not with `prompter:suggestion` objects.

That's fine, you can remove the duplicates in the PLUMP objects.

aadcg:
They way I've implemented now I get 0.7 seconds on a page with 2000 hints to deduplicate them.  That's much better.

Ambrevar:
Neat!

source/mode/hint.lisp
```diff
@@ -303,8 +296,14 @@ FUNCTION is the action to perform on the selected elements."
                 :hide-suggestion-count-p (fit-to-prompt-p (find-submode 'hint-mode))
                 :sources (make-instance 'hint-source
                                         :multi-selection-p multi-selection-p
-                                        :constructor (lambda (source) (declare (ignore source))
-                                                       (add-hints :selector selector)))
+                                        :constructor
+                                        (lambda (source) (declare (ignore source))
```
Ambrevar, line 300: Conventionally we keep the `(declare ...)` statements on their own line.

aadcg: Oh, sure. Lack of attention!

aadcg:
Look at all of my commit since they show all of my attempts. Commit 6229001e6 is the most ambitious attempt, with the aim or making it correct, but it's slow.

The final version is a compromise among all things considered. 

Any ideas you'd like to explore or shall I squash and merge? @Ambrevar @aartaka

source/mode/hint.lisp
```diff
@@ -303,8 +296,13 @@ FUNCTION is the action to perform on the selected elements."
...
                                         :multi-selection-p multi-selection-p
-                                        :constructor (lambda (source) (declare (ignore source))
-                                                       (add-hints :selector selector)))
+                                        :constructor
+                                        (lambda (source)
+                                          (declare (ignore source))
+                                          (delete-duplicates
+                                           (add-hints :selector selector)
+                                           :test (lambda (h1 h2) (and h1 h2 (string= h1 h2)))
+                                           :key (alex:rcurry #'plump:attribute "href"))))
```
aartaka, line 305: Note that `rcurry` (like `curry`) should be imported into all the Nyxt-based packages, so you can use it without a package prefix (★^O^★)

aadcg: Is that a shameless plug? :)

aartaka:
> Any ideas you'd like to explore or shall I squash and merge? @Ambrevar @aartaka

What I'd like to try is, doing things in stages, like:
- Get `href`-s.
- Remove `nil` ones.
- Delete duplicates.
If the biggest problem here is actually deleting duplicates (which I imply to be at least O(n log(n)) in complexity), then this would speed things up immencely by freeing us from two `null` checks per iteration.

aadcg:
> What I'd like to try is, doing things in stages, like:
> 
>     Get href-s.
>     Remove nil ones.
>     Delete duplicates.
>     If the biggest problem here is actually deleting duplicates (which I imply to be at least O(n log(n)) in complexity), then this would speed things up immensely by freeing us from two null checks per iteration.
> 

It seems to be that you roughly described my approach in 6229001e6.

Notice that you can't get rid of the `nil`s! You need to get rid of them in the sense that you'll filter through  the subsequence that doesn't feature them. You do need in the end, as they're valid hints.

aartaka:
Ahhhh, damn, right ヽ(≧Д≦)ノ

aadcg:
@aartaka, just for completeness and reference, for instance `button`s don't have the `href` attribute.

I'm squashing and merging.

aadcg:
Merged via 44bac5781

Issue #2535, opened 2022-09-02
Cannot list messages or downloads


franburstall:
**Describe the bug**

I attempt to `list-messages` and get an error:
```
<WARN> [10:33:54] Warning: Error on separate thread: too many colons in &quot;X&quot;

  Stream: #&lt;dynamic-extent STRING-INPUT-STREAM (unavailable) from &quot;/:X:/R/S...&quot;&gt;
```
and the same thing happens with `list-downloads`.

**Information**
- OS name+version:
```sh
debian-11
```
- Desktop environment / Window manager name+version: exwm
- How you installed Nyxt (Guix pack, package manager, build from source): source
- Information from `show-system-information`:
```
Nyxt version: 3-pre-release-1-341-gf49c768d
Renderer version: GI-GTK
Operating system kernel: Linux 5.10.0-17-amd64
Lisp implementation: SBCL 2.1.1 (Dynamic space size: 3221225472)
Features: (:QUICKLISP :NYXT-GI-GTK :NYXT-GTK :WEBKIT2 :WEBKIT2-2.36
 :WEBKIT2-CORS-ALLOWLIST :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
 :GLIB-2-30 :GLIB-2-32 :GLIB-2-34 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42
 :GLIB-2-44 :GLIB-2-46 :GLIB-2-48 :GLIB-2-50 :GLIB-2-52 :GLIB-2-54 :GLIB-2-56
 :GLIB-2-58 :GLIB :NYXT-3.1 :NYXT-3 :CUSTOM-HASH-TABLE-NATIVE :SLYNK
[...]

PR #2575, opened 2022-09-15
Fail safe profile

### LINKED ISSUE #2386: Add "noload" profile
In some cases and in particular for testing it's useful to ask Nyxt to not load any file from disk to ensure reproducible behaviour.

We often do `nyxt --no-config` but it's not enough: there is also the auto-config, the history, the bookmarks, and potentially more.

Wouldn't be nice to have a single knob to disable all file reads?  Actually, we can already programmatically do it by adding a profile which does just that.  

Since we have `nosave` already, I suggest we add `noload`.  Then it be enough to call

```lisp
nyxt --profile noload
```

to disable all file loads.

Should `noload` also disable file writes?  I would say yes, since it's not very intuitive to be "write only".

Thoughts?

Ambrevar:
# Description

This adds a long overdue `failsafe` command line parameter which should easily allow users to start Nyxt regardless of the their config, history, etc.

Fixes #2386.
Also addresses https://github.com/atlas-engineer/nyxt/issues/2535#issuecomment-1236731207

# Discussion

- Should `--failsafe` really modify the `*options*` values?  It makes for short code, but technically we lose "integrity" in the `*options*` variable.

- Do you like the `nofile-profile` name?

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -33,9 +33,7 @@ inherited from the superclasses.")
     :documentation "Unique identifier for a buffer.")
    ;; TODO: Or maybe a dead-buffer should just be a buffer history?
    (profile
-    (alex:if-let ((profile-class (find-profile-class (getf *options* :profile))))
-      (make-instance profile-class)
-      *global-profile*)
+    *global-profile*
     :type nyxt-profile
     :documentation "Buffer profiles are used to specialize the behaviour of
 various parts, such as the path of all data files.")
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -777,9 +777,13 @@ core to finalize the instance.")))
             (:nxref :command 'nyxt:toggle-debug-on-error)
             " command to enable Nyxt-native debugger and see the reasons of these. Based on
 this information, you can report a bug using " (:nxref :command 'nyxt:report-bug) ".")
+        (:p "You can also try to start the browser with the " (:code "--failsafe")
+            " command line option and see if you can reproduce your issue then.  If not,
+then the issue is most likely due to your configuration, an extension, or some
+corrupt data file like the history.")
         (:p "Note that often errors, hangs, and crashes happen on the side of renderer and
 thus are not visible to the Nyxt-native debugger and fixable on the side of
-Nyxt. See below for some of those"))
+Nyxt. See below."))
 
       (:nsection :title "Playing videos"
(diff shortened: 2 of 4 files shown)


aartaka:
> * Should `--failsafe` really modify the `*options*` values?  It makes for short code, but technically we lose "integrity" in the `*options*` variable.

Hmmmm. This may be a radical suggestion, but still: what it we propagate CLI options not through the `*options*` global variable, but through function arguments. This way we can solve the problem of `:failsafe` modifying too much global state (by altering the _values passed to functions_, instead of the global variable), and have our startup sequence to retain much more information in e.g., backtraces.

But this gotta be a huge refactoring, so I'm not sure it's worth it.

> * Do you like the `nofile-profile` name?

Sounds rhyming and fun :D But then, it's too easy to confuse with `nosave` one—they both seem to play with the filesystem jargon. So maybe use something more radical, like `null-profile` or `empty-profile`?

But naming things is definitely one of the hardest problems in computing :D

Ambrevar:
> what it we propagate CLI options not through the `*options*` global variable, but through function arguments.

Example?

> they both seem to play with the filesystem jargon.

That's exactly what they do, so it's a good thing they sound related, no?

> like `null-profile` or `empty-profile`?

I find this actually more confusing, especially to non-programmers.

aartaka:
For example, the `:no-socket` CLI option is used in `start-browser`. We can rewrite it as a keyword argument, like so
``` lisp
(defun start-browser (url-strings &key no-socket) ... 
```
and then call it in `start`, like so:
``` lisp
...
      (t
       (with-protect ("Error: ~a" :condition)
         (start-browser urls :no-socket no-socket))))
    (unless *run-from-repl-p* (uiop:quit))))
```
instead of getting it from the global variable.

Ambrevar:
We cannot do this because any piece of code may want to access the options.  Try grepping `*options*`: it's scattered throughout the code base.

Having global access is very useful in this case.

Ambrevar:
OK to merge?

aadcg:
Let me test it.

aadcg:
Ok to merge!

PR #2579, opened 2022-09-16
Describe unexported

### LINKED ISSUE #2576: Describe* links from doc-strings non-functional?
**Describe the bug**

Not sure if this is a bug or a feature I don't understand.

Some links in describe-* pages lead to a dead end.

Do `C-h C buffer` and click in the `buffer-delete` link in the documentation of `buffer-delete-hook`.

*What I expect*: some description of `buffer-delete`.

*What actually happens*: I get the dialog in this screenshot which is not easy to get out of (`C-g` does not always dismiss it) and conveys no information.

Observation: I get this behaviour for all links of the following form:
```
javascript:%28function%20%28%29%20%7B%0A%20%20%20%20var%20url%20%3D%20%27lisp%3A%2F%2F1739%2Fdescribe-any%27%3B%0A%20%20%20%20var%20request%20%3D%20fetch%28url%2C%20%7B%20%27mode%27%20%3A%20%27no-cors%27%20%7D%29%3B%0A%20%20%20%20if%20%28null%29%20%7B%0A%20%20%20%20%20%20%20%20__PS_MV_REG%20%3D%20%5B%5D%3B%0A%20%20%20%20%20%20%20%20return%20chain%28request%2C%20then%28function%20%28response%29%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20__PS_MV_REG%20%3D%20%5B%5D%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20return%20at%28response%2C%20ok%29%20%3F%20chain%28response%2C%20json%28%29%29%20%3A%20null%3B%0A%20%20%20%20%20%20%20%20%7D%29%2C%20then%28null%29%29%3B%0A%20%20%20%20%7D%3B%0A%7D%29%28%29%3B
```
I think these links are built by `resolve-backtick-quote-links` and possibly shouldn't be?

**Information**
- OS name+version: debian-11
- Graphics card and driver:
- Desktop environment / Window manager name+version: exwm
- How you installed Nyxt (Guix pack, package manager, build from source): source
```
Nyxt version: 3-pre-release-1-470-ge81aceba
Renderer version: GI-GTK
Operating system kernel: Linux 5.10.0-18-amd64
Lisp implementation: SBCL 2.1.1 (Dynamic space size: 3221225472)
Features: (:QUICKLISP :NYXT-GI-GTK :NYXT-GTK :WEBKIT2 :WEBKIT2-2.36
 :WEBKIT2-CORS-ALLOWLIST :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :GLIB-2-58 :GLIB :NYXT-3.1 :NYXT-3 :CUSTOM-HASH-TABLE-NATIVE :SLYNK
 :PLUMP-UTF-32 :PARENSCRIPT :FSET-EXT-STRINGS :GLOBAL-VARS :DECLARE-TYPES
 :NAMED-READTABLES :SWANK :LPARALLEL :21BIT-CHARS :CLOSER-MOP :CL-PPCRE-UNICODE
 :CL-UNICODE :CL-PPCRE :CHUNGA :FLEXI-STREAMS :BORDEAUX-THREADS
 :LPARALLEL.WITH-CLTL2 :LPARALLEL.WITH-CAS :LPARALLEL.WITH-STEALING-SCHEDULER
 :SPLIT-SEQUENCE CFFI-FEATURES:FLAT-NAMESPACE CFFI-FEATURES:X86-64
 CFFI-FEATURES:UNIX :CFFI CFFI-SYS::FLAT-NAMESPACE ALEXANDRIA::SEQUENCE-EMPTYP
 :FAST-IO-SV :FAST-IO :CL-JSON-CLOS :CL-JSON :SBCL-USES-SB-ROTATE-BYTE
 CHIPZ-SYSTEM:GRAY-STREAMS :THREAD-SUPPORT :ASDF3.3 :ASDF3.2 :ASDF3.1 :ASDF3
 :PACKAGE-LOCAL-NICKNAMES :SB-LDB :SB-PACKAGE-LOCKS :SB-THREAD :SB-UNICODE
 :SBCL :UNIX)

ASDF version: 3.3.1
Critical dependencies: (/home/fran/common-lisp/nyxt/_build/cl-cffi-gtk/gtk/cl-cffi-gtk.asd
 /home/fran/common-lisp/nyxt/_build/cl-gobject-introspection/cl-gobject-introspection.asd
 /home/fran/common-lisp/nyxt/_build/cl-webkit/webkit2/cl-webkit2.asd)


```

Ambrevar:
# Description

Previously we had no way to describe unexported symbols, inherited symbols, etc.

We now use multiple sources for functions, variables, etc.
- 1 source for the "export Nyxt symbols" + "all nyxt-user symbols".
- 1 source for "Nyxt internals"
- 1 source for "exported symbols in non-nyxt packages".

Of course now it's trivial to create new sources with whatever listing we like (for instance "inherited symbols in all packages").

Bonus: New `Status` prompter attribute.

Fixes #2576.

# Discussion

I've removed the `universal` argument, since it seems a bit redundant now.  Having more sources is more powerful in my opinion.

Does anyone feel like `universal` should be re-introduced?  If so, then we should probably add another flag for "nyxt internal symbols".

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:


Ambrevar:
Note: It builds but I haven't had the time to test this yet.  That said, do you like the API?

source/command.lisp
```diff
@@ -257,49 +257,59 @@ See also `nyxt-packages'."
...
-If PACKAGES is NIL, return all external symbols in all packages."
-  (let* ((user-packages (mapcar #'find-package (alex:ensure-list user-packages)))
-         (packages (or (alex:ensure-list packages)
-                       (set-difference
-                        (list-all-packages)
-                        user-packages)))
-         (symbols))
+(defun non-nyxt-packages ()
+  "Return the packages that are not related to Nyxt.
+It's the complement of `nyxt-packages' and `nyxt-user-packages'."
```
aadcg, line 262: It might be too pedantic, but perhaps "It's the complement of the union between `nyxt-packages` and `nyxt-user-packages`"?

source/command.lisp
```diff
@@ -257,49 +257,59 @@ See also `nyxt-packages'."
...
+(defun symbol-status (symbol)
+  (nth 1 (find-symbol (symbol-name symbol) (symbol-package symbol))))
+
+(-> package-symbols
+    ((list-of package) &key (:status (member :internal :external :inherited :any)))
+    (list-of symbol))
+(defun package-symbols (packages &key (status :any))
+  "Return the list of all symbols from PACKAGES.
+If STATUS is specified, only include symbols with that status.
+If PACKAGES is NIL, process all packages."
```
aadcg, line 275: This won't work when you pass `nil` as the `packages` argument.

See 9d540b6fc.

source/command.lisp
```diff
@@ -257,49 +257,59 @@ See also `nyxt-packages'."
...
-                        user-packages)))
-         (symbols))
+(defun non-nyxt-packages ()
+  "Return the packages that are not related to Nyxt.
+It's the complement of `nyxt-packages' and `nyxt-user-packages'."
+  (set-difference (list-all-packages)
+                  (append (nyxt-packages) (nyxt-user-packages))))
+
+(defun symbol-status (symbol)
+  (nth 1 (find-symbol (symbol-name symbol) (symbol-package symbol))))
```
aadcg, line 267: This needs to be refactored. The argument doesn't seem to a symbol, but a list of symbols.

aadcg: Since this function is always used in the following form:

```lisp
(sera:filter (lambda (s) (eq status (symbol-status s)))
             foo-sequence)
```

Perhaps the abstraction above would be the best one.

aadcg:
Overall, I think this is a good move forward. I left some comments.

aartaka:
I guess we don't need `universal` with such a solution!

source/command.lisp
Ambrevar: Hahaha, yes, that's a bit pedantic, but if you insist, why not :)

source/command.lisp
Ambrevar: Misunderstanding: the docstring says to explicit pass `NIL`, not to call `(package-symbols)`.

Now which one is better?  Actually I'm thinking there is little benefit in having this `nil` rule at all when we can just `(package-symbols (list-all-packages))` which has the benefit of being explicit.

source/command.lisp
Ambrevar: We can have both.

source/command.lisp
aadcg: I don't insist!

source/command.lisp
aadcg: Indeed, I misunderstood it!

Feel free to revert my commit and to refactor it s you find more appropriate.

source/command.lisp
Ambrevar: Done.

source/command.lisp
Ambrevar: I've removed the implicit rule, now NIL packages won't do anything.

Ambrevar:
OK to merge?

source/describe.lisp
```diff
@@ -3,16 +3,86 @@
...
 (in-package :nyxt)
 
-(define-class function-source (prompter:source)
-  ((universal nil)
-   (prompter:name "Functions")
-   (prompter:constructor (lambda (source)
-                           (apply #'package-functions
-                                  (when (universal source)
-                                    (list (list-all-packages)))))))
+(defun description-constructor (lister)
```
aadcg, line 6: `describe-constructor` sounds more correct to me, would you agree?

source/describe.lisp
```diff
@@ -3,16 +3,86 @@
...
+  "LISTER is a function return a list of symbols from the given packages.
```
aadcg, line 7: There's minor type here. Either "is a function returning" or "is a function that returns".

source/describe.lisp
aadcg: I understand that the goal to define this function is to avoid repeating this pattern for every constructor.  But it seems odd that it assumes that `source` is an implicit argument. Can we make it better?

source/describe.lisp
```diff
@@ -35,7 +105,9 @@
              (first-line (documentation (find-package symbol) t)))
             (t
              (first-line (documentation symbol 'variable))))
-          ""))))
+          ""))
+    ("Status"
```
aartaka, line 109: `Status` is confusing. `Visibility` or `Access` maybe? Java/C#/Python use "access modifiers" for things like `public`, `private` etc.

source/describe.lisp
```diff
@@ -492,10 +595,9 @@ A command is a special kind of function that can be called with
 
 (define-internal-page-command-global describe-class
     (&key
-     universal
      (class (prompt1
              :prompt "Describe class"
-             :sources (make-instance 'class-source :universal universal))))
+             :sources 'class-source)))
```
aartaka, line 600: Wait, why a mere `class-source` without the `non-nyxt` one etc.?

source/mode/download.lisp
```diff
@@ -2,7 +2,8 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (nyxt:define-package :nyxt/download-mode
-    (:documentation "Mode to manage downloads and the download listing page."))
+    (:documentation "Mode to manage downloads and the download listing page.")
+  (:shadow #:status))
```
aartaka, line 6: Yet another reasong to replace `status` with `visibility`. Note that `command` has `visibility` method too, and it's quite resemblant of package symbols' visibility, so `visibility` is a good metaphor for `describe-*` sources here.

aadcg:
What's the status here?

aadcg:
@aartaka I believe commit 3dd84a71f is incomplete. Call `describe-any` and see that the following warnings are raised at the REPL:

```
WARNING: Error on separate prompter thread: Unknown &KEY argument: :VISIBILITY
```

aartaka:
Ahhhhhh, now I see why it was initially called `status` in this PR: CLHS calls it that in `find-symbol` docs!
```
find-symbol string &optional package => symbol, status
```

Still, I find that name more confusing than useful, even though it's in CLHS. Because:
- CLHS itself only uses it once, as an identifier for the return value, and uses other words, like `accessible` and `present` in all the other contexts.
- `status` is already occupied by `buffer` and `download` slots.
- `status` is much more opaque than `visibility`, which is both:
  - a piece of recongiseable programming lingo, and
  - a good metaphor for symbol accessibility: you see it—you get it!

aadcg:
I agree with your reasoning @aartaka.

I've made some minor changes. Care to have a final look?

source/describe.lisp
```diff
@@ -71,9 +60,22 @@ See `package-functions' for an example."
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name)))
 
+(defmacro describe-constructor (fn)
```
aartaka, line 63: Doesn't look right. We lose backtraces and other niceties of functions. I mean, yeah, the function implementation implies `source` argument, but, as long as it's not exported, it can imply anything...

source/describe.lisp
```diff
@@ -35,7 +107,9 @@
              (first-line (documentation (find-package symbol) t)))
             (t
              (first-line (documentation symbol 'variable))))
-          ""))))
+          ""))
+    ("Visibility"
```
aartaka, line 111: Can probably be shortened to one line now that it's a mere `symbol-visibility` call.

source/describe.lisp
```diff
@@ -492,10 +598,11 @@ A command is a special kind of function that can be called with
 
 (define-internal-page-command-global describe-class
     (&key
-     universal
      (class (prompt1
              :prompt "Describe class"
-             :sources (make-instance 'class-source :universal universal))))
+             :sources (list (make-instance 'class-source)
```
aartaka, line 603: Oh wait, we can use lists of symbols here...

source/describe.lisp
aadcg: Agree, cc1f74e51.

source/command.lisp
```diff
@@ -308,38 +319,29 @@ See `package-symbols' for details on the arguments."
...
-  "Return the list of all method symbols in PACKAGES and USER-PACKAGES.
-See `package-symbols' for details on the arguments."
-  (loop for sym in (package-symbols packages user-packages)
+                     (class-slots class-sym :visibility visibility)))
+           (package-classes packages)))
+
+(defun package-methods (packages &key (visibility :any)) ; TODO: Unused.  Remove?
+  "Return the list of all method symbols in PACKAGES.
+See `package-symbols'."
+  (loop for sym in (package-symbols packages :visibility visibility)
```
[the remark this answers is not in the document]
aadcg: I don't see why not. c2b190015.

source/describe.lisp
aartaka: I mean, yeah, but this syntactic sugar could well be a function generator of the same power (or even more power, given that functions do type-check when compiled, while macros simply inline things and dissapear, dissolving the responsibility), so why lose features for no gain?

Function-level backtraces are useful, because they narrow the area of responsibility to a single function, thus easing the debugging. And you want that if something goes wrong! Look at `spinneret:with-html-string`—it's hard to debug because it's all one monolythic macro that generates little to no functions and whose backtraces are unintelligible :D

I mean, yes, this thing is internal and one-shot. But still. The mantra of "if you can use a function—use a function" is there for a reason (╯°□°）╯︵ ┻━┻

aartaka: I sound like a legit C++ developer, don't I? :D

aadcg: I agree with everything you said, but I'm not convinced with regards to the case at hand.

Still, I've reverted the commit.

aadcg:
Ok to merge?

aartaka:
Yes!

aartaka:
Cool! Thanks <3

Ambrevar:
Thanks for merging this, peeps! :)

Issue #2364, opened 2022-06-14
Protect lisp:// URLs?  Allow any value in nyxt://?


Ambrevar:
See https://github.com/atlas-engineer/nyxt/pull/2306#discussion_r884151097 for a discussion.  We may have to tweak our schemes a bit.

aartaka:
> Because you said
> 
> > lisp:// URLs are not protected by WebKit, and they are CORS-ready, which makes them extemely open to any request (and exploitation).

But I've said a sentence after:

> That's why we check the buffer and URL of all the lisp:// requests in the scheme callback—to avoid exploitation. I don't see a better way here yet.

Which seems to be reliable enough to me :)

Ambrevar:
I'm not knowledgeable enough about this, let me study a little bit and come back to you.

aartaka:
Note that e81e9d02b7f3b220e76bb7940dac97307306a75f allows us to read anything safe, which is probably the best of what `nyxt://` URLs can do.

`lisp://` URLs are still the same -- checking buffer they are invoked from and thus presumably safe enough :)

PR #2586, opened 2022-09-20
Mark "nyxt:" URLs as secure.

aartaka:
# Description

This marks the `nyxt:` URLs as `:secure-p`, thus allowing outside resources, like HTTP(S) pages, to link to Nyxt-specific `nyxt:` URLs. This way, one can post example URLs and make them clickable and executable in Nyxt.

# Discussion

This is one shot at #2364, and I see two options there
- Either restricting `nyxt:` URLs locally, and unlocking their query parameters to be any `read`-able value, and maybe ever non-readable ones;
- Or opening `nyxt:` URLs up to the world, while leaving their syntax restricted.

I've picked the second way, because:
- It opens more possibilities to both `nyxt:` URLs (like embedding stylesheets and other interface elements into internal pages), and to the outside world (we can now provide links to `nyxt:` pages, like `describe-class`, on our forums, for example).
- The syntax of `nyxt:` pages doesn't have much further to move, because passing non-printable objects (like buffers) will not be easily possible however much we change the syntax. Playing with the accessibility of the links is more productive, than playing with the restricted syntax in the restricted accessibility.

But does that sound sane? @Ambrevar, @aadcg, @jmercouris? Maybe we should not even change anything in the current state of affairs? Maybe we should have a proper security audit before changing the security settings of `nyxt:` pages, given their elevated privileges? Any other opinion?

A huge security problem is that marking those URLs as secure will:
- Allow embedding JS script from the outside world into `nyxt:` pages.
- These scripts, being controlled by neither Nyxt developers, not exrension/config writers, can contain arbitrary code and change in arbitrary ways.
- Executing those in the environment of `nyxt:` pages would allow executing arbitrary Lisp code from JS.
- Executing supply chain attacks on the scripts used on `nyxt:` pages could thus lead to the full control over the browser via these compromized scripts. 
Do we even need to gaze into this Pandora box, let alone open it?

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -393,7 +393,7 @@ guarantee of the same result."
                 ;; buffer and `form' to have this buffer as the buffer-var.
                 (setf (url buffer) (quri:uri url))
                 (apply (form internal-page) args)))))))
-  :local-p t)
+  :secure-p t)
 
 (ps:defpsmacro nyxt/ps::lisp-eval ((&key (buffer '(nyxt:current-buffer)) title callback) &body form)
   "Request the lisp: URL and invoke CALLBACK when there's a successful result.


Ambrevar:
> * Allow embedding JS script from the outside world into `nyxt:` pages.

How so?

What you describe sounds a big no no.  I don't really understand how we can consider this for merging then.

Another issue, anyways, is that while `nyxt://` pages are deterministic, they technically execute code.  And while most (all?) our current `nyxt://` pages are harmless, we can't predict what the user, an extension writer or even ourselves will introduce in the future.  If a `nyxt://` page executes code depending on its arg, we open a pandora box.

I suggest we refrain from going there.  My intention with #2364 was to go the other way around, that is, add a gate that prevents any non-well controlled context from entering.

aartaka:
> > * Allow embedding JS script from the outside world into `nyxt:` pages.
> 
> How so?

Bu adding `<script>` tags :)

> What you describe sounds a big no no. I don't really understand how we can consider this for merging then.

I mean, lots of products live with this. Our situation is more dangerous, because we expose Lisp, but we're conceptually as secure as the other projects using the CDN-distributed JS libraries.

Supply chain attacks are terrible, but they are a mere consequence of depending on someone else's code. Dependencies are a mere necessity when doing complicated products. So, if the change enables using some cool JS libraries, like D3 or Mermaid, then having a risk of supply chain attacks that is as high as in all the Web-based products is enough of a trade-off, I believe.

As an additional security measure (for when/if we merge this PR), I suggest we don't use any JS dependencies unless extremely necessary in Nyxt core internal pages. This way we can be sure that we don't ship anything compromized, while allowing one to use any dependency they like in their own pages.

> Another issue, anyways, is that while `nyxt://` pages are deterministic, they technically execute code. And while most (all?) our current `nyxt://` pages are harmless, we can't predict what the user, an extension writer or even ourselves will introduce in the future. If a `nyxt://` page executes code depending on its arg, we open a pandora box.

I'm not sure I follow. What do you mean by "executes code depending on its arg"?

> I suggest we refrain from going there. My intention with #2364 was to go the other way around, that is, add a gate that prevents any non-well controlled context from entering.

So what's your suggested way, again? Not sure I get this idea of "non-well controlled context" there...

Ambrevar:
> Bu adding `<script>` tags :)

I don't understand.  Nyxt is responsible for generating the HTML content, so how would  third-party website produce a `nyxt://` link to injects JS?  Can you provide a more complete example?



> I suggest we don't use any JS dependencies

What does this mean?

> I'm not sure I follow. What do you mean by "executes code depending on its arg"?

Take `describe-class` for instance, it generates content that depends on its arg.  So if there is a security flaw in our code, an attacker could produce a URL with the right arg the gets `describe-class` to go through the flaw branch.


> So what's your suggested way, again? Not sure I get this idea of "non-well controlled context" there...

Still have to think about it...

I'm a bit confused with your point: pure-JS supply chain attacks in regular browsers may compromise the _browsing data_ of the user, to some limited extent.
But that's nothing compared to a compromised `nyxt://` URL which may execute arbitrary Lisp, and thus compromise the entire system of the user.

aartaka:
> > Bu adding `<script>` tags :)
> 
> I don't understand. Nyxt is responsible for generating the HTML content, so how would third-party website produce a `nyxt://` link to injects JS? Can you provide a more complete example?

No, no, no, I mean using JS on `nyxt://` pages, not using `nyxt://` pages to _produce_ JS usable elsewhere! Which means that I suggest having `nyxt://` pages that have `<script>` tags linking to JS libraries if necessary. This was not possible until this PR.

> > I suggest we don't use any JS dependencies
> 
> What does this mean?

Again, this PR unlocks the ability to link to things beyond `nyxt://` URLs, like HTTP links to JS files, for example. This allows us to use JS libraries on our internal pages, but comes with this security tradeoff of exposing Lisp to JS. Because of that, I suggest not using any JS `<script>` tags in the core internal pages, leaving this feature to the extensions.

> > I'm not sure I follow. What do you mean by "executes code depending on its arg"?
> 
> Take `describe-class` for instance, it generates content that depends on its arg. So if there is a security flaw in our code, an attacker could produce a URL with the right arg the gets `describe-class` to go through the flaw branch.

But wait, `nyxt://` URLs only allow constant data as arguments. This ensures that the content-producing code is quite deterministic (and cacheable too). This also means that one has to sweat a lot to even bake anything `eval`-like into an internal page. And `describe-class`, even with "the most dangerous args", cannot do more than show the class documentation...

> > So what's your suggested way, again? Not sure I get this idea of "non-well controlled context" there...
> 
> Still have to think about it...
> 
> I'm a bit confused with your point: pure-JS supply chain attacks in regular browsers may compromise the _browsing data_ of the user, to some limited extent. But that's nothing compared to a compromised `nyxt://` URL which may execute arbitrary Lisp, and thus compromise the entire system of the user.

Indeed, that's quite a lot. That's why I suggest:
- Unlocking `nyxt://` URLs, marking them as `:secure-p`.
- Leaving the `nyxt://` URL parser intact to make `nyxt://` URLs secure at least from the outside.
- Never using JS `<script>`s in core internal pages.
  - And thus leaving it for extension writers and individual configs, narrowing the impact and traceability of a potential security breach.

And note that there are ways to check the script integrity that are baked into `<script>`, like `integrity` attribute, checking the SHA signature of the loaded file. Which means that we can load only what we trust, at the level of file hash.

aartaka:
@Ambrevar, I switched `nyxt:` URLs to be `display-isolated-p` instead of being an all-in `secure-p`. What this means:
- `nyxt:` URLs _cannot_ be embedded into pages with any other scheme, which kills the iframe-embedding as the biggest danger for the `secure-p` schemes.
- `nyxt:` URLs are still universally linkable, but _are not_ universally accessible/fetchable.
- `nyxt:` URLs, even being `secure-p`, were not CORS-enabled, and neither they are now.

All in all, I believe that `display-isolated-p` is the best type for `nyxt:` URLs, because it prevents malicious use, while still expanding the abilities of the scheme for much more use-cases.

aartaka:
No, wait, `:display-isolated-p` is too restrictive for what I described :(

aartaka:
Okay, I found the optimal set of settings: none! Seems like not marking `nyxt:` URLs as either `:secure-p` nor `:display-isolated-p` gives us the perfect set of settings that I previously attributed to `:display-isolated-p`:
- `<iframe>` embedding doesn't work.
- Redirection (I tried JS `window.location.(href|assign|replace)`, but regular code 301 should work too!) works.
- Pages are linkable from outside!

aadcg:
I'm not sure if I have enough background to provide much insight. I think @Ambrevar is better qualified.

aartaka:
@Ambrevar, any feedback on this iteration (removing security settings, falling back to default scheme settings WebKit provides with properties described [in this comment](https://github.com/atlas-engineer/nyxt/pull/2586#issuecomment-1325294786))?

aartaka:
@Ambrevar, firendly ping :) Should not take more than ten minutes to review this, starting with https://github.com/atlas-engineer/nyxt/pull/2586#issuecomment-1325294786!

Ambrevar:
Sorry, a bit flooded these days, will come back to it.

From your comment, I'd say it sounds sane, but I'd like to study the WebKitGTK doc and...

Well,, actually could you add some tests for this?  It's critical after all and it's the perfect place to check whether third-party sites can violate access or not! :)

aartaka:
> From your comment, I'd say it sounds sane, but I'd like to study the WebKitGTK doc and...

There's not much documentation, mostly trial and error setting one security setting or another...

> Well,, actually could you add some tests for this? It's critical after all and it's the perfect place to check whether third-party sites can violate access or not! :)

I doubt that it's sufficient, but be2518cf4ddfa8d9a51f807e4e61388e41263cd4 should test the most dangerous scenario :)

aadcg:
What's the status here?

aartaka:
> What's the status here?

Ready to merge on my side, but @Ambrevar may disagree, so let's wait on him!

Ambrevar:
OK, looks good to me!

Sorry for taking so long to come back to this, no excuse really considering how trivial it is :p  

Can you add a comment for the `nyxt` internal scheme definition, explaining why all other options would not work, and why "no option" does what we want?

aartaka:
> OK, looks good to me!
> 
> Sorry for taking so long to come back to this, no excuse really considering how trivial it is :p

No problem :)

> Can you add a comment for the `nyxt` internal scheme definition, explaining why all other options would not work, and why "no option" does what we want?

Yes, done in 55328c5289978026d1b39b67fb1a0c0552080e84. Merging then?

Ambrevar:
Yes!

tests/renderer-offline/nyxt-url-security.lisp
```diff
@@ -0,0 +1,40 @@
...
+  ;; Same settings as HTTPS, imitating it.
+  :secure-p t)
+
+(define-test nyxt-url-not-iframe-embeddable ()
+  (let ((ready-channel (nyxt::make-channel 1)))
+    (nyxt:start :no-config t :no-auto-config t
+                :headless t
+                :socket "/tmp/nyxt-test.socket"
+                :profile "test")
+    (hooks:once-on (nyxt:after-startup-hook *browser*) (browser)
```
aadcg, line 33: This needs to be rewritten now that #2740 has been merges no?

tests/renderer-offline/nyxt-url-security.lisp
```diff
@@ -0,0 +1,40 @@
...
+         (:span#text-check ""))))
+  ;; Same settings as HTTPS, imitating it.
+  :secure-p t)
+
+(define-test nyxt-url-not-iframe-embeddable ()
+  (let ((ready-channel (nyxt::make-channel 1)))
+    (nyxt:start :no-config t :no-auto-config t
+                :headless t
+                :socket "/tmp/nyxt-test.socket"
+                :profile "test")
```
aadcg, line 32: Why not just start with `:failsafe t :headless t`?

aartaka: Because that's how other tests do it. For consistency.

And other tests do it in the way that excludes user configuration, but doesn't exclude filesystem exisstence, just in case. For reproducibility.

tests/renderer-offline/nyxt-url-security.lisp
aartaka: Done.

aartaka:
Merging now that tests are adjusted?

tests/renderer-offline/nyxt-url-security.lisp
aadcg: If that's how it's done elsewhere, then I think we should follow the example.

source/urls.lisp
```diff
@@ -402,6 +402,17 @@ guarantee of the same result."
         (error "There's no nyxt:~a internal-page defined" symbol))))
 
 (define-internal-scheme "nyxt"
+    ;; NOTE: We use no security settings because:
+    ;; - :local-p makes nyxt:// scheme inaccessible from other schemes at all.
```
aadcg, line 406: If you want to emphasize, it would sound better to say: 
"- :local-p makes nyxt:// scheme completely inaccessible from other schemes."

source/urls.lisp
```diff
@@ -402,6 +402,17 @@ guarantee of the same result."
         (error "There's no nyxt:~a internal-page defined" symbol))))
 
 (define-internal-scheme "nyxt"
+    ;; NOTE: We use no security settings because:
```
aadcg, line 405: Which security settings?

source/urls.lisp
```diff
@@ -402,6 +402,17 @@ guarantee of the same result."
...
+    ;; - :local-p makes nyxt:// scheme inaccessible from other schemes at all.
+    ;; - :display-isolated-p does not allow displaying the scheme even inside
+    ;;   the page of the same scheme.
```
aadcg, line 408: Suggestion:
- :display-isolated-p does not allow loading a nyxt:// page from a page of the same scheme.

source/urls.lisp
```diff
@@ -402,6 +402,17 @@ guarantee of the same result."
...
+    ;; - :secure-p and :cors-enabled-p are both dangerously open for a scheme
+    ;;   that allows evaluating Lisp code.
```
aadcg, line 410: Open means set to true? Maybe rephrase?

source/urls.lisp
```diff
@@ -402,6 +402,17 @@ guarantee of the same result."
...
+    ;; That's why we've gone with no security settings:
```
aadcg, line 411: The comment starts with "We use no security settings because:". The lines after this one continue to list the reasons, right?

 Do we need "That's why we've gone with no security settings:"?

source/urls.lisp
```diff
@@ -402,6 +402,17 @@ guarantee of the same result."
...
+    ;; - <iframe> embedding and exploitation doesn't work.
```
aadcg, line 412: Instead  of "doesn't work", I'd say "are impossible".

source/urls.lisp
```diff
@@ -402,6 +402,17 @@ guarantee of the same result."
...
+    ;; - Redirection (mainly tried with window.location.(href|assign|replace), but
+    ;;   regular code 301 should work too!) works.
```
aadcg, line 414: I suggest "window.location.{href,assign,replace}" so that the the parenthesis introduced in the sentence don't confuse the reader.

Instead of "regular code 301", I'd say "HTTP status code 301".

tests/renderer-offline/nyxt-url-security.lisp
```diff
@@ -0,0 +1,41 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
```
aadcg, line 1: Shouldn't this file be called nyxt-scheme-security?

aadcg:
I left some minor comments.

tests/renderer-offline/nyxt-url-security.lisp
aartaka: Well, it concerns all `nyxt://`-schemed URLs, and the conventional term for those (synonymous to the `nyxt-url` function creating them) is "Nyxt URLs".

source/urls.lisp
aartaka: I've rephrased it, but in a slightly different and wordy way, to err on the side of caution.

aartaka:
@aadcg, all addressed, I believe!

source/urls.lisp
```diff
@@ -402,6 +402,18 @@ guarantee of the same result."
         (error "There's no nyxt:~a internal-page defined" symbol))))
 
 (define-internal-scheme "nyxt"
+    ;; NOTE: We use no security settings (no :local-p, :secure-p etc.) because:
+    ;; - :local-p makes nyxt:// scheme completely inaccessible from other schemes.
+    ;; - :display-isolated-p does not allow embedding the nyxt:// page even
+    ;;   inside the page of the same scheme. Self-embedding might be a use-case
```
aadcg, line 408: Inside "a" page of the same scheme, not "the".

source/urls.lisp
```diff
@@ -402,6 +402,18 @@ guarantee of the same result."
         (error "There's no nyxt:~a internal-page defined" symbol))))
 
 (define-internal-scheme "nyxt"
+    ;; NOTE: We use no security settings (no :local-p, :secure-p etc.) because:
```
aadcg, line 405: It is very confusing to use "no" when referring to a predicate. 

How about "we don't set security setting such as :local-p and :secure-p to true because:"

source/urls.lisp
```diff
@@ -402,6 +402,18 @@ guarantee of the same result."
...
+    ;; - :local-p makes nyxt:// scheme completely inaccessible from other schemes.
+    ;; - :display-isolated-p does not allow embedding the nyxt:// page even
```
aadcg, line 407: "a nyxt:// page".

source/urls.lisp
```diff
@@ -402,6 +402,18 @@ guarantee of the same result."
...
+    ;;   inside the page of the same scheme. Self-embedding might be a use-case
+    ;;   in the future, thus :display-isolated-p is not suitable.
```
aadcg, line 409: "thus setting :display-isolated-p to NIL (if I understood it correctly?) is not suitable."?

source/urls.lisp
```diff
@@ -402,6 +402,18 @@ guarantee of the same result."
...
+    ;; - :secure-p and :cors-enabled-p are both dangerously permissive for a
```
aadcg, line 410: "Setting :secure-p and :cors-enable-p to T..."

source/urls.lisp
aartaka: T.

source/urls.lisp
```diff
@@ -414,8 +426,7 @@ guarantee of the same result."
                 ;; FIXME: This allows `find-internal-page-buffer' to find the
                 ;; buffer and `form' to have this buffer as the buffer-var.
                 (setf (url buffer) (quri:uri url))
-                (apply (form internal-page) args)))))))
-  :local-p t)
+                (apply (form internal-page) args))))))))
```
aadcg, line 429: Since the comment refers to so many predicates, wouldn't it be sensible to explicitly add them? I know we'd be setting them to default values but it would make things easier to understand.

aartaka: Not really. We can change slot initforms anytime, or we can add a new renderer supporting another set of security settings. Passing them literally is neither portable nor compatible with slot initforms.

source/urls.lisp
aartaka: 0_o

aadcg:
> @aadcg, all addressed, I believe!

It may seem I'm being picky but I'm really having a hard time processing that comment. I left some more comments.

aartaka:
@aadcg, I've made yet another revision. Note that the meaning of the security settings keywords shall not be explained in detain in this comment—it's the role of WebKitGTK docs and `scheme` slots documentation to explain what those settings do and what to set them to.

My comment is merely to explain why I set none of those settings to anything. API explanation is beyond the scope.

source/urls.lisp
```diff
@@ -402,6 +402,20 @@ guarantee of the same result."
         (error "There's no nyxt:~a internal-page defined" symbol))))
 
 (define-internal-scheme "nyxt"
+    ;; NOTE: We don't set any security settings (such as :local-p, :secure-p
+    ;; etc.)  because:
+    ;; - :local-p makes nyxt:// scheme completely inaccessible from other schemes.
+    ;; - :display-isolated-p does not allow embedding a nyxt:// page even inside
+    ;;   the page of a same scheme. Self-embedding might be a use-case in the
```
aadcg, line 409: You changed the wrong "the" to "a"!

":display-isolated-p does not allow embedding a nyxt:// page even inside a page of the same scheme."

source/urls.lisp
aadcg: You disagree with my comment?

aadcg:
> API explanation is beyond the scope.

Absolutely.

Feel free to merge.

source/urls.lisp
aartaka: Yes, that's what I've been suspicious about in https://github.com/atlas-engineer/nyxt/pull/2586#discussion_r1086496083

PR #2593, opened 2022-09-28
Predict

jmercouris:
# Description

This library adds the ability to predict commands for the user. It utilizes the new composite-sequence functionality built into the analysis library. This library is able to scan multiple sequences of varying lengths and make predictions.

# Discussion

There is one problem here. By using a sliding window of one element, we are forced into the following pattern.

Run `execute-command`.
Run `select-next-source` to view the prediction.

Now, the model has been trained, what follows `execute-command` is always `select-next-source`. This will always be your prediction. I did validate that the model still works, however, this prediction is hardly useful. Suggested remedy:

+ Bind the prediction to a keybinding
OR
+ Add a list of commands which are ignored by the model

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [ ] Documentation:
- [X] Compilation and tests:


jmercouris:
P.S. yay!

source/changelog.lisp
```diff
@@ -537,4 +537,8 @@ the selection, and the string to paste instead of the clipboard (respectively)."
    (:li "New package nicknames:"
         (:ul
          (:li  (:code "time") " for " (:code "local-time"))
-         (:li  (:code "types") " for " (:code "trivial-types"))))))
+         (:li  (:code "types") " for " (:code "trivial-types"))))
+   (:li "New prediction capabilities. Nyxt can now predict your next command, it will
+show up automatically in the execute-command menu. Nyxt uses a stochastic model
+to generate predictions for what your next command will be. The model is stored
+locally and is cleared after every session.")))
```
Ambrevar, line 544: I think you mean "kept in memory", since it's not stored to disk.

By the way, why not store it to disk?

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+;; Given the following sequence:
+;; 1 2 3 4 5
+;;
+;; We would record the following sequences/outcomes.
+;;
+;; Chain length of 1:
```
Ambrevar, line 9: This is a bit confusing because the chain length is 2 here.
So either write "chain length of 2" or "chain with 1 transition".

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+
+(defmethod add-edge ((from-node node) (to-node node))
+  (alexandria:ensure-gethash (element to-node)
+                             (edges from-node) to-node))
+
+(defmethod list-edge-elements ((node node))
+  (mapcar #'element (alexandria:hash-table-values (edges node))))
+
+(defmethod increment ((node node))
+  (incf (occurrences node)))
```
Ambrevar, line 54: Useless method?

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+
+(defmethod add-record ((model sequence-model) sequence)
+  (let* ((list-but-last-element (butlast sequence))
+         (last-element (car (last sequence)))
```
Ambrevar, line 58: Style: Prefer `first` over `car`.

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+         (leaf (alexandria:ensure-gethash list-but-last-element
+                                          (edges model)
+                                          (make-instance 'node))))
```
Ambrevar, line 61: Instead of using `butlast` and `last` separately, you can use `(sera:halves sequence -1)` which traverses the list only once.

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+    (increment (add-edge leaf (make-instance 'element-node :element last-element)))))
+
+(defmethod add-record-subsequence ((model sequence-model) sequence)
+  "Add a record for all sub sequences. E.g. transform '(3 2 1)' into:
```
Ambrevar, line 65: Typo "subsequences"

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+'(3 2 1), '(2 1), '(1)"
+  (let ((sequence (copy-list sequence)))
+    (loop while (> (length sequence) 1)
```
Ambrevar, line 68: Because it's `>` and not `>=` it seems that you are not adding the sequence of length 1.

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+         (last-element (car (last sequence)))
+         (leaf (alexandria:ensure-gethash list-but-last-element
+                                          (edges model)
+                                          (make-instance 'node))))
+    (increment (add-edge leaf (make-instance 'element-node :element last-element)))))
+
+(defmethod add-record-subsequence ((model sequence-model) sequence)
+  "Add a record for all sub sequences. E.g. transform '(3 2 1)' into:
+'(3 2 1), '(2 1), '(1)"
+  (let ((sequence (copy-list sequence)))
```
Ambrevar, line 67: You don't have to `copy-list` if instead of `pop` you just assign the `rest` of the list of `sequence`.

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+          do (pop sequence))))
+
+(defmethod predict ((model sequence-model) sequence)
+  (let* ((leaf (gethash sequence (edges model)))
+         (edges (alexandria:hash-table-values (edges leaf))))
+    (first (sort edges #'> :key #'occurrences))))
+
+(defmethod predict-subsequence-simple ((model sequence-model) sequence)
+  "Predict a sequence's next value based on all subsequence predictions. This is
+a naieve implementation which simply considers the amount of occurences without
```
Ambrevar, line 79: Typo: "naive"

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+regard to the weight of different chain lengths."
```
Ambrevar, line 80: So what about the weighted version?  Shouldn't be too hard to work on.  Any blocker?

libraries/analysis/composite-sequence.lisp
```diff
@@ -0,0 +1,88 @@
...
+  (let* ((sequence (copy-list sequence))
```
Ambrevar, line 81: Same here, no need for `copy-list` if you replace `pop`.

source/command-commands.lisp
```diff
@@ -45,6 +45,19 @@ While disabled-mode commands are not listed, it's still possible to call them
...
   (:metaclass user-class))
 
+(define-class predicted-command-source (prompter:source)
+  ((prompter:name "Predicted Command")
+   (prompter:constructor
+    (lambda (source)
+      (declare (ignore source))
+      (list (analysis:element
+             (analysis:predict (command-model *browser*)
+                               (list (last-command *browser*))))))))
```
Ambrevar, line 55: Does this properly handle the case when there is no last command or command model?

Ambrevar (commented):
Nice!  I'll test this as soon as possible.

Ambrevar:
> * Bind the prediction to a keybinding
>       OR
> 
>     * Add a list of commands which are ignored by the model

I suggest both! :)

Ambrevar:
Tests seem to be stuck into an infinite loop.

source/changelog.lisp
jmercouris: Sure, it is kept in memory. I wanted to emphasize to the user that it does not rely on ML coming from some server! I'll clarify the wording.

libraries/analysis/composite-sequence.lisp
jmercouris: The chain length is in fact one. Because there is one value -> one outcome. I can put a delimeter to make it more obvious.

libraries/analysis/composite-sequence.lisp
jmercouris: I don't believe so. I don't wish to expose the implementation details even within the program itself.

libraries/analysis/composite-sequence.lisp
jmercouris: I do usually prefer first, not sure why I used car here. I can't remember my line of reasoning.

libraries/analysis/composite-sequence.lisp
jmercouris: I see. I was translating from German where the umlaut would make `ï->ie`

source/command-commands.lisp
jmercouris: No, and I thought about that, however I did not implement it since the prompt buffer already handles this.

libraries/analysis/composite-sequence.lisp
jmercouris: No blocker!

libraries/analysis/composite-sequence.lisp
jmercouris: Clever. Done!

libraries/analysis/composite-sequence.lisp
jmercouris: True!

libraries/analysis/composite-sequence.lisp
jmercouris: Done!

jmercouris:
> Tests seem to be stuck into an infinite loop.

This is odd. The tests I've added pass instantly on my machine. Also, the tests did not seem to fail on the tests I added.

jmercouris:
> > * Bind the prediction to a keybinding
> >   OR
> >   
> >   * Add a list of commands which are ignored by the model
> 
> I suggest both! :)

Done!

jmercouris:
Ready for review two!

source/input.lisp
```diff
@@ -86,10 +86,16 @@ KEYCODE-LESS-DISPLAY (KEYCODE-DISPLAY)."
...
-  (analysis:add-record (command-model *browser*) (list (last-command *browser*)
-                                                       (symbol-function function)))
-  (setf (last-command *browser*) (symbol-function function))
-  (run-async function))
+  (let ((ignored-commands '(execute-command
+                            nyxt/prompt-buffer-mode:select-next
+                            nyxt/prompt-buffer-mode:select-previous
+                            nyxt/prompt-buffer-mode:select-next-source
+                            nyxt/prompt-buffer-mode:select-previous-source)))
+    (unless (find (name (symbol-function function)) ignored-commands :test #'equal)
```
Ambrevar: No need for `:test #'equal` here.

aartaka:
Note that there's alex:lastcar for exactly this idiom :)

aadcg:
What's the status here?

jmercouris:
Status is ready to merge.

jmercouris:
I shall merge it, let me resolve changelog issues, and then do so.

PR #2597, opened 2022-10-06
Simplify repl

aartaka:
# Description

The makes REPL more approachable and more malleable on both the side of those preferring mouse-based interfaces (lots of buttons to click) and to those on keyboard-only side (commands behind those buttons are also keyboard-invokable).

In particular, with screenshots:

## Cell reformatting:
Before:
reformat-before
After:
reformat-after

## Cell cleaning
You can press "Clean" button and you'll get a nice'n'clean new empty cell.
after-clean

# Discussion

Any suggestions for keybindings for `delete-cell`, `move-cell-*`, `clean-cell`, `delete-cell`, and `reformat-cell`?

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] My code has been reviewed by at least one peer.  (The peer review to approve a PR counts.  The reviewer must download and test the code.)
- [X] Documentation:
- [X] Compilation and tests:


aartaka:
I wanted to add comment cells and Lisp file import/export, but I realized I must stop, as this PR is growing too big to bea reasonable bite-sized one :)

Instead, I'll add those features to #2113 and leave this PR for review, as it's ready!

aartaka:
I've rebased over master, it should be perfectly mergeable now.

aadcg:
When evaluating a cell the following error appears:

```
<WARN> [14:34:16] Warning: JavaScript error: GError: Domain: "WebKitJavascriptError", Code: 699, Message: nyxt:nyxt/repl-mode:repl:1:50: TypeError: null is not an object (evaluating 'document.querySelector('#evaluation-result-2904').innerHTML = '<div>\n v2904 = <code>33</code>\n</div>'')
; in: LAMBDA ()
;     (NYXT/REPL-MODE::EVALUATE-BINDING
;      (NYXT::BINDING-KEYS 'NYXT/REPL-MODE:EVALUATE-CELL :MODES
;                          (LIST NYXT/REPL-MODE:REPL-MODE)))
; 
; caught STYLE-WARNING:
;   The variable NYXT/REPL-MODE::EVALUATE-BINDING is defined but never used.
; 
; compilation unit finished
```

Could you please take a look @aartaka?

source/mode/repl.lisp
```diff
@@ -359,34 +423,89 @@ Features:
...
+                                                      (:title "add-cell-below")
+                                                      (add-cell :id (1+ order))))
+                                     :title "Add a new empty cell below this one."
+                                     "Add cell below")
+                                    (:button.button
+                                     :onclick (ps:ps (nyxt/ps:lisp-eval
+                                                      (:title "clean-cell")
+                                                      (clean-cell :id order)))
+                                     :title "Clean the cell contents."
+                                     "🧹 Clean")
```
aadcg, line 477: The font I'm using seems not to have this glyph. I'm not sure if we should it.

aadcg:
We have in-cell buttons and 2 buttons below all cells. ("Add a cell" and "Delete all"). Since there's "Add cell below" as an in-cell button it seems that "Add a cell" isn't very useful.

aartaka:
> We have in-cell buttons and 2 buttons below all cells ("Add a cell" and "Delete all"). Since there's "Add cell below" as an in-cell button it seems that "Add a cell" isn't very useful.

Indeed. But still, it's this global action that's not making you make a (small, yet existent) mental juggling deciding where you want a new cell. Don't care, add a cell in the bottom, done.

source/mode/repl.lisp
aartaka: Mine doesn't have it either 🤔

aartaka:
Initial cell added and "Clean" emoji is no longer there ;)

aartaka:
Someone, stop me from adding new features!

aartaka:
@aadcg, merging then?

aadcg:
Not so fast.

aadcg:
bc971a601 is a good example of a commit a reviewer doesn't need to see, in my opinion. It makes sense to explain in detail the mental process behind a given change when it has _already_ landed on master.

aadcg:
Bug report. Start the REPL, create two cells and click the delete button on the cell at the top.  The wrong cell gets deleted.

aartaka:
Oh wow...

source/mode/repl.lisp
```diff
@@ -343,50 +407,125 @@ Features:
...
                                                             order)))
-                                              (input evaluation)))
+                                              (input evaluation))
+                                   (:br)
+                                   (:button.button.accent
+                                    :onclick (ps:ps (nyxt/ps:lisp-eval
+                                                     (:title "evaluate-cell")
+                                                     (evaluate-cell :id order)))
+                                    :title "Evaluate the current cell and show the result below."
+                                    "Evaluate")
```
aadcg, line 443: I'd perhaps rename this to Eval, just to keep the Lisp spirit alive.

aartaka:
> Bug report. Start the REPL, create two cells and click the delete button on the cell at the top. The wrong cell gets deleted.

Fixed in e1fe099b51b31a127e32bae584cffa37ed4986fc.

source/mode/repl.lisp
aartaka: But then, "Evaluate" is just a more verbose and understandable version of "Eval", so it kinda is OK ¯\\\_(ツ)_/¯

aartaka:
> [bc971a6](https://github.com/atlas-engineer/nyxt/commit/bc971a6018c0e9d2ecf963f259812590129b80b8) is a good example of a commit a reviewer doesn't need to see, in my opinion. It makes sense to explain in detail the mental process behind a given change when it has _already_ landed on master.

Squashed!

source/mode/repl.lisp
aartaka: Done in c9764fe080f52fdd8756210a7cff3555abae124f ;)

aartaka:
@aadcg, anything else?

aadcg:
Add a bunch of cells and type a sexp in each of them, without evaluating any of those. When you evaluate any of those, you will loose what you have not yet evaluated. Does fixing this require rethinking the architecture?

aadcg:
We have 3 global buttons that are shown below all cells. From a UI perspective, I believe that they should be visible at al times as they're global. I'd put them at the top. Would that be too difficult?

aartaka:
> Add a bunch of cells and type a sexp in each of them, without evaluating any of those. When you evaluate any of those, you will loose what you have not yet evaluated. Does fixing this require rethinking the architecture?

Not necessarily, you can just add an `:onchange` event for every cell and modify its text on the Lisp side. I was hesitant to do that to not trigger additional computations, but it should be easy enough to circumvent. 

Note that this problem was there before this PR. I'm going to add more un-debatable niceties to the REPL once we merge, including this persistence thing—it's a bit out of scope for this PR (which've grown quite monstrous already).

aartaka:
> We have 3 global buttons that are shown below all cells. From a UI perspective, I believe that they should be visible at al times as they're global. I'd put them at the top. Would that be too difficult?

It would be easy, but I myself don't see the use in putting them at the top. Because:
- The proximity of the "+ Add a cell" button and the freshle added cell gets broken—the button is at the top, while the new cell is added at the bottom.
- The interface should grow downwards, as this is the most intuitive growth direction on the Web. Thus, the actions responsible for growth (or the growth reversal, like "✕ Delete all" button) should be at the bottom.

aadcg:
When the cells span through more than one screen and you add a cell below the last one it will do it but the page gets reloaded so you loose track of the view.

aadcg:
I believe there should be a user option that would automatically add a cell below the cell that has just been evaluated. This would be set to true by default.

aartaka:
> When the cells span through more than one screen and you add a cell below the last one it will do it but the page gets reloaded so you loose track of the view.

Yes, I'm aware. And I'm already thinking of how to fix it :)

aadcg:
> Note that this problem was there before this PR. I'm going to add more un-debatable niceties to the REPL once we merge, including this persistence thing—it's a bit out of scope for this PR (which've grown quite monstrous already).

Oh I see. It seems that I lost track of how bare the previous REPL was!

aartaka:
@aadcg, I've added all of your suggestions to #2113. I'll go through them later, if you don't mind :)

aartaka:
@aadcg, merging then (I feel bad for asking it that much)?

aadcg:
> I myself don't see the use in putting them at the top. Because:
> 
>     * The proximity of the "+ Add a cell" button and the freshly added cell gets broken—the button is at the top, while the new cell is added at the bottom.
> 
>     * The interface should grow downwards, as this is the most intuitive growth direction on the Web. Thus, the actions responsible for growth (or the growth reversal, like "✕ Delete all" button) should be at the bottom.

It seems we don't disagree regarding the fact that global buttons should always be visible. 

Regarding the position itself, I don't care much. It could even be on the side! I guess the point here is that it's important to have a "toolbox panel" visible at all times.

aadcg:
Thank you for adding my suggestions to the issue. I'll review the suggestions as well. 

Let me have a last look at the code.

aartaka:
> It seems we don't disagree regarding the fact that global buttons should always be visible.
> 
> Regarding the position itself, I don't care much. It could even be on the side! I guess the point here is that it's important to have a "toolbox panel" visible at all times.

Indeed, that could be useful. And I even know how to add it :D

source/mode/repl.lisp
```diff
@@ -343,50 +409,127 @@ Features:
...
+                                                     functions)))
+                                      (reload-buffer (buffer repl-mode)))))
+                   :title "Edit the source of one of Nyxt commands in REPL."
+                   "Edit Nyxt function")
+                  (:button.button
+                    :onclick (ps:ps (nyxt/ps:lisp-eval
+                                     (:title "delete-all-cells")
+                                     (setf (evaluations repl-mode) nil)
+                                     (reload-buffer (buffer repl-mode))))
+                    :title "Delete all cells in this pane."
```
aadcg, line 534: [According to this definition](https://github.com/atlas-engineer/nyxt/issues/2113#issuecomment-1275886156), this title seems to be a bit off. Perhaps "Delete all cells in the current REPL buffer"?

aartaka: Yes!

aartaka: See 3230772618d4f2a4d1875dbeb8f3e32bd87a60ab.

aartaka:
All cool!

aartaka:
So, am I merging now?

source/mode/repl.lisp
```diff
@@ -410,8 +410,7 @@ Features:
 (define-internal-page-command-global repl ()
     (repl-buffer "*REPL*" 'repl-mode)
   "Show Nyxt REPL, a multi-pane environment to experiment with code."
-  (let* ((repl-mode (find-submode 'nyxt/repl-mode:repl-mode repl-buffer))
-         (evaluate-binding (nyxt::binding-keys 'evaluate-cell :modes (list repl-mode))))
+  (let* ((repl-mode (find-submode 'nyxt/repl-mode:repl-mode repl-buffer)))
```
aadcg, line 415: Then `let` suffices right?

aartaka: Right. 02d8952289febbe64f0909e99c8477e2e55a186e.

aadcg:
Thanks @aartaka!

source/mode/repl.lisp
```diff
@@ -157,17 +167,21 @@ Features:
 (defun package-short-name (package)
   (first (sort (append (package-nicknames package)
                        (list (package-name package)))
-               #'string<)))
+               #'<
+               :key #'length)))
```
Ambrevar, line 171: I don't understand what this fixes.  @aartaka Can you explain?

aartaka: It fixes:
```
(string< "PPCRE" "CL-PPCRE")
;; => NIL
```
In other words, `string<` is comparing strings for sorting in alphabetic order, while what we want is a string-length-based order.

Ambrevar: Oh right!  Silly me!

PR #2605, opened 2022-10-28
Review all package definitions and symbol management

aadcg:
# Description

This is a long patch that I've been working for roughly one week. Unfortunately, it needed to be handled in this monolithic way.

The goal is to standardize the handing of packages and symbols. 

Please review commit-wise.

# Discussion

It can only be merged once we fix the warnings introduced in cb694aa0eca7efbe212bd1a67fad14818b52c245. @aartaka, could be fix this?

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:


aadcg:
I need to review this PR myself!

aartaka (approved):
All good, just some stylistic suggestions! ... Most of which you've already fixed anyway and thus they are an artifact of commit-based review :P

libraries/prompter/tests/package.lisp
```diff
@@ -2,7 +2,6 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (uiop:define-package prompter/tests
-  (:use :cl :lisp-unit2)
-  (:import-from :prompter))
+  (:use :cl :lisp-unit2 :prompter)
```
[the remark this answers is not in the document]
Ambrevar: Which practice?

libraries/prompter/tests/package.lisp
```diff
@@ -1,6 +1,8 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
+(in-package :cl-user)
```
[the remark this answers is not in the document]
Ambrevar: Not needed indeed.

libraries/analysis/analysis.lisp
```diff
@@ -3,6 +3,8 @@
 
 (in-package :analysis)
 
+;; FIXME It's a bad practice to export a class.  Export the constructor instead.
```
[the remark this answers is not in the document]
Ambrevar: Agreed with @aartaka, first time I hear it may be bad practice.

libraries/prompter/tests/package.lisp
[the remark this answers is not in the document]
aartaka: The practice of `:use`-ing the package being tested in the tests package.

documents/README.org
```diff
@@ -543,6 +543,8 @@ We've also developed some of our own:
 
 - Prefer =first= and =rest= over =car= and =cdr= respectively.
 - Use =define-class= instead of =defclass=.
+- Use =uiop:define-package= instead of =defpackage= since it has a more reliable
+  behaviour when reloading a system.
```
Ambrevar, line 547: This is confusing if when consider the item below about `nyxt:define-package`.  I think they belong together, and we should explain that we use `nyxt:define-package` for everything that belongs to Nyxt (e.g. modes), and `uiop:define-package` for Nyxt-independent stuff.

libraries/analysis/analysis.lisp
```diff
@@ -85,15 +87,21 @@ operations available, some examples:
...
 
+(export-always 'keywords)
 (defmethod keywords ((document document) &optional document-collection)
   (if document-collection
       (sort (loop for word in (dictionary document)
                   collect (cons word (term-frequency-inverse-document-frequency
-                                      document document-collection word))) #'> :key #'cdr)
+                                      document document-collection word)))
+            #'>
+            :key #'cdr)
```
Ambrevar, line 97: While we are at it, `cdr` -> `rest`.

libraries/analysis/analysis.lisp
```diff
@@ -85,15 +87,21 @@ operations available, some examples:
...
       (sort (loop for word in (dictionary document)
-                  collect (cons word (term-frequency document word))) #'> :key #'cdr)))
+                  collect (cons word (term-frequency document word)))
+            #'>
+            :key #'cdr)))
```
Ambrevar, line 101: Same.

libraries/analysis/package.lisp
```diff
@@ -4,14 +4,5 @@
...
-   #:document
-   #:keywords
-   #:extract-keywords
-   #:extract-sections
-   #:summarize-text
-   #:sequence-model
-   #:add-record
-   #:predict
-   #:element))
+  (:use :cl)
```
Ambrevar, line 7: Why rename `common-lisp` to `cl`?
I've noted that we are not consistent here.  I suggest we stick to non-acronyms, to keep `common-lisp` here and rename all other `cl` instances.

libraries/history-tree/tests/tests.lisp
```diff
@@ -2,67 +2,69 @@
...
+  (:import-from :class-star #:define-class))
+
 (in-package :history-tree/tests)
 
 (defvar *owner* "test-owner")
 
-(defun make (&key (owner *owner*) (key 'identity))
-  (htree:make :initial-owners (list owner) :key key))
+(defun %make (&key (owner *owner*) (key 'identity))
+  (make :initial-owners (list owner) :key key))
```
Ambrevar, line 15: Disagree: Tests should _not_ `use` the target package, because tests are here to validate the API and thus should not have unchecked access to private symbols.

I suggest to revert all these changes and to not `use` `htree`.

libraries/prompter/tests/tests.lisp
```diff
@@ -8,12 +8,10 @@
   (trivial-package-local-nicknames:add-package-local-nickname :sera :serapeum))
 
 (defun source1-suggestions (prompter)
-  (mapcar #'prompter:value (prompter:suggestions
-                            (first (prompter:sources prompter)))))
+  (mapcar #'value (suggestions (first (sources prompter)))))
```
Ambrevar, line 11: Same here, don't remove the package prefix.

nyxt-ubuntu-package.asd
```diff
@@ -4,7 +4,7 @@
 ;; TODO: Can we move this file to build-scripts?  Looks like linux-packaging
 ;; fails to find the produced binary then.
 (defsystem "nyxt-ubuntu-package"
-  :defsystem-depends-on (linux-packaging)
+  :defsystem-depends-on ("linux-packaging")
```
Ambrevar, line 7: Why the quotes?  We are consistently _not_ using quote for dependencies, so why include them?
I find quotes overly noisy.

There are multiple styles here, let's pick one and stick to it.

libraries/analysis/analysis.lisp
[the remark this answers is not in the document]
aadcg: I read it from

https://lispcookbook.github.io/cl-cookbook/clos.html#creating-objects-make-instance

And it seemed reasonable. Did I misunderstand it?

tests/benchmark-package.lisp
```diff
@@ -2,5 +2,8 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (benchmark:define-benchmark-package nyxt/benchmark
-  (:use #:common-lisp)
-  (:import-from #:nyxt))
+  (:use :nyxt)
+  ;; trivial-benchmark:start and nyxt:start are both exported
+  (:shadow #:start))
+
+(setf lparallel:*kernel* (lparallel:make-kernel (or (serapeum:count-cpus) 1)))
```
Ambrevar, line 9: This should be wrapped as a function which _does not_ reset the kernel on reload.  See `initialize-lparallel-kernel` in the core.

tests/offline/define-configuration.lisp
```diff
@@ -2,51 +2,50 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (in-package :nyxt/tests)
-(use-nyxt-package-nicknames)
 
 (define-test simple-configuration ()
   (let ((test-url (quri:uri "about:blank")))
-    (nyxt:define-configuration nyxt:browser
-      ((nyxt:default-new-buffer-url test-url)))
+    (define-configuration browser
```
Ambrevar, line 8: Same, don't remove the `nyxt` prefix.

tests/offline/global-history.lisp
```diff
@@ -2,23 +2,22 @@
...
 (defmethod files:read-file ((profile test-profile) (file nyxt-file) &key &allow-other-keys)
-  "This method guarantees FILE will not be loaded from disk in TEST-PROFILE."
+  "This method guarantees FILE will not be loaded from disk in PROFILE."
   nil)
 
 (define-test global-history ()
   ;; Set profile to nosave to inhibit serialization / deserialization.
   ;; TODO: We should still test serialization and deserialization.
-  (let* ((nyxt:*global-profile* (make-instance 'test-profile))
+  (let* ((*global-profile* (make-instance 'test-profile))
```
Ambrevar, line 17: Same, don't remove the `nyxt:` prefix.

tests/online/urls.lisp
```diff
@@ -70,30 +71,30 @@
   ;;"Valid IP URL with path"
   (assert-no-error t (valid-url-p "http://192.168.1.1/foo"))
   ;; "same schemeless URLs"
-  (assert-true (nyxt::url-equal (quri:uri "http://example.org")
-                                (quri:uri "https://example.org/")))
+  (assert-true (url-equal (quri:uri "http://example.org")
```
Ambrevar, line 74: Same, don't remove the `nyxt:` prefix.

Ambrevar:
Good job, André!  Thanks a lot.
I agree with most things, see my comments.
My main disagreement (and it's important) is that we should not remove the target package prefix from test files.

libraries/history-tree/tests/tests.lisp
```diff
@@ -2,9 +2,11 @@
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
 (in-package :cl-user)
+
 (uiop:define-package history-tree/tests
   (:use :cl :lisp-unit2)
   (:import-from :class-star #:define-class))
+
```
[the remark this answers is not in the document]
aadcg: Do you mean the blank line before or after `uiop:define-package`? Or both?

I think it makes sense to separate package definitions from setting the `*package*`.

documents/README.org
aadcg: Good point.

libraries/analysis/analysis.lisp
aadcg: See 538b64b52.

libraries/analysis/package.lisp
aadcg: See my commit message for e9de44956ea490eb883e2916af9232a9b9f98b2d.

> Since we always refer to common-lisp-user package as cl-user, it seems sensible.

Do you prefer to stick to `common-lisp` and `common-lisp-user`?

libraries/prompter/tests/package.lisp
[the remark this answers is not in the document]
aadcg: Is 3ab90ae4a what you have in mind?

libraries/prompter/tests/package.lisp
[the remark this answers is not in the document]
aadcg: Right, I forgot about external libraries. If you prefer, please open an issue in `ntemplate` and I'll do it myself @aartaka 

On this topic: how do we draw the line between libraries that land in Nyxt's repository and outside of it?  For instance `prompter` seems to be fully independent from Nyxt.

tests/online/urls.lisp
aadcg: Why? It's an exported symbol.

tests/offline/global-history.lisp
aadcg: Because it's a global variable?

libraries/analysis/package.lisp
Ambrevar: Good point.  Well, no strong opinion as long as we are consistent.

libraries/prompter/tests/package.lisp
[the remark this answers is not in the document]
Ambrevar: Nooooooo! :p  This is bad practice in my opinion, this should be reverted.
Tests are also here to test the API: if the package is `use`d, there is no good way to test this.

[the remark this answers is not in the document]
Ambrevar: @aadcg There is no line, just maturity and our previous practices.  Today, I believe that if we feel confident something should be a library, you can start straight away with a separate repo.

libraries/history-tree/tests/tests.lisp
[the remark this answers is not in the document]
aartaka: I mean the one after it.

libraries/analysis/analysis.lisp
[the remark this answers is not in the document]
aartaka: But then, Cookbook says that it's about having control over argument list, while in our case controlling argument list is a rare use-case, compared to the flexibility of `make-instance`.

libraries/history-tree/tests/tests.lisp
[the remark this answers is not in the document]
aadcg: @Ambrevar any opinion?

source/debug.lisp
```diff
@@ -1,7 +1,8 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-(uiop:define-package :nyxt/debug ; TODO: Rename to `nyxt/develop' to avoid confusion with `debugger'?
+; TODO: Rename to `nyxt/develop' to avoid confusion with `debugger'?
```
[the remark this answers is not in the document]
aadcg: Right!

libraries/prompter/tests/package.lisp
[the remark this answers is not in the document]
aartaka: Yes!

libraries/prompter/tests/package.lisp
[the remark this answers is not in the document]
aadcg: > Nooooooo! :p This is bad practice in my opinion, this should be reverted.
Tests are also here to test the API: if the package is used, there is no good way to test this.

Absolutely disagree. I don't see any connection between testing and a package's API. Also, `:use`-ing the package on its testing package counterpart if the only way to have a reasonable way of reasonable about the package. 

A separate, but important, question is whether one wants to use package prefixes when testing packages (for exported symbols). That is case-specific, in my opinion.

libraries/theme/tests/test.lisp
```diff
@@ -24,40 +23,40 @@
...
+                  (themed-css *theme*
                     (body
-                     :color (if (theme:dark-p theme:theme)
-                                theme:background
-                                theme:on-background)
-                     :background-color theme:primary))))
+                     :color (if (dark-p theme)
+                                background
+                                on-background)
+                     :background-color primary))))
```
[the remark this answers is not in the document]
aadcg: What makes them special, particularly when compared to `with-theme` or `+light-theme+`?

libraries/prompter/tests/package.lisp
```diff
@@ -3,8 +3,9 @@
 
 (in-package :cl-user)
 
-(uiop:define-package prompter/tests
-  (:use :cl :lisp-unit2 :prompter)
-  (:import-from :class-star #:define-class))
+(uiop:define-package :prompter/tests
+  (:use :cl :lisp-unit2)
+  (:import-from :class-star #:define-class)
+  (:import-from :prompter))
```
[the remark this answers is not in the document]
aadcg: All `:import-from` does, from what I understood, is signaling ASDF that `prompter/tests` depends on `prompter`. In that sense, I think what you say regarding "importing" external symbols does make sense.

I was then able to `:use` the package being tested, but I don't recall exactly why...

libraries/history-tree/tests/tests.lisp
```diff
@@ -109,30 +108,26 @@
                 (htree::map-data (all-contiguous-owned-nodes
                                   (apply #'go-to-child
                                          "http://example.root/A2"
-                                         (multiple-value-list
-                                          (go-to-child
+                                         (multiple-value-list (go-to-child
```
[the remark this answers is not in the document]
aadcg: Ooops, very well noted.

libraries/history-tree/tests/tests.lisp
```diff
@@ -155,6 +150,7 @@
                                             :title "Example page"))
         (web-page2 (make-instance 'web-page :url "http://example.org"
                                             :title "Same page, another title")))
+    ;; simplify
```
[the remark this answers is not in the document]
aadcg: Yes, but actually I don't see the point so I'm getting rid of this comment.

libraries/analysis/analysis.lisp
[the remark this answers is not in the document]
aadcg: I see. I'll get rid of this comment.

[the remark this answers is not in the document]
aadcg: Perhaps the question here is whether `analysis` should depend on `:class-star`, but that's not related to this PR anyway.

libraries/theme/package.lisp
```diff
@@ -6,18 +6,4 @@
 (uiop:define-package :theme
   (:use :cl)
   (:import-from :serapeum #:export-always)
```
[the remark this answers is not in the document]
aadcg: Notice that `:theme` depends on `:cl-css`.

tests/benchmarks/prompter.lisp
```diff
@@ -16,8 +16,8 @@ The build time for the Guix package database is not taken into account."
   "Measure the time needed to match against all Nyxt command docstrings.
 Inputs are random character sequences taken from the docstrings."
   (let* ((suggestions (loop for command in nyxt::*command-list*
-                            if (> (length (slot-value command 'nyxt::docstring)) 0)
-                              collect (slot-value command 'nyxt::docstring)))
+                            unless (uiop:emptyp (documentation command t))
```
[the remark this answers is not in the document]
aadcg: Good point!

documents/README.org
aadcg: See 6fded53ed.

libraries/analysis/package.lisp
aadcg: I agree on consistency.

@aartaka, any strong opinion on `common-lisp(-user)` vs `cl(-user)`?

libraries/history-tree/tests/tests.lisp
aadcg: I've explained in another thread why I have doubts about it, but I'm happy to be convinced otherwise.

@jmercouris `:use`d `:analysis` when testing the mentioned package. What's your opinion on the topic?

nyxt-ubuntu-package.asd
aadcg: > Why the quotes? We are consistently not using quote for dependencies, so why include them?

When we define a system with `defsystem` we always use quotes. Since it's a `:defsystem-depends-on` clause, it seemed to me that it would be consistent to use quotes as well. Notice that I have not added quotes on any one of our `:depends-on` clauses!

aadcg: @aartaka actually I wasn't aware of that post...

I just tried to look at examples from our codebase and try to come up with a unified and consistent guideline.

tests/benchmark-package.lisp
aadcg: I copied from `libraries/prompter/tests/package.lisp`. See commit 75001b41aa801ef49f4fdb1a24ba480014b05fdb.

Are you suggesting that both of these instances should be re-written?

aadcg: I think that a372a9356 is what you had in mind.

Notice that I had to repeat the same block since they're defined in completely different systems.

tests/offline/global-history.lisp
aadcg: The relevant discussion is taking place in another thread.

tests/online/urls.lisp
aadcg: The relevant discussion is taking place in another thread.

tests/offline/define-configuration.lisp
aadcg: The relevant discussion is taking place in another thread.

aadcg:
>  My main disagreement (and it's important) is that we should not remove the target package prefix from test files.

I see that. I'm meditating on that... Honestly I have doubts and I believe that `:use`-ing the package being tested makes sense. Keeping the package prefixes seems to be unrelated to `:use`-ing the packge, and I'm fine with  that.

libraries/history-tree/tests/tests.lisp
Ambrevar: I think it's bad practice for the same reason: it hides potential issues with unexported symbols or wrongly exported symbols.

nyxt-ubuntu-package.asd
Ambrevar: Either we go full on with double quotes (more noisy but more reliable), or not at all.

tests/benchmark-package.lisp
Ambrevar: Yup we need the `unless` here at the very least..  It's unfortunate that lparallel does not have a helper for such a common case.

Ambrevar:
Why does it make sense to `use` the package?

In testing it's popular to talk about "black box testing" to test an API: you don't get to see the internals, you just fiddle with the exported stuff.  It's good practice, since the internals may hide a poor API design.

Testing also validates the API, which is good.

Last but not least, tests are not to be convenient to write, there are here to _validate_, even if they are cumbersome to write.  In the end, tests are not touched very frequently.

Ambrevar:
I also want to emphasize that one of our top priorities right now is to stabilize Nyxt, mostly by the means of improving the test suite.

`use`-ing packages in tests is a significant regression in this area... :/  Let's make our tests as useful as possible! :)

libraries/history-tree/tests/tests.lisp
aadcg: I'd like to listen to @jmercouris on this one.

nyxt-ubuntu-package.asd
aadcg: @Ambrevar, I believe we're on the same page! Since we were using double quotes for systems, I made the change we're discussing here. Am I misunderstanding it?

libraries/history-tree/tests/tests.lisp
jmercouris: I think it depends on what kinds of tests we are doing. probably we should decide if it is ok if tests use unexported symbols. If we are strictly testing the API, then Pierre is right. If not, then André and myself are right. Do we need an absolute rule for this? Why not a: "it depends". It will not hurt the program if the tests are all different. Some need to test just the API, some the internals as well to validate some important information about their state. Finally, if it comes down to a *must* do always one way, then I would choose Pierre's way.

tests/benchmark-package.lisp
aadcg: There's `check-kernel`. We could perhaps write a short utility based on this one.

libraries/history-tree/tests/tests.lisp
Ambrevar: In any case, no need to remove `use` packages when it was not done before, since the tests have already been written with this in mind.

nyxt-ubuntu-package.asd
Ambrevar: Nope, all good!

tests/benchmark-package.lisp
aadcg: Sorry, `check-kernel` isn't very useful since it is an interactive way of setting `*kernel*`. [I've sent a patch to `lparallel`](https://github.com/lmj/lparallel/pull/44).

aadcg: We're good with what we have thus far, and it's not a priority.

createyourpersonalaccount:
This is a nice set of commits, but I think it should be separated into multiple PRs.

For example, 64e1755ed7516bc27e1c33abd2b9853ae1f8cb75, 35c3d6f15a8f96e2123e2c732b0a86e8715ff9bd, 043db99b80b2336789e6da8a86f2501c75784148 are typos. You could create a new branch `fix-typos` and [cherry pick](https://www.atlassian.com/git/tutorials/cherry-pick) these commits. (This list of commits is not exhausting, there's other typo fixes like 2b19a903e71dcc257b0101e710537c370518ba32 for example.)

On the other hand, commits such as af9e1826e96ebd03b5cf988555f4c067369a2065, e0187cc4056ad6c0aef283038612a82204d11e4e have to do with how symbols are used. These can be cherry picked on a branch named `symbol-fix`

There are also commits such as 4561c64565a2d3db6d156b8c56b77d25ccaebf4b which are purely stylistic, and have equivalent behavior, and they can be cherry picked on a branch named `style-fix`.

There may be other branches you need to name, I briefly looked at the commits. I recommend that once you break up the commits in this way, you close this PR and open separate PRs with your new branches.

aadcg:
> In testing it's popular to talk about "black box testing" to test an API: you don't get to see the internals, you just fiddle with the exported stuff. It's good practice, since the internals may hide a poor API design.
>
> Testing also validates the API, which is good.
> 
> Last but not least, tests are not to be convenient to write, there are here to _validate_, even if they are cumbersome to write. In the end, tests are not touched very frequently.

Everything you mention is correct, but I don't see how any of that related to `:use`ing the package being tested. You can `:use` the package and still use one colon to stress that you're dealing with an exported symbol. Likewise, if you see a double colon, that raises a red flag.  

Also, this is not some kind of esoteric opinion of mine.  If you look at most CL libraries out there, they all go by this practice.

aadcg:
> I also want to emphasize that one of our top priorities right now is to stabilize Nyxt, mostly by the means of improving the test suite.

Yes, absolutely! But you can't stabilize anything without first stabilizing the team and its practices. (We all different Nyxts inside our minds, but this is a different subjects). Frequently, I start something and I go down the rabbit hole when there's lack of consistency. Notice that some of tests suites do `:use` the package.

> `use`-ing packages in tests is a significant regression in this area... :/ Let's make our tests as useful as possible! :)

Sure. I will reduce this PR to the relevant commit without touching this sensitive point.  But please keep this in mind for the future. When you `:use` the package being test, you get immediately one thing for free: the package had to be compiled and loaded. If something isn't smooth, it will scream, which is good.

libraries/history-tree/tests/tests.lisp
aadcg: I believe you're both confused... 

Testing the API had nothing to do with `:use`ing the package being tested. If you `:use` package `tomato` in `tomato/tests` and you want to check whether unexported symbols are being used just eyeball it and look for `::`.

Look for instance at the tests from `bordeaux-threads` which makes use of `bt::with-cv-access`. When you want to make the "black box" test, this is a red flag.  Not `:use`ing the target package doesn't change that in any way.

aadcg:
@createyourpersonalaccount will follow that advice.

Ambrevar:
> When you `:use` the package being tested, you get immediately one thing for free: the package had to be compiled and loaded. If that step doesn't go smoothly, it will scream.

You get the same effect with `:import-from`.  We already do this.

> You can `:use` the package and still use one colon to stress that you're dealing with an exported symbol.

I'm talking about the opposite: tests may rely on unexported symbols without prefixing with the package, thus not realizing that the symbol is unexported.  So no visible red flag here.

> Also, this is not some kind of esoteric opinion of mine. If you look at most CL libraries out there, they all go by this practice.

No, many don't.

aadcg:
> You get the same effect with :import-from. We already do this.

It is not the same actually.

Ambrevar:
> > You get the same effect with :import-from. We already do this.
> 
> It is not the same actually.

What is not the same?  Can you give an example?

libraries/analysis/composite-sequence.lisp
```diff
@@ -26,6 +26,8 @@
 
 (in-package :analysis)
 
+;; FIXME It's a bad practice to export a class.  Export the constructor instead.
```
Ambrevar, line 29: Why is it bad practice to export a class?

aadcg:
>      What is not the same? Can you give an example?

Yes. 

From this git checkout, go to the REPL and run `,test-system RET nyxt/prompter`. Notice that all tests pass smoothly.

Then apply the change below to `libraries/prompter/tests/package.lisp` and conclude that it's not the same.

```diff
 
 (uiop:define-package :prompter/tests
-  (:use :cl :lisp-unit2)
+  (:use :cl :lisp-unit2 :prompter)
   (:import-from :class-star #:define-class)
-  (:import-from :prompter))
+  ;; (:import-from :prompter)
+  )
```

libraries/prompter/tests/package.lisp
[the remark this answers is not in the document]
Ambrevar: `import-from` takes as argument a package and, optional, some of its external symbols to `use` in the defined package.

If no external symbol is given, then the effect of `import-from` is merely to signal a dependency to ASDF.

aadcg:
> I'm talking about the opposite: tests may rely on unexported symbols without prefixing with the package, thus not realizing that the symbol is unexported. So no visible red flag here.

I don't follow. In that case the tests wouldn't pass... Can you explain with an example?

> No, many don't.

Can you give examples?

libraries/analysis/tests/tests.lisp
```diff
@@ -1,9 +1,8 @@
 ;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-(in-package :cl-user)
-(uiop:define-package analysis/tests
-    (:use #:common-lisp #:lisp-unit2 :analysis))
+(uiop:define-package :analysis/tests
+  (:use :cl :lisp-unit2 :analysis))
```
Ambrevar, line 5: Do not use analysis here.  But since it was used before, maybe this is out of the scope of this PR.  Then just add a `TODO` so we remember to move it later.

libraries/analysis/composite-sequence.lisp
aadcg: Ups. Will fix it.

Ambrevar:
> Then apply the change below to `libraries/prompter/tests/package.lisp` and conclude that it's not the same.
> 
> ```diff
>  
>  (uiop:define-package :prompter/tests
> -  (:use :cl :lisp-unit2)
> +  (:use :cl :lisp-unit2 :prompter)
>    (:import-from :class-star #:define-class)
> -  (:import-from :prompter))
> +  ;; (:import-from :prompter)
> +  )
> ```

What I meant is that it's the same _with regard to marking `prompter` as a dependency_.  Of course it does not have the same effect overall, otherwise we would not have this discussion :)

createyourpersonalaccount:
@aadcg My understanding of the difference between `:use` and `:import-from` is that the former will make accessible all symbols while the latter gets to pick and choose.

Thus an accidental overwrite of a symbol value is less likely. An example is this:

```lisp
(defpackage #:foo
  (:use #:cl)
  (:export #:bar #:my-func))
(in-package #:foo)
(defun bar () 1)
(defun my-func () (bar))

(defpackage #:foo-test
  (:use #:cl :foo))
(in-package #:foo-test)
(defun bar 2)
; now (foo:my-func) returns 2!
```

Ambrevar:
@aadcg Forget it, I was confused for the whole discussion and I was talking about something that completely unrelated to this PR... :(

So the real reason why I prefer `:import-from` over `:use` is just a matter of style: it makes it apparent which are the tested API points as opposed to the standard CL / lisp-unit2 symbols.

Another benefit of `import-from` is that it avoids collision.  But this point is kind of moot if tests are written with `:use` in mind in the first place.

libraries/analysis/tests/tests.lisp
aadcg: Good point!

aadcg:
> What I meant is that it's the same with regard to marking prompter as a dependency. 

Clear. But this is exactly why I've raised the topic. Regardless, this is off-topic at this point :D

aadcg:
>  Another benefit of import-from is that it avoids collision. But this point is kind of moot if tests are written with :use in mind in the first place.

This a very justified concern. It did bite me before, but then I read the following

> (From `lisp-unit2` README) In some cases, particularly when converting from lisp-unit(1) we need our tests to be in a different package (because tests are functions in their name's package). In lisp-unit these tests would not conflict with a function named the same (because tests were not functions). To ease conversion, the reintern-package argument will reintern all test names and tags provided into a different package. Define test accepts a package argument that mirrors this functionality. Suggested usage is to either have tests be named differently from the functions they test or to have tests and tags be in an explicitly referenced package, eg: (define-test my-tests::test1 (:tags '(my-tests::tag1)) ...)

aadcg:
>  So the real reason why I prefer :import-from over :use is just a matter of style: it makes it apparent which are the tested API points as opposed to the standard CL / lisp-unit2 / test-package symbols.

Could you please explain again?

aadcg:
@Ambrevar I pushed the change to the analysis test package since it's so trivial.

The goals of this PR may seem fuzzy, so let me list them:

- make sure all tests (for all systems) run locally. For instance, running `(asdf:test-system "nyxt/benchmark")` was broken but I've fixed it in 2f52cbc0c.
- enforce consistency between all package definitions. For instance, make sure we settle on `:import-from` instead of `:use` when testing packages (@Ambrevar convinced me this a good thing); consistently define packages via `uiop:define-package` or `nyxt:define-package`.
- consistently refer to packages as keywords (has the benefit that the package can be easily inspected from SLIME/SLY) and "package symbols" as unintered symbols. Systems names are referred to as strings, but system dependencies as symbols (I believe this consistency helps when `grep`ing).

Regarding the last point, [I've come across a nice discussion on it](https://stackoverflow.com/questions/74348301/why-do-we-specifiy-packages-using-pound-colon-in-common-lisp).

Now that I understand (1) all the twists and turns of our packages and systems and (2) `lisp-unit2`, I can enrich our tests. 

With this in mind, please have a last look at this PR. Does it all make sense? @aartaka @Ambrevar @createyourpersonalaccount

Ambrevar:
> > So the real reason why I prefer :import-from over :use is just a matter of style: it makes it apparent which are the tested API points as opposed to the standard CL / lisp-unit2 / test-package symbols.
> 
> Could you please explain again?

Example:

```lisp
(defun avg (seq)
  ...)

(assert-eq (max (1 (avg (seq-builder ...)))
```

In this made-up example, it's not which API point is being tested, since all symbols are unprefixed.  
- `assert-eq` comes from Lisp-Unit2
- `max` from standard CL
- `avg` from the test package.
- Only `seq-builder` comes from the package being tested.

If we don't `use` the tested package, this enforces the use of package prefixes for the test targets, thus making clearer to the reader what's being tested.

tests/renderer-package.lisp
```diff
@@ -1,27 +1,25 @@
...
 
-(class-star:define-class nyxt-user::test-profile (nyxt:nyxt-profile)
+(define-class nyxt-user::test-profile (nyxt:nyxt-profile)
   ()
   (:documentation "Test profile."))
 
-(defmethod files:write-file ((profile nyxt-user::test-profile) (file files:file) &key)
+(defmethod nfiles:write-file ((profile nyxt-user::test-profile) (file nfiles:file) &key)
   "Don't persist test data."
   nil)
```
Ambrevar, line 14: Why `nfiles` instead of `files`?

Ambrevar (commented):
Looks good to me beside some final nits! :)

createyourpersonalaccount:
Some of the style changes I don't know enough to judge, so I can't comment, but I did not spot anything that I consider an issue.

aadcg:
@Ambrevar all clear about what you meant by a "matter of style". Actually, in our case, it's even more critical since we might be testing `nyxt` but symbols are coming form all sorts of `nyxt/foo` packages. So, again, I am convinced by your arguments. Thanks!

tests/renderer-package.lisp
aadcg: Sorry, my mistake when re-organzing my commits. Will fix.

libraries/user-interface/package.lisp
```diff
@@ -1,19 +1,16 @@
...
+           #:buffer
+           #:update
+           #:text
+           #:action
+           #:to-html-string
+           #:paragraph
+           #:progress-bar
+           #:button
+           #:percentage
+           #:connect))
```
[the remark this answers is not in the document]
aadcg: See aba622101.

libraries/theme/theme.lisp
```diff
@@ -74,6 +76,16 @@ out from all of the other theme colors.")
...
+(export-always 'theme)
+(export-always 'background)
+(export-always 'on-background)
+(export-always 'primary)
+(export-always 'on-primary)
+(export-always 'secondary)
+(export-always 'on-secondary)
+(export-always 'accent)
+(export-always 'on-accent)
+(export-always 'font-family)
```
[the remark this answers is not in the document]
aadcg: Well noted @Ambrevar thanks.

@aartaka that would go against our style guide.

aadcg:
Thanks Everyone!

Ambrevar:
Thank you for your patience :)

aadcg:
Thank you @Ambrevar! I like to be convinced :)

PR #2612, opened 2022-11-01
Move renderer code to their own package

### LINKED ISSUE #2390: Move renderers to their own package
We've need this forever, time we did it!

At the very least, if we want to support multiple renderers they should be in separate packages to avoid name clashes.

I suggest the `nyxt/renderer/gtk` naming convention.

Don't we have an issue about this already?

Ambrevar:
# Description

This paves the way to third-party renderer support.

It's cleaner since renderers no longer pollute the core.

It also helps highlighting which parts of the foreign-interface API should be
exported.

Fixes #2390.

# Discussion

- [x] I had to extract the explicit reliance on GTK code from `web-extension-callbacks.lisp` to `foreign-interface.lisp`.
@aartaka Can you confirm this is OK?

- [ ] Is the `nyxt/renderer/gtk` naming convention OK?  It's not consistent with our naming convention for modes, e.g. `nyxt/blocker-mode`.  The `nyxt/renderer/gtk` convention is better because 
  1. It enables better completion.
  2. It's compatible with package inferred systems.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [x] Compilation and tests:


source/renderer/gi-gtk.lisp
```diff
@@ -11,19 +11,28 @@
...
 
-(in-package :nyxt)
+(nyxt:define-package :nyxt/renderer/gi-gtk
+    (:documentation "GTK renderer leveraging GObject Introspection.
+For now it is also partly based on `nyxt/renderer/gtk'."))
+(in-package :nyxt/renderer/gi-gtk)
 
-(setf +renderer+ "GI-GTK")
+#+sbcl
+(progn
```
aartaka, line 20: What's this about, just out of curiosity?

aartaka (approved):
All cool, WebExtensions fix if fine too!

source/renderer/gi-gtk.lisp
aartaka: Got it!

Ambrevar:
@aartaka I've reverted the global unlocking of packages with 3d4527a9ce0415dd31d1cc4a4c4ee1f01dd50328.

`renderer-thread-p` is now a method, but clearly this  calls for a better abstraction.  Suggestions?

In any case, I'm happy with this PR.  Let me know if it's OK to merge.

Ambrevar:
@aadcg @aartaka What about the question of package naming convention mentioned in the original post?  We have `nyxt/renderer/gtk` here, but our modes are named `nyxt/blocker-mode` for instance.

aadcg:
You mention that:

> The nyxt/renderer/gtk convention is better because it enables better completion

I think that boils down to a matter of taste. Both approaches have pros and cons, I'd say. If I want to complete `nyxt/blocker-mode`, I could argue it's easier to complete than `nyxt/mode/blocker` since the prefix `nyxt/bl` most likely narrows down to a single item. On the other than the prefix `nyxt/mode/` would show all package mode symbols. 

I'm on the side of consistency (between `nyxt/mode` and `nyxt/renderer`), whatever the decision is.

About inferred systems, I think we should defer that decision as much as possible since it's yet another layer of complexity. Making a decision based on a possible future might not be very useful.

aartaka:
> @aadcg @aartaka What about the question of package naming convention mentioned in the original post? We have `nyxt/renderer/gtk` here, but our modes are named `nyxt/blocker-mode` for instance.

That's an uneasy one. On the one hand:
- `blocker-mode` is a command synonymous to the mode.
- Resolving mode names is easier if you just suffix them with `-mode`.
- It's more or less backwards-compatible.
On the other hand:
- `nyxt/mode/blocker` mirrors directory hierarchy.
- And, maybe, ASDF module hierarchy too.
- It's indeed much more completion-friendly too, like `dotted.package.convention` or any other hierarchical convention.

All in all, I want to stay with `nyxt/blocker-mode`, because it's kinda synonymous to the mode name and is backwards-compatible. And because the benefits of the `nyxt/mode/name` are not exactly required for Nyxt moving forward. But that's not a certain decision for me, so I can accept `nyxt/mode/name` convention and use it just fine :)

aadcg:
On another topic: these renderer packages depend on `nyxt`, so shouldn't all of those symbols be external in `nyxt`? 

Here's a list of them:

```
nyxt::*context-menu-commands*
nyxt::*headless-p*
nyxt::*headless-p*
nyxt::*headless-p*
nyxt::*interactive-p*
nyxt::*run-from-repl-p*
nyxt::*schemes*
nyxt::+renderer+
nyxt::active-buffer
nyxt::active-prompt-buffers
nyxt::buffer-hide
nyxt::fullscreen-p
nyxt::get-keymap
nyxt::javascript-error-handler
nyxt::list-commands
nyxt::make-channel
nyxt::maximized-p
nyxt::open-urls
nyxt::panel-buffers
nyxt::print-message
nyxt::prompt-buffer-p
nyxt::run-async
nyxt::status
nyxt::tls-help
nyxt::update-prompt-input
nyxt::webkit-web-context
nyxt::without-package-locks
```

aartaka:
> On another topic: these renderer packages depend on `nyxt`, so shouldn't all of those symbols be external in `nyxt`?

But then, most of those are `nyxt`-internal implementation details. Not sure if we should export those.

Ambrevar:
About Nyxt exports: some could definitely be exported.  We just got to be careful here, because something exported cannot be unexported after 3.0.

Ambrevar:
> since the prefix `nyxt/bl` most likely narrows down to a single item. On the other hand, the prefix `nyxt/mode/` would narrow to all package mode symbols.

That's not what I meant.  The prefix `nyxt/bl` completes against any Nyxt package starting with `bl`, not just modes.  `nyxt/mode/bl` completes against _modes_ starting with `bl`.  While `nyxt/mode/` lists all mode packages.

aadcg:
[Citing @Ambrevar's article](https://nyxt.atlas.engineer/article/technical-design.org), `nyxt` is a controller. Thus, it must make it easy to inter-op with any render. In other words, when I see that unexported symbols are being called on a renderer package, it raises a red flag. That's my naive understanding.

source/renderer/gtk.lisp
```diff
@@ -1,9 +1,11 @@
...
 ;;;; SPDX-License-Identifier: BSD-3-Clause
 
-(in-package :nyxt)
+(nyxt:define-package :nyxt/renderer/gtk
+    (:documentation "GTK renderer using direct CFFI bindings."))
+(in-package :nyxt/renderer/gtk)
 
-(setf +renderer+ "GTK")
+(setf nyxt::+renderer+ "GTK")
 (push :nyxt-gtk *features*)
```
aadcg, line 10: It might a completely silly idea since I don't have the big picture, but does this belong here? Shouldn't `+renderer+` be perhaps set elsewhere? Perhaps at `browser.lisp`?

Building upon what I've mentioned earlier: the fact that these renderer packages rely on unexported raises a flat, but that doesn't mean that all of those symbols should be exported. Rather, perhaps some of the logic should be transferred somewhere else.

Ambrevar:
> when I see that unexported symbols are being called on a renderer package, it raises a red flag.

You are right, it's just that we haven't really designed a proper renderer API as of today.

source/renderer/gtk.lisp
Ambrevar: Yes, this is not so good design.

I'm wondering if we'd rather remove this `+renderer+` global and instead rely on the `renderer-browser` class name, e.g. `gtk-browser`.  Then we would add `gi-gtk-browser` (which would inherit from `gtk-browser`).

In any case, this is out of the scrope of this pull request.

aadcg: Sounds fair to me. Could you please open an issue or update an existing issue so that we don't miss this insight? I'd do it myself, but I believe you have a better picture.

aadcg:
I believe this is ready to be merged, since the concerns I have raised are outside the scope of this PR.

But there's an open question: should we address them before or after 3.0? @aartaka @Ambrevar

aartaka:
> I believe this is ready to be merged, since the concerns I have raised are outside the scope of this PR.
> 
> But there's an open question: should we address them before or after 3.0? @aartaka @Ambrevar

The API for renderers should be solidified before 3.0, because we'll have much less flexibility after the major release.

Ambrevar:
But we've got a long way to go before implementing a second renderer.  And it's too late to postpone 3.0 release.

I suggest we solidify as much as we can, but realistically, support for a second renderer will probably require 4.0.

aadcg:
If I got both of you right, it seems that we shouldn't care too much about the API at this point, since support for the 2nd renderer will take time. And it seems that @Ambrevar is implicitly saying that we shouldn't focus too much on our webengine/qt renderer (the `nyxt/qt` system). It's ok to have keep it, but 3.0 is expected to be reliably used with webkitGTK.

Ok to merge then?

aartaka:
OK to merge!

aadcg:
Merged via c1c398617.

Thanks everyone!

Ambrevar:
Well, we haven't really answered the package naming question! :p

PR #2615, opened 2022-11-01
libraries/user-interface: fix `eval' with runtime solution

createyourpersonalaccount:
As recommended by the README of spinneret:
<https://github.com/ruricolist/spinneret#interpreting-trees>

# Description

Removed `TODO` in code that asked to remove `eval`.

# Discussion

The spinneret project claims that the function `interpret-html-tree` does not support the full range of features. It should be fine for the marshalled expressions. It could be tested with a unit test.

Feel free to comment on this commit and offer tips, it's my first in the project.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/libraries/user-interface/package.lisp
+++ b/libraries/user-interface/package.lisp
@@ -11,10 +11,9 @@
    #:update
    #:text
    #:action
-   #:object-expression
+   #:to-html-string
    #:paragraph
    #:progress-bar
    #:button
    #:percentage
-   #:object-string
    #:connect))
--- a/libraries/user-interface/user-interface.lisp
+++ b/libraries/user-interface/user-interface.lisp
@@ -8,6 +8,9 @@
 (defun unique-id ()
   (format nil "ui-element-~d" (incf *id*)))
 
+(defgeneric to-html-string (object)
+  (:documentation "The string HTML representation of OBJECT."))
+
 (defclass ui-element ()
   ((id :accessor id)
    (buffer :accessor buffer :initarg :buffer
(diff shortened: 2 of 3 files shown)


aartaka:
> I foresee one problem though: I'm not sure how much control we have over the content of `object-expression`, and thus `interpret-html-tree` could potentially hit an unsupported expression.

And it eventually will :(

> @aartaka Thoughts on this?

I mean, `interpret-html-tree` is nice, but maybe we're trying to solve the wrong problem here? How about this: maybe we should define a method on interface elements that would return the exact HTML for those, no matter which way it's produced? So, instead of 
``` lisp
(defmethod object-expression ((button button))
  `(:button :id ,(id button)
            :class "button"
            :title ,(alt-text button)
            :onclick ,(action button)
            ,(text button)))
```
have
``` lisp
(defmethod object-string ((button button))
  (spinneret:with-html-string
   (:button :id (id button)
            :class "button"
            :title (alt-text button)
            :onclick (action button)
            (text button))))
```

I can vaguely recall that this `object-expression`/`object-string` dichotomy was due to `cl-markup` (our previous HTML-generation library) being extremely inflexible. And `user-interface` library wasn't updated for a loooooong time, maybe even sinse `cl-markup` times...

Or, even more radical: maybe we don't actually need `user-interface`? We have `spinneret-tags.lisp` now, and it is already full of useful widgets that are as easy to embed as writing any other HTML via Spinneret. Do we actually need to move further than that for UI?

Ambrevar:
No strong opinion, but if there is resistance for the second suggestion, then we can at least implement Artyom's first suggestion as a simple step to remove the `eval`.

createyourpersonalaccount:
I am returning the exact HTML now. I didn't know what is meant by the more radical suggestion, but I'll keep it in mind for later.

aartaka:
Thanks, @createyourpersonalaccount :)

PR #2611, opened 2022-10-31
Use nsymbols.

aartaka:
# Description

This uses the newly-abstracted `nsymbols` library to handle most of symbol resolution problems. Not much to discuss there, it's mostly deleting the code that was moved to `nsymbols`.

# Checklist:
Everything in this checklist is required for each PR.  Please do not approve a PR that does not have all of these items.

- [X] I have pulled from master before submitting this PR
- [X] There are no merge conflicts.
- [X] I've added the new dependencies as:
  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    git submodule add https://gitlab.common-lisp.net/nyxt/py-configparser _build/py-configparser
    ```
  - [x] and Guix dependencies.
- [X] My code follows the style guidelines for Common Lisp code. See:
  - [Norvig & Pitman's Tutorial on Good Lisp Programming Style (PDF)](https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf)
  - [Google Common Lisp Style Guide](https://google.github.io/styleguide/lispguide.xml)
- [X] I have performed a self-review of my own code.
- [x] My code has been reviewed by at least one peer.  (The peer review to approve a PR counts.  The reviewer must download and test the code.)
- [X] Documentation:
  - [X] All my code has docstrings and `:documentation`s written in the aforementioned style.  (It's OK to skip the docstring for really trivial parts.)
  - [X] I have updated the existing documentation to match my changes.
  - [X] I have commented my code in hard-to-understand areas.
  - [X] I have updated the `changelog.lisp` with my changes if it's anything user-facing (new features, important bug fix, compatibility breakage).
  - [X] I have added a `migration.lisp` entry for all compatibility-breaking changes.
- [X] Compilation and tests:
  - [X] My changes generate no new warnings.
  - [X] I have added tests that prove my fix is effective or that my feature works.  (If possible.)
  - [X] New and existing unit tests pass locally with my changes.

PR #2617, opened 2022-11-03
New subsystem for documentation coverage

createyourpersonalaccount:
# Description

The documentation coverage subsystem nyxt/documentation-coverage creates a file
named documentation-coverage.txt at the root directory of the Nyxt
repository. The file contains statistics on undocumented symbols of internal
packages of the nyxt system. One may view a particular package with

```
grep MYPKG: documentation-coverage.txt | fold -sw 80
```

or any other command that searches for the string `MYPKG:` in the text file.

# Discussion

This subsystem can be loaded only on a fresh lisp image. The reason is that it records the loaded packages with `list-all-packages` twice: once when the nyxt external dependencies are loaded, and once when the nyxt system is loaded. Then it performs statistics on the difference of the two lists, which contains the nyxt "internal" packages.

Another solution which will work on any lisp image would be to use `nyxt-packages`, however some internal packages are not picked up by it.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:


dcov.lisp
```diff
@@ -0,0 +1,73 @@
...
+             :documented ,documented
+             :undocumented ,undocumented))
+      (incf total)
+      (if (or (documentation s 'function)
+              (documentation s 'variable)
+              (documentation s 'structure)
+              (documentation s 'type)
+              (documentation s 'setf)
+              (documentation s 'compiler-macro)
+              (documentation s 'method-combination))
```
aartaka, line 38: Could also be `(documentation s t)` as the last option.

aartaka:
But why making it a shell script? How about following the `nyxt/build-scripts/report-unbound-exports.lisp` as an example of a proper Lisp library for CI? You can see `nyxt/.github/workflows/tests.yaml` for how it's invoked. No Quicklisp required either in this scenario :P

dcov.lisp
createyourpersonalaccount: Because `s` is a symbol I think this won't work. On sbcl it throws the warning

    WARNING: unsupported DOCUMENTATION: doc-type T for object of type SYMBOL

multiple times, messing up the output.

createyourpersonalaccount:
@aartaka Okay, thank you. I will do this.

aartaka:
> @aartaka Okay, thank you. I will do this.

Aou can actually make it a separate library and a system, if you want shell reusability. As a (shameless plug) reference, you can look at how it's made in my [Sade](https://github.com/aartaka/sade) and copy it from there, or refer to [CL Cookbook](https://lispcookbook.github.io/cl-cookbook/scripting.html#with-asdf).

createyourpersonalaccount:
@aartaka How do you propose that it becomes part of gh actions? It is triggered on every commit and shows statistics? I am looking under https://github.com/atlas-engineer/nyxt/actions/workflows/tests.yml but I get the impression that the purpose of these actions is to run error/warning tests, but this is just information. A lot of the jobs get a green mark even though they have output, for example

```
;   In UNBOUND-EXPORTS: Unused lexical variable PACKAGE
```

which I think is just hiding it. Do you think it's okay if we leave it as a script for now? I can make it an asdf subsystem as well.

I went through the asdf subsystems and wrote down all the packages that they load. I think this approach might be crude. Another idea is to have a `:before` hook on the `:nyxt` system load, which will allow me to record all packages loaded from dependencies, and then I can remove them from the output of `cl:list-all-packages`.

aartaka:
> @aartaka How do you propose that it becomes part of gh actions? It is triggered on every commit and shows statistics? I am looking under https://github.com/atlas-engineer/nyxt/actions/workflows/tests.yml but I get the impression that the purpose of these actions is to run error/warning tests, but this is just information. 

But wait, this is not just a mere information, it's a coverage report. So, if we want to stick to the good standard in terms of documentation coverage, we have to document _everything_. In this case, the dcov script can indeed be run on every commit _and error out if there is **any** undocumented exported symbol_.

> A lot of the jobs get a green mark even though they have output, for example
> 
> ```
> ;   In UNBOUND-EXPORTS: Unused lexical variable PACKAGE
> ```
> 
> which I think is just hiding it. 

This one is not critical, because it's part of test suite and not part of Nyxt code.

> Do you think it's okay if we leave it as a script for now? I can make it an asdf subsystem as well.

I'd prefer that most things that we use are in Lisp without external non-Lisp dependencies, because this way we have much more control over them, and a much more homogenous development environment.

> I went through the asdf subsystems and wrote down all the packages that they load. I think this approach might be crude. Another idea is to have a `:before` hook on the `:nyxt` system load, which will allow me to record all packages loaded from dependencies, and then I can remove them from the output of `cl:list-all-packages`.

What do you mean? If you mean listing symbols belonging to Nyxt, then you can look at the package-related code (`list-all-maybe-subpackages` to `non-nyxt-packages`) in source/command.lisp and re-use some of it in `dcov`. Also note that there's #2611 which abstracts this part into an even easier-to-reuse `nsymbols` library.

createyourpersonalaccount:
> > @aartaka How do you propose that it becomes part of gh actions? It is triggered on every commit and shows statistics? I am looking under https://github.com/atlas-engineer/nyxt/actions/workflows/tests.yml but I get the impression that the purpose of these actions is to run error/warning tests, but this is just information.
> 
> But wait, this is not just a mere information, it's a coverage report. So, if we want to stick to the good standard in terms of documentation coverage, we have to document _everything_. In this case, the dcov script can indeed be run on every commit _and error out if there is **any** undocumented exported symbol_.

Yes but new contributions will fail because there are undocumented symbols. The test I have is not sophisticated enough to look for only new symbols added by the commit.

> > I went through the asdf subsystems and wrote down all the packages that they load. I think this approach might be crude. Another idea is to have a `:before` hook on the `:nyxt` system load, which will allow me to record all packages loaded from dependencies, and then I can remove them from the output of `cl:list-all-packages`.
> 
> What do you mean? If you mean listing symbols belonging to Nyxt, then you can look at the package-related code (`list-all-maybe-subpackages` to `non-nyxt-packages`) in source/command.lisp and re-use some of it in `dcov`. Also note that there's #2611 which abstracts this part into an even easier-to-reuse `nsymbols` library.

The `list-all-maybe-subpackages` function gives me packages such as `serapeum`, `swank`, etc, which are external. What I mean is something like this

```lisp
(defvar *dependency-packages* '())
(defsystem "asdf-list-packages"
  ;; system/component metadata
  :description "Automatically list created packages."
  :long-description "This demonstrates how to list the created
packages of a system."

  :depends-on (alexandria)

  ;; files loaded by asdf for system asdf-list-packages
  :components ((:file "main"
                :description "This is the main file, it has one
package called MY-PACKAGE."))
  :perform (prepare-op :after (o c)
                       (setq *dependency-packages* (list-all-packages)))
  :perform (load-op :after (o c)
                    (format t "~{~A~%~}"
                      (remove-if (lambda (x) (member x *dependency-packages*))
                                 (list-all-packages)))))
```

which will only list the pakcages defined in the `main.lisp` file. I'm trying to do this in [./nyxt.asd] but I'm still struggling with some issue around defining a subsystem for it.

aartaka:
But... there's e.g. `nyxt-packages`...

createyourpersonalaccount:
@aartaka Ah very nice, `nyxt-packages` will work. My solution that I just force-pushed picks up some more packages, e.g. `HISTORY-TREE` which is listed in `non-nyxt-packages`. As far as I can tell, that is also Nyxt code. Do we want it in the documentation coverage?

createyourpersonalaccount:
I will be making some changes to better recognize which symbols are documented and which aren't. Feel free to postpone your review until I update.

libraries/nyxt-asdf/documentation-coverage.lisp
```diff
@@ -0,0 +1,64 @@
...
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt-asdf)
+
+(defun documentation-coverage (package)
+  "Documentation coverage data on PACKAGE.
+
+# Return value
+
+A property list with keys
```
aartaka, line 11: Maybe use multiple value returns then?

libraries/nyxt-asdf/documentation-coverage.lisp
```diff
@@ -0,0 +1,64 @@
...
+          percentage
+          package
+          undocumented))
+
+(export-always 'write-documentation-coverage)
+(defun write-documentation-coverage (packages &optional (destination t))
+  "Write a documentation coverage report for the given PACKAGES to DESTINATION.
+
+DESTINATION may be nil, t, a stream, or a string with a fill pointer."
+  (loop for (percentage package undocumented)
```
aartaka, line 52: This whole loop could've been a `(mapcar (lambda (package) (multiple-values-list (documentation-coverage package))) packages)` if you used multiple value returns. Why this exact design then?

aartaka (commented):
I mean, it looks OK, but the whole ASDF tinkering is not cool. Sorry for not answering your remark about `history-tree` earlier: `history-tree` and other Nyxt libraries are all listed in `*nyxt-extra-packages*` variable, so you can easily list all the packages to test with 
``` lisp
(append nyxt::*nyxt-extra-packages* (nyxt::nyxt-packages))
```

aartaka:
> @aartaka after some thought, I decided to keep it as-is. The difficulty was in detecting slot documentation, but I decided to skip it entirely on the idea that accessors should be documented instead. What do you think about this?

Slot documentation is perfectly detectable (although somewhat exponential in complexity). See describe.lisp for how we list all package slots with their respective classes. From this, you can do `mopu:class-slots` and then use e.g. `slynk-mop:slot-definition-documentation` on it.

> Also, I have noticed that for example my code detects [`mark-set`](https://github.com/atlas-engineer/nyxt/blob/b407778c0fb1265d0fc7c993011f277a6c96cc96/source/mode/visual.lisp#L78) from https://github.com/atlas-engineer/nyxt/blob/master/source/mode/visual.lisp as undocumented. I am not sure why, do you know why it would be displayed under exported symbols? Perhaps because it is a symbol and all mode symbols are exported?

Because it's an exported slot that has no documentation :)

> Because of this issue, it will be difficult to make it a github action; it would complain about undocumented symbols that the user can't document anyway...

Which ones, for example? CL documentation system is pretty comprehensive, so I doubt we can't document something because of language limitations.

createyourpersonalaccount:
> I mean, it looks OK, but the whole ASDF tinkering is not cool.

What don't you like, is it the added `extra-dependencies` subsystem or that I added a subsystem in nyxt-asdf?

>Sorry for not answering your remark about `history-tree` earlier: `history-tree` and other Nyxt libraries are all listed in `*nyxt-extra-packages*` variable, so you can easily list all the packages to test with
> 
> ```lisp
> (append nyxt::*nyxt-extra-packages* (nyxt::nyxt-packages))
> ```

Some seem to be missing, like:

```
ACCLIMATION
CL-CUSTOM-HASH-TABLE
CLUFFER
CLUFFER-BASE
CLUFFER-INTERNAL
CLUFFER-SIMPLE-BUFFER
CLUFFER-SIMPLE-LINE
CLUFFER-STANDARD-BUFFER
CLUFFER-STANDARD-LINE
CLUMP-2-3-TREE
CLUMP-BINARY-TREE
NYXT-USER
PROMPTER
THEME
USER-INTERFACE
```

Do we want these to be listed in `nyxt::*nyxt-extra-packages*`?

createyourpersonalaccount:
@aartaka 
> Slot documentation is perfectly detectable (although somewhat exponential in complexity). See describe.lisp for how we list all package slots with their respective classes. From this, you can do `mopu:class-slots` and then use e.g. `slynk-mop:slot-definition-documentation` on it.

Okay, I will do it that way, you're right it's better.

aartaka:
> > I mean, it looks OK, but the whole ASDF tinkering is not cool.
> 
> What don't you like, is it the added `extra-dependencies` subsystem or that I added a subsystem in nyxt-asdf?

I don't like that there's any change in ASDF-related file. They are much harder to debug than regular Lisp files. If we can avoid editing those (and we can, in this particular case)—we should avoid eiditing them.

> Some seem to be missing, like:
> 
> ```
> ACCLIMATION
> CL-CUSTOM-HASH-TABLE
> CLUFFER
> CLUFFER-BASE
> CLUFFER-INTERNAL
> CLUFFER-SIMPLE-BUFFER
> CLUFFER-SIMPLE-LINE
> CLUFFER-STANDARD-BUFFER
> CLUFFER-STANDARD-LINE
> CLUMP-2-3-TREE
> CLUMP-BINARY-TREE
> NYXT-USER
> PROMPTER
> THEME
> USER-INTERFACE
> ```
> 
> Do we want these to be listed in `nyxt::*nyxt-extra-packages*`?

`theme`, `prompter`, and `user-interface`—yes. Others—no, as they are not specific to Nyxt and we don't exactly control them.

createyourpersonalaccount:
> > > I mean, it looks OK, but the whole ASDF tinkering is not cool.
> > 
> > 
> > What don't you like, is it the added `extra-dependencies` subsystem or that I added a subsystem in nyxt-asdf?
> 
> I don't like that there's any change in ASDF-related file. They are much harder to debug than regular Lisp files. If we can avoid editing those (and we can, in this particular case)—we should avoid eiditing them.

Okay, it's true that with `nyxt::...` utilities for loaded packages it can be avoided. I can do it that way. 

> > Do we want these to be listed in `nyxt::*nyxt-extra-packages*`?
> 
> `theme`, `prompter`, and `user-interface`—yes. Others—no, as they are not specific to Nyxt and we don't exactly control them.

Okay, I'll look into that

createyourpersonalaccount:
> > > > I mean, it looks OK, but the whole ASDF tinkering is not cool.
> > > 
> > > 
> > > What don't you like, is it the added `extra-dependencies` subsystem or that I added a subsystem in nyxt-asdf?
> > 
> > 
> > I don't like that there's any change in ASDF-related file. They are much harder to debug than regular Lisp files. If we can avoid editing those (and we can, in this particular case)—we should avoid eiditing them.
> 
> Okay, it's true that with `nyxt::...` utilities for loaded packages it can be avoided. I can do it that way.
> 
> > > Do we want these to be listed in `nyxt::*nyxt-extra-packages*`?
> > 
> > 
> > `theme`, `prompter`, and `user-interface`—yes. Others—no, as they are not specific to Nyxt and we don't exactly control them.
> 
> Okay, I'll look into that

@aartaka Now that I looked at it some more, I realized that `*nyxt-extra-packages*` is hard-coded into the source. Don't you think the subsystem solution is nicer? It auto-detects everything. I don't know yet why some 3rd party packages are detected (like acclimation).

libraries/nyxt-asdf/documentation-coverage.lisp
aartaka: Friendly ping: Using multiple values will shorten your code two-fold, if I'm not missing something :P

nyxt.asd
```diff
@@ -294,6 +297,50 @@
...
+documentation is ignored, instead it is recommended to document the
+accessors via `defgeneric'.
+
+This subsystem must be loaded on a fresh lisp image. The output is in
+the file documentation-coverage.txt found in the root directory of the
+distribution. The file is replaced if it exists.
+
+To search for a particular package (on unix) one may use
+
+    grep MYPKG: documentation-coverage.txt | fold -sw 80
```
aartaka, line 318: But why generate it, if you can either:
- Write it to stdout, or
- Return all the requried values from the `documentation-coverage` function?

What I keep hinting at is that Lisp-based software needs no bash, grep, and files—Lisp has lots of data structures and language tools to pass things around and to  get to useful results. Using OS-specific scripting languages and other parts of the OS is almost never necessary (let alone the fact that it hinders portability to unusual OS-es) :P

aartaka:
@createyourpersonalaccount, this particular task requires a manual approach, given that some of our dependencies we don't control and we need to allow-list them first. Merely updating `*nyxt-extra-packages*` should be enough and would provide more control than ASDF subsystem does. Especially given that you're testing coverage in _packages_, while ASDF manages _systems_, and those can be wildly different in nature and contents.

Also note that you're stripping away from `"nyxt"` system the vital piece of information: its dependencies. Putting those in a separate place may confuse someone packaging Nyxt and makes our (already complicated) ASD file harder to read :P

aadcg:
I'm taking a first look now.

Let me ask conceptual questions. How is this metric important for us? Do you think it will force us to be more serious writing docstrings?

createyourpersonalaccount:
@aadcg I made the program to check where the interface is undocumented, so that I can find places to contribute (for example https://github.com/atlas-engineer/nyxt/pull/2639 documents just two files). The suggestion to incorporate it with the CI/CD would be for making sure that all commits contain documentation, but it is not as important in my opinion.

createyourpersonalaccount:
@aartaka okay, I will create a function that prints the results and add the missing packages in `*nyxt-extra-packages*`, and then I will leave it up to whoever wishes to run the function to see its results.

aartaka:
> @aartaka okay, I will create a function that prints the results and add the missing packages in `*nyxt-extra-packages*`, and then I will leave it up to whoever wishes to run the function to see its results.

Yup, works too. If you don't mind, I'll take over this PR after that :)

aadcg:
> @aadcg I made the program to check where the interface is undocumented, so that I can find places to contribute (for example #2639 documents just two files). The suggestion to incorporate it with the CI/CD would be for making sure that all commits contain documentation, but it is not as important in my opinion.

Clear. I think such a utility is useful, but I don't think it belongs to CI/CD as I'm sure there will always be some kind of corner case where documentation is not really needed. However, being able to load a system and getting such a report is indeed an interesting concept.  Thank you for working on this.

I should mention that we do have documentation for most things in our codebase. Still, it needs to be reviewed and improved. The same can't be said regarding testing. Our coverage ratio is frankly low (another suggestion of a small utility that we could benefit from). 

Regarding the technicalities of this PR, I agree with all points @aartaka raises and I'm sure he's the best person to finish the review with you.

createyourpersonalaccount:
@aadcg Good idea about the test coverage!

createyourpersonalaccount:
@aartaka You can take over now, use `write-documentation-coverage` in the new file `source/documentation-coverage.lisp` to get the report. Class slots are not tested for documentation as-is.

jmercouris:
Thank you Nick, I am eager to see this merged!

aartaka:
Okay, I've simplified documentation-coverage.lisp and moved it to build-scripts/, where it belongs better than source/. I've also allowed myself a liberty of squashing the commits in one and merging as 78fa0f1614f6bda4c7643629e9999070a2057dc0.

aadcg:
Compiling the merged file fails with:

```log
documentation-coverage.lisp:12:3:
  warning: 
    undefined variable: NYXT::IT
    --> LET WITH-PACKAGE-ITERATOR MULTIPLE-VALUE-BIND MULTIPLE-VALUE-CALL 
    --> FUNCTION LET MACROLET SB-LOOP::WITH-SUM-COUNT LET 
    --> SB-LOOP::WITH-SUM-COUNT LET SB-LOOP::WITH-LOOP-LIST-COLLECTION-HEAD 
    --> LET* TAGBODY WHEN 
    ==>
      (IF NYXT::IT
          (SETQ NYXT::TOTAL (1+ NYXT::TOTAL)))
    

Compilation failed.
```

aartaka:
Fixed in 36f4e11b9b45e4b6184a2f8d12e22761767c186c.

aadcg:
I think this file needs instructions at the top on how to use the script.

aadcg:
@createyourpersonalaccount @aartaka please provide me an example on how to use it :)

aartaka:
@aadcg, compile the code in the file and do `(write-documentation-coverage (list :nyxt))`.

aadcg:
```lisp
(write-documentation-coverage '(:nyxt/hint-mode))
=>
52% NYXT/HINT-MODE: AUTO-FOLLOW-HINTS-P COMPUTE-HINTS-IN-VIEW-PORT-P FIT-TO-PROMPT-P HIGHLIGHT-SELECTED-HINT HINTS-ALPHABET HINTS-SELECTOR IDENTIFIER SHOW-HINT-SCOPE-P UNHIGHLIGHT-SELECTED-HINT
```

It seems that class slots are marked as undocumented, when in fact they are documented.

Would it be saner to keep each undocumented symbol is a new line?

aartaka:
Slot documentation detection fixed in 81a3348ee9413f5d61435c851aba69f7aaa738f5.

Another bug I noticed: if a generic function is implicitly defined via `defmethod`, then the documentation of `defmethod` does not propagate to `defgeneric`. Which is intuitive, but hints us that we should also look at method documentation.

No, new lines are not cool, especially mandatory.

aadcg:
Thanks @aartaka.

Sorry that I've started to look deeper into this now. @aartaka makes a good point that this checks for documentation coverage in packages, not systems. But, actually, why do we care about packages? Why not doing it exclusively for systems?

CC @createyourpersonalaccount

createyourpersonalaccount:
The check is for undocumented symbols that are external in the nyxt package or any of its package dependencies in `nyxt::*nyxt-extra-packages*` or in `nyxt::nyxt-packages`.

Do you mean that it should look for all packages created by a given system? It is also possible. This tool can be done in a number of ways...

aadcg:
My view is that is would be nice if there would be `:perform` actions on each of our systems that would run this report. But I don't know ASDF well enough...

PR #2621, opened 2022-11-05
renderer/gtk: Fix exit-code slots

hendursaga:
This fixes the following bug, in addition to the similar one in `gtk.lisp`:

```
<WARN> [17:07:37] Warning: Error: When attempting to read the slot's value (slot-value), the slot
NYXT/RENDERER/GI-GTK::EXIT-CODE is missing from the object
#&lt;BROWSER {100B5BCE13}&gt;.
It has a slot NYXT::EXIT-CODE, while NYXT/RENDERER/GI-GTK::EXIT-CODE is requested.
```

### DIFF
--- a/source/renderer/gi-gtk.lisp
+++ b/source/renderer/gi-gtk.lisp
@@ -46,7 +46,7 @@ interface. On Darwin, we must run the GTK thread on the main thread."
           (unless nyxt::*run-from-repl-p*
             (bt:join-thread main-thread)
             ;; See comment about FreeBSD in gtk.lisp
-            (uiop:quit (slot-value browser 'exit-code) #+freebsd nil)))
+            (uiop:quit (slot-value browser 'nyxt::exit-code) #+freebsd nil)))
         #+darwin
         (main-func)))
 
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -275,7 +275,7 @@ the renderer thread, use `defmethod' instead."
           ;; UIOP:QUIT with FINISH-OUTPUT = NIL in the first place.
           ;;
           ;; FIXME: This also can be true for other BSD systems.
-          (uiop:quit (slot-value browser 'exit-code) #+freebsd nil)))
+          (uiop:quit (slot-value browser 'nyxt::exit-code) #+freebsd nil)))
       #+darwin
       (progn
         (setf gtk-running-p t)


createyourpersonalaccount:
Can't you use `nyxt:exit-code` with one colon instead?

aartaka:
> Can't you use `nyxt:exit-code` with one colon instead?

No, because `nyxt::exit-code` has `:export nil` in the slot definition, which makes it unexported.

aartaka:
Thanks @hendursaga!

aadcg:
I believe there is a misunderstand here.

@createyourpersonalaccount raises an important question. @aartaka, it doesn't matter whether the symbols is exported. It does matter if you were to invoke `(exit-code *browser*)`, but `(slot-value *browser* 'exit-code)` is valid. 

Proof:
- `(asdf:load-system :nyxt/gi-gtk)`
- `(nyxt:start)`
- `(slot-value *browser* 'exit-code)` returns gracefully.

Another proof is commit 9aa384f6b0c036db7abd228fc9ad72c99fdbf91f by @Ambrevar.

So, I'm actually not sure what problem @hendursaga is trying to solve since renderer packages `:use` the `nyxt` package. @hendursaga can you provide a recipe of how you were getting the mentioned warning?

createyourpersonalaccount:
@hendursaga If it is used by others, I think it should be exported, but it's not clear to me and so I will think about this later if I have more insight of the situation.

aartaka:
> @hendursaga If it is used by others, I think it should be exported, but it's not clear to me and so I will think about this later if I have more insight of the situation.

See #2612 discussion between @Ambrevar and @aadcg, with the bottom line being:
```
But we've got a long way to go before implementing a second renderer. And it's too late to postpone 3.0 release.

I suggest we solidify as much as we can, but realistically, support for a second renderer will probably require 4.0.
```
Which means: we have to live with API instability and use of unexported symbols in renderer code, because we won't have a clear API understanding until we revive a second renderer. Which means: it's okay to use unexported symbols in renderer-specific code!

aadcg:
@hendursaga, that is obvious! That's why I am asking you what problem you're trying to solve. You're not supposed to have access `*browser*` internals from renderer packages. 

If that warning you mention was raised when you were Nyxt, then we need to address it. I just don't see how it could have possible have happened!

I just as you've mentioned and I got no warning at all.

@createyourpersonalaccount  it is not exported for the reasons @hendursaga mention.

@Ambrevar please take a look. I think we should revert this commit.

aadcg:
> Which means: we have to live with API instability and use of unexported symbols in renderer code, because we won't have a clear API understanding until we revive a second renderer. Which means: it's okay to use unexported symbols in renderer-specific code!

This is true, but it doesn't apply here.  The case we face here is actually the opposite. We had unexported symbols or, in other words, a well-thought out private API and we've just made it public for obscure reasons...

Ambrevar:
> @Ambrevar please take a look. I think we should revert this commit.

You mean revert this pull request?  I don't understand why, this patch seems correct to me.

aadcg:
There's nothing incorrect here. But it's a no-op that obcures the API... 

When I define a function, I do it with `defun`, not with `common-lisp:defun` right?

aadcg:
To sum up, I'm saying that the three are equivalent because renderer packages `:use` the `nyxt` package.

- `(slot-value *browser* 'exit-code)`
- `(slot-value *browser* 'nyxt:exit-code)`
- `(slot-value *browser* 'nyxt::exit-code)`

Notice that `(slot-value *browser* 'exit-code)` and `(exit-code *browser*)` are two different incantations. For the latter you are required to export `exit-code`.

hendursaga:
@aadcg maybe we're using two different SBCL versions???! I'm using `2.2.10`. And both of the `ffi-initialize` methods aren't accessing `*browser*`, they're explicitly passed a `browser` instance..

Am I the only person who can reproduce my issue??

I'll defer to @Ambrevar's judgment. Otherwise, I'm going to have to tap out of this one.

aadcg:
Anyway, this is not very important and all of @hendursaga's feedback is very welcome :)

aadcg:
> That it uses :nyxt makes no difference, since the exit-code slot is unexported. use-ing does not make unexported symbol accessible.

True but irrelevant.

You don't need to export `exit-code` when you access it via `slot-value`. But I'm repeating myself at this point... 

Again, doesn't matter much.

aartaka:
> You don't need to export `exit-code` when you access it via `slot-value`. But I'm repeating myself at this point...

Wait, what? If you don't export slot name, you still need to use the unexported prefix...

``` lisp
(uiop:define-package :hello)
(in-package :hello)
(defclass foo () ((what :initform 5)))
(in-package :cl-user)
(defvar foo (make-instance 'hello::foo))
(slot-value foo 'hello:what)
; Debugger entered on #<SB-INT:SIMPLE-READER-PACKAGE-ERROR "The symbol ~S is not external in the ~A package." {1005D2FF63}>
(slot-value foor 'hello::what)
; => 5 (3 bits, #x5, #o5, #b101)
```

Ambrevar:
> You don't need to export `exit-code` when you access it via `slot-value`. But I'm repeating myself at this point...

This is not true.

PR #2625, opened 2022-11-06
Review theme library

aadcg:
# Description

Expand README and add test suite to make sure we don't forget about the invariants of the theme library.

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/libraries/theme/tests/test.lisp
+++ b/libraries/theme/tests/test.lisp
@@ -59,3 +59,46 @@ h1 { color: magenta; }
                                 theme:background
                                 theme:on-background)
                      :background-color theme:primary))))
+
+(defun hex-to-rgb (hex)
+  "Convert HEX to an RGB triple."
+  (declare (string hex))
+  (let ((hex (if (uiop:string-prefix-p "#" hex) (subseq hex 1) hex)))
+    (mapcar (lambda (x) (/ x 255.0))
+            (list (parse-integer (subseq hex 0 2) :radix 16)
+                  (parse-integer (subseq hex 2 4) :radix 16)
+                  (parse-integer (subseq hex 4 6) :radix 16)))))
+
+(defun relative-luminance (hex)
+  "Compute relative luminance of HEX."
+  ;; See https://www.w3.org/TR/WCAG20-TECHS/G18.html
+  (loop for const in '(0.2126 0.7152 0.0722)
+        for rgb-component in (hex-to-rgb hex)
+        sum (* const (if (<= rgb-component 0.03928)
+                         (/ rgb-component 12.92)
+                         (expt (/ (+ rgb-component 0.055) 1.055) 2.4)))))
+
+(defun contrast-ratio (hex1 hex2)
+  "Compute contrast ratio between HEX1 and HEX2."
+  (let ((ratio (/ (+ (relative-luminance hex1) 0.05)
+                  (+ (relative-luminance hex2) 0.05))))
+    (max ratio (/ ratio))))
+
+(defvar *minimum-contrast-ratio* 7.0
+  "The minimum contrast ratio required between color and on-color.")
+
+(define-test contrast-ratio-between-color-and-on-color ()
+  (flet ((assert-contrast-ratio (hex)
+           (assert-true
+            ;; This avoids handling the fact that on-colors are set as color
+            ;; strings (i.e. "black" and "white")
+            (or (> (contrast-ratio hex "ffffff") *minimum-contrast-ratio*)
+                (> (contrast-ratio hex "000000") *minimum-contrast-ratio*)))))
+    ;; No need to test the ratio between background and on-background since it's
+    ;; trivially close to the largest possible value of 21:1.
+    (assert-contrast-ratio (theme:primary-color theme:+light-theme+))
+    (assert-contrast-ratio (theme:secondary-color theme:+light-theme+))
+    (assert-contrast-ratio (theme:accent-color theme:+light-theme+))
+    (assert-contrast-ratio (theme:primary-color theme:+dark-theme+))
+    (assert-contrast-ratio (theme:secondary-color theme:+dark-theme+))
+    (assert-contrast-ratio (theme:accent-color theme:+dark-theme+))))
--- a/libraries/theme/README.org
+++ b/libraries/theme/README.org
@@ -1,9 +1,77 @@
-#+TITLE:theme library for Nyxt
+#+TITLE: Theme library for Nyxt
 
-Based on Google Material Design Guidelines, we've decided to have consistent themes working all across Nyxt. These themes consist of contrasting pairs of colors, mirroring Google's "on-*" colors approach. We've ended up with 4 pairs:
-- ~background-color~ and ~on-background-color~ :: background and text colors, respectively.
-- ~primary-color~ and ~on-primary-color~ :: the color of interactive interface elements and the most neutral palette color we can find. ~on-primary-color~ is, respectively, the color of the text on interface elements.
-- ~secondary-color~ and ~on-secondary-color~ :: the color that more or less contrasts with ~primary-color~ and can be used for more varied interface styles. With its own ~on-*~ color.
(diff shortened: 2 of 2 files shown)


jmercouris:
Brilliant! I like!

libraries/theme/tests/test.lisp
```diff
@@ -59,3 +59,18 @@ h1 { color: magenta; }
                                 theme:background
                                 theme:on-background)
                      :background-color theme:primary))))
+
+(defvar *minimum-contrast-ratio* 7.0
+  "The minimum contrast ratio required between color and on-color.")
+
+(define-test contrast-ratio-between-color-and-on-color ()
```
aartaka, line 66: But wait, should we maybe compare the `*-color` and `on-*-color` instead just the colors against balck and white?

libraries/theme/theme.lisp
```diff
@@ -160,3 +160,27 @@ headings have border of secondary color.
...
+            (list (parse-integer (subseq hex 0 2) :radix 16)
+                  (parse-integer (subseq hex 2 4) :radix 16)
+                  (parse-integer (subseq hex 4 6) :radix 16)))))
+
+(defun relative-luminance (hex)
+  "Compute relative luminance of HEX."
+  ;; See https://www.w3.org/TR/WCAG20-TECHS/G18.html
+  (loop for const in '(0.2126 0.7152 0.0722)
+        for rgb-coord in (hex-to-rgb hex)
+        sum (* const (if (<= rgb-coord 0.03928)
```
aartaka, line 178: This feels like an abstractable (`defun`? `flet`?) function of a descriptive name rather than a mere expression.

libraries/theme/theme.lisp
```diff
@@ -160,3 +160,27 @@ headings have border of secondary color.
...
+  (let ((hex (if (eq 7 (length hex)) (subseq hex 1) hex)))
+    (mapcar (lambda (x) (/ x 255.0))
+            (list (parse-integer (subseq hex 0 2) :radix 16)
+                  (parse-integer (subseq hex 2 4) :radix 16)
+                  (parse-integer (subseq hex 4 6) :radix 16)))))
+
+(defun relative-luminance (hex)
+  "Compute relative luminance of HEX."
+  ;; See https://www.w3.org/TR/WCAG20-TECHS/G18.html
+  (loop for const in '(0.2126 0.7152 0.0722)
```
aartaka, line 176: Will `(reduce #'+ (mapcar (lambda ...) '(0.2126 ...) (hex-to-rgb hex)))` be shorter and nicer here?

libraries/theme/theme.lisp
```diff
@@ -160,3 +160,27 @@ headings have border of secondary color.
   `(with-theme ,theme
     (cl-css:css
      (list ,@(mapcar #'requote-rule rules)))))
+
+(defun hex-to-rgb (hex)
+  "Convert HEX to an RGB triple."
+  (declare (string hex))
+  (let ((hex (if (eq 7 (length hex)) (subseq hex 1) hex)))
```
aartaka, line 167: Well that's quite a loose assumption. Maybe do `(uiop:string-prefix-p "#" hex)` and `(ignore-errors (parse-integer ...))`?

Alll in all, only parsing hex feels incomplete, because we may also have `rgb(X, X, X, X)`, mere color names, and other CSS functions. And supporting at least RGB is necessary, because that's quite an frequent way to set colors too.

aartaka (commented):
It's definitely an improvement, but what I'd also want to see here is:
- A more reliable HEX detection.
- RGB parsing.
- Maybe color name constants (see the list in renderer/gtk.lisp, you just have to find their values).

For now, the heuristics for HEX detection and the assumptions behind the test set are too loose, I believe.

libraries/theme/tests/test.lisp
aadcg: You're obviously correct, but we can change that when our `on-color`s will be something other than white or black, which is unlikely anytime soon...

libraries/theme/theme.lisp
aadcg: I disagree because I don't care about that about any of the parts of `relative-luminance`. It's just some random computation that I want to hide.

libraries/theme/theme.lisp
aadcg: That is a viable option but I have doubts it will look less nasty. I can try, but is it worth it?

libraries/theme/theme.lisp
aadcg: Good point regarding the loose assumption.

Absolutely, but what do we gain from all of that right now? This only matters for the default themes we provide. It's literally 4 colors for each theme and we always encode those as hex.

aadcg:
Overall I agree with everything you mention @aartaka. But the goal here is to prevent us from forgetting why we have settled on the default theme the colors. 

I will write down all of the gotchas and TODOs for the future.

aadcg:
> For now, the heuristics for HEX detection and the assumptions behind the test set are too loose, I believe.

Let me stress this again.  We're not testing over any theme that users might come up with. We're testing the 2 singles themes (light and dark) that we have chosen. Will we define those with both RGB and HEX? Unlikely. Will we use random strings like "LavenderBlush"? Unlikely. Will we use smth other than white and black for on colors? Unlikely. Even if we do, there's no need to have a full blown library to handle all of that.  Besides, that library may exist.

libraries/theme/tests/test.lisp
aadcg: If you ask me, I'd just use HEX for all or our default themes so that we don't have these annoying "white" and "black". But I know you'd not approve :D

libraries/theme/theme.lisp
aadcg:
```
(defun relative-luminance (hex)
  "Compute relative luminance of HEX."
  ;; See https://www.w3.org/TR/WCAG20-TECHS/G18.html
  (reduce #'+ (mapcar #'*
                      '(0.2126 0.7152 0.0722)
                      (mapcar (lambda (rgb-component)
                                (if (<= rgb-component 0.03928)
                                    (/ rgb-component 12.92)
                                    (expt (/ (+ rgb-component 0.055) 1.055) 2.4)))
                              (hex-to-rgb hex)))))
```

Do you prefer it?

aadcg:
Given @aartaka comments, one thing that could be done would be to move the utilities relative to computing the contrast ratio to the tests package. The reason why I have not done so is because it's still marginally useful for other purposes. Say I'm an advanced user and I'm crafting my theme. Then I could perhaps make use of those utilities.   

No strong opinion from my side.

libraries/theme/tests/test.lisp
aartaka: > You're obviously correct, but we can change that when our `on-color`s will be something other than white or black, which is unlikely anytime soon...

Fair.

aartaka:
> Given @aartaka comments, one thing that could be done would be to move the utilities relative to computing the contrast ratio to the tests package. 

Yes!

> The reason why I have not done so is because it's still marginally useful for other purposes. Say I'm an advanced user and I'm crafting my theme. Then I could perhaps make use of those utilities.

If so, these utilities should be more general, as I'm suggesting! There are lots of ways people create themes in, and locking them to (arguably) relatively opaque HEX is not cool :(

aadcg:
> If so, these utilities should be more general, as I'm suggesting! There are lots of ways people create themes in, and locking them to (arguably) relatively opaque HEX is not cool :(

It doesn't influence the decision to move the utils to the tests packages but I think this is not the right way to look at it. The theme library promises to ease the process of crafting themes. The utilities don't frustrate any of that. On the contrary, it provides an extra tool for theme crafters aiming at a certain contrast standard.  In effect, we'd be giving the users the ability to reason about themes the way we did. These would be unexported of course.

aadcg:
@aartaka did you read the README? I made changes to the original you had written.

aartaka:
> @aartaka did you read the README? I made changes to the original you had written.

README is fine, yes.

aadcg:
Thanks!

aartaka:
Thank you for sustaining through this, André `(´∀｀)♡`

aadcg:
It's me being annoying :D

All arguments and counter-arguments are healthy. I'm happy to say that it's often the case that someone convinces me otherwise, or proves me plain wrong!

PR #2626, opened 2022-11-07
Use LASS in libraries/theme.

aartaka:
# Description

This is a prototype to gather feedback on whether we want LASS in our themes. The benefits are:
- Less macrology (we can remove `theme::requote-rule` and I'd be glad when this abomination is gone).
- Support for media queries and a more powerful rule syntax with `:and` and `:or` selector composition.
- Built-in value injection with `#(syntax)`.
The downsides are:
- We need to quote every rule (this might change if I research it some more).
- The syntax for value injection if not familiar.
- We now have to use quasi-quotes to allow inline code evaluation.
  - Is it a downside, actually? Especially given that there's no more `requore-rule` that was enabling the inline evaluation, but was sacrifising lots of flexibility.

# Discussion

@aadcg, @Ambrevar, @jmercouris, does `theme:themed-css2` look right to you? If it does, I'll replace `themed-css` with it, use it everywhere, and send a patch with LASS to Guix.

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [X] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/libraries/theme/theme.lisp
+++ b/libraries/theme/theme.lisp
@@ -151,13 +151,13 @@ headings have border of secondary color.
                             :dark-p t
                             :on-background-color \"red\"
                             :accent-color \"blue\")
-           '(|h1,h2,h3,h4,h5,h6|
+           `(|h1,h2,h3,h4,h5,h6|
              :border-style \"solid\"
              :border-width \"1px\"
-             :border-color #(theme:secondary))
+             :border-color ,theme:secondary)
            `(p
              :color ,(if (theme:dark-p theme:theme) theme:accent theme:secondary)
-             :background-color #(theme:background)))"
+             :background-color ,theme:background))"
   `(with-theme ,theme
      (funcall #'%themed-css theme:theme
               ;; NOTE: This loop allows to omit quotes for most trivial rules,
--- a/libraries/theme/theme.lisp
+++ b/libraries/theme/theme.lisp
@@ -110,19 +110,9 @@ out from all of the other theme colors.")
      ,@body))
 
 (defmacro themed-css (theme &body rules)
-  "Generate a CSS styled according to the THEME.
+  "Generate a CSS styled according to the THEME and RULES.
 
-RULES is a list of LASS rules. There are special symbols that this macro
-:let-binds with LASS:
-- `theme:background';
-- `theme:on-background';
-- `theme:primary';
-- `theme:on-primary';
-- `theme:secondary';
-- `theme:on-secondary';
-- `theme:accent';
-- `theme:on-accent';
-- `theme:font-family';
+RULES is a list of LASS rules.
 
 Any LASS-friendly syntax (including quasi-quotes) works, so you can evaluate
 arbitrary code before the rules are compiled. For the convenience of
--- a/libraries/theme/theme.lisp
+++ b/libraries/theme/theme.lisp
@@ -109,38 +109,39 @@ out from all of the other theme colors.")
           (theme:font-family (font-family theme:theme)))
      ,@body))
 
-(defun plist-p (object)
-  "Return non-nil if OBJECT is a plist."
-  (and (listp object)
-       (alexandria:proper-list-p object)
-       (evenp (length object))
-       (loop :for x :in object :by #'cddr
-             :always (keywordp x))))
-
-(defun requote-rule (rule)
-  (if (plist-p (rest rule))
-      (cons 'list (mapcar (lambda (elem)
-                            (typecase elem
-                              (symbol (if (equalp "theme" (package-name (symbol-package elem)))
-                                          elem
-                                          `(quote ,elem)))
-                              (atom (if (constantp elem)
-                                        elem `(quote ,elem)))
-                              (list elem)
-                              (t elem)))
-                          rule))
-      (cons 'list (mapcar (lambda (x) `(quote ,x)) rule))))
+(defun %themed-css (theme &rest rules)
+  (lass:compile-and-write
+   (append `(:let
+                ,(list (list (quote theme:background) (background-color theme))
+                       (list (quote theme:on-background) (on-background-color theme))
+                       (list (quote theme:primary) (primary-color theme))
+                       (list (quote theme:on-primary) (on-primary-color theme))
+                       (list (quote theme:secondary) (secondary-color theme))
+                       (list (quote theme:on-secondary) (on-secondary-color theme))
+                       (list (quote theme:accent) (accent-color theme))
+                       (list (quote theme:on-accent) (on-accent-color theme))
+                       (list (quote theme:font-family) (font-family theme))))
+           rules)))
 
 (defmacro themed-css (theme &body rules)
   "Generate a CSS styled according to the THEME.
 
-RULES is a list of CL-CSS rules (with one level of nesting removed). There are
-special symbols that this macro will substitute for theme elements, if
-encountered. See `with-theme' for a list of those.
+RULES is a list of LASS rules. There are special symbols that this macro
+:let-binds with LASS:
+- `theme:background';
+- `theme:on-background';
+- `theme:primary';
+- `theme:on-primary';
+- `theme:secondary';
+- `theme:on-secondary';
+- `theme:accent';
+- `theme:on-accent';
+- `theme:font-family';
 
-Any non-atomic s-expression put as value of CSS property will be
-evaluated too, so you can put arbitrarily complex expressions as
-property values.
+Any LASS-friendly syntax (including quasi-quotes) works, so you can evaluate
+arbitrary code before the rules are compiled. For the convenience of
+quasi-quoted forms, this macro let-binds the set of symbols above around the
+RULES.
 
 Example: color all the paragraph text in accent color if the theme is dark, and
 in secondary color otherwise. Use the text color as background color. Make
@@ -150,13 +151,28 @@ headings have border of secondary color.
                             :dark-p t
                             :on-background-color \"red\"
                             :accent-color \"blue\")
-           (|h1,h2,h3,h4,h5,h6|
-            :border-style \"solid\"
-            :border-width \"1px\"
-            :border-color theme:secondary)
-           (p
-            :color (if (theme:dark-p theme:theme) theme:accent theme:secondary)
-            :background-color theme:background))"
+           '(|h1,h2,h3,h4,h5,h6|
+             :border-style \"solid\"
+             :border-width \"1px\"
+             :border-color #(theme:secondary))
+           `(p
+             :color ,(if (theme:dark-p theme:theme) theme:accent theme:secondary)
+             :background-color #(theme:background)))"
   `(with-theme ,theme
-    (cl-css:css
-     (list ,@(mapcar #'requote-rule rules)))))
+     (funcall #'%themed-css theme:theme
+              ;; NOTE: This loop allows to omit quotes for most trivial rules,
+              ;; mostly preserving backwards-compatibility.
+              ,@(loop for rule in rules
+                      for first = (first rule)
+                      ;; FIXME: This is not perfect, but it's good enough for
+                      ;; 99% of cases. Maybe somehow parse selectors with LASS?
+                      if (or (stringp first)
+                             (keywordp first)
+                             (eq '* first)
+                             (and (symbolp first)
+                                  (eq :internal (nth-value 1 (find-symbol (symbol-name first)
+                                                                          (symbol-package first)))))
+                             (and (listp first)
+                                  (keywordp (first first))))
+                        collect (cons 'quote (list rule))
+                      else collect rule))))
--- a/libraries/theme/theme.lisp
+++ b/libraries/theme/theme.lisp
@@ -109,20 +109,6 @@ out from all of the other theme colors.")
           (theme:font-family (font-family theme:theme)))
      ,@body))
 
-(defun %themed-css (theme &rest rules)
-  (lass:compile-and-write
-   (append `(:let
-                ,(list (list (quote theme:background) (background-color theme))
-                       (list (quote theme:on-background) (on-background-color theme))
-                       (list (quote theme:primary) (primary-color theme))
-                       (list (quote theme:on-primary) (on-primary-color theme))
-                       (list (quote theme:secondary) (secondary-color theme))
-                       (list (quote theme:on-secondary) (on-secondary-color theme))
-                       (list (quote theme:accent) (accent-color theme))
-                       (list (quote theme:on-accent) (on-accent-color theme))
-                       (list (quote theme:font-family) (font-family theme))))
-           rules)))
-
 (defmacro themed-css (theme &body rules)
   "Generate a CSS styled according to the THEME.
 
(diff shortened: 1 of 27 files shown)


aadcg:
> Less macrology (we can remove theme::requote-rule and I'd be glad when this abomination is gone)

This makes me very happy too!

Does this change require revising our `style` slots? If it does, then I'd say it's post 3.0, otherwise it's a no brainer.

aartaka:
> Does this change require revising our `style` slots? If it does, then I'd say it's post 3.0, otherwise it's a no brainer.

Yes, it does require at least quoting these rules and changing value injection syntax. But that's not too much work, especially if I figure out the way to discern LASS rules from everything else.

aadcg:
I agree this is very welcome! I just don't see it as a priority right now.

aartaka:
@aadcg, with LASS, the whole problem of #2625 is solved by redefinging `.accent` like this:
``` lisp
(.accent
 :background-color #(theme:accent)
 :color #(theme:on-accent)
 ((:or p h1 h2 h3 h4 h5 h6 span)
  :color #(theme:accent)
  :background #(theme:background)))
```

source/manual.lisp
```diff
@@ -538,25 +538,13 @@ every individual class controlling Nyxt interface elements. All such classes hav
             ((style
               (theme:themed-css (theme *browser*)
                 (*
-                 :background-color (str:concat
-                                    (if (theme:dark-p theme:theme)
-                                        theme:background
-                                        theme:on-background)
-                                    " !important")
```
aadcg, line 545: This seems to be outdated on master right?

aartaka: Why?

aadcg: Why is the `if` clause needed?

aartaka: See the reasoning in 8c3410e8077333dd3a8c11c53a891dd593db0d64 for why I removed it.

aadcg:
I see @aartaka, but let's keep the discussion about that class in the other PR.

source/manual.lisp
aartaka: To change colors things conditionally on the theme being dark or not. 

Actually, I'm lo longer sure on the necessity of distinguishing dark and light themes, given that we test them for contrast now....

aadcg: `background-color` is set when you create a theme object. In other words, `theme:background` is always correct, and there's no need for this `if` clause... 

> Actually, I'm lo longer sure on the necessity of distinguishing dark and light themes, given that we test them for contrast now....

I don't see any connection between testing for contrast ratios and the necessity of distinguishing between dark and light themes. If you're referring to the need of having the `dark-p` slot, I find it useful.  Based on that variable I can define logic in other contexts. Example "turn on web extension X that inverts CSS when `dark-p` is set`.

aadcg: It would be more correct to set `dark-p` after the initialization of the object based on the value of `background-color`, and I have thought about this. But it's not very important and it's not clear how you'd define that when the values aren't white or black. At what point of grey can we qualify it as dark/light...? Not worth the effort IMO.

aartaka: In this particular case, `if` is setting the color of whatever to the darkes available color, which is
- `background-color` for dark themes,
- and `on-background-color` for light ones. 
It's the easiest thing one can do to find an unconditionally dark color in whatever theme. Iterating through theme is OK too, but much more work for no immediate benefit.

Anyway, this piece of code is removed now (but there's a duplicate in `style-mode` :P), and discussing it is beyond this PR :3

aadcg: > Anyway, this piece of code is removed now (but there's a duplicate in style-mode :P), and discussing it is beyond this PR :3

Yes. I mentioned it because it seemed there was something wrong in master. Therefore it would make sense to make that change outside the scope of this PR.

It seems I don't quite follow what that example wants to show but that's a separate topic.

aartaka:
@aadcg, so the PR looks good to you? Then we wait until my LASS patch is accepted to Guix and merge after it?

aadcg:
Just give me a bit more time.

aartaka:
LASS is on Guix!

aartaka:
> Just give me a bit more time to review.

@aadcg, friendly ping :)

aadcg:
The styling of REPL seems off. Could you take a look?

aartaka:
> The styling of REPL seems off. Could you take a look?

Fixed with 1ab0c8dd54a64dbad0fffc21678e28ebc68b0c55 and 235f20ac609d03f2810423b13f40934c42b0d1bf!

source/buffer.lisp
```diff
@@ -65,12 +65,12 @@ various parts, such as the path of all data files.")
              :outline "inherit")
             (.button
              :all "unset"
-             :background-color theme:primary
-             :color theme:on-primary
+             :background-color #(theme:primary)
```
aadcg, line 68: Personally I find this syntax is a bit off-putting. Why does `lass` use it? It gives the impression that we passing an array to `:background-color`. Sorry, I didn't get too deep into the library.

aadcg:
It seems that you find backwards compatibility important. It's not so obvious to me. Why should we do it now that we're making a new release?

aadcg:
You forgot to port to the syntax in `window`, `search-buffer`, `repl`, `hint` (skipped one occurrence)

aartaka:
Wait, where exactly in source/window.lisp and source/mode/search-buffer.lisp did I forget it?

aartaka:
> It seems that you find backwards compatibility important. It's not so obvious to me. Why should we do it now that we're making a new release?

Indeed backwards compatibility is important to me. Why break what works perfectly except for naming? The problem is: CL-CSS doesn't work anywhere close to perfectly :P 
- It requires either read-time evaluation or quirky requoting of forms to insert any runtime value into CSS, 
- it has a spotty support for CSS media queries,
- it's not extensible,
- and it has no facilities to shorten the code in any significant way—it's merely CSS in Lisp, with no benefits Lisp usually provides.

With this set of problems, CL-CSS is not good for moving forward, and it will result in the linear growth of code bloat (which is better than exponential, but worse than code shortening). LASS, in this regard, has facilities to 
- shorten the code,
- extend the syntax,
- and insert any runtime value without painfully processing the whole cons tree.

So yeah, LASS is a matter-of-survival choice for our CSS use here :P

aartaka:
@aadcg, again, have I missed some piece of styling in window.lisp and mode/search-buffer.lisp?

aadcg:
> Wait, where exactly in source/window.lisp and source/mode/search-buffer.lisp did I forget it?

My bad.  It seems those files were not automatically reverted when changing branches.

aadcg:
That `lass` is superior to `cl-css` has always been clear.

But I don't follow what do you mean by:

>  Indeed backwards compatibility is important to me. Why break what works perfectly except for naming?

What exactly are we trying to preserve in terms of backwards compatibility?

aartaka:
> > Indeed backwards compatibility is important to me. Why break what works perfectly except for naming?
> 
> What exactly are we trying to preserve in terms of backwards compatibility?

Function/class/slot names, and the ways to use those (function signatures, class methods). Which is mostly about naming. That's why I say "except for naming"—if something works, then the last thing we should change about it is the name, because that'll immediately break all the dependent code. But, obviously, it's not only about the name. In case of LASS, it's also about syntax of the macro, which it sucks to change, but the benefits overweight the problems.

aadcg:
The way I see it is that we will loose backwards compatibility in the sense that the following will not longer evaluate to the correct CSS string:

```lisp
(theme:themed-css (theme *browser*)
  (.class1
   :background-color theme:secondary
   :color theme:on-secondary)
  (.class2
   :color theme:on-accent))
```

With `lass` we need to surround the theme variables with `#()`.

libraries/theme/theme.lisp
```diff
@@ -150,13 +151,28 @@ headings have border of secondary color.
...
+           `(p
+             :color ,(if (theme:dark-p theme:theme) theme:accent theme:secondary)
+             :background-color #(theme:background)))"
   `(with-theme ,theme
-    (cl-css:css
-     (list ,@(mapcar #'requote-rule rules)))))
+     (funcall #'%themed-css theme:theme
+              ;; NOTE: This loop allows to omit quotes for most trivial rules,
+              ;; mostly preserving backwards-compatibility.
+              ,@(loop for rule in rules
```
aadcg, line 165: So what is being achieved here?

aartaka: Here we preserve a mostly backwards-compatible unquoted syntax for blocks, wherever intuitive. Because we lost the ability to insert any code piece into property values given LASS migration and `requote-rule` removal.

aartaka: And because quotes are ugly :P

aartaka:
Yes, we lose backwards-compatibility, but we lose only the value insertion part of it, which is less of an annoyance than changing the whole thing. I've tried to minimize breakages wherever possible!

aadcg:
I see. Can you give an example of how it would look like without avoiding breaking it?

aartaka:
> Can you give an example of how it would look like without avoiding breaking it?

``` lisp
(theme:themed-css *theme*
  '(h1
    :color #(theme:accent))
  `(body
    :color ,(if (theme:dark-p theme:theme)
                theme:background
                theme:on-background)
    :background-color #(theme:primary)))
```

So both LASS insertion syntax and mandatory quotes.

aadcg:
My opinion is that backwards compatibility, generally speaking, should be either total or none.

In this concrete case, when I mentioned:

> Does this change require revising our style slots? If it does, then I'd say it's post 3.0, otherwise it's a no brainer.

My point was not that we should aim at backwards comparability, but rather that we should focus our efforts on more pressing issues right now. From our experience, no little change in Nyxt goes without a proper discussion (and that's a good thing!).

In that sense, I maybe have confused you and you made an extra effort to seek backwards compatibility.

My opinion, now that I had a proper look at `lass` is that using the quoting is good and should be hidden under the rug!

Just look:

```lisp
(theme:themed-css +light-theme+
  `(.class1
   :background-color ,theme:secondary
   :color ,theme:on-secondary)
  `(.class2
   :background-color ,theme:accent
   :color ,theme:on-accent))
=> 
".class1{
    background-color: #A6A6A6;
    color: black;
}

.class2{
    background-color: #37A8E4;
    color: black;
}"
```

Uraa!

In that sense, see 015c2c5d5.

aadcg:
Just a reminder that we have some `cl-css:css` calls when using `spinneret:with-html-string`, so that also needs to be refactored.

aadcg:
Also, I find that `#()` syntax atrocious and advocate towards using the usual Lisp backquote syntax we know and love.

aartaka:
The suggestion is good overall, I thought about outsourcing things to LASS too. The only thing I'd change is: retain the `:let`-bound variables, so that LASS `#(syntax)` is accessible alongside the quoted one.

But then, I like convenient syntaxes (and you might've seen by `:ncode`) and I'd enjoy having no quotes for simple style blocks and only using them for the non-trivial ones. I mean, it's more lines of code in `theme.lisp`, but less irritation everywhere else :)

aartaka:
@aadcg, I've completed 66a4cfbe351ffb6fdd0de8c56154b463d9792b00 with Guix change :)

aadcg:
> retain the :let-bound variables, so that LASS #(syntax) is accessible alongside the quoted one

It's really horrendous... And as the `lass` README mentions, it makes sense in LASS-only files. I'd insist on this one.

> I'd enjoy having no quotes for simple style blocks and only using them for the non-trivial ones

Ok, I can agree with that.

aartaka:
@aadcg, how about settling on a middle ground, which my "smart" unquoted-plus-quotes version was?

aartaka:
I am a terrible persion, but I actually like LASS syntax. I believe we have to accomodate for maximum aesthetic choices possible here. @aadcg, mind reverting 015c2c5d5736a1d7cde24ba83665457771b82725?

I am not ever against using backtick-comma syntax in our code, though—but I won't be the one to rewrite it :P

aadcg:
You can revert it. I just wanted to show to you how I see it.

aartaka:
015c2c5d5736a1d7cde24ba83665457771b82725 dropped. @aadcg, want to rewrite the rules in a prettier backtick-comma syntax?

aadcg:
Are you saying you're keeping the #() syntax when we can use backquote?

aadcg:
5f3308a47. Probably has mistakes...

aadcg:
Not as many as I thought.

aadcg:
@aartaka I'm sorry but I can't agree with the direction this PR is taking.

Only now have I fully understood how subtle your implementation is. I wondered how these two lexical environments were working together, and it can be summed up by a simple example.

```lisp
(let ((color1 "#000000"))
  (lass:compile-and-write
   `(:let ((color2 "#ffffff"))
      (div :background #(color2))
      (a :background ,color1))))
```

The vector syntax is a hack that `lass` provides to generate css files from lass files. We don't need that since we will be generating css by calling `themed-css`. It makes no sense to support syntax just for its own sake (the syntax is also non-Lispy and unintuitive). The nested lexical environments are exactly the same and thus it just adds up complexity for no good reason.

Regardless, my opinion is just a mere opinion. It's just that "I can't put my signature on this one". Maybe you can convince @Ambrevar @jmercouris otherwise.

aartaka:
@aadcg, OK, let it be backticks then. I've removed the `:let` block in 7d81d4b498d87595d70c2981220e636c184ea768. I've also propagated the newly correct backticks everywhere `theme:themed-css` is used. And I don't like the fact that we have to quote even the most trivial rules. But I can live with it nonetheless.

How about restoring the rule quotes guessing, at least? This way, we can at least omit quotes for trivial rules and use them only for themes or other non-trivial logic.

aadcg:
> How about restoring the rule quotes guessing, at least? This way, we can at least omit quotes for trivial rules and use them only for themes or other non-trivial logic

No strong opinion, unlike the topic I've raised above which is more critical. 

Personally I'd keep all quotes because it actually helps to understand what's going on, i.e. how `lass` handles the blocks. Less added complexity by not adding an implicit quote logic (fallback on the author's library decisions frees us from the burden). Notice that it looks a bit odd when some blocks have backquotes while others have none (what kind of sorcery is that?). 

Another approach that would result in uniform syntax would be to backquote all blocks (adding that logic in `themes-css`), but I also find that highly undesirable. 

Regardless, this is not very important.

aadcg:
If you're interested in taking a look how it looks with all quotes you can apply the patch below. 

[0001-libraries-theme-theme-Replace-CSS-dependency.patch.txt](https://github.com/atlas-engineer/nyxt/files/10001502/0001-libraries-theme-theme-Replace-CSS-dependency.patch.txt)

aartaka:
@aadcg, I've already added backquotes to all the rules in 49c92d4133e34655e6aa97e0163c9552e9350355, that's why I'm concerned about how they look :P  

But I can live with that given the benefits of reduced complexity and higher extensibility LASS provides out of the box.

So, merging then?

aadcg:
Why use `backquote` everywhere?

aadcg:
I believe we need to do some git operation to get rid of the `cl-css` submodule correct?

aartaka:
> Why use `backquote` everywhere?

For consistency ¯\\\_(ツ)_/¯

aartaka:
> I believe we need to do some git operation to get rid of the `cl-css` submodule correct?

Oh right, I'll remove it.

aartaka:
> I believe we need to do some git operation to get rid of the `cl-css` submodule correct?

Done in 14b82365a1c533bfde2c7d8d056e1c6fbbbb3e92. Was a mere `o k cl-css` in Magit.

aadcg:
Thanks @aartaka.

Some (personal) remarks:

- The merged version's indentation is off (I'll push a fix);
- A single commit would have sufficed when merging.  Why pollute our git log with iterations that never landed on master? Notice that the several commit were valuable while we're working on the PR. I'm not suggesting that the PR should have been opened with a single commit! 
- `rules` is a concept perhaps is residual from the old library (`cl-css`). I believe `lass` refers to "rules" as blocks.

aadcg:
See 15243c3df.

aartaka:
> * The merged version's indentation is off (I'll push a fix);

Uh oh :D 

> * A single commit would have sufficed when merging.  Why pollute our git log with iterations that never landed on master? Notice that the several commit were valuable while we're working on the PR. I'm not suggesting that the PR should have been opened with a single commit!

I mean, yeah, but I believed it better to split it in at least several, given that amount of discussion we had over each one. Note that I squashed a lot already when merging :P

> * `rules` is a residual concept perhaps from the old library (`cl-css`). I believe `lass` refers to "rules" as blocks.

WIll fix!

aartaka:
ebfd5e7f07d68bcd18d0274dec84504ef8fe330b fixes it!

Issue #2583, opened 2022-09-17
Buffer IDs do not persist


franburstall:
**Describe the bug**

Once more I do not know if this is a bug or a feature request.

I would like to persist a list of favoured buffers and had thought that I could simply save a list of buffer id's to disk.

However, I discover that the mapping from id's to buffers does not survive a restart.   That is, the buffers in a restored session have different ids after a restart..

Should this be the case?  I thought ids used to persist.

If not, is there a way to accomplish my goal?

**Information**
- OS name+version: debian 11
- Desktop environment / Window manager name+version: exwm
- How you installed Nyxt (Guix pack, package manager, build from source): source
- Information from `show-system-information`:
```
Nyxt version: 3-pre-release-1-478-g870c8d28
Renderer version: GI-GTK
Operating system kernel: Linux 5.10.0-18-amd64
Lisp implementation: SBCL 2.1.1 (Dynamic space size: 3221225472)
Features: (:QUICKLISP :NYXT-GI-GTK :NYXT-GTK :WEBKIT2 :WEBKIT2-2.36
 :WEBKIT2-CORS-ALLOWLIST :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
 :GLIB-2-30 :GLIB-2-32 :GLIB-2-34 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42
 :GLIB-2-44 :GLIB-2-46 :GLIB-2-48 :GLIB-2-50 :GLIB-2-52 :GLIB-2-54 :GLIB-2-56
 :GLIB-2-58 :GLIB :NYXT-3.1 :NYXT-3 :CUSTOM-HASH-TABLE-NATIVE :SLYNK
[...]

PR #2627, opened 2022-11-07
WIP: Persist buffer state to history (for modes, tags and more)

### LINKED ISSUE #2036: Webkit settings do not get persisted in buffers
**Describe the bug**
Often times, I like to navigate the web without media or images enabled by creating a mode similar to `no-image-mode`.

```lisp
(define-mode disable-media-mode ()
  "Disables media in the browser."
  ((constructor (lambda (mode)
                  (ffi-buffer-auto-load-image (buffer mode) nil)
                  (ffi-buffer-enable-media (buffer mode) nil)))   
   (visible-in-status-p nil)))
```

This works great for the most part. But I still get media when I access a media URL directly, such as opening an [image](https://nyxt.atlas.engineer/image/nyxt_128x128.png).

Further, when I delete the buffer and I restore it via `reopen-buffer`, the Webkit settings that were applied before are not anymore. I remember having seen an issue regarding buffer persistence somewhere in the project's issue tracker, but I can't recall which one it is and searching for it yielded no relevant results.

Open a buffer with the aforementioned `disable-media-mode`, kill it, and re-open it via `reopen-buffer`, yielding the default Webkit settings and not the ones I specified.

For website-specific issues:
Can you reproduce this issue with Epiphany / GNOME Web (https://wiki.gnome.org/Apps/Web)?
No

**Information**
- OS name+version: guix (GNU Guix) ea11be50e9dbc5c043d3a18be8364b2e1b2507a5
- Graphics card and driver: Intel Corporation UHD Graphics
- Desktop environment / Window manager name+version: EXWM
- How you installed Nyxt (Guix pack, package manager, build from source): Guix pack

```
Nyxt version: 2.1.1-1293-g45c399e4
Renderer version: GI-GTK
Operating system kernel: Linux 5.15.7
Lisp implementation: SBCL 2.1.9 (Dynamic space size: 3221225472)
Features: (:SERIES :SERIES-ANSI :SLYNK :WEBKIT2 :WEBKIT2-2.34 :WEBKIT2-CORS-ALLOWLIST
 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING :WEBKIT2-MUTE :WEBKIT2-EMOJI
 :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22 :GTK-3-20 :GTK-3-18 :GTK-3-16
 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6 :GTK-3-4 :GTK :GDK-3-22
 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12 :GDK-3-10 :GDK-3-8 :GDK-3-6
 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF :GLIB-2-30 :GLIB-2-32 :GLIB-2-34
 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42 :GLIB-2-44 :GLIB-2-46 :GLIB-2-48
 :GLIB-2-50 :GLIB-2-52 :GLIB-2-54 :GLIB-2-56 :GLIB-2-58 :GLIB :NYXT-2
 :FSET-EXT-STRINGS :CUSTOM-HASH-TABLE-NATIVE :SWANK :PLUMP-UTF-32 :GLOBAL-VARS
 :DECLARE-TYPES :PARENSCRIPT :SBCL+SAFE-STANDARD-READTABLE :NAMED-READTABLES
 :LPARALLEL :21BIT-CHARS :CHUNGA :FLEXI-STREAMS :CLOSER-MOP :CL-PPCRE-UNICODE
 :CL-UNICODE :CL-PPCRE :CL-JSON-DOUBLE-FLOAT-IS-SUBSUMED
 :CL-JSON-SINGLE-FLOAT-IS-SUBSUMED :BORDEAUX-THREADS :LPARALLEL.WITH-CLTL2
 :LPARALLEL.WITH-CAS :LPARALLEL.WITH-STEALING-SCHEDULER :SPLIT-SEQUENCE
 CHIPZ-SYSTEM:GRAY-STREAMS CFFI-FEATURES:FLAT-NAMESPACE CFFI-FEATURES:X86-64
 CFFI-FEATURES:UNIX :CFFI CFFI-SYS::FLAT-NAMESPACE ALEX::SEQUENCE-EMPTYP
 :FAST-IO-SV :FAST-IO :SBCL-USES-SB-ROTATE-BYTE :CL-JSON-CLOS :CL-JSON
 :THREAD-SUPPORT :ASDF3.3 :ASDF3.2 :ASDF3.1 :ASDF3 :ASDF2 :ASDF :OS-UNIX
 :NON-BASE-CHARS-EXIST-P :ASDF-UNICODE :X86-64 :GENCGC :64-BIT :ANSI-CL
 :COMMON-LISP :ELF :IEEE-FLOATING-POINT :LINUX :LITTLE-ENDIAN

ASDF version: 3.3.5
Critical dependencies: (/gnu/store/s7w144g61qsmschm6mzy41bvq0qvjfrs-cl-cffi-gtk-0.11.2-2.e9a46df/share/common-lisp/source/cl-cffi-gtk/gtk/cl-cffi-gtk.asd
 /gnu/store/w5awybp5s66bn7g74mbvlr10ivhr6m2v-cl-gobject-introspection-0.3-1.d0136c8/share/common-lisp/source/cl-gobject-introspection/cl-gobject-introspection.asd
 /gnu/store/gcqfm1frfrcdiizn6swqsy2znvyvmdqc-cl-webkit-3.5.0/share/common-lisp/source/cl-webkit/webkit2/cl-webkit2.asd)

Guix version: 
```

  If you can't run `show-system-information`, try `nyxt --system-information` from
  a shell.  It this still does not work, please provide the following:
  - Nyxt version (from the `nyxt-version` command or `nyxt --version` from a shell):
  - Lisp implementation/version (if built from source):
  - Kernel name+version:
  - WebKitGTK+ | QtWebEngine version:

### LINKED ISSUE #2064: Modes are not restored when restoring session
It seems that modes are not restored when restoring sessions.
In particular, when help buffers are restored, they don't have the `help-mode`.
Which is probably why we experience https://github.com/atlas-engineer/nyxt/issues/2057.

In `restore-history-buffers`, there is no mention of modes.
And indeed, they are not stored in the global history.  This is probably related to #1095, #2036.

But here is the heart of the problem: where do we store modes?
We can store arbitrary data in `history-entry`, but these are meant for individual URLs, not for history owners (which are the buffers).  Owners don't have such arbitrary data.  Shall we add it to the `htree` library then?

Thoughts?

Ambrevar:
# Description

This persists buffer modes (and more buffer data) in history.  This is useful so that on restart, modes are restored (which can be critical e.g. for privacy).

Fixes #2064 and #2583 (not yet).
Fixes #2036.

# How to test

- Make sure your cl-prevalence is up-to-date.

# Discussion

- [x] Is the design sane?
  In my opinion, yes, as it makes only minimal modifications to the Htree library, while adding a Nyxt core class that's easily extensible to add more buffer-related data to persist in the future.

- [x] Can we better serialize and deserialize URIs and timestamp?  The problem is that until now we used to serialize them as strings, but then how can Nyxt know which string is a URI, a timestamp, or just a string?
  An option would be to use reader macros, for instance `(local-time:enable-read-macros)`.  But that would break history backward compatibility.  Thoughts?

  - Another suggestion by @aartaka: Introduce `uri-sepcifier` and `timestamp-specifier` which accept strings, then write accessors to automatically replace the string with the structured version.  With a bit of MOP we could derive a general solution.
  - Better option: `coerce-slot`, which automatically coerce the value to the desired type.

# To do:

- [ ] Replace current dead buffers with serialized buffers.  Benefit: less risk for memory leaks.
- [x] Squash commits since this PR commit history is a mess.
- [x] Update cl-prevalence in submodules and Guix once patch is merged upstream. 
- [ ] Tests!
- [ ] Add a `tags` slot and persist it.
- [ ] Add tag-related commands.  See https://github.com/atlas-engineer/nyxt/issues/2583#issuecomment-1298379178
- [ ] Fix auto-rules which prevents any good use of this PR:
  - [ ] Currently auto-rules remove modes from the list of disabling them.
  - [ ] Reloading a page removes the manually enabled modes.  Example:
    - Go to https://en.wikipedia.org
    - Enable no-script-mode.
    - Reload: BOOM!  no-script-mode is gone.

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -348,11 +348,12 @@ Return non-NIL of history was restored, NIL otherwise."
                        (gethash (htree:creator-id owner) old-id->new-id)))
                (htree:owners history))
       (setf (htree:owners history) new-owners))
-    (alex:when-let ((latest-id (first
-                                (first
-                                 (sort-by-time (alex:hash-table-alist (htree:owners history))
-                                               :key (compose #'htree:last-access #'rest))))))
-      (switch-buffer :buffer (buffers-get latest-id)))))
+    (alex:when-let* ((latest-id (first
+                                 (first
+                                  (sort-by-time (alex:hash-table-alist (htree:owners history))
+                                                :key (compose #'htree:last-access #'rest)))))
+                     (buffer (buffers-get latest-id)))
+      (switch-buffer :buffer buffer))))
 
 (defmethod files:deserialize ((profile nyxt-profile) (file history-file) raw-content &key)
   "Restore the global/buffer-local history and session from the PATH."
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -199,14 +199,11 @@ lot."
         (*print-length* nil))
     ;; We need to make sure current package is :nyxt so that symbols are printed
     ;; with consistent namespaces.
-    (write
-     (with-input-from-string (in (with-output-to-string (out)
-                                   (s-serialization:serialize-sexp
-                                    (list +version+ (files:content file))
-                                    out)))
-       ;; We READ the output of serialize-sexp to make it more
-       ;; human-readable.
-       (safe-read in))
+    (pretty-print
+     (with-output-to-string (out)
+       (s-serialization:serialize-sexp
+        (list +version+ (files:content file))
+        out))
      :stream stream)))
 
 ;; REVIEW: This works around the issue of cl-prevalence to deserialize structs
--- a/source/history.lisp
+++ b/source/history.lisp
@@ -334,10 +334,14 @@ Return non-NIL of history was restored, NIL otherwise."
                                     (htree:owner history owner-id))))
                  ;; Node-less owners can safely be ignored.
                  (when current-node
-                   (let ((new-buffer (make-buffer :title (title (htree:data current-node))
-                                                  :history-file history-file
-                                                  :url (url (htree:data current-node))
-                                                  :load-url-p nil)))
+                   (let ((new-buffer (alex:if-let ((data (htree:data owner)))
+                                       (with-input-from-string (in data)
+                                         (s-serialization:deserialize-sexp in))
+                                       ;; In case buffer was not serialized:
+                                       (make-buffer :title (title (htree:data current-node))
+                                                    :history-file history-file
+                                                    :url (url (htree:data current-node))
+                                                    :load-url-p nil))))
                      (setf (gethash owner-id old-id->new-id) (id new-buffer))
                      (setf (gethash (id new-buffer) new-owners) owner))))))
             (alex:hash-table-alist (htree:owners history)))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -267,6 +267,10 @@ This method should be called by the renderer after instantiating the web view
 of BUFFER."
   (unless no-hook-p
     (hooks:run-hook (buffer-make-hook browser) buffer))
+  ;; Modes may be initialized without a `buffer' slot (e.g. when deserialized).
+  (mapc (lambda (mode)
+          (setf (buffer mode) buffer))
+        (slot-value buffer 'modes))
   (mapc #'enable (slot-value buffer 'modes))
   (enable-modes* (append (reverse (default-modes buffer))
                          (uiop:ensure-list extra-modes))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -38,7 +38,8 @@ inherited from the superclasses.")
     :documentation "Buffer profiles are used to specialize the behavior of
 various parts, such as the path of all data files.
 See also the `profile' slot in the `browser' class.")
-   (url (quri:uri ""))
+   (url (quri:uri "")
+        :type quri:uri)
    (url-at-point (quri:uri ""))
    (title "")
 
@@ -284,6 +285,16 @@ buffer, with a meaningful result."
 To access all modes, including disabled ones, use `slot-value'."
   (sera:filter #'enabled-p (slot-value buffer 'modes)))
 
+(defmethod (setf modes) (value (buffer modable-buffer))
+  "Persist buffer state into history when changing mode."
+  (setf (slot-value buffer 'modes) value)
+  (when (context-buffer-p buffer)
+    (files:with-file-content (history (history-file buffer)
+                              :default (make-history-tree))
+      (alex:when-let ((owner (gethash (id buffer) (htree:owners history))))
+        (setf (htree:data (gethash (id buffer) (htree:owners history)))
+              (serialize-buffer buffer))))))
+
 (define-class input-buffer (buffer)
   ((keyscheme
     keyscheme:cua
@@ -412,6 +423,7 @@ down."))
 (define-class context-buffer (buffer)
   ((last-access
     (time:now)
+    :type time:timestamp
     :export nil
     :documentation "Timestamp when the buffer was last switched to.")
    (search-engines
@@ -872,6 +884,10 @@ Return the created buffer."
   ;; Background buffers are invisible to the browser.
   buffer)
 
+(defun serialize-buffer (buffer)
+  (with-output-to-string (out)
+    (s-serialization:serialize-sexp buffer out)))
+
 (defmethod customize-instance :after ((buffer context-buffer)
                                       &key parent-buffer no-history-p
                                       &allow-other-keys)
@@ -893,7 +909,8 @@ Return the created buffer."
                                                 (global-history-p buffer)
                                                 (not (nosave-buffer-p buffer))
                                                 (not (nosave-buffer-p parent-buffer)))
-                                       (id parent-buffer))))))
+                                       (id parent-buffer))
+                         :data (serialize-buffer buffer)))))
   buffer)
 
 (define-command update-document-model (&key (buffer (current-buffer)))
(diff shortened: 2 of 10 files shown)


aadcg:
@Ambrevar Could you please re-word the discussion section? It seems off.

aadcg:
Could you please explain the concept of a `tag` in this context?

source/history.lisp
```diff
@@ -45,6 +45,14 @@ It's a list of a form (Y &OPTIONAL X)."))
...
 The total number of visit for a given URL is (+ explicit-visits implicit-visits)."))
 
+(define-class buffer-history-entry ()   ; TODO: Export?
+  ((modes
+    '()
+    :type (list-of symbol)
+    :documentation "List of modes that were last active for buffer."))
+  (:accessor-name-transformer (class*:make-name-transformer name))
+  (:documentation "Buffer state to restore when restoring history."))
+
```
aadcg, line 55: `history.lisp` is loaded after `buffer.lisp`, so it's possible to access the `modes` slot from `modable-buffer`. Would it be a bad idea to dump the enabled modes to the history file before exiting Nyxt?

Wouldn't this functionality be achieved in a cleaner way via the new "`auto-mode`"?

Sorry if my understanding is completely off as I really don't know how we manage history!

source/history.lisp
```diff
@@ -344,7 +352,8 @@ Return non-NIL of history was restored, NIL otherwise."
                                                   :url (url (htree:data current-node))
                                                   :load-url-p nil)))
                      (setf (gethash owner-id old-id->new-id) (id new-buffer))
-                     (setf (gethash (id new-buffer) new-owners) owner))))))
+                     (setf (gethash (id new-buffer) new-owners) owner)
+                     (enable-modes (modes (htree:data owner)) new-buffer))))))
```
aadcg, line 356: I'm a bit confused by this. Could you explain?

Ambrevar:
> @Ambrevar Could you please re-word the discussion section? It seems off.

Done.

> Could you please explain the concept of a `tag` in this context?

A buffer tag is an arbitrary string (or keyword?).  Users can then filter / categorize buffers.
See #2583 for a discussion.

source/history.lisp
Ambrevar: History stores the list of buffer IDs and their histories.
When we restore the history, we recreate the corresponding buffers and associate them with these histories.

auto-rules associate URLs with modes.  Here we want to restore modes for specific buffers.

aadcg: Clear.

source/history.lisp
aadcg: To be more precise: `htree:data` can be whatever, so it feels a bit arbitrary to use it to restore the modes that were active in a buffer.

Ambrevar:
Thought more about it: manually serializing modes and tags might be a bit shortsighted.

Instead, we should probably add a `s-serialization` method for buffers, which could work independently from history altogether.

For instance, create buffer B, serialize it with the new method, kill the buffer, then deserialize the serialized content to recreate the buffer.

aadcg:
That sounds better to me @Ambrevar!

source/history.lisp
aartaka: Actually, we can try using history for all this `mast-active-modes` business of `auto-mode`/auto-rules (ʃ⌣́,⌣́ƪ)

source/history.lisp
aartaka: `htree:data` is an accessor for the arbitrarily-typed payload inside the owner/node. This is a design decision that we made to not tie ourselves to any particular design of histoy nodes and to make `htree` more reusable.

Ambrevar:
By the way, this would address the long-standing issue of how to abstract "dead buffers".  Two (or three?) birds with one stone.

aartaka:
Yes, serializing buffers to history was the initial solution I thought of for buffer tagging :D

Ambrevar:
@aartaka It seems that auto-rules are broken.  For instance reloading a page disables the manually enabled modes, which is wrong.  (See the original post for an example).

Can you have a look?

Ambrevar:
Good news!  I'm able to serialize and restore buffers with all their modes!

aartaka:
> @aartaka It seems that auto-rules are broken. For instance reloading a page disables the manually enabled modes, which is wrong. (See the original post for an example).

Ahhhhhh, I might know the reason—I've removed the handler from `(en|dis)able-mode-hook`, and we lost the ability to preserve the most recent set of modes. I know how to fix it and will fix it tomorrow!

aartaka:
@Ambrevar, toggled modes preservation fixed in 16483000fbcc279ccfee050c71400bd52dddc692!

Ambrevar:
Thanks!

Ambrevar:
Before going further we need to merge #2654 and https://github.com/40ants/cl-prevalence/pull/22.

Ambrevar:
I've rebased and squashed. 

Prevalance still needs a fix: https://github.com/40ants/cl-prevalence/pull/25

Then there are a few more issues to iron out with this patch (like internal page URLs failing for some reason), but we are close!

aartaka:
Internal page URLs may fail because they are being loaded before the context has all the schemes initialized. But that's a wild guess :)

Ambrevar:
I'm working on the problem that URLs / timestamps are serialized as strings and thus need some manual code to be deserialized properly.  Of course the manual code is easy to write, the problem is that it needs to be set in many places which is cumbersome and error prone.

The following seems to work to ensure that `buffer`'s `url` slot is always a quri:uri:

```lisp
(defmethod (setf slot-value-using-class) :around (new-value
                                                  (class user-class)
                                                  (instance buffer)
                                                  slot)
  (if (eq 'url (c2mop:slot-definition-name slot))
      (call-next-method (ensure-url new-value) class instance slot)
      (call-next-method new-value class instance slot)))
```

Still, it needs to be done for all classes + it's verbose.  Can we do better?

I'm thinking of a few options:

1. Add a `:type-coercion` slot option that does the above by calling the value (a function) on the slot value and assigning the slot to the result.  Example

```lisp
(define-class ...
  ((foo "blah"
    :type string
    :type-coercion #'princ-to-string)
    ...))
```

Way less verbose, but still needs to be done for all slots.

2. Add a class option that does just the same for some types.
3. Add a global option / method that's called by user classes.  Then maybe this would do (untested):

```lisp
(defmethod (setf slot-value-using-class) :around (new-value
                                                  (class user-class)
                                                  instance
                                                  slot)
  (typecase (c2mop:slot-definition-type slot)
    ((quri:uri)
     (call-next-method (ensure-url new-value) class instance slot))
    (t
     (call-next-method new-value class instance slot))))
```

The drawback is that this is global, so if we want a bit more flexibility for a slot, we would need to override that method for a given class.  Or... just use a more flexible type.  So maybe that's not much of a drawback after all.

Thoughts?

aartaka:
> I'm working on the problem that URLs / timestamps are serialized as strings

But why not override cl-prevalence serialization for these?

Ambrevar:
> But why not override cl-prevalence serialization for these?

I guess you meant DEserialization.

We cannot do it based on the type, because the type is `string`, and we obviously cannot deserialize all strings to `quri:uri` or `local-time:timestamp`.

So we would have to do it for every quri:uri / timestamp slot of every class.  Which is what this pull request does.  It's cumbersome and error prone since it's easy to forget some slots.  Plus it's an extra burden for extension writers / user configuration.

Ambrevar:
Before this  can be tested properly #2895 must be addressed.

jmercouris:
Please reopen when ready.

PR #2629, opened 2022-11-08
Fix startup page

aadcg:
# Description

Fix UI styling.

Fixes [this issue](https://github.com/atlas-engineer/nyxt/commit/f6307328626601e7358dee0d2c04c116b2e044fa#commitcomment-88844270).

# Discussion

I took a different approach than that suggested by @aartaka in the aforementioned issue. I find it simpler and more effective.

@aartaka can you confirm that the UI on the REPL's side looks as it should? Thanks.

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -75,6 +75,9 @@ various parts, such as the path of all data files.")
              :border-width "0px"
              :padding "6px"
              :margin "2px")
+            (.button.accent
+             :background-color theme:accent
+             :color theme:on-accent)
             (.link
              :all "unset"
              :text-decoration "underline"
@@ -83,8 +86,7 @@ various parts, such as the path of all data files.")
             (".link:hover"
              :opacity 0.8)
             (.accent
-             :background-color theme:accent
-             :color theme:on-accent)
+             :color theme:accent)
             (|.button:hover|
              :opacity 0.8)
             (|.button:visited|


aartaka:
I suggest yet a different of an approach: define .accept per tag type! E.g.:
```  lisp
.accent
:color theme:on-accent
```
``` lisp
"p.accent, h1.accent, h2.accent ..."
:background-color theme:background
:color theme:accent
```

It preserves the conventional/convenient `.accent` class, while fixing accents for text. And, actually, all accented text elements should better be goreground-accented by default. Whic means: no need to add more classes, just adjust the behavior of existing ones!

aadcg:
You have confused me now! How does that differ for your former suggestion?

aadcg:
Also, say you set the accent class to the heading present in our startup page. Then it will match both the accent class and the accent class relative to `h1`. In short, you'll end up with both the foreground and background covered by the `accent-color`.

aadcg:
To make it clear, apply the following onto master and check the startup page.

```diff
diff --git a/source/buffer.lisp b/source/buffer.lisp
index 870abc89c..c88b007f7 100644
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -85,6 +85,8 @@ various parts, such as the path of all data files.")
             (.accent
              :background-color theme:accent
              :color theme:on-accent)
+            (h1.accent
+             :color theme:accent)
             (|.button:hover|
              :opacity 0.8)
             (|.button:visited|
```

aartaka:
> You have confused me now! How does that differ from your former suggestion?

It doesn't :P Because using the selectors mechanism that CSS provides is more effective than simulating it manually, especially for such DWIM-ish universal styles as `.accent`.

aartaka:
> To make it clear, apply the following onto master and check the startup page.
> 
> ```diff
> diff --git a/source/buffer.lisp b/source/buffer.lisp
> index 870abc89c..c88b007f7 100644
> --- a/source/buffer.lisp
> +++ b/source/buffer.lisp
> @@ -85,6 +85,8 @@ various parts, such as the path of all data files.")
>              (.accent
>               :background-color theme:accent
>               :color theme:on-accent)
> +            (h1.accent
> +             :color theme:accent)
>              (|.button:hover|
>               :opacity 0.8)
>              (|.button:visited|
> ```

Yes, the background will be accented, because you only redefined `:color` for `h1`. You didn't specifically redefine `:background-color`, as I _explicitly_ suggested in my commentary ｜−・;）

aartaka:
> I disagree with the per element approach. We're dealing with a CSS class that is literally used in 2 spots (REPL and startup page)! 

This can (and will) change. The fact it's used in just two places doesn't mean it won't be used anywhere else. And the fact it's already incomplete hints that it will inevitably be incomplete once we add it anywhere else.

> Why should we generalize something for which we lack experience? To be honest, I even find the use of the accent CSS class backwards. If you take a closer look, we usually define CSS classes for concrete purposes and then we make use of the `accent` color.

Yes, and this pattern is so frequent that it may deserve a CSS class. Compare:
- You define a `.button-accent :color theme:accent` and then use it as `:button.button-accent`, or
- You re-use the existing class with mere `:button.accent`.

> On the other hand, assuming we'd take your approach, I really don't like it because it hardcodes the background, as you have explicitly suggested.

Yes, it does. And that's for good: text elements, even accented ones, should only appear on a text background.

Alternative to the form I suggested it in, you can make `.accent` to be foreground-accented by default, and define a background accented version for e.g. `.button.accent`. Use your judgment of which of the two will make you write less code.

aadcg:
>  Yes, it does. And that's for good: text elements, even accented ones, should only appear on a text background.

Why making such an opinionated decision up front? It sounds reasonable to imagine you'd have a background colored in either `primary` or `secondary` where text elements would be in `accent`.

aartaka:
> > Yes, it does. And that's for good: text elements, even accented ones, should only appear on a text background.
> 
> Why making such an opinionated decision up front? It sounds reasonable to imagine you'd have a background colored in either `primary` or `secondary` where text elements would be in `accent`.

What about the alternative then?

aadcg:
Your suggestion, if I understood correctly, only differs from the one I started this PR with in the sense that I called the classes `accent-foreground` and `accent-background` and you advocate for `accent` and `accent-background`. See the latest commits.

aartaka:
> Your suggestion, if I understood correctly, only differs from the one I started this PR with in the sense that I called the classes `accent-foreground` and `accent-background` and you advocate for `accent` and `accent-background`. See the latest commits.

I'm not advocating for any additional classes, I advocate for dispatching an intuitive `.accent` CSS class over HTML tag names—an ability that CSS gives us from its first versions (҂⌣̀_⌣́) See latest comments in #2626 for an example.

aadcg:
I thought we have agreed that that suggestion is short-sighted since:

> Why making such an opinionated decision up front? It sounds reasonable to imagine you'd have a background colored in either primary or secondary where text elements would be in accent.

aadcg:
I was referring to this when I mentioned i was unsure if I have followed you correctly:

> Alternative to the form I suggested it in, you can make .accent to be foreground-accented by default, and define a background accented version for e.g. .button.accent. Use your judgment of which of the two will make you write less code.

aartaka:
OK, so I was referring to having
``` css
.accent { ... }
p.accent, h1.accent, h2.accent, ... { ... }
```
not 
``` css
.accent { ... }
.accent-background /* or any other name for this second distinct class*/ { ... } 
```

aadcg:
That is clear to me. 

What I don't follow is how you can adopt that approach and avoid hardcoding the background color to `theme:background` since:

> It sounds reasonable to imagine you'd have a background colored in either primary or secondary where text elements would be in accent.

aadcg:
Actually, I may have understood after all.

Check the overall diff of this PR. I agree with this. If you agree too, I'll squash and merge.

aartaka:
> > Alternative to the form I suggested it in, you can make .accent to be foreground-accented by default, and define a background accented version for e.g. .button.accent. Use your judgment of which of the two will make you write less code.
> 
> I think I didn't understand what you meant here.
> 
> It seems quite different from:
> 
> ```css
> .accent { ... }
> p.accent, h1.accent, h2.accent, ... { ... }
> ```

Yes, because it's an alternative to it. To sum up, I suggest TWO options that would work perfectly without introducing new classes and breaking our CSS API:
1. Define per-text-tag `.accent` color the foreground in accent, with non-accent background:
``` css
.accent { ... }
p.accent, h1.accent, h2.accent, ... { ... }
```
2. Or redefine the default `.accent` to be foreground-only (which has the benefit of preserving the background color of the element `.accent` is applied to), while defining the background-accented version for classes and tags that need the background version:
```
.accent {color: accent;}
.button.accent, div.accent ... {background-color: accent; color: on-accent}
```

I'm leaning to the first option, because I feel that accented background is more frequent of a use-case than the foreground one is.

source/buffer.lisp
```diff
@@ -75,6 +75,9 @@ various parts, such as the path of all data files.")
              :border-width "0px"
              :padding "6px"
              :margin "2px")
+            (button.accent
```
aartaka, line 78: I'd rather do `.button.accent`, because it's not only `<button>`s that we style—it's also `<select>`s and `<details>`.

aadcg: Ah, I see!

aadcg:
Woohoo.

Keep in mind, as I am keeping it as well, your suggestion for the future. Nothing is set in stone.

Issue #2329, opened 2022-05-31
Generate image from commit 


aadcg:
I'd like to build Nyxt on commit cac685372 (from March 2021).

Generating the image, via the mechanism specified in `guix.scm` and `nyxt-guix.el`, should work but I get the error below.

I don't understand why is complains about `cl-markup`. We don't use it anymore, but since it's specified in `guix.scm`, it should work. 

Here's how I set `sly-lisp-implementations`:

```lisp
(sly-lisp-implementations '((sbcl-nyxt (lambda () (nyxt-make-guix-sbcl-for-nyxt
                                                "~/common-lisp/nyxt"
                                                :force t
                                                :preserve '("GDK_SCALE"
                                                            ;; "GDK_DPI_SCALE"
                                                            "GTK_IM_MODULE"
                                                            "QT_IM_MODULE"
                                                            "XMODIFIERS"
                                                            "CLUTTER_IM_MODULE")
                                                :no-grafts t
                                                :ad-hoc '("emacs"
                                                          "git"
                                                          "gnupg"
                                                          "guix"
                                                          "password-store"
                                                          "xdg-utils"))))))
```

CC @Ambrevar 

```
This is SBCL 2.2.2, an implementation of ANSI Common Lisp.
More information about SBCL is available at <http://www.sbcl.org/>.

SBCL is free software, provided as is, with absolutely no warranty.
It is mostly in the public domain; some portions are provided under
BSD-style licenses.  See the CREDITS and COPYING files in the
distribution for more information.
WARNING: redefining PERFORM (#<STANDARD-CLASS ASDF/BUNDLE:IMAGE-OP> #<STANDARD-CLASS ASDF/SYSTEM:SYSTEM>) in DEFMETHOD
WARNING: System definition file #P"/gnu/store/bx3y7726s7vqz9bhsqw51izkkpjpnbvz-cl-calispel-0.1-1.e9f2f9c/share/common-lisp/source/cl-calispel/calispel.asd" contains definition for system "calispel-test". Please only define "calispel" and secondary systems with a name starting with "calispel/" (e.g. "calispel/test") in that file.
[...]

PR #2633, opened 2022-11-10
Fix repeat mode

### LINKED ISSUE #2610: repeat-mode is broken
See https://github.com/atlas-engineer/nyxt/pull/2490 for a discussion.  Currently it seems to be failing silently.

aadcg:
# Description

@aartaka I've cherry picked these commits of yours from #2490 since the scope if not related to the mentioned PR.

Does that make sense?

Fixes #2610.

# Discussion

Work in progress.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:


source/mode/repeat.lisp
```diff
@@ -50,15 +56,19 @@ It takes a `repeat-mode' instance as argument.")))
...
-    (when seconds
-      (enable-modes 'repeat-mode (current-buffer)
-                    (list :repeat-interval seconds :repeat-action function)))))
+  (alex:when-let ((seconds (or seconds
+                               (ignore-errors
+                                (parse-integer
+                                 (prompt1 :prompt "Repeat every X seconds"
+                                          :input "5"
+                                          :sources 'prompter:raw-source))))))
+    (disable-modes :modes 'repeat-mode :buffers (current-buffer))
```
aartaka, line 65: Wait, but this syntax is from the auto-rules branch, isn't it?

aartaka:
Noooo, it does not fix `repeat-mode` and should be left inside the auto-rules branch, because it's a part of the refactor there :(

aadcg:
The part relative to the refactor I've kept in #2490. [See the relevant commit](https://github.com/atlas-engineer/nyxt/pull/2490/commits/a74a657244281ef4afa251cdb0cd682fe258df7a#diff-6a5544c9864339bba3a8d5c3721ed5d125f8f303c1b53cfa641237e6cc0e6afc).

source/mode/repeat.lisp
aadcg: Well, it will be merged when the mode is fixed!

aadcg:
Rebased over master. Would be interested in taking a look @aartaka?

source/mode/repeat.lisp
```diff
@@ -50,16 +56,19 @@ It takes a `repeat-mode' instance as argument.")))
...
-                     :repeat-interval seconds
-                     :repeat-action function))))
+  (alex:when-let ((seconds (or seconds
+                               (ignore-errors
+                                (parse-integer
+                                 (prompt1 :prompt "Repeat every X seconds"
+                                          :input "5"
+                                          :sources 'prompter:raw-source))))))
+    (disable-modes :modes 'repeat-mode :buffers (current-buffer))
+    (apply #'enable-modes
```
aartaka, line 66: Why is it still `enable-modes` and not `enable-modes*`?

aartaka:
Does it work? Does it fix `repeat-mode`?

source/mode/repeat.lisp
aadcg: Right!

aadcg:
The bug stems from the fact that the mode object isn't always garbage collected and so it may be re-initialized with some state left from older instances. 

The key point is how we handle `thread-terminated-p`.

There are several ways to fix this:

- Add a cleanup to `repeat-mode` that resets `thread-terminated-p` and let the `cleanup` be the last thing executed in the `disable` from `process-mode` (what I did)
- Set this `cleanup` to be the default one in `process-mode`, which in other words is equivalent to rewrite `disable` in `process-mode` as:

```lisp
(defmethod disable ((mode process-mode) &key)
  (unless (thread-terminated-p mode)
    (destroy-thread* (thread mode)))
  (setf (thread-terminated-p mode) t)
  (setf (thread-terminated-p mode) nil) 
  (and (cleanup mode)
       (funcall (cleanup mode) (path-url mode) mode)))
```

Another topic: I'm not sure what are the consequences of having the `cleanup` run BEFORE destroying the thread.

@aartaka I'll follow your judgement on this one, as you have a better picture.

aadcg:
@aartaka there are still some commits of yours related to `repeat-mode` while you worked on #2490. I'm keeping those as I'll soon open a new PR that reviews docstrings and minor things related to this mode.

aartaka:
Ahhh, my comment didn't get published. Here we go again:

How about removing `thread-terminated-p` altogether? This way we don't have to inspect anything but `thread`, and thread termination errors won't happen in our `destroy-thread*` anyway.

aadcg:
@aartaka Sure, makes sense to me. Let me explore it.

aadcg:
@aartaka are you able to to use `repeat-key` from this git checkout?  I'm not able to do it, but I wonder if my window manager is playing tricks on me...

aadcg:
It seems that the bug is actually deeper and somewhat unrelated to this.

I believe that there's an issue with the implementation of `enables-modes*`. Running 

```lisp
(enable-modes* 'nyxt/watch-mode:watch-mode (current-buffer) :repeat-interval 10)
```

Results in an object where the argument passed `repeat-internal` is not honored. 

How hard did you test `enables-modes*` @aartaka? :p

This function was introduced in #2490.

aadcg:
@aartaka it's unrelated to #2490, sorry about that.

The following returns nil before that PR. I'm trying to hunt for a time when this actually returns t...

```lisp
(let ((*browser* (make-instance 'browser))
      (buffer (make-instance 'modable-buffer))
      (arg-value 10))
  (with-current-buffer buffer
    (setf (url (current-buffer)) (quri:uri "test"))
    (= arg-value
       (nyxt/repeat-mode:repeat-interval
        (first (modes (enable-modes 'nyxt/watch-mode:watch-mode
                                    buffer
                                    (list :repeat-interval arg-value))))))))
```

aadcg:
It's quite hard to do that due to [an issue I've identified a long time ago](https://github.com/atlas-engineer/nyxt/issues/2329#issuecomment-1144965682).

@aartaka This PR solves what is strictly related to `repeat-mode`.

I'll start a draft PR related to `enable-modes*` and then hopefully `repeat-mode` will be fully working.

Makes sense? Ok to merge?

source/mode/process.lisp
```diff
@@ -40,35 +40,28 @@ Accepts the path to the acted-on document and `process-mode' instance.")
...
-                                  do (progn
-                                       (setf (thread-terminated-p mode) t)
-                                       (disable-modes* (sera:class-name-of mode) (buffer mode))
-                                       (return))
-                                else
-                                  when condition-value
-                                    do (with-current-buffer (buffer mode)
-                                         (when (action mode)
-                                           (funcall (action mode) (path-url mode) mode))))))))
+        (thread mode) (run-thread "process"
```
Ambrevar, line 48: This loses reference to a previously unterminated thread, if a thread is hung and the process-mode is re-enabled (without disablling first).

source/mode/process.lisp
```diff
@@ -40,35 +40,28 @@ Accepts the path to the acted-on document and `process-mode' instance.")
...
+                                     (return))
+                              else
+                                when condition-value
+                                  do (with-current-buffer (buffer mode)
+                                       (when (action mode)
+                                         (funcall (action mode) (path-url mode) mode)))))))
 
 (defmethod disable ((mode process-mode) &key)
   (and (cleanup mode)
        (funcall (cleanup mode) (path-url mode) mode))
```
Ambrevar, line 66: This can be replaced with `funcall*`.

aadcg: Yes, indeed! 

I have been keeping a `review-repeat-mode` branch with that change, but I'll share once we merge this and #2645. Let's keep this one focused :)

source/mode/process.lisp
aadcg: Are you suggesting bringing back `thread-terminated-p`?

aartaka: - Or using `bt:thread-alive-p` as an additional check before setting the new one.
- Or ensuring that `disable` was called before the `enable` somehow (@Ambrevar, I have a vague memory that it should happen automatically, but does it actually happen this way?)

aadcg: Good point. Let me fix that.

aadcg:
Since modes that derive from `process-mode` will most likely be dropped in the 3.0 release I'm writing a test suite to make sure it won't be too difficult to bring them back in.

aadcg:
Test suite added and I've also left some comments with some open questions.

@aartaka @Ambrevar

source/mode/process.lisp
```diff
@@ -40,24 +40,27 @@ Accepts the path to the acted-on document and `process-mode' instance.")
...
-    :documentation "Whether the thread has terminated."))
+    :documentation "Thread that `action' happen in."))
   (:toggler-command-p nil))
 
+(defmethod thread-alive-p ((mode process-mode))
+  ;; No action should mean no thread, right?
+  (and (thread mode) (bt:thread-alive-p (thread mode))))
+
+;; It would be nice this method would return the mode object but for some reason
+;; it doesn't...
```
aadcg, line 51: I mean, even when I add `mode` as the last s-exp in the method. Can you make sense of that??

aadcg: Oh right, I'm an idiot :D

10c4b79e7980ff98f43bd291c93e223a6d164113

aadcg:
Ready for review.

source/mode/repeat.lisp
```diff
@@ -95,6 +104,7 @@ It takes a `repeat-mode' instance as argument.")))
   (setf (command-dispatcher (current-window)) #'dispatch-command
         (input-skip-dispatcher (current-window)) #'dispatch-input-skip))
 
+;; FIXME doesn't work
```
Ambrevar, line 107: Nit: We seem to use `TODO` more often than `FIXME`, I suggest we stick to one convention.

source/mode/repeat.lisp
```diff
@@ -35,7 +21,30 @@
...
+It takes a `repeat-mode' instance as argument.")
+   (nyxt/process-mode:firing-condition
+    #'(lambda (path-url mode)
+        (declare (ignore path-url))
+        (sleep (repeat-interval mode))
+        (cond ((repeat-count mode)
+               (if (zerop (repeat-count mode))
+                   :return
+                   (decf (repeat-count mode))))
+              (t t)))
```
Ambrevar, line 33: This `cond` can be replaced by an `(or (and ...) t)`.

source/mode/repeat.lisp
```diff
@@ -35,7 +21,30 @@
...
-It takes a `repeat-mode' instance as argument.")))
+It takes a `repeat-mode' instance as argument.")
+   (nyxt/process-mode:firing-condition
+    #'(lambda (path-url mode)
+        (declare (ignore path-url))
+        (sleep (repeat-interval mode))
+        (cond ((repeat-count mode)
+               (if (zerop (repeat-count mode))
+                   :return
+                   (decf (repeat-count mode))))
```
Ambrevar, line 32: Maybe a potential race-condition on the counter?
Could use either a mutex or an atomic counter.  In any case, the race-safety should be handled on `process-mode` side.

tests/offline/mode/process.lisp
```diff
@@ -0,0 +1,86 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt/tests)
+(use-nyxt-package-nicknames)
+
+(define-test basic-functionality ()
```
Ambrevar, line 7: Nit: Test name is a bit generic.

tests/offline/mode/process.lisp
```diff
@@ -0,0 +1,86 @@
...
+    (let* ((action-run-p)
+           (mode (make-instance 'nyxt/process-mode:process-mode
+                                :buffer (make-instance 'modable-buffer)
+                                :path-url (quri:uri "test")
+                                :action
+                                (lambda (path-url mode)
+                                  (declare (ignore path-url mode))
+                                  (setf action-run-p t)))))
+      (enable mode)
+      (sleep 0.5)
```
Ambrevar, line 20: `sleep` is often a red flag that signals we may miss a way to _wait_ on the desired termination. What are we waiting for exactly here?  Is it for the thread termination?  If so, we have `bt:join-thread`.

tests/offline/mode/process.lisp
```diff
@@ -0,0 +1,86 @@
...
+(define-test cleanup ()
+  (let* ((clean-p)
+         (mode (make-instance 'nyxt/process-mode:process-mode
+                              :buffer (make-instance 'modable-buffer)
+                              :path-url (quri:uri "test")
+                              :cleanup
+                              (lambda (path-url mode)
+                                (declare (ignore path-url mode))
+                                (setf clean-p t)))))
+    ;; Notice that an instance of process-mode can disabled without being
```
Ambrevar, line 71: Typo: "be disabled"?

tests/offline/mode/process.lisp
```diff
@@ -0,0 +1,86 @@
...
+    ;; enabled!  Bug or feature?
```
Ambrevar, line 72: @aartaka?

tests/offline/mode/process.lisp
```diff
@@ -0,0 +1,86 @@
...
+    (assert-true clean-p)))
+
+(define-test thread-handling ()
+  (let ((mode (make-instance 'nyxt/process-mode:process-mode
+                             :buffer (make-instance 'modable-buffer)
+                             :path-url (quri:uri "test")
+                             :action
+                             (lambda (path-url mode)
+                               (declare (ignore path-url mode))))))
+    ;; Ensure that re-enabling the mode doesn't overwrite an alive thread
```
Ambrevar, line 83: Nit: Ensure to end your comment sentences with a period.

source/mode/repeat.lisp
aadcg: Sure!

source/mode/repeat.lisp
aadcg: Why not..

aadcg: Actually it's not that simple. It's more readable as is.

source/mode/repeat.lisp
aadcg: This is a good point.

I'm not very familiar with the primitives of `bordeaux-threads` so if either @Ambrevar or @aartaka want to push a fix, go ahead. Otherwise I'll leave a note in the codebase.

tests/offline/mode/process.lisp
aadcg: Good point. And it also made me realize that I should pass the `firing-condition` argument explicitly, since I don't know what the future default value will be :)

aadcg: Renamed to `run-action-ad-eternum`.

tests/offline/mode/process.lisp
aadcg: > Is it for the thread termination?

No, in this test case, the thread doesn't terminate! Notice that I have uses `join-thread` in other tests.

>  What are we waiting for exactly here?

I'm waiting for the variable action-run-p to set to `t`. Is there a better way to do this? Watch a variable until it is set?

tests/offline/mode/process.lisp
aadcg: I also prefer that!

source/mode/repeat.lisp
aartaka: See 03d3d56ddf62a76455a7fe1cedeb51cb106c7fe4 for a fix.

tests/offline/mode/process.lisp
aartaka: But wait, @aadcg, why are you instantiating them instead of `enable-mode`-ing them? `make-instance MODE-SYMBOL` should not be used except for when needing to make a dummy mode.

aartaka: But we can actually add an `:around` method code checking the mode being enabled.

source/mode/repeat.lisp
aadcg: Thanks!

tests/offline/mode/process.lisp
aadcg: > But wait, @aadcg, why are you instantiating them instead of enable-mode-ing them? make-instance MODE-SYMBOL should not be used except for when needing to make a dummy mode.

There's nothing preventing it. I'm doing it because I can, just like users won't rest until they won't break the program they're using :p

I agree with dee7d5d. Thanks.

tests/offline/mode/process.lisp
Ambrevar: lparallel `promise` seems to be the right way to do this.
You could also use serapeum `monitor` or `synchronized`.
@aartaka Any other suggestion?

aartaka: @Ambrevar, no Lparallel primitives are long due adding to Nyxt instead of things that we use now.

aadcg:
Looks ready to be merged from my side.

tests/offline/mode/process.lisp
aartaka: > There's nothing preventing it. I'm doing it because I can, just like users won't rest until they won't break the program they're using :p

Still, you're testing the API, and you're supposedly expected to _use_ the API (((( ;°Д°))))

source/mode/repeat.lisp
```diff
@@ -50,31 +59,31 @@ It takes a `repeat-mode' instance as argument.")))
...
 
 (define-command-global repeat-times (&optional times function)
   "Prompt for FUNCTION to be run a number of TIMES."
-  (let ((times (or times
-                   (ignore-errors
-                    (parse-integer
-                     (prompt1 :prompt "Repeat for X times"
-                              :input "4"
-                              :sources 'prompter:raw-source))))))
-    (when times
```
aartaka, line 72: Why remove `times`?

source/mode/repeat.lisp
```diff
@@ -50,31 +59,31 @@ It takes a `repeat-mode' instance as argument.")))
 
 (define-command-global repeat-every (&optional seconds function)
   "Prompt for FUNCTION to be run every SECONDS."
-  (let* ((seconds (or seconds
-                      (ignore-errors
-                       (parse-integer
-                        (prompt1 :prompt "Repeat every X seconds"
-                                 :input "5"
-                                 :sources 'prompter:raw-source))))))
-    (when seconds
```
aartaka, line 59: Why remove `seconds`?

aartaka (approved):

source/mode/repeat.lisp
aadcg: Did you see commit 0f7d79815a2db66e33353a004f84bb57d3085d06?

aartaka: Ah, right, `alex:when-let`...

aartaka:
Looks good to me too!

aadcg:
Thanks everyone!

PR #2639, opened 2022-11-11
Add documentation

createyourpersonalaccount:
# Description

Add documentation strings and re-implement `last-word`.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:


source/utilities.lisp
```diff
@@ -25,6 +29,12 @@
 
 (export-always 'defmemo)
 (defmacro defmemo (name params &body body) ; TODO: Replace with https://github.com/AccelerationNet/function-cache?
+  "Define a new function with a hash table cache.
```
aartaka, line 32: "Memoized" instead of "with a hash table cache" would work fine too, because most Lispers know of memoization.

source/utilities.lisp
```diff
@@ -25,6 +29,12 @@
...
+
+Functionally equivalent to (defun NAME PARAMS BODY) but the function
```
aartaka, line 34: "Functionally equivalent to defun"?

source/utilities.lisp
```diff
@@ -53,7 +63,8 @@ Particularly useful to avoid errors on already terminated threads."
 
 (export-always 'ensure-file-exists)
 (defun ensure-file-exists (pathname)
-  (open pathname :direction :probe :if-does-not-exist :create))
+  "Create file PATHNAME if it does not exist."
+  (open filepath :direction :probe :if-does-not-exist :create))
```
aartaka, line 67: Where does `filepath` come from? The argument is called `pathname`...

While we're at it, how about returning `pathname` from the function so that it's more composable?

source/utilities.lisp
```diff
@@ -152,9 +163,10 @@ Return the lambda s-expression as a second value, if possible."
 (-> last-word (string) string)
 (export-always 'last-word)
 (defun last-word (s)
-  (if (uiop:emptyp s)
-      ""
-      (the (values string &optional) (alex:last-elt (sera:words s)))))
+  "Last substring of alphanumeric characters, or empty if none."
+  (let ((words (sera:words s)))
+    (the (values string)
+         (if words (alex:last-elt words) ""))))
```
aartaka, line 169: I'd better return NIL, because it's a conventional "nothingness" value, and it's much easier to work with than a mere empty string. And the behavior of `last-word` would be more intuitive there, when returning NIL:
- If there's any word, then return the last word, as a string;
- If there are none, then return NIL to signify absence of words.

This is also consistent with underlying `sera:words`.

source/utilities.lisp
createyourpersonalaccount: Good catch! I fixed it.

source/utilities.lisp
createyourpersonalaccount: It means you would use the macro as you would use defun.

source/utilities.lisp
createyourpersonalaccount: Memoization has many techniques, this is using a simple hash table. It explains a bit of the internals in this way. If you'd like, I can change the docstring to `Define a memoized function.` and add a comment about the method of memoization.

source/utilities.lisp
createyourpersonalaccount: Maybe I can do this on a follow up PR later?

aartaka: But that change is one of the main points of _this_ PR 0_o

source/utilities.lisp
aartaka: Exactly. That's why it's not necessary to expand it as `(defun NAME PARAMS BODY)`— a mere `defun` mention should be enough.

source/utilities.lisp
aartaka: Yes, that would be preferable. 

Alternatively, explain the mechanism on some docstring line that's not the first one. The first line of the docstring is for function purpose, not for its implementation details.

source/utilities.lisp
createyourpersonalaccount: I think this is okay as it is.

aartaka: I tend to not agree. What's the use in returning an empty string? One has to use expensive (both CPU-wise and mental juggling-wise) `string=` or `str:emptyp` to check that the `last-word` does not indeed exist.

Instead, if `last-word` returns NIL, one can easily test it with `null`, `not`, make boolean operations with it (like `(or (last-word ...) "default")`), or execute something conditionally:
``` lisp
(alex:if-let ((word (last-word ...)))
  (do-something-with word)
  (error "No word!"))
``` 
This last snippet would be painfully clunky with empty string version:
``` list
(let ((word (last-word ...)))
  (if (not (str:emptyp word))
      (do-something-with word)
      (error "No word!")))
```

I mean, `last-word` is a small utility and I may sound picky, but we have to preserve the intuitiveness at all levels of our APIs there :)

source/utilities.lisp
aartaka: So, leaving it as-is? 

Bear in mind: if you refer to some cons structure in the docstring, it'd better be reproduced as precisely as possible. In this case, the form should rather be `(defun NAME PARAMS . BODY)` or `(defun NAME PARAMS &body BODY)` so that one understands that BODY is not a single form and is expanded into `defun` body.

At this point I believe that referring to defun is only meaningful as a mere `defun` (instead of this `(defun ....)` thing), because it's less confusing and bears the same meaning. You've said it yourself: "It means you would use the macro as you would use defun.", which is even better of a documentation than providing an incorrect expansion :P

source/utilities.lisp
createyourpersonalaccount: I am just keeping the behavior as it was. I did not want to change the behavior of `last-word`, just to fix the bug. The behavior can change in a future PR.

aartaka: @aadcg, pull us apart, I'm getting stubborn about this small thing :P

source/utilities.lisp
createyourpersonalaccount: Okay, I fixed it.

aartaka: Thanks!

source/utilities.lisp
createyourpersonalaccount: @aartaka I am not going to change the function now, it can be done in a later commit. Don't worry about it.

aartaka: Let it be then.

aartaka:
Merging then?

source/utilities.lisp
```diff
@@ -13,10 +13,14 @@
...
-(alex:define-constant +newline+ (string #\newline) :test #'equal)
+(alex:define-constant +newline+ (string #\newline)
+  :test #'equal
+  :documentation "A string containing a newline.")
 
 (export-always '+escape+)
-(alex:define-constant +escape+ (string #\escape) :test #'equal)
+(alex:define-constant +escape+ (string #\escape)
+  :test #'equal
+  :documentation "A string containing the escape character.")
```
Ambrevar, line 23: Not sure these two bring anything.

source/mode/document.lisp
```diff
@@ -119,11 +119,15 @@ It does not assume being online."))
 
 (export-always '%clicked-in-input?)
 (defun %clicked-in-input? (&optional (buffer (current-buffer)))
+  "Get the class of the active element in BUFFER."
```
Ambrevar, line 122: "class" is ambiguous here because Common Lisp has classes two.
Plus this is a predicate (as indicated by the `?` suffix) and I don't think we ever use the return value, so it's safer to write "Return non-nil if an HTML element is selected for input."

source/utilities.lisp
```diff
@@ -25,6 +29,11 @@
 
 (export-always 'defmemo)
 (defmacro defmemo (name params &body body) ; TODO: Replace with https://github.com/AccelerationNet/function-cache?
+  "Define a new memoized function named NAME in the global environment.
+
+Functionally equivalent to `defun' but the function stores its
+computations in a hash table, and remembers its passed parameters when
```
Ambrevar, line 35: "hash table" is an implementation detail that is prone to change, you can leave it out.

source/mode/document.lisp
createyourpersonalaccount: It is used in https://github.com/atlas-engineer/nyxt/blob/bd1d1f7686111d534631a2e8e032cc3d87a16ec1/source/mode/document.lisp#L136-L137

source/utilities.lisp
createyourpersonalaccount: done

source/mode/document.lisp
```diff
@@ -119,11 +119,15 @@ It does not assume being online."))
...
   (ps-eval :buffer buffer (ps:@ document active-element tag-name)))
 
 (export-always 'input-tag-p)
 (-> input-tag-p ((or string null)) boolean)
 (defun input-tag-p (tag)
+  "Is this TAG inputtable?
```
aadcg, line 128: We never write docstrings in question form. Please use "whether".

source/utilities.lisp
```diff
@@ -53,6 +63,7 @@ Particularly useful to avoid errors on already terminated threads."
 
 (export-always 'ensure-file-exists)
 (defun ensure-file-exists (pathname)
+  "Create file PATHNAME if it does not exist."
```
aadcg, line 66: "unless it exists".

source/utilities.lisp
aadcg: Sorry somehow this went under my radar...

source/mode/document.lisp
aadcg: @createyourpersonalaccount so you're saying that `%clicked-in-input?` is not a predicate despite its misleading naming, correct?

aadcg: Please change the name of `%clicked-in-input?`. And, indeed, this is not a class but an HTML element.

Ambrevar: @createyourpersonalaccount Good catch!

source/mode/document.lisp
```diff
@@ -117,9 +117,9 @@ It does not assume being online."))
        "space" 'scroll-page-down
        "pagedown" 'scroll-page-down)))))
 
-(export-always '%clicked-in-input?)
-(defun %clicked-in-input? (&optional (buffer (current-buffer)))
-  "Get the class of the active element in BUFFER."
+(export-always 'tag-of-active-element)
+(defun tag-of-active-element (&optional (buffer (current-buffer)))
```
Ambrevar, line 121: What about the maybe more idiomatic `active-element-tag`?

aadcg:
I've reworked the commits and pushed to master. @createyourpersonalaccount see 62e8b704b, 20c1e89c6 and 20c1e89c6.

Usually when a PR is ready to merge, we rebase it, squash it and send a message saying that it's ready to be shipped. 

Thank you for your patience.

PR #2645, opened 2022-11-17
Fix enable modes

aadcg:
# Description

We need to handle arguments passed to modes when using `enable-modes*`.

I've added a simple test. When that passes, we're good to merge.

# Discussion

@aartaka I believe you'd be the best qualified person to take a look at this. If you're too busy, I can do it.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:


aartaka:
I see the test, I understand it's broken, but I've no clue how to fix it yet. But yes, I guess I'm taking it, given that I have to polish auto-rules anyway :P

aadcg:
Just to clarify @aartaka, the bug was NOT introduced by auto rules at all! 

The only reason why I said that you might be the best qualified person to take a look is because you have recently introduced `enable-modes*` and the cascading method combination for mode enabling.

aartaka:
Yup, makes sense!

aartaka:
@aadcg, 182fef9ed0aa6211319b43dfae076b81ae596937 should fix mode initargs not being propagated to mode slots. However, I'm not satisfied with the implementation—do we actually need to do all this slot inspection there? I mean, we can do it, and that's a blessing, but do we have to? Maybe there's a better way, like using `reinitialize-instance` etc.?

CC @Ambrevar, @jmercouris, @createyourpersonalaccount.

createyourpersonalaccount:
@aartaka What is the test testing for? My opinion for slots is that they're private and instead accessors, which are methods, should be documented. I don't think that's the status quo in Nyxt; I am not sure how to evaluate this issue, what does it mean "set slot values from initargs"? Isn't initargs just a keyword for `make-instance` for a slot value?

aartaka:
> @aartaka What is the test testing for? 

To make sure that the arguments passed to `enable` method end up in the method slots.

> My opinion for slots is that they're private and instead accessors, which are methods, should be documented. 

Not necessarily. Slots are not private, and most are even exported in most CL object-oriented code. Some libraries don't even use accessors and only export slot names to be used in `with-slots` or `slot-value`. So no, slots are not private, and they are valid API too.

Another matter is validation and completing the values of slots:
- `slot-value` gets you the raw value of the slot, not doing anything to it but getting it from the object. It also allows to set the slot to anything, not doing much checks but type ones (only if the `safety` is high, though).
- Custom accessors allow you to check the value being fetched/set and somehow modify it. They are more of an external API, doing intutitive actions, and should be used before `slot-value` if exported.

This particular case is different and requires using `slot-value`, let me explain below.

> I don't think that's the status quo in Nyxt; I am not sure how to evaluate this issue, what does it mean "set slot values from initargs"? Isn't initargs just a keyword for `make-instance` for a slot value?

Yes, you most often see inititargs in `make-instance` and `(re)initialize-instance` arguments, but any code that implicitly uses those also works with the idea of initargs. `enable` for Nyxt modes is a kind of second initialization, setting up some values and launching vital actions. That's one reason to talk about "initargs" as `enable` arguments here.

Another reason is that we don't initialize modes ourselves, we do that under the carpet of `enable`, passing an already initialized mode instance to it.

-----------------------------

So, to rephrase my initial question: is there any better way to set slot values in `enable`, short of manually parsing keyword arguments and using `slot-value`? I'm leaning more and more to `reinitialize-instance`.

createyourpersonalaccount:
> To make sure that the arguments passed to `enable` method end up in the method slots.

I see. You could put this in the docstring of the unit test if you'd like.

aartaka:
@aadcg, not that the fix is in, who should take care of merging and other details? Should I finish it?

aadcg:
Let me please have a final look - [maybe tomorrow, maybe today](https://www.youtube.com/watch?v=SURGrWNK4-4&t=25).

Ambrevar:
`reinitialize-instance` seems fine, but can you share an example of what you are trying to fix?  If it's just passing slot values, one problem is that `reinitialize-instance` is going to do too much, like instantiating twice (especially bad the first time the mode is enabled).

aartaka:
> `reinitialize-instance` seems fine, but can you share an example of what you are trying to fix? If it's just passing slot values, one problem is that `reinitialize-instance` is going to do too much, like instantiating slots twice the first time the mode is enabled.

So yes, I'm trying to set slots from the keyword arguments of `enable`. And `reinitialize-instance` I doubted about using there, because it's indeed an overkill for such a use-case. But then, 
- we don't have any complex `shared-initialize`/`initialize-instance` logic for modes, 
- and our `enable` indeed is a second initialization for the mode,

so I think it is a reasonable stretch to rely on `reinitialize-instance` there.

Ambrevar:
Can we somehow skip the `reinitialize-instance` call when the mode is first instantiated?  We might need an extra level of indirection.

Ambrevar:
> * we don't have any complex `shared-initialize`/`initialize-instance` logic for modes,

Indeed, but we might in the future, plus this would be prohibitive for extension writers who rely on heavy slot instantiation.

aartaka:
I am leaning to somehow entangling `initialize-instance` with `enable` to escape this double initialization problem...

aadcg:
> I am leaning to somehow entangling `initialize-instance` with `enable` to escape this double initialization problem...

That seems saner to me.

Keeping it open right?

aadcg:
Rebased over master and squashed uninteresting details.

aartaka:
Okay, I've reverted the `reinitialize-instance` commit, because indeed, `reinitialize-instance` does too much to be used as-is. I believe this is ready for merging :P

source/mode.lisp
```diff
@@ -102,7 +102,14 @@ functionality. A `cascade' method combination is used for that.
...
 
 (defmethod enable :before ((mode mode) &rest keys &key &allow-other-keys)
-  (apply #'reinitialize-instance mode keys))
+  ;; Using class-direct-slots here because `enable' will cascade to parent modes anyway.
+  ;; FIXME: An easier way to initialize slots given initargs?
+  (loop with slot-defs = (closer-mop:class-direct-slots (class-of mode))
+        for (key value) on keys by #'cddr
+        do (alex:when-let ((slot-name (loop for slot-def in slot-defs
+                                            when (member key (c2mop:slot-definition-initargs slot-def))
+                                              do (return (c2cl:slot-definition-name slot-def)))))
```
Ambrevar, line 111: You can use `mopu` and `find` to simplify this inner `loop`.

source/mode.lisp
```diff
@@ -102,7 +102,14 @@ functionality. A `cascade' method combination is used for that.
...
+             (setf (slot-value mode slot-name) value))))
```
Ambrevar, line 112: You should use the slot writer if any, nah?

aartaka: No! See https://github.com/atlas-engineer/nyxt/pull/2645#issuecomment-1319868401 for my reasoning there :)

source/mode.lisp
aartaka: I remember you suggesting that already, and I remember that `member`/`union` as a `:test` function has convinced me against going with `find` there. Can you expand on what fourm you have in mind? I cannot conceive anything both short and readable, which would fulfill the same purpose as this `loop`...

source/mode.lisp
aadcg: After reading the comment you mention, I'm still confused I'm afraid. You seem to suggest that using accessors instead of `slot-value` has some undesired effect in this particular case. Could you specify?

Ambrevar: Accessors may do non-functional things, for instance depending on the current time or the value of other slots.  This is why using them may give you a more relevant result than using `slot-value` (which would give you a result closer to the original instance).

In any case, it's debatable.  Maybe add a `TODO` comment then?

aadcg: Where can I find an example or a reference for that?

aartaka: We're not using accessors in `initialize-instance` (because we don't have access to it; but even if we did have access to `intiialize-instance` insternals, we probably won't use accessors), because that's setting slot values to their initial values before the object starts doing any meaningful activity and needing any accessor intuitive actions. That's why I'm also using `slot-value` in `enable :before` method—it's also part of mode initialization. 

Note that particular `enable` methods can use accessors if they wish, I'm only conserned with setting proper slot values `:before` those start happening.

I'll add the `TODO` comment, yes.

aadcg: Ok, now it all makes sense!

aadcg:
Ready to merge then? @aartaka @Ambrevar

aartaka:
Ready on my side, yes!

aadcg:
Thanks!

PR #2653, opened 2022-11-18
Prompter object attributes display

### LINKED ISSUE #2652: Separate attribute text and their display code in `promtper:object-attributes`
In ed5e2f8a5e42f0a500cb2b1543f34df2997f264c, I've introduced a good idea for a color picker: displaying the color suggestions in the color they represent. Inline, as suggestions. It certainly is a good idea, but I had to hack it to be able to preserve both
- `tab`-completion of suggestions, and
- pretty HTML-ized display.

The problem is: when you provide a single string as an attribute value, you can style it with HTML, but `tab`-completing it will also spill some HTML into the prompt input. I solved this with a (temprorary, hopefully) split of the color source into two attributes: one, default, listing jsut the name of the color, and the other one displaying it in a pretty way. But that must go.

The simplest solution I envision is adding an optional third list element to every column in the `prompter:object-attributes` listing, solely responsible for display of the suggestion. This way, we both preserve `prompter` ignorance of the display (it only uses two of the list elements, not more), and have a simple way to split content from display. And no need for `html-string-p` too!

So that this:
``` lisp
(defmethod prompter:object-attributes ((color string) (source color-source))
  `(("Name" ,color)
    ("Color" ,(let ((spinneret:*html-style* :tree))
                (spinneret:with-html-string
                  (:span :style (format nil "background-color: ~a; color: ~a; border: 0.3empx solid ~a; border-radius: 0.1em"
                                        color (contrasting-color color) (contrasting-color color))
                         color))))))
```
becomes this:
``` lisp
(defmethod prompter:object-attributes ((color string) (source color-source))
  `(("Color" ,color ,(let ((spinneret:*html-style* :tree))
                       (spinneret:with-html-string
                         (:span :style (format nil "background-color: ~a; color: ~a; border: 0.3empx solid ~a; border-radius: 0.1em"
                                               color (contrasting-color color) (contrasting-color color))
                                color))))))
```

We can also:
- Define methods for display, as @Ambrevar said.
- Redefine `prompter::default-attribute-value`, hacking the display/tab-completion for a particular source. But that requires definining our own class of suggestions, which is too much hustle for such a seemingly simple problem. It should be easier.

**Additional context**
Quite a long time ago, I allowed the attribut values to contain HTML, so that we can style those. It was a good solution to the problem of our prompter being dull in appearance by default, but I didn't consider the fact that attribute values are also `tab`-completed.

aartaka:
# Description

One of the possible shots at #2652: add a third value to `prompter:object-attributes` listing. Simplest solution, and quite intutive one, I believe.

Fixes #2652

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [x] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -403,8 +403,11 @@ call."))
                   (princ-to-string (rest pair))))
                object))
       (t (call-next-method))))
-  (:documentation "Return an alist of non-dotted pairs (ATTRIBUTE-KEY ATTRIBUTE-VALUE) for OBJECT.
+  (:documentation "Return an alist of non-dotted lists (ATTRIBUTE-KEY ATTRIBUTE-VALUE ...) for OBJECT.
 Attributes are meant to describe the OBJECT in the context of the SOURCE.
+
+The attributes after the first two are for the application specific purposes,
+like format strings or code for element display.
 
 Both returned attribute-keys and attribute-values are strings (if not, they are
 automatically converted to `princ-to-string'). If the attribute value is a
@@ -454,11 +457,6 @@ Suggestions are made with the `suggestion-maker' slot from `source'."))
        (or (not (listp (rest object)))
            (null (rest (rest object))))))
 
-(defun undotted-pair-p (object)
-  (and (listp object)
-       (listp (rest object))
-       (null (rest (rest object)))))
-
 (defun alist-p (object)
   "Return non-nil if OBJECT is an alist, dotted or undotted."
   (and (listp object)
@@ -467,7 +465,7 @@ Suggestions are made with the `suggestion-maker' slot from `source'."))
 (defun undotted-alist-p (object &optional value-type)
   "If VALUE-TYPE is non-nil, check if all values are of the specified type."
   (and (listp object)
-       (every #'undotted-pair-p object)
+       (every #'listp object)
        (or (not value-type)
            (every (lambda (e) (typep (first e) value-type))
                   (mapcar #'rest object)))))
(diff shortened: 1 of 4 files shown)


libraries/prompter/prompter-source.lisp
```diff
@@ -399,9 +399,12 @@ call."))
                   (princ-to-string (rest pair))))
                object))
       (t (call-next-method))))
-  (:documentation "Return an alist of non-dotted pairs (ATTRIBUTE-KEY ATTRIBUTE-VALUE) for OBJECT.
+  (:documentation "Return an alist of non-dotted lists (ATTRIBUTE-KEY ATTRIBUTE-VALUE ...) for OBJECT.
 Attributes are meant to describe the OBJECT in the context of the SOURCE.
 
+The attributes after the first two are for the application specific purposes,
+like format strings or code for element display.
+
```
Ambrevar, line 407: Maybe explain why they are not pairs and that the calling site may leverage this.

Ambrevar (commented):
Looks good to me!

libraries/prompter/prompter-source.lisp
aartaka: Done in baa95fb0332a9430406f5616b978ae7d1fede2e8. Note that the non-pair character of attributes is mentioned one paragraph above, in:
```
Return an alist of non-dotted lists (ATTRIBUTE-KEY ATTRIBUTE-VALUE ...) for OBJECT.
```

aartaka:
This feature should be documented in prompter/prompt-buffer docs, but given that those are practically non-existent at the moment, I'll leave this to the future documenter of these :P (I remember @aadcg saying something about this being necessary and even opening an issue about it, but I can't find the issue now 0_o)

Merging this.

aartaka:
Cool, André!

Issue #2527, opened 2022-08-29
Rename `marks` to `mark`?


aadcg:
#2520 introduced `marks-actions`, which adds to `return-actions` and `selection-actions`.

However, `marks-actions` sounds really weird and it's hard to pronounce. 

I'd rename `marks` to `mark`. Indeed, `selection` and `mark` are radically different since the former is always a singleton but the latter isn't. They're actually both lists, but for very different reasons related to the way they're implemented. 

It doesn't sound odd to me that `mark` would a be concept defined by a word in the singular form that refers to a set of entities.

aadcg:
On second thoughts, I'd suggest something even more radical.

We have the following concepts: suggestions, selection and marks.

Wouldn't it be saner to have suggestions, current-suggestion and selection?

Suggestions are the prompt options; current-suggestion is a singleton of suggestions; and selection is a subset of suggestions.

aartaka:
I guess the initial naming comes from Helm. If the new naming is drains less cognitive resources, then why not :)

I'll leave that to @Ambrevar, though, as he wanted to abstract `promtper` into a separate library, and that is likely when the renaming should happen.

Ambrevar:
Yes, the naming comes from Helm.  "Mark" is not very familiar outside of this context, so yes, I'm in favour of André's renaming suggestion.

Shall it wait for 4.0 though?  We've already got a lot on our hands...

aadcg:
Yes, this is not particularly important.

PR #2655, opened 2022-11-18
Rename prompter concepts

### LINKED ISSUE #2601: Rename `prompter` concepts
While working on #2555, I noticed that `prompter`'s concepts can be difficult to internalize.

I suggest the following rename (reasoning below):

- `selection` -> `current-suggestion`
- `selection-actions` -> `current-suggestion-actions`
- `marks` -> `selection`
- `marks-actions` -> `selection-actions`

The key observation is that `marks` only exist when `multi-selection-p` is non-nil, which immediately makes us think of a selection (a set of items). But that collides with our concept for "the current suggestion", which we call it `selection`...

In #2527 I have proposed `marks` to be a singular concept, and `selection` falls in that category too. This proposal makes #2527 obsolete.

With this rename the overview of the `prompter` library would be as follows.  
 

> The key objects are =prompters=, =sources= and =suggestions=.
> 
> A =prompt= is an interface for user interactions that holds one or more
> =sources=, and each of those are populated by =suggestions=.
> 
> Other central concepts include:
> 
> - =prompt=
>    + =current-suggestion= :: A single =suggestion=, if any.  Think of it as the currently
>      selected =suggestion=, belonging to a =source=.
> - =source=
>    + =selection= :: A list of =suggestions=, when =multi-selection-p= is non-nil.
>    + =actions= :: A list of functions that run on =suggestions=.
>       - =return-actions= :: General purpose.
>       - =selection-actions= :: On =selection= change (event-driven).
>       - =current-suggestion-actions= :: On =current-suggestion= change (event-driven).
>

aadcg:
# Description

Fixes #2601 #2527 

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/libraries/prompter/README.org
+++ b/libraries/prompter/README.org
@@ -8,28 +8,31 @@ A =prompt= is an interface for user interactions that holds one or more
 Other central concepts include:
 
 - =prompt=
-   + =selection= :: A single =suggestion=, in any.  Think of it as the currently
-     selected =suggestion=, belonging to a =source=.
+   + =current-suggestion= :: A single =suggestion=, if any.
 - =source=
-   + =marks= :: A list of =suggestions=, when =multi-selection-p= is non-nil.
+   + =marks= :: A list of =suggestions=, when =enable-marks-p= is non-nil.
    + =actions= :: A list of functions that run on =suggestions=.
-      - =return-actions= :: General purpose.
-      - =marks-actions= :: On =marks= change (event-driven).
-      - =selection-actions= :: On =selection= change (event-driven).
+      - =actions-on-return= :: Meant to be called over =marks= (or
+        =current-suggestion=, when =enable-marks-p= is nil) when successfully
+        returning =prompt=.  By default, it's the =identity= function.
+      - =actions-on-marks= :: On =marks= change (event-driven).
+      - =actions-on-current-suggestion= :: On =current-suggestion= change
+        (event-driven).
 
 Example: Find below a graphical visualization of a single prompt with sources 1
-and 2, and suggestions A, B, C and D.  The =marks= is the list composed by
-Suggestions A and C.  The =selection= is Suggestion B.
+and 2, and suggestions A, B, C and D.  =Marks= is the list composed by
+Suggestions A and C.  The =current-suggestion= is Suggestion B.
 
 [[file:example.png]]
 
 Remarks:
 
-A =prompt= always has a single =selection=, whereas a =source= has either a
-single =selection= or none (when =prompt= has multiple =sources=).
+When =prompt= has multiple =sources=, while =current-suggestion= is always
+defined for =prompt=, it is empty for all but one of its =sources=.
 
-=marks= is a concept related to =source= not =prompt=, unlike that of
-=selection=.
+=Marks= is a concept related to =source= not =prompt=, unlike that of
+=current-suggestion=.  Thus, from the point of view of a =prompt='s object, the
+=marks= are the union of the =marks= of each of the =prompt='s =source=s.
 
 * Features
 
@@ -45,8 +48,8 @@ Non-exhaustive list of features:
 - Notifications sent when suggestion list is updated.
 - Per-source history.
 - Resumable prompters.
-- Selection actions (automatically run persistent actions on selection change).
-- Marks actions (automatically run persistent actions on marks change).
+- Marks actions (event-driven on marks change).
+- Current suggestion actions (event-driven on current suggestion change).
 - Automatically return the prompt when narrowed down to a single suggestion.
 
 This library is heavily inspired by Emacs' minibuffer and [[https://emacs-helm.github.io/helm/][Helm]].  It only deals
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -110,18 +110,18 @@ list.")
     :documentation "The list of `suggestion' values which have been marked by
 the user.
 
-Marking is only allowed when `multi-selection-p' is non-nil.  When suggestions
-are marked, subsequent `return-actions' run over all marked suggestions.
+Marking is only allowed when `enable-marks-p' is non-nil.  When suggestions
+are marked, subsequent `actions-on-return' run over all marked suggestions.
 
 We store the values instead of the `suggestion' because `suggestion' objects are
 reinstantiated between each input processing.")
 
-   (marks-actions
+   (actions-on-marks
     '()
     :type (or null function function-symbol
               (cons (or function function-symbol) *))
-    :documentation "The first function of this list is called automatically on
-the mark when it's changed.
+    :documentation "The first function of this list is called automatically when
+the marks change.
 It does not interrupt or return the prompter.
 For convenience, it may be initialized with a single function, in which case it
 will be automatically turned into a list.")
@@ -207,7 +207,7 @@ recompute the match-data for instance.")
     :documentation "A predicate used to sort the `suggestion's once filtered.
 The predicate works the same as the `sort' predicate.")
 
-   (return-actions
+   (actions-on-return
     #'identity
     :type (or null
               (or function function-symbol)
@@ -215,7 +215,7 @@ The predicate works the same as the `sort' predicate.")
     :accessor nil
     :export nil
     :documentation "List of funcallables that can be run on `suggestion's of
-this source.  This is the low-level implementation, see the `return-actions'
+this source.  This is the low-level implementation, see the `actions-on-return'
 function for the public interface.
 For convenience, it may be initialized with a single function or symbol, in
 which case it will be automatically turned into a list.")
@@ -276,7 +276,7 @@ See `attribute-channel'.")
     :documentation "Channel used to communicate attributes to `attribute-thread'
 to compute asynchronously.")
 
-   (multi-selection-p
+   (enable-marks-p
     nil
     :type boolean
     :documentation "Whether multiple `suggestion's can be marked.")
@@ -288,27 +288,27 @@ to compute asynchronously.")
 the prompter is resumed.
 See `resume-sources'.")
 
-   (selection-actions
+   (actions-on-current-suggestion
     '()
     :type (or null function function-symbol
               (cons (or function function-symbol) *))
     :documentation "The first function of this list is called automatically on
-the selection when it's changed.
+the current-suggestion when it's changed.
 It does not interrupt or return the prompter.
 For convenience, it may be initialized with a single function, in which case it
 will be automatically turned into a list.")
 
-   (selection-actions-enabled-p
+   (actions-on-current-suggestion-enabled-p
     nil
     :type boolean
-    :documentation "Whether the first of `selection-actions' is automatically
-executed.  Also see `selection-actions-delay'.")
+    :documentation "Whether the first of `actions-on-current-suggestion' is
+automatically executed.  Also see `actions-on-current-suggestion-delay'.")
 
-   (selection-actions-delay
+   (actions-on-current-suggestion-delay
     0.0
     :type alex:non-negative-real
-    :documentation "Time in seconds after which `selection-actions' run.
-Also see `selection-actions-enabled-p'."))
+    :documentation "Time in seconds after which `actions-on-current-suggestion'
+run.  Also see `actions-on-current-suggestion-enabled-p'."))
   (:export-class-name-p t)
   (:export-accessor-names-p t)
   (:accessor-name-transformer (class*:make-name-transformer name))
@@ -331,11 +331,11 @@ call."))
 
 (defmethod (setf marks) (value (source prompter:source))
   (setf (slot-value source 'marks) value)
-  (alex:when-let ((action (first (marks-actions source))))
-    (run-thread "Prompter mark action thread" (funcall action (marks source)))))
+  (alex:when-let ((action (first (actions-on-marks source))))
+    (run-thread "Prompter marks action thread" (funcall action (marks source)))))
 
-(defmethod default-selection-action ((source prompter:source))
-  (first (selection-actions source)))
+(defmethod default-action-on-current-suggestion ((source prompter:source))
+  (first (actions-on-current-suggestion source)))
 
 (export-always 'object-attributes)
 (defgeneric object-attributes (object source)
@@ -536,8 +536,8 @@ Attributes are set with `object-attributes'."
                  :value value
                  :attributes (object-attributes value source)))
 
-(defmethod default-return-action ((source source))
-  (first (slot-value source 'return-actions)))
+(defmethod default-action-on-return ((source source))
+  (first (slot-value source 'actions-on-return)))
 
 (define-class yes-no-source (source)
   ((name "Confirm")
@@ -563,7 +563,7 @@ Attributes are set with `object-attributes'."
    (filter nil)
    (hide-attribute-header-p :always)
    (hide-suggestion-count-p t)
-   (multi-selection-p nil))
+   (enable-marks-p nil))
   (:export-class-name-p t)
   (:accessor-name-transformer (class*:make-name-transformer name))
   (:documentation "Prompt source for raw user input.
@@ -587,7 +587,7 @@ If you are looking for a source that just returns its plain suggestions, use `so
    (filter-preprocessor 'make-word-suggestions)
    (filter nil)
    (hide-attribute-header-p :always)
-   (multi-selection-p t))
+   (enable-marks-p t))
   (:export-class-name-p t)
   (:accessor-name-transformer (class*:make-name-transformer name))
   (:documentation "Prompt source for user input words."))
@@ -628,10 +628,11 @@ If you are looking for a source that just returns its plain suggestions, use `so
         (calispel:! wait-channel t)))
     ;; Wait until above thread has acquired the `initial-suggestions-lock'.
     (calispel:? wait-channel))
-  (setf (selection-actions source) (uiop:ensure-list (selection-actions source)))
-  (setf (marks-actions source) (uiop:ensure-list (marks-actions source)))
-  (setf (slot-value source 'return-actions)
-        (uiop:ensure-list (slot-value source 'return-actions)))
+  (setf (actions-on-current-suggestion source)
+        (uiop:ensure-list (actions-on-current-suggestion source)))
+  (setf (actions-on-marks source) (uiop:ensure-list (actions-on-marks source)))
+  (setf (slot-value source 'actions-on-return)
+        (uiop:ensure-list (slot-value source 'actions-on-return)))
   source)
 
 (export-always 'attributes-keys-non-default)
--- a/libraries/prompter/prompter.lisp
+++ b/libraries/prompter/prompter.lisp
@@ -58,13 +58,13 @@ with `make-instance' and the result is used in this slot.
 
 See also `make'.")
 
-     (selection
+     (current-suggestion
       '()
       ;; TODO: Index by (source-index suggestion-index) instead?
       ;; TODO: Use structure?
       :type list
       :export nil
-      :reader selection
+      :reader current-suggestion
       :documentation "A pair of source and suggestion index.")
 
      (constructor
@@ -84,15 +84,13 @@ It's called before the sources are cleaned up.")
       :type (or null function)
       :documentation "Last function called with no parameters when calling the
 `destroy' function over this prompter.
-It's called after the sources are cleaned up.
-
-Note that the function is executed *before* performing a return-action.")
+It's called after the sources are cleaned up.")
 
      (auto-return-p
       nil
       :type boolean
-      :documentation "Whether the default `return-action' automatically runs
-when the suggestions are narrowed down to just one item.")
+      :documentation "Whether the action returned by `default-action-on-return'
+automatically runs when the suggestions are narrowed down to just one item.")
 
      (history
       (make-history)
@@ -103,7 +101,7 @@ If nil, no history is used.")
      (result-channel
       (make-channel 1)
       :type calispel:channel
-      :documentation "Channel to which the selection is sent on exit.
+      :documentation "Channel to which the `current-suggestion' is sent on exit.
 Caller should also listen to `interrupt-channel' to know if the prompter was cancelled.")
 
      (interrupt-channel
@@ -159,32 +157,33 @@ computation is not finished.")))
                        (t (error "Bad source specifier ~s." source-specifier))))
                    (uiop:ensure-list specifiers))))
     (alex:appendf (sources prompter) (ensure-sources sources)))
-  (select-first prompter)
+  (first-suggestion prompter)
   (maybe-funcall (constructor prompter) prompter)
   (update-sources prompter (input prompter))
   prompter)
 
-(defmethod (setf selection) (value (prompter prompter))
-  (setf (slot-value prompter 'selection) value)
-  (call-selection-action prompter))
+(defmethod (setf current-suggestion) (value (prompter prompter))
+  (setf (slot-value prompter 'current-suggestion) value)
+  (run-action-on-current-suggestion prompter))
 
-(export-always 'call-selection-action)
-(defmethod call-selection-action ((prompter prompter))
-  (sera:and-let* ((source (selected-source prompter))
-                  (_ (selection-actions-enabled-p source))
-                  (action (default-selection-action source))
-                  (suggestion (selected-suggestion prompter)))
-    (let ((delay (selection-actions-delay source)))
+(export-always 'run-action-on-current-suggestion)
+(defmethod run-action-on-current-suggestion ((prompter prompter))
+  (sera:and-let* ((source (current-source prompter))
+                  (_ (actions-on-current-suggestion-enabled-p source))
+                  (action (default-action-on-current-suggestion source))
+                  (suggestion (%current-suggestion prompter)))
+    (let ((delay (actions-on-current-suggestion-delay source)))
       (if (plusp delay)
-          (run-thread "Prompter selection action thread"
+          (run-thread "Prompter current suggestion action thread"
             (sleep delay)
             (funcall action (value suggestion)))
           (funcall action (value suggestion))))))
 
-(export-always 'set-selection-action)
-(defmethod set-selection-action (value (prompter prompter))
-  (setf (selection-actions (selected-source prompter))
-        (cons value (delete value (selection-actions (selected-source prompter))))))
+(export-always 'set-action-on-current-suggestion)
+(defmethod set-action-on-current-suggestion (value (prompter prompter))
+  (setf (actions-on-current-suggestion (current-source prompter))
+        (cons value
+              (delete value (actions-on-current-suggestion (current-source prompter))))))
 
 (export-always 'input)
 (defmethod (setf input) (text (prompter prompter))
@@ -193,7 +192,7 @@ computation is not finished.")))
     (unless (string= old-input text)
       (setf (slot-value prompter 'input) text)
       (update-sources prompter text)
-      (select-first prompter)))
+      (first-suggestion prompter)))
   text)
 
 (export-always 'destroy)
@@ -210,12 +209,13 @@ Signal destruction by sending a value to PROMPTER's `interrupt-channel'."
   (calispel:! (sync-interrupt-channel (sync-queue prompter)) t)
   (calispel:! (interrupt-channel prompter) t))
 
-(defun select (prompter steps &key wrap-over-p)
-  "Select `suggestion' by jumping STEPS forward.
+(defun set-current-suggestion (prompter steps &key wrap-over-p)
+  "Set PROMPTER's `current-suggestion' by jumping STEPS forward.
 If STEPS is 0, do nothing.
 If STEPS is negative, go backward.
-If the currently selected suggestion is the last one of the current source, go
-to next source, or previous source if STEPS is negative."
+When the current suggestion is the last (resp. first) one of the current source,
+return the first (resp. last) suggestion of the next (resp. previous) source
+when STEPS is positive (resp. negative)."
   (unless (= 0 steps)
     (labels ((index->source (index &optional (sources (sources prompter)))
                (let ((limit (length (suggestions (first sources)))))
@@ -232,8 +232,8 @@ to next source, or previous source if STEPS is negative."
       (let ((limit (source-length (sources prompter))))
         (declare (type unsigned-byte limit))
         (unless (= 0 limit)
-          (let* ((previous-sources (previous-sources (first (selection prompter))))
-                 (index (+ (second (selection prompter))
+          (let* ((previous-sources (previous-sources (first (current-suggestion prompter))))
+                 (index (+ (second (current-suggestion prompter))
                            (source-length previous-sources)))
                  (new-index (+ index steps)))
             (setf new-index
@@ -244,34 +244,31 @@ to next source, or previous source if STEPS is negative."
                    (relative-index (- new-index
                                       (source-length (previous-sources new-source)))))
               (declare (type unsigned-byte relative-index))
-              (setf (selection prompter)
+              (setf (current-suggestion prompter)
                     (list new-source relative-index)))))))))
 
-(export-always 'select-next)
-(defun select-next (prompter &optional (steps 1))
-  "Select element by jumping STEPS forward.
-If STEPS is 0, do nothing.
-If STEPS is negative, go backward."
-  (select prompter steps))
+(export-always 'next-suggestion)
+(defun next-suggestion (prompter &optional (steps 1))
+  "See `set-current-suggestion'."
+  (set-current-suggestion prompter steps))
 
-(export-always 'select-previous)
-(defun select-previous (prompter &optional (steps 1))
-  "Select element by jumping STEPS forward.
-If STEPS is 0, do nothing.
-If STEPS is negative, go forward."
-  (select prompter (- steps)))
+(export-always 'previous-suggestion)
+(defun previous-suggestion (prompter &optional (steps 1))
+  "See `set-current-suggestion'."
+  (set-current-suggestion prompter (- steps)))
 
 (defun empty-source-p (source)
   (not (suggestions source)))
 
-(export-always 'select-next-source)
-(defun select-next-source (prompter &optional (steps 1))
-  "Jumping STEPS non-empty source forward and select first suggestion.
+(export-always 'next-source)
+(defun next-source (prompter &optional (steps 1))
+  "Set `current-suggestion' after jumping STEPS non-empty sources.
 If STEPS is 0, do nothing.
-If STEPS is negative, go backward and select last suggestion."
+When STEPS is positive (resp. negative), set the current suggestion to be the
+first (resp. last) one of destination source."
   (unless (= 0 steps)
     (sera:and-let* ((nonempty-sources (remove-if #'empty-source-p (sources prompter)))
-                    (source-index (or (position (selected-source prompter)
+                    (source-index (or (position (current-source prompter)
                                                 nonempty-sources)
                                       0))
                     (new-source (nth (alex:clamp (+ steps source-index) 0 (1- (length (sources prompter))))
@@ -279,49 +276,45 @@ If STEPS is negative, go backward and select last suggestion."
                     (suggestion-index (if (< 0 steps)
                                           0
                                           (1- (length (suggestions new-source))))))
-      (setf (selection prompter)
+      (setf (current-suggestion prompter)
             (list new-source suggestion-index)))))
 
-(export-always 'select-previous-source)
-(defun select-previous-source (prompter &optional (steps 1))
-  "Jumping STEPS source backward and select last suggestion.
-If STEPS is 0, do nothing.
-If STEPS is negative, go forward and selection first suggestion."
-  (unless (= 0 steps)
-    (select-next-source prompter (- steps))))
+(export-always 'previous-source)
+(defun previous-source (prompter &optional (steps 1))
+  (next-source prompter (- steps)))
 
 (defun nonempty-source-p (source)
   (suggestions source))
 
-(export-always 'select-first)
-(defun select-first (prompter)
-  "Select first element.
+(export-always 'first-suggestion)
+(defun first-suggestion (prompter)
+  "Set `current-suggestion' to PROMPTER's first suggestion.
 Empty sources are skipped."
   (let ((first-non-empty-source
           (or (find-if #'nonempty-source-p (sources prompter))
               (first (sources prompter)))))
-    (setf (selection prompter)
+    (setf (current-suggestion prompter)
           (list first-non-empty-source 0))
     (when (and (auto-return-p prompter)
                (sera:single (all-suggestions prompter)))
-      (return-selection prompter))))
+      (run-action-on-return prompter))))
 
-(export-always 'select-last)
-(defun select-last (prompter)
-  "Select last element."
+(export-always 'last-suggestion)
+(defun last-suggestion (prompter)
+  "Set `current-suggestion' to PROMPTER's last suggestion."
   (let ((last-non-empty-source
           (or (find #'nonempty-source-p (sources prompter)
                     :from-end t)
               (first (last (sources prompter))))))
-    (setf (selection prompter)
+    (setf (current-suggestion prompter)
           (list last-non-empty-source
                 (1- (length (suggestions last-non-empty-source)))))))
 
 (export-always 'toggle-mark)
 (defun toggle-mark (prompter)
-  (when (multi-selection-p (selected-source prompter))
+  (when (enable-marks-p (current-source prompter))
     (multiple-value-bind (suggestion source)
-        (selected-suggestion prompter)
+        (%current-suggestion prompter)
       (with-accessors ((marks marks)) source
         (let ((value (value suggestion)))
           (if (find value marks)
@@ -330,15 +323,15 @@ Empty sources are skipped."
 
 (export-always 'mark-all)
 (defun mark-all (prompter)
-  (let ((source (selected-source prompter)))
-    (when (multi-selection-p source)
+  (let ((source (current-source prompter)))
+    (when (enable-marks-p source)
       (alex:unionf (marks source)
                    (mapcar #'value (suggestions source))))))
 
 (export-always 'unmark-all)
 (defun unmark-all (prompter)
-  (let ((source (selected-source prompter)))
-    (when (multi-selection-p source)
+  (let ((source (current-source prompter)))
+    (when (enable-marks-p source)
       (with-accessors ((marks marks)
                        (suggestions suggestions))
           source
@@ -348,8 +341,8 @@ Empty sources are skipped."
 
 (export-always 'toggle-mark-all)
 (defun toggle-mark-all (prompter)
-  (let ((source (selected-source prompter)))
-    (when (multi-selection-p source)
+  (let ((source (current-source prompter)))
+    (when (enable-marks-p source)
       (with-accessors ((suggestions suggestions)
                        (marks marks))
           source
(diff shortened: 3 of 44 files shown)


libraries/prompter/prompter-source.lisp
```diff
@@ -288,27 +288,27 @@ to compute asynchronously.")
 the prompter is resumed.
 See `resume-sources'.")
 
-   (selection-actions
+   (suggestion-actions
```
aartaka, line 291: Well, that's why I was suggesting `cursor` or `focus` instead of `current-suggestion`—`suggestion-actions` seems too vague to immediately infer its meaning :(

Not that `cursor` is much better, but at least it hints at the fact that we choose one thing from many, while `suggestion-actions` may refer to many suggestions, if we only judge those by name...

aadcg: You have a point.  Notice that actually I'm going against what I proposed in the issue. I proposed `current-suggestion-actions`, not `suggestion-actions`.

Would `focus` and `focus-actions` sound better? Either that or `current-suggestion`/`current-suggestion-actions`...

aartaka: `focus` is good, as it has at least some associations for almost anyone.

Or yes, staying consistent and using `current-suggestion-actions`.

I'd personally prefer `focus`, but, due to this PR, I want to dig into similar interfaces and their terminology to find the most accurate set of terms for what we have here :D

aadcg:
I can only change the status of this PR by the time I finish a draft on the prompt buffer's manual section.

libraries/prompter/prompter-source.lisp
aadcg: Open to suggestions. As of today, I'm still on the side of `current-suggestion`.

aadcg:
I've added a draft of the prompt buffer manual section.  It's in org, since I can't imagine myself editing big chunks of text directly with `spinneret`.

Too early to take a look.

aadcg:
First acceptable draft concluded.

aadcg:
@aartaka has convinced me to keep the concept of marks, instead of renaming it to selection as I've proposed.

I remind that we can only be sure if the concepts make sense globally when the judge how easy it is to write a manual section on the topic of prompt buffer.

@aartaka take a look a first look the draft org file.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+enable-marks-p slot is non-nil.  For instance, the delete-buffer command deletes
+all selected buffers at once.
+
+Every collection of suggestions belongs to a source.  Sources define consistent
+umbrellas for specific use cases.  For instance, the command describe-any
+features separate sources for variables, functions and classes.
+# and other things I want to omit.
+
+# I just realized how hard it is to describe how users can understand which
+# source is the current one.  Shouldn't this UI be re-thought?  Thomas?
```
aartaka, line 25: The current source header is always at the top, just below the input area. Additionally, the header above the current suggestion highlighting is always the current source header. Is that not enough?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+select-previous-page
+select-next-page
+select-previous-source
+select-next-source
+# rename all of these; select is both confusing and inaccurate
+# previous-suggestion
+# next-suggestion
+# first-suggestion
+# last-suggestion
+# previous-suggestion-page
```
aartaka, line 52: Maybe just `(next|previous)-page`? We don't have any concept of suggestion pages, after all...

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+In some cases, the argument you will submit is a list of suggestions, i.e. the
+marks.  The following commands add suggestions to the marks:
+
+mark-all
+unmark-all
+toggle-mark
+toggle-mark-backwards
+toggle-mark-all
+# rename these commands to:
+# mark-all-suggestions
```
aartaka, line 70: I mean, it should be obvious that if we're in prompt buffer, then we're marking suggestions and not somethign else. So I'd stay with at least `mark-all` and `unmark-all`. Other commands would benefit from some descriptiveness, though.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+# mark-complementary-suggestions
+
+You can exit the prompt buffer at any given moment with command cancel-input.
+
+# rename cancel-input to exit-prompt-buffer.
+
+** Notable sources
+Should we refer to some of them?
+
+buffer-source
```
aartaka, line 83: My suggestions: `prompter:raw-source`, `bookmark-source`, and an `if-confirm` as a special macro.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+hint-source
+file-source
+
+** Attributes
+Attributes describe in detail the suggestions of each source.  These are shown
+in a columnar way.
+
+# mention toggle-attributes-display
+
+# is there a global way of disable showing all attributes?
```
aartaka, line 95: No. Why?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+
+** Actions
+# maybe rename actions to
+# action-on-return
```
aartaka, line 99: I actually thought of suggesting this, because the `on-*` language automatically implies that it's something dependent and event-ish already. Like `onclick` in HTML.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+# action-on-marks
+# action-on-current-suggestion
+Actions are functions that take a single suggestion or marks as arguments.
+
+There are three kinds of them.
+
+Return actions define what happens when exiting the prompt buffer.  For
```
aartaka, line 106: "exiting the prompt buffer... normally"

By the way, should we add `cancel-actions`, or are those usually managed by the calling code? Should it be this way?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+instance, in the case of buffers, these could be deleting or bringing it to the
+foreground.
+
+Marks actions run everytime the marks change.
+# hard to find a simple example... the example on hint-mode is too complex.
```
aartaka, line 111: No ideas :D

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+# ideas?
+
+Current suggestion actions run everytime the current suggestion change.  For
```
aartaka, line 114: This paragraph consists of two absolutely unrelated sentences. More so: it's inconsistent with the previous ones, which follow the "explanation -> example" structure. 

Maybe extract the part about highlighted suggestion to the part when you explain what current suggestion is, and add a good example, like hint/search highlightling, instead of it?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+instance, while searching on a buffer the search match corresponding to the
+current suggestion is highlighted in a distinctive way.
+
+** Passwords
```
aartaka, line 118: I don't like the name of the section, but I don't like the non-descriptive and unintuitive name of the slot either. So meybe we can find some other name? StumpWM uses `:password t`, which is overly narrow, I believe.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+Nyxt.  When invisible-input-p is non-nil, the input is replaced by a placeholder
+character to conceal it.
+
+** History
+Oftentimes you may want to re-introduce something you had already typed in the
+input area.  Everytime the prompt buffer returns successfully, that information
+is saved for the future.  To access any of those saved inputs, call the command
+nyxt/prompt-buffer-mode:history.
+
+# rename to prompt-input-history
```
aartaka, line 129: No, it's not necessarily an input history! Specific sources can re-define it to be more useful and serve something beyond inputs!

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,157 @@
...
+thus far.
+
+Contrary to what you may expect, there is no global way to set the way filtering
+is done.  Instead, this is an option within the context of each source.  In
+other words, the source (i.e. the specific context of a given collection of
+suggestions) determines which kind of filtering suits best.
+
+Generally speaking, prompter:filter is set to fuzzy-match.
+
+# Is giving an overview of fuzzy-match out of scope?  I think so.
```
aartaka, line 147: It is in scope, because it showcases that prompt-buffer is not a mere prefix search dumbness (like the one StumpWM has ヽ(ｏ`皿′ｏ)ﾉ), but something smarter!

aartaka (commented):
Only reviewed the .org file :)

source/manual-prompt-buffer.org
aadcg: I'm referring to a different issue. Our source glyph is a sun (or whatever that is) and it's not clear that headings introduced by that glyph are source names. Also, below that the user finds attributes names. 

Input area, source header, attributes header... It's a lot to chew in one go. I think Thomas could help us here.

aadcg: See #2687.

source/manual-prompt-buffer.org
aadcg: That is true, but the concept of `page` is also not clear. The concept of `pagedown` however, is commonplace!

What about `pagedown-suggestions` and `pageup-suggestions`?

source/manual-prompt-buffer.org
aadcg: By the same token, you could have argued that `previous` or `next` are descriptive enough, since it should be obvious that in the prompt buffer we're dealing with suggestions.

Being clear doesn't hurt us here. These commands are very seldom called by name interactively, so its length should not worry us. Programmatically, it does help a lot. Don't forget that `mark` is used in a radically way in `visual-mode` (notice that commands such as `toggle-mark` exist there).

source/manual-prompt-buffer.org
aartaka: True, page is not clear. But then, `pagedown` is not helping much either. `next-screen`? `next-bunch`? `next-screenful-of-suggestions`? At some point we have to sacrifice total intelligibility in favor of some more or less understandable concept.

source/manual-prompt-buffer.org
aartaka: If the command names are the same, but they are all isolated in their own special modes, that's not much of a problem, because it's somewhat context-specific.

source/manual-prompt-buffer.org
aadcg: Wait. We're writing a user manual. Your suggestions seem better suited at a reference manual that details Nyxt's internals for developers.

With that in mind, the source examples we should highlight are the specific and "advanced" ones. I.e. the ones that the user has no way of guessing that they do more than what they would except. `file-source` seemed like a good candidate.

What do you think?

source/manual-prompt-buffer.org
aadcg: We have a toggler that disables some or all of the attributes. There should be an option to start the prompt buffer without any attributes. The default would be to show all attributes (as we do now).

source/manual-prompt-buffer.org
aadcg: Regarding the section name, we're on the same page. 

Regarding the slot, that's a fair point. A suggestion that comes to mind is `conceal-input-p`. It's better but not much better perhaps.

source/manual-prompt-buffer.org
aadcg: Ok, that's interesting. Could you provide an example?

source/manual-prompt-buffer.org
aadcg: Honestly I don't know why I wrote this, since I agree with you (and thus disagree with what I wrote).

source/manual-prompt-buffer.org
aadcg: I don't follow.

The example illustrate `current-suggestion-actions`. While searching a buffer, everytime the current suggestion changes, the action runs (in this case, it highlights the search match).

Oh. I see your confusion. You think I'm referring to highlighting in the prompt buffer! Is that it?

source/manual-prompt-buffer.org
aadcg: Also, `marks-actions` is terrible (as noticed in #2527).

source/manual-prompt-buffer.org
aadcg: @Ambrevar?

source/manual-prompt-buffer.org
aadcg: Why does not `pagedown-suggestions` help? Everyone associates that concept with what you suggest `next-screen`, `next-bunch`, etc.

aadcg:
Generally speaking me and @aartaka have agreed on most topics.

To move forward, I need input from @jmercouris and @Ambrevar. 

Let me summarize all of it in a table.

The last table adds a third column. If the row is non-empty, it means @aartaka has a different suggestion. 

I'm not happy about `mark-current-suggestion-forward/backwards`. Oddly, @aartaka didn't raise the topic.

```log
| old concept       | new concept                  |
|-------------------+------------------------------|
| return-action     | action-on-return             |
| marks-actions     | action-on-marks              |
| selection-actions | action-on-current-suggestion |
| selection         | current-suggestion           |



| old slot               | new slot                          |
|------------------------+-----------------------------------|
| multi-selection-p      | enable-marks-p                    |



| old command            | new command (aadcg)               | new command (aartaka) |
|------------------------+-----------------------------------+-----------------------|
| select-previous        | previous-suggestion               |                       |
| select-next            | next-suggestion                   |                       |
| select-first           | first-suggestion                  |                       |
| select-last            | last-suggestion                   |                       |
| select-previous-page   | pagedown-suggestions              | next-page             |
| select-next-page       | pageup-suggestions                | previous-page         |
| select-previous-source | previous-source                   |                       |
| select-next-source     | next-source                       |                       |
|------------------------+-----------------------------------+-----------------------|
| mark-all               | mark-all-suggestions              | mark-all              |
| unmark-all             | unmark-all-suggestions            | unmark-all            |
| toggle-mark            | mark-current-suggestion-forward   |                       |
| toggle-mark-backwards  | mark-current-suggestion-backwards |                       |
| toggle-mark-all        | mark-complementary-suggestions    |                       |
|------------------------+-----------------------------------+-----------------------|
| cancel-input           | exit-prompt-buffer                |                       |
|------------------------+-----------------------------------+-----------------------|
| return-selection       | run-return-action-over-marks      |                       |
| return-marks-action    | set-return-action                 |                       |
```

source/manual-prompt-buffer.org
aadcg: I'm afraid I'm confused. Do you mean that returns actions run when the prompt buffer exists successfully? I need to find a way to communicate that in a non-scary and user-friendly way.

aadcg: > By the way, should we add cancel-actions, or are those usually managed by the calling code? Should it be this way?

Not sure... Perhaps let's think about an example?

aadcg:
Well, I'll take the silence as approval and continue the work :p

Ambrevar:
Sorry for the delay, I'll come back to this as soon as possible.

aadcg:
The renaming was done as per the table below.

I now need to go back to the manual to see if everything makes sense.

```log
| old concept                                  | new concept                   |
|----------------------------------------------+-------------------------------|
| return-action                                | actions-on-return             |
| marks-actions                                | actions-on-marks              |
| selection-actions/current-suggestion-actions | actions-on-current-suggestion |
| selection                                    | current-suggestion            |

| old slot               | new slot                          |
|------------------------+-----------------------------------|
| multi-selection-p      | enable-marks-p                    |

| old command             | new command                     |
|-------------------------+---------------------------------|
| select-previous         | previous-suggestion             |
| select-next             | next-suggestion                 |
| select-first            | first-suggestion                |
| select-last             | last-suggestion                 |
| select-previous-page    | next-page                       |
| select-next-page        | previous-page                   |
| select-previous-source  | previous-source                 |
| select-next-source      | next-source                     |
|-------------------------+---------------------------------|
| toggle-mark             | mark-forward                    |
| toggle-mark-backwards   | mark-backwards                  |
| toggle-mark-all         | mark-complement                 |
|-------------------------+---------------------------------|
| cancel-input            | exit-prompt-buffer              |
|-------------------------+---------------------------------|
| return-selection        | run-action-on-return-over-marks |
| return-marks-action     | set-action-on-return            |
```

aadcg:
I've reviewed the renaming. It seems to make sense conceptually, and it passed all of my inspection routines to ensure that I'm not introducing bugs.

If it looks good then I will merge, but only after:

- extracting the `.org` manual file
- rebase into master
- squash commits

Waiting for review. @jmercouris @aartaka @Ambrevar

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+
+** Basic Usage
+When invoked, the prompt buffer occupies the lower region of the screen, and the
+cursor is set at the top bar, which features the input area at the center, the
+active modes on the right and the prompter message on the left.  The suggestions
+are listed vertically below the top bar and are displayed before you start
+typing.
+
+As you change the content of the input area, the list of suggestions is updated.
+The commands defined by input-edit-mode are available.  The current suggestion
```
aartaka, line 27: I'd not refer to `input-edit-mode` without explaining that its commands are for text editing. So maybe "The commands useful for text editing (provided by `input-edit-mode`) are provided"?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+In some cases, the argument you will submit are marks.  The following commands
+add suggestions to the marks:
+
+mark-all
+unmark-all
+mark-forward
+mark-backwards
+mark-complement
+
+To submit a suggestion or the marks, you can run the command
```
aartaka, line 49: This sentence feels quite artificial and formalistic. No clue what to do with this, though.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+run-action-on-return-over-marks, bound to RET by default.
```
aartaka, line 50: Here a good heuristic might be: tell the douwn-to-the-earth action ("press RET") first, and then explain the guts of it ("bound to the `run-action-on-return-over-marks` by default").

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+
+In some specific situations, several actions-on-return are available and you can
+choose which one will run with command set-action-on-return.
```
aartaka, line 53: Keybinding? Note that it works with only one action too, and this sentence does not clearly mention it.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+
+At any given moment, you can exit the prompt buffer with command
+exit-prompt-buffer.
+
+** Attributes
+Attributes describe in detail the suggestions of each source.  These are shown
+in a columnar way.
+
+The command toggle-attributes-display
```
aartaka, line 62: Seems dropped? And what about `prompter:object-attributes`?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+
+** Actions
+Actions are functions that take a single suggestion or marks as arguments.
```
aartaka, line 65: "that take a list of either a single suggestion or >1 marks as a sole argument"?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+
+There are three kinds of them.
```
aartaka, line 67: "them" is an anaphoric pronoun, which might be confusing, especially in a paragraph of its own. "three kinds of actions", maybe?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+
+actions-on-return define what happens when submitting arguments to the prompt
+buffer.  For instance, in the case of buffers, these could be deleting or
+bringing it to the foreground.
```
aartaka, line 71: "switching to it", maybe? "bringing it to the foreground" implies a visual/conceptual layering, which is not the case with either the implementation nor the concepts we use outside this manual entry.

source/manual-prompt-buffer.org
aartaka: An example of redefining the `prompter:object-attribtues` for a known source won't hurt too.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+
+actions-on-current-suggestion run everytime the current suggestion change.  For
```
aartaka, line 73: Typo: "changes" (third person, present tense, "suggestion" as the actor)?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+instance, while searching on a buffer the search match corresponding to the
```
aartaka, line 74: "search hint" is the term we use for search highlights right now. See `remove-search-hints` command.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+current suggestion is highlighted in a distinctive way.
+
+actions-on-marks run everytime the marks change.
+
+** Concealing information
+You may need to enter a password or other sensitive information in Nyxt.  When
+invisible-input-p is non-nil, the input is replaced by a placeholder character
```
aartaka, line 81: Maybe mention that it's a keyword argument to `prompt(1)`?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+to conceal it.
+
+** History
+You may want to re-introduce something you had already typed in the input area.
+Everytime the prompt buffer returns successfully (i.e. unless you have run
+exit-prompt-buffer), that information is saved for the future.  To access any of
+those saved inputs, call the command nyxt/prompt-buffer-mode:history.
```
aartaka, line 88: Keybinding?

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+
```
aartaka, line 89: Related to history: mention prompt resuming? `resume-prompt` etc.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+** Filtering
+# The concept used by prompt buffer is filter, since it filters all suggestions
+# against user input.  Completion is more common concept, but different: it
+# entails binding certain keys to update the list of suggestions (usually TAB).
+
+Filtering is a mechanism that helps you find suggestions as you type.  Nyxt is
+able to guess what you meant based on what you have typed thus far.
+
+Contrary to what you may expect, there is no global way to set the way filtering
```
aartaka, line 98: I'd leave out this "Contrary to what you may expect", as it's better to not guess one's expectations but to simply write what's there :)

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
...
+scroll-other-buffer-up
+scroll-page-down-other-buffer
+scroll-page-up-other-buffer
+
+** Notable sources
+As mentioned before, sources are designed with specific use cases in mind.
+Besides setting the suggestions accordingly, actions and other parameters are
+specialized for the case at hand.
+
+Inspect sources with the command describe-class.  buffer-source, file-source,
```
aartaka, line 128: You'll add `:nxref`s everywhere necessary when you turn that into a manual entry, right?

aadcg:
I didn't explain myself very well so let me make it clearer. I drafted the manual together with the renaming to ensure that the renaming actually makes sense. Imagine if I had made the renaming first only to realize later, while writing the manual section, that I needed to re-do some renaming.

When I mentioned the bit below, I was expecting a review of the renaming itself.

> If it looks good then I will merge, but only after:

>   - extracting the .org manual file
>  - rebase into master
>  - squash commits

My plan is to get feedback on the renaming, and then rebase, delete `.org` file, squash and merge. Then, in another PR, we will continue to shape the manual section.

Since @aartaka reviewed this incomplete draft of the manual (sorry for the incomplete sections, I wasn't expecting a review there), I will answer all of his comments in this PR but the changes will be reflected in the PR I will open later. Is that ok @aartaka?

source/manual-prompt-buffer.org
aadcg: Yes. When I open the other PR we will work in `org` for a while. I want to move to Nyxt's manual format when we agree on all of the text.

aartaka:
Ohhh, sorry about that `(^_^;)` Yeah, the renaming is fine by me!

aadcg:
Your valuable comments will not be forgotten! 

I'd like to wait for @Ambrevar to have a final word about the renaming.

Also, do the changes in the changelog look good?

source/changelog.lisp
```diff
@@ -31,7 +31,7 @@ particular revision.")
    (:li "Search engine autocompletion. See the search engine suggestions in
         the prompt buffer as you type. By default, support for DuckDuckGo and
         Wikipedia is included.")
-   (:li (:code "set-url") " and related commands now allow for multiple selection.")
+   (:li (:code "set-url") " and related commands now allow marks.")
```
aartaka, line 34: I'd rather leave the hystoric parts of changelog intact, as this was the term at the time ir was written. Otherwise we risk all the risks rewriting history entails :P

source/changelog.lisp
```diff
@@ -191,7 +191,7 @@ particular revision.")
    (:li (:code "list-buffers") " can now display the the buffers as a tree."))
   (:h3 "Platform support")
   (:ul
-   (:li "Fix slow buffer and prompt-buffer creation on FreeBSD.  (Thanks to
+   (:li "Fix slow buffer and prompt buffer creation on FreeBSD.  (Thanks to
```
aartaka, line 194: Either is fine, not sure which side I'm on.

source/changelog.lisp
```diff
@@ -496,11 +496,57 @@ to open a file, save it, switch buffer or delete current buffer.")
...
         (:code "M-y") " in Emacs scheme of "
         (:nxref :class-name 'nyxt/document-mode:document-mode) ".")
-   (:li "Prompt-buffer now has familiar bindings for text cutting.")
-   (:li "Add " (:nxref :command 'nyxt/prompt-buffer-mode:set-selection-action)
+   (:li "Prompt buffer now has familiar bindings for text cutting.")
+   (:li "Add " (:nxref :command 'nyxt/prompt-buffer-mode:set-action-on-current-suggestion)
         ", bound to " (:code "C-c C-j") "by default.")
-   (:li (:code "return-selection-over-action") " renamed to "
-        (:nxref :command 'nyxt/prompt-buffer-mode:return-marks-action)
+   (:li (:code "return-selection") " renamed to "
```
aartaka, line 502: But why not put it into 3.0.0 section below? It's not 3-pre-release-2 change, after all...

source/changelog.lisp
aadcg: Ok, let's do it that way.

source/changelog.lisp
aadcg: I have settled on non hyphenation in prose (docstrings, manual, etc). I had a note about it in a previous version of the `.org` file.

source/changelog.lisp
aadcg: Makes sense.

aadcg: See 2609489aa.

source/changelog.lisp
aartaka: Got it :)

aadcg:
@Ambrevar could I ask you for your feedback?

Please read both of my comments above:

- [comment 1](https://github.com/atlas-engineer/nyxt/pull/2655#issuecomment-1354860058)
- [comment 2](https://github.com/atlas-engineer/nyxt/pull/2655#issuecomment-1357755502)

source/mode/prompt-buffer.lisp
```diff
@@ -330,77 +330,79 @@ current unmarked selection."
...
-  "Mark selection.
-Only available if current PROMPT-BUFFER source `multi-selection-p' is non-nil.
+(define-command-prompt toggle-actions-on-current-suggestion-enabled (prompt-buffer)
+  "Toggle whether `prompter:actions-on-current-suggestion' are enabled for PROMPT-BUFFER."
+  (prompter:toggle-actions-on-current-suggestion-enabled prompt-buffer)
+  (echo "Current suggestion actions: ~:[dis~;en~]abled."
+        (prompter:actions-on-current-suggestion-enabled-p (current-source prompt-buffer))))
+
+;; TODO The direction logic must be extracted to a function, since this command
+;; exposes too many implementation details to the user.
```
Ambrevar, line 373: I don't understand, what is exposed?

Ambrevar:
Thank you @aadcg for this amazing work, and very sorry for taking so long to get back to it! :/ 

Here are the few names I would like to work on:

- `mark-complement`: "complement" is not used much in our code base.  I like "toggle" better.
- `exit-prompt-buffer`: we actually use "quit" more often.
- `mark-forwards` and `mark-backwards`: the previous name highlighted that the command is a toggler, which is kind of cool.  If you didn't like the asymmetry, what about `toggle-mark-forwards` and `toggle-mark-backwards`?  Obviously your approach has the benefit of being more concise.  So no strong opinion here.
- `run-action-on-return-over-marks `: Why not just `run-action-on-return`?

libraries/prompter/README.org
```diff
@@ -45,8 +45,8 @@ Non-exhaustive list of features:
 - Notifications sent when suggestion list is updated.
 - Per-source history.
 - Resumable prompters.
-- Selection actions (automatically run persistent actions on selection change).
-- Marks actions (automatically run persistent actions on marks change).
+- Marks actions (event-driven on marks change).
+- Current suggestion actions (event-drive on current suggestion change).
```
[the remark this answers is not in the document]
aadcg: Thanks!

source/mode/prompt-buffer.lisp
aadcg: I'll rephrase.

aadcg:
I agree with all of your comments and made the changes accordingly.

Now, I need to review this huge change from top to bottom with care!

aadcg:
I've reviewed it and it looks good to me.  Would you like to have a final look @aartaka?

aartaka (commented):
All in all, my comments are rather requests for future work. One significant thing: get `return` back. Rationale:
- We have `actions-on-return`, but we have no `return` that these happen `on`. Which is quite a gap in our prompter concepts.
- `run-action-on-on-return` is concerned with much more than _just running the actions_. Which is a bad sign. There should be some other infrastructural function to do channel and history management, like `return`. Action runner should not be concerned with anything beyond actions.
- `return` is just so good of a name and API point in general, while `run-actions-on-return` feels like quite a specialized thing. No amount of documentation can fix the confusing API structure, so we better have a single point of exit that's clearly named—`return`.

On the subject on naming, you seem to have renamed `cancel-input` to `quit-prompt-buffer`, but left the `prompt-buffer-canceled` error with its old name. Why so? Maybe rename it too? What I'd personally do is:
- Retain the meaningful `cancel` in the name of the command, so it's something like `cancel-prompt-buffer`. `cancel-input` is not perfect, but it conveys a clear message of bailing out, instead of awfully neutral `quit-*`.
- Don't rename the condition, leave it as is. It's clear enough.

libraries/prompter/README.org
```diff
@@ -8,28 +8,28 @@ A =prompt= is an interface for user interactions that holds one or more
...
-     selected =suggestion=, belonging to a =source=.
+   + =current-suggestion= :: A single =suggestion=, if any.
 - =source=
-   + =marks= :: A list of =suggestions=, when =multi-selection-p= is non-nil.
+   + =marks= :: A list of =suggestions=, when =enable-marks-p= is non-nil.
    + =actions= :: A list of functions that run on =suggestions=.
-      - =return-actions= :: General purpose.
-      - =marks-actions= :: On =marks= change (event-driven).
-      - =selection-actions= :: On =selection= change (event-driven).
+      - =actions-on-return= :: General purpose.
```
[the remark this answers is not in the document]
aadcg: This is the prompter library, so it doesn't know what the Return key is. The return in `actions-on-returns` refers to returning (the output of a computation) the prompt.

`actions-on-marks` and `actions-on-current-suggestion` are triggered when that data changes, whereas `actions-on-return` needs to be explicitly called. 

I'll extend the explanation on `actions-on-return`.

libraries/prompter/README.org
```diff
@@ -8,28 +8,28 @@ A =prompt= is an interface for user interactions that holds one or more
...
 
-A =prompt= always has a single =selection=, whereas a =source= has either a
-single =selection= or none (when =prompt= has multiple =sources=).
+When =prompt= has multiple =sources=, while =current-suggestion= is always
+defined for =prompt=, it is empty for all but one of its =sources=.
 
-=marks= is a concept related to =source= not =prompt=, unlike that of
-=selection=.
+=Marks= is a concept related to =source= not =prompt=, unlike that of
+=current-suggestion=.
```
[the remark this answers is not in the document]
aadcg: Sure.

libraries/prompter/prompter.lisp
```diff
@@ -159,32 +157,33 @@ computation is not finished.")))
...
-                  (_ (selection-actions-enabled-p source))
-                  (action (default-selection-action source))
-                  (suggestion (selected-suggestion prompter)))
-    (let ((delay (selection-actions-delay source)))
+(defmethod (setf current-suggestion) (value (prompter prompter))
+  (setf (slot-value prompter 'current-suggestion) value)
+  (run-action-on-current-suggestion prompter))
+
+(export-always 'run-action-on-current-suggestion)
+(defmethod run-action-on-current-suggestion ((prompter prompter))
```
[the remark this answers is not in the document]
aadcg: It's always a single one. See `default-action-on-current-suggestion`.

libraries/prompter/prompter.lisp
```diff
@@ -244,84 +244,75 @@ to next source, or previous source if STEPS is negative."
...
                     (list new-source relative-index)))))))))
 
-(export-always 'select-next)
-(defun select-next (prompter &optional (steps 1))
-  "Select element by jumping STEPS forward.
-If STEPS is 0, do nothing.
-If STEPS is negative, go backward."
-  (select prompter steps))
+(export-always 'next-suggestion)
+(defun next-suggestion (prompter &optional (steps 1))
```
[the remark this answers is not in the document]
aadcg: Ok, let's  do that.

libraries/prompter/prompter.lisp
```diff
@@ -244,84 +244,75 @@ to next source, or previous source if STEPS is negative."
...
 
 (defun nonempty-source-p (source)
   (suggestions source))
 
-(export-always 'select-first)
-(defun select-first (prompter)
-  "Select first element.
+(export-always 'first-suggestion)
+(defun first-suggestion (prompter)
+  "Set `current-suggestion' to the first suggestion.
```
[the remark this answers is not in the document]
aadcg: Good point.

libraries/prompter/prompter.lisp
```diff
@@ -403,24 +394,26 @@ If input is already in history, move to first position."
...
                      (input prompter))))
 
-(export-always 'return-selection)
-(defun return-selection (prompter
-                         &optional (return-action (default-return-action prompter)))
-  "Call RETURN-ACTION over `marks' and send the results to PROMPTER's `result-channel'.
+(export-always 'run-action-on-return)
+(defun run-action-on-return (prompter &optional (action-on-return
+                                                 (default-action-on-return prompter)))
+  "Call ACTION-ON-RETURN over `marks' and send the results to PROMPTER's
```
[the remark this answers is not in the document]
aadcg: There's a line below that addresses that concern.

libraries/prompter/prompter.lisp
```diff
@@ -403,24 +394,26 @@ If input is already in history, move to first position."
...
 
-(export-always 'toggle-selection-actions-enabled)
-(defun toggle-selection-actions-enabled (prompter
-                                         &optional (source (selected-source prompter)))
-  "Toggle `selection-actions-enabled-p' in SOURCE."
-  (setf (selection-actions-enabled-p source) (not (selection-actions-enabled-p source))))
+(export-always 'toggle-actions-on-current-suggestion-enabled)
+(defun toggle-actions-on-current-suggestion-enabled
+    (prompter &optional (source (current-source prompter)))
+  "Toggle `actions-on-current-suggestion-enabled-p' in SOURCE."
```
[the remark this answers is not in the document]
aadcg: I don't see much benefit in that.

source/changelog.lisp
```diff
@@ -572,6 +572,59 @@ auto-mode-rules.lisp)."))
    (:li "Startup is more robust against corrupted history files.")
    (:li (:code "diff-mode") " is removed."))
 
+  (:h3 "Bindings")
```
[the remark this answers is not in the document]
aadcg: Makes sense!

aadcg:
Thanks!

@aartaka I suggest opening issues regarding the architecture of the library.

I'll soon continue to work on the prompt buffer's manual section, while not forgetting the review that @aartaka did on that.

aadcg:
> On the subject on naming, you seem to have renamed cancel-input to quit-prompt-buffer, but left the prompt-buffer-canceled error with its old name. Why so? Maybe rename it too? What I'd personally do is:

> -  Retain the meaningful cancel in the name of the command, so it's something like cancel-prompt-buffer. cancel-input is not perfect, but it conveys a clear message of bailing out, instead of awfully neutral quit-*.
> -   Don't rename the condition, leave it as is. It's clear enough.

This was discussed above when @ambrevar mentioned:

> exit-prompt-buffer: we actually use "quit" more often.

I don't see much a problem in the fact that we have a command called `quit-prompt-buffer` (that is user-facing) and a condition called `prompt-buffer-canceled` (that is developer-facing).

Ambrevar:
@aadcg @aartaka Was an issue opened about the `return` issue?

Note that I was planning (Soon Enough©) to replace Calispel with Lparallel in the Prompter library, which should simplify a bunch of things, in particular remove lots of channels.

aartaka:
> > On the subject on naming, you seem to have renamed cancel-input to quit-prompt-buffer, but left the prompt-buffer-canceled error with its old name. Why so? Maybe rename it too? What I'd personally do is:
> 
> > * Retain the meaningful cancel in the name of the command, so it's something like cancel-prompt-buffer. cancel-input is not perfect, but it conveys a clear message of bailing out, instead of awfully neutral quit-*.
> > * Don't rename the condition, leave it as is. It's clear enough.
> 
> This was discussed above when @Ambrevar mentioned:
> 
> > exit-prompt-buffer: we actually use "quit" more often.
> 
> I don't see much a problem in the fact that we have a command called `quit-prompt-buffer` (that is user-facing) and a condition called `prompt-buffer-canceled` (that is developer-facing).

This was discussed in the context of you renaming `cancel-input` to `exit-prompt-buffer`, but only as a matter of framing `exit-prompt-buffer` in a more conventional way. So it's a different layer of the arguement:
- Pierre was saying about the framing of the new name, not discussing the meaning of the new name itself (@Ambrevar, correct me if I'm wrong),
- I'm talking about the meaning of the new name, and that this meaning is slightly misleading.

To exemplify it with a more down-to-earth example:
- When I review a PR and leave a comment like "Here a `loop` would suit better than `do*`", this does not immediately mark PR as approved. It's just a stylistic comment.
- When I click "Approve" checkbox on GitHub or write something like "Looks good to me!", then I unambiguosly approve a given PR.

aartaka:
> @aadcg @aartaka Was an issue opened about the `return` issue?

Now there is: https://github.com/atlas-engineer/nyxt/issues/2753

> Note that I was planning (Soon Enough©) to replace Calispel with Lparallel in the Prompter library, which should simplify a bunch of things, in particular remove lots of channels.

Yay!

aadcg:
@aartaka, since [December 2](https://github.com/atlas-engineer/nyxt/pull/2655#issuecomment-1335886948) I have suggested and renamed `cancel-input` to `exit-prompt-buffer`. On that topic, the only feedback I had came from @Ambrevar that advocated for `quit-prompt-buffer`. I think there was plenty of time for everyone to chime in.

If there's any topic that needs to be further discussed, please open an issue.

aartaka:
> @aartaka, since [December 2](https://github.com/atlas-engineer/nyxt/pull/2655#issuecomment-1335886948) I have suggested and renamed `cancel-input` to `exit-prompt-buffer`. On that topic, the only feedback I had came from @Ambrevar that advocated for `quit-prompt-buffer`. I think there was plenty of time for everyone to chime in.

Yes, that I agree with—I was not attentive enough when reviewing this PR, and was unable to spot this naming thing earlier. Sorry about that. Still, I noteced it now, and, as they say, it's better late than never. 

I've been in a rush, so sorry if I sounded off. I didn't intend to downplay your work on this PR—it's huge and it made a lot of things much more consistent `٩(๑ơలơ)۶♡`

> If there's any topic that needs to be further discussed, please open an issue.

Yes, see #2755.

source/manual-prompt-buffer.org
```diff
@@ -0,0 +1,129 @@
+* Prompt buffer
+
+The prompt buffer is special buffer where Nyxt commands read arguments, such as
+URLs, file names or Common Lisp expressions.  We call it prompt buffer because
+it requests textual input from the user.
```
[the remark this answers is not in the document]
aadcg: Agree.

source/manual-prompt-buffer.org
aadcg: Agree.

source/manual-prompt-buffer.org
aadcg: Agree, but I'm also a bit clueless... Will massage it.

source/manual-prompt-buffer.org
aadcg: Agree.

source/manual-prompt-buffer.org
aadcg: I think "you can choose which one" indicates that you can choose one from (possibly) many.

source/manual-prompt-buffer.org
aadcg: > An example of redefining the prompter:object-attribtues for a known source won't hurt too.

I think that a user manual should mention variables (or slots) that exist for configuration. However, it is not the purpose of a user manual to explain and provide concrete example of how to achieve something more specific. That is the responsibility of either: a cookbook, a user manual section that explains the "programming interface" in a user manual, or a standalone "programming interface manual" when it becomes too bulky.

source/manual-prompt-buffer.org
aadcg: Agree.

source/manual-prompt-buffer.org
aadcg: Agree.

source/manual-prompt-buffer.org
aadcg: Agree.

source/manual-prompt-buffer.org
aadcg: Yes!

source/manual-prompt-buffer.org
aadcg: Ok!

source/manual-prompt-buffer.org
aadcg: Agree.

source/manual-prompt-buffer.org
aadcg: Yes, good idea.

source/manual-prompt-buffer.org
aadcg: Agree.

aadcg:
Thanks for the manual review @aartaka. I'll make a new version of it.  I'll probably open a PR with the org file so that we continue to go through it.  At the last minute, I'll integrate it in `manual.lisp`. Notice that by using `(:nxref :command 'foo-command)`, the keybinding will be mentioned so let's not worry about that while reviewing the `.org` file.

source/manual-prompt-buffer.org
aartaka: Yes.

source/manual-prompt-buffer.org
aartaka: Our manual—due to the ambiguous position of our users in between the passive users and active developers—is both a reference manual and user manual. Thus, we shold play both simplicity and depth (which is not always easy). Anyway, more links and code samples are always better than less.

source/manual-prompt-buffer.org
aartaka: We don't use it yet, but we can easily provide `:history`to any `prompt` call, providing a custom history.

PR #2657, opened 2022-11-21
More intuitive history heuristics

aartaka:
# Description

This adds two history improvements utilizing the `htree` and Nyxt features:
- Remembering the requests that were cancelled mid-air, and adding those to history as parents of the ones that replace them.
- Going back in history instead of creating a new child, if the URL opened is already present on the current branch. This way, we utilize the branching nature of the history and re-create the structure of the page in the history tree. The current implementation has downsides:
  - If the "hub" URL (i.e., the one that person returns to) is not the first one they visit, then some of the URLs visited from the hub might actually be recognized as parents of the hub. We need to intelligently sort the history for parent-child/index-page relations and swap the neighbouring nodes if we see them being thus related and misplaced. This all might've sounded unintelligeble, so here's an illustration of what the current logic does and how it might fail:
```
github.com/.../pull/1 ->  github.com/notifications -> github.com/.../pull/3 -> github.com/notifications
                                                                               ^ current node here

Gets optimized into this structure with this PR:

github.com/.../pull/1 ->  github.com/notifications -> github.com/.../pull/3
                          ^ current node here

But then, if one decides to visit github.com/.../pull/1 from the "hub" (github.com/notifications), 
then we end up with a mental mismatch of a model with history structure:

github.com/.../pull/1 ->  github.com/notifications -> github.com/.../pull/3
^ current node here

While it should've been
github.com/notifications -> github.com/.../pull/3
                       |-> github.com/.../pull/1
                           ^ current node here

Because the mental map of the user most probably has /notifications 
as the hub URL that all the journeys start from. github.com/.../pull/1 
is merely a coincidental URL that the browsing started from in one particular session.
```

# Discussion

I'll be happy if we resolve some more problems of our history interface (@jmercouris, note that I'm saying interface here, it's not about `htree` as a structure, it's about us not watching all the corner-cases and blindly putting things into `htree`) as part of this PR, but I am okay with posponing it until later:
- Better canceled link inspection. Simply putting them into history is not always the right decision (can't remember the cases that break it, but there certainly were some, because of which I didn't open this PR three months ago). We need better heuristics for when link canceling is meaningful, and when it's not.
- Properly handling redirection URLs. Right now if there'a redirect from `foo.org#hello` to `foo.org#hi`, we only remember the first one and ignore the second one based on it having the "same URL". There are two issues at play here:
  - Flawed (in some cases like this) URL equality operators, easy to fix.
  - Broken redirection storage, only storing the first URL that started redirect chain. I am not certain about the reasons for that. 
  - I guess that the right behavior for redirects would be to store only the final redirect chain URL in history.
    - Alternatively, we can store all the redirections and all the chain links in history, but mark them as such, so that when one goes `history-backwards` to a redirect URL, they get prompted or automatically intuitively moved to before it, to not trigger the inifinite cycle of redirections.

# Checklist:
- [X] I have pulled from master before submitting this PR.
  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [x] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -243,6 +243,49 @@ follows.")
                                                      :fallback-url "https://reddit.com")
                                       %slot-value%))))))
 
+      (:nsection :title "History"
+        (:p "Nyxt history model is a tree whose nodes are URLs. It branches out through all
+the buffers. If you create a new buffer (via " (:nxref :command 'nyxt/hint-mode:follow-hint-new-buffer)
+" or " (:nxref :command 'make-buffer) "), it becomes a new history branch originating
+from the branch of the previous buffer.")
+        (:p "History can be navigated with the arrow keys in the status buffer, or with
+commands like " (:nxref :command 'nyxt/history-mode:history-backwards) " and "
+            (:nxref :command 'nyxt/history-mode:history-forwards)
+            " (which the arrows are bound to).")
+        (:p "If the beyond-buffer-boundaries behavior sounds like too much to you, or you
+prefer the behavior of Nyxt 2, where the history was still a tree, but was not
+spilling across the buffers, then configure "
+            (:nxref :slot 'nyxt/history-mode:conservative-history-movement-p
+              :class-name 'nyxt/history-mode:history-mode)
+            " to be T. This would make all buffers to have their own history, not connected
+to the other buffers at all. All the history commands (like "
+            (:nxref :command 'nyxt/history-mode:history-backwards) " and "
+            (:nxref :command 'nyxt/history-mode:history-forwards)
+            ") will only work inside the buffer history then.")
+        (:p "Nyxt supra-buffer history has benefits, though: it optimizes browsing patterns
+into more intuitive and productive structures. One particular pattern Nyxt
+history optimizes is hub-and-spoke search, where you keep returning to a certain
+hub to start your search/navigation from a familiar point. You can enable the
+optimization (merely going back in history to the hub page, instead of creating
+a new history node) for this strategy by configuring "
+            (:nxref :slot 'nyxt/history-mode:backtrack-to-hubs-p
+              :class-name 'nyxt/history-mode:history-mode)
+            " to T.")
+        (:p "Another useful side to Nyxt tree-like history are braching-aware history
+commands, like "
+            (:nxref :command 'nyxt/history-mode:history-forwards-query)
+            ", allowing one to choose which branch of history they are going to visit, if
+there are several. If there's only one branch, then this command behaves much
+like regular " (:nxref :command 'nyxt/history-mode:history-forwards) ".")
+        (:p "There are commands that allow to move across all the history before or after
+the current node:")
+        (:ul (list-command-information '(nyxt/history-mode:history-backwards-query
+                                         nyxt/history-mode:history-forwards-all-query
+                                         nyxt/history-mode:history-all-query)))
+        (:p "If you need to know more: most of the optimizations and data structures are
+in " (:nxref :package :history-tree) " library, while most of the Nyxt-specific interface is in "
+(:nxref :package :nyxt/history-tree-mode) "."))
+
       (:nsection :title "Downloads"
         (:p "See the " (:nxref :command 'nyxt/download-mode:list-downloads) " command and the "
             (:nxref :slot 'download-path :class-name 'buffer) " buffer slot documentation."))
--- a/source/mode/history.lisp
+++ b/source/mode/history.lisp
@@ -21,6 +21,20 @@ search.")
     nil
     :type boolean
     :documentation "Whether history navigation is restricted by buffer-local history.")
+   (backtrack-to-hubs-p
+    nil
+    :type boolean
+    :documentation "Go back in history when encountering an already visited page.
+For example
+example.com/.../1 ->  example.com/hub -> example.com/.../3 -> example.com/hub
+                                                              ^ current node here
+
+Gets optimized into this structure with this setting on:
+
+example.com/.../1 ->  example.com/hub -> example.com/.../3
+                      ^ current node here
+
+Experimental, may not always produce intuitive enough history trees.")
    (keyscheme-map
     (define-keyscheme-map "history-mode" ()
       keyscheme:default
@@ -382,6 +396,35 @@ internally, but this display is clearer and more navigable."
     (:p (format nil "The last ~a history entries:" limit))
     (:ul (:raw (nyxt::history-html-list :limit limit)))))
 
+(-> history-add (quri:uri &key (:title string) (:buffer buffer)) *)
+(defun history-add (url &key (title "") (buffer (current-buffer)))
+  "Add URL to the global/buffer-local history.
+If `backtrack-to-hubs-p' is on, go back to the matching parent instead.
+The `implicit-visits' count of the `htree:current' node is incremented in any
+case."
+  (files:with-file-content (history (history-file (current-buffer))
+                            :default (nyxt::make-history-tree))
+    (unless (or (url-empty-p url)
+                ;; If buffer was not registered in the global history, don't
+                ;; proceed.  See `buffer's `customize-instance' `:after' method.
+                (not (htree:owner history (id buffer))))
+      (let* ((mode (find-submode 'history-mode))
+             (parent-position (when mode
+                                (and (backtrack-to-hubs-p mode)
+                                     (position url (htree:all-parents history :owner (id buffer))
+                                               :test #'quri:uri-equal
+                                               :key (compose #'url #'htree:data))))))
+        (if parent-position
+            (htree:backward history (id buffer) (1+ parent-position))
+            (htree:add-child (make-instance 'nyxt::history-entry
+                                            :url url
+                                            :title title)
+                             history
+                             (id buffer))))
+      (let* ((entry (htree:data (htree:current (htree:owner history (id buffer))))))
+        (setf (title entry) title)
+        (incf (nyxt::implicit-visits entry))))))
+
 (export-always 'blocked-p)
 (defun blocked-p (url mode)
   (find-if (rcurry #'str:starts-with? (render-url url))
@@ -404,8 +447,8 @@ internally, but this display is clearer and more navigable."
         ;; WARNING: We add buffer's URL instead of the URL argument because in
         ;; case of some redirects (like Wikipedia's) the buffer URL is the final
         ;; one.
-        (nyxt::history-add (url buffer) :title (title buffer)
-                                        :buffer buffer)))
+        (history-add (url buffer) :title (title buffer)
+                                  :buffer buffer)))
     url))
 
 (defmethod nyxt:on-signal-notify-uri ((mode history-mode) url)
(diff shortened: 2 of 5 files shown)


aadcg:
I'm stepping out of this discussion as I don't know how our history facilities evolved over times.

aartaka:
Well, you can compare the current state of things with this PR and report whether it's convenient or not, at least (: This PR, after all, is not about structure, but rather about the UX.

Ambrevar:
The first cancellation fix is welcome, of course!

About hub-URL-revisitation, since it's debatable but has its uses, I suggest to make it an option.
We could make add a slot to the `history` class, but that would mean toggling the option globally, + it's yet another change to this library.

Can't we implement the same thing from `mode/history.lisp`?  Should be easy enough: check if a parent matches, if so call `htree:visit-all`, if not call `htree:add-child`.

Then this behaviour can become a mode option for history-mode, and thus be local.  Thoughts?

aartaka:
> About hub-URL-revisitation, since it's debatable but has its uses, I suggest to make it an option. We could make add a slot to the `history` class, but that would mean toggling the option globally, + it's yet another change to this library.
> 
> Can't we implement the same thing from `mode/history.lisp`? Should be easy enough: check if a parent matches, if so call `htree:visit-all`, if not call `htree:add-child`.

Indeed, sounds like a better option to me!

aartaka:
@Ambrevar, redirections problem reported in #2660.

aartaka:
Hub back-tracking implemented in (failing) d8cac127efa2ad26759ed2a665a0931f74f05058.

source/mode/history.lisp
```diff
@@ -21,6 +21,11 @@ search.")
     nil
     :type boolean
     :documentation "Whether history navigation is restricted by buffer-local history.")
+   (backtrack-to-hubs-p
+    nil
+    :type boolean
+    :documentation "Go back in history when encountering an already visited page.
```
Ambrevar, line 27: Maybe add an example here, because this sentence alone is not very clear.

source/mode/history.lisp
```diff
@@ -381,6 +386,32 @@ internally, but this display is clearer and more navigable."
...
 
+(-> history-add (quri:uri &key (:title string) (:buffer buffer)) *)
+(defun history-add (url &key (title "") (buffer (current-buffer)))
+  "Add URL to the global/buffer-local history.
+The `implicit-visits' count is incremented."
+  (files:with-file-content (history (history-file (current-buffer))
+                            :default (nyxt::make-history-tree))
+    (unless (or (url-empty-p url)
+                ;; If buffer was not registered in the global history, don't
+                ;; proceed.  See `buffer's `customize-instance' `:after' method..
```
Ambrevar, line 397: Spurious double "..".

Ambrevar (commented):
Looks good to me!

aartaka:
I've added some tests for history backtracking, and they seem to fail, although backtracking works just fine when using Nyxt from this branch. @Ambrevar, any idea which part of the state I may be missing here?

Ambrevar:
The error is 

```lisp
  | ERROR: There is no applicable method for the generic function
  |          #<STANDARD-GENERIC-FUNCTION HISTORY-TREE:DATA (2)>
  |        when called with arguments
  |          (NIL).
```

You'd need a backtrace to be sure (you can run lisp-unit2 tests with a debugging context), but I'm guessing the error comes from.

I don't think the problem is with state, but rather with timing: the programmatic version might be too fast and thus some history operations might get skipped.

aartaka:
That is strange. Should it not synchronize the contents of the file between calls to `files:content`?

Ambrevar:
> Should it not synchronize the contents of the file between calls to files:content?

Not that simple: it does, but writing to disk might be slower than the call, so if you check the content of the file externally, it might not reflect the content in memory.

aartaka:
> > Should it not synchronize the contents of the file between calls to files:content?
> 
> Not that simple: it does, but writing to disk might be slower than the call, so if you check the content of the file externally, it might not reflect the content in memory.

Maybe block `files:content` until the disk operations are done then?

Ambrevar:
No, because the point of `nfiles:content` is to be fast. 

Actually `nfiles:content` has a `wait-p` option.  So this guarantees that the threaded operation is done upon return whant `wait-p` is T.  However this does not guarantee that the OS is done reading/writing the file.

Ambrevar:
In any case, I don't have a clear idea of the problem.  Not sure it has to do with Nfiles.  I'd need to have a closer look.

aadcg:
What's the status here?

aartaka:
Need to rebase and test this with the updates to `nfiles` and see whether those fix the problem of `nfiles:content` returning outdated results sometimes.

aartaka:
Rebased over `master`, squashed all the fix/patch commits, ensured all the experimental features (hub backtracking) are behind an opt-in flag—should be ready to merge!

aartaka:
The problem I attributed to Nfiles was rather the problem of parent-finding logic in my code （＠´＿｀＠）

Should be fixed now.

aartaka:
I just realized we have no manual entry for history and its configuration. Have to fix it. 

@aadcg, if you feel like another round of review, then I'll be glad to see your comments. In the meanwhile, I'll write up the history entry of the manual!

source/manual.lisp
```diff
@@ -243,6 +243,45 @@ follows.")
                                                      :fallback-url "https://reddit.com")
                                       %slot-value%))))))
 
+      (:nsection :title "History"
+        (:p "Nyxt history model is a tree—and it's a tree that branches out through all the
+buffers. If you create a new buffer (via " (:nxref :command 'nyxt/hint-mode:follow-hint-new-buffer)
```
aadcg, line 248: I think the first sentence should read: "Nyxt history model is a tree whose nodes are URLs."

source/manual.lisp
```diff
@@ -243,6 +243,45 @@ follows.")
...
+" or " (:nxref :command 'make-buffer) "), it becomes a new history branch originating
+from the branch of the current buffer.")
```
aadcg, line 250: It's not very explicit for the reader. The concept buffer is very confusing because it's never clear whether it's the one before or after running a command that creates a buffer and focus it.

source/manual.lisp
```diff
@@ -243,6 +243,45 @@ follows.")
...
+        (:p "If this behavior sounds like too much to you, or you prefer the behavior of
+Nyxt 2, where the history was still a tree, but was not spilling across the
+buffers, then configure "
+            (:nxref :slot 'nyxt/history-mode:conservative-history-movement-p
+              :class-name 'nyxt/history-mode:history-mode)
```
aadcg, line 255: This a doubt I have as a user. What's the difference between `conservative-history-movement-p` and `global-history-p`?

source/manual.lisp
```diff
@@ -243,6 +243,45 @@ follows.")
...
+            " to be T. This would make all buffers to have their own history, not connected
+to the other buffers at all. All the history commands (like "
+            (:nxref :command 'nyxt/history-mode:history-backwards) " and "
+            (:nxref :command 'nyxt/history-mode:history-forwards)
+            ") will only work inside the buffer history then.")
+        (:p "Nyxt supra-buffer history has benefits, though: it optimizes browsing patterns
+into something more or less intuitive and productive. One particular pattern
```
aadcg, line 262: "more or less" is not an expression suitable to be used in a manual.

source/mode/history.lisp
```diff
@@ -21,6 +21,20 @@ search.")
...
+    :documentation "Go back in history when encountering an already visited page.
+For example
+github.com/.../pull/1 ->  github.com/notifications -> github.com/.../pull/3 -> github.com/notifications
+                                                                               ^ current node here
+
+Gets optimized into this structure with this setting on:
+
+github.com/.../pull/1 ->  github.com/notifications -> github.com/.../pull/3
+                          ^ current node here
+
```
aadcg, line 36: I think it would be more informative and concise to name these example URLs differently: "URL-A", "URL-B" and "URL-C".

source/manual.lisp
aartaka: Waaaaaaaait...

source/manual.lisp
aartaka: Right!

tests/offline/global-history.lisp
```diff
@@ -40,10 +43,26 @@
...
         (assert-eq 2
                    (length (htree:all-data (files:content file))))
+        (sleep 0.2)
+        ;; Enable hub backtracking.
+        (setf (nyxt/history-mode:backtrack-to-hubs-p mode) t)
+        ;; Add the same URL and test if backtracking worked.
+        (nyxt/history-mode::history-add (quri:uri "http://example.org"))
+        (assert-eq 2
+                   (length (htree:all-data (files:content file))))
+        (sleep 0.2)
```
Ambrevar, line 57: Why the `sleep` calls?

source/mode/history.lisp
```diff
@@ -382,6 +396,31 @@ internally, but this display is clearer and more navigable."
...
+  "Add URL to the global/buffer-local history.
+The `implicit-visits' count is incremented."
+  (files:with-file-content (history (history-file (current-buffer))
+                            :default (nyxt::make-history-tree))
+    (unless (or (url-empty-p url)
+                ;; If buffer was not registered in the global history, don't
+                ;; proceed.  See `buffer's `customize-instance' `:after' method.
+                (not (htree:owner history (id buffer))))
+      (alex:if-let ((parent-position (and (find-submode 'history-mode)
+                                          (backtrack-to-hubs-p (find-submode 'history-mode))
```
Ambrevar, line 410: Let-bind `history-mode`.

Ambrevar (commented):
Looks good to me beside 2 nits :)

source/mode/history.lisp
aartaka: Done! A shame there's no `if-let*` in any of the libraries we use...

tests/offline/global-history.lisp
aartaka: It was an artifact of the times I tried to fix the problem of backtracking not getting tested. Removed now!

aartaka:
Should be all covered now!

Ambrevar:
Nice, feel free to merge.

Ambrevar:
One last nit: Maybe you can extend the `history-add` docstring to explain the new optional behaviour.

In any case, feel free to merge.

aartaka:
Docsting extended, commits squashed, branch merged :D

source/manual.lisp
```diff
@@ -243,6 +243,49 @@ follows.")
                                                      :fallback-url "https://reddit.com")
                                       %slot-value%))))))
 
+      (:nsection :title "History"
+        (:p "Nyxt history model is a tree whose nodes are URLs. It branches out through all
```
aadcg, line 247: Pity that it has been merged already. Typo "Its".

aadcg:
The manual section is not satisfactory but I'll rewrite it some day.

Issue #2619, opened 2022-11-04
Polish renderer API


aadcg:
We now have each renderer in its own package but the API could be improved.

The symptom of the issue manifests itself from `nyxt/renderer/foo` being required to use unexported symbols from the `nyxt` package. Find a below a list of those:

```
nyxt::*context-menu-commands*
nyxt::*headless-p*
nyxt::*interactive-p*
nyxt::*run-from-repl-p*
nyxt::*schemes*
nyxt::+renderer+
nyxt::active-buffer
nyxt::active-prompt-buffers
nyxt::buffer-hide
nyxt::fullscreen-p
nyxt::get-keymap
nyxt::javascript-error-handler
nyxt::list-commands
nyxt::make-channel
nyxt::maximized-p
nyxt::open-urls
nyxt::panel-buffers
nyxt::print-message
nyxt::prompt-buffer-p
nyxt::run-async
nyxt::status
nyxt::tls-help
nyxt::update-prompt-input
nyxt::webkit-web-context
nyxt::without-package-locks
```

As the support for more renderers grow, the API should support a mechanism for users to switch between them.

The following operation should be possible:

- Start Nyxt with a given renderer enabled.
- Quit Nyxt, without killing the inferior Lisp process.
- Switch renderer using a single function call.
- Start Nyxt with another renderer enabled.

It would be great if it'd be possible to set the renderer when spawning windows :)

Follow the relevant discussions in 937262e and #2612.

Ambrevar:
Renderer switching is now possible from the same REPL instance. see c52e797abef80e646daaea72733cf44ac2fca43f.

aadcg:
@Ambrevar can we get an example on how to switch renderer from the REPL somewhere (comment, docstring, etc).

Also, could you add a simple test that captures this invariant? Thanks.

Ambrevar:
I've added an example in commit f5181edf520674cd79d1c3aa83d24bf54488afb3.
However it does not work anymore.  I tried commit c52e797abef80e646daaea72733cf44ac2fca43f of the pull request where it was tested, and it's failing there too.

Either I'm using the wrong invocation, or something else broke.  Possibly a Guix update broke it.

aadcg:
@Ambrevar, hmm, that's interesting... Maybe WebKitGTK internals changed slightly.

[...]

Issue #2609, opened 2022-10-31
Testing the `nyxt/gi-gtk` system seems to be broken


aadcg:
While working on #2605, I noticed that I couldn't test `nyxt/gi-gtk` on master.

From the REPL, I issued `,test-system RET nyxt/gi-gtk` and got:

> WARNING: System definition file #P"/home/aadcg/common-lisp/lisp-unit2/lisp-unit2.asd" contains definition for system "lisp-unit2-test". Please only define "lisp-unit2" and secondary systems with a name starting with "lisp-unit2/" (e.g. "lisp-unit2/test") in that file.
> WARNING: System definition file #P"/gnu/store/r860vmxghysv74435fvjj4g52xzhw41y-cl-symbol-munger-0.0.1-1.97598d4/share/common-lisp/source/cl-symbol-munger/symbol-munger.asd" contains definition for system "symbol-munger-test". Please only define "symbol-munger" and secondary systems with a name starting with "symbol-munger/" (e.g. "symbol-munger/test") in that file.
> WARNING: redefining HU.DWIM.DEFCLASS-STAR::PROCESS-SLOT-DEFINITION in DEFUN
> ; compiling file "/home/aadcg/common-lisp/nyxt/source/renderer/gi-gtk.lisp" (written 17 OCT 2022 03:10:34 PM):
> 
> ; wrote /home/aadcg/.cache/common-lisp/sbcl-2.2.6-linux-x64/home/aadcg/common-lisp/nyxt/source/renderer/gi-gtk-tmpRF6DISR1.fasl
> ; compilation finished in 0:00:00.048
> ; compiling file "/home/aadcg/common-lisp/nyxt/tests/renderer-package.lisp" (written 31 OCT 2022 01:04:22 PM):
> 
> ; wrote /home/aadcg/.cache/common-lisp/sbcl-2.2.6-linux-x64/home/aadcg/common-lisp/nyxt/tests/renderer-package-tmp3OXMRLFZ.fasl
> ; compilation finished in 0:00:00.020
> ; compiling file "/home/aadcg/common-lisp/nyxt/tests/renderer-offline/set-url.lisp" (written 31 OCT 2022 01:04:22 PM):
> 
> ; wrote /home/aadcg/.cache/common-lisp/sbcl-2.2.6-linux-x64/home/aadcg/common-lisp/nyxt/tests/renderer-offline/set-url-tmp2XEJNW23.fasl
> ; compilation finished in 0:00:00.020
[...]

Issue #1089, opened 2021-01-04
libraries/class-star: Make define-class anaphoric?


aartaka:
Anaphoric class definitions can save us code lines and bring more power in class definions. One of the use-cases is the issue of resizing `status-buffer` recently brought up on IRC. CSS of `status-buffer` is not dependent on `height` of it, while it ideally should, so that if user redefines height of `status-buffer`, font gets resized too. That's how I imagine definition of `status-buffer` then (`%this-class` is an anaphoric variable referring to the class being defined):

``` lisp
(define-class status-buffer (user-internal-buffer)
  ((height 20
           :type integer
           :documentation "The height of the status buffer in pixels.")
   (style #.(cl-css:css
             `((body
                :background "rgb(200, 200, 200)"
                :font-size ,(str:concat (object-string (- (height %this-class) 6)) "px")
                :color "rgb(32, 32, 32)"
                :padding 0
                :margin 0
                :line-height ,(str:concat (object-string (height %this-class)) "px"))
               (".arrow"
                :width ,(str:concat (object-string (/ (height %this-class) 2)) "px")
                :height ,(str:concat (object-string (height %this-class)) "px"))
               (".arrow-right"
                :clip-path "polygon(0 100%, 100% 50%, 0 0)")
               (".arrow-left"
                :clip-path "polygon(0 50%, 100% 100%, 100% 0)")
               ("#container"
                :display "grid"
                ;; Columns: controls, arrow, url, arrow, tabs, arrow, modes
                :grid-template-columns ,(format nil "115px ~apx 1fr ~apx 2fr ~apx 250px"
                                                (/ (height %this-class) 2)
                                                (/ (height %this-class) 2)
                                                (/ (height %this-class) 2))
                :overflow-y "hidden")
               ;; Other slots...
               ))))
  (:export-class-name-p t)
  (:export-accessor-names-p t)
  (:export-predicate-name-p t)
  (:accessor-name-transformer #'class*:name-identity))
```

Other possible features/applications:
- MOP and slot initform rewriting would be used under the hood.
- If the slot is not present in class declaration, we can look at the slots of parent classes that the defined one inherits.
- Having this, we can set the style of all internal/web buffers through buffer slots and style slots referring to them (e.g. having `background-color`, `foreground-color` slots of buffer referenced  in `style` slot definition of the same class), all in lispy syntax with good type-checking and simple redefinitions. A good step towards Nyxt theming :)
[...]

PR #2666, opened 2022-11-24
Fix global profile and make renderer first class objects

### LINKED ISSUE #2663: `history-restoration` tests fails
Defined in `/tests/offline/global-history.lisp`.

Could someone that knows how we manage history fix it?  

Run this single test  `(with-summary () (run-test 'history-restoration))` from the REPL with `*package*` set to `nyxt/test`.

Ambrevar:
# Description

In an attempt to make tests more reproducible and less stateful, and more generally everything less stateful, I've implemented some important changes:

- `*global-profile*` is now `(global-profile)`.
- Renderers are now first class objects!
  - It's possible to switch renderers between two instances from the same REPL!
  - The dirty package lock violations are all gone.
  - It's now easier to see thanks to `install` where each renderer hooks itself in.

Fixes #2663 and #2619.

Prepares the fix for #2609.

# Discussion

The `(history-file browser)` initialization is too specific:

```lisp
(defmethod initialize-instance :after ((browser browser)
                                       &key (history-file nil history-file-p)
                                       &allow-other-keys)
  "Ensure `history-file' uses the browser profile."
  (setf (history-file browser)
        (if history-file-p
            history-file
            (make-instance 'history-file :profile (profile browser)))))
```

Can we be more general than that?

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1148,7 +1148,8 @@ This is a low-level function.  See `buffer-delete' for the high-level version."
 
 (export-always 'window-list)
 (defun window-list ()
-  (alex:hash-table-values (windows *browser*)))
+  (when *browser*
+    (alex:hash-table-values (windows *browser*))))
 
 (defun dummy-buffer-p (buffer)
   (eq 'buffer (type-of buffer)))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -33,10 +33,11 @@ inherited from the superclasses.")
     :documentation "Unique identifier for a buffer.")
    ;; TODO: Or maybe a dead-buffer should just be a buffer history?
    (profile
-    *global-profile*
+    (global-profile)
     :type nyxt-profile
     :documentation "Buffer profiles are used to specialize the behavior of
-various parts, such as the path of all data files.")
+various parts, such as the path of all data files.
+See also the `profile' slot in the `browser' class.")
    (url (quri:uri ""))
    (url-at-point (quri:uri ""))
    (title "")
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -441,7 +441,9 @@ query is not a valid URL, or the first keyword is not recognized.")
     :documentation "Select a download engine to use, such as `:lisp' or
 `:renderer'.")
    (history-file
-    (history-file *browser*)
+    (if *browser*
+        (history-file *browser*)
+        (make-instance 'history-file))
     :type history-file
     :documentation "File where to save the global history used by this buffer.
 See also `history-file' in `browser' for the global history restored on startup,
@@ -1006,7 +1008,11 @@ BUFFER's modes."
 
 (hooks:define-hook-type buffer (function (buffer)))
 
-(define-command make-buffer (&rest args &key (title "") modes (url (default-new-buffer-url *browser*)) parent-buffer
+(define-command make-buffer (&rest args &key (title "") modes
+                             (url (if *browser*
+                                      (default-new-buffer-url *browser*)
+                                      (quri:uri (nyxt-url 'new))))
+                             parent-buffer
                              no-history-p (load-url-p t) (buffer-class 'web-buffer)
                              &allow-other-keys)
   "Create a new buffer.
--- a/source/global.lisp
+++ b/source/global.lisp
@@ -78,10 +78,10 @@ is 4005, default set to 4006 in Nyxt to avoid collisions).")
   "The port that Slynk will open a new server on (default Emacs Sly port
 is 4005, default set to 4006 in Nyxt to avoid collisions).")
 
-(defparameter +renderer+ nil "The renderer used by Nyxt. This value is meant to
-be set to a string by the renderer itself. This variable exists to allow for
-reporting by users, it does not create any functional differences in the
-execution of Nyxt.")
+(declaim (type (maybe renderer) *renderer*))
+(defparameter *renderer* nil
+  "The renderer used by Nyxt.
+It can be changed between two runs of Nyxt when run from a Lisp REPL.")
 
 (alex:define-constant +nyxt-critical-dependencies+
     '(:cl-cffi-gtk
--- a/source/renderer/gi-gtk.lisp
+++ b/source/renderer/gi-gtk.lisp
@@ -16,40 +16,62 @@
 For now it is also partly based on `nyxt/renderer/gtk'."))
 (in-package :nyxt/renderer/gi-gtk)
 
-(setf nyxt::+renderer+ "GI-GTK")
 (push :nyxt-gi-gtk *features*)
-(handler-bind ((warning #'muffle-warning))
-  (let ((renderer-thread-name "Nyxt renderer thread"))
-    (defmethod nyxt/renderer/gtk::renderer-thread-p ((renderer (eql "GI-GTK")) &optional (thread (bt:current-thread)))
-      (string= (bt:thread-name thread)
-               #+darwin
-               "thread"
-               #-darwin
-               renderer-thread-name))
-    (defmethod ffi-initialize ((browser nyxt/renderer/gtk:gtk-browser) urls startup-timestamp)
-      "On GNU/Linux we can create a separate thread to launch the GTK
-interface. On Darwin, we must run the GTK thread on the main thread."
-      (declare (ignore urls startup-timestamp))
-      (log:debug "Initializing GI-GTK Interface")
-      (setf (uiop:getenv "WEBKIT_FORCE_SANDBOX") "0")
-      ;; TODO: Do not run the GTK loop again when T?
-      (setf nyxt/renderer/gtk::gtk-running-p t)
-      (flet ((main-func ()
-               (with-protect ("Error on GI-GTK thread: ~a" :condition)
-                 (glib:g-set-prgname "nyxt")
-                 #+GTK-3-4
-                 (gdk:gdk-set-program-class "Nyxt")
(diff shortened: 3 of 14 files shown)


aartaka:
But why do we have to pass `history-file`? Why not just an initform?

... Ahhhh, I see, `(profile browser)`. Then how about introducing constructor slot values, as per #1089?

source/buffer.lisp
```diff
@@ -1005,7 +1008,11 @@ BUFFER's modes."
 
 (hooks:define-hook-type buffer (function (buffer)))
 
-(define-command make-buffer (&rest args &key (title "") modes (url (default-new-buffer-url *browser*)) parent-buffer
+(define-command make-buffer (&rest args &key (title "") modes
+                             (url (if *browser*
+                                      (default-new-buffer-url *browser*)
```
aartaka, line 1013: Alternatively, dispatch `default-new-buffer-url` over `null`. Or is it too much of a fuzzy CLOS trickery?

source/buffer.lisp
```diff
@@ -1141,7 +1148,8 @@ This is a low-level function.  See `buffer-delete' for the high-level version."
 
 (export-always 'window-list)
 (defun window-list ()
-  (alex:hash-table-values (windows *browser*)))
+  (when *browser*
```
aartaka, line 1151: This `(when *browser* ...)` and `(if *browser* ...)` is certainly safe-enough, but it doesn't bring sense of safety to me. Maybe make it a function too (`(current-browser)` :D)?

Or am I over-engineering?

source/global.lisp
```diff
@@ -78,10 +78,10 @@ is 4005, default set to 4006 in Nyxt to avoid collisions).")
   "The port that Slynk will open a new server on (default Emacs Sly port
 is 4005, default set to 4006 in Nyxt to avoid collisions).")
 
-(defparameter +renderer+ nil "The renderer used by Nyxt. This value is meant to
-be set to a string by the renderer itself. This variable exists to allow for
-reporting by users, it does not create any functional differences in the
-execution of Nyxt.")
+(declaim (type (maybe renderer) *renderer))
```
aartaka, line 81: Missing asterisk?

source/renderer/gi-gtk.lisp
```diff
@@ -16,40 +16,62 @@
...
+        ;; See comment about FreeBSD in gtk.lisp
+        (uiop:quit (slot-value browser 'nyxt::exit-code) #+freebsd nil)))
+    #+darwin
+    (main-func)))
+
+(nyxt/renderer/gtk:define-ffi-method ffi-kill-browser ((browser gi-gtk-browser))
+  (unless nyxt::*run-from-repl-p*
+    (gir:invoke ((gir:ffi "Gtk" "3.0") 'main-quit))))
+
+(defmethod install ((renderer gi-gtk-renderer))
```
aartaka, line 71: What about `uninstall`? Is it necessary here?

source/renderer/gtk.lisp
```diff
@@ -223,7 +210,7 @@ the renderer thread, use `defmethod' instead."
     `(defmethod ,name ,args
        ,@(sera:unsplice docstring)
        ,@declares
-       (if (renderer-thread-p nyxt::+renderer+)
+       (if (renderer-thread-p nyxt::*renderer*)
```
aartaka, line 213: But then, should `*renderer*` be exported? I feel like it should...

source/renderer/gtk.lisp
```diff
@@ -2282,3 +2241,33 @@ See `make-buffer' for a description of the other arguments."
              :sources (list (make-instance 'prompter:raw-source :name "New context")
                             'context-source))))
   (apply #'make-buffer args))
+
+(defmethod install ((renderer gtk-renderer))
+  (flet ((set-superclasses (renderer-class-sym+superclasses)
+           (closer-mop:ensure-class (first renderer-class-sym+superclasses)
+                                    :direct-superclasses (rest renderer-class-sym+superclasses)
+                                    :metaclass 'interface-class)))
+    (mapc #'set-superclasses '((renderer-browser gtk-browser)
```
aartaka, line 2250: Wow, that's a smart one :D

source/renderer/gtk.lisp
```diff
@@ -2282,3 +2241,33 @@ See `make-buffer' for a description of the other arguments."
...
+                               (nyxt/download-mode:renderer-download gtk-download)
+                               (renderer-request-data gtk-request-data )
+                               (renderer-scheme gtk-scheme)
+                               (nyxt/user-script-mode:renderer-user-style gtk-user-style)
+                               (nyxt/user-script-mode:renderer-user-script gtk-user-script)))))
+
+(defmethod uninstall ((renderer gtk-renderer))
+  (flet ((remove-superclasses (renderer-class-sym)
+           (closer-mop:ensure-class renderer-class-sym
+                                    :direct-superclasses '()
```
aartaka, line 2262: This one feels wrong. So, not even `standard-class`? Maybe compute the current set of classes and remove all the classes in `nyxt*` packages?

aartaka (commented):
The description of this PR over email looked scarier that it turns out to be! Wonderful work, Pierre! I've marked some moments that stood out to me, but in general I am eager to see this merged!

source/buffer.lisp
Ambrevar: But how would `current-browser` fix the problem here?  `(windows (current-browser))` would still raise an error if there is no browser.

Alternatively, we could have a macro, like `(with-renderer ...)`, but that would be just as long.

source/global.lisp
Ambrevar: Oops!

source/renderer/gi-gtk.lisp
Ambrevar: No, because `(uninstall gi-gtk-renderer)` calls the next method which is `(uninstall gtk-renderer)`.  The beauty of CLOS!

source/renderer/gtk.lisp
Ambrevar: Me too, but let's deal with the exports when we are nearing 3.0 release.

source/renderer/gtk.lisp
Ambrevar: Nope, this is  correct CLOS.  Try this:

```lisp
(closer-mop:ensure-class 'foo :direct-superclasses '())
(closer-mop:class-direct-superclasses (find-class 'foo))
; => (#<STANDARD-CLASS COMMON-LISP:STANDARD-OBJECT>)
```

source/buffer.lisp
Ambrevar: On the one hand CLOS approach is better because it applies everywhere.

On the other hand, we would need to do it for all `browser` entries.

Wait... What about ensuring `*browser*` is _always_ bound to a default browser? This way no need to fallback on the initform, `(default-new-buffer-url *browser*)` would do the right thing!

Ambrevar:
> ... Ahhhh, I see, `(profile browser)`. Then how about introducing constructor slot values, as per #1089?

This could be done with MOP.  Maybe for another pull request?

source/buffer.lisp
Ambrevar: See https://github.com/atlas-engineer/nyxt/issues/2670

source/renderer/gi-gtk.lisp
aartaka: Right, I missed that part :D

source/renderer/gtk.lisp
aartaka: Sure!

source/renderer/gtk.lisp
aartaka: Can't stop being amazed with the deliberation that went into CL `(/^▽^)/`

aartaka:
> > ... Ahhhh, I see, `(profile browser)`. Then how about introducing constructor slot values, as per #1089?
> 
> This could be done with MOP. Maybe for another pull request?

Sure :)

Ambrevar:
Ok to merge?

source/buffer.lisp
aartaka: We can move the `(make-instance 'browser)` into the `current-browser` body, this way lazily offloading it until the browser is actually requred and will likely be initializeable. That would solve the problem of NIL browser, because a necessary one will be created on demand.

aartaka:
Please, go through the checklist again and make sure you've got the documentation and tests written/updated in regards to this change.

Ambrevar:
Done.

Ambrevar:
OK to merge?

source/buffer.lisp
Ambrevar: A drawback of `(current-browser)` is that it makes it harder to let-bind a browser.

Is there a way to dynamically override functions in CL?

aartaka: Yes: `flet`?

aartaka:
Yes, OK to merge!

source/buffer.lisp
Ambrevar: No, it's lexical:

```lisp
> (defun foo () 17)
FOO
> (defun bar () (foo))
BAR
$> (flet ((foo () 18))
				(bar))
17
```

Ambrevar:
Merged with c52e797abef80e646daaea72733cf44ac2fca43f.

PR #2671, opened 2022-11-25
Fix nil browser

### LINKED ISSUE #2670: Always bind `*browser*`
In https://github.com/atlas-engineer/nyxt/pull/2666#discussion_r1031497582 we discussed the problem of having to protect against NIL `*browser*` all over the place.  Then having to extract some default value, which duplicates it since it's already in the slot initforms.

Suggestion: always bind `*browser*` to at least `(make-instance 'browser)`.

I believe this should do the right thing.  Thoughts?

Ambrevar:
# Description

Follow-up to #2666.

This ensures `*browser*` is always bound, thus removing a whole class of potential errors.

Fixes #2670.

# Discussion

Can we set `*browser*` at compile time and save it in the image?

# To do:

- [ ] Merge #2666 first.
- [ ] Test!

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/start.lisp
+++ b/source/start.lisp
@@ -179,9 +179,9 @@ It skips configuration files and other data files like the history."))))
       (mapc #'destroy-thread* (non-terminating-threads *browser*))
       (ffi-kill-browser *browser*)
       ;; Reset global state.
-      (setf *browser* nil
-            *options* nil)
       (uninstall *renderer*)
+      (setf *browser* (make-instance 'browser)
+            *options* nil)
       ;; On FreeBSD this may cause freeze. Also we have to pass
       ;; FINISH-OUTPUT = NIL in FFI-INITIALIZE.
       #-freebsd
@@ -541,6 +541,10 @@ Finally, run the browser, load URL-STRINGS if any, then run
   (log:info "Source location: ~s" (files:expand *source-directory*))
 
   (install *renderer*)
+
+  (unless (find-if (sera:eqs 'messages-appender) (log4cl:all-appenders)
+                   :key #'sera:class-name-of)
+    (log4cl:add-appender log4cl:*root-logger* (make-instance 'messages-appender)))
 
   (when (getf *options* :profile)
     (alex:if-let ((profile-class (find-profile-class (getf *options* :profile))))
--- a/source/user-files.lisp
+++ b/source/user-files.lisp
@@ -13,13 +13,8 @@
   (:documentation "With the default profile all data is persisted to the
 standard locations."))
 
-(export-always 'global-profile)
-(defun global-profile ()
-  "The profile to use in the absence of buffers and on browser-less variables."
-  (profile *browser*))
-
 (define-class nyxt-file (files:gpg-file)
-  ((files:profile (global-profile))
+  ((files:profile (profile *browser*))
    (files:on-external-modification 'files:reload)
    (editable-p
     t
--- a/source/user-files.lisp
+++ b/source/user-files.lisp
@@ -16,11 +16,7 @@ standard locations."))
 (export-always 'global-profile)
 (defun global-profile ()
   "The profile to use in the absence of buffers and on browser-less variables."
-  (or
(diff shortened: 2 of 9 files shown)


source/start.lisp
```diff
@@ -179,9 +179,9 @@ It skips configuration files and other data files like the history."))))
       (mapc #'destroy-thread* (non-terminating-threads *browser*))
       (ffi-kill-browser *browser*)
       ;; Reset global state.
-      (setf *browser* nil
-            *options* nil)
       (uninstall *renderer*)
+      (setf *browser* (make-instance 'browser)
+            *options* nil)
```
Ambrevar, line 184: To do: wrap this around `(when *run-from-repl-p*...)` since it does not need to be done otherwise.

source/start.lisp
```diff
@@ -542,6 +542,10 @@ Finally, run the browser, load URL-STRINGS if any, then run
 
   (install *renderer*)
 
+  (unless (find-if (sera:eqs 'messages-appender) (log4cl:all-appenders)
+                   :key #'sera:class-name-of)
+    (log4cl:add-appender log4cl:*root-logger* (make-instance 'messages-appender)))
+
```
Ambrevar, line 548: Test this!

source/user-files.lisp
```diff
@@ -13,17 +13,8 @@
...
 
-(export-always 'global-profile)
-(defun global-profile ()
-  "The profile to use in the absence of buffers and on browser-less variables."
-  (or
-   (when *browser* (profile *browser*))
-   (alex:when-let ((profile-class (find-profile-class (getf *options* :profile))))
-     (make-instance profile-class))
-   (make-instance 'nyxt-profile)))
-
```
Ambrevar, line 24: Tests need to be updated to reflect `global-profile` disappearance.

aartaka:
That feels strange—initializing the browser at load-time. Especially given that parent class juggling and `initialize-instance :after` are involved.

aartaka:
Can it provoke any undesired side-effects?

Ambrevar:
To clarify: this initial `*browser*` is a dummy value.  We discard it when we run `start`.  So no undesired side effects here.

aartaka:
Oh, I see, `browser` has no `initialize-instance :after` and, supposedly, no renderer-specific side-effects!

Ambrevar:
Note that the dummy renderer is an instantiation of the browser class _before_ the renderer is installed.  In effect, it's a renderer-less browser.

aartaka:
But then `update-instance-for-redefined-class` might kick in any moment and make it a renderer-enabled object with all the consequences, no?

Ambrevar:
Fair point, but there is no consequence for the moment since instantiated a renderer browser has no side effects I believe.

Also ultimately I'd like to define `update-instance-for-redefined-class` for browser, buffer, window, etc., so that we can switch renderer at run time :)

aadcg:
>  But then update-instance-for-redefined-class might kick in any moment and make it a renderer-enabled object with all the consequences, no?

Before we merge, I'd like to address this point since it seems extremely important to me.

Did any of you actually tried to programmatically follow the steps @aartaka suggests?

Ambrevar:
Which steps?

aadcg:
My understanding is that the dummy renderer is a renderer-less instantiation of the browser class, but after loading a renderer package that can have side-effects. Did I misunderstand it?

Ambrevar:
Theoretically it can, but in practice it does not (for now).

If some day it would, then we could either ensure the side effects work also on the pre-instantiated browser, or we could specialize `update-instance-for-different-class`.

Ambrevar:
Oops!  Accidentally closed this.  See #2679 for a follow up.

PR #2673, opened 2022-11-25
Add color utils to libraries/theme

aartaka:
# Description

This is a more complete implementation of what @aadcg suggested in #2625: color parsing and contrast detection utilities as part of Nyxt theme library. The biggest complain I had to André's implementation was the it only implied hex-encoded colors, which is quite a rare and unreadable of a color, so we'd better have at least some more formats.

# Discussion

- Note that Nyxt-side use of this library relies on GDKRgba parsing, because it's more reliable. But it's a bummer, because having our own Lispy library for color would be much cooler. Does that make any sense to re-implement all of GDK parsing? @aadcg, am I over-optimizing?
- Note that there's `cl-colors` library (beaing a collateral dependency of Nyxt already), which is pretty close in functionality (getting color triples by names or hexes), but we require much more power than it provides, so re-implementation is an option.

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:


aartaka:
Note that [`gdk_rgba_parse`](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkrgba.c#L185) is _just_ 150 lines of C code, so it _should be_ pretty easy to implement (★^O^★)

libraries/theme/color-utils.lisp
```diff
@@ -0,0 +1,244 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :theme)
+
+(ql:update-all-dists)
```
Ambrevar, line 6: Why this line?

libraries/theme/color-utils.lisp
```diff
@@ -0,0 +1,244 @@
...
+   "Violet" '(0.93333334 0.50980395 0.93333334)
+   "Wheat" '(0.9607843 0.87058824 0.7019608)
+   "White" '(1.0 1.0 1.0)
+   "WhiteSmoke" '(0.9607843 0.9607843 0.9607843)
+   "Yellow" '(1.0 1.0 0.0)
+   "YellowGreen" '(0.6039216 0.8039216 0.19607843)))
+
+(export-always '*colors*)
+(defvar *colors* (alexandria:hash-table-keys *colors->triples*)
+  "List of CSS color names CSS.")
```
Ambrevar, line 191: Typo?

libraries/theme/color-utils.lisp
```diff
@@ -0,0 +1,244 @@
...
+
+(defun name->triple (color)
+  (gethash color *colors->triples*))
+
+(export-always 'parse-color)
+(defun parse-color (color)
+  "Parse the COLOR to a triple of (RED GREEN BLUE) float numbers in a [0, 1] range."
```
Ambrevar, line 198: Nit: "in a [0.0, 1.0] range" would highlight it's a float.

libraries/theme/color-utils.lisp
```diff
@@ -0,0 +1,244 @@
...
+  (cond
+    ((or (and (uiop:string-prefix-p "#" color)
+              (or (= 4 (length color))
+                  (= 7 (length color))))
```
Ambrevar, line 202: Why 4 and 7?  Same underneath, why 3 and 6?

libraries/theme/color-utils.lisp
```diff
@@ -0,0 +1,244 @@
...
+(export-always '*parsing-function*)
+(defvar *parsing-function* #'parse-color
+  "The function to parse color string to a (RED GREEN BLUE) triple.
+Defaults to `parse-color'.")
+
+(export-always 'relative-luminance)
+(defun relative-luminance (color)
+  "Compute relative luminance of COLOR."
+  ;; See https://www.w3.org/TR/WCAG20-TECHS/G18.html
+  (loop for const in '(0.2126 0.7152 0.0722)
```
Ambrevar, line 221: Better name for `const`?

libraries/theme/color-utils.lisp
```diff
@@ -0,0 +1,244 @@
...
+        for rgb-component in (parse-color color)
+        sum (* const (if (<= rgb-component 0.03928)
+                         (/ rgb-component 12.92)
+                         (expt (/ (+ rgb-component 0.055) 1.055) 2.4)))))
```
Ambrevar, line 225: Would it be possible to give meaningful names to these magic values?

aadcg:
> This is a more complete implementation of what @aadcg suggested in #2625: color parsing and contrast detection utilities as part of Nyxt theme library. The biggest complain I had to André's implementation was the it only implied hex-encoded colors, which is quite a rare and unreadable of a color, so we'd better have at least some more formats.

This is exactly the main source of confusion. I didn't propose myself to implement any of that. There are invariants in the way we think about Nyxt themes. Namely, we ensure that there's a minimum contrast ratio of 7:1 between colors and on-colors. Since those invariants were only in our minds, adding tests ensures we would be careful when making changes. That goal was achieved.  The reason I went with hex-encoded colors is because that's how we specify colors in our themes. Since I has no other goal, I didn't see any point in going further.

Also, implementing conversion utilities among colors specified via string, hex or RGB is the subject for another library that has little to do with `nyxt/theme`, and it already done or partially done is other CL libraries.

aadcg:
> - Note that Nyxt-side use of this library relies on GDKRgba parsing, because it's more reliable. But it's a bummer, because having our own Lispy library for color would be much cooler. Does that make any sense to re-implement all of GDK parsing? @aadcg, am I over-optimizing?
- Note that there's cl-colors library (beaing a collateral dependency of Nyxt already), which is pretty close in functionality (getting color triples by names or hexes), but we require much more power than it provides, so re-implementation is an option.

Can we start this from another perspective? 

What are you trying to achieve? :)

libraries/theme/color-utils.lisp
aadcg: Well, that depends on how you look at it. I read that interval as an interval of real numbers, but I'm not a programmer.

libraries/theme/color-utils.lisp
aadcg: `rgb-component-const`

libraries/theme/color-utils.lisp
aadcg: I'm sure there is a meaning to it, but it's out of scope. There's a link in a comment above that presents these magic values. It's a black box and, for our purposes, it's ok.

libraries/theme/color-utils.lisp
```diff
@@ -0,0 +1,244 @@
...
+
+(export-always 'contrast-ratio)
+(defun contrast-ratio (color1 color2)
+  "Compute contrast ratio between COLOR1 and COLOR2."
+  (let ((ratio (/ (+ (relative-luminance color1) 0.05)
+                  (+ (relative-luminance color2) 0.05))))
+    (max ratio (/ ratio))))
+
+(export-always 'contrasting-color)
+(defun contrasting-color (color)
```
aadcg, line 235: Ok, this is your goal. Now I understand the purpose of this PR.

My question is: is it worth to go to such lengths to beautify our `color-source`?

Or do think that making this computation is useful in other contexts?

libraries/theme/color-utils.lisp
```diff
@@ -0,0 +1,244 @@
...
+  "Find the contrasting color for COLOR.
+This computed color will look good as either the background for COLOR or text
+color on top of COLORed background."
+  (if (< (relative-luminance color)
+         ;; Roughly the luminance of #777, the color
+         ;; equally contrasting with black and white.
+         0.2)
+      "white"
+      "black"))
```
aadcg, line 244: I believe this is not correct.

Firstly you should compute the `luminance` of a color. If it's higher than 0.5, "white", otherwise "black".

Where `luminance`:

```lisp
(defun luminance (r g b)
  (/ (+ (* .2126 r) (* .7152 g) (* .0722 b)) 255))
```

libraries/theme/color-utils.lisp
aartaka: Yes, it's worth it, because it makes the color source more useful and immediate.

aartaka: Computing those things is also useful in theme initialization, for example. So that we don't need setting `on-*` colors explicitly.

aadcg: Clear. I just don't find this particularly important right now

aadcg:
>  Note that Nyxt-side use of this library relies on GDKRgba parsing, because it's more reliable. But it's a bummer, because having our own Lispy library for color would be much cooler. Does that make any sense to re-implement all of GDK parsing?

I think we should not implement any of that in Lisp, since GDK determines which color model the toolkit uses. If tomorrow we change to another toolkit that uses another color model, then we have just done work for nothing. I don't think it is desirable, for our purposes, to manipulate any that. 

> Note that there's cl-colors library (beaing a collateral dependency of Nyxt already), which is pretty close in functionality (getting color triples by names or hexes), but we require much more power than it provides, so re-implementation is an option

Note that there's [cl-colors2](https://notabug.org/cage/cl-colors2), a more up-to-date fork. I think we should fork it (or try to send them our changes) with the following addition:

- in the same spirit as they do `parse-x11-color`, add `parse-gdk-color`. If for some reason generating it is difficult, then manually add a `collornames-gdk` file (you have already parsed it manually).

Notice that all of the representation conversions, and color model conversions are already present. It's a tested and matured library, no need to reinvent the wheel.

Regarding `relative-luminance`, `contrast-ratio` and `contrasting-color`, I think they belong to `nyxt/themes` and not to `cl-colors2`. So, in my view, `nyxt/themes` is a library that depends on `cl-colors2` so that those 3 functions I mentioned earlier return irrespective of how the colors are passed (hex, RGB or color name).

Do that sound reasonable to you? @aartaka

aartaka:
Yes, it does sound perfectly reasonable! `cl-colors2` I haven't heard of, thanks for enlightening me :D

aadcg:
Beautiful. Sorry for some misunderstandings from my side. I didn't understand much about colors and how toolkits handle them. I thought that doing this GDK parsing would be like conjuring black magic...

aartaka:
I've opened a PR with GDK-parsed colors in `cl-colors2`: https://notabug.org/cage/cl-colors2/pulls/5/files

I'm also in the process of extracting color contrast utils there too.

libraries/theme/color-utils.lisp
```diff
@@ -229,14 +229,18 @@ Defaults to `parse-color'.")
                   (+ (relative-luminance color2) 0.05))))
     (max ratio (/ ratio))))
 
+(defun luminance (color)
+  "Compute absolute luminance of COLOR."
+  ;; See https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_Colors_and_Luminance#modeling_light_color_and_vision
+  (destructuring-bind (r g b)
+      (parse-color color)
+    (/ (+ (* .2126 r) (* .7152 g) (* .0722 b)) 255)))
```
aadcg, line 237: You perhaps added this after a comment I've made right? 

I actually have removed that comment, since I was incorrect.

aadcg:
> I've opened a PR with GDK-parsed colors in cl-colors2: https://notabug.org/cage/cl-colors2/pulls/5/files

If we don't get an answer whithin 2 weeks, then we should host a fork at github.com/atlas-engineer. Only afterwards should we package it in Nyxt and add the git submodule.

> I'm also in the process of extracting color contrast utils there too.

Are you suggesting that utils like `relative-luminance` belong to `cl-color2`? I thought you had agreed with what I've mentioned above:

> Regarding relative-luminance, contrast-ratio and contrasting-color, I think they belong to nyxt/themes and not to cl-colors2. So, in my view, nyxt/themes is a library that depends on cl-colors2 so that those 3 functions I mentioned earlier return irrespective of how the colors are passed (hex, RGB or color name).

aartaka:
I agree. I'm contributing most of this PR to `cl-colors2` and then getting back here.

aartaka:
Closing in favor of #2902.

aadcg:
@aartaka, when closing a PR, take into account that you need to delete the remote branch as well.

aartaka:
Right! I forgot it because GitHub deletes the branch automatically if merging, but, apparently, not when closing the PR :D

PR #2679, opened 2022-12-01
Fix nil browser

### LINKED ISSUE #2670: Always bind `*browser*`

Ambrevar:
# Description

Follow up to #2671 and #2666.

This ensures `*browser*` is always bound, thus removing a whole class of potential errors.

Fixes #2670.

# Discussion

Can we set `*browser*` at compile time and save it in the image?

# To do:

- [ ] Test!

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -48,7 +48,10 @@ Without handler, return ARG.  This is an acceptable `combination' for
 
 (define-class browser (renderer-browser)
   ((profile
-    (global-profile)
+    (or
+     (alex:when-let ((profile-class (find-profile-class (getf *options* :profile))))
+       (make-instance profile-class))
+     (make-instance 'nyxt-profile))
     :type nyxt-profile
     :documentation "Global profile used to specialize the behavior of
 various parts, such as the path of all data files.
@@ -258,10 +261,7 @@ prevents otherwise.")
   "Ensure `history-file' uses the browser profile."
   (setf (history-file browser) history-file))
 
-(defmethod theme ((ignored (eql nil)))
-  "Fallback theme in case there `*browser*' is NIL."
-  (declare (ignore ignored))
-  (make-instance 'theme:theme))
+(setf *browser* (make-instance 'browser))
 
 (defmethod external-editor-program ((browser browser))
   (alex:ensure-list (slot-value browser 'external-editor-program)))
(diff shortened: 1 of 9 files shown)


Ambrevar:
Turns out that instantiating a `browser` right away is not an easy task.
The reason is that `browser` has `nyxt-file` slots, which in turn needs to resolve a profile which is stored either in `*browser*`...  Hence the circularity.

We could break this circularity, but clearly the globality of the profile needs to be better designed somehow.

aartaka:
That's what I meant by side-effects browser instantialion implies! How about the lazy `(current-browser)` solution wrapping around something like `%browser%` special variable? Seems to work for buffers...

aartaka:
> That would not solve the issue: what would `(current-browser)` return at compilation time?

But do we need it at compile-time? Do we need it _that_ early? What's the use-case/bug/problem we're tryign to cover? Even if there's one, we may use `(current-browser)` and instantiate the browser there as well, but this time exactly when it's required to compute it, be it at run-time or compile-time.

> What we are trying to do here is instantiate a browser very early on.

Yes, so I'm suggesting to intantiate it lazily, which kind of implies that it may be initialized really early too, but without the downsides of being initialized _too early_.

Ambrevar:
`*browser*` is needed very early indeed, look at `*default-hostlist*` in blocker-mode: it needs to know the global profile, and thus depends on `*browser*`.

One fix would be to not have the global profile be resolve in browser but separately, like we used to.

So we could partially revert #2666.

aartaka:
> `*browser*` is needed very early indeed, look at `*default-hostlist*` in blocker-mode: it needs to know the global profile, and thus depends on `*browser*`.
> 
> One fix would be to not have the global profile be resolve in browser but separately, like we used to.
> 
> So we could partially revert #2666.

Yes, that sounds sanest.

aadcg:
What's the status here? This might [solve the blocker mentioned in this PR](https://github.com/atlas-engineer/nyxt/pull/2693#issuecomment-1352848266).

Ambrevar:
It's a tough one, so I need to find time to sit down and think through it.  It's high priority, obviously.

aadcg:
No worries @Ambrevar and please ignore my previous comment, since the PR I mentioned has been already merged!

jmercouris:
Please reopen when ready.

PR #2680, opened 2022-12-01
Move some build-script tests to nyxt-asdf

Ambrevar:
# Description

Follow up to #2305.

What's done:
- error out on compilation warning : now it's done _at all times_.
- unbound export check: moved to a regular nyxt-asdf test.
- manual generation: it's enough to call it from `nyxt/test`.

What's left out because ASDF cannot call itself recusively (need some different approach):
- "Validate make-instance symbols"
- "Keymap type checking"

# Discussion

What about the names?  In particular, `nyxt/tests/compilation` and `compilation-tests` does not seem too appropriate for unbound export checks.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:


aartaka:
> What's done:
> 
> * error out on compilation warning : now it's done _at all times_.
> * unbound export check: moved to a regular nyxt-asdf test.
> * manual generation: it's enough to call it from `nyxt/test`.

Cool!

> What's left out because ASDF cannot call itself recusively (need some different approach):
> 
> * "Validate make-instance symbols"

I may come around as out of context, but would it not be solved by this?
``` lisp
(defmethod asdf:perform :before ((op asdf:load-op) (system (eql (asdf:find-system :nyxt/tests/validate-make-instance))))
  (push :nyxt-debug-make-instance *features*))
```

> * "Keymap type checking"

Same here: just push a thing in `perform :before` method.

> # Discussion
> What about the names? In particular, `nyxt/tests/compilation` and `compilation-tests` does not seem too appropriate for unbound export checks.

`nyxt/tests/load-time` maybe?

Ambrevar:
I don't think it would work with a `before` method, because you'd need to reload the system.  And ASDF does not support calling itself recursively.

Ambrevar:
It's ready!  I'll fine-tune the documentation.
Otherwise OK to merge?  @aartaka

libraries/nyxt-asdf/tests.lisp
```diff
@@ -77,3 +77,23 @@ If the NYXT_TESTS_NO_NETWORK environment variable is set, tests with the `:onlin
       (format t "~a (~a sample~:p):~%" (first mark)
               (getf (rest (second mark)) :samples))
       (mapc #'print-times (rest mark)))))
+
+(defun redefinition-p (condition)       ; From Slynk.
+  (and (typep condition 'style-warning)
+       (every #'char-equal "redefin" (princ-to-string condition))))
```
aartaka, line 83: Maybe use `search` instead?

aartaka (approved):
All cool, yes!

nyxt.asd
```diff
@@ -229,11 +230,17 @@
...
+  :class :nyxt-compilation-test-system
+  :depends-on (nyxt)
+  :packages (:nyxt))
+
 ;; TODO: Test that Nyxt starts and that --help, --version work.
 (defsystem "nyxt/tests"
   :defsystem-depends-on ("nyxt-asdf")
   :class :nyxt-test-system
-  :depends-on (nyxt)
+  :depends-on (nyxt nyxt/tests/compilation nyxt/documentation)
```
aadcg, line 243: Commit 421e77acb7b2364925d4579d12b5d6a370d06099 feels a bit odd to me.

Wouldn't it be possible to add a `:in-order-to` to make the checks that the system `nyxt/tests/compilation` you introduced does?

In other words, I think it's wrong to add `nyxt/tests/compilation` as a dependency to systems. What you want to say is rather different: "for a given system, make these compilation checks".

Am I misinterpreting it?

libraries/nyxt-asdf/tests.lisp
Ambrevar: I wonder why Slynk wrote it this way...

nyxt.asd
Ambrevar: I don't think that would work, because `in-order-to` override the default method, here the `load-op`, with other method calls.

So it would look like this:

```lisp
  :in-order-to ((load-op (load-op "nyxt/documentation")
                         CALL-NEXT-METHOD))
```

The problem: What would you write instead of CALL-NEXT-METHOD?  I don't know if `in-order-to` can do that.

aadcg:
I've added a minor review to the readme.

nyxt.asd
Ambrevar: We've been bitten multiple times by `nyxt/documentation` not working because of actual bugs in Nyxt, so this has proven to be a precious test!  Hance its inclusion in the test suite.

Or would you rather have it tested some place else?

aadcg: Ok, I agree that this is a precious test.

Ideally, I believe that it should belong to the `:in-order-to` clause from the `nyxt` system. That way we could even get rid of the `nyxt/documentation` system. Smth along the lines of:

```lisp
(defsystem "nyxt"
  ...
  :in-order-to ((compile-op "manual genration logic")
                (test-op (test-op "nyxt/tests")
                         (test-op "nyxt/history-tree/tests")
                         (test-op "nyxt/class-star/tests")
                         (test-op "nyxt/prompter/tests"))))
```

My ASDF knowledge is rudimentary. Anyway, this is not very important.

aadcg:
From my perspective, it's ready to merge. 

Merging?

nyxt.asd
Ambrevar: That's a good point, will do.

aadcg: Beautiful! I like it :)

aadcg:
Merging then?

Ambrevar:
Done!

aadcg:
Thanks @Ambrevar :)

aartaka:
Yay, less bash and more ASDF! ＼（＠￣∇￣＠）／

Issue #1517, opened 2021-06-16
Feature: mouse hover to see overlong text on prompt-fields


pdelfino:
For instance, it is hard to read the `Docstring` info:

![image](https://user-images.githubusercontent.com/7698207/122287435-53857100-cec7-11eb-91d5-3f481a4d042f.png)

 Something like this would help:
![image](https://user-images.githubusercontent.com/7698207/122287530-6a2bc800-cec7-11eb-8c49-7bb6198c222c.png)

This could be useful for other prompt-fields, such as hint-descriptions. 

It is possible to find the information (e.g., what a command does), it just takes more actions, which hinders productivity.

I discussed this feature with "waleee" on IRC and he made great comments. I guess his user on Github  is @dtw-waleee

Ambrevar:
This is a good idea!

As a side note, you can press `C-]` to toggle the display of some columns, which can give more space to display the largest columns.

aartaka:
Fixed in 15f03f933459f996f4f065769b6a01d51a589b5e.

PR #2682, opened 2022-12-02
Auto-adjust the prompt listing to the content in it.

### LINKED ISSUE #2675: possibilities for Docstring element of the Execute Command Window
Ciao a tutti,
I notice that the Docstring element of the Command Execute Window has some cut off descriptions. Has anyone experimented in the past with it? 

I see a few potential solutions: (though I do not know what thought process has gone into what exists now)
​​​​​​​1. text wrapping
2. max character limits for descriptions.
3. re-writing lengthy descriptions to be more concise.

Thank you

docstring element

aartaka:
# Description

This adds auto-adjustment of suggestion listing to all prompt sources. It's convenient, because it allows to display all the content at once in the auto-laid-out table. The downsides are that:
- Prompt is jittering a lot when scrolling, because we create a new table on every scroll event.
- The columns can get terribly uneven, and one of those can occupy all the allocated space, so we have to take care to now put anything extremely long in those.

As an example of both these adjantages and disadvantages, open Nyxt with this patch and trigger `execute-command`.

Fixes #2675.

# Discussion

Maybe there are other ways to make it more intuitive of a display, without this jitter (it annoys me, yes). Maybe render all the table below the screen space? This way it's going to auto-adjust to a more of a median value, I guess...

CC @jmercouris, @aadcg, @lansingthomas

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -111,7 +111,7 @@ selection over prompt buffer suggestions.")
           :color ,theme:on-background
           :margin-left "16px"
           :width "100%"
-          :table-layout "fixed"
+          :table-layout auto ; So that content is laid-out in a smarter way.
           (td
            :white-space "nowrap"
            :height "20px"


Ambrevar:
I don't think #2675 is a problem, I would say it's a feature.

One thing that would be nice to have is the ability to define column width ratio on a per-source basis.

For truncated text, I propose to display it when hovering with the mouse.

Last, I would love to have the Helm feature where one can "dump" the prompt buffer into a regular buffer: in this dump, we could easily display the full content in a flexible table.

(In Helm try `C-x C-s` from helm-grep or helm-occur.)

Ambrevar:
Also the jittering is a big no-no in my opinion.

aartaka:
Truncated text is already displayed on hover (#1517) :D

But still, it's not fully obvious there, so having a more intuitive layout (preferably relying on WebKit layout smartness) won't hurt.

aartaka:
> One thing that would be nice to have is the ability to define column width ratio on a per-source basis.

Right! We can use the fourth element of the attribute there, given that attributes can have arbitrary number of elements and `prompter` is only concerned with the first two :P

> Last, I would love to have the Helm feature where one can "dump" the prompt buffer into a regular buffer: in this dump, we could easily display the full content in a flexible table.
> 
> (In Helm try `C-x C-s` from helm-grep or helm-occur.)

This one is a totally separate issue, and I suggest you open it as one :P

aadcg:
> I don't think #2675 is a problem, I would say it's a feature.

Absolutely.

> One thing that would be nice to have is the ability to define column width ratio on a per-source basis.

Agree. For instance, the width allocated for hint string in `hint-mode` is excessive.

aartaka:
See #2684. Closing this.

Ambrevar:
Yes!

aadcg:
Let's fix some terminology. Column jittering means having the columns of a table change their width.

The following questions arise. What kind of column jittering is bearable? None? Some? If some, could you specify it? I'm asking as to understand in what direction should #2686 move. Particularly interested in your views @Ambrevar.

Ambrevar:
In my opinion there should never be any column jittering.

aadcg:
@Ambrevar we're on the same page then.

aartaka:
But then, it's not noticeable/annoying if it's just by several millimeters |_・)

Ambrevar:
I still believe that it's heavily discomforting for most people, in terms of user experience.
At best it looks like a bug.  At worst, it makes everyone epileptic :(

aartaka:
Well, epilepsy argument does not apply, as our suggestions already change quite instantaneously with one's typing, so column movement is not changing the trigger-ness of the prompt much.

And what's discomforting for most people is not fast change of something, what's discomforting is the violation of expectations. Smooth change of column width (as in #2686, check it and see!) is not discomforting, as most of the columns don't move much and are still where one expects them to be. The change in #2682 (here) _is discomforting_, because the change in column width is so big and unpredictable that one easily loses track of where the columns were!

Given that #2686 (or, well, my idea/part of #2686, as @aadcg may've changed everything already and I'd have to test it again) adds a reasonable heuristic for column width calculation while having a bearable amount of column width changes that are mostly trackable and not violating one's expectations. It's an overall positive change, I believe.

Ambrevar:
@aartaka Your last comment may come across as offensive towards @aadcg.  Please, let's remain all peaceful on this, there is no need to start a flamewar on such trivial matters.

(Not that there ever is a need to start a flamewar... of course!)

PR #2686, opened 2022-12-02
prompt-buffer: Adjust prompt column widths based on content.

### LINKED ISSUE #2675: possibilities for Docstring element of the Execute Command Window

### LINKED ISSUE #2684: Prompt column width ratios
It would be nice to have columns to be adjustable in e.g. `prompter:object-attributes`, because it's more than natural that some columns are much bigger than the others.

Originally posted by @Ambrevar in https://github.com/atlas-engineer/nyxt/issues/2682#issuecomment-1334892685_

aartaka:
# Description

A second shot at #2682, this time looking at prompt content and deciding the size based on the average length of the content in the suggestion attributes. Here's how it looks:

adjusted-prompt-execute
adjusted-prompt-buffers

Fixes #2675.
Fixes #2684.

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -683,6 +683,13 @@ If the `active-attributes' slot is NIL, return all attributes keys."
           (cons (attributes-keys-default source)
                 (apply #'remove-from-seq (attributes-keys-non-default source) value)))))
 
+(export-always 'all-attribute-values)
+(defmethod all-attribute-values (attribute-key (source source))
+  "Return all ATTRIBUTE-KEY values from SOURCE."
+  (mapcar (alex:compose #'first
+                        (alex:rcurry #'str:s-assoc-value attribute-key))
+          (mapcar #'attributes (suggestions source))))
+
 (export-always 'active-attributes)
 (defmethod active-attributes ((suggestion suggestion)
                               &key (source (error "Source required"))
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -347,6 +347,13 @@ call."))
     (loop for pair in (call-next-method)
           for key = (first pair)
           for value = (second pair)
+          ;; NOTE: Duplicate keys are bad, because searching the alist by key
+          ;; will always return the first occurrence, and never the second.
+          when (member key keys :test #'string-equal)
+            do (warn "Duplicate attribute names found in ~a: ~a.
+Attribute names should be unique for prompter to correctly filter those."
+                     source key)
+          collect key into keys
           ;; FIXME: Having six (string-t for keys and string-function-t for
           ;; values) branches would be more correct, but does that matter enough
           ;; to bother?
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -296,13 +296,14 @@ See also `show-prompt-buffer'."
                "Scale each element of WIDTHS to ratios that sum up to one."
                (mapcar (rcurry #'/ (reduce #'+ widths)) widths)))
       (loop for key in (prompter:active-attributes-keys source)
-            for width = (funcall width-function
-                                 (cons key
-                                       ;; TODO Filter empty suggestions in
-                                       ;; all-attribute-values (add as a
-                                       ;; keyword). Let all-attribute-values
-                                       ;; accept a list of keys.
-                                       (prompter:all-attribute-values key source)))
+            ;; Ensures that the Attribute key string fits into the column.
+            for width = (max (length key)
+                             (funcall width-function
+                                      ;; TODO Filter empty suggestions in
+                                      ;; all-attribute-values (add as a
+                                      ;; keyword). Let all-attribute-values
+                                      ;; accept a list of keys.
+                                      (prompter:all-attribute-values key source)))
             collect width into widths
             ;; Attributes keys set to empty strings aren't allowed, so widths is
             ;; never a list whose elements are all zeros.  In such a scenario,
@@ -327,6 +328,9 @@ an integer."))
     (when (prompter:suggestions source)
       (:table :class "source-content"
               (:colgroup
+               ;; This gets called everytime the prompt-buffer is redrawn, not
+               ;; good.
+               ;; What should be called are the ratio value from the attributes.
                (dolist (width (attribute-widths source))
                  (:col :style (format nil "width: ~,2f%" (* 100 width)))))
               (:tr :style (if (or (eq (prompter:hide-attribute-header-p source) :always)
@@ -393,6 +397,9 @@ This does not redraw the whole prompt buffer, unlike `prompt-render'."
                            (if (prompter:ready-p source)
                                ""
                                "(In progress...)"))
+                     ;; render-attributes, which calls attribute-widths, is
+                     ;; called many times (see that source->html gets called in
+                     ;; a loop)
                      (:raw (render-attributes source prompt-buffer))))))
       (ps-eval :buffer prompt-buffer
         (setf (ps:@ (nyxt/ps:qs document "#suggestions") |innerHTML|)
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -304,12 +304,22 @@ See also `show-prompt-buffer'."
                                        ;; accept a list of keys.
                                        (prompter:all-attribute-values key source)))
             collect width into widths
-  (:documentation "Compute the widths of SOURCE attribute columns (as percent).
-            finally (return (clip-extremes (width-normalization widths))))))
-Returns a list of ratios that sum up to one.
-Uses the WIDTH-FUNCTION (by default computing average length of non-blank
-attribute values) to derive the width of the list of attribute
-values. WIDTH-FUNCTION is a function accepting a list of strings and returning
+            ;; Attributes keys set to empty strings aren't allowed, so widths is
+            ;; never a list whose elements are all zeros.  In such a scenario,
+            ;; width-normalization would error with DIVISION-BY-ZERO.
+            finally (return (width-normalization widths)))))
+  (:documentation "Return a list of unit ratio representative of the length of each
+of SOURCE's attributes in terms of number of characters.
+
+The sum of all ratios equals one.
+
+Useful to set column width for each of SOURCE's attributes. The ratios sum to
+one.
+
+Uses WIDTH-FUNCTION to derive the representative width of the list of attribute
+values.
+;; WIDTH-FUNCTION should a rations accepting a list of numbers returning a number.
+WIDTH-FUNCTION is a function accepting a list of strings and returning
 an integer."))
 
 (defun render-attributes (source prompt-buffer)
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -310,11 +310,9 @@ See also `show-prompt-buffer'."
                                 into new-widths
                             finally (return new-widths))))
                    (t widths))))
-             (ratio-widths (widths)
-               "Compute the percentage of the screen that attributes with WIDTHS should occupy."
-               (let* ((total (reduce #'+ widths)))
-                 (mapcar (lambda (w) (if (zerop total) 0 (/ w total)))
-                         widths))))
+             (width-normalization (width-list)
+               (let ((total (reduce #'+ widths)))
+                 (mapcar (lambda (w) (/ w total)) widths))))
       ;; FIXME: This loop ignores attribute names in the attribute width
       ;; computation. Because of this, attribute names could theoretically get
       ;; cropped if the values are short enough. This is bad, but not exactly
@@ -326,14 +324,14 @@ See also `show-prompt-buffer'."
                                        (mapcar #'fourth (prompter:active-attributes (first suggestions) :source source)))
                                       (some-fourth (some #'identity fourth-attributes))
                                       (coefficients (substitute 1 nil fourth-attributes)))
-                        (return (ratio-widths coefficients)))
+                        (return (width-normalization coefficients)))
             for key in (prompter:active-attributes-keys source)
             for width
               = (funcall width-function (mapcar (compose #'first (rcurry #'str:s-assoc-value key))
                                                 attributes))
             collect width into widths
-            finally (return (clip-extremes (ratio-widths widths))))))
   (:documentation "Compute the widths of SOURCE attribute columns (as percent).
+            finally (return (clip-extremes (width-normalization widths))))))
 Returns a list of ratios that sum up to one.
 Uses the WIDTH-FUNCTION (by default computing average length of non-blank
 attribute values) to derive the width of the list of attribute
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -312,25 +312,13 @@ See also `show-prompt-buffer'."
                             finally (return new-widths))))
                    (t widths))))
              (width-normalization (width-list)
-      ;; FIXME: This loop ignores attribute names in the attribute width
-      ;; computation. Because of this, attribute names could theoretically get
-      ;; cropped if the values are short enough. This is bad, but not exactly
-      ;; critical.
-      (loop with suggestions = (prompter:suggestions source)
-            with attributes = (mapcar (rcurry #'prompter:active-attributes :source source) suggestions)
-            ;; This is to process column width as fourth object attribute element.
-            initially (sera:and-let* ((fourth-attributes
-                                       (mapcar #'fourth (prompter:active-attributes (first suggestions) :source source)))
-                                      (some-fourth (some #'identity fourth-attributes))
-                                      (coefficients (substitute 1 nil fourth-attributes)))
-                        (return (width-normalization coefficients)))
-            for key in (prompter:active-attributes-keys source)
-            for width
-              = (funcall width-function (mapcar (compose #'first (rcurry #'str:s-assoc-value key))
-                                                attributes))
                ;; this should not eval to (/ 0 0) by construction
                (let ((total (reduce #'+ width-list)))
                  (mapcar (lambda (w) (/ w total)) width-list))))
+      (loop for key in (prompter:active-attributes-keys source)
+            for width = (funcall width-function
+                                 (cons key
+                                       (prompter:all-attribute-values key source)))
             collect width into widths
   (:documentation "Compute the widths of SOURCE attribute columns (as percent).
             finally (return (clip-extremes (width-normalization widths))))))
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -283,20 +283,29 @@ See also `show-prompt-buffer'."
   "Compute the widths of SOURCE attribute columns (as percent).
 Returns a list of integers, the sum of which should be roughly equal to 100.
 Uses the average length of attribute values to derive the width."
-  (loop with suggestions = (prompter:suggestions source)
-        with mandatory-space = (* 10 (length (prompter:active-attributes-keys source)))
-        for key in (prompter:active-attributes-keys source)
-        for width
-          = (loop for suggestion in suggestions
-                  sum (length (first (str:s-assoc-value
-                                      (prompter:active-attributes suggestion :source source) key))))
-        collect width into widths
-        sum width into total
-        finally (return (mapcar (lambda (w) (+ 10 (round (* (- 100 mandatory-space)
-                                                            (if (zerop total)
-                                                                0
-                                                                (/ w total))))))
-                                widths))))
+  (flet ((ratio-widths (widths total mandatory-space)
+           (mapcar (lambda (w) (+ 10 (round (* (- 100 mandatory-space)
+                                               (if (zerop total)
+                                                   0
+                                                   (/ w total))))))
+                   widths)))
+    (loop with suggestions = (prompter:suggestions source)
+          with mandatory-space = (* 10 (length (prompter:active-attributes-keys source)))
+          ;; This is to process column width as fourth object attribute element.
+          initially (sera:and-let* ((fourth-attributes
+                                     (mapcar #'fourth (prompter:active-attributes (first suggestions) :source source)))
+                                    (some-fourth (some #'identity fourth-attributes))
+                                    (coefficients (substitute 1 nil fourth-attributes))
+                                    (total (reduce #'+ coefficients)))
+                      (return (ratio-widths coefficients total mandatory-space)))
+          for key in (prompter:active-attributes-keys source)
+          for width
+            = (loop for suggestion in suggestions
+                    sum (length (first (str:s-assoc-value
+                                        (prompter:active-attributes suggestion :source source) key))))
+          collect width into widths
+          sum width into total
+          finally (return (ratio-widths widths total mandatory-space)))))
 
 (defun render-attributes (source prompt-buffer)
   (spinneret:with-html-string
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -312,8 +312,6 @@ See also `show-prompt-buffer'."
                             finally (return new-widths))))
                    (t widths))))
              (width-normalization (width-list)
-               (let ((total (reduce #'+ widths)))
-                 (mapcar (lambda (w) (/ w total)) widths))))
       ;; FIXME: This loop ignores attribute names in the attribute width
       ;; computation. Because of this, attribute names could theoretically get
       ;; cropped if the values are short enough. This is bad, but not exactly
@@ -331,6 +329,8 @@ See also `show-prompt-buffer'."
               = (funcall width-function (mapcar (compose #'first (rcurry #'str:s-assoc-value key))
                                                 attributes))
                ;; this should not eval to (/ 0 0) by construction
+               (let ((total (reduce #'+ width-list)))
+                 (mapcar (lambda (w) (/ w total)) width-list))))
             collect width into widths
   (:documentation "Compute the widths of SOURCE attribute columns (as percent).
             finally (return (clip-extremes (width-normalization widths))))))
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -288,8 +288,10 @@ See also `show-prompt-buffer'."
   ;; to the empty string, and therefore
   ;; (mapcar #'length (remove-if #'str:blankp attribute-values))
   ;; can be NIL.
-  (alex:median (uiop:ensure-list (mapcar #'length
-                                         (remove-if #'str:blankp attribute-values)))))
+  (alex:if-let ((non-blank-lengths
+                 (uiop:ensure-list (mapcar #'length (remove-if #'str:blankp attribute-values)))))
+    (alex:median non-blank-lengths)
+    0))
 
 (defgeneric attribute-widths (source &key width-function)
   (:method (source &key (width-function #'average-attribute-width))
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -280,7 +280,11 @@ See also `show-prompt-buffer'."
                                        (sort-modes-for-status (modes prompt-buffer)))))))))
 
 (defun attribute-widths (source)
-  (loop for key in (prompter:active-attributes-keys source)
+  "Compute the widths of SOURCE attribute columns (as percent).
+Returns a list of integers, the sum of which should be roughly equal to 100.
+Uses the average length of attribute values to derive the width."
+  (loop with suggestions = (prompter:suggestions source)
+        for key in (prompter:active-attributes-keys source)
         for width
           = (loop for suggestion in suggestions
                   sum (length (second (assoc key (prompter:active-attributes suggestion :source source)
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -294,6 +294,7 @@ Uses the average length of attribute values to derive the width."
                                                      (/* w total))))))
                      widths)))
     (loop with suggestions = (prompter:suggestions source)
+          with attributes = (mapcar (rcurry #'prompter:active-attributes :source source) suggestions)
           ;; This is to process column width as fourth object attribute element.
           initially (sera:and-let* ((fourth-attributes
                                      (mapcar #'fourth (prompter:active-attributes (first suggestions) :source source)))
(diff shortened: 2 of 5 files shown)


aadcg:
I'll take a deeper look next week. Thanks @aartaka.

Note to self: test performance in extreme conditions (for instance, `describe-any`).

aartaka:
> Note to self: test performance in extreme conditions (for instance, `describe-any`).

It should be as good as it used to be, because all the attributes are computed in advance, and there's only one additional loop processing all of the attributes. The slowdown should be negligible.

source/prompt-buffer.lisp
```diff
@@ -279,6 +279,22 @@ See also `show-prompt-buffer'."
...
+(defun attribute-widths (source)
+  "Compute the widths of SOURCE attribute columns (as percent).
+Returns a list of integers, the sum of which should be roughly equal to 100.
+Uses the average length of attribute values to derive the width."
+  (loop with suggestions = (prompter:suggestions source)
+        with mandatory-space = (* 10 (length (prompter:active-attributes-keys source)))
+        for key in (prompter:active-attributes-keys source)
+        for width
+          = (loop for suggestion in suggestions
+                  sum (length (second (assoc key (prompter:active-attributes suggestion :source source)
```
aadcg, line 291: I'd suggest using `str:s-assoc-value` instead of `assoc` here. 

Unfortunately `(prompter:active-attributes suggestions :source source)` doesn't returned a list of dotted pairs. If it would, there wouldn't be the need to call `first` after `str:s-assoc-value`.

aadcg:
```lisp
(str:s-assoc-value '(("Name" . "my-mode") ("Bindings" . "C-x p")) "Bindings")
=>
"C-x p"
("Bindings" . "C-x p")
```

aartaka:
#2684 is covered with c3331d5f91bbf4302161d389c6a7019764791793.

aadcg:
This is a welcome change.

Regarding the implementation, I think it would be wiser to compute the median, not the average.

For each of the N attributes, uniformly allocate 100/N width to each of them (let's call it the default width per attribute). Then, compute the median for each attribute. Two situations can happen: either the combined computed median width falls above or below 100. If it falls below, then redistribute uniformly over all attributes. Otherwise, deduct uniformly over the clipping attributes. Clipping attributes are those whose computed median width exceeds the default width.

Example 1: 4 attributes - A, B, C and D. The median is 10, 40, 50 and 5 (which amounts to 105). The clipping attributes are B and C. Therefore the final width will be 10, 37.5, 47.5, 5.

Example 2: 3 attributes - A, B and C. The median is 10, 40, 20 and 5 (which amounts to 75). The final width will be 16.25, 46.25, 26.25 and 11.25.

aartaka:
> Regarding the implementation, I think it would be wiser to compute the median, not the average.

But then, we might have lots of empty attribute values (think "Documentation" attribute in `describe-*` sources). To the extent that those might get to the median. Average is better here.

> For each of the N attributes, uniformly allocate 100/N width to each of them (let's call it the default width per attribute). Then, compute the median for each attribute. Two situations can happen: either the combined computed median width falls above or below 100. If it falls below, then redistribute uniformly over all attributes. Otherwise, deduct uniformly over the clipping attributes. Clipping attributes are those whose computed median width exceeds the default width.

Clipping I wanted to do too, but having this base-size and adding average is much easier to implement and seems to work just fine too.

aartaka:
Note that the extreme examples from the screenshots in the initial comment are not as extreme as they used to be, given that I allocate at least 10% for every attribute now.

aadcg:
> But then, we might have lots of empty attribute values (think "Documentation" attribute in describe-* sources). To the extent that those might get to the median. Average is better here.

That's exactly why the median is superior to the average. Extreme values have less of an effect.

aartaka:
> That's exactly why the median is superior to the average. Extreme values have less of an effect.

But still, median is just one value in the middle of the distribution, not accounting for anything on either side...

> Empty attributes should be discarded.

Yes, this I'm working on.

aadcg:
Since prompt buffer is the apple of @Ambrevar's eyes, perhaps we should wait for him so that he can take a final look.

I'd suggest:
- Clearly separate the width computation of a single attribute, so that the width computation for a source is more intelligible;
- The method that heuristically computes width attribute should be an argument with a default (right now, it's mean but it's not possible to abstract it);
- Add basic tests with mock sources and attributes, capturing the rationale.

aartaka:
> Since prompt buffer is the apple of @Ambrevar's eyes, perhaps we should wait for him so that he can take a final look.

Well, it's mostly a UI-related change, not an architectural one. But let it be this way :)

> * Clearly separate the width computation of a single attribute, so that the width computation for a source is more intelligible;

It won't become more intelligeble, because attributes are scattered over suggestions and are reliably not accessible in any way but `prompter:active-attributes` of every single `suggestion`. Fetching only the first attribute, then only the second one, calling `prompter:active-attributes` every time etc... is likely to make the algorithm more complex/slow because of the data it operates on being re-shuffled in a totally new structure.

I mean, it could be the problem with me not having an intuitive enough understanding/API for `prompter` that this code is born from. But given that attributes are `suggestion`-specific and are an undotted alist, it's more than understandable that trying to transpose those ends up being quite a fun of a problem.

> * The method that heuristically computes width attribute should be an argument with a default (right now, it's mean but it's not possible to abstract it);

Given that the heuristic involves the whole set of attributes to compute the average, it's only replaceable at the level of `attribute-widths` as a whole.

> * Add basic tests with mock sources and attributes, capturing the rationale.

Yes!

aadcg:
> It won't become more intelligeble, because attributes are scattered over suggestions and are reliably not accessible in any way but prompter:active-attributes of every single suggestion. Fetching only the first attribute, then only the second one, calling prompter:active-attributes every time etc... is likely to make the algorithm more complex/slow because of the data it operates on being re-shuffled in a totally new structure.

I think I was not very clear. My point is that the outer loop of width computation traverses each of the attributes, for a given source. The meat of the computation is in each of the attributes. I'm just suggesting to capture that meat in a separate place. After all, for a given source, all suggestions have the same attributes isn't it so?

aadcg:
>  Given that the heuristic involves the whole set of attributes to compute the average, it's only replaceable at the level of attribute-widths as a whole.

I'm not sure I follow. 

The heuristic to arrive at an estimation of the typical length for a column attribute is independent from all others. Of course that that does not provide the final width for a given attribute, but that's a separate operation. This is important since it makes it more flexible.  Say that I'm a user who disable the "Docstrings" attribute for all prompts. Then I'd probably go with `max` to define the width for my attributes.

aartaka:
Fair, I've abstracted it and made `attribute-widths` a generic so that one can override it with `:around` method or re-define it for a certain source.

aartaka:
@aadcg, I've gone with a mixture of yours&mine in the algorithm for width computation:
- It computes average width (that's adjustable with `&key width-function`).
- Then, if there are any widths lover than 100/N/2 (because 100/N is too big for a minimal size), then
  - substract from those above this minimal size proportional to them being more than this minimal size,
  - and complete the lacking ones to the minimal size.
- If the widths were provided by the `prompter:object-attributes` writer or don't have any widths below minimal (even if it's some `(13 13 13 61)`), then leave it as is.

aadcg:
I'm taking a deeper look at this right now.

aadcg:
Let's take the case when, for each attribute, all suggestions have the same length.

The results are in the table below. @aartaka do you find it satisfactory?

```
| #Attributes | Width per attributes list |
|           1 | (100)                     |
|           2 | (33 67)                   |
|           3 | (17 33 50)                |
|           4 | (12 20 29 39)             |
|           5 | (10 13 19 26 32)          |
```

aadcg:
Sorry, I had to delete a comment I made.

aadcg:
Two observations thus far:

- Why `round`ing the results?
- You seem to have forgotten the fact that the length of the attribute name (the title bar at the prompt buffer's bar) also matters while computing the width.

aadcg:
Could you explain what you meant by:

> because 100/N is too big for a minimal size

aadcg:
> Why rounding the results? CL is arguably one of the most elegant languages in the way it handles rational numbers.

In that spirit, I'd suggest that the total width would sum to 1, not 100.

I'm not sure if I agree with the commit that allows setting the relative width as an attribute. I don't see much profit in it, and I don't like the idea of adding more parameters to the attributes structure - `("Name" (transform object))`.


I've added some tests that generalize the one you wrote and they should some cracks in your implementation. See bca37209d.

aartaka:
> * Why `round`ing the results? CL is arguably one of the most elegant languages in the way it handles rational numbers.

Yes, CL is. But CSS is not, and the main purpose of these widths is to be inserted verbatim into CSS code. I may try making it into floats, but rationals are a big no if we're to interop with CSS.

> * You seem to have forgotten the fact that the length of the attribute name (the title bar at the prompt buffer's bar) also matters while computing the width.

Yes, I did forget that. But, more often than not, attribute name is less than attribute contents and/or less important than the content. So it should not be much of a problem there.

aartaka:
> Could you rephrase? In particular I don't understand what you meant by "because 100/N is too big for a minimal size" below:
> 
> > ```
> > * Then, if there are any widths lover than 100/N/2 (because 100/N is too big for a minimal size), then
> >   
> >   * substract from those above this minimal size proportional to them being more than this minimal size,
> >   * and complete the lacking ones to the minimal size.
> > ```

Yes, sure. So, we have a total of 100% of the space. If we set the minimal/default bound to 100/N to each attribute, we'd have no space for maneuvres at all—all the space is already distributed evenly. Subtracting while being aware of all the attribute widths is a pain, so the algorithm that would distribute 100/N first and then re-distribute the widths will likely end up much longer and much less readable than the current one.

If we take 100/N/2 of the space per every attribute, we'll have 50% left for distribution between the attributes and a reasonable minimal width for an attribute (just my observation—100/N is the most reasonable uniform width, while 100/N/2 is twice as small, while still maintaining most of the visibility).

> Does this mean that you set a lower bound on the width percentage per attribute equal to `100/N/2` where `N` is the number of attributes?

Yes.

aartaka:
> > Why rounding the results? CL is arguably one of the most elegant languages in the way it handles rational numbers.
> 
> In that spirit, I'd suggest that the total width would sum to 1, not 100.

While it definitely would be elegant and mathy and correct, if would imply yet another round of conversion for CSS generation. I don't see much value in this change, given that both models (1.0 vs. 100%) are relatively equal in the explanatory power.

> I'm not sure if I agree with the commit that allows setting the relative width as an attribute. I don't see much profit in it, and I don't like the idea of adding more parameters to the attributes structure - `("Attribute Name" (transform object))`.

I don't like it either, but it's often is a necessity and our attribute structures already allow third attribute (per #2652), so why not add the fourth one and make it more configurable?

> I've added some tests that generalize the one you wrote and they show some cracks in your implementation. See [37a9574](https://github.com/atlas-engineer/nyxt/commit/37a9574aaef375b99820559efb93d696a851b793). You can run the failing test with `(with-summary () (run-test 'ratio-between-two-attributes))`.

The algorithm works fine under the assumption that attribute names are unique relative to each other. Which your test attributes are not. I've fixed it. Note that these tests also fail because of ratio/integer inconsistency, but that I've listed rationale for, so unless there's a dire necessity in rationals, I'd leave it to you to rewrite the ratios part of the test :)

aadcg:
>  Yes, CL is. But CSS is not, and the main purpose of these widths is to be inserted verbatim into CSS code. I may try making it into floats, but rationals are a big no if we're to interop with CSS.

That's what I imagined, but then the rounding only needs to happen in the final moment (i.e. when you pass it to CSS), right? `attribute-widths`, in the CL world, doesn't need any rounding.

aadcg:
> Yes, I did forget that. But, more often than not, attribute name is less than attribute contents and/or less important than the content. So it should not be much of a problem there.

It is still a bug, so if we're not fixing it, we should at least add a note on why we ignore it.

aartaka:
> > Yes, I did forget that. But, more often than not, attribute name is less than attribute contents and/or less important than the content. So it should not be much of a problem there.
> 
> It is still a bug, so if we're not fixing it, we should at least add a note on why we ignore it.

2211f30fe01955f2f46e3378ea90f1ec1872ffe5.

aartaka:
Sorry, 2211f30fe01955f2f46e3378ea90f1ec1872ffe5 fixes a different model shortcoming (⌒_⌒;)

9d2f01249a82e46ee8999c24d63d18f72ca25189 adds the comment about attribute name width not being considered.

aartaka:
> > Yes, CL is. But CSS is not, and the main purpose of these widths is to be inserted verbatim into CSS code. I may try making it into floats, but rationals are a big no if we're to interop with CSS.
> 
> That's what I imagined, but then the rounding only needs to happen in the final moment (i.e. when you pass it to CSS), right? `attribute-widths`, in the CL world, doesn't need any rounding.

That is fair indeed. And would likely produce much more correct of a result set.

aartaka:
> > I don't like it either, but it's often is a necessity and our attribute structures already allow third attribute (per #2652), so why not add the fourth one and make it more configurable?
> 
> I now realize I should have said something about that! I think we're adding more structure in the wrong place. There should be another data structure to keep those case-specific pieces of data. Regardless, that's another subject.

No, this data belongs to attributes, because:
- Attributes are responsible for the display of suggestions, that's their main purpose.
- `prompter` allows additional attribute values for application specific information passing. It already is a structure dedicated for display of suggestions, we just have to breathe Nyxt-specific meaning into it.
- HTML display as a third attribute value allows to fine-tune the display of the single suggestion value.
- Column width as the fourth value allows to fine-tune the area display happens on.
- There could be more, like changing the order of columns (think flexbox CSS attributes) or the highlight style. The space is unexplored.

> The most important point is: is setting the relative width so important? I don't think it is.

It may be important, because we have a preemptive knowledge about the content of attributes and we need to make the most informative display possible for it. Adjusting the amount of space certain attribute gets may make it much more informative.

aadcg:
> How wildly does it jump, exactly? Because I have a feeling that selecting a reasonably big subset from the suggestion set is likely to produce average widths similar to the set-total average...

As wildly fast as you type I guess!

aadcg:
Let me take this from here. I'll let you know when to take a look.

libraries/prompter/prompter-source.lisp
```diff
@@ -347,6 +347,13 @@ call."))
     (loop for pair in (call-next-method)
           for key = (first pair)
           for value = (second pair)
```
Ambrevar, line 349: These 3 lines could be replaced by `(loog for (key val) in (call-next-method)`

libraries/prompter/prompter-source.lisp
```diff
@@ -347,6 +347,13 @@ call."))
...
+          ;; NOTE: Duplicate keys are bad, because searching the alist by key
+          ;; will always return the first occurrence, and never the second.
+          when (member key keys :test #'string-equal)
+            do (warn "Duplicate attribute names found in ~a: ~a.
+Attribute names should be unique for prompter to correctly filter those."
+                     source key)
+          collect key into keys
```
Ambrevar, line 356: Good point, but is there a way to perform this check only once?  Because this is a relatively tight loop.

source/prompt-buffer.lisp
```diff
@@ -279,14 +279,107 @@ See also `show-prompt-buffer'."
                                (mapcar (curry #'mode-status status-buffer)
                                        (sort-modes-for-status (modes prompt-buffer)))))))))
 
+(defun average-attribute-width (attribute-values)
+  (let* ((values (remove-if #'str:blankp attribute-values))
+         (total (reduce #'+ values :key #'length)))
+    ;; this should not eval to (/ 0 0) by construction
+    (if (zerop (length values))
+        0
+        (/ total (length values)))))
```
Ambrevar, line 288: Note that Alexandria has `mean` and `median`, maybe they could be useful here.

source/prompt-buffer.lisp
```diff
@@ -279,14 +279,107 @@ See also `show-prompt-buffer'."
...
+
+(defgeneric attribute-widths (source &key width-function)
+  (:method (source &key (width-function #'average-attribute-width))
+    (labels ((clip-extremes (widths)
+               (let* ((length (length widths))
+                      (minimal-size (/ 1 length 2)))
+                 (cond
+                   ((= 1 length)
```
Ambrevar, line 296: `(sera:single widths)` maybe?

source/prompt-buffer.lisp
```diff
@@ -279,14 +279,107 @@ See also `show-prompt-buffer'."
...
+                    (list 1))
+                   ((some (rcurry #'< minimal-size) widths)
+                    (let* ((lacking (remove-if-not #'plusp
+                                                   (mapcar (curry #'- minimal-size) widths)))
+                           (abundant (set-difference widths lacking))
+                           (lack (reduce #'+ lacking)))
```
Ambrevar, line 302: You could use `sera:partition` here.

source/prompt-buffer.lisp
```diff
@@ -279,14 +279,107 @@ See also `show-prompt-buffer'."
...
+               ;; this should not eval to (/ 0 0) by construction
+               (let ((total (reduce #'+ width-list)))
+                 (mapcar (lambda (w) (/ w total)) width-list))))
+      (loop for key in (prompter:active-attributes-keys source)
+            for width = (funcall width-function
+                                 (cons key
+                                       (prompter:all-attribute-values key source)))
+            collect width into widths
+  (:documentation "Compute the widths of SOURCE attribute columns (as percent).
+            finally (return (clip-extremes (width-normalization widths))))))
```
Ambrevar, line 324: Looks like you are not returning `clip-extremes (width-normalization ...)`?  Looks like they are never called :p

tests/offline/prompt-buffer.lisp
```diff
@@ -0,0 +1,46 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt/tests)
+
+(unless lparallel:*kernel* (setf lparallel:*kernel*
+                                 (lparallel:make-kernel (or (serapeum:count-cpus) 1))))
```
Ambrevar, line 7: This should probably go in the tests/package.lisp file.  We do this for `tests/benchmark-package.lisp` too.

tests/offline/prompt-buffer.lisp
```diff
@@ -0,0 +1,46 @@
...
+                           result)))
+        ((eq index (length lst)) (nreverse result)))))
+
+;; TODO Add to lisp-unit2.
+(defun random-in-range (start end)
+  (+ start (random (+ 1 (- end start)))))
+
+(define-test constant-size-attributes ()
+  (let ((n-attributes (random-in-range 1 10)))
+    ;; How to shup up the fact that the method is being redefined?
```
Ambrevar, line 33: You can use `c2mop:ensure-method`.

tests/offline/prompt-buffer.lisp
```diff
@@ -0,0 +1,46 @@
...
+    (set-test-source-object-attributes (make-list n-attributes :initial-element 1))
+    (assert-equal (make-list n-attributes :initial-element (/ 1 n-attributes))
+                  (nyxt::attribute-widths (make-instance 'test-source)))))
+
+(define-test ratio-between-two-attributes ()
+  (let ((ratio (random-in-range 1 4)))
```
Ambrevar, line 39: Specific reason why `4`?

Ambrevar:
It's a bit difficult to follow what's happening with all the list manipulations (see the tests for instance).  Maybe use a class?

aadcg:
@Ambrevar I should have told you not to look at this yet, since it's an intermediary state between how @aartaka left it and how I've picked it up.

tests/offline/prompt-buffer.lisp
```diff
@@ -0,0 +1,46 @@
...
+  "Set `prompter:object-attributes' following LST properties:
+Its length LST dictates the number of attributes.
+Each element dictates the number of times a suggestion is concatenated."
+  (defmethod prompter:object-attributes ((object string) (source test-source))
+    (do ((index 0 (1+ index))
+         (result '() (cons `(,(str:repeat (nth index lst)
+                                (str:concat "Foo" (princ-to-string index)))
+                             ,(str:repeat (nth index lst) object))
+                           result)))
+        ((eq index (length lst)) (nreverse result)))))
```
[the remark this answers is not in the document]
aadcg: No, because `dotimes` is used for side-effects. I'm building a list.

tests/offline/prompt-buffer.lisp
aadcg: Agree.

libraries/prompter/prompter-source.lisp
aadcg: Actually I don't understand why we need to filter duplicate attribute names.

tests/offline/prompt-buffer.lisp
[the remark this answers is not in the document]
Ambrevar: You can also built a list with dotimes:

```lisp
(let ((result '())
  (dotimes (_ (length lst) result)
    (push ... result)))
```

libraries/prompter/prompter-source.lisp
aartaka: > Good point, but is there a way to perform this check only once? Because this is a relatively tight loop.

This loop only happens once per attribute computation. I guess there's not way to make it even more optimal, because checking it while setting attributes would be even more resource-hungry...

aartaka: > Actually I don't understand why we need to filter duplicate attribute names.

Because we use alist-ish structure and `assoc`more than once when working with attributes. And `assoc` only gets the first pair that starts with the given key. Thus we'll always ignore the second eponymous attribute unconditionally, as it happened in the auto-generated tests for attribute widths.

tests/offline/prompt-buffer.lisp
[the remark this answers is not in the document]
aartaka: Maybe a 
```
(loop for index below (length ...)
     for ... in ...
     collect (list ...))
```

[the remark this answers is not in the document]
aadcg: I'll see which variant is the most readable.

aadcg:
> For each of the N attributes, uniformly allocate 100/N width to each of them (let's call it the default width per attribute). Then, compute the median for each attribute. Two situations can happen: either the combined computed median width falls above or below 100. If it falls below, then redistribute uniformly over all attributes. Otherwise, deduct uniformly over the clipping attributes. Clipping attributes are those whose computed median width exceeds the default width.

> Example 1: 4 attributes - A, B, C and D. The median is 10, 40, 50 and 5 (which amounts to 105). The clipping attributes are B and C. Therefore the final width will be 10, 37.5, 47.5, 5.

I have now realized that my suggestion is extremely silly. Why should clipping attributes be favored when distributing the excess? Being below the default width (as probed by `mean`, `median` or other measure) doesn't mean that values of a given attribute key fit in the column width for all suggestions...

aadcg:
It's even worse than that. I've made this way more complicated than it is.

> For each of the N attributes, uniformly allocate 100/N width to each of them (let's call it the default width per attribute). Then, compute the median for each attribute. Two situations can happen: either the combined computed median width falls above or below 100.

This is absolute bullshit. All there is to do is extremely simple. For each attribute, take all suggestions and  come up with a representative width (via `median`, `mean`, etc). Then it's just a matter of normalizing the results. 

Example. Attributes A, B, C and D whose "typical" width is 20, 30, 55 and 90 (measured in numbers of characters). It's trivial to normalize that data so that the sum of all ratios equals 1.

aartaka:
> It's even worse than that. I've made this way more complicated than it is.
> 
> > For each of the N attributes, uniformly allocate 100/N width to each of them (let's call it the default width per attribute). Then, compute the median for each attribute. Two situations can happen: either the combined computed median width falls above or below 100.
> 
> This is absolute bullshit. All there is to do is extremely simple. For each attribute, take all suggestions and come up with a representative width (via `median`, `mean`, etc). Then it's just a matter of normalizing the results.
> 
> Example. Attributes A, B, C and D whose representative width is 20, 30, 55 and 90 (measured in numbers of characters). It's trivial to normalize that data so that the sum of all ratios equals 1.

That's exactly what I initially did, and it made some attributes shrink into non-existence. Mere normalization without clipping distribution is not enough. The implementation I ended up with is the sanest heuristic, as it gives us both sane minimal size for narrow attributes and relative content-based width for wide ones.

aadcg:
After talking to @jmercouris, we came to the conclusion that the key aspect regarding this feature is getting the amount of column jittering (if any).

Please run Nyxt from this branch as it stands now. Don't look at the code. Just use the prompt buffer extensively, with as many different sources and you can, change its input, scroll through suggestions, etc. 

Let me know if this amount of columns jittering is acceptable or not. I'm concealing my opinion for now :) 

@aartaka @Ambrevar @jmercouris

source/prompt-buffer.lisp
```diff
@@ -393,6 +398,9 @@ This does not redraw the whole prompt buffer, unlike `prompt-render'."
                            (if (prompter:ready-p source)
                                ""
                                "(In progress...)"))
+                     ;; render-attributes, which calls attribute-widths, is
```
aartaka, line 401: But it's called once per source, so it's the same as it used to be with per-source loop, just a bit more abstracted!

aadcg:
Following @Ambrevar's [comment](https://github.com/atlas-engineer/nyxt/pull/2682#issuecomment-1353131515), I will soon push a draft with the adjustment without column jittering.

aartaka:
@aadcg, the last set of commits (since 05a08b0b0ad681410a3481cc1a8fc31ab05f6b12) is a regression to me, as:
1. The case of empty attribute suggestions is not handled and errors out. Previous version used to handle this case. I've fixed it in 7ea482d633554295740f1227441fb1541f618e4e, but it's not nice that I had to :P Broken sources: `global-history-source` (doesn't even start if there're no deprecated "Last-Access" entries in history).
2. The name of `average-attribute-width` is clearly inconsistent with the fact that `alexandria:median` is used.
3. `clip-extremes` is missing, and some attributes are _tiny_ because of that. My implementation used to `clip-extremes` and ensured that the minimal size of the columns is both visible enough and consistent. Broken sources: `global-history-source` ("Last-Access" column shunk into non-existence), `file-source` ("Directory" column is dominating over all the space, with the "Name" and "Extension" columns shunk beyond visual balance).
4. It does not take into account function-valued attributes. That's why my implementation used `prompter:active-attributes`—to get the computed values of attributes. Broken source: `buffer-source` (doesn't even show anything due to type mismatch).
5. Let alone the fact that there's an unnecessary and a clearly case-specific change in `libraries/prompter`: `prompter:all-attribute-values`. Removing blank values is violating one's expectations that the length of the source should be the same as the length of attribute values (even if all of them are blank). I'd remove it and resurrect what was there before it in `prompt-buffer.lisp`, because that's where this logic belongs. And, even if retain it for this use-case (which I wouldn't do), then at least rely on `prompter:active-attributes` to fix the problem number 4.

aartaka:
Note that I'm not complaining about normalization to one and function renames. I even am okay with `alex:median`... more or less. I would have renamed and normalize things myself eventually, based on our discussions.

What I'm not okay is regressions in the functionality with no clear benefit, which the last set of commits is clearly doing [I was even unable to insert the angry kaomoji here, because `nx-kaomoji:kaomoji-source` seems to error out with this PR too] :(

aadcg:
I said: don't look at the code. 

I said: reach consensus regarding column jittering. In my book, that consensus has been reached. But stubbornness seems to burn inside yourself... 

I have said it all, and now I'll move into silence since I feel that my time could be put to good use. I'm dropping out of this thread that makes me anxious for no good reason.

Ambrevar:
Peeps, please cool down, no need to get angry over such a trivial discussion.

aartaka:
@aadcg, I'm sorry. I indeed am somewhat irritated and stubborn about this PR, because I feel it to be a positive change and am willing to see it merged with the concepts and ideas I put into it. Sorry if I shifted to personal remarks—I shouldn't've had.

> I said: don't look at the code.

This I honestly tried, as a good boy! But then, the very first prompt I tried to open errored out and didn't appear. There's not much one can do in such a situation but to try fixing the code. I've fixed the issue I've encountered (7ea482d633554295740f1227441fb1541f618e4e) and went forward with testing other prompts. I only started looking at the code seriously after I encountered half a dosen errors stemming from the issues I highlighted. My tone was not nice, yes, but I sincerely tried my best to follow your directions and failed, this being the reason of my irritation.

> I said: reach consensus regarding column jittering. In my book, that consensus has been reached. But stubbornness seems to burn inside yourself...

The consensus is quite a shaky one:
- @Ambrevar: no jittering.
- @aadcg: ?
- @aartaka: jittering in adequate bounds which don't violate the expectations.

My view is a result of chaotic UX learning (@lansingthomas may be the judge in this argument, actually), and I've said what my opinion is based on. I'm not meaninglessly stubborn, I stand by what I believe to be a positive UX change.

> I said: I will soon push a draft with the adjustment without column jittering.

This I was not debating, even though I believe it to be a regression too. I'd have to see and try it first to make decisions.

Ambrevar:
So, allow me to act as a diplomat / dictator here and make a decision for everyone :)

- Let's implement @aartaka's design and make it _opt-in_.  This means, add a boolean slot which by default has column adjustment _disabled_.  

Rationale: Because we couldn't reach an agreement on a new default, I believe the best course of action is to keep the current default and make everything else optional.

I believe this discussion is a good lesson for the future: when a fiery discussion arises around _new defaults_, I propose that everyone keeps in mind that _it's not a big deal_ since Nyxt is so configurable, it's just a _default_ after all.  So if no consensus is found, let's call it a day and keep the current default.

aartaka:
Yes, thanks for concluding it! Will do!

aartaka:
Closing in favor of #2741.

PR #2689, opened 2022-12-05
Unify height and width handling for buffers & other widgets

aartaka:
# Description

We used to have a disjoint zoo of height&width methods:
- `ffi-window-status-buffer-height`,
- `ffi-buffer-height`,
- `ffi-buffer-width`,
- `ffi-window-panel-buffer-width`,
- `ffi-window-prompt-buffer-height`.

Now we have just two: `ffi-width` and `ffi-height`, which are dispatchabel over most Nyxt widget-related classes (which means `window` and `buffer` with descendants.

# Discussion

What I don't like is that status, message, and panel buffer containters are stored inside the `window`, so adjusting the height of those requires always resorting to `window`. While it's not much problem, it's not exactly of a design, and we're best to move those `gtk-object`s of these special buffers into their own classes and not touch `window` in `ffi-width`/`ffi-height` at all.

As an additional pain point, message buffer doesn't even have a class, which made me:
- Retain `ffi-window-message-buffer-height`,
- And leave `message-buffer-style` with `message-buffer-height` inside `window`. 

Given that we have at least several properties that clearly belong to `message-buffer`, I believe we have to move it into a separate class. This way, we'll have a more unified API for message buffer (`style`, `ffi-height`) and will be able to backport all the niceties regular buffers have, to `message-buffer` too.

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/window.lisp
+++ b/source/window.lisp
@@ -77,10 +77,11 @@ the generic functions on `status-buffer'.  Finally set the `window'
         :margin 0)))
    (prompt-buffer-open-height
     ;; Ensures that 10 suggestions are vertically shown, while the last entry
-    ;; appears truncated thus making it clear that more follow.
-    ;; It's also set exactly to a third of the default Nyxt window height
-    ;; (768:256).
-    256
+    ;; appears truncated thus making it clear that more follow. It's also set
+    ;; exactly to a third of Nyxt window height.
+    :unbound
+    :reader nil
+    :writer t
     :documentation "The height of the prompt buffer when open.")
    (input-dispatcher
     'dispatch-input-event
@@ -123,6 +124,12 @@ The handlers take the window as argument."))
     (setf (id window) (new-id))
     (setf (slot-value browser 'last-active-window) window))
   window)
+
+(defmethod prompt-buffer-open-height ((window window))
+  (if (slot-boundp window 'prompt-buffer-open-height)
+      (slot-value window 'prompt-buffer-open-height)
+      (setf (slot-value window 'prompt-buffer-open-height)
+            (round (/ (ffi-height window) 3)))))
 
 (defmethod print-object ((window window) stream)
   (print-unreadable-object (window stream :type t :identity t)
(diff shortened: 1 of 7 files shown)


aadcg:
> Given that we have at least several properties that clearly belong to message-buffer, I believe we have to move it into a separate class. 

Sounds reasonable to me!

source/window.lisp
```diff
@@ -124,6 +125,12 @@ The handlers take the window as argument."))
     (setf (slot-value browser 'last-active-window) window))
   window)
 
+(defmethod prompt-buffer-open-height ((window window))
+  (if (slot-boundp window 'prompt-buffer-open-height)
+      (slot-value window 'prompt-buffer-open-height)
+      (setf (slot-value window 'prompt-buffer-open-height)
+            (round (/ (ffi-height window) 3)))))
```
aadcg, line 132: We can't guarantee that the last visible suggestion will appear truncated with this change so we revise that comment.

source/window.lisp
```diff
@@ -77,10 +77,11 @@ the generic functions on `status-buffer'.  Finally set the `window'
         :margin 0)))
    (prompt-buffer-open-height
     ;; Ensures that 10 suggestions are vertically shown, while the last entry
-    ;; appears truncated thus making it clear that more follow.
-    ;; It's also set exactly to a third of the default Nyxt window height
-    ;; (768:256).
-    256
+    ;; appears truncated thus making it clear that more follow. It's also set
```
aadcg, line 80: This comment here.

source/window.lisp
aartaka: We were unable to guarantee that with 256 either. It all depended on screen size and font setting on the system. So, this is not even a regression (although it definitely exposes the fact that we hard-code the number of displayed suggestions).

aadcg: That is correct. The only reason why I left that comment was to expose the rationale behind the hardcoded value. Not a single thing in this PR can be regarded as a regression :)

aartaka:
Yes!

aadcg:
The CI failed for `ccl`, but the tests passed for me locally with `ccl`. Ignoring and merging.

aadcg:
Thanks @aartaka.

PR #2713, opened 2022-12-20
Add customizable status bar placement

efimerspan:
# Description

This allows users to choose whether to place the status buffer at the top of the GTK window. I'm of the belief that status bars should be at eye-level for optimal ergonomics. 

# Discussion

Mention there any suspicious parts of the new code, or the ideas that you'd like to discuss in regards to this change.

# Checklist:

- [-] I've added the new dependencies as:
  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -451,7 +451,8 @@ response.  The BODY is wrapped with `with-protect'."
                   prompt-buffer-view
                   status-buffer status-container
                   message-container message-view
-                  key-string-buffer) window
+                  key-string-buffer)
+         window
        (unless gtk-object
          (setf gtk-object (make-instance 'gtk:gtk-window
                                          :type :toplevel
@@ -486,6 +487,9 @@ response.  The BODY is wrapped with `with-protect'."
          ;; receive input, for instance to create a new buffer.
          (setf nyxt::active-buffer (make-instance 'input-buffer))
 
+         (when (eq (status-buffer-position window) :top)
+           (gtk:gtk-box-pack-start root-box-layout status-container :expand nil))
+
          ;; Add the views to the box layout and to the window
          (gtk:gtk-box-pack-start main-buffer-container (gtk-object nyxt::active-buffer) :expand t :fill t)
          (gtk:gtk-box-pack-start horizontal-box-layout panel-buffer-container-left :expand nil)
@@ -499,7 +503,9 @@ response.  The BODY is wrapped with `with-protect'."
          (setf (gtk:gtk-widget-size-request message-container)
                (list -1 (message-buffer-height window)))
 
-         (gtk:gtk-box-pack-end root-box-layout status-container :expand nil)
+         (when (eq (status-buffer-position window) :bottom)
+           (gtk:gtk-box-pack-end root-box-layout status-container :expand nil))
+
          (gtk:gtk-box-pack-start status-container (gtk-object status-buffer) :expand t)
          (setf (gtk:gtk-widget-size-request status-container)
                (list -1 (height status-buffer)))
--- a/source/window.lisp
+++ b/source/window.lisp
@@ -63,6 +63,10 @@ The channel is popped when a prompt buffer is hidden.")
 To modify the status buffer appearance and behavior, subclass it and specialize
 the generic functions on `status-buffer'.  Finally set the `window'
 `status-buffer' slot to an instance of this subclass.")
+   (status-buffer-position
+    :bottom
+    :type (member :top :bottom)
+    :documentation "The position where to place the status buffer in the GTK window.")
    (message-buffer-height
     16
     :documentation "The height of the message buffer in pixels.")


source/window.lisp
```diff
@@ -63,6 +63,11 @@ The channel is popped when a prompt buffer is hidden.")
 To modify the status buffer appearance and behavior, subclass it and specialize
 the generic functions on `status-buffer'.  Finally set the `window'
 `status-buffer' slot to an instance of this subclass.")
+   (status-buffer-top-p
```
aartaka, line 66: That feels slightly off. I'd rather make it
```
   (status-buffer-position
     :bottom
     :type (member :top :bottom)
     :documentation "Where in the GTK window should the status buffer reside.") 
```

Because "position" is not exactly as binary top vs. bottom concept, and so binary/boolean choice is not fully matching the iea of status buffer position. Having a multiple-choice keywords with the possibility of extending those to include e.g., `:side` or `:below-message-buffer` is a more reliable choice here.

source/renderer/gtk.lisp
```diff
@@ -451,89 +451,99 @@ response.  The BODY is wrapped with `with-protect'."
...
-         (setf main-buffer-container (make-instance 'gtk:gtk-box
-                                                    :orientation :vertical
-                                                    :spacing 0))
-         (setf prompt-buffer-container (make-instance 'gtk:gtk-box
-                                                      :orientation :vertical
-                                                      :spacing 0))
-         (setf message-container (make-instance 'gtk:gtk-box
+                  key-string-buffer)
+         window
+       (flet ((pack-status-buffer ()
```
aartaka, line 456: If we stay with boolean slot, then this is fine. But, as boolean slot doesn't feel right, a more reliable choice would be to just spill over all of these `setf`s some 
```
...
(when (eq (status-buffer-position window) :top)
  (gtk:gtk-box-pack-end root-box-layout status-container :expand nil))
...
```
setting the status buffer into the current GTK state.

aadcg:
@efimerspan amazing work, thanks for your time!

@aartaka jumped right into the code itself, but there are key conceptual questions to massage first.

If we add the possibility to add the status buffer to the top, we lose the clear distinction between the buffer and the message buffer (see the screenshot below).

Notice that when the status buffer is drawn below (our default), you get (from top to bottom): buffer, status buffer and message buffer. 

I don't approve this change until we have a solution for this issue. Therefore, all of the technical implementation details are less important right now.

aartaka:
@aadcg, that's a UI issue, and it's solved easily by adding a constrasting border to `message-buffer-style` :P

aartaka:
See 24785c8a011b4b7fe40c10521369d4dc58b31b9a for the fix :)

aadcg:
@aartaka can we please go one step at a time?

I raised a blocking UI issue. You agreed it's an issue. 

Before actually implementing a solution, it's time to let people reach consensus on a solution. Then it can be implemented. 

You suggest adding a border around the message buffer. I  think that doesn't solve the issue because:

- it draws a useless border when the prompt buffer is located below (drawing it conditionally is an equally awful idea because users need consistent UIs)
- when the prompt buffer is located above, the border doesn't look very well (aesthetically)

More importantly, we have to deal with the following: our message buffer allocates a line on the screen, that is present at _all_ times (even when no message is being echoed). When the message buffer is delimited by the prompt buffer, I think that's fine. But, on the other hand, when the message buffer is a dangling and lonely line that borders with the main buffer... That's bad. It looks bad, and it feels like some kind of esoteric allocation of screen space.

On most browsers, the equivalent to our message buffer is an area that typically occupies a fraction of a line and is drawn on top of the main buffer for as long as it is actually echoing any information. I don't like that UI either.

I suggest the following: when the prompt buffer is drawn at the top, the message buffer must be drawn on top of it. 

To make it clear. Here's how things should look from top to bottom:

- prompt buffer below:
  - main buffer
  - prompt buffer
  - message buffer
- prompt buffer above:
  - message buffer
  - prompt buffer
  - main buffer

efimerspan:
@aadcg I partially agree with your sentiment, although I think you might have mistaken the prompt buffer for the status buffer in your previous comment, right? 

In my opionion, I don't think placing the message buffer at the top makes much sense since most of the time it doesn't convey useful information to the user, which is instead located in the status buffer, plus it takes focus away from the status buffer. 

I've been recently playing with the idea of using the Emacs header line as the mode line and it works and looks better from my short time testing it. The Emacs "message area" always stays at the bottom, most of the times without any content in it, which I personally see no issue with in regards to wasted screen space since it's insignificant. If Nyxt is trying to go for an "Emacsy" look, then I believe it would only make sense to be in line with Emacs's UI. However, if we truly want to mimic Emacs's UI, then we'd have to consider that when the prompt buffer is opened, the message buffer would need to be merged with the prompt buffer's input area, which I think overcomplicates things, or we could simply place it above the prompt buffer for now.

aartaka:
Or we can dynamically hide it when emptied and show when used, that's not hard at all :)

aadcg:
> I think you might have mistaken the prompt buffer for the status buffer in your previous comment, right?

Yes, sorry! I've edited it.

> In my opionion, I don't think placing the message buffer at the top makes much sense since most of the time it doesn't convey useful information to the user which is instead located in the status buffer, plus it takes focus away from the status buffer.

I understand your argument, but all of that is also true when you place the prompt buffer below... 

> I've been recently playing with the idea of using the Emacs header line as the mode line and it works and looks better from my short time testing it. 

Interesting. I've always wanted to do this!

> when the prompt buffer is opened, the message buffer would need to be merged with the prompt buffer's input area, which I think overcomplicates things, or we could simply place it above the prompt buffer for now.

You're raising another very good point.

The way I see this whole discussion is as follows.

For us to merge this change, we need to ensure that drawing the status buffer at the top or at the bottom results in a symmetrical UI. This means that, in both cases, the placing of prompt, status and message buffers must be consistent.

Another way to move forward would be to say that, from now on, we will rethink our UI and the status buffer will be place at the top. Since we wouldn't officially be supporting alternative UIs, we don't need to ensure any symmetry (as described in the paragraph above).

Yet another way is as follows. @efimerspan, we really appreciate your effort, but it seems to be that you're still experimenting and it seems premature to accept these changes. Still, your work was not in vain! You can add this to you config file and keep experimenting. We can help you with that if you get stuck.

aartaka:
@aadcg, symmetry does not always apply to interfaces, and it certainly needn't apply here. Moving status buffer is only moving status buffer, no more. Better have it as a (possibly unstable and un-aesthetic) toggle than not have it. I mean, yeah, the first iteration with a boolean I don't like, but not because the idea of moving the status buffer is bad (it's not), but because the implementation does not allow for all the quirks of individual configurations there might be :)

Yes, message buffer gets obscured by the main content, which is not the problem of moving status buffer per se and not a valid reason to deny status movement. It's a problem of message buffer contrast and visibility, which I tried to address in 24785c8a011b4b7fe40c10521369d4dc58b31b9a. If that's a solution that you don't like (I've reverted it), let's iterate on it, in a separate issue/PR. Now that the problem of UI elements being naively non-self-sufficient is evident, we can work on it. But that's not related to status buffer movement and is a separate thing.

I can work on making message buffer distinctive enough, if you think it necessary. Should not be hard. But not here, it's unrelated.

aartaka:
If one would have to come up with perfect symmetry of Emacs buffer splitting, one'd end up with Visual Studio Code, not Emacs （;≧皿≦）

aadcg:
In short, I don't oppose this change!

aadcg:
> If one would have to come up with perfect symmetry of Emacs buffer splitting, one'd end up with Visual Studio Code, not Emacs 

I seems I confused you with the word "symmetry". I used it as meaning "invariant".

The symmetry I alluded before is the follows: the status buffer is placed in between the message buffer and the main buffer to ensure that the content of the main and messages buffers are sharply distinct. 

Dropping this invariant (when the status is drawn at the top) is a bad thing, in my opinion, but I can live with that!

efimerspan:
Alright, addressed the points you raised @aartaka.

aartaka:
@efimerspan, looks perfect to me. Can you squash all the commits into one so that we can merge?

efimerspan:
@aadcg Done.

aartaka:
@efimerspan, thanks a lot `(‘∀’●)♡`

PR #2716, opened 2022-12-21
Silent CCL compiler warnings.

### LINKED ISSUE #2715: Mode tests fail with CCL
See https://github.com/atlas-engineer/nyxt/actions/runs/3749661391/jobs/6368388504.

@aadcg Can you fix it?

aadcg:
# Description

Fixes #2715

# Discussion

The problem was not directly related to testing.

When a slot is set to a value, CCL seems to be very strict and only allows re-setting it to another value of the same type (regardless of what was passed as `:type` for that slot).

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [x] Documentation:
  - [x] (If this changes something about the features showcased on Nyxt website) I have these changes described in the new/existing article at Nyxt website or will notify one of maintainters to do so.
- [x] Compilation and tests:

### DIFF
--- a/source/mode/repeat.lisp
+++ b/source/mode/repeat.lisp
@@ -42,7 +42,7 @@ It takes a `repeat-mode' instance as argument.")
     #'(lambda (path-url mode)
         (declare (ignore path-url))
         ;; Needed since the mode object might not have been garbage collected.
-        (setf (repeat-action mode) nil
+        (setf (repeat-action mode) (constantly nil)
               (repeat-count mode) nil
               (repeat-interval mode) 1.0))
     :documentation "See `nyxt/process-mode:cleanup'.")))
--- a/source/mode/watch.lisp
+++ b/source/mode/watch.lisp
@@ -34,7 +34,7 @@
 (define-mode watch-mode (nyxt/repeat-mode:repeat-mode)
   "Reload the current buffer every 5 minutes."
   ((rememberable-p t)
-   (nyxt/repeat-mode:repeat-interval 300)
+   (nyxt/repeat-mode:repeat-interval 300.0)
    (nyxt/repeat-mode:repeat-action
     #'(lambda (mode)
         (reload-buffer (buffer mode))))))


source/mode/repeat.lisp
```diff
@@ -42,7 +42,7 @@ It takes a `repeat-mode' instance as argument.")
     #'(lambda (path-url mode)
         (declare (ignore path-url))
         ;; Needed since the mode object might not have been garbage collected.
-        (setf (repeat-action mode) nil
+        (setf (repeat-action mode) (constantly nil)
```
Ambrevar, line 45: This is odd, because the type of the slot is `(or null (function (repeat-mode)))`, so `nil` should work here.

@aartaka Any idea what's going on?

aadcg: From my top post (sorry if have actually read it @Ambrevar):

> When a slot is set to a value, CCL seems to be very strict and only allows re-setting it to another value of the same type (regardless of what was passed as `:type` for that slot).

Ambrevar: I read it, but this is not "normal" behaviour.

aartaka: No idea, but that doesn't look normal to me either 0_o Maybe make a small reproducible example and send to CCL team?

Ambrevar: Yes, let's do that!

aartaka:
I found it! Type inference in `nclasses` is working on inferited slots too, so `nyxt/repeat-mode:repeat-action` in `watch-mode` is inferred to be a `function` and thus unsuitable for NIL value. Try `macroexpand-1` a couple of times on `watch-mode` definition and you'll see that there's nothing specific to CCL—CCL is just more attentive to types :P

The solution to that would be not inferring the type of inherited slots. Is that possible? Dunno...

@Ambrevar, any idea?

Ambrevar:
> The solution to that would be not inferring the type of inherited slots. Is that possible? Dunno...

I think it is, because we are already doing some similar checks with regard to exporting.  So what we could do is check if the parent slots has a type inference before inferring one.

Ambrevar:
On second though: what if the child initform does not match the parent inferred type?  Then we get an warning.  Probably what we want though, considering the same slot names in related classes have different type, the user should probably write the type explicitly.

aartaka:
Yes.

aadcg:
Is one of you taking over this PR? @aadcg @Ambrevar

Ambrevar:
Yes, I'm working on it.

aadcg:
@Ambrevar will you create another PR or force-push here?

Ambrevar:
Probably here.

Ambrevar:
Closing in favour of #2764.

aadcg:
Branch deleted.

PR #2685, opened 2022-12-02
Test mode togglers

aadcg:
# Description

Add a template mode toggler test for each mode.

The `vi` test is the only one active at the moment. I've tested, and it would have prevented us from making mistakes as the one that was fixed by 18e19553d.

Fixes #2683

# Discussion

Either merge this is as-is, or give me some time to turn the dummy tests into proper ones for all modes.

Notice that each mode requires slightly different conditions, so it's not a simple blind copy and paste operation.

# Checklist:
Everything in this checklist is required for each PR.  Please do not approve a PR that does not have all of these items.

- [ ] I have pulled from master before submitting this PR
- [ ] There are no merge conflicts.
- [ ] I've added the new dependencies as:
  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    git submodule add https://gitlab.common-lisp.net/nyxt/py-configparser _build/py-configparser
    ```
  - [ ] and Guix dependencies.
- [ ] My code follows the style guidelines for Common Lisp code. See:
  - [Norvig & Pitman's Tutorial on Good Lisp Programming Style (PDF)](https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf)
  - [Google Common Lisp Style Guide](https://google.github.io/styleguide/lispguide.xml)
- [ ] I have performed a self-review of my own code.
- [ ] My code has been reviewed by at least one peer.  (The peer review to approve a PR counts.  The reviewer must download and test the code.)
- [ ] Documentation:
  - [ ] All my code has docstrings and `:documentation`s written in the aforementioned style.  (It's OK to skip the docstring for really trivial parts.)
  - [ ] I have updated the existing documentation to match my changes.
  - [ ] I have commented my code in hard-to-understand areas.
  - [ ] I have updated the `changelog.lisp` with my changes if it's anything user-facing (new features, important bug fix, compatibility breakage).
  - [ ] I have added a `migration.lisp` entry for all compatibility-breaking changes.
  - [ ] (If this changes something about the features showcased on Nyxt website) I have these changes described in the new/existing article at Nyxt website or will notify one of maintainters to do so.
- [ ] Compilation and tests:
  - [ ] My changes generate no new warnings.
  - [ ] I have added tests that prove my fix is effective or that my feature works.  (If possible.)
  - [ ] New and existing unit tests pass locally with my changes.

PR #2717, opened 2022-12-22
Fix the architecture of some modes

aadcg:
# Description

In PR #2685 I made the following observation: instantiating a `modable-buffer` is _not_ always enough to `enable` a mode in it.

This seems odd. If I am a mode, all I should need is a modable-buffer to be enabled. 

When a mode is enabled, it means that it can _potentially_ act the way it was developed. Why only potentially? Because some pre-requisites might not be met. 

And why should we enable a mode when pre-requisites aren't met? If Nyxt was a program where we don't trust our users, it would be a good idea to limit and prevent the user from doing something off the charts.

An alternative to my proposal is: when pre-requisites aren't met, the mode is not allowed to be enabled. But, instead of raising a condition, it should gracefully exit the attempt to enable it.

No strong opinion, but it would be good to have everyone on the same page here. @Ambrevar @aartaka @jmercouris 

Take a look at the changes for a concrete example - `certificate-exception-mode`.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/mode/certificate-exception.lisp
+++ b/source/mode/certificate-exception.lisp
@@ -19,10 +19,16 @@ See the `add-domain-to-certificate-exceptions' command."
                            :type (list-of string))))
 
 (defmethod enable ((mode certificate-exception-mode) &key)
-  (setf (certificate-exceptions (buffer mode)) (certificate-exceptions mode)))
+  (let ((buffer (buffer mode)))
+    (if (network-buffer-p buffer)
+        (setf (certificate-exceptions buffer) (certificate-exceptions mode))
+        (echo "~a is not a network buffer. Nothing to be done." buffer))))
 
 (defmethod disable ((mode certificate-exception-mode) &key)
-  (setf (certificate-exceptions (buffer mode)) nil))
+  (let ((buffer (buffer mode)))
+    (if (network-buffer-p buffer)
+        (setf (certificate-exceptions buffer) nil)
+        (echo "~a is not a network buffer. Nothing to be done." buffer))))
 
 (define-command add-domain-to-certificate-exceptions (&optional (buffer (current-buffer)))
   "Add the current hostname to the buffer's certificate exception list.
--- a/tests/offline/mode/certificate-exception.lisp
+++ b/tests/offline/mode/certificate-exception.lisp
@@ -4,7 +4,7 @@
 (in-package :nyxt/tests)
 
 (define-test toggle-certificate-exception-mode ()
-  (let ((buffer (make-instance 'network-and-modable-buffer)))
+  (let ((buffer (make-instance 'modable-buffer)))
     (with-current-buffer buffer
       (assert-true (enable-modes* 'nyxt/certificate-exception-mode:certificate-exception-mode
                                   buffer))


aartaka:
I have several (quite polarized) stances on this PR. Neither of those I will bias as the one I'm leaning towards (I am in fact leaning towards one of those), so let's just take it for three separate people/worldviews talking this.

# Logic

While `modable-buffer` sounds like something all modes should work in, it's not. `modable-buffer` only implies that modes are there, not that they work. The prerequisites for the modes may be in other buffer mix-ins, or even beyond any classes/mix-ins. 

To rephrase it in a form of with a logical rule (that head of my school once tried to deny):
```
Every fish is an animal, 
BUT
not all animal is fish.
```
in our case
```
Any buffer mode works in is modable, 
BUT
not all modable buffers should work with all modes.
```

Thus, the suggestion: **give up on testing modes against `modable-buffer` and test them agains the feature set they require (= change no thing)**.

As a continuation of this logical point of view, a pragmatical one:

# Pragma

While `modable-buffer` and other mix-ins may sound like a minimal necessary thing to test a mode against, this is not a practical thing. Modes are usually run in one of:
- `web-buffer`,
- `prompt-buffer`, or
- `editor-buffer`.

While can introduce other classes of buffers still based on existing mix-ins, they will still imply some scenarios and sets of modes that will work in them.

Thus, the suggestion: **test modes against one of the *practical* buffer classes that they *will* be enabled on in practice**.

# Discipline

Yes, we have to test modes and check for the conditions they are enabled in. BUT! 
1. We have to error out if those conditions are not met, because that gives a much clearer and debuggable indication of what went wrong.
2. And we have to clearly define which buffer class a given mode belongs to. E.g. `editor-mode` belongs to `editor-buffer`, `document-mode` belongs to `document-buffer`, and, if one enables those for something else, then error out terribly (1). Which means that we have to extend `define-mode` to provide the buffer class mode should be enabled in OR add a slot checking that mode prerequisites are met. Why? Because `enable` should only be concerned with enabling the mode, not checking whether it's enableable. `enable :around` — maybe, but not `enable` and not per mode. We need a more general solution.

Thus, the suggestion: **it's not strict, structured, and type-based enough, make it to be so**.

aadcg:
@aartaka you're leading this conversation to a place where we should not go. Even though I understand perfectly well the point you're making (and I can guess what you'll suggest).

Testing the mode toggler is not the same as testing the _functionality_ of mode! I didn't not test any mode (except `process`). When testing mode togglers, you should rely on the absolute minimum. 

I have some ideas to improve my top post suggestions, also taking into account your observations.

aartaka:
> Testing the mode toggler is not the same as testing the _functionality_ of mode! I didn't not test any mode (except `process`). When testing mode togglers, you should rely on the absolute minimum (which is creating a modable buffer).

If we restrict ourselves to always _just_ testing mode togglers, then yes, you're perfectly correct. But what we, in fact, are testing in the tests for every mode we toggle, is that it works (successfully enables, disables, and TODO does it's direct responsibilities) in the conditions it should work in. These conditions include a buffer that mode gets enabled in, and, often, `*browser*` too. And `enable`, being the rightful part of mode API/functionality, has all the rights to require these conditions too.

So why should we test the mode (even its handlers) in the conditions it's not intended to be used in (absent `*browser*`, barebones impractical `modable-buffer`)?

aadcg:
> So why should we test the mode (even its handlers) in the conditions it's not intended to be used in (absent *browser*, barebones impractical modable-buffer)?

Exactly. This represents your point of view. You get really scared when you see me writing tests as if I'm 4-years old and I never looked at Nyxt's codebase. Well, I do it on purpose. Tests should be naive and separate things as much as possible (as much as the architecture allows). 

If you don't test with the minimal amount of paraphernelia you need, you're not testing the API. 

I think I will probably close this PR, as it seems that I have convinced myself that the dummy buffer I've created (such as `network-and-modable-buffer`) make perfect sense.

In short, I disagree with something I wrote in the post post:

> This seems odd. If I am a mode, all I should need is a modable-buffer to be enabled.

aartaka:
(・ε・)

aadcg:
A modables buffers is a necessary condition for enabling a mode, but it's not sufficient. Because enabling a mode is more than adding objects to the `modes` slot.

Closing. Thanks.

aartaka:
> A modable buffer is a necessary condition for enabling a mode, but it's not sufficient. Because enabling a mode is more than adding objects to the `modes` slot.

QED.

PR #2718, opened 2022-12-22
Test start

### LINKED ISSUE #2710: Test `start.lisp`
The first battery of tests should be done in the context of `nyxt` since doing it on the level of `nyxt/renderer` raises the complexity bar. 

For all `start` parameters check that:

- `(nyxt:start :parameter t)` => variable corresponding to `parameter` is rightly set
- Don't clean Nyxt's state
- `(nyxt:start)` => ensure that `parameter` fallbacks to the default value.

In the context of `nyxt/renderer`, there are some important checks:

- To test `entry-point` and `start`:
  - either add an option to pass a handler after init;
  - or have `start` reuse an existing `*browser*`;
  - or (best option in my opinion), have a client listen to the socket for readiness, then send instruction to quit.
- To test all other browser initialization features, bypass `start` and initialize `browser` ourselves.

aadcg:
Fixes #2710 

# Discussion

Work in progress. I'd like you to take a look at the strategy I'm taking. @ambrevar @aartaka @jmercouris 

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/start.lisp
+++ b/source/start.lisp
@@ -163,40 +163,41 @@ Without --remote, it also disables socket use."))))
 
 (define-command quit (&optional (code 0))
   "Quit Nyxt."
-  (when (slot-value *browser* 'ready-p)
-    (progn
-      (hooks:run-hook (before-exit-hook *browser*))
-      ;; Unready browser:
-      ;; - after the hook, so that on hook error the browser is still ready;
-      ;; - before the rest, so to avoid nested `quit' calls.
-      (setf (slot-value *browser* 'ready-p) nil)
-      (setf (slot-value *browser* 'exit-code) code)
-      (dolist (window (window-list))
-        (window-delete window :force-p (/= 0 code)))
-      (when (socket-thread *browser*)
-        (destroy-thread* (socket-thread *browser*))
-        ;; Warning: Don't attempt to remove socket-path if socket-thread was not
-        ;; running or we risk remove an unrelated file.
-        (let ((socket (files:expand *socket-file*)))
-          (when (uiop:file-exists-p socket)
-            (log:info "Deleting socket ~s." socket)
-            (uiop:delete-file-if-exists socket))))
-      (mapc #'destroy-thread* (non-terminating-threads *browser*))
-      (ffi-kill-browser *browser*)
-      ;; Reset global state.
-      (setf *browser* nil
-            *options* nil)
-      (uninstall *renderer*)
-      ;; On FreeBSD this may cause freeze. Also we have to pass
-      ;; FINISH-OUTPUT = NIL in FFI-INITIALIZE.
-      #-freebsd
-      (unless *run-from-repl-p*
-        (run-thread "force-quitter"
-          ;; Force-quit in case `ffi-kill-browser' hangs.  Must be run in a
-          ;; separate thread because the renderer loop is waiting for this
-          ;; function to finish.
-          (sleep 1)
-          (uiop:quit code nil))))))
+  (if (not (and *browser* (slot-value *browser* 'ready-p)))
+      (progn (echo "There's no browser instance to kill.") nil)
+      (progn
+        (hooks:run-hook (before-exit-hook *browser*))
+        ;; Unready browser:
+        ;; - after the hook, so that on hook error the browser is still ready;
+        ;; - before the rest, so to avoid nested `quit' calls.
+        (setf (slot-value *browser* 'ready-p) nil)
+        (setf (slot-value *browser* 'exit-code) code)
+        (dolist (window (window-list))
+          (window-delete window :force-p (/= 0 code)))
+        (when (socket-thread *browser*)
+          (destroy-thread* (socket-thread *browser*))
+          ;; Warning: Don't attempt to remove socket-path if socket-thread was not
+          ;; running or we risk remove an unrelated file.
+          (let ((socket (files:expand *socket-file*)))
+            (when (uiop:file-exists-p socket)
+              (log:info "Deleting socket ~s." socket)
+              (uiop:delete-file-if-exists socket))))
+        (mapc #'destroy-thread* (non-terminating-threads *browser*))
+        (ffi-kill-browser *browser*)
+        ;; Reset global state.
+        (setf *browser* nil
+              *options* nil)
+        (uninstall *renderer*)
+        ;; On FreeBSD this may cause freeze. Also we have to pass
+        ;; FINISH-OUTPUT = NIL in FFI-INITIALIZE.
+        #-freebsd
+        (unless *run-from-repl-p*
+          (run-thread "force-quitter"
+            ;; Force-quit in case `ffi-kill-browser' hangs.  Must be run in a
+            ;; separate thread because the renderer loop is waiting for this
+            ;; function to finish.
+            (sleep 1)
+            (uiop:quit code nil))))))
 
 (define-command quit-after-clearing-session (&key confirmation-p) ; TODO: Rename?
   "Close all buffers then quit Nyxt."
--- a/source/start.lisp
+++ b/source/start.lisp
@@ -514,7 +514,8 @@ Examples:
       (t
        (with-protect ("Error: ~a" :condition)
          (start-browser urls))))
-    (unless *run-from-repl-p* (uiop:quit))))
+    (unless *run-from-repl-p* (uiop:quit)))
+  *browser*)
 
 (defun load-or-eval (&key remote)
   (when remote
--- a/tests/offline/start.lisp
+++ b/tests/offline/start.lisp
@@ -1,0 +1,35 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt/tests)
+
+(define-class dummy-renderer (renderer)
+  ((name "Dummy renderer"))
+  (:export-class-name-p t)
+  (:export-accessor-names-p t)
+  (:accessor-name-transformer (class*:make-name-transformer name)))
+
+(defmethod install ((_ dummy-renderer)) t)
+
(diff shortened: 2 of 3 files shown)


source/start.lisp
```diff
@@ -158,40 +158,41 @@ It skips configuration files and other data files like the history."))))
...
-      ;; FINISH-OUTPUT = NIL in FFI-INITIALIZE.
-      #-freebsd
-      (unless *run-from-repl-p*
-        (run-thread "force-quitter"
-          ;; Force-quit in case `ffi-kill-browser' hangs.  Must be run in a
-          ;; separate thread because the renderer loop is waiting for this
-          ;; function to finish.
-          (sleep 1)
-          (uiop:quit code nil))))))
+  (if (and *browser* (slot-value *browser* 'ready-p))
```
Ambrevar, line 161: Maybe wrap the condition with `(not ...)` and move the `(echo ...)` in the first branch to make the code more readable.

tests/offline/start.lisp
```diff
@@ -0,0 +1,37 @@
...
+  (:accessor-name-transformer (class*:make-name-transformer name)))
+
+(defmethod install ((_ dummy-renderer)) t)
+
+(defmethod uninstall ((_ dummy-renderer)) t)
+
+(setf nyxt::*renderer* (make-instance 'dummy-renderer))
+
+(define-test null-quit ()
+  (assert-true (nyxt:quit)))
```
Ambrevar, line 19: The return value of `nyxt:quit` is not explicitly typed.

tests/offline/start.lisp
```diff
@@ -0,0 +1,37 @@
...
+
+;; Question: should the socket be created when there's no browser object?
```
Ambrevar, line 21: I'd would we should be able to, depending on the instantiation options.  I can be useful to test remote communication.

In most tests there is no need for the socket however, and it can help to _not have it_ since we avoid the risk of leaving broken sockets behind.

tests/offline/start.lisp
```diff
@@ -0,0 +1,37 @@
...
+
+(define-test stateless-headless-argument ()
+  (nyxt:start :headless t :failsafe t)
+  ;; TODO find a way to avoid these calls to sleep.
+  (sleep 4)
```
Ambrevar, line 26: Currently the only proper way is via the `*after-startup-hook`.

But clearly it's not convenient for testing.  A few options:

- Add a `callback` option to `nyxt:start` which is executed when `after-startup-hook` is run.  This callback could either run the test code or fulfill a promise which the test code is waiting on.
- Add an (unexported?) promise slot to browser, and have it fulfilled when `after-startup-hook` is run.
  This way the caller can `force` the promise (which means "wait for the browser to be started").
- Make `nyxt:start` synchronous, which means we would put all the promise/force logic inside `nyxt:start` itself and only return from the function once the renderer is started.
  The drawback is that `nyxt:start` would now be blocking.  But... considering there is nothing to be done in the meantime, is this really a drawback?

Ambrevar (commented):
Looks good so far!

tests/offline/start.lisp
aartaka: - Finally move `wait-on(-handler)` to `nhooks`? not exactly a solution, but makes tests much prettier.

Ambrevar: I'm not fully satisfied with the `wait-on` approach: it's still a cumbersome approach that forces nesting over nesting...

The CPS approach is more interesting in this regard: https://github.com/atlas-engineer/nyxt/pull/2054#issuecomment-1043593315

aartaka: Or promises/futures and `lparallel:plet`?

source/start.lisp
aadcg: Good point.

tests/offline/start.lisp
aadcg: Good point.

tests/offline/start.lisp
aadcg: All clear.

tests/offline/start.lisp
aadcg: I think the best and cleaner solution would be to make `nyxt:start` synchronous. 

>  The drawback is that nyxt:start would now be blocking. But... considering there is nothing to be done in the meantime, is this really a drawback?

No. That's why I think this is the best solution.

This change would also gracefully handle `(progn (nyxt:start) (nyxt:quit))`.

aadcg: @Ambrevar are you interested in working on this or should I do it?

Ambrevar: If you think you have a clear idea of how to implement it, please go ahead :)

aartaka: > I think the best and cleaner solution would be to make `nyxt:start` synchronous.
> 
> > The drawback is that nyxt:start would now be blocking. But... considering there is nothing to be done in the meantime, is this really a drawback?
>
> No. That's why I think this is the best solution.
> 
> This change would also gracefully handle `(progn (nyxt:start) (nyxt:quit))`.

Okay, this one feels quite contradictory. 
- I understood that you want to make `nyxt:start` to happen synchronously. The only clear option I see for that is `bt:join-thread` on the thread that `nyxt:start` spawns. 
  - An alternative solution would be waiting on some `after-startup-hook` (or some alternative) in `nyxt:start`, but that is architecturally impossible, because `*browser*` is not yet initialized when we run `nyxt:start`.
- Given that we join threads, one cannot simply do `(progn (nyxt:start) (nyxt:quit))`, because `nyxt:start` does not terminate until `nyxt:quit` is called. Which is a vicious circle.
  - If we wait on some event (which is quite bad for an startup routine), then we'd be basically baking the waiting logic from the tests into the `nyxt:start`. And it does not belong there.
- All in all, I don't see much use in making `nyxt:start` synchronous, because it's either
  - Creating inconveniences for the otherwise trivial use-case (`bt:join-thread` I can do myslef), or
  - Introducing backwards references to the extremely fragile startup sequence.

I must be missing something here. I mean, I'd like to make our startup more reproducible, but not in those two ways I'm imagining, and not necessarily by making `nyxt:start` synchronous/blocking, so there should be a better way you're conjuring, @aadcg.

aadcg: Honestly, I don't want to work on the architecture of `start.lisp`. There is too much historical background that I lack, while @aartaka  and @Ambrevar  know it quite well. 

@Ambrevar suggestion on making `nyxt:start` seems sane to me, because I don't see the point in having the ability to do something while the renderer is loading (and @Ambrevar seems to agree).

With regards to the semantics of `(progn (nyxt:start) (nyxt:quit)` I think it should mean: "Start Nyxt, block it while the renderer loads and then, when `nyxt:start` exists, proceed and run `nyxt:quit`". I'm sure this is possible to do. But I understand that it would re-thinking a lot of the architecture. 

I can help with testing and making sense of the start process of course.

To answer @Ambrevar's comment:

>  If you think you have a clear idea of how to implement it, please go ahead :)

No, I have no clear idea. I just have a clear idea of how things ought to be :)

aartaka:
@Ambrevar, wait, should this be closed? @aadcg, there was something in addition to #2740 in this PR, right?

Ambrevar:
GitHub did this for me, so yes, this should remain open.

aadcg:
Yes, there is lots of work to be done on this front. Not on testing itself, but on making a proper start implementation. And the implementation specification is dictated by the test I write tests (that ought to pass).

aadcg:
`(nyxt:start)` returns `NIL` and so does `(nyxt:quit)` _after_ Nyxt was started.

What should be return value from running `(nyxt:quit)` when there's no browser instance? `NIL`?

CC @ambrevar @aartaka

tests/offline/start.lisp
aadcg: This is now solved.

Ambrevar:
Makes sense if `nyxt:start` returns `*browser*`.
`(nyxt:quit)` should probably return nothing / NIL.

aadcg:
> Makes sense if nyxt:start returns *browser*.

Currently it returns NIL, but I agree that it should return `*browser*`.

tests/offline/start.lisp
```diff
@@ -0,0 +1,35 @@
...
+(setf nyxt::*renderer* (make-instance 'dummy-renderer))
+
+(define-test null-quit ()
+  (assert-false (nyxt:quit)))
+
+;; See https://github.com/atlas-engineer/nyxt/issues/2754
+;; (define-test start-quit-cycle ()
+;;   (assert-false (progn (nyxt:start :failsafe t) (nyxt:quit))))
+
+;; Fails.
```
aadcg, line 25: @Ambrevar @aartaka @jmercouris 

How critical do you think that this is?

For the user that starts Nyxt via the OS shell, it doesn't matter much.

For the developer that starts Nyxt from the REPL it's critical.

Ambrevar: Oops, fell under my radar.

It's not critical production-wise, as you pointed out.
Still, this highlights that our startup sequence is not as functional as we would hope, which indicates we wrote poor code :(

aadcg: Agreed!

I'll soon come back to this PR by adding all tests that would capture the invariants of a good startup sequence. If some are not met, then I'll leave them commented out.

jmercouris: Pierre covered it!

aadcg: What do you mean @jmercouris?

aadcg: Got it!

jmercouris:
Please reopen when ready.

PR #2724, opened 2022-12-26
Discernible status and message buffers

### LINKED ISSUE #2719: Make message buffer more self-sufficient and discernible.
After #2713, we can move status buffer to the top of the view. Which exposes a problem: our message buffer, being a thin white/black strip at the bottom of the screen, is not distinctive enough to be discerned from the page content. We have to make it a bit more self-sufficient UI-wise. Several options there:
- Make a contrasting border (rejected by @aadcg as too simplistic?)
- Make the design of message buffer slightly more UI-like (primary color background? angular arrows? something else?)
- Make message buffer appear only when it's necessary and hide it otherwise (not exactly a solution, but should work nicely with the other two). 

Clearly, the solution is actually a fusion of those three, not any single of those (although the idea with a contrasting border I believe to be quite sufficient and effective).

CC @lansingthomas, @aadcg, @jmercouris

aartaka:
# Description

This makes our message and status buffers a bit more UI-like, adding a slight groove border to both. It's quite vintage, but it seems perfect for this case of UI highlighting.

Fixes #2719.

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:


aadcg:
As I've mentioned in a [comment](https://github.com/atlas-engineer/nyxt/issues/2719#issuecomment-1363184158):

> I don't oppose anything. Except if you decide to change something about the default UI!



This PR proposes changing the background color of the message buffer of our default UI, so you should expect that everyone will oppose (our users included). 

For us to change the default UI, we need to identify the problem we're solving. Issue #2719 identifies a problem related to a UI that is _not_ the default one. So, I think it's fair to say that there's no problem related to the default UI.

I've also mentioned the following:

> drawing a border around the message buffer is a viable solution when done **conditionally** (when the status buffer is at the top).


Meaning that changing the UI conditionally is OK, since it doesn't touch the default one. 

Perhaps you consider that consistency is more important that preserving the default UI. My point from the beginning is that we could meet both goals if we were to draw the message buffer above the status buffer when the status buffer is drawn at the top.

Regardless of all of the above, I think this change is aesthetically inferior to what we have now.

Offloading the decision to @jmercouris.

aadcg:
At some point we'll have @lansingthomas thinking about this issue, but this is not a priority.

aartaka:
Closing, as we'll eventually come to some model, but right now it's all undecided.

PR #2738, opened 2023-01-09
Enable vi-insert whenever we see an input field.

### LINKED ISSUE #2727: vi insert not always getting activated on forms
**Describe the bug**
Many times I select a form entry like login/pass or search bar and nyxt is not on insert so screws up my writing since r and o and other special keys will get triggered and screw up my page or behaving in a way not intended, and I have to manually go into command mode ":" and type "vi-insert.." and enter to be able to finally type correctly

Happens more than not, rarely it does what is supposed I will tab, mouse there or  use f to select the form and will auto trigger into vi-insert mode
but most times it does not :(
Screenshots attached with some examples a video will be better but...

OS: Arch GNU/Linux
Wayland/Sway
Nyxt latest nightly from git using AUR Arch.

more info in the nyxt forum: https://discourse.atlas.engineer/t/vim-mode-form-issues/592

[edit] The screenshots did not upload here in order but you  get the idea.
[edit2] Just notice someone posted a similar bug report https://github.com/atlas-engineer/nyxt/issues/2701 sorry

swappy-20221226-042436
swappy-20221226-042234
swappy-20221226-001402
swappy-20221226-042356

aartaka:
# Description

This is a catch-all fix to VI insert mode not being enabled in some input field cases, like using Tab to focus it or invoking a non-hint command. `on-signal-key-press` (now a generic function in buffer.lisp) should be reliable (though possibly a bit wasteful) enough for that.

@fictitiousexistence, @r3k2, can you check?

# Discussion

Going through all the modes of all the buffers on every keypress feels like a lot of CPU cycles wasted, doesn't it? CC @Ambrevar, @aadcg

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -681,6 +681,7 @@ regular commands, such as "
 
   (:h3 "Bug fixes")
   (:ul
+   (:li "VI insert mode is triggered in more cases where it should be triggered.")
    (:li "Invoke the right WebKit command when cutting text with " (:nxref :function 'ffi-buffer-cut))
    (:li "Fix the display of history suggestions when going forward in history.")
    (:li "Security: all the non-ASCII domain names are shown as IDN punycodes in addition
--- a/source/mode/document.lisp
+++ b/source/mode/document.lisp
@@ -223,51 +223,19 @@ It does not assume being online."))
   "Select all the text in the text field."
   (ffi-buffer-select-all buffer))
 
-(export-always 'element-focused)
-(defgeneric element-focused (mode) ; TODO: Make hook instead?  Or use both, have the default method call hook.
-  (:method ((mode t))
-    nil)
-  (:documentation "Method run when `focus-element' is called."))
-
-(defmacro focus-element ((&optional (buffer '(current-buffer))) &body element-script)
-  "Select the element pointed to by ELEMENT-SCRIPT.
-ELEMENT-SCRIPT is a Parenscript script that is passed to `ps:ps'."
-  (alex:with-gensyms (element)
-    (alex:once-only (buffer)
-      `(progn
-         (ps-eval :buffer ,buffer (let ((,element (progn ,@element-script)))
-                                    (ps:chain ,element (focus))
-                                    (ps:chain ,element (select))))
-         (dolist (mode (modes ,buffer))
-           (element-focused mode))))))
-
-(define-command focus-first-input-field (&key (type-blacklist '("hidden"
-                                                                "checkbox"
-                                                                "button")))
-  "Move the focus to the first input field of `buffer'."
-  ;; The list of input types can be found below.
-  ;; https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input
-  ;; TODO: The following results in 2 DOM traversal.  We should probably do the
-  ;; whole thing in a single Parenscript instead.
-  (ps-labels
-    ((nth-input-type
-      (i)
-      (let* ((input (ps:chain document
-                              (get-elements-by-tag-name "INPUT")))
-             (item (when input (ps:chain input (item (ps:lisp i))))))
-        (when item
-          (ps:chain item type)))))
-    (let ((i (do ((i 0 (1+ i)))
-                 ((notany
-                   (lambda (type) (equalp (nth-input-type i) type))
-                   type-blacklist)
-                  i))))
-      (focus-element ()
-        (let* ((input (ps:chain document
-                                (get-elements-by-tag-name "INPUT")))
-               (item (when input (ps:chain input (item (ps:lisp i))))))
-          (when item
-            item))))))
+(define-command focus-first-input-field (&key (buffer (current-buffer)))
+  "Move the focus to the first input field of BUFFER."
+  ;; There are two basic ways to have an editable widget on a webpage:
+  ;; - Using <input>/<textarea>,
+  ;; - or marking any other element as contenteditable:
+  ;;   https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable
+  (loop with inputs = (clss:ordered-select "input, textarea, [contenteditable]" (document-model buffer))
+        for input across inputs
+        when (ps-eval :buffer buffer
+               (nyxt/ps:element-editable-p
+                (nyxt/ps:qs-nyxt-id document (ps:lisp (nyxt/dom:get-nyxt-id input)))))
+          do (nyxt/dom:focus-select-element input)
+          and do (return input)))
 
 (defmethod nyxt:on-signal-load-committed ((mode document-mode) url)
   (declare (ignore mode url))
--- a/source/mode/vi.lisp
+++ b/source/mode/vi.lisp
@@ -94,9 +94,12 @@ See also `vi-normal-mode' and `vi-insert-mode'."
       (enable-modes* 'nyxt/passthrough-mode:passthrough-mode buffer))))
 
 (defun vi-insert-on-input-fields (buffer)
-  (when (nyxt/document-mode:input-tag-p
-         (ps-eval :buffer buffer (ps:@ document active-element tag-name)))
-    (enable-modes* 'vi-insert-mode buffer)))
+  (if (ps-eval :buffer buffer
+        (if (and (ps:@ document active-element)
+                 (nyxt/ps:element-editable-p (ps:@ document active-element)))
+            ps:t ps:false))
+      (enable-modes* 'vi-insert-mode buffer)
+      (enable-modes* 'vi-normal-mode buffer)))
 
 (defmethod on-signal-load-finished ((mode vi-insert-mode) url)
   (declare (ignore url))
@@ -104,6 +107,10 @@ See also `vi-normal-mode' and `vi-insert-mode'."
   (vi-insert-on-input-fields (buffer mode)))
 
 (defmethod on-signal-button-press ((mode vi-normal-mode) button-key)
+  (when (string= "button1" (keymaps:key-value button-key))
+    (vi-insert-on-input-fields (buffer mode))))
+
+(defmethod on-signal-button-press ((mode vi-insert-mode) button-key)
   (when (string= "button1" (keymaps:key-value button-key))
     (vi-insert-on-input-fields (buffer mode))))
 
--- a/source/mode/vi.lisp
+++ b/source/mode/vi.lisp
@@ -118,8 +118,8 @@ See also `vi-normal-mode' and `vi-insert-mode'."
   (when (equal "tab" (keymaps:key-value key))
     (vi-insert-on-input-fields (buffer mode))))
 
-(defmethod nyxt/document-mode:element-focused ((mode vi-normal-mode))
-  (enable-modes* 'vi-insert-mode (buffer mode)))
+(defmethod nyxt/dom:focus-select-element :after ((element plump:element))
+  (vi-insert-on-input-fields (current-buffer)))
 
 (defmethod nyxt:mode-status ((status status-buffer) (vi-normal vi-normal-mode))
   (spinneret:with-html-string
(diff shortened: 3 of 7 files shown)


aadcg:
Can't help much here... :(

source/mode/vi.lisp
```diff
@@ -93,16 +93,24 @@ See also `vi-normal-mode' and `vi-insert-mode'."
...
-    (when (and (string= "button1" (keymaps:key-value button-key))
-               (nyxt/document-mode:input-tag-p
-                (ps-eval :buffer buffer (ps:@ document active-element tag-name))))
-      (enable-modes* 'vi-insert-mode buffer))))
+    (when (string= "button1" (keymaps:key-value button-key))
+      (vi-insert-on-input-fields buffer))))
+
+(defmethod on-signal-key-press ((mode vi-normal-mode) button-key)
+  (let ((buffer (buffer mode)))
+    (vi-insert-on-input-fields buffer)))
```
Ambrevar, line 113: Wait, doesn't this prevent VI commands when an input field is focused?  We need to be able to press `escape` and the VI commands.

aartaka: Oh, right, it does prevent command mode :cold_sweat: I guess we'd have to do something more complex here...

aartaka: Okay, cd1eeb9f3b3a977bd34a570e8fd9f3085d4b5e87 seems to be a good-enough fix. But I don't like that things are hard-coded :( Any idea for a better design here?

Ambrevar: I think the logic is wrong anyways.  For instance, some websites support keyboard shortcuts and the user might have the wrong behaviour because of the logic you wrote.

As far as I can tell, Qutebrowser et al. only enable insert mode _on focus_ (e.g. click or tab).  In particular, an input field _being focused_ should not trigger the insert mode in my opinion.

aartaka: > I think the logic is wrong anyways. For instance, some websites support keyboard shortcuts and the user might have the wrong behaviour because of the logic you wrote.

This method is merely a watcher, it cannot influence the way we interact with the renderer and consume the keys, so don't worry :)

> As far as I can tell, Qutebrowser et al. only enable insert mode _on focus_ (e.g. click or tab). In particular, an input field _being focused_ should not trigger the insert mode in my opinion.

Yes, that seems sensible! Will implement this option then :)

aartaka:
Okay, click and Tab are there. Not sure it solves the initial problem, but for these cases there's 027ca1e652585531052ac53fd02f78175a930be9. I've tried to make all the commands to check for inputs based on it, but I thought it to be overkill. So, until we find a better solution, this brick lies unused.

aartaka:
All in all, this is rebased over master and tested. Should work alright in more cases than `master`.

r3k2:
Thanks for working on  this! and happy new years, hopefully this works because I am almost close to ditch nyxt for luakit is been a pain ride for the last 2 months specially trying to enter 2FA tokens into websites etc.. I read above someone saying that VIM keys should be on when insert mode and no, only key that should be enable on insert mode is the one to get out of insert mode back to command mode picture a password with : or  lol I even having problems typing this now :( anyways lets hope this fix it thanks

aartaka:
Okay, the last set of commits should make it all much more coherent with regards to insert mode being enabled whenever intuitive. @Ambrevar, merging?

fictitiousexistence:
@aartaka Sorry just seeing this. Will try to try this morning. (I'm on the east coast us.)

fictitiousexistence:
@aartaka 
This did not work for me. 

Here's my steps to test in case I did something wrong. 

`git clone --branch vi-insert-on-input-fields --recurse-submodules https://github.com/atlas-engineer/nyxt.git`
`cd nyxt`
`make all`
`./nyxt -CI`

Went to https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_input_test

Pressed ctrl+space to select vi normal mode
Clicked with mouse in the input fields. 
Still in vi-normal

https://user-images.githubusercontent.com/20089581/212071868-2f06d5ee-c77f-44cd-94a0-66c979d5967a.mp4

aartaka:
That's because the inputs are hidden behind an IFrame on w3schools. Any other website it breaks on?

fictitiousexistence:
@aartaka 
Go figure, I chose the worse site for an example. 

I just tried on the 2 main problem sites I use and its working well.

aartaka:
Cool! I believe we're doing most of what other VI-inspired browsers do. Iframes are a bit beyond our capabilities there, unfortunately :(

@Ambrevar, merging then?

aadcg:
Can we wait for @Ambrevar to review this PR? I think waiting one week wouldn't hurt, sinceit will certainly land on 3.0 regardless.

source/changelog.lisp
```diff
@@ -608,6 +608,7 @@ regular commands, such as "
 
   (:h3 "Bug fixes")
   (:ul
+   (:li "VI insert mode is triggered in more cases where it should be triggered.")
```
aadcg, line 611: Suggestion: "Review logic relative to automatically triggering VI insert mode."

source/parenscript-macro.lisp
```diff
@@ -100,18 +100,20 @@
...
   "Whether ELEMENT is editable."
   `(let ((tag (chain ,element tag-name)))
-     (if (or (string= tag "INPUT")
+     (if (or (and (string= tag "INPUT")
+                  (not (chain ([] "hidden" "checkbox" "button") (includes (chain ,element type))))
+                  (not (chain ,element disabled)))
              (string= tag "TEXTAREA")
              (chain ,element is-content-editable))
-         t nil)))
+         t f)))
```
aadcg, line 108: Doesn't this change break other functionality? See `select-input-field` in `external-editor.lisp`.

aartaka: No, it makes it more reliable on the Lisp-side, while preserving the behavior on the JS side. I.e.:
```
| before this fix | JS                    | Lisp         |
|-----------------+-----------------------+--------------|
| true            | true (true-y value)   | t            |
| false           | null (false-y value)  | :null (bug!) |
|-----------------+-----------------------+--------------|
| after this fix  |                       |              |
|-----------------+-----------------------+--------------|
| true            | true (true-y value)   | t            |
| false           | false (false-y value) | nil          |
```

source/changelog.lisp
aartaka: That's so vague...

source/parenscript-macro.lisp
aadcg: Clear!

source/changelog.lisp
aadcg: As you wish.

Ambrevar (commented):
Excellent!

Just tune the doc a little bit and we are ready to go!

aartaka:
@Ambrevar, what about this move of the T/NIL returns into the command and skip dispatchers? I don't feel like it's necessary and nice anymore, so I'm interested to know what you think of it :)

source/mode/document.lisp
```diff
@@ -222,51 +222,15 @@ It does not assume being online."))
...
-                                    (ps:chain ,element (select))))
-         (dolist (mode (modes ,buffer))
-           (element-focused mode))))))
-
-(define-command focus-first-input-field (&key (type-blacklist '("hidden"
-                                                                "checkbox"
-                                                                "button")))
-  "Move the focus to the first input field of `buffer'."
-  ;; The list of input types can be found below.
-  ;; https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input
```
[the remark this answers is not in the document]
aartaka: Yes, moved to `nyxt/ps:element-editable-p`.

Ambrevar:
> @Ambrevar, what about this move of the T/NIL returns into the command and skip dispatchers? I don't feel like it's necessary and nice anymore, so I'm interested to know what you think of it :)

Why isn't it necessary anymore?  We need a way to tell whether to accept or forward the input, right?

If not with a return value, we could do it with conditions, but I'm not sure it would be much cleaner.

Any other idea?

aartaka:
> > @Ambrevar, what about this move of the T/NIL returns into the command and skip dispatchers? I don't feel like it's necessary and nice anymore, so I'm interested to know what you think of it :)
> 
> Why isn't it necessary anymore? We need a way to tell whether to accept or forward the input, right?

Well, it seems like all the other VI-keyed browsers simply look at the current element after:
- page loading,
- mouse click,
- and some contextual actions, like our `follow-hint`.

No one actually inspects things at every keypress, so 3a37ead81560d649344c2280e67ca3d2aaec5956 may've been a premature optimization that'd cause us pains afterwards :/

> If not with a return value, we could do it with conditions, but I'm not sure it would be much cleaner.

Conditions are underrated indeed, but let's first understand whether we actually need this feature in the first place :P

Ambrevar:
I think I don't understand what you tried to optimize with https://github.com/atlas-engineer/nyxt/commit/3a37ead81560d649344c2280e67ca3d2aaec5956.  Can you explain?

So if I get you right, you are saying that we don't need to inspect the page (to see whether an input tag is focused) on every key press?  If so, I agree!

Ambrevar:
> Well, it seems like all the other VI-keyed browsers simply look at the current element after:
> 
>     * page loading,
> 
>     * mouse click,
> 
>     * and some contextual actions, like our `follow-hint`.

Isn't it what we were doing before with the `element-focused` method?

aartaka:
> > Well, it seems like all the other VI-keyed browsers simply look at the current element after:
> > ```
> > * page loading,
> > * mouse click,
> > * and some contextual actions, like our `follow-hint`.
> > ```
> 
> Isn't it what we were doing before with the `element-focused` method?

Roughly, yes. But `focus-element` (the host of `element-focused` function) was not used anywhere, even in hint-mode. And now I've tried to make it more rooted into Nyxt by relying on `nyxt/dom` and its methods (which are used in much more places now).

So this PR is rather a set of more reliable and architectuarally sound heuristics than there used to be. And, based on the feedback of those using VI bindings, this PR accomplisshes exactly what's needed—reliable context-dependent VI insert mode toggling.

aartaka:
Okay, I removed the questionable commit with delegation of `dispatch-input-event` logic to `dispatch-command` and `dispatch-input-skip`. Now this PR should be a simple and easy fix for the problem it was intended to fix, without any architectural complexities introduced.

aartaka:
Merging? @Ambrevar, @aadcg?

source/mode/document.lisp
```diff
@@ -223,51 +223,19 @@ It does not assume being online."))
...
+  ;; - Using <input>/<textarea>,
+  ;; - or marking any other element as contenteditable:
+  ;;   https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable
+  (loop with inputs = (clss:ordered-select "input, textarea, [contenteditable]" (document-model buffer))
+        for input across inputs
+        when (ps-eval :buffer buffer
+               (nyxt/ps:element-editable-p
+                (nyxt/ps:qs-nyxt-id document (ps:lisp (nyxt/dom:get-nyxt-id input)))))
+          do (nyxt/dom:focus-select-element input)
+          and do (return input)))
```
aadcg, line 238: What about the logic introduces in commit c5527f690448fb02a5ace2b08d02d4429ab35721?

aadcg:
From my side, all good.

source/mode/document.lisp
```diff
@@ -223,51 +223,19 @@ It does not assume being online."))
...
-                   type-blacklist)
-                  i))))
-      (focus-element ()
-        (let* ((input (ps:chain document
-                                (get-elements-by-tag-name "INPUT")))
-               (item (when input (ps:chain input (item (ps:lisp i))))))
-          (when item
-            item))))))
+(define-command focus-first-input-field (&key (buffer (current-buffer)))
+  "Move the focus to the first input field of BUFFER ."
```
aadcg, line 227: Extra space before the period.

aartaka:
Cool, thanks!

Ambrevar:
Thanks!

PR #2740, opened 2023-01-11
start: Make start-browser synchronous.

Ambrevar:
# Description

There is no point in having it asynchronous, since anything run in parallel before browser is ready is likely to trigger a race conditions or undefined behaviour.

It simplifies testing a lot.

Fixes https://github.com/atlas-engineer/nyxt/pull/2718#discussion_r1064384172.

# Discussion

- [ ] Edit tests to not use `nyxt:after-startup-hook` since it's not needed anymore.
- [ ] Better name for the promise?  @aartaka ?
- [ ] I tried to see if we could generalize the concept, for instance come up with a macro that would automatically `synchronize` asynchronous functions.  It's difficult, because we don't really have a convention for these async functions and it heavily depends on WebKitGTK.  For instance, `buffer-load` is "finished" when the `on-signal-load-finished` signal is triggered, in which we call the `buffer-loaded-hook`.  This is completely different from `start-browser`.  Can we devise a convention somehow?

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [x] Documentation:
- [ ] Compilation and tests:


aartaka:
I like how it looks! But I want to see the changes to tests too, so that we see how it works in headless scenarios.

> * Better name for the promise?  @aartaka ?

No, seems fine. Along the lines of `after-startup-hook` etc.

> * I tried to see if we could generalize the concept, for instance come up with a macro that would automatically `synchronize` asynchronous functions.  It's difficult, because we don't really have a convention for these async functions and it heavily depends on WebKitGTK.  For instance, `buffer-load` is "finished" when the `on-signal-load-finished` signal is triggered, in which we call the `buffer-loaded-hook`.  This is completely different from `start-browser`.  Can we devise a convention somehow?

Hmmmmm. The case of `buffer-load` is even more complex, though: is it finished when `on-signal-load-finished` fires, or when all the resources are loaded? The latter is not hookable, although it's only then that the page is fully interactive and the loading is done, as far as user point of view is concerned...

So there seems to be no solid convention due to lack of clear meaning for things. That's the same problem I've encountered when trying to abstract our concurrency patterns: we don't have many patterns at all. There are lots of unique cases which are not abstractable because of their.

The only reliable option I can think of for this (a)synchronization is to store the promise/completion predicate inside the function itself. Completion predicate would mean less `lpara:fulfill` across the code and more autonomy, but more evaluation issues. In any case, storing things inside a function is complicated: MOP, our own function class, `define-function` macro as a sibling to `define-class` smartness. That's kinda scary, but curious in the possibilities it opens :D

aadcg:
If ``nyxt:start` is synchronous shouldn't the following work?  

```log
CL-USER> (progn (nyxt:start :failsafe t) (nyxt:quit))
Nyxt version 3-pre-release-2-812-g007b59d1b
<INFO> [11:10:28] Source location: #P"/home/aadcg/common-lisp/nyxt/"
<INFO> [11:10:28] Profile: "nofile"
<INFO> [11:10:28] GTK extensions directory: #P"/home/aadcg/common-lisp/nyxt/libraries/web-extensions/"
<INFO> [11:10:29] GTK extensions directory: #P"/home/aadcg/common-lisp/nyxt/libraries/web-extensions/"
<INFO> [11:10:29] Loading #P"/home/aadcg/.local/share/nyxt/history/default.lisp".
NIL
<WARN> [11:10:30] Warning: decide policy error: There is no applicable method for the generic function
                       #<STANDARD-GENERIC-FUNCTION NYXT/RENDERER/GTK::ON-SIGNAL-DECIDE-POLICY (1)>
                     when called with arguments
                       (#<WEB-BUFFER 1269 {100AF24363}>
                        #<CL-WEBKIT2:WEBKIT-NAVIGATION-POLICY-DECISION {100CB77AD3}>
                        :WEBKIT-POLICY-DECISION-TYPE-NAVIGATION-ACTION).
See also:
<WARN> [11:10:30] Warning: decide policy error: There is no primary method for the generic function
                       #<STANDARD-GENERIC-FUNCTION NYXT:BUFFERS (4)>
                     when called with arguments
                       (NIL).
See also:
  The ANSI Standard, Section 7.6.6.2
```

aadcg:
Also, I see little sense in having a separate PR for this. Can I move this commit and the thread to #2718? @aartaka

aartaka:
Hmmmm. I'd rather separate those two, because synchronous `start` is quite a set of concepts and problems. Would be beneficial to resolve this PR on its own to not bring more issues into the same context.

aadcg:
Ok, then I'll wait for developments on this PR.

Ambrevar:
> If `nyxt:start` is synchronous shouldn't the following work?

Yes it should.  Will investigate.

Ambrevar:
> There are lots of unique cases which are not abstractable because of their.

@aartaka Looks like you forgot a word :)

Ambrevar:
@aartaka Tests don't use `after-startup-hook` anymore.

aadcg:
Thanks @ambrevar. Ok to merge?

aartaka:
Neat! This PR looks just perfect to me now 🤩

Ambrevar:
Done!

aartaka:
Thanks, Pierre `(´∀｀)♡`

PR #2741, opened 2023-01-11
Make adjustable attribute widths optional (rebase & rework of #2686)

### LINKED ISSUE #2684: Prompt column width ratios

aartaka:
# Description

This is a second shot at #2686, with the column ratios being an opt-in feature.

I've also allowed myself a liberty of cutting out the latest set of commits that refused to compile dues to unmatched parentheses. I've tried to preserve most of the useful changes, though.

Fixes #2684

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -580,8 +580,9 @@ settings changes.")
    (:li "When started with " (:code "--remote") " and without " (:code "--quit")
         ", Nyxt now reads s-expression from standard input and sends it to the remote process. "
         "This avoids the performance penalty of a new process startup on each iteration.")
-   (:li "Prompt buffer attribute columns adjust to their content, allowing for
-a better overview of lengthy attributes."))
+   (:li "With " (:nxref :slot 'dynamic-attribute-width-p :class-name 'prompt-buffer)
+        " on, prompt buffer attribute columns adjust to their content, allowing
+for a better overview of lengthy attributes."))
 
   (:h3 "Bindings")
   (:ul
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -579,7 +579,9 @@ auto-mode-rules.lisp)."))
 settings changes.")
    (:li "When started with " (:code "--remote") " and without " (:code "--quit")
         ", Nyxt now reads s-expression from standard input and sends it to the remote process. "
-        "This avoids the performance penalty of a new process startup on each iteration."))
+        "This avoids the performance penalty of a new process startup on each iteration.")
+   (:li "Prompt buffer attribute columns adjust to their content, allowing for
+a better overview of lengthy attributes."))
 
   (:h3 "Bindings")
   (:ul
--- a/source/command-commands.lisp
+++ b/source/command-commands.lisp
@@ -18,7 +18,7 @@
     `(("Name" ,(string-downcase (closer-mop:generic-function-name command)))
       ("Bindings" ,(format nil "~{~a~^, ~}" bindings))
       ("Docstring" ,(or (first (sera::lines (documentation command 'function)))
-                        ""))
+                        "") nil 4)
       ("Mode" ,(let ((package-name (str:downcase (uiop:symbol-package-name (closer-mop:generic-function-name command)))))
                  (if (sera:in package-name "nyxt" "nyxt-user")
                      ""
@@ -35,7 +35,8 @@
                            (sort-by-time
                             (list-commands
                              :global-p (global-p source)
-                             :mode-symbols (mapcar #'sera:class-name-of (sera:filter #'enabled-p (modes (buffer source)))))))))
+                             :mode-symbols (mapcar #'sera:class-name-of (sera:filter #'enabled-p (modes (buffer source))))))))
+   (prompter:attributes-keys-default '("Name" "Bindings" "Docstring")))
   (:export-class-name-p t)
   (:accessor-name-transformer (class*:make-name-transformer name))
   (:documentation "Prompter source to execute commands.
@@ -115,7 +116,7 @@ Includes all commands and modes, and adds arbitrary Lisp functions on top of tha
     `(("Expression" ,(format nil "~s" extended-command))
       ("Arguments" ,(remove #\newline (format nil "~{~a~^ ~}" (arglist function))))
       ("Documentation" ,(or (first (sera::lines (documentation function 'function)))
-                            "")))))
+                            "") nil 4))))
 
 (defmethod prompter:object-attributes ((extended-command symbol) (source extended-command-source))
   (declare (ignore source))
@@ -124,10 +125,10 @@ Includes all commands and modes, and adds arbitrary Lisp functions on top of tha
         (let ((function (symbol-function extended-command)))
           `(("Expression" ,(format nil "~s" extended-command))
             ("Arguments" ,(remove #\newline (format nil "~{~a~^ ~}" (arglist function))))
-            ("Documentation" ,(or (first (sera::lines (documentation function 'function))) ""))))
+            ("Documentation" ,(or (first (sera::lines (documentation function 'function))) "") nil 4)))
         `(("Expression" ,(prini-to-string extended-command))
           ("Arguments" "")
-          ("Documentation" ,(or (documentation extended-command 'variable) ""))))))
+          ("Documentation" ,(or (documentation extended-command 'variable) "") nil 4)))))
 
 (define-command execute-command ()
   "Execute a command by name.
--- a/source/describe.lisp
+++ b/source/describe.lisp
@@ -94,19 +94,20 @@ See `sym:package-functions' for an example."
 (defmethod prompter:object-attributes ((symbol symbol) (source prompter:source))
   (declare (ignore source))
   `(("Name" ,(prini-to-string symbol))
+    ("Visibility" ,(prini-to-string (sym:symbol-visibility symbol)))
     ("Documentation"
      ,(or (cond
-            ((fboundp symbol)
-             (first-line (documentation symbol 'function)))
-            ((and (find-class symbol nil)
-                  (mopu:subclassp (find-class symbol) (find-class 'standard-object)))
-             (first-line (documentation symbol 'type)))
-            ((find-package symbol)
-             (first-line (documentation (find-package symbol) t)))
-            (t
-             (first-line (documentation symbol 'variable))))
-          ""))
-    ("Visibility" ,(prini-to-string (sym:symbol-visibility symbol)))))
+           ((fboundp symbol)
+            (first-line (documentation symbol 'function)))
+           ((and (find-class symbol nil)
+                 (mopu:subclassp (find-class symbol) (find-class 'standard-object)))
+            (first-line (documentation symbol 'type)))
+           ((find-package symbol)
+            (first-line (documentation (find-package symbol) t)))
+           (t
+            (first-line (documentation symbol 'variable))))
+          "")
+     nil 4)))
 
 (defmethod prompter:object-attributes ((package package) (source prompter:source))
   (declare (ignore source))
@@ -115,7 +116,7 @@ See `sym:package-functions' for an example."
                           ;; Old ASDF/UIOP don't know about package-local-nicknames.
                           (ignore-errors (uiop:symbol-call
                                           :uiop :package-local-nicknames package))))
-    ("Documentation" ,(or (first-line (documentation package t)) ""))))
+    ("Documentation" ,(or (first-line (documentation package t)) "") nil 4)))
 
 (define-class class-source (describe-nyxt-source)
   ((prompter:name "Classes")
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -277,6 +277,94 @@ See also `show-prompt-buffer'."
                                (mapcar (curry #'mode-status status-buffer)
                                        (sort-modes-for-status (modes prompt-buffer)))))))))
 
+(defun attribute-widths (source)
+  "Compute the widths of SOURCE attribute columns (as percent).
+Returns a list of integers, the sum of which should be roughly equal to 100.
+Uses the average length of attribute values to derive the width."
+  (flet ((ratio-widths (widths total)
+           (mapcar (lambda (w) (+ (/ 100 (length widths) 2)
+                                  (round (* 50 (if (zerop total)
+                                                   0
+                                                   (/ w total))))))
+                   widths)))
+    (loop with suggestions = (prompter:suggestions source)
+          ;; This is to process column width as fourth object attribute element.
+          initially (sera:and-let* ((fourth-attributes
+                                     (mapcar #'fourth (prompter:active-attributes (first suggestions) :source source)))
+                                    (some-fourth (some #'identity fourth-attributes))
+                                    (coefficients (substitute 1 nil fourth-attributes))
+                                    (total (reduce #'+ coefficients)))
+                      (return (ratio-widths coefficients total)))
+          for key in (prompter:active-attributes-keys source)
+          for width
+            = (loop for suggestion in suggestions
+                    sum (length (first (str:s-assoc-value
+                                        (prompter:active-attributes suggestion :source source) key))))
+          collect width into widths
+          sum width into total
+          finally (return (ratio-widths widths total)))))
+
+(defun render-attributes (source prompt-buffer)
+  (spinneret:with-html-string
+    (when (prompter:suggestions source)
+      (:table :class "source-content"
+              (:colgroup
+               (dolist (width (attribute-widths source))
+                 (:col :style (format nil "width: ~d%" width))))
+              (:tr :style (if (or (eq (prompter:hide-attribute-header-p source) :always)
+                                  (and (eq (prompter:hide-attribute-header-p source) :single)
+                                       (sera:single (prompter:active-attributes-keys source))))
+                              "display:none;"
+                              "display:revert;")
+                   (loop for attribute-key in (prompter:active-attributes-keys source)
+                         collect (:th (spinneret::escape-string attribute-key))))
+              (loop
+                ;; TODO: Only print as many lines as fit the height.
+                ;; But how can we know in advance?
+                with max-suggestion-count = 10
+                repeat max-suggestion-count
+                with cursor-index = (prompter:current-suggestion-position prompt-buffer)
+                for suggestion-index from (max 0 (- cursor-index (/ max-suggestion-count 2)))
+                for suggestion in (nthcdr suggestion-index (prompter:suggestions source))
+                collect
+                (let ((suggestion-index suggestion-index)
+                      (cursor-index cursor-index))
+                  (:tr :id (when (equal (list suggestion source)
+                                        (multiple-value-list (prompter:%current-suggestion prompt-buffer)))
+                             "selection")
+                       :class (when (prompter:marked-p source (prompter:value suggestion))
+                                "marked")
+                       :onmousedown (when (mouse-support-p prompt-buffer)
+                                      (ps:ps
+                                       (lambda (event)
+                                         (nyxt/ps:lisp-eval
+                                          (:title "choose-this-suggestion"
+                                                  :buffer prompt-buffer)
+                                          (prompter::set-current-suggestion
+                                           (current-prompt-buffer)
+                                           (- suggestion-index cursor-index)))
+                                         (cond
+                                           ;; Ctrl-click to mark a single suggestion.
+                                           ;; TODO: Shift-click to mark a region.
+                                           ((or (ps:@ event meta-key)
+                                                (ps:@ event ctrl-key))
+                                            (nyxt/ps:lisp-eval
+                                             (:title "mark-this-suggestion"
+                                                     :buffer prompt-buffer)
+                                             (funcall (read-from-string "nyxt/prompt-buffer-mode:toggle-mark")
+                                                      prompt-buffer)))
+                                           (t (nyxt/ps:lisp-eval
+                                               (:title "return-this-suggestion"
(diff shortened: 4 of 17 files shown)


lansingthomas:
Ciao Atlasers,
John and I had a productive conversation yesterday about this attribute width challenge. Sorry it has taken me song long to weigh in. Hopefully this plan will be effective -- we have tried to be thoughtful. I notice that some code has been written.  I want to recognize that we are all working hard on this. Thank you for your diligence and patience. 

---

### John/Thomas Discussion Overview:
#### we discussed possible plans for the attribute widths, and agreed on some priorities.
- I will elaborate some solution requirements
- I will elaborate the plans we discussed
- I will run you those plans as well on our calls tomorrow (@aartaka @aadcg ) :)

---

### Understanding that any solution will have some requirements...
Potential Solutions:
Any solution MUST have at least **one fixed width column** at the left alignment of the primary attributes to maintain some nice visual continuity across prompt buffers to help users recognize the structure across use cases.
- I recommend a primary attribute width that accommodates 40 characters for `Name`/`URL`/`Hint` (which is right about where the width is is now)
   - reasoning: the longest `Name` attribute in the Execute Command Prompt Area is ~41 characters, with the significant majority below 30 characters. 
   - a typical URL has 8 chars for the prefix (https://), ~20 chars for 1-3 words, and 5 chars for the first navigation suffix (/home or /about) -- all that sums to ~33 words, so the length (40 characters) is enough plus some space to display part of longer URLs.
   - With `Hints` the eyeball focus is on the webpage first, and the attributes second, so the void created by the extra width in this primary attribute column is fine.
   
Any solution SHOULD have: fixed width for some secondary attributes when they appear (like `Bindings` -- very common across use cases, easily predictable # of characters)
- immediate recognition for `bindings` is important -- as the user will not be looking at `bindings` unless they are trying to do something fast.
- shortening this attribute column width creates space for the longer columns where we may want more space.
- I recommend that `Bindings` accommodate  

A future solution COULD have: dynamic widths for longer secondary attributes like `Docstring`
- in this context a solution MUST calculate the width once and display it, so that width does not change while scrolling.
- also in this context a solution SHOULD constrain the longer descriptions and suggestions to minimize the need for visual real space in the UI. -(perhaps with a max character length, or similar)

---

### we discussed some possibilities...

1. STATIC OPTION: totally static attribute column widths.
- specific widths for each attribute column
   - a two attribute prompt area would follow  a 3-7 (30%, 70%) width structure
   - **A three attribute prompt area would follow a 3-2-5 width structure**
      - (as in 30% available width to attribute-column 1, 20% to attribute-column 2, and 50% to attribute-column 3)
      - Fibonacci would do it, 2-3-5 but he did not have URLs getting cut off too early so here we are.
   - a four attribute prompt area would revert to 25%, 25%, 25%, 25%
- this approach accommodates the ~40 character max for the first attribute
- this approach fits the `bindings` in column two structure very nicely
- this approach allows for longer descriptions in the third column
- this approach keeps our attribute width the same across all use cases which pleases my visual brain and satisfies subtle user expectations
- this approach vertically aligns the first attribute when multiple sources are displayed. 
 
URL attribute width image


2. DYNAMIC OPTION: dynamic across all use cases
- each instance of the prompt area calculates widths for the attributes based on character length and displays
- this approach has some problems with continuity and alignment. We can do better!
- John and I at least agreed that any changes to column width DURING scrolling is NOT okay because it blasts skimmability, and if we are willing to do that we have better options (like text wrapping).

3. STATIC + DYNAMIC OPTION
- Static column width for the first attribute, across all use cases.
- A mix of fixed column widths and dynamic column widths for attributes 2-n.
- this approach keeps the anchor point of a common width for the first column across prompt area uses cases 
- this approach has the potential to be more extensible
- this approach would take more time and effort

### I am calling for alignment on option 1 - STATIC OPTION
- it is better than the current version
- we have a shared time/effort priority
- we have good opportunity to re-arrange the display order of the attributes for each common use case after deploying anything, and this may be a blocker for some other UI decisions. 
- it is a great starting place even IF we decide to go with option 3 - STATIC + DYNAMIC
-  Of course we will talk about it tomorrow on call

---

Here are some images of STATIC OPTION:
2 attributes
3 attributes
STATIC OPTION with 4 attributes

lansingthomas:
okay!
I have spoken with @jmercouris, @aartaka, & @aadcg and we have alignment on the first option:
STATIC OPTION

See above for details.

Thank you

aadcg:
@lansingthomas great job at fitting all of the pieces of the puzzle together!  I think your specification makes complete sense.

aartaka:
Static option is on by default (in 8e9ddbe51e74b37ecfd42acb0e183c3f2e813566) with dynamic one behind a `dynamic-attribute-width-p' toggle. I've also added lots of more-or-less intuitive static widths for most sources in Nyxt in 1d4dd57250e1db22f90ccc938281bd9cb6a6ae7c.

aartaka:
I'll fix the CI errors next week. In the meanwhile, anyone willing to review the code?

aadcg:
> I'll fix the CI errors next week. In the meanwhile, anyone willing to
> review the code?

I'll take a look at this next week.

aadcg:
@aartaka I can't even compile this checkout.

```log
; file: /home/aadcg/common-lisp/nyxt/source/prompt-buffer.lisp
; in: DEFGENERIC ATTRIBUTE-WIDTHS
;     (RETURN (NYXT::WIDTH-NORMALIZATION NYXT::COEFFICIENTS))
; 
; caught ERROR:
;   return for unknown block: NIL
```

aadcg:
Also, @lansingthomas specified that:

> a two attribute prompt area would follow a 3-7 (30%, 70%) width structure
> a three attribute prompt area would follow a 3-2-5 width structure

But the prompt buffer invoked by the `set-url` command doesn't follow the specification.

lansingthomas:
@aartaka thank you for your quick deployment!

I think we are still using the wrong tool for the job. There should not be a need for calculating these average width values right now.

### Please set the widths to static values like this:

#### For Source Headings with one Attribute Column; 
- the column width covers 100% of our column width area

#### For Source Headings with two Attribute Columns;
- the first column width covers 30% of the available attribute column width
- some gutter exists in between the two columns
- the remainder of the available attribute column width is for the second attribute column (70% minus the gutter)

#### For Source Headings with three Attribute Columns;
- 30% width goes to the first attribute
- gutter
- 20% width goes to the second
- gutter
- 50% (minus the gutters) goes to the third attribute column

#### For Source Headings with four attribute columns
- 30% width goes to the first attribute
- gutter
- 20% width goes to the second
- gutter
- 20% goes to the third attribute
- gutter
- 50% (minus the gutters) goes to the fourth attribute

#### For Source Headings with greater than four attributes.
- equal width allocated for each column.

lansingthomas:
@aartaka I appreciate you going through the sources like this: I think the table you shared will inform future decisions for this. Right now I have no way of understanding the sources in the context of use -- Which means that I may be asking you to change some of these again soon. So for now we really want the columns to line up nicely in our most common use cases.
Thank you

aartaka:
> * which sources are these from the two attribute bucket?
>   1, 4

`autofill-source` for `autofill` command.

>   1, 3

Search engine source for `query-selection-in-search-engine`.

>   1, 3

`window-source` for `delete-window`.

>   5, 1

`new-url-or-search-source` (the one with the search suggestions) for `set-url`.

lansingthomas:
from: Artyom & Thomas' conversation -- we agreed on some excellent attribute column widths for Artyom to implement. We arrived at these choices by going through many of the cases and balancing functionality with alignment across prompt buffers.

--- 

## the rules:

### for the 2 attribute source 
1,1 at 50% -50%
for the rest 1, 3 
the lone 5-1 is the exception

### for the 3 attribute sources
- 3,2,1 for most cases
- 1,1,4 for others (1,4,1 --> 1,1,4)
- 1,1,1, is the exception
- for remembrance (3,2,2 try 3,2,1 -- make a choice) 

### for the special case of `execute command`
- change the default display to a 3 attribute version (by removing mode)

### for the four attribute sources
- standard is either 1,1,1,1 or 4,1,1,2 (whichever seems to fit best for each case)
- the 4,x,x,x we want to change to 3,x,x,x, (like bookmarks) 

### the date attribute
- readability issues xD

aartaka:
Adjusted all the attributes to the spec and rebased over master. Should be mergeable now.

aartaka:
Merging?

aartaka:
@Ambrevar, care to review yet another time? Or is it mergeable already?

source/changelog.lisp
```diff
@@ -579,7 +579,10 @@ auto-mode-rules.lisp)."))
 settings changes.")
    (:li "When started with " (:code "--remote") " and without " (:code "--quit")
         ", Nyxt now reads s-expression from standard input and sends it to the remote process. "
-        "This avoids the performance penalty of a new process startup on each iteration."))
+        "This avoids the performance penalty of a new process startup on each iteration.")
+   (:li "With " (:nxref :slot 'auto-attribute-width-p :class-name 'prompt-buffer)
+        " on, prompt buffer attribute columns adjust to their content, allowing
+for a better overview of lengthy attributes."))
```
Ambrevar, line 585: Maybe explain why it's not on by default.

source/command-commands.lisp
```diff
@@ -18,11 +18,7 @@
     `(("Name" ,(string-downcase (closer-mop:generic-function-name command)))
       ("Bindings" ,(format nil "~{~a~^, ~}" bindings))
       ("Docstring" ,(or (first (sera::lines (documentation command 'function)))
-                        ""))
-      ("Mode" ,(let ((package-name (str:downcase (uiop:symbol-package-name (closer-mop:generic-function-name command)))))
-                 (if (sera:in package-name "nyxt" "nyxt-user")
-                     ""
-                     (str:replace-first "nyxt/" "" package-name)))))))
```
Ambrevar, line 25: Displaying the mode is critical since there may be duplicate command names.

If you really think it takes too much space, hide it by default, but leave it available for display for those who want it.

source/describe.lisp
```diff
@@ -94,19 +94,20 @@ See `sym:package-functions' for an example."
...
-             (first-line (documentation symbol 'function)))
-            ((and (find-class symbol nil)
-                  (mopu:subclassp (find-class symbol) (find-class 'standard-object)))
-             (first-line (documentation symbol 'type)))
-            ((find-package symbol)
-             (first-line (documentation (find-package symbol) t)))
-            (t
-             (first-line (documentation symbol 'variable))))
-          ""))
-    ("Visibility" ,(prini-to-string (sym:symbol-visibility symbol)))))
```
Ambrevar, line 109: Same here, you can leave "visibility" which is very useful, but hide it by default.

source/changelog.lisp
```diff
@@ -579,7 +579,10 @@ auto-mode-rules.lisp)."))
 settings changes.")
    (:li "When started with " (:code "--remote") " and without " (:code "--quit")
         ", Nyxt now reads s-expression from standard input and sends it to the remote process. "
-        "This avoids the performance penalty of a new process startup on each iteration."))
+        "This avoids the performance penalty of a new process startup on each iteration.")
+   (:li "With " (:nxref :slot 'auto-attribute-width-p :class-name 'prompt-buffer)
```
Ambrevar, line 583: Isn't it "dynamic-attribute-width-p"?

tests/offline/prompt-buffer.lisp
```diff
@@ -0,0 +1,51 @@
...
+  (defmethod prompter:object-attributes ((object string) (source test-source))
+    (loop for i from 1 upto length
+          collect (list (str:concat "Attribute" (princ-to-string i)) (str:repeat i object) nil i))))
+
+(defun random-in-range (start end)
+  (+ start (random (+ 1 (- end start)))))
+
+(define-test constant-size-attributes ()
+  (let ((n-attributes (random-in-range 1 10)))
+    ;; How to shup up the fact that the method is being redefined?
```
Ambrevar, line 27: Typo: `shup`.

`defmethod` gives a warning, so either use 

```lips
(handler-bind ((warning #'ignore-warning))
     ...))
```

or use `c2cl:ensure-method`.

tests/offline/prompt-buffer.lisp
```diff
@@ -0,0 +1,51 @@
...
+    (set-test-source-object-attributes n-attributes)
+    (assert-equal (loop for i from 1 upto n-attributes
+                        sum i into total
+                        collect i into widths
+                        finally (return (mapcar (rcurry #'/ total) widths)))
+                  (nyxt::attribute-widths (make-instance 'test-source)))))
+
+(define-test dynamic-widths ()
+  (let (ratio)
+    (setf ratio 2)
```
Ambrevar, line 37: Can set `ratio` in the `let`.

tests/offline/prompt-buffer.lisp
```diff
@@ -0,0 +1,51 @@
...
+    (set-test-source-object-attributes ratio)
+    (assert-equal
+     (list 1/3 2/3)
+     (nyxt::attribute-widths (make-instance 'test-source) :dynamic-p t))
+    (setf ratio 3)
```
Ambrevar, line 42: I would let-bind `ratio` each time, to make it more functional.

source/prompt-buffer.lisp
```diff
@@ -54,6 +54,14 @@ some point.")
       :type boolean
       :documentation "Whether to allow mouse events to set and return the
 current suggestion in the prompt buffer.")
+     (dynamic-attribute-width-p
+      nil
+      :type boolean
+      :documentation "Whether to adjust the widths of columns to the content in it.
```
Ambrevar, line 60: "Whether to adjust the column widths to their content."

source/prompt-buffer.lisp
```diff
@@ -277,14 +285,138 @@ See also `show-prompt-buffer'."
...
+                finally (return (clip-extremes (width-normalization widths))))
+          (sera:and-let* ((suggestions (prompter:suggestions source))
+                          (fourth-attributes
+                           (mapcar #'fourth (prompter:active-attributes (first suggestions) :source source)))
+                          (coefficients (substitute 1 nil fourth-attributes)))
+            (width-normalization coefficients)))))
+  (:documentation "Compute the widths of SOURCE attribute columns (as percent).
+
+If DYNAMIC-P, compute widths based on content length. Otherwise, rely on the
+fourth value of `prompter:object-attributes' for the given source.
```
Ambrevar, line 343: This should be documented in the `object-attributes` `defgeneric`.

source/prompt-buffer.lisp
```diff
@@ -277,14 +285,138 @@ See also `show-prompt-buffer'."
...
+      ;; FIXME: This loop ignores attribute names in the attribute width
+      ;; computation. Because of this, attribute names could theoretically get
+      ;; cropped if the values are short enough. This is bad, but not exactly
+      ;; critical.
+      (if dynamic-p
+          (loop with suggestions = (prompter:suggestions source)
+                with attributes = (mapcar (rcurry #'prompter:active-attributes :source source) suggestions)
+                for key in (prompter:active-attributes-keys source)
+                for width
+                  = (funcall width-function (mapcar (compose #'first (rcurry #'str:s-assoc-value key))
```
Ambrevar, line 331: `funcall*`?  What if the function is not provided?

aadcg:
@lansingthomas did you validate @aartaka's implementation?

I remind that @lansingthomas isn't an engineer, so you should probably schedule a screen sharing session to go over the changes.

aadcg:
I don't think the logic behind setting the width manually is correct. Here's an example: 3 attributes. Set the first to 200 and the second to 400. The third one is assigned 1 (since a value wasn't set explicitly) so the ratios become 200/601, 400/601 and 1/601.  

I think there are two approaches to handle it: 

1. Check whether width values were assigned for all attributes. When an attribute has a missing width value, fallback into uniform distribution for all. In this setting, it's not necessary to enforce values within certain bounds since they will be normalized.
2. Enforce width values within certain bounds (say [0,1] or [0,100]). Uniformly distribute over the attributes whose width was not specified.

aartaka:
> @lansingthomas did you validate @aartaka's implementation?
> 
> I remind that @lansingthomas isn't an engineer, so you should probably schedule a screen sharing session to go over the changes.

That's exactly what we did, more than once. We've gone through lots of sources and Tom has certified that these ratios are good.

aadcg:
> But still, you're stretching it too far, I believe. All the attribute widths are not more than 5 at the moment, and they are unlikely to get much higher unless one (user or extension writer) actually wants it, in which case we're not the ones to set obstacles.

The issue I'm raising is not related to the number of attributes.

Run the following recipe and inspect how the width is set in the prompt.

```lisp
(define-class test-source (prompter:source)
  ((prompter:name "Test")
   (prompter:constructor '("suggestion"))))

(defmethod prompter:object-attributes ((object string) (source test-source))
  `(("Attribute-1" ,object nil 50)
    ("Attribute-2" ,object)))

(prompt :prompt "test" :sources 'test-source)
```

Besides 50, you can also try 1 and 0.5. 

What conclusions do you draw?

aartaka:
> > But still, you're stretching it too far, I believe. All the attribute widths are not more than 5 at the moment, and they are unlikely to get much higher unless one (user or extension writer) actually wants it, in which case we're not the ones to set obstacles.
>
> The issue I'm raising is not related to the number of attributes.
> 
> Run the following recipe and inspect how the width is set in the prompt.
> 
> ```lisp
> (define-class test-source (prompter:source)
>   ((prompter:name "Test")
>    (prompter:constructor '("suggestion"))))
> 
> (defmethod prompter:object-attributes ((object string) (source test-source))
>   `(("Attribute-1" ,object nil 50)
>     ("Attribute-2" ,object)))
> 
> (prompt :prompt "test" :sources 'test-source)
> ```
> 
> Besides 50, you can also try 1 and 0.5.
> 
> What conclusions do you draw?

Same conclusions as in my previous comment. Sorry if it was sloppily worded. Current sources don't have width ratios (not the number of attribute widths, but widths themselves) above 5, and are unlikely to ever have. Thus hypothesizing that widths may be 50, 200, or 10000 is irrelevant. The current implementation is alright (except for comments Pierre made, and I'm fixing those right now).

source/command-commands.lisp
aartaka: With 03a247fbd95ccfc2792922a74c7144e70df5e6c0 and 681e0844cde02127a275f9683f86f6d96ee6bad4, the relevance of "Modes" column is uch lower. But yes, I'll retain it and hide by default.

source/changelog.lisp
aartaka: We didn't have a consensus about this feature, and that's the only solid reason for it not being a default. Other reasons are arguable, so let's leave it as is :)

aadcg:
> Current sources don't have width ratios (not the number of attribute widths, but widths themselves) above 5, and are unlikely to ever have. Thus hypothesizing that widths may be 50, 200, or 10000 is irrelevant.

I honestly do not understand.

What's the meaning of the numbers in [the table above](https://github.com/atlas-engineer/nyxt/pull/2741#issuecomment-1384236865)?

source/prompt-buffer.lisp
aartaka: But wait, it cannot be, because `object-attributes' `defgeneric' explicitly states that the attributes after the second one are application specific. I am not sure where to put the values of all the attributes (manual? some commentary in the code of Nyxt? commentary in `prompter'?), so let's have at least this.

source/prompt-buffer.lisp
aartaka: If it's not provided, it's an exceptional situation that throws up a condition. This function should be provided.

aartaka:
> What's the meaning of the numbers in [the table above](https://github.com/atlas-engineer/nyxt/pull/2741#issuecomment-1384236865)?

Proportions of the attributes, i.e. how much of the common space of the prompt-buffer they occupy.  For example, if the proportions/ratios are `4-1-1-2`, then it means that out of absolutely equal and arbitrary 8 (4+1+1+2) units of widths, first attribute takes up 4, second and third—one each, and fourth—2.

Which means that the first one takes up 50% of the width, second and third take 12.5% each, and fourth takes 24%.

source/prompt-buffer.lisp
Ambrevar: Oh right.

aartaka:
So, anything I missed? CI is passing, things discussed with @lansingthomas, @Ambrevar comments addressed. Merging?

tests/offline/prompt-buffer.lisp
```diff
@@ -0,0 +1,48 @@
...
+  (defmethod prompter:object-attributes ((object string) (source test-source))
+    (loop for i from 1 upto length
+          collect (list (str:concat "Attribute" (princ-to-string i)) (str:repeat i object) nil i))))
+
+(defun random-in-range (start end)
+  (+ start (random (+ 1 (- end start)))))
+
+(define-test constant-size-attributes ()
+  (let ((n-attributes (random-in-range 1 10)))
+    ;; How to shut up the fact that the method is being redefined?
```
Ambrevar, line 27: Spurious comment now, it seems.

Ambrevar:
OK for me, see my last comment.

aartaka:
Squashed and merged as 8d71ac1917d45da4d8ab500e5bdc2d1df4d5e68e!

PR #2764, opened 2023-02-02
libraries/class-star: Infer type from parent slot if any.

### LINKED ISSUE #2715: Mode tests fail with CCL

Ambrevar:
# Description

This changes class-star to infer type from inherited slot, if any.

Fixes #2715 and supersedes #2716.

# Discussion

This may not do the right thing if the parent class is not finalized.  For now I opted to just ignore unfinalized classes.

Any idea if we can do better about this? @aadcg @aartaka 

# To do:

- [x] Merge #2765 first.
- [x] Protect against unfinalized classes.  Problem can be seen when loading `prompter`.
- [ ] Ensure bug from #2715 is fixed.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/libraries/class-star/class-star.lisp
+++ b/libraries/class-star/class-star.lisp
@@ -156,16 +156,23 @@ The predicate returns non-nil when the argument is of the `name' class.")
   (declare (ignore args))
   (intern (format nil "~a-P" name) *package*))
 
-(defun inherited-slot-type (slot-name superclasses)
-  (let ((parent (find slot-name superclasses
-                      :test (lambda (slot class)
-                              ;; REVIEW: We ignore unfinalized classes, is it wise?
-                              (find slot (ignore-errors (mopu:slot-names class)))))))
-    (when parent
-      (values
-       (getf (mopu:slot-properties parent slot-name)
-             :type)
-       parent))))
+(defun slot-type-maybe-inherited (definition type-inference superclasses)
+  (let ((slot-name (first definition))
+        (all-parents-finalized-p
+          (and
+           (every (lambda (s) (find-class s nil)) superclasses)
+           (every #'closer-mop:class-finalized-p
+                  (mapcar (lambda (s) (find-class s nil)) superclasses)))))
+    (cond
+      ((and all-parents-finalized-p
+            (find slot-name (apply #'append (mapcar #'mopu:slot-names superclasses))))
+       (let ((parent (find (list slot-name) superclasses
+                           :key #'mopu:slot-names :test #'intersection)))
+         (when parent
+           (getf (mopu:slot-properties parent slot-name) :type))))
+      (all-parents-finalized-p
+       (funcall type-inference definition))
+      (t nil))))
 
 (defun process-slot-initform (definition &key ; See `hu.dwim.defclass-star:process-slot-definition'.
                                            superclasses
@@ -182,13 +189,9 @@ The predicate returns non-nil when the argument is of the `name' class.")
           (not (definition-type definition))
           type-inference)
      (append definition
-             (multiple-value-bind (type parent)
-                 (inherited-slot-type (first definition) superclasses)
-               (if parent               ; Found parent slot, type possibly nil.
-                   (list :type type)
-                   (let ((result-type (funcall type-inference definition)))
-                     (when result-type
-                       (list :type result-type)))))))
+             (let ((type (slot-type-maybe-inherited
+                          definition type-inference superclasses)))
+               (when type (list :type type)))))
     ((and (not (initform definition))
           initform-inference)
      (append definition
--- a/libraries/class-star/class-star.lisp
+++ b/libraries/class-star/class-star.lisp
@@ -156,7 +156,18 @@ The predicate returns non-nil when the argument is of the `name' class.")
   (declare (ignore args))
   (intern (format nil "~a-P" name) *package*))
 
+(defun inherited-slot-type (slot-name superclasses)
+  (let ((parent (find slot-name superclasses
+                      :test (lambda (slot class)
+                              (find slot (mopu:slot-names class))))))
+    (when parent
+      (values
+       (getf (mopu:slot-properties parent slot-name)
+             :type)
+       parent))))
+
 (defun process-slot-initform (definition &key ; See `hu.dwim.defclass-star:process-slot-definition'.
+                                           superclasses
                                            initform-inference
                                            type-inference)
   (unless (consp definition)
@@ -170,9 +181,13 @@ The predicate returns non-nil when the argument is of the `name' class.")
           (not (definition-type definition))
           type-inference)
      (append definition
-             (let ((result-type (funcall type-inference definition)))
-               (when result-type
-                 (list :type result-type)))))
+             (multiple-value-bind (type parent)
+                 (inherited-slot-type (first definition) superclasses)
+               (if parent               ; Found parent slot, type possibly nil.
+                   (list :type type)
+                   (let ((result-type (funcall type-inference definition)))
+                     (when result-type
+                       (list :type result-type)))))))
     ((and (not (initform definition))
           initform-inference)
      (append definition
@@ -242,6 +257,8 @@ argument, right after the slot name.
 
 The same applies to the types with the `:type-inference' option, the
 `*type-inference*' default and the `:type' argument respectively.
+If the slot is found in a superclass, the inferred type is then that of the
+parent slot.
 
 If `:predicate-name-transformer' is non nil (defaults to
 `*predicate-name-transformer*'), a predicate of the resulting name is generated.
@@ -278,6 +295,7 @@ If `:export-predicate-name-p' is non-nil, which defaults to
            ,(mapcar (lambda (definition)
                       (process-slot-initform
                        definition
+                       :superclasses supers
                        :initform-inference initform-inference
                        :type-inference type-inference))
              slots)
--- a/libraries/class-star/class-star.lisp
+++ b/libraries/class-star/class-star.lisp
@@ -159,7 +159,8 @@ The predicate returns non-nil when the argument is of the `name' class.")
 (defun inherited-slot-type (slot-name superclasses)
   (let ((parent (find slot-name superclasses
                       :test (lambda (slot class)
-                              (find slot (mopu:slot-names class))))))
+                              ;; REVIEW: We ignore unfinalized classes, is it wise?
+                              (find slot (ignore-errors (mopu:slot-names class)))))))
     (when parent
       (values
        (getf (mopu:slot-properties parent slot-name)
--- a/libraries/class-star/tests/tests.lisp
+++ b/libraries/class-star/tests/tests.lisp
@@ -97,3 +97,27 @@
              (getf (mopu:slot-properties 'foo-type-infer 'fun) :type))
   (assert-eq nil
              (getf (mopu:slot-properties 'foo-type-infer 'composite) :type)))
+
+(define-test inherited-type-inference ()
+  (class-star:define-class parent ()
+    ((name "foo")
+     (age nil
+          :type (or null number))))
+  (class-star:define-class child (parent)
+    ((name nil
+           :type (or null string))
+     (age 17)))
+  (let ((p (make-instance 'parent))
+        (c (make-instance 'child)))
+    ;; Perform some  assignments: CCL can catch type errors here.
+    (setf (slot-value p 'age) 18)
+    (setf (slot-value c 'name) "bar")
+    (setf (slot-value c 'age) nil)
+    (assert-eq 'string
+               (getf (mopu:slot-properties 'parent 'name) :type))
+    (assert-equal '(or null number)
+                  (getf (mopu:slot-properties 'parent 'age) :type))
+    (assert-equal '(or null string)
+                  (getf (mopu:slot-properties 'child 'name) :type))
+    (assert-equal '(or null number)
+                  (getf (mopu:slot-properties 'child 'age) :type))))
--- a/libraries/class-star/tests/tests.lisp
+++ b/libraries/class-star/tests/tests.lisp
@@ -98,26 +98,51 @@
   (assert-eq nil
              (getf (mopu:slot-properties 'foo-type-infer 'composite) :type)))
 
+(class-star:define-class parent ()
+  ((name nil
+         :type (or null string))
+   (age nil
+        :type (or null number))))
+
 (define-test inherited-type-inference ()
-  (class-star:define-class parent ()
-    ((name "foo")
-     (age nil
-          :type (or null number))))
   (class-star:define-class child (parent)
-    ((name nil
-           :type (or null string))
-     (age 17)))
-  (let ((p (make-instance 'parent))
-        (c (make-instance 'child)))
-    ;; Perform some  assignments: CCL can catch type errors here.
-    (setf (slot-value p 'age) 18)
-    (setf (slot-value c 'name) "bar")
-    (setf (slot-value c 'age) nil)
-    (assert-eq 'string
-               (getf (mopu:slot-properties 'parent 'name) :type))
-    (assert-equal '(or null number)
-                  (getf (mopu:slot-properties 'parent 'age) :type))
-    (assert-equal '(or null string)
+    ((name "foo" :type string)          ; inherited.
+     (age 17)                           ; inherited.
+     (address "Dummy address")))
+  (let ((c (make-instance 'child)))
+    ;; Perform some assignments: CCL can catch type errors here.
+    (setf (slot-value c 'age) 15)
+    #+ccl
+    (assert-error 'error
+                  (setf (slot-value c 'name) nil))
+    (assert-equal 'string
                   (getf (mopu:slot-properties 'child 'name) :type))
-    (assert-equal '(or null number)
-                  (getf (mopu:slot-properties 'child 'age) :type))))
+    ;; Given that the parent may or may not be finalized by then, we have to way
+    ;; to test it but checking it falls into a reliable set of options (NIL =
+    ;; parent not finalized, type = parent finalized).
+    (assert-true (member (getf (mopu:slot-properties 'child 'age) :type)
+                         (list nil '(or null number))
+                         :test #'equal))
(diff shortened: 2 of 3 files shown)


Ambrevar:
The newly introduced tests are failing the first time, but pass when run again:

```lisp
(define-test inherited-type-inference ()
  (class-star:define-class parent ()
    ((name "foo")
     (age nil
          :type (or null number))))
  (class-star:define-class child (parent)
    ((name nil
           :type (or null string))
     (age 17)))
  (let ((p (make-instance 'parent))
        (c (make-instance 'child)))
    ;; Perform some  assignments: CCL can catch type errors here.
    (setf (slot-value p 'age) 18)
    (setf (slot-value c 'name) "bar")
    (setf (slot-value c 'age) nil)
    (assert-eq 'string
               (getf (mopu:slot-properties 'parent 'name) :type))
    (assert-equal '(or null number)
                  (getf (mopu:slot-properties 'parent 'age) :type))
    (assert-equal '(or null string)
                  (getf (mopu:slot-properties 'child 'name) :type))
    (assert-equal '(or null number)
                  (getf (mopu:slot-properties 'child 'age) :type))))
```

This is because the first time the parent class is not finalized and thus the child `age` slot get its type inferred to `integer`.

I'm unsure we can fix this at the macro level.  Of course MOP can do this, but the point of this library is to not enforce any metaclass (otherwise you wouldn't be able to use class-star over your own metaclasses).

Any other idea?
@aartaka

Ambrevar:
I'm tempted to ditch this pull request and simply add a warning in the readme of class-star and in the docstring of the type inference option.

Thoughts?  @aadcg @aartaka

aartaka:
Hmmmmmmm. I'm thinking of several heuristics for non-finalized classes there:
- If the slot name has a different `symbol-package` from the current package, don't infer.
- If there's only an initform (no doc, no type, no allocation), don't infer.
- If neither of those is certain, add a one-shot `initialize-instance :after` for an instance of this class to decide the type in the runtime.
- Or, if the parent classes are not finalized yet, don't infer anything at all...

I'll be extremely sorry if this useful change does not end up included into class-start, because it's good to rely on parent classes when those are available!

Ambrevar:
> I'll be extremely sorry if this useful change does not end up included into class-star, because it's good to rely on parent classes when those are available!

Yes, but it's worth to have some code yield 2 different evaluations when reevaluated :(

aartaka:
> > I'll be extremely sorry if this useful change does not end up included into class-star, because it's good to rely on parent classes when those are available!
> 
> Yes, but it's worth to have some code yield 2 different evaluations when reevaluated :(

That's why I'm suggesting these heuristics: to have the first evaluation to be less erroneous, while having the second to be more correct. Yes, we're playing with image state here, and it seems inevitable for such a case, but we can still make it at least useful.

aartaka:
Yet another crazy idea: if the parent classes are not finalized, generate a `satisfies` type that will check those parent classes when invoked and either return `t` if they are not finalized yet, or return the actual type when those are finalized.

aartaka:
But, in general, I'd go with:
- If parents are finalized, look at those and infer the type.
- If parents are not finalized, don't infer the types of slots at all.

This would make the behavior consistent enough to not:
- cause wrong type inference (by refusing to infer anything in the absence of reliable information), and
- cause reproducibility problems (the type is either correct or not there).
----------------------------------------------------------------
As for the tests, how about structuring it this way:
- First, define a parent class.
- Then, define a child class.
- Check the type inference being non-existent and parent class being finalized.
- Then, finalize the parent class and re-define the child one.
- Check the type inference firing and parent class being finalized.
- Et voila!

I mean, the fact that classes get finalized in between test runs is a matter of test cleanup, not the matter of library working the way it should, right?

Ambrevar:
> This would make the behavior consistent enough to not:
> 
>     * cause wrong type inference (by refusing to infer anything in the absence of reliable information), and
> 
>     * cause reproducibility problems (the type is either correct or not there).

I don't understand how this would fix the reproducibility problem.  On second evaluation, the parent is finalized and thus the slot type would _change_.

How can we prevent that?

The heursitic you provided above does not deal with unfinalized parents (beside the last point of course).

aartaka:
> > This would make the behavior consistent enough to not:
> > ```
> > * cause wrong type inference (by refusing to infer anything in the absence of reliable information), and
> > 
> > * cause reproducibility problems (the type is either correct or not there).
> > ```
> 
> I don't understand how this would fix the reproducibility problem. On second evaluation, the parent is finalized and thus the slot type would _change_.
> 
> How can we prevent that?

What if I say that we don't have to? :)

> The heursitic you provided above does not deal with unfinalized parents (beside the last point of course).

The last point is the only one that's needed to actually deal with unfinalized parents! Finalization of classes, in my perspective, is part of image state (much like the state of WebKit things, or threads in the image) and thus is something that we should respect, but not expect. If the class is finalized—cool, let's look at it. If it's not—let's not do anything that it may break when finalized.

Ambrevar:
> let's not do anything that it may break when finalized.

Setting the type to whatever value would break when finalized.

Without getting into image considerations, double-loading the system / tests breaks in this case.  Which is very bad behaviour.

aartaka:
> > let's not do anything that it may break when finalized.
> 
> Setting the type to whatever value would break when finalized.

Well, then let's not set it at all, as I already suggested :)

aartaka:
I mean this, not giving up on type inference :P

> But, in general, I'd go with:
> - If parents are finalized, look at those and infer the type.
> - If parents are not finalized, don't infer the types of slots at all.

Ambrevar:
> > 
> 
> Well, then let's not set it at all, as I already suggested :)

I mean, not setting it would also break, because once the class is finalized, the next evaluation would set it, thus create  conflicting types.

Ambrevar:
If there is still a misunderstanding, could you maybe write a code sample to exhibit what you have in mind?

aartaka:
> I mean, not setting it would also break, because once the class is finalized, the next evaluation would set it, thus create conflicting types.

Yes, the slot type will be set on the second evaluation, but there will be no conflict and no error, because [redefining classes in CL (CLHS link)](http://www.lispworks.com/documentation/HyperSpec/Body/04_cf.htm) also includes redefining their types. Redefining
``` lisp
(defclass foo ()
 ((slot-a :initform 4)))
```
to 
``` lisp
(defclass foo ()
 ((slot-a :initform 4 :type (or null integer))))
```
is an absolutely valid thing and macroexpanding to it is fine too.

aadcg:
Feel free to make a decision together.

Ambrevar:
@aartaka The problem is not with the class redefinition, but with later assumptions from the compiler.

Take the watch-mode issue: the problem is with `setf` taking an argument of the wrong type.  With your strategy, the compiler would not signal a problem in case of type error, thus rendering type inheritance ineffective.

Or take the test form this pull request.  They currently won't pass your suggestion when evaluated twice, which seems wrong.

libraries/class-star/tests/tests.lisp
```diff
@@ -97,3 +97,27 @@
...
+    ((name "foo")
+     (age nil
+          :type (or null number))))
+  (class-star:define-class child (parent)
+    ((name nil
+           :type (or null string))
+     (age 17)))
+  (let ((p (make-instance 'parent))
+        (c (make-instance 'child)))
+    ;; Perform some  assignments: CCL can catch type errors here.
```
aartaka, line 112: SBCL can too, with `(declaim (optimize (safety 3) (debug 3)))`.

aartaka:
_Been rewriting this void knows how much times..._

So okay, all approaches are insufficient and imperfect, but, given that we try to avoid errors now, let's assess the error possibility vs. correctness for different approaches. Our situations for types are scaled along: 
- unfinalized vs. finalized superclass,
- and whether the type of the superclass slot is more/less exact, same, absolutely different from the subclass one, or absent.

Now, transform it into tables:

For ditching type inference:
```
| superclass slot rel. to subclass slot \ superclass | finalized | unfinalized |
|----------------------------------------------------+-----------+-------------|
| same                                               | nothing   | nothing     |
| more exact                                         | nothing   | nothing     |
| less exact (or no type)                            | nothing   | nothing     |
| not matching                                       | nothing   | nothing     |
| no type                                            | correct   | correct     |
```

For inference ignoring unfinalized superclasses (this PR):
```
| superclass slot rel. to subclass slot \ superclass | finalized | unfinalized                          |
|----------------------------------------------------+-----------+--------------------------------------|
| same                                               | correct   | nothing                              |
| more exact                                         | correct   | error (loose, more values allowed)   |
| less exact                                         | correct   | error (too narrow)                   |
| not matching                                       | correct   | error (wrong type)                   |
| no type                                            | correct   | error (presence of type)             |
```

For type inference only when all superclasses are finalized (my suggestion):
```
| superclass slot rel. to subclass slot \ superclass | finalized | unfinalized |
|----------------------------------------------------+-----------+-------------|
| same                                               | correct   | nothing     |
| more exact                                         | correct   | nothing     |
| less exact                                         | correct   | nothing     |
| not matching                                       | correct   | nothing     |
| no type                                            | correct   | correct     |
```

Now, if we take
- correct = 1
- nothing = 0
- error = -1

then
- no type inference = 2
- ignoring unfinalized superclasses = 1
- inferring only the finalized options = 6

Note that I may be forging results in my own favor here :)

libraries/class-star/tests/tests.lisp
Ambrevar: Nice, didn't know this!

Ambrevar:
I think we are looking at different directions.

> For type inference only when all superclasses are finalized (my suggestion):

My point is that class finalization is, in general, something that's out of the reach of the CL developer.  
Thus in the general case, with this PR (or your suggestion), we can expect errors.  Which is not acceptable in my opinion.

aartaka:
But still, we can minimize these errors to a single case: type narrowing on reload. Which, while being a violation of one's expectations, is not harmful. It's even useful.

We've ran nclasses in Nyxt for a long time, and it's been solid enough and extremely helpful for lots of cases. Add parent slot type inheritance, and the type inference will probably be as useful as it can ever be, fixing the only (there was also the `nil = boolean` problem, but it wasn't that big and kinda made sense) problem we've encountered with nclasses in all this time—wrong type inference of inherited slots.

I'm uneasy about the irreprodusibility and statefulness of type inference here too, but the same statefulness happens with reader macros when packages with those get reloaded (that's why we don't `*read-eval*` the sources of things in `describe-*` commands). Yes, strong discipline of not putting side effects into reader macros can nullify these errors, but so can strategically placed `closer-mop:ensure-finalized` too. We can even add an option to `define-class` that would automatically ensure all the parents are finalized. And we can mention this side effect of this particular algo in the README/comments, as we do with lots of other tradeoffs in our libraries.

But still, the message I'm trying to convey is: it works and it's useful, no need to ditch it all just because it's not perfect.

Ambrevar:
> Which, while being a violation of one's expectations, is not harmful. It's even useful.

Can you give an example of usefulness?
In my experience, it's the opposite: it's a distraction which conveys the wrong information.



> We can even add an option to `define-class` that would automatically ensure all the parents are finalized.

I think such an option is unrealistic.  Class finalization is something that's up to the compiler and `c2mop:ensure-finalized` should only be used with extreme care in very specific cases.



> Add parent slot type inheritance, and the type inference will probably be as useful as it can ever be, fixing the only (there was also the `nil = boolean` problem, but it wasn't that big and kinda made sense) problem we've encountered with nclasses in all this time—wrong type inference of inherited slots.

I like this argument but I conclude the opposite: the lack of inference inheritance has been a problem only _once_ since the dawn of time, so why bother?  It's such a trivial matter in the end :p

Ambrevar:
Maybe this example will clarify why I'm against this change:

```lisp
(declaim  (optimize (debug 3) (safety 3)) )
(progn
  (class-star:define-class parent3 ()
    ((name "foo")
     (age nil
          :type (or null number))))
  (class-star:define-class child3 (parent)
    ((name nil
           :type (or null string))
     (age 17)))
  (let ((p (make-instance 'parent3))
        (c (make-instance 'child3)))
    (setf (slot-value c 'age) nil)
))
```

The above raises an _error_ when first compiled.  So it's not just a style warning.  The generated code is deemed _wrong_ by the compiler.
In practice, this means that class-star users cannot rely on `(declaim  (optimize (debug 3) (safety 3)) )` or CCL, because every first compilation of the above will fail.

aartaka:
@Ambrevar, 5a8b4cee4a55409e352fef61209a6c0c3143d9b6 implements my suggestion. Try the example code against it—it doesn't error (but don't forget to provide `:name` argument to `make-instance 'child3` and refer to `parent3` instead of `parent`—those are errors all of their own)!

libraries/class-star/class-star.lisp
```diff
@@ -156,16 +156,26 @@ The predicate returns non-nil when the argument is of the `name' class.")
...
+           (every #'closer-mop:class-finalized-p
+                  (mapcar (lambda (s) (find-class s nil)) superclasses)))))
+    (cond
+      ((and all-parents-finalized-p
+            (find slot-name (alexandria:mappend
+                             (lambda (class) (ignore-errors (mopu:slot-names class)))
+                             superclasses)))
+       (let ((parent (find slot-name superclasses
+                           :test (lambda (slot class)
+                                   (find slot (ignore-errors (mopu:slot-names class)))))))
```
Ambrevar, line 173: No need for `ignore-errors` if parents are finalized.

libraries/class-star/class-star.lisp
```diff
@@ -182,13 +192,9 @@ The predicate returns non-nil when the argument is of the `name' class.")
...
           type-inference)
      (append definition
-             (multiple-value-bind (type parent)
-                 (inherited-slot-type (first definition) superclasses)
-               (if parent               ; Found parent slot, type possibly nil.
-                   (list :type type)
-                   (let ((result-type (funcall type-inference definition)))
-                     (when result-type
-                       (list :type result-type)))))))
+             (alexandria:when-let ((type (slot-type-maybe-inherited
```
Ambrevar, line 195: Don't use alexandria as `nclass` / `class-star` does not depend on it.

Ambrevar:
Thanks, @aartaka , we can go forward with this.

Can you update the tests so that it is reproducible?  I think the only fix required is that we finalize the `parent` class before defining the `child` class.

aartaka:
> Can you update the tests so that it is reproducible? I think the only fix required is that we finalize the `parent` class before defining the `child` class.

Yes, almost. Another fun thing that I encountered (and adjusted the tests for) was that SBCL (with safety and debug cranked up) does not allow redefining inherited slot to be of a wider type than the parent class slot. So `child`'s `name` cannot be `(or null string)` while `parent`'s one is a mere `string` :P

aartaka:
It's mergeable now, I believe!

libraries/class-star/tests/tests.lisp
```diff
@@ -98,26 +98,42 @@
   (assert-eq nil
              (getf (mopu:slot-properties 'foo-type-infer 'composite) :type)))
 
+(inherited-type-inference)
```
Ambrevar, line 101: Can you explain this?

libraries/class-star/tests/tests.lisp
```diff
@@ -98,26 +98,42 @@
...
 (define-test inherited-type-inference ()
   (class-star:define-class parent ()
-    ((name "foo")
+    ((name nil
+           :type (or null string))
```
Ambrevar, line 105: The point of testing 2 slots is that we test 2 logics:
- 1 slot is NIL in the parent and non-nil in the child.
- And vice versa.

With your change, both slots follow the same logic, so we are not testing the other way around.

libraries/class-star/tests/tests.lisp
aartaka: Ouch, test testing code :D

libraries/class-star/tests/tests.lisp
aartaka: Indeed, we don't check in both directions. But it's not exactly possible, see my comment above:
> SBCL (with safety and debug cranked up) does not allow redefining inherited slot to be of a wider type than the parent class slot. So child's name cannot be (or null string) while parent's one is a mere string :P

Ambrevar: It would still work if you set the type explicitly, but keep the initform as I wrote them, right?

aartaka:
Uh oh, doesn't pass ( ꒪Д꒪)ノ

aartaka:
A note for the future: we need to properly handle the case of `:unbound` initform, and not infer its type to be `keyword`, as we currently do (checked on a fresh checkout of Nyxt without this PR and with debug/safety optimizations).

Ambrevar:
Can you open an issue about this in nclasses?

Ambrevar:
Update the tests: @aartaka OK with the change?

We can squash and merge I believe.

Ambrevar:
@aartaka Tests still fail in weird ways.

On my machine, it failed as in the CI, but then I reran it and it passed.  Now I can't reproduce the issue locally.

Any idea?

aartaka:
Maybe try calling it `child2` to not reuse the old definition. In my case deleting the cache also seemed to solve the problem once.

Ambrevar:
Found the problem: `define-class` is a macro and thus type inference happens at read  time, before `(closer-mop:ensure-finalized (find-class 'parent))` is called.

Ambrevar:
OK with the result, @aartaka ?  We should probably squash some commits though.

aartaka:
All cool, yes!

aartaka:
@Ambrevar, which commits do you mean squashing? Can you do it yourself?

Ambrevar:
Done.

I suggest we don't merge this on master, since it would needlessly conflct with #2784.

Ambrevar:
I'll merge this directly onto nclasses.

Ambrevar:
Closing then?

aartaka:
Yes!

PR #2776, opened 2023-02-08
List internal pages in status buffer

### LINKED ISSUE #2771: Nyxt:New Buffer --please add a navigation object for internal pages to connect me with the status bar navigation system.
Gentle Engineers & Super Wizards,
Please help.

When I create a new buffer with an internal page {`New buffer` or `tutorial`}, it would really help me understand my position in the system if the status bar displayed a text object to reflect this new space. 

Please add a text object in the status buffer to represent the creation of these new buffers. 

**Why:**
- consistency (the status bar displays buffers, these are buffers)
- recognition is better than recall (this helps us identify where we are surfing without having to remember that there is some arbitrary internal page between two URLs)
- when users discover internal pages like `History Tree` this fix will help keep them oriented

Thank you : )

aartaka:
# Description

This adds a small "nyxt://" chip to status buffer listing of buffer domains. Behind it—internal pages. See the screenshot.

internal-buffer-chip

Fixes #2771.

# Discussion

@lansingthomas, does it look good to you?

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:


Ambrevar:
It's a bit inconsistent that special pages are listed via their scheme, while other pages are listed via they domain.

What about calling it `special` or `internal`?

No strong opinion though.

aartaka:
> It's a bit inconsistent that special pages are listed via their scheme, while other pages are listed via they domain.
> 
> What about calling it `special` or `internal`?

Yeah, it's inconsistent, but then they are special, so forcing consistency may hurt. `special` or `internal` might work too... But I'd delegate this decision to @lansingthomas :)

aadcg:
I second @Ambrevar's remark.

lansingthomas:
I am open to the chip, but not sold on it. I need to be certain that the chip fits with the `switch-buffer-next/previous` structure.

---

**3 Questions**
1. Would the chip accommodate multiple internal pages simultaneously? (like a `nyxt:manual` and a `nyxt:new`).
   - would I see two chips? Or would the internal pages stack in the chip?
  
2. What are the advantages of a chip like this?

3. How are URLs in the status bar currently ordered?
   - perhaps: recent use? order opened? scheme (I don't know what this is)?


I am hoping that internal nyxt pages, like `nyxt:manual` can be treated the same way as URLs in the status buffer. All buffers are not created equal, but at the task level, for navigation, they should be.

---

#### I think switching between buffers (groups of pages, or tasks) is one of the important things we provide as a browser, and a great opportunity for us.

In my way of thinking;
- we must provide a way to get users around in Nyxt using a **familiar** system.
   - The familiar system is traversing main-buffers (webpages + internal pages) horizontally, with some keys.
- IF this system clearly shows a users position in the status bar, it teaches and re-enforces important basic navigation commands (`switch-buffer-next` and `switch-buffer-previous`).
- IF we treat internal pages in the same ways (at this task level), we have a more seamless browser experience.

### Any structure for buffer navigation should:
- [ ] show position -- The primary purpose of any chip, or highlighted text object, would be to visually afford the user with their position in the system.
   - **Show me where I am, at any time,** while I am translating around with `switch-buffer-next` or `new-buffer` or `manual`, or after I click something from `list-buffers`.
   - A crucial feature is that the structure can NOT be something a user must remember, they must be able to see it.
- [ ] preserve secondary purposes -- The secondary purposes are; a clickable button, and showing the successful closing of a buffer (as the object vanishes from view).
- [ ] meet navigation expectations - translations left and right with `switch-buffer-next/previous` need to correspond with what is displayed in the status buffer, and with user expectations for movement within that system.
   - for example: when I see three URLS, and my position is in the middle of the three, my expectation is to use `switch-buffer-previous` to to the first, or `switch-buffer-next` to go to the third. 
including nyxt internal pages in status buffer navigation
   - right now (Version pre-release 3) the order is all tangled up.
- [ ] stay still - any change in the location of a buffer must be either instigated by that user, or approved by them. Remember, our users are trying to either start a new task, return to a task, or switch to another task. They do not want to be shuffling buffers around, and wondering about the order of their tabs.

#### Someday soon we will solve the other problems with buffer navigation. Today I want to get the internal pages presented in the status bar for navigation, so I can see what the next and previous buffer will be.

aadcg:
I think there are lots of misunderstandings in this discussion. I think @lansingthomas is confused about how things in Nyxt today with respect to the how we represent open buffers in the status bar. Until he understands that clearly, I don't find it efficient to discuss any future action. My advice to @lansingthomas is: discuss with @jmercouris (or with me, if you prefer) these topics.

I'm not interested in participating in the debate, but I can summarize my view briefly. In all browsers I've used, the only URL shown at the status bar is the current one.  And I think that's smart, because that's the only relevant URL. There are other interfaces to show all buffers, or their chronological ordering.   My favourite handling of the current URL is what safari does: it displays the scheme of the current URL in the status bar.  When you click on it, it shows the complete URL. In our case, "clicking on it" would be triggering the `set-url` command. Familiar, simple and easy to implement.

lansingthomas:
Dear computer magicians,

I am open to a vertical arrangement for listing open buffers, in fact I prefer it. 

We really want our users to be able to determine their position while navigating open buffers (and pages within them). 

### my understanding of the current state
To bring up a map for navigation:
1. Users need to invoke commands to bring up a buffer panel, or an internal page with; `buffers-panel`, `buffer-history-tree`, `list-buffers`
   - OR
2. Users can imagine a map and use `switch-buffer-next` and `switch-buffer-last`
   - OR
3. Users can use `switch-buffer` to bring up the prompt buffer with a list of suggestions.

- [ ] does this match your understand of the situation for a user navigating from page to page with Nyxt?

### What I need your help improving
Option number 2 for navigation is confusing because users get lost. We cannot expect them to maintain an accurate mental model of a map of open buffers. I know, as you know, that users can choose option 1, or option 3 to save themselves. I want to make option two better by building on the existing structure in the status buffer.

- [ ] are you with me?
- [ ] what am I missing?

Thank you

aartaka:
@lansingthomas, yes, I'm with you. Buffers are currently sorted by how long ago they were accessed, which is confusing. I'd rather sort them based on the underlying structure that we use for switching, yes.

aadcg:
@lansingthomas all good points!

> We really want our users to be able to determine their position while navigating open buffers (and pages within them).

@lansingthomas could you please explain what do you mean by position here? It seems to be me that you're tied to the idea that buffers ("tabs") have static ordinals associated with them. In normal browsers, there's the first, second, etc tabs and users can change their order by dragging them in the tab bar. So, indeed, in that case you can talk about position: "I can see that I'm at position tab 4 since it's the fourth tab in the tab bar."

In Nyxt, we think of it differently. At any given moment, there is the current buffer and all of the others. When you want to change buffer, you query "all of the others" and we display them sorted by last access time. In that sense, I don't think we can talk about "position". This is the logic behind `switch-to-buffer`, `switch-buffer-next` and `switch-buffer-previous`. It's the same idea as Ctrl + TAB in operating systems or other programs.

It's not clear to me what exactly you find confusing. I think you'd agree that "Ctrl + TAB" isn't confusing, yet there is something that we may be doing wrong. What is it? :)

Also, what do you mean by navigating pages within open buffers?



--

> I'd rather sort them based on the underlying structure that we use for switching

@aartaka could you explain?

aartaka:
Nyxt utilizes the tree-like history and buffer tree interrelations to shape the most intuitive buffer switching behavior in `switch-buffer-next`. The structure _is not_ sorted by access time, notably.

(I wonder if we should abstract it to a separate function, actually. So that it can be used in `switch-buffer` and other places across Nyxt.)

So, status buffer listing can be more intuitive and consistent than it is now. We just need to flatten the tree for it to work.

aadcg:
Thanks. As you can see, I don't know Nyxt myself!

lansingthomas:
Thank you both for your attentions in this matter.

## okay, I am starting to wrap my head around this. 

1. The navigation UI should be a reflection of our users mental models. (helping them resume tasks, start new tasks, switch between ongoing tasks)
2. Our users carry an imperfect mental model with them from a pattern of use with legacy browsers
   - that mental model looks like this:
buffer-history navigation mental model
3. We expect our users to intuit a new model
   - the new model is more like this:
buffer-history navigation mental model (1)

Cool, yes, I'm on board.

## help me help our poor status bar
IF we provide three different ways to get around, we need to make all three awesome and usable. I think we all agree that tabs and history as they are used in wide adoption now are bad and we can do better. Help me prioritize here my dear wizards. We cannot be hiding our best features.

AS `switch-buffer-next/previous` is commonly used in Nyxt --  let's make it better by showing the internal tabs in the status bar. It matches the mental model that users bring in from legacy browsers, and is also the method with least inputs for switching between buffers. We provide this status bar to coherently show system status.
OR
are we deliberately making the `switch-buffer-next/previous` navigation option confusing so that users try the other ones? right now we are worse than pseudo-tabs, we have mystery-pseudo-tabs in the Status Buffer where users need to read obscure documentation to understand it. Maybe I still don't understand it.


Why are we showing domains (or schemes) in the status bar at all? (seems like: visual affordance for one of our three navigation schemas. But not the best one!)

If it is too much -- okay. I can back off. Though I would really like to be able to see where `switch-buffer-next` and `switch-buffer-previous` will take me.

thank you

aadcg:
@lansingthomas I think this discussion has become so complex that I have serious doubts whether I'm able to follow it myself.

I think it would be more efficient to start with simple questions.

Let's ignore for a moment how "tabs" could be shown at the status buffer (its linear order and transpositions/permutations triggering by navigation). Do we want to show "tabs" in the status buffer? Can we reach a consensus on that topic?

> Though I would really like to be able to see where switch-buffer-next and switch-buffer-previous will take me.

Here I'm afraid that I still don't understand the problem you're addressing. Say that a user issues `switch-buffer-next` and it does something "unexpected". Then he can easily undo it with `switch-buffer-previous`. Perhaps you mean that you expect to see some kind of preview as depicted in the screenshot below. The issue with such a solution is that it doesn't scale and we expect our users to have lots of buffers open.

aadcg:
Remark. If we decide to display tabs it's important to keep `global-history-p` set to `nil` by default (as we do now). I believe this mirrors the behavior of mainstream browsers, where history is not shared among tabs.

I'm happy with whatever consensus that will be reached! No strong opinions from my side.

jmercouris:
> > Remark. If we decide to display tabs it's important to keep `global-history-p` set to `nil` by default (as we do now). I believe this mirrors the behavior of mainstream browsers, where history is not shared among tabs.
> 
> 
> 
> Wait, what?! Who put that in?! Why disable the most useful (and, supposedly, mentally reasonable) feature of Nyxt by default?! 

This kind of language is unhelpful and volatile. Calling things insane and mentally unreasonable carries the implication that holding said position makes you insane and unreasonable. I know that's not what you meant, I know you. I'm just trying to remind you that charging observations like that can give the wrong impressions and hurt others.

jmercouris:
I know times are difficult right now. We all have tough stuff going on. Let's just put a little more love and compassion in our communications :-). I love and appreciate all of you :-)

lansingthomas:
@aadcg 
the simple question approach is great. Thank you.

> Do we want to show "tabs" in the status buffer? Can we reach a consensus on that topic?
- I am a strong maybe. -- with my improved understanding from this discussion of how buffers are organized in Nyxt I am no longer convinced that we _must_ show "tabs" in the status buffer.
- I want to speak more about it, as it raises some other valuable questions  -- I will fit that into a conversation planned with John for later. I would love to hear your thoughts as well André -- we should catch up at a convenient time soon.

#### regarding the chip (@aartaka's merge)
- I think it is an improvement that does not take up too much real estate. AND we can implement it before resolving the larger "tabs" question. -- I am Thomas and I approve of the chip (for now)

aartaka:
> > > Remark. If we decide to display tabs it's important to keep `global-history-p` set to `nil` by default (as we do now). I believe this mirrors the behavior of mainstream browsers, where history is not shared among tabs.
> > 
> > 
> > Wait, what?! Who put that in?! Why disable the most useful (and, supposedly, mentally reasonable) feature of Nyxt by default?!
> 
> This kind of language is unhelpful and volatile. Calling things insane and mentally unreasonable carries the implication that holding said position makes you insane and unreasonable. I know that's not what you meant, I know you. I'm just trying to remind you that charging observations like that can give the wrong impressions and hurt others.

No, sorry, I didn't intend to translate this meaning (^_^;) What I meant is that global history is a _reasonable mental model_ and that disabling it makes it less useful. Not the best wording I've chosen, yes (⌒_⌒;)

aadcg:
> I think it is an improvement that does not take up too much real estate. AND we can implement it before resolving the larger "tabs" question. -- I am Thomas and I approve of the chip (for now)

I agree.

>    I would love to hear your thoughts as well André -- we should catch up at a convenient time soon.

Anytime!  I can tell you beforehand that I don't have a position on the topic. Instead, I have some opinions on what we should do when we take path A, B or C.

aartaka:
Okay, I've renamed the chip to `internal` and added a funky bit of behavior:

If there's only one internal page, list its name directly:
single-page-chip

And, if there are several, show a generic `internal` chip:
multiple-pages-chip

I'll rebase, squash, and merge.

PR #2781, opened 2023-02-09
Status buffer restyle

jmercouris:
# Description

This includes many minor fixes, for more information, please see the commit log.

Short summary: made modes area expand as needed instead of taking up a lot of space. Tweaked appearance of tabs. Increased contrast on mouseover. Use just domain in the status area, the full URL is 99% of the time cut off and therefore unreadable. Switch triangle shapes for forward/backward to arrows.

Before:
Screenshot_20230209_154901

After:
Screenshot_20230209_154755

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -769,7 +769,7 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :overflow "hidden"
               :white-space "nowrap"
               :z-index "3"
-              :flex-basis "8em")
+              :flex-basis "104px")
             `("#url"
               :background-color ,theme:primary
               :color ,theme:on-primary
@@ -782,7 +782,7 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :z-index "2"
               :flex-grow "3"
               :flex-shrink "2"
-              :flex-basis "10em")
+              :flex-basis "144px")
             `("#tabs"
               :background-color ,theme:secondary
               :color ,theme:on-secondary
@@ -795,7 +795,7 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :z-index "1"
               :flex-grow "10"
               :flex-shrink "4"
-              :flex-basis "10em")
+              :flex-basis "144px")
             `("#tabs::-webkit-scrollbar"
               :display "none")
             `(.tab
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -805,7 +805,7 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :padding-left "5px"
               :padding-right "5px")
             `(".tab:hover"
-              :opacity 0.8)
+              :opacity 0.6)
             `("#modes"
               :background-color ,theme:primary
               :color ,theme:on-primary
@@ -830,7 +830,7 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :background-color ,theme:accent
               :color ,theme:on-accent)
             `((:and .button :hover)
-              :opacity 0.8)
+              :opacity 0.6)
             `((:and .button (:or :visited :active))
               :color ,theme:background)
             `(.plain
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -764,13 +764,12 @@ store them somewhere and `ffi-buffer-delete' them once done."))
             `("#controls"
               :background-color ,theme:secondary
               :color ,theme:on-secondary
-              ;; :font-size "16px"
               :font-weight "700"
               :padding-left "5px"
               :overflow "hidden"
               :white-space "nowrap"
               :z-index "3"
-              :flex-basis "7em")
+              :flex-basis "8em")
             `("#url"
               :background-color ,theme:primary
               :color ,theme:on-primary
@@ -815,10 +814,7 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :padding-right "5px"
               :overflow-x "scroll"
               :white-space "nowrap"
-              :z-index "2"
-              :flex-grow "2"
-              :flex-shrink "1"
-              :flex-basis "10em")
+              :z-index "2")
             `("#modes::-webkit-scrollbar"
               :display "none")
             `(button
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -834,8 +834,8 @@ store them somewhere and `ffi-buffer-delete' them once done."))
             `((:and .button (:or :visited :active))
               :color ,theme:background)
             `(.plain
-              :padding-left "4px"
-              :padding-right "4px"
+              :padding-left "6px"
+              :padding-right "6px"
               :color ,theme:on-background
               :background-color ,theme:background))))
   (:export-class-name-p t)
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -802,8 +802,8 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :color ,theme:background
               :white-space "nowrap"
               :text-decoration "none"
-              :padding-left "5px"
-              :padding-right "5px")
+              :padding-left "8px"
+              :padding-right "8px")
             `(".tab:hover"
               :opacity 0.6)
             `("#modes"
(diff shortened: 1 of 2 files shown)


aadcg:
I agree with all the proposed changes!

aadcg (approved):

Ambrevar:
Fine with me.

@aartaka OK with the status bar URL->domain change?

aartaka:
I'm not OK with it at all, as it's a security risk! `(,,#ﾟДﾟ)`
- So, the [spec for URL display in browsers](https://url.spec.whatwg.org/#url-rendering-simplification) says that we can only hide scheme if there's only one scheme a browser ever uses, and that we can hide "trivial subdomains", like `www` or `m`. And that we can shorten it to a mere domain only when the space is severely constrained, which is certainly not our case here!
- Google Chrome started hinding the `http://` and `https://` schemes, and `www` subdomains, which [was rightfully labelled a security risk](https://bugs.chromium.org/p/chromium/issues/detail?id=883038). 

In the case of 083d22dfae304d29292795f5d18cd85428a0913c, we're not just stripping some "trivial" subdomain or scheme, we unconditionally hide most of the information there is to a URL, without providing any information necessary for a security decision. That's extremely bad. Either we use a much smarter display algo (at least the level of Google), or we don't use any. URLs are still important.

If anyone wants a compact display for the URLs in their Nyxt status—they can hack it up in their config, at their own risk.

jmercouris:
I've added the scheme back, I don't agree that it needs to be shown all the time, but better safe than sorry when it comes to security!

aartaka:
> I've added the scheme back, I don't agree that it needs to be shown all the time, but better safe than sorry when it comes to security!

No, this is not "better safe than sorry", it's still dangerously reckless. I've reverted the URL display changes that you've merged.

I would accept the changes to URL display, but only if they are at least as secure as the current state, which means:
- Protect against IDN homograph attacks for Unicode links (the whole reason behind showing the "safe" and "aesthetic" links alongside each other when Unicode is involved).
- Protect against subdomain spoofing, showing the full host.
  - OR showing at least the representative part of it and clearly stating it's incomplete (as recommended in the spec, `...domain.tld`)
- Show the scheme, or some equivalent of it, so that one is aware of which protocal/security measures are used.
  - OR only hide the scheme when it's intutive to do is (intuitiveness is to be discussed yet).
- Make sure the URL parts important for decision-making—path (mandatory) and query (recommended)—are accessible unconditionally, 
  - OR at least reachable via both keyboard and mouse.

--------------------------------------------------------------------------------------------------

Okay, I want our URL display to be pretty too. BUT, let's make it a reasonable redesign that does not endanger anyone. Areas where it's safe to change the display (according to the spec and common sense, in the order of decreasing safety):
- Stripping off only one scheme (HTTPS?) or replacing frequent schemes with emojis/labels/signs that uniquely identify those. Saves space, preserving the expectations and safety of web exploration.
- Not rendering URL username and password (WHATWG).
- Highlighting the host/domain of the URL, while painting the other parts gray/half-transparent.
- Showing the IDN Unicode-enabled URL as a host-only, to not repeat path/query/fragment twice—those are already there in IDN-encoded URL. Safe and aesthetic!
- Stripping `m.`, `www.`, `www[0-9]` from the host. Those WHATWG suggests, and they are relatively trivial.
- Stripping the trailing slash, index.html, index.php, and other home page suffixes. They are somewhat trivial (in my opinion) and are usually the same as a mere non-suffixed website host.

Finally, the aesthetic decision that lots of browsers (in particular and most of all Safari) made is putting the URL in the center of the status area—this way there's more than enough space for even the lengty URLs, and the focus of the user is rightfully on the current page URL. So we can:
- Increase URL area size, retaining its position;
- Move the URL area to the center, compacting buffers to favicons or abbreviations (not cool).
- Allow the flexbox layout of status area, so that CSS decides what should take the space, instead of us hard-coding pixels and making bad decisions on what to show, and what not to.

aartaka:
Much like with the discussion about smart attribute widths: don't enforce questionable defaults, especially if someone from the team is not okay with it. As I already suggested, make it optional or don't even make it: those that need it will hack it up in their own config files.

jmercouris:
The rationale for the URL only initially included the scheme- which I added. These additional considerations were not known to me, but I can integrate them as well.

The reason I made these changes to the URL display is because frequently the information is occluded. We basically just see a scheme, and the beginning of a URL. We don't see a path, or a title, and it is therefore not useful. The goal of this PR was to change the status bar to be more useful and space efficient.

Will update with a more comprehensive URL display!

aartaka:
> The rationale for the URL only initially included the scheme- which I added. 

No, it didn't only include the scheme. The linked resources were quire exact in the scope:
- Google Chrome issue tracker discussion was about schemes, 
  - _but also_ about subdomain spoofing.
- WHATWG document was quite detailed in regards of:
  - scheme display,
  - domains and subdomains display,
  - stripping off "trivial" subdomains,
  - decoding Unicode data in both
    - IDN-encoded hosts, and
    - in query parameters.

So yeah, I'll be glad if I won't have to link to those again for the next iteration of the display :)

> These additional considerations were not known to me, but I can integrate them as well.

Will be glad to see what comes out of it, yes! But this time wait for a review, please :P

> The reason I made these changes to the URL display is because frequently the information is occluded. We basically just see a scheme, and the beginning of a URL. We don't see a path, or a title, and it is therefore not useful. The goal of this PR was to change the status bar to be more useful and space efficient.

So, how does the fact of path being invisible lead to the idea that we need to stop showing it at all, only listing domain instead? Something's not aligning here...

So if you were concerned about URLs not being shown in full, then what about my suggestions above: make the flexbox do the layout itself? Or maybe move URL to the center of status area? Or maybe increase the size of the URL area in the status? There are many options there, and quite some of those are not endagering our users!

jmercouris:
I was only interested in the argument presented, not the linked articles. I assumed this was a specification that supported your notion and didn't delineate all of the requirements. Next time I will be more thorough!

Ambrevar:
@aartaka Considering how critical the `format-status-url` method is, it should be documented as such.

Mind adding a docstring explaining the potential attack vectors in there?

aartaka:
Done in 89df5aa3d92cb684310f9eb255792c1beeafadaa.

Ambrevar:
Thanks!

aadcg:
Unfortunately, there is plenty of disinformation in this thread.

An example of an IDN homograph atack: take the URL `https://google.com` and replace the `o` character by the Cyrillic character. In punycode, the URL becomes `https://xn--ggle-55da.com/`. 

The key observation is:

```lisp
(webkit:webkit-uri-for-display "https://xn--ggle-55da.com/") => "https://xn--ggle-55da.com/"
```

Therefore, `webkit:webkit-uri-for-display` behaves as the identity function in these cases and we can be assured that, **_in the context of the WebKitGTK port_**, we don't need to handle the URL display is any special way. This proves that the following commits needs to be reviewed: a376c05edc0e468190a5e0613f26d7544df377d2, 89df5aa3d92cb684310f9eb255792c1beeafadaa, 963e86b17661bbf471afa8d7fdcd127799ed6b53. 

It should be mentioned that, even if we did need to guard against IDN homograph attacks, the logic would still be deeply flawed. In the example below, the 2 returned values are the same because the check on non ASCII characters is not being made on the hostname. This would result on the status bar showing the URL as `URL-aesthetic (URL-safe)`.

```lisp
(render-url "https://ru.wikipedia.org/wiki/%D0%9F%D1%80%D0%B0%D0%B2%D0%B4%D0%B0_(%D0%B3%D0%B0%D0%B7%D0%B5%D1%82%D0%B0)")
=>
"https://ru.wikipedia.org/wiki/Правда_(газета)"
"https://ru.wikipedia.org/wiki/Правда_(газета)"
```

---------

No surprises here, and the [documentation for webkit:webkit-uri-for-display](https://webkitgtk.org/reference/webkitgtk/stable/func.uri_for_display.html) is crystal clear:


> The URIs used internally by WebKit may contain percent-encoded characters or Punycode, which are not generally suitable to display to users. **This function provides protection against IDN homograph attacks, so in some cases the host part of the returned URI may be in Punycode if the safety check fails.**


-------


All's well that ends well!

PR #2801, opened 2023-02-17
Fix prompt buffer actions

### LINKED ISSUE #2794: Should `prompter:actions-on-{return,marks,current-suggestion}` be set to functions?
Look at the following prompt buffer invocation. The action defined is valid, and runs as expected. An error is raised  when issuing command `set-action-on-return`. The reason being that `make-action-suggestion` expects symbols or commands (that's why actions are usually defined via `lambda-command` or `lambda-mapped-command`).

```lisp
(prompt :prompt "Test"
        :sources (make-instance 'prompter:source
                                :name "Test"
                                :constructor (list "Hello")
                                :actions-on-return
                                (list
                                 ;; Defining it with a lambda results in
                                 ;; a error when issuing command
                                 ;; set-action-on-return.
                                 (lambda (suggestion-list)
                                   "Greet world."
                                   (echo (str:concat (first suggestion-list)
                                                     " World!")))
                                 #'identity)))
```

I suggest that the elements of lists `prompter:actions-on-{return,marks,current-suggestion}` should either be symbols or commands.

aadcg:
# Description

Fixes #2794

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [ ] Documentation:
- [x] Compilation and tests:


source/prompt-buffer-source.lisp
```diff
@@ -0,0 +1,57 @@
...
+(define-class prompt-buffer-source (prompter:source)
+  ;; Override prompter:actions-* slots.  In Nyxt, we want actions to be function
+  ;; symbols instead of functions, since the later are opaque.
+  ;; Concretely, it's important when issuing command set-action-on-return.  See
+  ;; `nyxt/prompt-buffer-mode::make-action-suggestion'.
+  ((prompter:actions-on-return
+    'identity
+    :type (or null
+              function-symbol
+              (cons function-symbol *))
```
aadcg, line 15: The `:type` declaration doesn't seem to be honored since the following is valid:

`(make-instance 'prompt-buffer-source :name "test" :actions-on-return #'identity)` 

Is this a bug or am I missing something? CC @Ambrevar @aartaka

aartaka: It's not a bug, it's compiler settings. SBCL does not check slot types unles it's `(optimize (safety 3) (debug 3))` (But then it goes standard-compliant and ignores function-typed slots that ANSI CL prohibits—at least it used to do this in some 2.2.6. Things might've changed, but better to keep things as symbols anyway).

aadcg: Clear, thanks!

aadcg:
Ready for review.

CC @Ambrevar @aartaka.

source/mode/search-buffer.lisp
```diff
@@ -254,7 +254,8 @@ Example:
           :sources (make-instance 'search-buffer-source
                                   :case-sensitive-p case-sensitive-p
                                   :actions-on-return
-                                  (lambda (search-match)
+                                  (lambda-command identity (search-match)
```
aartaka, line 257: Maybe better to define it as a toplevel command?

source/prompt-buffer-source.lisp
```diff
@@ -0,0 +1,85 @@
+;;;; SPDX-FileCopyrightText: Atlas Engineer LLC
+;;;; SPDX-License-Identifier: BSD-3-Clause
+
+(in-package :nyxt)
+
+(define-class prompt-buffer-source (prompter:source)
+  ;; Override prompter:actions-* slots.  In Nyxt, we want actions to be function
```
aartaka, line 7: But then it should also be commands, right?

aartaka (commented):
The name seems unnecessarily verbose. Why not just `source` or `prompt-source`?

source/mode/search-buffer.lisp
aadcg: What would be the benefit? The commands hardly does anything, so I don't see the point in having it globally available.

source/prompt-buffer-source.lisp
aadcg: Yes, I forgot it!

aadcg:
> The name seems unnecessarily verbose. Why not just `source` or `prompt-source`?

Let's stick to `prompt-source` then.

aadcg:
Ready to merge from my side.

CC @Ambrevar @aartaka

source/mode/search-buffer.lisp
aartaka: I mean, there's a `'identity` and `'buffer-load*` all across the code, and it's quite uncertain what those are bound to. It'd be better if there was an introspectable toplevel form for it, than some local definition. Even though it doesn't do much.

aadcg: I don't follow... Are we discussing the `lambda-command` defined in `search-buffer`?

> I mean, there's a 'identity and 'buffer-load* all across the code, and it's quite uncertain what those are bound to.

What do you mean by "bound" here? Bound to a keybinding, or to what these symbols are bound (its definition as a function)?

> It'd be better if there was an introspectable toplevel form for it, than some local definition.

Is this a general remark? I think `lambda-command`, `lambda-mapped-command` and `lambda-unmapped-command` are all useful in the context of actions since they don't pollute the command list. It if often the case that the routine you want to call makes sense only when called in that context.

aartaka:
I've just pushed three commits that I believe to do The Right Thing ™️. 1d76f9e53a41bfe72e8f98439715477213878e96 makes sure it's actually commands or command symbols that we provide as actions, making things strict and not allowing any function in, even as a symbol. aba95c4ed7237be6f001e1eee70c822dab7a2e06 gets rid of `buffer-load*` being a function, by defining it as a variable instead (this is a terribly arguable change, so alternative solutions are welcome). And, finally, 9fc68576951dd06e6bf8e9cb85779a215931e299 gets rid of `identity` being used everywhere and being a function, by replacing it with NIL, which does not conflict with either `prompter` or `prompt-source` checks.

@aadcg, does this change make sense to you? Is this fitting our idea of making sure only commands are allowed for actions?

source/prompt-source.lisp
```diff
@@ -0,0 +1,81 @@
...
+  (let ((%value (uiop:ensure-list value)))
+    (if (valid-actions-p %value)
+        (setf (slot-value source 'prompter:actions-on-return) %value)
+        (error "Please provide actions as command symbols or commands."))))
+
+(defmethod (setf prompter:actions-on-current-suggestion) :after (value (source prompt-source))
+  (let ((%value (uiop:ensure-list value)))
+    (if (valid-actions-p %value)
+      (setf (slot-value source 'prompter:actions-on-current-suggestion) %value)
+      (error "Please provide actions as command symbols or commands."))))
```
aartaka, line 69: Here I'd rather make it a separate condition class and invoke it with `cerror` so that the error is continuable. This motivates for a slight rewrite, like:
``` lisp
(defmethod (setf prompter:actions-on-return) :after (value (source prompt-source))
  (let ((%value (uiop:ensure-list value)))
    (unless (valid-actions-p %value)
      (cerror "Use this value anyway" 'non-commands-as-prompt-actions ...))
    (setf (slot-value source 'prompter:actions-on-return) %value)))
```

aadcg:
> @aadcg, does this change make sense to you? Is this fitting our idea of making sure only commands are allowed for actions?

Regarding not allowing function symbols, I think that's OK.

Regarding `buffer-load*`, also OK.

What's not OK is setting actions to `NIL`. I think it's unacceptable from the user perspective. Yes, `identity` and `NIL` are the same (from the programming interface perspective), but not for the user! When a user runs `set-action-on-return`, it is unacceptable to find an empty list. There should always be an action - the identity one. We can set it a command, not as a function symbol.

```lisp
(prompt :prompt "Prompter"
        :sources (make-instance 'prompter:source
                                :name "Source"
                                :constructor (list "Hello")
                                :actions-on-return nil))
```

 [I've repeated this argument several times now](https://github.com/atlas-engineer/nyxt/issues/2794#issuecomment-1433146897), so please explain why you disagree. CC @ambrevar.

source/prompt-source.lisp
aadcg: Yes, that's a good idea. I didn't do it from the start since I don't much about conditions yet. I'll look into it.

aartaka:
Oh right, it's my attention span being a potato, sorry. How about setting it to a newly-defined command then? Like 
``` lisp
(export-always 'buffer-load*)
(defvar buffer-load*
  (lambda-command return-unaltered* (list)
    "Return all the suggestions from LIST unaltered."
    list))
```

aadcg:
Well, I'd just wrap `identity` into a `lambda-command`, but that's indeed the spirit!

aartaka:
365878b5559a9721dfee96e26d7fbee97e582711 should do it better.

aadcg:
Can I take over the PR now? Would you mind if I squash some of your commits into older ones? @aartaka

aartaka:
Go ahead, yes.

source/prompt-source.lisp
aadcg: I agree with defining a condition, but I'm not sure about the error being continuable.

aadcg:
@aartaka Could you explain the change from `initialize-instance :after` to `customize-instance :after`.

After reading the documentation of the later (below), I got the idea that `customize-instance :after` should be left for the user. Am I misinterpreting it?

```lisp
"Specialize this method to customize the default values and
behavior of some CLASS instance.

This method is run after the instance has been initialized (in particular, after
the `initialize-instance' :after method).

The standard method is reserved for user configuration.

Do not specialize the standard method in public code, prefer
`initialize-instance :after' instead to initialize slots, and
`customize-instance :after' for code that relies on finalized slot values."
```

aadcg:
Also, look at the following:

```lisp
NYXT> (define-class test-class ()
  ((slot 5)))

(defmethod initialize-instance :after ((test test-class) &key)
  (print "hello initialize"))

(defmethod customize-instance :after ((test test-class) &key)
  (print "hello customize"))

(make-instance 'test-class)

"hello initialize" 
#<TEST-CLASS {100502E093}>
```

Isn't this a bug? `customize-instance :after` seems to be ignored.

aartaka:
Yeah, `customize-instance` is a `user-class`-specific configuration method that e.g. `define-configuration` uses :)

source/prompt-source.lisp
aartaka: I mean, it's a convenience thing for production use. If someone configures an action that's a function—then why not, `prompter` will work with it. So binary build on Nyxt will easily use `continue` restart. 

What we're guarding against here is inconsistencies in the core that will immediately surface when running non-binary Nyxt. Even though the error is continuable, it's still an error, which means that something's gone wrong.

-------------------------------------------------------------------------

But then all the stuff I do personally in CL, I add `cerror` to, because this forces quite a flexible design on the developer and allows the user to not worry about recoverable/benign errors (which this action type mismatch error more or less is). So yes, I'm quite biased towards this design.

aadcg: I have a question that is related: how are conditions and restarts handles when you run a binary generated by `sbcl` or other implementations. Is it part of part of the CL specification or dependent on the implementation?

aadcg:
> @aartaka Could you explain the change from `initialize-instance :after` to `customize-instance :after`?

@aartaka could you answer the question above?

Also, what motivates the fact that you changed `#'nyxt:command-p` to `(read-from-string "nyxt:command-p")`?

aartaka:
> > @aartaka Could you explain the change from `initialize-instance :after` to `customize-instance :after`?
> 
> @aartaka could you answer the question above?

Yes, sure. So `customize-instance :after` is our point for finalization after all the user configuration is done (most of it is inside `customize-instance` (wthout a qualifier). In it, we can safely do things to an object knowing that its class won't change anymore. Thus, the change to `customize-instance :after` is a good tone towards user configuration of `prompt-source` and its subclasses (that all should be `user-class`-metaclasses), taking it into account.

> Also, what motivates the fact that you changed `#'nyxt:command-p` to `(read-from-string "nyxt:command-p")`?

This one is to not rush ahead of the file load order. Even though I don't knwo of any implementation that does that, there could be one that doesn't allow using function before it's defined. Thus this reader trick to lazy-resolve a symbol.

Bute we should probably use `uiop:find-symbol*` here...

aartaka:
@aadcg, I just pushed two commits trying to make `buffer-load*` and `identity*` into commands. This works, but the `execute-command` prompt lists those two alongside regular commands, which is not cool. I'm thinking of moving those into `prompt-buffer-mode`, but that is likely to create an avalanche of `SYMBOL not found in PACKAGE` errors, which I'm not exactly ready for.

-------------------------------------------------
@Ambrevar, we seem to have found yet anothe class of commands: not-user-reachable, type-only commands. They are commands, but only in type. They don't behave like commands. But they have to be ones. So our classification of commands becomes:
- Commands accessible anywhere and anytime: `:global`.
- Commands accessible based on the enabled modes: `:mode`.
- Inacessible frow anywhere: `:anonymous`.
- Accessible in the code, but not to the user: ???

I wonder if I'm dragging it too far here and we'd be better of with simply allowing function actions here. Or using some `lambda-(un)mapped-command` on some function...

aadcg:
I'll re-read #2801 and #2794 and write some conclusions.

source/prompt-source.lisp
```diff
@@ -0,0 +1,94 @@
...
+(defmethod customize-instance :after ((source prompt-source) &key)
+  (setf (actions-on-return source)
+        (slot-value source 'prompter:actions-on-return))
+  (setf (actions-on-current-suggestion source)
+        (slot-value source 'prompter:actions-on-current-suggestion))
+  (setf (actions-on-marks source)
+        (slot-value source 'prompter:actions-on-marks)))
+
+(defmethod (setf actions-on-return) :after (value (source prompt-source))
+  (let ((%value (%actions value)))
```
Ambrevar, line 59: Why the percent syntax?

source/prompt-source.lisp
```diff
@@ -0,0 +1,94 @@
...
+    (unless (valid-actions-p %value)
+      (cerror "Use this value anyway." 'invalid-prompt-buffer-actions :arg %value))
+    (setf (slot-value source 'prompter:actions-on-marks) %value)))
+
+(defun valid-actions-p (actions)
+  "Ensure that ACTIONS are funcallables."
+  (every (alex:disjoin 'functionp 'sym:function-symbol-p) (uiop:ensure-list actions)))
+
+(defun %actions (actions)
```
Ambrevar, line 80: Same, no need for the `%`-prefix I believe.

Ambrevar:
I'm a bit  confused about the problem we are trying to fix here.
Is it that when `set-action-on-return` is called on an empty list, we should _not_ show a prompt buffer with 0 suggestions?  If so, why not just edit `set-action-on-return`?

Sorry if is off track, I'm a bit lost with the discussion here :p

Ambrevar:
> This PR is mostly about making sure prompt actions are always commands and never functions.

What's the use case?  If it's about the user accidentally setting a function instead of a command, we could handle this in `action-on-return-source` and display a warning if a function is found instead of a command.

aadcg:
> I'm a bit confused about the problem we are trying to fix here.

[So I am, that's why I went back to the issue and tried to sum up everything that we've been discussing](https://github.com/atlas-engineer/nyxt/issues/2794#issuecomment-1441864946).

Let's please continue to discuss what we actually want in the issue, and then we can work on the PR.

aadcg:
Closing this one. Thanks for the input.

PR #2802, opened 2023-02-19
Flatpak

jmercouris:
I think it works :-O

can someone please try the resultant Flatpak? I tried it myself, but I want to make sure it isn't because I had installed something to make it work!


aadcg:
@jmercouris Thanks for working on this!

For those reviewing this PR that are on Guix. `flatpak` is available, while `flatpak-builder` isn't. I recommend installing the latter via `nix` (add `(service nix-service-type)` to your Guix services). Theoretically, `flatpak` could be installed via `nix` too, but if you're using `nix` as an addition package manager (not NixOS) [you're out of luck](https://nixos.wiki/wiki/Flatpak). 

If the flatpak commands won't work, append `--user` so that they're executed user-wide (instead of system-wide).

`flatpak remote-add --user --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo`
`flatpak install --user --noninteractive org.gnome.Sdk/x86_64/41`

aadcg:
@jmercouris I was able to run the Nyxt flatpak!

makefile
```diff
@@ -89,3 +76,21 @@ clean-submodules:
...
 clean: clean-submodules
+	rm -rf build
+
+.PHONY: flatpak-build
+flatpak-build:
+	@$(FLATPAK_BUILDER) --user --install --force-clean build $(FLATPAK_MANIFEST)
+
+.PHONY: flatpak-run
+flatpak-run:
+	@$(FLATPAK_COMMAND) run $(FLATPAK_APP_ID)
```
aadcg, line 87: Personally, I think `flatpak-run` doesn't belong to a makefile!

makefile
```diff
@@ -89,3 +76,21 @@ clean-submodules:
...
+
+.PHONY: flatpak-repository
+flatpak-repository:
+	mkdir -p $(FLATPAK_EXPORT_REPOSITORY)
+	@$(FLATPAK_BUILDER) --force-clean --repo=$(FLATPAK_EXPORT_REPOSITORY) build $(FLATPAK_MANIFEST)
+
+.PHONY: flatpak-bundle
+flatpak-bundle:
+	@$(FLATPAK_COMMAND) build-bundle $(FLATPAK_EXPORT_REPOSITORY) nyxt.flatpak $(FLATPAK_APP_ID)
```
aadcg, line 96: Do we need all of these: `flatpak-repository`, `flatpak-bundle` and `flatpak-build`? What's the difference between these?

.github/workflows/release.yaml
```diff
@@ -94,4 +94,4 @@ jobs:
       uses: ncipollo/release-action@v1
       with:
         bodyFile: release.txt
-        artifacts: "nyxt-*.tar.xz*,guix-recipe.txt,*.deb"
+        artifacts: "nyxt-*.tar.xz*,guix-recipe.txt,*.flatpak"
```
aadcg, line 97: What exactly is this generated flatpak file?

aadcg:
I'm confused about how we'll distribute the Nyxt flatpak.

There are 3 ways to install flatpaks. 

1. Via `flathub` ([we need to add the Nyxt flatpak to their repositories](https://github.com/flathub/flathub))
2. Via a so-called `.flatpakref` file that enables users to say `flatpak install /path/to/.flatpakref`.
3. Via a URL that points to the file mentioned above `flatpak install url-link-to-flatpakref`.

engineer.atlas.Nyxt.yaml
```diff
@@ -0,0 +1,61 @@
+app-id: engineer.atlas.Nyxt
+# Not a mistake! We need the sonames of shared libraries for SBCL during runtime.
+runtime: org.gnome.Sdk
+runtime-version: "41"
+sdk: org.gnome.Sdk
```
aadcg, line 5:
```log
Info: org.gnome.Sdk is end-of-life, with reason: The GNOME 41 runtime is no longer supported as of September 17, 2022. Please ask your application developer to migrate to a supported platform.
Info: org.gnome.Sdk.Locale is end-of-life, with reason: The GNOME 41 runtime is no longer supported as of September 17, 2022. Please ask your application developer to migrate to a supported platform.
Installing runtime/org.gnome.Sdk/x86_64/41
```

Do we really need this version? If so, can we add a comment explaining it?

jmercouris:
> @jmercouris I was able to run the Nyxt flatpak!

Yay!

engineer.atlas.Nyxt.yaml
jmercouris: You're right. I have updated it to 42! 43 is causing some gobject errors, so we will leave that for now.

.github/workflows/release.yaml
jmercouris: > I'm confused about how we'll distribute the Nyxt flatpak.
> 
> There are 3 ways to install flatpaks.
> 
>     1. Via `flathub` ([we need to add the Nyxt flatpak to their repositories](https://github.com/flathub/flathub))
> 
>     2. Via a so-called `.flatpakref` file that enables users to say `flatpak install /path/to/.flatpakref`.
> 
>     3. Via a URL that points to the file mentioned above `flatpak install url-link-to-flatpakref`.

We will support installation via two methods:

1. Flathub
2. A file called `nyxt.flatpak` which contains everything the user needs

what they do with this file is the following: `flatpak install nyxt.flatpak` and then it is available on their system :-)

makefile
jmercouris: Yes we do need them all. In order to make a `flatpak-bundle` (it generates the standalone nyxt.flatpak file), we need a `flatpak-repository.` In order to test and run our flatpak locally we need `flatpak-build` and `flatpak-run`

makefile
jmercouris: It makes a little bit easier than writing the tedious: `flatpak run engineer.atlas.Nyxt`, I've seen many other flatpak repositories that use this.

makefile
aadcg: Clear.

.github/workflows/release.yaml
aadcg: Clear - the CI will make `nyxt.flatpak` available on each release.

engineer.atlas.Nyxt.yaml
aadcg: Thanks.

aadcg:
@jmercouris same comment: please merge according to our standards and squash some commits. 

For instance, having a commit where you specify version 41 and then 42 is useless. No one wants to see your thinking process while working on something. All we care about is: before there was no flatpak generation logic, and now there is. We want to see a clean addition of this feature. On the other hand, in the future, that generation logic will be tweaked. In that case, yes, I want to see every little detail because it has already landed on master!

jmercouris:
@Ambrevar the files are still in the Git history. I removed them because I also delete code mostly instead of just commenting out. It is not gone, and since we are not using it, I think the clutter does not serve us well.

The work is not destroyed! It is still there! Do you disagree with my approach? Can you think of a better way I should have done it?

aartaka:
> @Ambrevar the files are still in the Git history. I removed them because I also delete code mostly instead of just commenting out. It is not gone, and since we are not using it, I think the clutter does not serve us well.

GIt history tends to be unreliable, as it can be changed any moment by a recless rebase. We have a convention to not do this, but we are humans, which means we happen to err. 

I personally am in favor of keeping the .deb, even though it's not working for all Ubuntu versions. @jmercouris, you keep advocating for installability of Nyxt, why remove yet another way to install it then?

Ambrevar:
@jmercouris There is no clutter if we keep the deb generation in a seperate file (I agree that keeping it commented would be ugly).

jmercouris:
> @jmercouris There is no clutter if we keep the deb generation in a seperate file (I agree that keeping it commented would be ugly).

I still think it would be clutter, but that is not my rationale for removing it.

As we had mentioned in the other issue, it will not work on newer versions of Ubuntu, and it will confuse users. If we have a single way of installing (other than Guix), I think we can more easily support it. Thoughts?

jmercouris:
> GIt history tends to be unreliable, as it can be changed any moment by a recless rebase. We have a convention to not do this, but we are humans, which means we happen to err.

I have thought about this argument, but a reckless rebase can destroy it even IF it is a file. Therefore, I don't believe there is a real difference between the two. If we happen to do such a rebase, well, it is OK, there are multiple copies of the Git history, one of us is bound to have it :-D 

> I personally am in favor of keeping the .deb, even though it's not working for all Ubuntu versions. @jmercouris, you keep advocating for installability of Nyxt, why remove yet another way to install it then?

Because we want installations that work, and the deb installation does not work, for a large quantity of the users. Many users on systems that do not support the deb will see it, and attempt to install it, then it won't work, and they will get frustrated. The flatpak always works, every time. Therefore, I think it is a better solution for our casual user. Plus, it is built into many Linux distributions! https://www.flatpak.org/setup/ I spent a lot of time evaluating the different casual user packaging options before narrowing down on Flatpak, which I believe to be the most promising.

aartaka:
> > GIt history tends to be unreliable, as it can be changed any moment by a recless rebase. We have a convention to not do this, but we are humans, which means we happen to err.
> 
> I have thought about this argument, but a reckless rebase can destroy it even IF it is a file. Therefore, I don't believe there is a real difference between the two. 

There is. Getting really metaphysical: it's much easier to destroy a things that _was_, rather than the thing that _is_. It's harder to destroy both a file and a history of it, than just erasing history :P

> > I personally am in favor of keeping the .deb, even though it's not working for all Ubuntu versions. @jmercouris, you keep advocating for installability of Nyxt, why remove yet another way to install it then?
> 
> Because we want installations that work, and the deb installation does not work, for a large quantity of the users. Many users on systems that do not support the deb will see it, and attempt to install it, then it won't work, and they will get frustrated. The flatpak always works, every time. Therefore, I think it is a better solution for our casual user. Plus, it is built into many Linux distributions! https://www.flatpak.org/setup/ I spent a lot of time evaluating the different casual user packaging options before narrowing down on Flatpak, which I believe to be the most promising.

This is a valid argument, but [not for Ubuntu 23.04+ (The Register newspiece)](https://www.theregister.com/2023/02/23/ubuntu_remixes_drop_flatpak/)

Ambrevar:
> As we had mentioned in the other issue, it will not work on newer versions of Ubuntu, and it will confuse users. If we have a single way of installing (other than Guix), I think we can more easily support it. Thoughts?

You just said it, we don't have a single way :p  Also we can always follow Ubuntu LTS, which some times will work for all latest versions of Ubuntu.  It's a matter of "Ubuntu period".

My point is not about that though, it's just about keeping this precious piece of code in the worktree.

jmercouris:
Don't worry, it is not lost, it is still in the Git history!

I fundamentally disagree about keeping a broken installation offering in our downloads.

Ambrevar:
It's not broken: if we want to release for the latest Ubuntu, it's a trivial change in the OS selection.

Hard work has been put into this, and the result is fantastic: a tailored .deb package that just works.  Please, let's be a bit more considerate here :p 

Being in the Git history is not a way to store code in a way that anyone  can realistically   find it back.

PR #2805, opened 2023-02-20
spinneret-tags(:p*): Add a smart paragraph tag.

aartaka:
# Description

Hey @aadcg, look, a smart paragraph thingie to not get irritated with quoted strings!

Still, it's quite incomplete and buggy. There's only so much we can do without hooking into the reader...

### DIFF
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -19,50 +19,51 @@ of Nyxt."))
           (config-file (namestring (files:expand *config-file*)))
           (rules-file (namestring (files:expand (make-instance 'auto-rules-file)))))
       (:nsection :title "Configuration"
-        (:p "Nyxt is written in the Common Lisp programming language which offers a
-great perk: everything in the browser can be customized by the user, even while
-it's running!")
-        (:p "To get started with Common Lisp, we recommend checking out
-    our web page: "
-            (:a :href "https://nyxt.atlas.engineer/learn-lisp" "Learn Lisp")
-            ". It contains numerous pointers to other resources, including
-        free books both for beginners and seasoned programmers.")
-        (:p "Nyxt provides a mechanism for new users unfamiliar with Lisp to
-customize Nyxt. Start by invoking the commands " (:nxref :command 'describe-class) "
-or " (:nxref :command 'describe-slot) ".  You can press the button marked 'Configure' to
-change the value of a setting. The settings will be applied immediately and
-saved for future sessions. Please note that these settings will not alter
-existing object instances.")
+        (:p* *Nyxt is written in the *Common *Lisp programming language which offers a
+          great perk\: everything in the browser can be customized by the user\, even while
+          it\'s running!)
+        (:p* *To get started with *Common *Lisp\, we recommend checking out
+          our web page\:
+          (:a :href "https://nyxt.atlas.engineer/learn-lisp" "Learn Lisp") \.
+          *It contains numerous pointers to other resources\, including
+          free books both for beginners and seasoned programmers.)
+        (:p* *Nyxt provides a mechanism for new users unfamiliar with Lisp to
+            customize *Nyxt. *Start by invoking the commands (:nxref :command 'describe-class)
+            or (:nxref :command 'describe-slot) \. *You can press the button marked
+            (:code "Configure") to change the value of a setting. *The settings will be
+            applied immediately and saved for future sessions. *Please note that these
+            settings will not alter existing object instances.)
         (unless (str:empty? auto-config-file)
-          (:p "Settings created by Nyxt are stored in " (:code auto-config-file) "."))
+          (:p* *Settings created by *Nyxt are stored in (:code auto-config-file) \.))
         (unless (str:empty? config-file)
-          (:p "Any settings can be overridden manually by " (:code config-file) "."))
-        (:p "The following section assumes knowledge of basic Common Lisp or a
-similar programming language.")
-        (:p "The user needs to manually create the Nyxt configuration file, and the parent folders if necessary."
-            (when (and (current-buffer) ; In case manual is dumped.
-                       (not (files:nil-pathname-p config-file))
-                       (not (uiop:file-exists-p config-file)))
-              (:p "You can also press the button below to create it."
-                  (:p (:a :class "button"
-                          :onclick (ps:ps
-                                     (nyxt/ps:lisp-eval
-                                      (:title "create-config-file")
-                                      (ensure-directories-exist config-file)
-                                      (ensure-file-exists config-file)
-                                      (echo "Configuration file created at ~s." config-file)))
-                          "Create configuration file")))))
-        (:p "Example:")
+          (:p* *Any settings can be overridden manually by (:code config-file) \.))
+        (:p* *The user needs to manually create the *Nyxt configuration file\,
+          and the parent folders if necessary.
+          (when (and (current-buffer) ; In case manual is dumped.
+                     (not (files:nil-pathname-p config-file))
+                     (not (uiop:file-exists-p config-file)))
+            (:p* *You can also press the button below to create it\:)
+            (:a :class "button"
+                :onclick (ps:ps
+                           (nyxt/ps:lisp-eval
+                            (:title "create-config-file")
+                            (ensure-directories-exist config-file)
+                            (ensure-file-exists config-file)
+                            (echo "Configuration file created at ~s." config-file)))
+                "Create configuration file")))
+        (:p* *The following section assumes knowledge of basic *Common *Lisp or a
+          similar programming language.)
+        (:p* *Example\:)
         (:ncode
           (define-configuration web-buffer
             ((default-modes (pushnew 'nyxt/no-script-mode:no-script-mode %slot-value%)))))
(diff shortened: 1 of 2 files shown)


aadcg:
@aartaka thanks for working on this!

One thing that we'd need to test is whether moving from `:p` to `:p*` would not take much more time compiling the manual file.

> Still, it's quite incomplete and buggy. 

Meaning that it's not surprise that the following error, right? `(spinneret:with-html-string (:p* *Hello |there,| are you (thee? (:span \"something?\") thou?) doing?))`

Still, the following already works, which is great! 

```(spinneret:with-html-string
  (:p* A sentence I can type freely.  Another one.  And yet another.))
```

aartaka:
> @aartaka thanks for working on this!
> 
> One thing that we'd need to test is whether moving from `:p` to `:p*` would not take much more time compiling the manual file.

This is not really dependent on the complexity of the macros (unless those stack-overflow xD), but rather on the length of forms, at least on SBCL. If we split the manual into separate toplevel forms (as with changelog or migration guide), it will compile lightning fast, no matter the macro complexity.

aadcg:
Thanks for the clarification. The most logical way would be to break up the manual into its sections then (if needed).

aartaka:
> Thanks for the clarification. The most logical way would be to break up the manual into its top sections then (if needed).

A long overdue change, yes.

aartaka:
> Meaning that it's not surprise that the following error, right? `(spinneret:with-html-string (:p* *Hello |there,| are you (thee? (:span \"something?\") thou?) doing?))`

If you remove the backslash before the quotes, it does |ω・）

aadcg:
> If you remove the backslash before the quotes, it does work |ω・）

Yes, that was silly of me!

> Still, it's quite incomplete and buggy.

I got confused with what you mentioned!

aartaka:
Also note that Spinneret has [Markdown support](https://github.com/ruricolist/spinneret#markdown) as a reasonable alternative for this change.

aadcg:
> Also note that Spinneret has [Markdown support](https://github.com/ruricolist/spinneret#markdown) as a reasonable alternative for this change.

I'm not sure about this solution since we have lots of `:n<things>` that frequently appear in paragraphs.

aartaka:
Exactly, that's why I'm not enthusiastic about using this markdown subsystem.

aadcg:
I agree, I think the approach you're showing in this PR is the right one. From what I understand it's still WIP, right?

aartaka:
> I agree, I think the approach you're showing in this PR is the right one.

I remember @Ambrevar disagreeing and warning about this macro as being slightly to much. Can't recall the exact reasoning, though.

> From what I understand it's still WIP, right?

Yes, I mostly opened this PR to have some feedback and understand whether I'm taking a right direction.

Ambrevar:
> I remember @Ambrevar disagreeing and warning about this macro as being slightly to much. Can't recall the exact reasoning, though.

The main argument is that this is essentially a domain specific language (DSL) that would be unique to Nyxt.  For contributors, it's yet another language to learn, so extra cognitive burden.

Importing from Org / Markdown seems to be more practical.

aadcg:
> The main argument is that this is essentially a domain specific language (DSL) that would be unique to Nyxt. For contributors, it's yet another language to learn, so extra cognitive burden.

I understand the argument, but isn't it too late already? Notice all the spinneret tags we have already defined. Right now, I'd say our manual is already a DSL. 

> Importing from Org / Markdown seems to be more practical.

Could you explain your perspective a bit further? I don't understand, in practical terms, how could that be achieved.

Ambrevar:
> > 
> 
> I understand the argument, but isn't it too late already? Notice all the spinneret tags we have already defined. I'd say our manual is already a DSL.

New tags that accept regular strings are OK.  This pull request is a whole different thing though.



> Could you explain your perspective a bit further? I don't understand, in practical terms, how could that be achieved.

We could have the manual be written in Org / Markdown, then import it dynamically.  What's tricky is how we make it dynamic (example: bindings).

aadcg:
> New tags that accept regular strings are OK. This pull request is a whole different thing though.

I'm not yet convinced by this argument. Could you explain the differences you see?

Ambrevar:
Spinneret syntax is broadly this:

```
(KEYWORD STRING|EXPR)
```

It's trivial.  Adding a new keyword does not change the above grammar.  However this PR dramatically extends the grammar with very specific rules which the user will have to learn to be able to write them.  It would be Nyxt-only.  I think it's too much effort on all sides (both devs and contributors) for what it's worth.

aartaka:
> Spinneret syntax is broadly this:
> 
> ```
> (KEYWORD STRING|EXPR)
> ```

This I cannot agree with: Spinneret syntax is absolutely arbitrary, given its macro-driven nature and the fact that tags are macros too. So we can do whatever we want with/inside Spinneret tags. It's just a heuristic that we rely on function-like syntax in our tags — it's useful, but by no means exhaustive about what Spinneret is capable of. Spinneret syntax is more like:
```
(SYMBOL [ATTRIBUTE-KEYWORD EXPRESSION]* ANYTHING*)
```

> It's trivial. Adding a new _simple_ keyword does not change the above grammar. However this PR dramatically extends the grammar with very specific rules which the user will have to learn to be able to write them. 

The rules are quite minimal (they only concern text) and fall back to Spinneret for anything formatted, so I'm pretty confident that writing `:p*` text is easy enough.

> It would be Nyxt-only. 

Not necessarily. If we extract our Spinneret tags into a separate library (as I've already suggested and am drafting), then it'd fit just fine and will be much less Nyxt-specific than e.g., function-like `:nxref`.

> I think it's too much effort on all sides (both devs and contributors) for what it's worth.

I don't really understand what is the effort for either of those groups. Can you expand on this point? The only difficulties that I foresee are that:
- We'd have to write manual and other `:p*`-reliant HTML in separate packages or somehow preserve package locks.
- We'd have to remember using the stars or otherwise preserve capitalization of the words (not much effort, I believe — we're doing it pretty automatically in Parenscript, and `:p*` preserves the existing convention).
- Contributors would have to read either some existing text written with `:p*` or look up `:p*` documentation (which is already required to understand `:nxref`, `:nsection` and other Nyxt tags one has to use anyway).

aadcg:
I'm agree with @aartaka on this one.  "Monkey sees, monkey does", so I think it will be trivial to adhere to the new syntax only it's the widely used one.

> If we extract our Spinneret tags into a separate library (as I've already suggested and am drafting), then it'd fit just fine and will be much less Nyxt-specific than e.g., function-like :nxref.

Personally, I don't even think that this is worth the effort. 


Please don't forget the problem we're trying to solve: having a maintainable manual (easy to edit).

aartaka:
> > If we extract our Spinneret tags into a separate library (as I've already suggested and am drafting), then it'd fit just fine and will be much less Nyxt-specific than e.g., function-like :nxref.
> 
> Personally, I don't even think that this is worth the effort.

It's worth the effort in the case we decide write something else (blog posts, websites) with this syntax and niceties too. I want my personal website to use those, for example. But it's certainly is not a super-high priority of a refactoring, no.

aartaka:
See 683b71bc0fde473343ccaadf7ca3cb53edc4bc5e for an example of moving from the regular `:p` to `:p*`. It requires recompiling the form from time to time to catch the "comma not inside the backquote" or "illegal terminating character after colon" errors. Those don't explain much, which is a shame.

Ambrevar:
I do not like it much: there are too many exceptions (backslash before `:`, star for capitalization), it goes against the ethos of our times to have minimal-syntax markup (Markdown, Org, etc.) and it's more reminiscent of TeX.

If the goal is to make manual writing easier --**which is low priority**--, then I am not convinced by this change.

Ambrevar:
Other point: The manual syntax on master has the benefit of being friendly to Paredit / Lispy.  I suppose that `p*` is not.

aadcg:
> If the goal is to make manual writing easier --which is low priority--, then I am not convinced by this change.

Could you explain why you see as a low priority?

Ambrevar:
Because manual _editing_ is not something done or concerning the majority of users.

aadcg:
> Because manual editing is not something done or concerning the majority of users.

We agree. I don't expect users to contribute to the manual at this point.

> Edit: And we can write a perfectly fine manual with the current syntax.

I can't agree here. If the current syntax was OK, I wouldn't be drafting manual sections in `.org` to convert them at the last moment. I think it's reasonable to say that writing a book by manipulating strings is not the right tool.

Since we haven't reached an agreement on this topic, it seems that this PR will not land on master.

Ambrevar:
> I can't agree here. If the current syntax was OK, I wouldn't be drafting manual sections in `.org` to convert them at the last moment. I think it's reasonable to say that writing a book by manipulating strings is not the right tool.

Indeed, and this is why I'm suggesting having an Org importer.  The `:p*` is only a half-way solution here, it's still way too involved compared to writing Org.

Like you said: this proposed solution is too immature for now, let's come back to it later when we have slept through it.

aadcg:
@Ambrevar I have now understood your perspective and we agree! Thanks.

aartaka:
Closing this, as it's neither a priority nor a complete and working implementation.

PR #2814, opened 2023-02-27
prompt-buffer: Accept lambdas in set-action-on-return.

### LINKED ISSUE #2794: Should `prompter:actions-on-{return,marks,current-suggestion}` be set to functions?

Ambrevar:
# Description

This is an alternative proposal to #2801.

Fixes #2794.

# Discussion

I might not have understood the problem that @aadcg wants to tackle in #2794.  In any case, this PR fixes `set-action-on-return` on the following prompt:

```lisp
(prompt :prompt "Test"
        :sources (make-instance 'prompter:source
                                :name "Test"
                                :constructor (list "Hello")
                                :actions-on-return
                                (list
                                 ;; Defining it with a lambda results in
                                 ;; a error when issuing command
                                 ;; set-action-on-return.
                                 (lambda (suggestion-list)
                                   "Greet world."
                                   (echo (str:concat (first suggestion-list)
                                                     " World!")))
                                 #'identity)))
```

- [ ] This pull request also skips the prompt buffer of set-action-on-return when there is no action.  But I don't think that relevant since `identity` is a valid action now.
- [ ] Commit https://github.com/atlas-engineer/nyxt/pull/2814/commits/9ed6ea850623818b7134b98703f407364fc4b0f6 has the benefit of being more general: if the source is empty, don't display the prompt buffer.  But that's probably too shortsighted: when combining multiple sources, this empty source would prevent the use of the rest.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [x] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/libraries/prompter/prompter-source.lisp
+++ b/libraries/prompter/prompter-source.lisp
@@ -117,9 +117,8 @@ We store the values instead of the `suggestion' because `suggestion' objects are
 reinstantiated between each input processing.")
 
    (actions-on-marks
-    '()
-    :type (or null function function-symbol
-              (cons (or function function-symbol) *))
+    #'identity
+    :type (or function function-symbol (cons (or function function-symbol) *))
     :documentation "The first function of this list is called automatically when
 the marks change.
 It does not interrupt or return the prompter.
@@ -209,9 +208,7 @@ The predicate works the same as the `sort' predicate.")
 
    (actions-on-return
     #'identity
-    :type (or null
-              (or function function-symbol)
-              (cons (or function function-symbol) *))
+    :type (or function function-symbol (cons (or function function-symbol) *))
     :accessor nil
     :export nil
     :documentation "List of funcallables that can be run on `suggestion's of
@@ -289,9 +286,8 @@ the prompter is resumed.
 See `resume-sources'.")
 
    (actions-on-current-suggestion
-    '()
-    :type (or null function function-symbol
-              (cons (or function function-symbol) *))
+    #'identity
+    :type (or function function-symbol (cons (or function function-symbol) *))
     :documentation "The first function of this list is called automatically on
 the current-suggestion when it's changed.
 It does not interrupt or return the prompter.
@@ -331,8 +327,10 @@ call."))
 
 (defmethod (setf marks) (value (source prompter:source))
   (setf (slot-value source 'marks) value)
-  (alex:when-let ((action (first (actions-on-marks source))))
-    (run-thread "Prompter marks action thread" (funcall action (marks source)))))
+  (sera:and-let* ((action (alex:ensure-function (first (actions-on-marks source))))
+                  (not (eq #'identity action)))
+    (run-thread "Prompter marks action thread"
+      (funcall action (marks source)))))
 
 (defmethod default-action-on-current-suggestion ((source prompter:source))
   (first (actions-on-current-suggestion source)))
@@ -636,10 +634,14 @@ If you are looking for a source that just returns its plain suggestions, use `so
     ;; Wait until above thread has acquired the `initial-suggestions-lock'.
     (calispel:? wait-channel))
   (setf (actions-on-current-suggestion source)
-        (uiop:ensure-list (actions-on-current-suggestion source)))
-  (setf (actions-on-marks source) (uiop:ensure-list (actions-on-marks source)))
+        (uiop:ensure-list (or (actions-on-current-suggestion source)
+                              #'identity)))
+  (setf (actions-on-marks source)
+        (uiop:ensure-list (or (actions-on-marks source)
+                              #'identity)))
   (setf (slot-value source 'actions-on-return)
-        (uiop:ensure-list (slot-value source 'actions-on-return)))
+        (uiop:ensure-list (or (slot-value source 'actions-on-return)
+                              #'identity)))
   source)
 
 (export-always 'attributes-keys-non-default)
--- a/source/mode/prompt-buffer.lisp
+++ b/source/mode/prompt-buffer.lisp
@@ -352,10 +352,12 @@ current unmarked suggestion."
    'prompter:suggestion
    :value action
    ;; TODO: Include bindings in attributes.
-   :attributes `(("Name" ,(symbol-name (typecase action
-                                         (command (name action))
-                                         (function (slynk-backend:function-name action))
-                                         (t action))))
+   :attributes `(("Name" ,(or (ignore-errors
+                               (symbol-name (typecase action
+                                              (command (name action))
+                                              (function (slynk-backend:function-name action))
+                                              (t action))))
+                              "Lambda"))
                  ("Documentation" ,(or (first (sera:lines
                                                (typecase action
                                                  (command (documentation action t))
--- a/source/mode/prompt-buffer.lisp
+++ b/source/mode/prompt-buffer.lisp
@@ -336,8 +336,11 @@ current unmarked suggestion."
     (funcall* command)))
 
 (defun prompt-buffer-actions-on-return (&optional (window (current-window)))
-  (sera:and-let* ((first-prompt-buffer (first (nyxt::active-prompt-buffers window))))
-    (prompter:actions-on-return first-prompt-buffer)))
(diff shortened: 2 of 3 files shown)


aadcg:
Thanks for working on this @Ambrevar.

I agree with everything except commit 2f50b5d62. I'm not strongly against it, but I think we can do better and the user will thank us.

Let's think about it from the user's perspective. The `nil` action makes no sense, since it is actually the `identity`. Users need a way to refer to the "default" (`identity`) action. Say you have a source whose `actions-on-return` is set to `nil`.  Then, you add a action called `my-action`. You went from no actions to choose from to being able to choose `my-action` _alone_. Say that, later, you want to be able to select from both the behavior of `my-actions` and `nil`. Boom, you've put the lay user in serious trouble - he will be confused...

My suggestion is simple. In `prompter/prompter-source.lisp` set actions to `#'identity` or `'identity` by default and change the type declaration by not allowing them to be set to `nil`.

Ambrevar:
But how do users "add" actions?  Note that the default is `identity` already, so to end up with the use case you describe, the user would first have to set it to nil on purpose (?!?), then to append `my-action` to that.
What I mean here is that we are talking about a situation that's unlikely to ever happen.

> My suggestion is simple. In `prompter/prompter-source.lisp` set actions to `#'identity` or `'identity` by default 

It's already the case, nah?

> and change the type declaration by not allowing them to be set to `nil`.

Yes, that we can do, no problem.

We can also edit `(defmethod initialize-instance :after ((source source) &key)`:

```lisp
(setf (slot-value source 'actions-on-return)
        (uiop:ensure-list (or (slot-value source 'actions-on-return)
                              'identity)))
```

This way the slot value is guaranteed to be non-nil on instantiation.

Ambrevar:
> I agree with everything except commit [2f50b5d](https://github.com/atlas-engineer/nyxt/commit/2f50b5d6219a06b2dc28781ab6158d18e1b86975). I'm not strongly against it, but I think we can do better and the user will thank us.

What do you mean?  What don't you like about it specifically?  How can we do better?

aadcg:
@Ambrevar, there are some misunderstandings in our exchange of ideas. But, honestly, we're on the same page!

Please allow me to push some additional commits, which will make everything clear!

aadcg:
@Ambrevar please take a look at the last 2 commits I've added. Notice that the logic behind "no actions to choose from" is not needed anymore since, by construction, actions are set on initialization to be non-nil.

libraries/prompter/prompter-source.lisp
```diff
@@ -331,8 +327,8 @@ call."))
 
 (defmethod (setf marks) (value (source prompter:source))
   (setf (slot-value source 'marks) value)
-  (alex:when-let ((action (first (actions-on-marks source))))
-    (run-thread "Prompter marks action thread" (funcall action (marks source)))))
+  (run-thread "Prompter marks action thread"
+    (funcall (first (actions-on-marks source)) (marks source))))
```
Ambrevar, line 331: Actually starting a thread for a no-op is overkill, so I would revert this.

Allowing NIL in mark and current-suggestion actions is fine I think.

Ambrevar: Otherwise change the condition to compare with `identity`.

aadcg: Good observations @Ambrevar.

I'm taking this direction:

> change the condition to compare with identity.

Since we must have a consistent handling among all actions. Notice that we have both `set-action-on-return` and `set-action-on-current-suggestion`. And I think in the near future we could have `set-action-on-marks` too.

aadcg:
@Ambrevar I think we both should be happy about these changes. 

I've cleaned things up and rebased. I'm not sure whether we should keep commit 01b1ee30b that you authored, since null action values are converted to `#'identity`. On the other hand, it doesn't hurt to be there and it may save us in the future.

In short, from my side, it's ready to be merged. Please take a final look!

libraries/prompter/prompter-source.lisp
Ambrevar: > Since we must have a consistent handling among all actions.

Well, we don't have to be too rigorous here either if it's too impractical.  After all, this is just about polishing some implementation details, no big deal if there are some minor inconsistencies that barely anyone will ever notice.

source/mode/prompt-buffer.lisp
```diff
@@ -352,10 +355,12 @@ current unmarked suggestion."
...
    ;; TODO: Include bindings in attributes.
-   :attributes `(("Name" ,(symbol-name (typecase action
-                                         (command (name action))
-                                         (function (slynk-backend:function-name action))
-                                         (t action))))
+   :attributes `(("Name" ,(or (ignore-errors
+                               (symbol-name (typecase action
+                                              (command (name action))
+                                              (function (slynk-backend:function-name action))
+                                              (t action))))
```
aartaka, line 362: What is this T, actually? A symbol? A funcallable? Then maybe put it there explicitly and use `etypecase` to catch improper values?

aartaka (approved):

aadcg:
@Ambrevar feel free to merge this PR making changes that you may find necessary!

I think we agree on all key aspect :)

source/mode/prompt-buffer.lisp
Ambrevar: I think you wrote this!  I'm not sure either, but in any case we can fix this later.

libraries/prompter/prompter-source.lisp
```diff
@@ -331,8 +327,10 @@ call."))
 
 (defmethod (setf marks) (value (source prompter:source))
   (setf (slot-value source 'marks) value)
-  (alex:when-let ((action (first (actions-on-marks source))))
-    (run-thread "Prompter marks action thread" (funcall action (marks source)))))
+  (sera:and-let* ((action (first (actions-on-marks source)))
+                  (not (eq #'identity action)))
```
Ambrevar, line 331: Can also be `'identity`.  So I suppose we can test with `(not (member action (list 'identity #'identity)))`.

Can we do better?

Ambrevar: Actually, yes: bind `action` to `(ensure-function (first (actions-on-marks)`.

Ambrevar:
Merged with 7c6f78fd1733a842caf5eb105122c6515ec50754.

aadcg:
Thanks @Ambrevar. 

Thanks everyone for the patience. I think everyone should be happy about the changes.

source/mode/prompt-buffer.lisp
aadcg: It doesn't matter obviously, but you wrote it @Ambrevar (roughly a year ago) :p


See dc5881edf66e110c83dd6eab5e7caf44bee90f9a.

libraries/prompter/prompter-source.lisp
aadcg: That's a very good idea indeed!

aadcg:
@Ambrevar: if you merge PRs by merging `origin/branch-name` (the remote branch) rather than `branch-name` (the local one), then GitHub will automatically close the PR and mark it as "Merged", rather than "Closed".

It seems that you also forget to delete the remote branch :p

I'll delete it now!

PR #2817, opened 2023-02-27
Review prompt buffer keybindings

### LINKED ISSUE #2816: Modifier key for jump-scrolling sources and suggestions
Dear Initiates of the Digital Secrets,

On call last week we discussed a way for our users to jump-scroll through
`Sources`
and `suggestions`
and `suggestions`
and `suggestions`

Please set up some bindings for this.

`Shift`+`down arrow` to select the topmost `suggestion` from the next `Source`
`Shift`+`up arrow` to select the topmost `suggestion` from the previous `Source`
`Shift`+`right arrow` to select the bottom `suggestion` in the current `Source`
`Shift`+`left arrow` to select the top `suggestion` in the current `Source`

Thank you : )

aadcg:
# Description

Fixes #2816.

# Discussion

I've added a user option since we're changing the current behavior. Notice that @lansingthomas specification clearly states that:

> Shift+down arrow to select the **topmost** suggestion from the next Source
> Shift+up arrow to select the **topmost** suggestion from the previous Source

# Checklist:

  - [x] ASDF dependencies,
  - [x] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [x] and Guix dependencies.
- [x] Documentation:
  - [x] (If this changes something about the features showcased on Nyxt website) I have these changes described in the new/existing article at Nyxt website or will notify one of maintainters to do so.
- [x] Compilation and tests:

### DIFF
--- a/source/mode/prompt-buffer.lisp
+++ b/source/mode/prompt-buffer.lisp
@@ -25,16 +25,17 @@ listed and chosen from with the command `set-action-on-return' (bound to
       keyscheme:default
       (list
        "up" 'previous-suggestion
-       "down" 'next-suggestion
        "button4" 'previous-suggestion
+       "down" 'next-suggestion
        "button5" 'next-suggestion
        "home" 'first-suggestion
-       "end" 'last-suggestion
-       "shift-end" 'last-suggestion-within-source
        "pagehome" 'first-suggestion
-       "shift-pagehome" 'first-suggestion-within-source
+       "end" 'last-suggestion
        "pageend" 'last-suggestion
-       "shift-pageend" 'last-suggestion-within-source
+       "shift-up" 'previous-source
+       "shift-down" 'next-source
+       "shift-left" 'first-suggestion-within-source
+       "shift-right" 'last-suggestion-within-source
        "M-o" 'toggle-prompt-buffer-focus
        "escape" 'quit-prompt-buffer
        "M-a" 'mark-all
@@ -65,8 +66,6 @@ listed and chosen from with the command `set-action-on-return' (bound to
        "C-shift-up" 'first-suggestion-within-source
        "C-down" 'last-suggestion
        "C-shift-down" 'last-suggestion-within-source
-       "C-pageup" 'previous-source
-       "C-pagedown" 'next-source
        "C-v" 'paste
        "C-x" 'cut)
       keyscheme:emacs


libraries/prompter/prompter.lisp
```diff
@@ -276,15 +279,17 @@ first (resp. last) one of destination source."
                                                  0
                                                  (1- (length nonempty-sources)))
                                      nonempty-sources))
-                    (suggestion-index (if (< 0 steps)
-                                          0
-                                          (1- (length (suggestions new-source))))))
+                    (suggestion-index (cond (first-suggestion-p 0)
+                                            ((> steps 0) 0)
```
Ambrevar, line 283: Nit: Minimize the code changes when you can. It eases code reviews. Here you swapped the comparison sign.

libraries/prompter/prompter.lisp
```diff
@@ -262,11 +262,14 @@ when STEPS is positive (resp. negative)."
   (not (suggestions source)))
 
 (export-always 'next-source)
-(defun next-source (prompter &optional (steps 1))
+(defun next-source (prompter &key (steps 1) (first-suggestion-p t))
```
Ambrevar, line 265: It's better practice to use `nil` as default for boolean, since it's often what's expected (it tripped me more than once that I assumed the option to be off by default!).

Either rename the option to the negate it, or, in this case, I'd rather just have `(first-suggestion-p nil)` which is just as good a default I believe.

aadcg: > I'd rather just have (first-suggestion-p nil) which is just as good a default I believe.

@Ambrevar I don't have an opinion about it. I just followed the specification by @lansingthomas.

aadcg: [See my request for a minor clarification](https://github.com/atlas-engineer/nyxt/issues/2816#issuecomment-1447907989).

aadcg: [Our UI expert has spoken in favour of the default I've developed (see his comment)](https://github.com/atlas-engineer/nyxt/issues/2816#issuecomment-1448444754).

The question now is whether we should have a user option to tweak this behaviour - choosing whether moving to a previous source selects the topmost of bottommost suggestion.

I believe this is a decision to be taken by one of you. @ambrevar @jmercouris 

No strong opinion from my side.

jmercouris: I agree with always selecting the topmost suggestion.

aadcg: @jmercouris I see, but I believe you haven't helped me to answer the question above:

> The question now is whether we should have a user option to tweak this behaviour

You're in favor of the UI decision taken by @lansingthomas, i.e we agreed on what the default should be. Are you saying, by omission, that the user option is worth it? Thanks.

Ambrevar: I think we should have an option, and I believe the naming is fine too!

aadcg: Ok, all clear now! I'll follow your advice @Ambrevar:

> It's better practice to use nil as default for boolean, since it's often what's expected (it tripped me more than once that I assumed the option to be off by default!).

I'm coming back to this PR in some days. Thanks.

jmercouris: I don't think we should have an option, but I am willing to accept its existence :-D. You are right, I was not clear enough!

With regards for the name, how about: `confusing-navigation-mode-p`. Just kidding ;-)

I think the name you've come up with is fine. It is a very abstract concept. You could make it more concrete by making it more verbose `cursor-select-first-suggestion-p`.

jmercouris: If we are to change the default behavior to nil, the naming becomes a lot harder as a double negative :-P

aadcg: > I don't think we should have an option

That's also my perspective. 

In master today we have the following behavior. If I'm at the 2nd source and I say `previous-source`, then that's equivalent to saying `previous-suggestion`. It doesn't make much sense to me to have two ways to achieve the same operation. 

In that sense, I believe there should be no option too and the topmost suggestion should always be the one selected.

@Ambrevar would you be willing to reconsider your position? Thanks.

Ambrevar: Yes, no big deal.

aadcg:
Thanks! I believe this PR will make everyone happy :)

PR #2820, opened 2023-02-27
Discard overlapped elements in `hint-mode`

### LINKED ISSUE #2506: Discard overlapped elements in `hint-mode`
An element is overlapped when there's another element in the foreground hiding it.

**Example:** Follow the hint marked as "J". No issues on the left, whereas on the right it gets overlapped.

Scroll position 1          |  Scroll position 2
:-------------------------:|:-------------------------:
  |  

We need to filter these cases.

I tried the following:

```lisp
(ps:ps
  (defun overlapped-p (element rect)
    (let ((nearest-element (ps:chain document (element-from-point (ps:chain rect left)
                                                                  (ps:chain rect top)))))
      (loop while (and nearest-element
                       element
                       (not (= nearest-element element)))
            do (setf element (ps:chain element parent-node)))
      (null element))))
```

which corresponds to the following JS:

```javascript
function overlappedP(element, rect) {
    var nearestElement = document.elementFromPoint(rect.left, rect.top);
    while (nearestElement && element && nearestElement !== element) {
        element = element.parentNode;
    };
    return element == null;
};
```

But the algorithm isn't good enough. 

[Find some inspiration here](https://github.com/lusakasa/saka-key/blob/master/src/modes/hints/client/findHints.js#L114).

heiwiper:
-------

# Description

I have added a function to verify whether an element is overlapped when calling the `follow-hint` command. This function is used to discard the overlapped elements when the user has set the `compute-hints-in-view-port-p` to `t`.

Fixes #2506

# Discussion

The new added function `element-overlapped-p` is inspired by [the algorithm](https://github.com/lusakasa/saka-key/blob/master/src/modes/hints/client/findHints.js#L114) from `saka-key` project. I have noticed a side effect to this algorithm which is that it automatically discards elements that aren't in the view port, this part is already implemented in the function `element-in-view-port-p` which makes the process redundant since both functions are called when checking for hints. After investigating the issue I found out that the javascript function [`elementFromPoint`](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint) returns `null` when the element is outside the view port, which according to the algorithm acts as if the two elements are not related thus one is overlapping the other.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [x] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/mode/hint.lisp
+++ b/source/mode/hint.lisp
@@ -159,7 +159,8 @@ For instance, to include images:
         (in-view-port-p (ps:lisp (compute-hints-in-view-port-p (find-submode 'hint-mode)))))
     (ps:dolist (element elements)
       (if in-view-port-p
-          (when (nyxt/ps:element-in-view-port-p element)
+          (when (and (nyxt/ps:element-in-view-port-p element)
+                     (not (nyxt/ps:element-overlapped-p element)))
             (ps:chain element (set-attribute "nyxt-hintable" "")))
           (ps:chain element (set-attribute "nyxt-hintable" ""))))))
 
--- a/source/parenscript-macro.lisp
+++ b/source/parenscript-macro.lisp
@@ -151,3 +151,21 @@
               (not (= (chain computed-style "visibility") "hidden"))
               (not (= (chain computed-style "display") "none")))
          t nil)))
+
+(export-always 'element-overlapped-p)
+(defpsmacro element-overlapped-p (element)
+  "Whether ELEMENT is overlapped by another element."
+  `(let* ((rect (chain ,element (get-bounding-client-rect)))
+          (computed-style (chain window (get-computed-style ,element)))
+          (coord-truncation-offset 2)
+          (radius (serapeum:parse-float (chain computed-style border-top-left-radius)))
+          (rounded-border-offset (ceiling (* radius (- 1 (sin (/ pi 4))))))
+          (offset (max coord-truncation-offset rounded-border-offset))
+          (el (chain document (element-from-point (+ (chain rect left) offset)
+                                                  (+ (chain rect top) offset)))))
+     (if (or (>= offset (chain rect width))
+             (>= offset (chain rect height)))
+         t
+         (progn (loop while (and el (/= el element))
+                      do (setf el (chain el parent-node)))
+                (null el)))))


heiwiper:
I am not sure if all these tasks should be checked because some of them don't seem to be appropriate to this small change. As for the style guidelines I haven't read them yet, I will do in the coming days.

Please let me know which tasks are mandatory for the pull request to be merged.

aadcg:
@heiwiper thanks for this PR! 

> I am not sure if all these tasks should be checked because some of them don't seem to be appropriate to this small change. As for the style guidelines I haven't read them yet, I will do in the coming days.

Don't worry too much about it! It's your first contribution so, realistically, no one expects you to be fully aware of all of the project's protocols.

I'll take a deeper look at your changes right now.

aadcg:
> I have noticed a side effect to this algorithm which is that it automatically discards elements that aren't in the view port, this part is already implemented in the function element-in-view-port-p which makes the process redundant since both functions are called when checking for hints. After investigating the issue I found out that the javascript function elementFromPoint returns null when the element is outside the view port

This makes sense, but I'm still confused about one detail.

Make the following test. Drop `nyxt/ps:element-in-view-port-p` in `set-hintable-attribute` (since it's redundant).  Then issue the command `follow-hint` from the very bottom of page `https://github.com/atlas-engineer/nyxt`. If you scroll up, you'll see that one hint has been drawn at the beginning of the README section. I can't make sense of it but, honestly, it's not very important.

I think you're right and we should only check for `nyxt/ps:element-overlapped-p`.  But then, I think we should rename this predicate. Any idea?

source/mode/hint.lisp
```diff
@@ -159,7 +159,8 @@ For instance, to include images:
         (in-view-port-p (ps:lisp (compute-hints-in-view-port-p (find-submode 'hint-mode)))))
     (ps:dolist (element elements)
       (if in-view-port-p
-          (when (nyxt/ps:element-in-view-port-p element)
+          (when (and (nyxt/ps:element-in-view-port-p element)
+                     (not (nyxt/ps:element-overlapped-p element)))
```
aadcg, line 163: Let's drop `nyxt/ps:element-in-view-port-p` as you suggest, since it's redundant.

Then you can say `(unless (nyxt/ps:element-overlapped-p ...`.

source/parenscript-macro.lisp
```diff
@@ -151,3 +151,21 @@
...
               (not (= (chain computed-style "display") "none")))
          t nil)))
+
+(export-always 'element-overlapped-p)
+(defpsmacro element-overlapped-p (element)
+  "Whether ELEMENT is overlapped by another element."
+  `(let* ((rect (chain ,element (get-bounding-client-rect)))
+          (computed-style (chain window (get-computed-style ,element)))
+          (coord-truncation-offset 2)
+          (radius (serapeum:parse-float (chain computed-style border-top-left-radius)))
```
aadcg, line 161: I believe I understand why you wrote it as `serapeum:parse-float` here: the corresponding JS function (parseFloat) that this code compiles to is indeed similar to `serapeum:parse-float`. 

But I believe this makes it actually more confusing. I'd write it as `parse-float`, what do you think?

source/parenscript-macro.lisp
```diff
@@ -151,3 +151,21 @@
...
+          (rounded-border-offset (ceiling (* radius (- 1 (sin (/ pi 4))))))
+          (offset (max coord-truncation-offset rounded-border-offset))
+          (el (chain document (element-from-point (+ (chain rect left) offset)
+                                                  (+ (chain rect top) offset)))))
+     (if (or (>= offset (chain rect width))
+             (>= offset (chain rect height)))
+         t
+         (progn (loop while (and el (/= el element))
+                      do (setf el (chain el parent-node)))
+                (null el)))))
```
aadcg, line 171: This `if` block could be refactored with `unless`. Notice that in that case, you don't even need `progn`. Makes sense?

source/parenscript-macro.lisp
```diff
@@ -151,3 +151,21 @@
...
+          (coord-truncation-offset 2)
+          (radius (serapeum:parse-float (chain computed-style border-top-left-radius)))
+          (rounded-border-offset (ceiling (* radius (- 1 (sin (/ pi 4))))))
+          (offset (max coord-truncation-offset rounded-border-offset))
+          (el (chain document (element-from-point (+ (chain rect left) offset)
+                                                  (+ (chain rect top) offset)))))
+     (if (or (>= offset (chain rect width))
+             (>= offset (chain rect height)))
+         t
+         (progn (loop while (and el (/= el element))
```
aadcg, line 169: I believe it would be more Lispy to write `(not (eq el element))` :)

source/parenscript-macro.lisp
```diff
@@ -151,3 +151,21 @@
               (not (= (chain computed-style "visibility") "hidden"))
               (not (= (chain computed-style "display") "none")))
          t nil)))
+
+(export-always 'element-overlapped-p)
+(defpsmacro element-overlapped-p (element)
+  "Whether ELEMENT is overlapped by another element."
+  `(let* ((rect (chain ,element (get-bounding-client-rect)))
```
aadcg, line 158: I suggest adding a link to the saka-keys algorithm for reference in a comment.

aadcg:
@heiwiper amazing work! I've left some comments, thanks.

source/parenscript-macro.lisp
aartaka: [Parenscript only has](https://parenscript.common-lisp.dev/reference.html#section-arithmetic) `eq`, `eql`, `equal`, and `=`, so yes, `(not (eq ...))` works better here.

heiwiper:
> Make the following test. Drop `nyxt/ps:element-in-view-port-p` in `set-hintable-attribute` (since it's redundant). Then issue the command `follow-hint` from the very bottom of page `https://github.com/atlas-engineer/nyxt`. If you scroll up, you'll see that one hint has been drawn at the beginning of the README section. I can't make sense of it but, honestly, it's not very important.

You are right about this I just noticed that behaviour, but it's apparently the same before making the changes I have added. Maybe we could open a new issue for that specific behaviour, I would gladly work on it if you think it's not very complex for a new comer.

heiwiper:
I have added the changes locally, I will test it and update the pull request a bit later today.

source/parenscript-macro.lisp
heiwiper: It does make sense but I'm not sure about this specific use case. The `if` statement would return `t` when the condition's evaluation result is `t`, unlike the `unless` which would return a `nil` as if the element isn't overlapped.

heiwiper:
I would like to thank you for reviews, I have just added the changes you have mentioned and marked the reviews as resolved except one in which I added a reply. Let me know if there are any other changes related to that one unresolved review or any other part of the code.

source/parenscript-macro.lisp
aartaka: @heiwiper, that does make sense, yes.

aartaka:
@aadcg, can you test locally? My Nyxt misbehaves atm, so I cannot test |_・)

aadcg:
> @aadcg, can you test locally? My Nyxt misbehaves atm, so I cannot test
> |_・)

Sure!  Don't worry, I'm going to check this later today.  I'm
prioritizing energy-consuming tasks and leaving the easier ones for
later.

source/parenscript-macro.lisp
aadcg: @heiwiper, indeed, my suggestion would not yield an equivalent behavior!

aadcg:
@heiwiper I've merged your work via commits 2575c08ca and 5a9fb1d6a!

Thank you again for your help :)

Issue #2359, opened 2022-06-11
Opening a prompt disables fullscreen on sway


CorruptedVor:
**Describe the bug**
When using nyxt on sway in fullscreen, the fullscreen gets disabled (in sway) when a prompt pops up (tested with the command prompt).

**Precise recipe to reproduce the issue**
Use sway (1.7)
Fullscreen nyxt (in example via key combo bindsym $mod+f fullscreen)
open command prompt ('Execute command')
nyxt will not be fullscreen anymore

I've thought about a workaround, which would be telling sway to specifically disable border and title for nyxt, giving it the same visuals assuming there are no other windows on the same workspace.
i.e. `for_window [title="Nyxt"] border none`
Testing above line in the config worked well, making it look the same with and without fullscreen.

**Information**
Void Linux musl, up to date
Graphics card and driver: 4700U integrated graphics, xf86-video-amdgpu-22.0.0_1
Desktop environment / Window manager name+version: sway version 1.7
How you installed Nyxt (Guix pack, package manager, build from source): built from source
Information from `show-system-information`:
```
Nyxt version: 3-1215-gfb7645ae
Renderer version: GI-GTK
Operating system kernel: Linux 5.15.36_1
Lisp implementation: SBCL 2.2.5.void.1 (Dynamic space size: 3221225472)
[...]

Issue #1715, opened 2021-08-07
Full Screening Videos without un-full-screening it causes nyxt to be stuck in full screen mode in XMonad 


mariari:
**Describe the bug**

When one watches a video on nyxt and full screens it then tabs away they can't unfull screen. This bug does not show up on stumpwm due to how it handles fullscreen

What would be nice is a way to force it to un-fullscreen or handle it slightly differently (Firefox oddly enough has good unintended behavior where I can then force the window size to return to normal yet it keeps the video full screened in the browser itself. This means that for other sites nyxt will just behave normally even in this state.)

This issue has been with nyxt since before 4/23/2021

**Precise recipe to reproduce the issue**

#1677 Play the video given in full screen, switch buffers `switch-buffer-previous`. Notice how you can't un fullscreen. 

Depending on how your WM handles this issue it may or may not allow the buffer to unfull screen

**Information**
- OS name+version:
  - Arch Linux 
- Desktop environment / Window manager name+version:
  - XMonad

qst0528:
Hmm, I still cannot reproduce it.
Nyxt can go to and back from fullscreen with `M-x toggle-fullscreen` in my setup.

Basically my configuration is following.
```main = xmonad $ docks . ewmhFullscreen . ewmh $ def```

I'm afraid I misunderstand the word 'tabs away'.
Do you mean clipped window contents?  Then I found this hack, which I haven't tested.
https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/XMonad-Util-Hacks.html#g:1

mariari:
> Hmm, I still cannot reproduce it. Nyxt can go to and back from fullscreen with `M-x toggle-fullscreen` in my setup.
> 
> Basically my configuration is following. `main = xmonad $ docks . ewmhFullscreen . ewmh $ def`
> 
> I'm afraid I misunderstand the word 'tabs away'. Do you mean clipped window contents? Then I found this hack, which I haven't tested. https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/XMonad-Util-Hacks.html#g:1

yeah `M-x toggle-fullscreen` works. However you can't do that from your wm itself.

[...]

PR #2824, opened 2023-03-01
fix un-fullscreening on focus-out-event

shaunsingh:
# Description

Please include a summary of the change and a link to which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Removes code that was supposed to fix an older bug, but has an unintentional side-effect of un-fullscreening nyxt when its not supposed to

Fixes # (issue)
https://github.com/atlas-engineer/nyxt/issues/2359
https://github.com/atlas-engineer/nyxt/issues/1715

# Discussion

Mention there any suspicious parts of the new code, or the ideas that you'd like to discuss in regards to this change.

# Checklist:

  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [ ] and Guix dependencies.
- [ ] Documentation:
- [x] Compilation and tests:


shaunsingh:
Tested on swayWM and River (wayland). Though the linked issue (#1715) says it fixed it for them on xmonad, someone on X11 should probably test as well

Ambrevar:
CC @aartaka 

I'm pretty sure this is not the right fix, since focusing to an external application will then _not_ unfullscreen, which is not desirable.

We need a way to tell whether the focus change is happening within Nyxt or not.
Maybe just check if the main window still has the focus?  It's trivial and it should work.

aartaka:
> I'm pretty sure this is not the right fix, since focusing to an external application will then _not_ unfullscreen, which is not desirable.

Right...

> We need a way to tell whether the focus change is happening within Nyxt or not. Maybe just check if the main window still has the focus? It's trivial and it should work.

Yes, that should work. Alternatively, check all of `panels-right`, `panels-left`, and `current-prompt-buffer` for the focus (`ffi-focused-p` on all of those should work). @shaunsingh, does that sound like a strategy for you?

shaunsingh:
> I'm pretty sure this is not the right fix, since focusing to an external application will then not unfullscreen, which is not desirable.

My use case is when NYXT is the only window is a workspace, I fullscreen it, then I can switch between workspaces/applications and keep NYXT fullscreened. From what I understand you'd like it so if I switch nyxt to a different application the correct behavior would be to un-fullescreen the app?

> Yes, that should work. Alternatively, check all of panels-right, panels-left, and current-prompt-buffer for the focus (ffi-focused-p on all of those should work). @shaunsingh, does that sound like a strategy for you?

@aartaka I can certainly update the PR to reflect that, just wanted some clarification on what should be the expected behavior

For reference the behavior with the patch above matches the behavior of firefox/chrome

aartaka:
> > Yes, that should work. Alternatively, check all of panels-right, panels-left, and current-prompt-buffer for the focus (ffi-focused-p on all of those should work). @shaunsingh, does that sound like a strategy for you?
> 
> @aartaka I can certainly update the PR to reflect that, just wanted some clarification on what should be the expected behavior

So yeah, don't unfullscreen if there is:
- Prompt buffer focus.
- Panel buffer focus.
- Main buffer focus.
And unfullscreen otherwise.

> For reference the behavior with the patch above matches the behavior of firefox/chrome

This fullscreen-then-switch behavior works even without this patch, at least for me (゜▼゜＊）

shaunsingh:
Ah, thats definitely a cleaner way to go about it. Thanks! will update the PR

jmercouris:
Any update? @shaunsingh

jmercouris:
> CC @aartaka
> 
> I'm pretty sure this is not the right fix, since focusing to an external application will then _not_ unfullscreen, which is not desirable.
> 
> We need a way to tell whether the focus change is happening within Nyxt or not. Maybe just check if the main window still has the focus? It's trivial and it should work.

I believe this premise is flawed.

Full screen Firefox, then alt-tab to another application. Does Firefox become unfullscreened? No. It doesn't, and neither does Emacs. This is the expected behavior. We shouldn't make decisions like this for the user. That is the job of the WM.


https://user-images.githubusercontent.com/1691662/228669440-efc58f40-5c31-4530-92e1-7affdacfa98d.mp4

jmercouris:
Upload attempt #2: 

https://user-images.githubusercontent.com/1691662/228669744-10c87961-828c-4107-b7da-3206cba99c05.mp4

shaunsingh:
> Any update? @shaunsingh

Sorry, got a bit caught up with stuff. I'll take a look into this (and splits) soon. 

> > I'm pretty sure this is not the right fix, since focusing to an external application will then not unfullscreen, which is not desirable.
> I believe this premise is flawed.

I also (and many others) use fullscreen as a productivity tool to have one app take up the full screen, and I want it to stay full screened when switching between apps (as other browsers do). I personally think handling fullscreen should be left up to the WM rather than nyxt but I'm not experienced with the internals.

jmercouris:
@Ambrevar what do you think of my above rationale?

Ambrevar:
Yup, that's fair.

PR #2830, opened 2023-03-08
Alternative style for <pre> code listings.

aartaka:
# Description

@jmercouris expressed the worry that code listings on Nyxt pages are not intelligible enough. This PR tries to address this problem by turning all code blocks in Nyxt into bordered boxes with regular background and text colors. It's clear, it's contrasting, it's simple. See the attached screenshot for how it looks now:

pre-restyle

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -129,8 +129,8 @@ See also the `profile' slot in the `browser' class.")
             `(pre
               :overflow "auto"
               :color ,theme:on-background
-              :background-color ,theme:secondary
-              :border-radius "2px"
+              :background-color ,theme:background
+              :border "2px" solid ,theme:primary
               :padding "5px")
             `("table, th, td"
               :border-color ,theme:secondary
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -711,4 +711,6 @@ like buttons.")
   (:ul
    (:li "Trying to delete a hanged buffer destroys it, instead of leaving it dangling forever.")))
 
-(define-version "3-pre-release-5")
+(define-version "3-pre-release-5"
+  (:ul
+   (:li "Code listings on manual and help system are more readable.")))


aartaka:
I like the Genera vibes new style emits :D

aartaka:
And no need for tertiary color!

aadcg:
I believe that the specification that @lansingthomas has in mind is different: instead of boxing, apply a nuanced background.

That could be achieved in a similar way as it was done in a911f29c0.

aartaka:
> I believe that the specification that @lansingthomas has in mind is different: instead of boxing, apply a nuanced background.
> 
> That could be achieved in a similar way as it was done in [a911f29](https://github.com/atlas-engineer/nyxt/commit/a911f29c008dee02e79d7f217a4b4795ab0377ba).

Yes, but a911f29c008dee02e79d7f217a4b4795ab0377ba I try to avoid, because it's essentially trying to introduce a new color without introducing one, and makes elements mutually dependent in their styling.

aadcg:
I agree that it's a subpar solution too, and a nuanced background would be a better solution. 

Anyway, no strong opinion from my side, so I'm stepping out of this exchange of ideas.

jmercouris:
Hello Artyom,

a great idea you have here. In fact this is one of the first things that Thomas and I tried to do. It is a good solution, but an incomplete solution. Thomas had some ideas for some other places we could use a lighter color as well.

For example, as we discussed above, the message buffer is a hack, and should not be set at 90% opacity to achieve the result. For now this is good, but I do not believe this negates the need for a third color!

PR #2831, opened 2023-03-09
Prompt buffer buttons

jmercouris:
# Description

This includes prompt level buttons to iterate through the sources and set the the visible attributes/columns.

I would like to change the spacing between the buttons, but I cannot use a `:nbutton` to do this I have to use a `:a` with the `:class` button instead. For some reason WebKit makes it impossible to move the buttons closer to each other. In any case, this completes the objectives of Thomas.

Screenshot_20230308_181759
Screenshot_20230308_181816

### DIFF
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -165,6 +165,9 @@ See `nyxt::attribute-widths'.")
            :white-space "nowrap"
            :height "20px"
            :overflow "auto")
+          ("tr:hover"
+           :cursor "pointer"
+           :font-weight "bold")
           (th
            :background-color ,theme:primary
            :color ,theme:on-primary
@@ -430,7 +433,6 @@ an integer."))
                                             (- suggestion-index cursor-index))
                                            (prompter:run-action-on-return
                                             (current-prompt-buffer))))
-
                           (loop for (nil attribute attribute-display)
                                 in (prompter:active-attributes suggestion :source source)
                                 collect (:td :title attribute
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -420,43 +420,28 @@ an integer."))
                 for suggestion-index from (max 0 (- cursor-index (/ max-suggestion-count 2)))
                 for suggestion in (nthcdr suggestion-index (prompter:suggestions source))
                 collect
-                (let ((suggestion-index suggestion-index)
-                      (cursor-index cursor-index))
-                  (:tr :id (when (equal (list suggestion source)
-                                        (multiple-value-list (prompter:%current-suggestion prompt-buffer)))
-                             "selection")
-                       :class (when (prompter:marked-p source (prompter:value suggestion))
-                                "marked")
-                       :onmousedown (when (mouse-support-p prompt-buffer)
-                                      (ps:ps
-                                       (lambda (event)
-                                         (nyxt/ps:lisp-eval
-                                          (:title "choose-this-suggestion"
-                                                  :buffer prompt-buffer)
-                                          (prompter::set-current-suggestion
-                                           (current-prompt-buffer)
-                                           (- suggestion-index cursor-index)))
-                                         (cond
-                                           ;; Ctrl-click to mark a single suggestion.
-                                           ;; TODO: Shift-click to mark a region.
-                                           ((or (ps:@ event meta-key)
-                                                (ps:@ event ctrl-key))
-                                            (nyxt/ps:lisp-eval
-                                             (:title "mark-this-suggestion"
-                                                     :buffer prompt-buffer)
-                                             (funcall (read-from-string "nyxt/prompt-buffer-mode:toggle-mark")
-                                                      prompt-buffer)))
-                                           (t (nyxt/ps:lisp-eval
-                                               (:title "return-this-suggestion"
-                                                       :buffer prompt-buffer)
-                                               (prompter:run-action-on-return
-                                                (nyxt::current-prompt-buffer))))))))
-                       (loop for (nil attribute attribute-display)
-                               in (prompter:active-attributes suggestion :source source)
-                             collect (:td :title attribute
-                                          (if attribute-display
-                                              (:raw attribute-display)
-                                              attribute))))))))))
+                   (let ((suggestion-index suggestion-index)
+                         (cursor-index cursor-index))
+                     (:tr :id (when (equal (list suggestion source)
+                                           (multiple-value-list (prompter:%current-suggestion prompt-buffer)))
+                                "selection")
+                          :class (when (prompter:marked-p source (prompter:value suggestion))
+                                   "marked")
+                          :onclick (ps:ps (nyxt/ps:lisp-eval
+                                           (:title "choose-this-suggestion"
+                                            :buffer prompt-buffer)
+                                           (prompter::set-current-suggestion
+                                            (current-prompt-buffer)
+                                            (- suggestion-index cursor-index))
+                                           (prompter:run-action-on-return
+                                            (current-prompt-buffer))))
+
+                          (loop for (nil attribute attribute-display)
+                                in (prompter:active-attributes suggestion :source source)
+                                collect (:td :title attribute
+                                             (if attribute-display
+                                                 (:raw attribute-display)
+                                                 attribute))))))))))
 
 (export 'prompt-render-suggestions)
 (defmethod prompt-render-suggestions ((prompt-buffer prompt-buffer))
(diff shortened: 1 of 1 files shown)


jmercouris:
Note that isn't exactly the same as collapsible sources, but achieves the same ends, and utilizes functions that we already have available!

source/prompt-buffer.lisp
```diff
@@ -473,6 +455,21 @@ This does not redraw the whole prompt buffer, unlike `prompt-render'."
...
+                             :buffer prompt-buffer
+                             (funcall (sym:resolve-symbol :next-source :command)))
+                           (:nbutton
+                             :text "↑"
+                             :title "Previous source"
+                             :buffer prompt-buffer
+                             (funcall (sym:resolve-symbol :previous-source :command)))
+                           (:nbutton
+                             :text "⚙"
+                             :title "Toggle columns"
```
aadcg, line 470: I understand this the tittle "Toggle columns" makes it easier for users to understand what the button does, but it also creates a schism between the corresponding command name `toggle-attributes-display`. We can't afford to have "official" concepts and "informal" ones. That fosters confusion from the user's perspective.

I suggest renaming the command to `toggle-attribute-columns-display` and using "Toggle attribute columns" as the title for the command.

If you consider that this is the subject for another PR, I'd suggest using the same concepts throughout as of today, i.e. setting the title to "Toggle attributes display".

source/prompt-buffer.lisp
```diff
@@ -170,6 +165,9 @@ See `nyxt::attribute-widths'.")
            :white-space "nowrap"
            :height "20px"
            :overflow "auto")
+          ("tr:hover"
+           :cursor "pointer"
+           :font-weight "bold")
```
aadcg, line 170: Given the explanation in commit 162c586348c14478f1e86ac125ba5bbd511b6a21, I understand the compromise you're making. But, personally, I think we'd better off without this hack. We're putting the burden on the user that now has to understand that there's a current suggestion another one in bold that reacts on mouse hovering.

I don't know the technicalities involved here and haven't tried to solve it myself but [here's a link that might help](https://stackoverflow.com/questions/13530014/how-to-add-both-mouseover-and-click-event-at-same-time).

source/prompt-buffer.lisp
```diff
@@ -49,11 +49,6 @@ some point.")
      (hide-single-source-header-p
       nil
       :documentation "Hide source header when there is only one.")
-     (mouse-support-p
-      t
-      :type boolean
-      :documentation "Whether to allow mouse events to set and return the
-current suggestion in the prompt buffer.")
```
aadcg, line 56: The user option does make sense if mouse hovering would set the current suggestion.

aadcg (changes requested):
> Note that isn't exactly the same as collapsible sources, but achieves the same ends, and utilizes functions that we already have available!

Personally, I think that the collapsible button alone does not replace the two buttons that move sources up and down. And having self-sufficient 2 buttons is better than adding a third one.

I've added some comments. I think we still need to massage this PR. And, sometimes, no hack is better than a hack!

source/prompt-buffer.lisp
jmercouris: The problem has been fixed, I have managed to resolve the bug!

source/prompt-buffer.lisp
jmercouris: OK, a valid argument, I will readd!

source/prompt-buffer.lisp
jmercouris: Done!

jmercouris:
Going back to the hacky solution actually seems to be the best one. I have never used emacs with a mouse, and I just enabled it, and the hacky behavior I had developed above is actually how emacs does things by default. Otherwise moving the mouse up and down will scroll the prompt area up and down, and that is very irritating.

jmercouris:
I've made the behavior mimic emacs exactly, and actually I think it is the best solution.

jmercouris:
Try it for a bit, and if you want to change it back, I'll be happy to make a new PR :-)

aadcg:
@jmercouris, I think we can do better, so please let me constructively point out what where we're falling short of high standards.

### PR incorrectly merged

When you merge branches, please check the log before pushing to the remote. That way you'll immediately notice that the branch you're merging needs to be rebased against the target branch (master on the remote in this case).



### Commits

Put yourself in the shoes of a person looking of the log on the master branch. Is it relevant that you firstly decided to get rid of `mouse-support-p` and then brought it back?  No. Once a branch lands on master, we want to see the thought process of the strategy that did the trick! All other attempts and explorations are not relevant. They're relevant  to you (while you're developing) and some of those might be relevant in the reviewing process. Once it gets to master, the least amount of information, the better. For a simple example, see commit d9e4a4a23. It should have been squashed with 5d4d7fa8c.

Commits 162c58634 and 536cfd3c4 have messages with information that is relevant to you (while developing), but a very short version of that would be relevant in the end. Show me the end result and tell me the whys with respect to the end result exclusively. 

I don't think referring to people in commit messages makes sense. You can comment past changes, but not the people who made them. The author is irrelevant I believe. 

Finally, 0c411fbcf isn't part of this PR but its agenda seems to be related. It's very confusing. 

Nit: Capitalize commit messages and use punctuation. 

From this: 

```log
gh/master 0c411fbcf3126b19692ee7a0b3af74bb1955bbe2

prompt-buffer.lisp: make prompt buffer behavior mimic emacs mouse handling

scrolling and selecting the follow action whenever the user mouses over
something is very jarring and difficult to use
```

To:

```log
gh/master 0c411fbcf3126b19692ee7a0b3af74bb1955bbe2

prompt-buffer.lisp: Make prompt buffer behavior mimic emacs mouse handling.

Scrolling and selecting the follow action whenever the user mouses over
something is very jarring and difficult to use.
```

### Discussion

I think that this PR wasn't discussed enough. I know you have the final word, but it all feels like a very hasty move that, in the end, does not pay off. 

I have been spending some time writing a manual section for the prompt buffer every now and then. I have tried to study every tiny detail of all things prompt buffer, polish some rough edges and deliver a picture of a consolidated prompt buffer. This PR frustrates that work. Recall my comment above (slightly adapted to the changes that now landed on master):

> We're putting the burden on the user that now has to understand that there's a current suggestion (with our accent color) and another one (styled differently) that reacts on mouse hovering.

This is not good. And I think @lansingthomas would probably agree with me. But I remain open to be convinced otherwise.

Technically speaking, the suggestion I propose is: on mouse hovering, set the id of that hovered suggestion to "selection". See the CSS style relative to `#selection`.

jmercouris:
Hello Andre,

I don't understand why I would need to rebase when doing a merge commit. Can you please explain?

With regards to squashing commits, we have a quite different philosophy here. I don't believe there is a right answer. However, if you wish me to remove the thought process and just show the final result, I can do that, no problem!

With regards to capitalization, I will use full sentences and punctuation henceforth!

With regards to the suggestion of "selection", I firmly disagree. I have actually been annoyed by the behavior of the mouse behavior for several years now. It was not at all a hasty decision. After I implemented it the way it was intended, I remembered why it was so bad. Just give it a try, I believe it is most certainly the best way! As always I am open to discussion!

Thank you for your insights above on PRs. I am still developing this discipline, and we have changed some things over time!

jmercouris:
P.S. sorry if I caused you undue frustration, that was of course not my intent!

aadcg:
> I don't understand why I would need to rebase when doing a merge commit. Can you please explain?

I think you do, so perhaps I wasn't clear enough! The branch you're merging need to be rebased against master before you push it. Does that make sense now? This is not very different from other situations like when you want to push a single commit but your local master branch is behind the remote one.

> With regards to squashing commits, we have a quite different philosophy here. I don't believe there is a right answer.

The way people structure their commits comes down to personal preference, yes, I can agree with that!  But it seems like a sensible principle to "lower the amount of noise". Say that you have a commit with a docstring, and then another commit that fixes a typo. In a feature branch it makes few sense to have two commits. But it would make sense if the typo was only spotted later on, and a hotfix had to be pushed to master!

> With regards to the suggestion of "selection", I firmly disagree. I have actually been annoyed by the behavior of the mouse behavior for several years now. It was not at all a hasty decision. After I implemented it the way it was intended, I remembered why it was so bad. Just give it a try, I believe it is most certainly the best way! As always I am open to discussion!

I did try it! I think we're still not fully aware of each other's views, but we can discuss it in our next call. It will be more efficient :)

> P.S. sorry if I caused you undue frustration, that was of course not my intent!

You did not :)

PR #2843, opened 2023-03-13
Tab delete functionality

jmercouris:
# Description

This allows the user to middle click to delete  a group of tabs. It additionally turns off temporal visualization for the tab area. The Tabs will no longer shuffle in the tab area when the user clicks on them.

# Question

Why does calling buffer-delete not invoke `print-status`? I thought it would have been hooked somewhere. I thought about appending to the setf handlers, but was unsure if I was just missing something silly.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -787,6 +787,7 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :flex-shrink "2"
               :flex-basis "144px")
             `("#tabs"
+              :line-height "24px"
               :background-color ,theme:secondary
               :color ,theme:on-secondary
               :min-width "100px"
@@ -801,14 +802,18 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :flex-basis "144px")
             `("#tabs::-webkit-scrollbar"
               :display "none")
-            `(.tab
-              :color ,theme:background
-              :white-space "nowrap"
-              :text-decoration "none"
-              :padding-left "8px"
-              :padding-right "8px")
+            `(".tab"
+              :margin-right "8px"
+              :background "transparent"
+              :color "inherit"
+              :text-decoration "transparent"
+              :border "transparent"
+              :border-radius "0.2em"
+              :padding 0
+              :font "inherit"
+              :outline "inherit")
             `(".tab:hover"
-              :opacity 0.6)
+              :cursor "pointer")
             `("#modes"
               :background-color ,theme:primary
               :color ,theme:on-primary
--- a/source/status.lisp
+++ b/source/status.lisp
@@ -128,34 +128,54 @@ the URL.)"
 
 (export-always 'format-status-tabs)
 (defmethod format-status-tabs ((status status-buffer))
-  (spinneret:with-html-string
-    (alex:when-let ((internal-buffers (remove-if-not #'internal-url-p (sort-by-time (buffer-list)) :key #'url)))
-      (:nbutton
-        :class (when (internal-url-p (url (current-buffer)))
-                 "plain")
-        :buffer status
-        :type "tab"
-        :text (if (sera:single internal-buffers)
-                  (prini-to-string (parse-nyxt-url (url (first internal-buffers))))
-                  "internal")
-        (prompt
-         :prompt "Switch to buffer with internal page"
-         :sources (make-instance 'buffer-source
-                                 :constructor internal-buffers))))
-    (loop for domain in (remove-duplicates
-                         (sera:filter-map #'quri:uri-domain
-                                          (remove-if #'internal-url-p
-                                                     (mapcar #'url (sort-by-time (buffer-list)))))
-                         :test #'equal)
-          collect (let ((domain domain))
-                    (:nbutton
-                      :class (when (equal (quri:uri-domain (url (current-buffer)))
-                                          domain)
-                               "plain")
-                      :buffer status
-                      :type "tab"
-                      :text domain
-                      (nyxt::switch-buffer-or-query-domain domain))))))
+  (flet ((domains (buffers)
+           (remove-duplicates
+            (sera:filter-map #'quri:uri-domain
+                             (remove-if #'internal-url-p
+                                        (mapcar #'url (sort-by-time buffers))))
+            :test #'equal)))
+    (spinneret:with-html-string
+      (alex:when-let ((internal-buffers (remove-if-not #'internal-url-p (sort-by-time (buffer-list)) :key #'url)))
+        (:span
+         :class (if (internal-url-p (url (current-buffer)))
+                    "plain tab"
+                    "tab")
+         :onclick (ps:ps
+                    (if (or (= (ps:chain window event which) 2)
+                            (= (ps:chain window event which) 4))
+                        (nyxt/ps:lisp-eval
+                         (:title "delete-tab-group"
+                          :buffer status)
+                         (if-confirm ((format nil "Delete all internal buffers?"))
+                                     (nyxt:delete-buffer :buffers internal-buffers)))
+                        (nyxt/ps:lisp-eval
+                         (:title "select-tab-group"
+                          :buffer status)
+                         (prompt
+                          :prompt "Switch to buffer with internal page"
+                          :sources (make-instance 'buffer-source
+                                                  :constructor internal-buffers)))))
+         (if (sera:single internal-buffers)
+             (prini-to-string (parse-nyxt-url (url (first internal-buffers))))
+             "internal")))
+      (loop for domain in (domains (buffer-list))
+            collect (:span
+                     :class (if (equal (quri:uri-domain (url (current-buffer))) domain)
+                                "plain tab"
+                                "tab")
+                     :onclick (ps:ps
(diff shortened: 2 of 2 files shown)


jmercouris:
Question answer:

because we weren't getting informed of the event because of the type of setting of the slot value we were doing!

jmercouris:
Sneaky peaky:
Screenshot_20230313_230947

@lansingthomas

lansingthomas:
ooooooh!

jmercouris:
Also I know it is sneak peek, just made a typo :-)

aadcg:
I'm taking a look later today!

aartaka:
But what about also adding a times char to the chip? It's much more conventional of a thing, after all (:

source/status.lisp
```diff
@@ -246,4 +257,9 @@ See also `define-setf-handler'."
     (lambda (buffer)
       (when (window status-buffer)
         (when (eq buffer (active-buffer (window status-buffer)))
-          (print-status (window status-buffer)))))))
+          (print-status (window status-buffer))))))
+  (define-setf-handler browser buffers status-buffer
+    (lambda (browser)
+      (declare (ignore browser))
+      (loop for window in (window-list)
+            do (print-status window)))))
```
Ambrevar, line 265: Replace the `loop` with `(mapc #'print-status (window-list))`.

jmercouris:
> But what about also adding a times char to the chip? It's much more conventional of a thing, after all (:

What do you mean? What is a times char?

source/status.lisp
```diff
@@ -128,34 +128,45 @@ the URL.)"
...
-                               "plain")
-                      :buffer status
-                      :type "tab"
-                      :text domain
-                      (nyxt::switch-buffer-or-query-domain domain))))))
+  (let* ((buffers (if (display-tabs-by-last-access-p status)
+                      (sort-by-time (buffer-list))
+                      (buffer-list)))
+         (domain-deduplicated-urls (remove-duplicates
+                                    (mapcar #'url buffers) :test #'equal :key #'quri:uri-domain)))
```
aadcg, line 135: Prefer the equivalent case-sensitive counterpart of `equal` for strings: `string=`.

source/buffer.lisp
```diff
@@ -787,6 +790,7 @@ store them somewhere and `ffi-buffer-delete' them once done."))
               :flex-shrink "2"
               :flex-basis "144px")
             `("#tabs"
+              :line-height "22px"
               :background-color ,theme:secondary
               :color ,theme:on-secondary
               :min-width "100px"
```
aadcg, line 796: It might sound a bit picky, but I think having the most striking CSS rules at the top account for better readability. Background color and color make sense to be at the very top.

source/buffer.lisp
```diff
@@ -801,14 +805,18 @@ store them somewhere and `ffi-buffer-delete' them once done."))
...
+            `(".tab"
+              :margin-top "1px"
+              :margin-right "8px"
+              :background "transparent"
+              :color "inherit"
+              :text-decoration "transparent"
+              :border "transparent"
+              :border-radius "1px"
+              :font "inherit"
+              :outline "inherit")
```
aadcg, line 817: Same argument with respect to ordering CSS rules.

source/buffer.lisp
```diff
@@ -836,9 +844,13 @@ store them somewhere and `ffi-buffer-delete' them once done."))
...
-              :padding-right "6px"
+            `(.selected-tab
+              :display "inline-block"
+              :clip-path "polygon(20px 0%, 100% 0%, calc(100% - 20px) 100%, 0% 100%)"
+              :padding-left "18px"
+              :padding-right "18px"
+              :margin-right "0"
+              :margin-left "0"
               :color ,theme:on-background
               :background-color ,theme:background))))
```
aadcg, line 855: And here :p

source/buffer.lisp
```diff
@@ -733,6 +733,9 @@ store them somewhere and `ffi-buffer-delete' them once done."))
    (glyph-mode-presentation-p
     nil
     :documentation "Display the modes as a list of glyphs.")
+   (display-tabs-by-last-access-p
+    nil
+    :documentation "Show the tabs in a dynamic order by their last access time.")
```
aadcg, line 738: If you look at other examples, I usually start docstrings of predicates with "whether".

I'd suggest: "Whether tabs are dinamically ordered by last access time."

source/status.lisp
aadcg: I believe the indentation should look like this:

```lisp
(domain-deduplicated-urls (remove-duplicates (mapcar #'url buffers)
                                             :test #'string=
                                             :key #'quri:uri-domain))
```

source/status.lisp
```diff
@@ -128,34 +128,45 @@ the URL.)"
...
+    (spinneret:with-html-string
+      (loop for url in domain-deduplicated-urls
+            collect
+               (let* ((internal-buffers (remove-if-not #'internal-url-p (buffer-list) :key #'url))
+                      (presentation (if (internal-url-p url)
```
aadcg, line 140: Presentation is a bit odd isn't it? Perhaps `%domain` or anything resembling `domain`?

source/status.lisp
```diff
@@ -128,34 +128,45 @@ the URL.)"
...
+                                        "internal"
+                                        (quri:uri-domain url)))
+                      (url url)
+                      (domain (quri:uri-domain url)))
+                 (:span
+                  :class (if (equal (quri:uri-domain (url (current-buffer (window status)))) (quri:uri-domain url))
```
aadcg, line 146: `equal` -> `string=`.

Personally I think that indenting the 2 values you're comparing vertically account for better readability.

aadcg:
I agree with the changes thus far, but I don' think this PR is ready to be merged.

------

The concept of a tab bar entails that new tabs are always added to the right, and I don't think any other behavior is as intuitive. I think we should have that behavior when `display-tabs-by-last-access-p` set to `nil`. Right now, it seems that new tabs are always  added to the left (unless a URL with the same domain already exists).

-------

Another expected feature of a tab tab, is that tabs aren't shuffled around unless the user asks for it. Switching buffers moves the tabs around, which seems to contradict the fact that `display-tabs-by-last-access-p` is set to `nil`.


https://user-images.githubusercontent.com/45483512/225129046-32f115e7-dac8-4217-a69c-33b3339668e1.mp4

-------

Aesthetically, I'm not sold on the UI of each single tab. Parallelograms don't blend well with the chevron-based UI. Also, I think that a thin frame/border around the tab is really lacking. How to I distinguish between tabs Wikipedia and GitHub in the screenshot below? As a user, I need to understand that each of those looks like a button.



----

See #2849, but I don't think that's important right now.

aadcg:
@jmercouris I've left of comments from the engineering point of view (nothing very important there) and from the user point of view.

source/status.lisp
jmercouris: It is not the domain, it is what the user is presented. What about tab-display?

jmercouris:
I have fixed the engineering changes!

There are some UI changes, but those are things I would like to do in subsequent requests.

For example:

+ Thomas and I worked on a proposal for a set of lines that would appear next to the tab to indicate that there are multiple elements there. It would look like a stack of tabs.

+ We could add parallelogram lines around the non selected tab that are subtle. This would allow for a separation of them.

With regards to the tabs being shuffled around, they are never shuffled around actually. They always stay in the same position.

With regards to their order, indeed it should probably be reversed!

jmercouris:
I've added a way for all tabs to be visible. If only our theme had another color. Alas... I will have to add it ;-)

jmercouris:
Latest look:

Screenshot_20230314_193548

cc: @aadcg @lansingthomas

aadcg:
> With regards to the tabs being shuffled around, they are never shuffled around actually. They always stay in the same position.

Yes, I was completely inattentive! I still have some comments on that topic, but let's first tackle other more important issues.

@jmercouris, thanks for the changes! I think we're very very close now :)

The theme change is unrelated to this PR, so let me take your commits and push a new PR. Also, I'll add some commits to those of yours since I some ideas on about it. When that one will me merged we can get back to this one.


------

An observation.

Notice that, from the very beginning, naming this branch `tab-delete-functionality` didn't match the initial changes since you added the user option `display-tabs-by-last-access-p`. `refactor-tabs-ui` would be a good candidate perhaps.

From our documentation:

> The names of the branches really matter since the merge commit references them,
so please take that into account!

jmercouris:
Right indeed, can we rename a branch? What do you intend to change about theme work?

aadcg:
> Right indeed, can we rename a branch?

[As I expected, you have to make a new PR](https://stackoverflow.com/a/20014444/9674523).

> What do you intend to change about theme work?

Nothing of major importance. You'll see in some minutes :)

aadcg:
@jmercouris the new theme color has landed. We need to close this PR, and open a new one with a new branch name and rebase against master.

Do you want me to do it?

jmercouris:
Yes please :-)

aadcg:
@jmercouris I'm closing this PR since it has been superseded by #2855.

I'll re-read this whole thread and make a brief summary of the small bits we need to close before merging.

PR #2845, opened 2023-03-14
Verbose CUA modifiers

### LINKED ISSUE #2841: Labeling of the `shift` key
In the codebase we refer to the Shift key as `"shift"`, whereas in the documentation we refer to it as `"s"`.

Find an example below.

Help interface:

Codebase:

```lisp
"shift-left" 'backward-char-with-selection
"shift-right" 'forward-char-with-selection
"shift-up" 'backward-line-with-selection
"shift-down" 'forward-line-with-selection
```

This means that the codebase is actually more approachable than the Help interface. 

I suggest referring to the Shift key as `"shift"` everywhere.

@lansingthomas, does it make sense?

aartaka:
# Description

This addresses the fact that our notation may be intimidating to CUA/new users, by defining full names of modifiers in CUA and disabling `nkeymaps:*print-shortcut*` (can't stop being amazed at the amount of foresight @Ambrevar has) for it. This way, `M-C-S-s-x` turns into a gorgeously long `Alt-Win-Ctrl-Shift-x`, or, on Mac, `Option-Command-Ctrl-Shift-x` :D

Fixes #2841

# Discussion

I don't like the fact that we cannot use the long modifier names portably anymore (due to CUA rebinding e.g. `control` to `Ctrl` etc.). But, given that most of the engineers are Emacs/Vi people, we should probably mitigate that with a strict keybinding discipline. 

It should not be a problem for users either, because everyone sticks with just one keyscheme and thus has their own notation working perfectly.

Also note that tests fail :P I'll fix those soon-ish!

# Checklist:

  - [X] ASDF dependencies,
  - [X] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - [X] and Guix dependencies.
- [X] Documentation:
- [ ] Compilation and tests:


Ambrevar:
Regarding how we write bindings in the code base, I believe we should achieve the following, by order or priority:
1. Consistency
2. No ambiguity.

Thus switching all the "s" / "shift" to the same notation is what matters the most.
Then I'd lean towards using "shift" instead of "s" because it's less ambiguous, but it's secondary.

aartaka:
> Regarding how we write bindings in the code base, I believe we should achieve the following, by order or priority:
>     1. Consistency
>     2. No ambiguity.

Yes, that resonates with me!

> Thus switching all the "s" / "shift" to the same notation is what matters the most. 

That's why I've switched to "s" in this PR.

> Then I'd lean towards using "shift" instead of "s" because it's less ambiguous, but it's secondary.

But then, it's inconsistent with all the other modifiers, so I'm not sure.

aadcg:
[I strongly oppose this PR and I've presented my argument in the corresponding issue](https://github.com/atlas-engineer/nyxt/issues/2841#issuecomment-1467846958).

aartaka:
I've decided to go with string replacement in `binding-keys` after all, because `nkeymaps` is very strict about all the keyschemes having the same set of modifiers (which is absolutely justified).

Ready for review.

aadcg:
> Thus switching all the "s" / "shift" to the same notation is what matters the most.

The only instance that uses "s" (that I've found) is at `document.lisp`. See `scroll-page-up` in `vi-normal`.

aadcg:
I have now understood @aartaka's perspective fully.

I think that the decisions to be taken here are:

1. What's the de facto representation of the shift key that we want to use in our codebase? `"s"` or `"shift"`?
2. Do we want to support multiple aesthetical representation of keybindings in the Help system? Regarless of the answer, notice that all representations must be valid ways to specify keybindings programatically.

I don't hold any strong position, but I'd lean towards `"shift"` and no (at least right now). 

@Ambrevar and @jmercouris, I think you should have the final word :)

Ambrevar:
I'm also not super hot about  change 2. (the consistency fix is fine).

As usual, if we don't reach consensus for now, we can work this out later.

Ambrevar:
I think this PR is actually the right way to go: it does not change how keyspecs are written, which is a good thing, it only changes the printing of the keys as far as I understand.

This raises a problem however: with this  change, there is now a mismatch between how the keys are printed and how the keyspecs are written.  Which is admittedly confusing.

How to we get out of this dilemma?

jmercouris:
The decision has been made already, and Andre's arguments are sound. Let us proceed with that approach.

In any case, even if we do the specialized printing, it should be done in `nkeymaps` anyways.

jmercouris:
*If* you think we should change it, let us propose a vote. Otherwise, let us let it rest, for now. Don't worry, we'll eventually find the best way. But we need to move forward!

jmercouris:
I want to emphasize something that I think we have lost sight of. We are *always* ready to rewrite things when necessary! I think we have spent far too much time trying to get everything right in advance. We are not sooth sayers. Instead, let us not be afraid to rewrite!

Much like a book, we have to make many drafts and revisions until it is correct. We cannot anticipate all eventualities in advance.

Ambrevar:
The decision was not made with me...
Please, refrain from hasty merges.



> But we need to move forward!

Unfortunately here we've moved backwards, by making code unportable.

PR #2863, opened 2023-03-18
user-interface.lisp: Fix user interface.

shamazmazum:
Hi! Can you please check if download-mode works correctly in the most recent nyxt and, if not, apply this PR? User interface's `update` method was broken between nyxt 3-pre-release-1 and 3-pre-release-2 and I think the breakage should still be present in master. I cannot install the most recent nyxt myself because I have an old version of WebKit which is incompatible with cl-webkit2. Here is a commit message with clarification:

In 5a5f0cb a correct call to document.getElementById() was (accidentally?) replaced with document.querySelector(). Consequences of that are some broken modes, e.g. download-mode shows many JS errors and progress bars do not work anymore. This commit fixes these errors.

### DIFF
--- a/source/user-interface.lisp
+++ b/source/user-interface.lisp
@@ -10,7 +10,7 @@
    (user-interface:buffer paragraph)
    (ps:ps
      (setf (ps:chain document
-                     (query-selector (ps:lisp (user-interface:id paragraph)))
+                     (get-element-by-id (ps:lisp (user-interface:id paragraph)))
                      text-content)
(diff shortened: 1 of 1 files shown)


aadcg:
Hi @shamazmazum. With respect to commit 5a5f0cb, I think you're correct. It seems that I did a mistake there.

Regarding `download-mode` could someone please try it? I have never used this feature, so I'm not sure how it used to work.

jmercouris:
I just tested it, works great, thank you :-)

PR #2868, opened 2023-03-21
Use filter-exact-matches in much more places.

### LINKED ISSUE #2792: `fuzzy-match` behavior when input doesn't match any suggestions
It's reasonable to expect that a filtering mechanism returns a proper subset of suggestions when matching against input. In some cases that subset might even be the empty set.

However, `fuzzy-match` return all suggestions in some situations.

Example:

- Issue `execute-command`; 
- Run `(length (prompter:suggestions (current-source)))`; 
- Set the input to something bogus, say "sklfjslkfjslfjsk"; 
- Run `(length (prompter:suggestions (current-source)))`

Both calls to `(length (prompter:suggestions (current-source)))` return the same value. 

I find this behavior extremely strange. 

CC @lansingthomas @jmercouris @Ambrevar @aartaka.

aartaka:
This makes most of our sources more exact in matching the input. This is important for sources with URLs, medium-to-huge quantities of text, or many-attributes sources. 

Please, someone, just look at this change and accept that our sources are more exact and useful now :)

Closes #2792.

# Discussion

- Should I add test for that?
- Also note that `nyxt/gi-gtk` tests fail even on master :P

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [X] Documentation:
- [x] Compilation and tests:


source/changelog.lisp
```diff
@@ -717,7 +717,8 @@ like buttons.")
    (:li "The REPL features " (:nxref :class-name 'nyxt/repl-mode:shell-cell "shell cells")
         ", which pass commands to the underlying system shell.")
    (:li "The prompt buffer tab interface has been updated. Tabs are stably sorted and do
-not shuffle around. It is now possible to middle click to delete tabs."))
+not shuffle around. It is now possible to middle click to delete tabs.")
+   (:li "Many prompts became more exact."))
```
Ambrevar, line 721: It's not very clear what you mean.

source/recent-buffers.lisp
```diff
@@ -28,7 +28,8 @@
        (mapc #'reopen-dead-buffer buffer-list)
        (set-current-buffer (or (first (prompter:marks (current-source)))
                                (current-suggestion-value (current-prompt-buffer)))))
-     (lambda-mapped-command reopen-dead-buffer)))))
+     (lambda-mapped-command reopen-dead-buffer)))
+   (prompter:filter-preprocessor #'prompter:filter-exact-matches)))
```
Ambrevar, line 32: Why for buffers?

source/mode/history.lisp
```diff
@@ -279,6 +284,7 @@ Otherwise go forward to the only child."
 (define-class history-all-source (prompter:source)
   ((prompter:name "All history URLs")
    (buffer :initarg :buffer :accessor buffer :initform nil)
+   (prompter:filter-preprocessor #'prompter:filter-exact-matches)
```
Ambrevar, line 287: But it's not just URLs, it's also titles, so this may be too restrictive.
Same for all history sources.

source/recent-buffers.lisp
aartaka: Because buffers have URLs in them, and because `buffer-source` is used in `set-url` to match against buffers. In the latter case, it's extremely important to filter aggressively, so that the source disappears the very moment it becomes irrelevant.

source/changelog.lisp
aartaka: Reworded to "Many prompts are filtered more aggressively, resulting in more exact suggestions."

aadcg:
I'm taking a look at this later today or tomorrow. Thanks!

aadcg:
I believe @Ambrevar already did a good review. 

This seems to address issue #2792, and probably closes it too. I wonder if the default for `prompter:filter-preprocessor` shouldn't be changed, but I didn't think about it enough.

Feel free to proceed as you find most adequate!

source/mode/history.lisp
Ambrevar: Yup, but fuzzy-matching on titles is desirable.

aartaka:
> This seems to address issue #2792, and probably closes it too. 

Yes, I just was unable to find the issue xD

> I wonder if the default for `prompter:filter-preprocessor` shouldn't be changed, but I didn't think about it enough.

I lean to that too, because `filter-exact-matches` is probably the most intuitive matching algo we have. But I'll leave that for @Ambrevar to decide, given that `libraries/prompter` is to be someday abstracted out, and that will prompt (pun not intended) for all the necessary architectural redesigns.

source/mode/history.lisp
aartaka: I'd still say that most of people search things by space-separated chunks of text, so fuzzy matching that works backwards or matches every single letter is not as useful as matching clearly delineated word-like sequences.

PR #2870, opened 2023-03-22
Add a convenience `define-generic` macro.

aartaka:
# Description

This adds a `define-generic` macro, that targets a frequent use-case we have for `defgeneric`: one generic, one method. Using `defmethod` for these cases is quite tempting, because it's really simple to write a straigh-to-the-point
``` lisp
(defmethod name ((some specialized) args &key and more &allow-other-keys)
  "Method docs."
  (do-something))
```
instead of a noisy
``` lisp
(defgeneric name (some args &key &allow-other-keys)
  (:method ((some specialized) args &key and more &allow-other-keys)
    "Method docs."
    (do-something))
  (:documentation "Generic docs."))
```

------------------------------------------------------------------------------------------

`define-generic` relieves the `defgeneric` fatigue by optimizing the layout of things to resemble `defun`/`defmethod` and to smartly parse the arglist and options that `defgeneric` has. With it, the definition is exactly like `defmethod`
``` lisp
(define-generic name ((some specialized) args &key and more &allow-other-keys)
  "Generic docs."
  (do-something))
```
and it expands to
``` lisp
(defgeneric name (some args &key &allow-other-keys)
  (:method ((some specialized) args &key and more &allow-other-keys)
    (do-something))
  (:documentation "Generic docs."))
```

Neat, huh?

# Discussion

- I'm pretty sure that this macro works as intended, as I've tested it in lots of cases. Additional testing is welcome, though. 
- And let me know if this syntax makes sense to you!
- Does it maybe belong to `nclasses`? Any other convenience library we can contribute it to, maybe?

I'll propagate this syntax everywhere in Nyxt once we decide `define-generic` is a good shot at generics brevity.

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:


aartaka:
> I have doubts about this one. The whole point of using `defgeneric` is to specialize against several types/classes.

No, not necessarily. In our case, the main value of `defgeneric` (IIRC) is that it is better for the inspector, as it saves a definite source location for the function.

> ```lisp
> (defgeneric name (arg-list)
>   (:method ((instance class-foo)) ...)
>   (:method ((instance class-bar)) ...)
>   ...)
> ```

`define-generic` support this syntax too. I've gone a great length to make it fully (not sure, but was working in all the cases I could think of) backwards-compatible with `defgeneric` :)

More so: even this form can be enhanced with leading docstring, as in
``` lisp
(nyxt/utilities:define-generic generic-name (value &optional compact-p)
  "Some docstring"
  (:method (value &optional compact-p)
    (declare (ignore compact-p))
    (princ value))
  (:method ((value null) &optional compact-p)
    (declare (ignore compact-p))
    (princ value))
  (:method ((value string) &optional compact-p)
    (declare (ignore compact-p))
    (princ value)))
```

> There may be a single method but, even in that case, seeing it defined via `defgeneric` seems preferable.

The thing quite often is that there's only one default method. In Nyxt code, there are twenty such instances our of forty-ish `defgeneric`s.

Ambrevar:
It's a small improvement, but it's frequent enough that it makes sense to have it.

As part of Nclasses?  I don't know.  Generic functions are some abstract MOP objects, but not exactly _classes_, so it's questionable it belongs there, but that wouldn't be too far fetched either.

Hmmm... I'd vote for yes!

source/utilities.lisp
```diff
@@ -51,6 +51,95 @@ any expensive computation only takes place once."
...
+   (if rotate
+       (uiop/utility:strcat (princ-to-string b) a)
+       (uiop/utility:strcat a (princ-to-string b))))
+  (:method ((a integer) (b string) &key rotate)
+   (if rotate
+       (uiop/utility:strcat b (princ-to-string a))
+       (uiop/utility:strcat (princ-to-string a) b)))
+  ;; :method-combination is preserved.
+  (:method-combination progn)
+  (:documentation \"Concatenate integers to strings.\"))"
```
Ambrevar, line 126: Maybe move some of these examples to tests?  The docstring is a bit too long anyways :)

aartaka:
> As part of Nclasses? I don't know. Generic functions are some abstract MOP objects, but not exactly _classes_, so it's questionable it belongs there, but that wouldn't be too far fetched either.

My reasoning was that `nclasses` is about syntax sugar, so `define-generic`, as an intuitive enough of a syntax sugar belongs there.

> Hmmm... I'd vote for yes!

You mean moving it to `nclasses`? I'll open a PR there then.

source/utilities.lisp
aartaka: Two reasons to retain these as they are:
- `define-generic` has quite a set of behaviors, so it's important to showcase them (but I guess that having one long example is better than two noisy ones).
- I feel like testing macroexpansions may be dangerous and implementation-dependent. Haven't tried it, though.

Ambrevar:
> > 
> 
> You mean moving it to `nclasses`? I'll open a PR there then.

Yes, thanks.

aartaka:
Closing in favor of https://github.com/atlas-engineer/nclasses/pull/11

PR #2881, opened 2023-03-29
Add tree-style hinting

### LINKED ISSUE #2509: Incremental tree-style hinting
In Emacs, there's a couple of packages such as [avy](https://github.com/abo-abo/avy) where you can search for text "incrementally" in a character-by-character decision tree. I would love a similar feature for the hinting system (in addition to general search, too). That way, I could more easily select which hints I actually want, and faster. Also, it could help out in the cases where hints are too close together and it's hard to tell what sequence of letters they are.

heiwiper:
# Description

The hint-source pre-processor filter is modified to add a function that toggles the visibility of hint elements on and off according to whether they are matched by the input.

Fixes #2509

# Discussion

- I am not sure whether adding the super class's pre-processor filter function as a `funcall` is the proper way of adding more functionality on top of the already existing one.
- I have noticed there's a file `hint-prompt-buffer.lisp` which contains some functions related to hinting, so I don't know if the new added functions `show-hint-element` and `hide-hint-element` are supposed to be there.
- This change should would change is directly visible by end users, so there might a need to mention it in the `changelogs.lisp`, but I did not know where to put. 
- I still haven't finished reading the style guidelines for Common Lisp (I'm currently doing that) so I can't check that item in the checklist, but I try to follow the already existing code as much as possible.

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [ ] Documentation:
- [x] Compilation and tests:


aadcg:
@heiwiper thanks for you work! I'll get back to you in a couple of days :)

aadcg:
In the mean time, notice that the tests are failing.

heiwiper:
> In the mean time, notice that the tests are failing.

It should be all good now! sorry for the inconvenience, there was a conflict when using `visibility` as an argument of `ps:@` so I passed a string instead.

Also please note that this is the very basic implementation of tree-style hinting, I wouldn't mind adding more functionality on top of it, maybe something like qutebrowser where the input character is highlighted either in bold or different font color.

aartaka:
> It should be all good now! sorry for the inconvenience, there was a conflict when using `visibility` as an argument of `ps:@` so I passed a string instead.

Note that this doesn't always work with Parenscript and using uninterned symbols `#:visibility` is more reliable, especially when chaining properties/methods.

> Also please note that this is the very basic implementation of tree-style hinting, I wouldn't mind adding more functionality on top of it, maybe something like qutebrowser where the input character is highlighted either in bold or different font color.

I'd rather merge this immediate improvement PR and have the character highlighting in a separate PR (because I've a hunch that this may be less intuitive than expected).

heiwiper:
> > 
> 
> Note that this doesn't always work with Parenscript and using uninterned symbols `#:visibility` is more reliable, especially when chaining properties/methods.

Should I update my PR to use the method you recommended ? I actually followed [this code](https://github.com/atlas-engineer/nyxt/blob/07da7614a07cba612f1d62b6c17d5646aacb2811/source/parenscript-macro.lisp#L139) that's why I used a string in the first place.

source/mode/hint.lisp
```diff
@@ -220,9 +220,34 @@ For instance, to include images:
...
+      (setf (ps:@ %element style "visibility") "hidden"))))
+
+;; Toggle on the visibility of the hint element that corresponds to the given `hint`
+(define-parenscript-async show-hint-element (hint)
+  (let ((%element (nyxt/ps:qs document (ps:lisp (format nil "#nyxt-hint-~a"
+                                                        (identifier hint))))))
+    (when %element
+      (setf (ps:@ %element style "visibility") "visible"))))
+
 (define-class hint-source (prompter:source)
```
aadcg, line 237: I'd perhaps suggest having a single method here. Along the lines of:

```lisp
(define-parenscript-async set-hint-visibility (hint state)
  "Set visibility STATE of HINT element.

Consult https://developer.mozilla.org/en-US/docs/Web/CSS/visibility."
  (let ((elem (nyxt/ps:qs document (ps:lisp (format nil "#nyxt-hint-~a"
                                                    (identifier hint))))))
    (when elem (setf (ps:@ elem style "visibility") (ps:lisp state)))))
```

aadcg: Notice that naming variables `%foo` in PS isn't a very good idea since:

```lisp
NYXT> (ps:ps (let ((%foo "foo"))))
"(function () {
    var percentfoo = 'foo';
    return null;
})();"
```

source/mode/hint.lisp
```diff
@@ -220,9 +220,34 @@ For instance, to include images:
...
+    (lambda (suggestions source input)
+      (when (and (auto-follow-hints-p (find-submode 'hint-mode))
+                 (fit-to-prompt-p (find-submode 'hint-mode)))
+        (loop for suggestion in suggestions
+              do (if (str:starts-with-p input
+                                        (prompter:attributes-default suggestion)
+                                        :ignore-case t)
+                     (show-hint-element (prompter:value suggestion))
+                     (hide-hint-element (prompter:value suggestion)))))
+      (funcall #'prompter:delete-inexact-matches suggestions source input)))
```
aadcg, line 250: Do we really need this? I think we can simply return `suggestions`.

```diff
 (define-class hint-source (prompter:source)
   ((prompter:name "Hints")
    (prompter:actions-on-current-suggestion-enabled-p t)
    (prompter:filter-preprocessor
     (lambda (suggestions source input)
+      (declare (ignore source))
       (when (and (auto-follow-hints-p (find-submode 'hint-mode))
                  (fit-to-prompt-p (find-submode 'hint-mode)))
         (loop for suggestion in suggestions
@@ -247,7 +242,7 @@ define-class hint-source
                                         :ignore-case t)
                      (show-hint-element (prompter:value suggestion))
                      (hide-hint-element (prompter:value suggestion)))))
-      (funcall #'prompter:delete-inexact-matches suggestions source input)))
+      suggestions))
    (prompter:filter
     (if (and (auto-follow-hints-p (find-submode 'hint-mode))
              (fit-to-prompt-p (find-submode 'hint-mode)))
```

aadcg:
@heiwiper, functionality-wise, this is a good improvement! Thanks for your work.

> Also please note that this is the very basic implementation of tree-style hinting, I wouldn't mind adding more functionality on top of it, maybe something like qutebrowser where the input character is highlighted either in bold or different font color.

Let's follow @aartaka's advice and tackle that in a separate PR. In my view, the best UI for that feature would make the characters that have been typed thus far more transparent.

> Should I update my PR to use the method you recommended ? I actually followed this code that's why I used a string in the first place.

Personally, I think it's ok the way you did it :)

----

I left some minor comments, thanks!

source/mode/hint.lisp
heiwiper: The reason I added it is because it is [used](https://github.com/atlas-engineer/nyxt/blob/25bf3a481b07b43c24eb1dcd76fd4c0d56699c5a/libraries/prompter/prompter-source.lisp#L169) in the `filter-preprocessor` in `prompter:source` class, and since `hint-source` class inherits `prompter:source` I thought I would call `delete-inexact-matches` to keep the same behaviour of the super function just like how `fuzzy-match` function is called in `filter` function of `hint-source` class [here](https://github.com/atlas-engineer/nyxt/blob/25bf3a481b07b43c24eb1dcd76fd4c0d56699c5a/source/mode/hint.lisp#L234).

aadcg:
> This change should would change is directly visible by end users, so there might a need to mention it in the changelogs.lisp, but I did not know where to put.

@heiwiper just append to the end, under the `3-pre-release-6`. Does that make sense?

heiwiper:
I have just pushed the corrections, also I'm glad this will be part of `3-pre-release-6`.

>      Let's follow @aartaka's advice and tackle that in a separate PR. In my view, the best UI for that feature would make the characters that have been typed thus far more transparent.

I will try to make another pull request today concerning the hint UI as we have discussed last.

aadcg:
@heiwiper merged via 717ba6ebb. Thank you for your work!

> I will try to make another pull request today concerning the hint UI as we have discussed last.

Looking forward :)

PR #2882, opened 2023-03-30
buffer.lisp: Fix on-signal-load-canceled method

shamazmazum:
`on-signal-load-canceled` method for buffers calls `on-signal-load-redirected` for modes. Fix that.


Ambrevar:
CC @aartaka

aartaka:
Right, that's an oversight on my side. Thanks for fixing it @shamazmazum, merged with d2f94b39bov.

PR #2883, opened 2023-03-31
global: Fix compilation notes.

Ambrevar:
# Description

This was due to +version+ being a constant, and SBCL being smart enough to resolve the `if' condition at compile time, thus complaining that a branch had been elided.

# Checklist:

- [x] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,7 +67,7 @@ jobs:
     - name: Fetch Common Lisp third-party dependencies
       shell: bash
       run: |
-        ros ${{ matrix.rosargs }}  -e '(handler-bind ((error (lambda (a) (uiop:print-backtrace) (format *error-output* "Error: ~a~&" a) (uiop:quit 17)))) (asdf:load-system :nyxt/submodules))'
+        ros ${{ matrix.rosargs }}  -e '(handler-bind ((error (lambda (a) (princ a) (uiop:print-backtrace a) (uiop:quit 17)))) (asdf:load-system :nyxt/submodules))'
         mkdir -p ~/.config/common-lisp/source-registry.conf.d/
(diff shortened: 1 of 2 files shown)


aadcg:
@Ambrevar thanks for this fix! Notice that ccl doesn't seem to like this change... :(

aartaka:
Note that SBCL emits `sb-ext:code-deletion-note` on code deletion. If we can somehow handle this, we'll have no need for read-time hacks.

Ambrevar:
I wouldn't call this a hack, since it makes sense for the body of `version` to be defined at read-time.

However CCL does not know the value of the `+version+` constant.  We can wrap it in `eval-always`.

@aartaka Any idea how to catch compilation errors with SBCL?

aartaka:
> I wouldn't call this a hack, since it makes sense for the body of `version` to be defined at read-time.

Yes, that does make sense.

> However CCL does not know the value of the `+version+` constant. We can wrap it in `eval-always`.

Yes!

> @aartaka Any idea how to catch compilation errors with SBCL?

No idea except rebinding the debugger hook, which is wrong :(

Ambrevar:
CCL fails but I can't see why.  A read error for `(defun version ...)`.  Maybe a compiler bug?

Unless someone can find a good workaround, I suggest to drop this pull request.

Ambrevar:
> CCL fails but I can't see why. A read error for `(defun version ...)`. Maybe a compiler bug?

I should have noted that I cannot reproduce this locally, which is even more puzzling...

aartaka:
I've tried to make the backtraces and error printing more consistent, but it's still imperfect and doesn't reveal much more of our error...

jmercouris:
Please reopen when ready.

PR #2916, opened 2023-04-19
[WIP] Improve semantics of `hint-mode`

### LINKED ISSUE #2865: Improve semantics of `hint-mode`'s user options
We have, in very general terms, two de-facto ways to use `hint-mode`:

1. Use the prompt buffer and match against the hint strings, and other attributes of the hints (the default);
2. Only matches against the hint strings, collapses the prompt buffer and returns as soon as a matching hint string has been typed.

There are roughly 3 user options that enable the 2nd mode of operation above:

1. `nyxt/hint-mode:fit-to-prompt-p`
2.  `nyxt/hint-mode:auto-follow-hints-p`
3. `nyxt/hint-mode:compute-hints-in-view-port-p`

For the 1st option, it must be set to `t` for mode 2 and to `nil` on mode 1.

The 2nd option must be set to `nil` on mode 1 and, arguably, could be set to any boolean on mode 2 (but it would make much more sense to have set to `t`).

The 3rd option, arguably, could be set to any boolean in mode 1 (while it makes much more sense to set it to `nil`). On the other hand, for mode 2 it only makes sense to set to `t` (for performance).

In short, users need to set 3 options to enable mode 2. Also, not all combinations make sense. I think we must do better.

I'm concealing my personal view on a possible solution, so that I don't influence anyone's view.

It'd be great to get some feedback from our users, such as @heiwiper!

heiwiper:
# Description

Replace the following `hint-mode` slots with new slot `hinting-type`:
- `compute-hints-in-view-port-p`
- `auto-follow-hints-p`
- `fit-to-prompt-p`

Fixes #2865

# Discussion

This is a draft PR that attempts to address the issue following the suggestion mentioned by @aartaka .
I tried to remove the deprecated slots mentioned above but that was causing an issue in `changelog.lisp`.
What should I do for the users who have the deprecated slots set in their config files ?
I still haven't added changelogs since this is not yet approved.

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [x] Documentation:
  - (If this changes something about the features showcased on Nyxt website) I have these changes described in the new/existing article at Nyxt website or will notify one of maintainers to do so.
- [X] Compilation and tests:


aartaka:
> I tried to remove the deprecated slots mentioned above but that was causing an issue in `changelog.lisp`. 

Replace all the `:nxref`s mentioning the removed slots with `(:code "slot-name")` and it will compile properly. `:nxref` is a link to the symbol that ensures it exists, thus the errors.

> What should I do for the users who have the deprecated slots set in their config files? 

Add a migration.lisp entry. That's the best we could come up with so far (:

> I still haven't added changelogs since this is not yet approved.

Do add them, because if we merge it then we won't need to alter changelog about it :)

heiwiper:
> Add a migration.lisp entry. That's the best we could come up with so far (:
I have added a migration entry but I'm not sure how to test it. I tried to set the deprecated slot in my config and then run the new nyxt version where that configuration slot doesn't exist and that apparently caused the new slot to be ignored ?

I have also added a new changelog entry `3-pre-release-7` since the previous version `3-pre-release-6` was released yesterday so I thought I can no longer add notes there.

aadcg:
@heiwiper another great PR! I've merged your work with some minor corrections via commit 53f98d80b. 

Thanks again :)

heiwiper:
Thank you! I'm glad you were satisfied with those changes I will try to fix the shadow DOM issue next.

aadcg:
Looking forward, thanks again for your work!

PR #2942, opened 2023-05-01
Add and change prompt mouse/return actions for consistency and intuitiveness

### LINKED ISSUE #2628: Review all keybindings before releasing `3.0`
Adding it as a reminder.

@aartaka [advocated for rebinding `set-action-on-return` (ex `return-selection-over-actions`)](https://github.com/atlas-engineer/nyxt/pull/2555#issuecomment-1300309695).

> `set-action-on-return` (ex `return-selection-over-action`) should be bound to `shift-return` instead of `M-return`. The latter should be reserved for "strong" actions that close the prompt-buffer, analogous to `return` and `C-return` (return stands either for "activate this now" or "I'm done, commit this action" in CUA). If we dilute the responsibility one takes when pressing `M-return`, then the character of other actions bound to *-return will not be perceived as final and decisive for a given prompt.

### LINKED ISSUE #2931: match some keybindings with some mouse inputs (marking system)
I notice some inconsistency, I understand that it matches emacs environment. I am not sure that we need it to be this way in Nyxt.
Currently:
`click` invokes a suggestion
`Ctrl + click` marks an unmarked suggestion OR unmarks a marked suggestion. 

`return` invokes a suggestion
`Ctrl + spacebar` does the same 

It would be more consistent to have `Ctrl + return` to mark or unmark suggestions.
I am happy with `Ctrl+space` doing the same, though I would like to add `Ctrl+return` as the keybindings for this functionality - at least in CUA mode.

IF anyone has a better suggestion please let me know and I will fight for it.

thank you very much mighty Nyxt devs

aartaka:
# Description

This makes two changes:
- Adding more mouse actions, bridging the functionality mismatch between mouse use and keyboard use:
  - `C-button1` for marking.
  - `s-button1` for marking.
  - `M-button1` for return with another action (= `M-return`).
- Bringing the keyboard actions to consistency with these:
  - `C-return` for marking.
  - `s-return` for marking.

Fixes #2931
Also should resolve #2628 once we decide on what to do with the action on current suggestion.

# Discussion

@aadcg, we previously had this discussion about whether `run-action-on-current-suggestion` should be bound to `C-return`. Now that it stands for marking, it's resolved. But then, how does one run actions on current suggestion? Seems like it's not invokable with either mouse nor keyboard. Should it be `s-return`/`s-click`?

CC @lansingthomas 

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/mode/prompt-buffer.lisp
+++ b/source/mode/prompt-buffer.lisp
@@ -52,8 +52,10 @@ listed and chosen from with the command `set-action-on-return' (bound to
        "keypadenter" 'run-action-on-return
        "M-return" 'set-action-on-return
        "M-keypadenter" 'set-action-on-return
-       "C-return" 'run-action-on-current-suggestion
-       "C-keypadenter" 'run-action-on-current-suggestion
+       "C-return" 'toggle-mark-forwards
+       "C-keypadenter" 'toggle-mark-forwards
+       "s-return" 'toggle-mark-forwards
+       "s-keypadenter" 'toggle-mark-forwards
        "C-c C-j" 'set-action-on-current-suggestion
        "tab" 'insert-current-suggestion
        ; TODO: This is the Emacs Helm binding.  Better?
--- a/source/mode/prompt-buffer.lisp
+++ b/source/mode/prompt-buffer.lisp
@@ -56,6 +56,7 @@ listed and chosen from with the command `set-action-on-return' (bound to
        "C-keypadenter" 'toggle-mark-forwards
        "s-return" 'toggle-mark-forwards
        "s-keypadenter" 'toggle-mark-forwards
+       "C-j" 'run-action-on-current-suggestion
        "C-c C-j" 'set-action-on-current-suggestion
        "tab" 'insert-current-suggestion
        ; TODO: This is the Emacs Helm binding.  Better?
--- a/source/prompt-buffer.lisp
+++ b/source/prompt-buffer.lisp
@@ -431,26 +431,39 @@ an integer."))
                                    "marked")
                           :onclick (when (mouse-support-p prompt-buffer)
                                      (ps:ps
-                                       (if (ps:chain window event ctrl-key)
-                                           (nyxt/ps:lisp-eval
-                                            (:title "mark-this-suggestion"
-                                             :buffer prompt-buffer)
-                                            (prompter::set-current-suggestion
-                                             prompt-buffer
-                                             (- suggestion-index cursor-index))
-                                            (prompter:toggle-mark prompt-buffer)
-                                            (prompter::set-current-suggestion
-                                             prompt-buffer
-                                             (- cursor-index suggestion-index))
-                                            (prompt-render-suggestions prompt-buffer))
-                                           (nyxt/ps:lisp-eval
-                                            (:title "return-this-suggestion"
-                                             :buffer prompt-buffer)
-                                            (prompter::set-current-suggestion
-                                             prompt-buffer
-                                             (- suggestion-index cursor-index))
-                                            (prompter:run-action-on-return
-                                             (nyxt::current-prompt-buffer))))))
+                                       (cond
+                                         ((or (ps:chain window event ctrl-key)
+                                              (ps:chain window event shift-key))
+                                          (nyxt/ps:lisp-eval
+                                           (:title "mark-this-suggestion"
+                                            :buffer prompt-buffer)
+                                           (prompter::set-current-suggestion
+                                            prompt-buffer
+                                            (- suggestion-index cursor-index))
+                                           (prompter:toggle-mark prompt-buffer)
+                                           (prompter::set-current-suggestion
+                                            prompt-buffer
+                                            (- cursor-index suggestion-index))
+                                           (prompt-render-suggestions prompt-buffer)))
+                                         ((ps:chain window event alt-key)
+                                          (nyxt/ps:lisp-eval
+                                           (:title "return-this-suggestion-with-another-action"
+                                            :buffer prompt-buffer)
+                                           (prompter::set-current-suggestion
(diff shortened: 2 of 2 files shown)


source/mode/prompt-buffer.lisp
```diff
@@ -52,7 +52,8 @@ listed and chosen from with the command `set-action-on-return' (bound to
        "keypadenter" 'run-action-on-return
        "M-return" 'set-action-on-return
        "M-keypadenter" 'set-action-on-return
-       "C-return" 'run-action-on-current-suggestion
+       "C-return" 'toggle-mark-forwards
+       "s-return" 'toggle-mark-forwards
        "C-keypadenter" 'run-action-on-current-suggestion
```
jmercouris, line 57: what about this enter key? should this one also do toggle-mark-forwards?

aartaka: Riiiight! Added it :)

source/mode/prompt-buffer.lisp
```diff
@@ -52,8 +52,10 @@ listed and chosen from with the command `set-action-on-return' (bound to
        "keypadenter" 'run-action-on-return
        "M-return" 'set-action-on-return
        "M-keypadenter" 'set-action-on-return
-       "C-return" 'run-action-on-current-suggestion
-       "C-keypadenter" 'run-action-on-current-suggestion
```
aadcg, line 56: With this change CUA and VI keyscheme users won't have a keybinding bound to `run-action-on-current-suggestion`. 

What about setting it to `C-j` in the default keyscheme?

aartaka: Yes, `C-j` it is then! Pushed as 76e68b3d9cbbf0257968102aa842c169f77c8d58.

source/prompt-buffer.lisp
```diff
@@ -52,8 +52,12 @@ some point.")
      (mouse-support-p
       t
       :type boolean
-      :documentation "Whether to allow mouse events to set and return the
-current suggestion in the prompt buffer.")
+      :documentation "Whether to allow mouse events to act on prompt buffer suggestions.
+Current actions are:
```
aadcg, line 56: Now that I take a look at this for the second time, I've realized that "action" might not be the best word (to avoid confusion with prompt buffer actions).

I believe we could paraphrase as "The following mouse keybindings are available:"

source/prompt-buffer.lisp
```diff
@@ -52,8 +52,12 @@ some point.")
...
+- button1: return the suggestion.
+- C-button1: mark the suggestion.
+- s-button1: mark the suggestion.
+- M-button1: return with one of alternative actions.")
```
aadcg, line 60: I think we should mention "return actions" here.

aartaka:
@aadcg, mostly addressed! Take a second look, just in case ;)

source/prompt-buffer.lisp
```diff
@@ -52,8 +52,13 @@ some point.")
...
       :type boolean
-      :documentation "Whether to allow mouse events to set and return the
-current suggestion in the prompt buffer.")
+      :documentation "Whether to allow mouse events to act on prompt buffer suggestions.
+The following mouse keybindings are available:
+- button1: return the suggestion.
+- C-button1: mark the suggestion.
+- s-button1: mark the suggestion.
+- M-button1: return with one of alternative actions (prompt-finishing code
+  fulfilling a specific purpose).")
```
aadcg, line 61: Not terribly important, but I've suggested earlier I think it's important to say "return actions". "alternative actions" is very confusing for users. And "prompt-finishing code fulfilling a specific purpose" is confusing too.

Actually, what I'd so here would be to link these behaviors to the equivalent commands.

```text
The following mouse keybindings are available:
- button1: `run-action-on-return'
- C-button1: `toggle-mark-forwards'
- s-button1: `toggle-mark-forwards'
- M-button1: `set-action-on-return').")
```

Now that I think of it, why can't we bind these mouse keybindings at `source/mode/prompt-buffer.lisp`?

source/mode/prompt-buffer.lisp
aadcg: I've only seen your question now by the way! Sorry, that I didn't notice it before. Anyway, we made a decision on this subject.

> @aadcg, we previously had this discussion about whether run-action-on-current-suggestion should be bound to C-return. Now that it stands for marking, it's resolved. But then, how does one run actions on current suggestion? Seems like it's not invokable with either mouse nor keyboard. Should it be s-return/s-click?

source/prompt-buffer.lisp
aartaka: We cannot, because we need to act on certain suggestions, which are not necessarily the ones that are marked/selected. So it has to be JavaScript, at least to select the necessary suggestions :(

aartaka: > Actually, what I'd so here would be to link these behaviors to the equivalent commands.
> 
> ```
> The following mouse keybindings are available:
> - button1: `run-action-on-return'
> - C-button1: `toggle-mark-forwards'
> - s-button1: `toggle-mark-forwards'
> - M-button1: `set-action-on-return').")
> ```

Pushed as b8306f9acd41937efa72d93f8139396bde8fb9be!

aadcg: > We cannot, because we need to act on certain suggestions, which are not necessarily the ones that are marked/selected. So it has to be JavaScript, at least to select the necessary suggestions :(

Of course! What a silly suggestion of mine.

aadcg:
Looks good to me! Feel free to merge it :)

Issue #2622, opened 2022-11-06
Support XBEL bookmarks


aartaka:
**Is your feature request related to a problem? Please describe.**
There's a format for bookmarks called [XBEL](https://en.wikipedia.org/wiki/XBEL). There are browsers and app that use it. Moving from these browsers/apps to Nyxt would be hard if Nyxt doesn't have XBEL support.

**Describe the solution you'd like**
Much like how we support HTML bookmarks, we can create two functions:
- `import-bookmarks-from-xbel`
- `export-bookmarks-to-xbel`

And we should be good :)

**Describe alternatives you've considered**
Using some shell utility to reformat those? Naaaaaah.

**Additional context**
XBEL is XML, which is not cool. But, well, most of the world isn't s-exprs and we have to sustain through this until Lisp world dominance ヾ(@゜∇゜@)ノ

PR #2945, opened 2023-05-03
Supporf XBEL bookmarks #2622

msam010101:
# Description

Basic support for importing XBEL.

Fixes # (issue)
#2622

# Discussion

 Bookmarks are imported 'flat', ignoring the folders in the XBEL file. File assumed valid with no attempt so far to handle edge cases.

# Checklist:

(Awaiting first code review)


source/mode/bookmark.lisp
```diff
@@ -195,7 +195,7 @@ In particular, we ignore the protocol (e.g. HTTP or HTTPS does not matter)."
                  (let ((url-href (render-url (url bookmark))))
                    (:div
                     (:p (title bookmark))
-                    (:p (:a :href url-href :target "_blank" url-href))))))
+                    (:p (:a :href url-href url-href))))))
```
aartaka, line 198: Why remove this one? It's used in this panel so that the link gets opened in the main buffer. Otherwise it's opened in the panel buffer.

Although making a new buffer to open a bookmark is not cool, I guess.

Still, it doesn't belong to this PR :)

source/mode/bookmark.lisp
```diff
@@ -429,6 +429,84 @@ background buffers."
...
+;; fixes issue #2622
+
+(defclass import-xbel-seed ()
+  (
+    (current-bookmark-href :initarg :current-bookmark-href :initform nil)
+    (bookmarks :initarg :bookmarks :initform '())
+  )
+)
+
+;; For some reason reading the attributes works ONLY with :test #'string-equal
```
aartaka, line 445: This one is because the parser that cl-prevalence uses parses XML to Lisp symbols. `s-xml` (at least in this use-case) seems to parse to strings, and thus a different, string-aware comparison function is needed.

source/mode/bookmark.lisp
```diff
@@ -429,6 +429,84 @@ background buffers."
...
+;; even though this very function, get-attribute-value, is defined in 
+;; ./_build/cl-prevalence/src/serialization/xml.lisp using :test #'eq
+(defun get-attribute-value (name attributes)
+  (cdr (assoc name attributes :test #'string-equal)))
+  
+;; XBELs have a <title>Some text</title> either within a <folder> element
```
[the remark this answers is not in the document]
aartaka: I mean, can you descend into `<folder>` to extract bookmarks from there?

source/mode/bookmark.lisp
```diff
@@ -429,6 +429,84 @@ background buffers."
...
+                (let ((uri (quri:uri (car bk)))
+                      (title (cdr bk))
+                     )
+                    (bookmark-add uri
+                              :title title)))
+          )
+        )
+      (echo "The file doesn't exist or is not an XBEL file.")))      
+
+;; end support importing bookmarks from XBEL
```
aartaka, line 509: No need for this comment and for the opening one either—it's mostly clear from the functions that belong to it.

If there's documentation for the functions, then most comments are unnecessary :)

source/mode/bookmark.lisp
aartaka: Good, all clear!

source/mode/bookmark.lisp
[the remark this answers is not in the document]
msam010101: The folder structure is simple, `<folder><title>Folder name</title</folder>` ,  and a folder can contain an arbitrary number of bookmarks or other folders. I considered adding the complete hierarchy from the source XML but 

(1) The Nyxt bookmark API , `bookmark-add` , doesn't accept a container for the new bookmark, so I assumed Nyxt doesn't have folders for bookmarks (and uses tags instead).
(2) The function `import-bookmarks-from-html` , upon which I based my own function, just grabs all `<a>` tags and adds them as a flat list, which seems to support my hypothesis.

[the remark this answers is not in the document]
aartaka: Good, that makes sense. If you manage to actually set folder title as a tag, that would be extra cool. But no pressure :)

aartaka:
@msam010101, what's the status of this PR?

aartaka:
@msam010101, friendly ping :)

PR #2946, opened 2023-05-03
mode/bookmark.lisp: list-bookmark shows initially closed sections (WIP)

shamazmazum:
Hi! I've noticed that `list-bookmarks` command shows bookmarks with initially expanded sections. This way it's very difficult to find the bookmark I actually need. I usually remember in what section the bookmark is, but cannot remember its name. Therefore I need to manually close each section with exception of one and search there.

IMHO, it's better for sections to be initially closed. On the other hand, it's hard to do `C-s s` this way. Is it a rationale for why sections are initially expanded? Maybe add a button which expands / collapses all sections? What do you think?

Also, I think `request-id` is not needed here. SBCL says the same when `list-bookmarks` is executed:
``` lisp
; in: LAMBDA ()
;     (SPINNERET:WITH-HTML-STRING
;       (:NSTYLE
;         (NYXT:STYLE
;          (NYXT:FIND-SUBMODE 'NYXT/BOOKMARK-MODE:BOOKMARK-MODE
;                             (NYXT:CURRENT-BUFFER))))
;       (:H1 "Bookmarks")
;       (COND
;        ((ZEROP (HASH-TABLE-COUNT NYXT/BOOKMARK-MODE::BOOKMARKS))
;         (:P (FORMAT NIL "No bookmarks in ~s." #)))
;        (T (MAPHASH (LAMBDA # #) NYXT/BOOKMARK-MODE::BOOKMARKS))))
; --> LET SB-INT:DX-LET LET LET SPINNERET:WITH-HTML LET PROGN COND IF
; --> THE MAPHASH LET* SB-KERNEL:%COERCE-CALLABLE-TO-FUN
; ==>
;   1
;
; caught STYLE-WARNING:
;   The variable NYXT/BOOKMARK-MODE::REQUEST-ID is defined but never used.
;
; compilation unit finished
```
So this binding is removed.


aartaka (approved):

aartaka:
@shamazmazum, did you want to add something else? Do I merge it? Sorry for rebasing over master (⌒_⌒;)

shamazmazum:
@aartaka If you think it's OK, you can merge it. I'll send a PR with collapse/expand button later then.

aartaka:
OK, merging then!

aartaka:
Merged as 28c6ad3324ee392b292503f1da4120c844a1774a, thanks!

PR #2908, opened 2023-04-18
Refactor search buffer (Follow up to #2529 and #2424) 

aadcg:
# Description

This PR follows the strategy started in #2424, but the implementation is completely new.

It fixes long standing issues, and goes beyond the features of the old `search-buffer` capabilities.

Closes #1204.
Closes #2504.
Closes #1555. 
Closes #1237.
Closes #1169.

# Discussion

Mention there any suspicious parts of the new code, or the ideas that you'd like to discuss in regards to this change.

# Checklist:
Everything in this checklist is required for each PR.  Please do not approve a PR that does not have all of these items.

- [x] Git hygiene:
  - I have pulled from master before submitting this PR
  - There are no merge conflicts.
- [x] I've added the new dependencies as:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    git submodule add https://gitlab.common-lisp.net/nyxt/py-configparser _build/py-configparser
    ```
  - and Guix dependencies.
- [x] My code follows the style guidelines for Common Lisp code. See:
  - [Norvig & Pitman's Tutorial on Good Lisp Programming Style (PDF)](https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf)
  - [Google Common Lisp Style Guide](https://google.github.io/styleguide/lispguide.xml)
- [ ] I have performed a self-review of my own code.
- [ ] My code has been reviewed by at least one peer.  (The peer review to approve a PR counts.  The reviewer must download and test the code.)
- [ ] Documentation:
  - All my code has docstrings and `:documentation`s written in the aforementioned style.  (It's OK to skip the docstring for really trivial parts.)
  - I have updated the existing documentation to match my changes.
  - I have commented my code in hard-to-understand areas.
  - I have updated the `changelog.lisp` with my changes if it's anything user-facing (new features, important bug fix, compatibility breakage).
  - I have added a `migration.lisp` entry for all compatibility-breaking changes.
  - (If this changes something about the features showcased on Nyxt website) I have these changes described in the new/existing article at Nyxt website or will notify one of maintainters to do so.
- [ ] Compilation and tests:
  - My changes generate no new warnings.
  - I have added tests that prove my fix is effective or that my feature works.  (If possible.)
  - I ran the tests locally (`(asdf:test-system :nyxt)` and `(asdf:test-system :nyxt/gi-gtk)`) and they pass.

PR #2949, opened 2023-05-05
hint mode and search buffer mode must not share the same stylesheet

shamazmazum:
Hi! When you search for something with `seach-buffer` and then navigate with `follow-hint`, you get hints not shown as expected (it works also the other way around). This is because `nyxt/mode/hint` and `nyxt/mode/search-buffer` share the same stylesheet. This patch just renames stylesheets, so they can be used simultaneously.

This is what it looks like before the fix:
20230505_11h15m24s_grim

### DIFF
--- a/source/mode/hint.lisp
+++ b/source/mode/hint.lisp
@@ -158,7 +158,8 @@ For instance, to include images:
     (ps:chain element (remove-attribute "nyxt-hintable"))))
 
 (defun add-hints (&key selector (buffer (current-buffer)))
-  (add-stylesheet (style (find-submode 'hint-mode))
+  (add-stylesheet "nyxt-hint-stylesheet"
+                  (style (find-submode 'hint-mode))
                   buffer)
   (set-hintable-attribute selector)
   (update-document-model :buffer buffer)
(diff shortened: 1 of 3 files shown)


aadcg:
Thanks! This is indeed a bug that I've introduced in #2908. 

@aartaka, @jmercouris please review this PR, or I will next Tuesday.

P.S. @shamazmazum  хороший сайт)

aartaka:
Oh, it seems that we never clean up stylesheets from the page. I wonder if we should delete them, just as a hygienic thing.

Okay, that's a barely practical thought for some later time. This PR is absolutely sensible. Thanks @shamazmazum, I'm merging it!

aartaka:
Merged as 3986e57f45a9ac8bc30f51141b4bd0b590930568!

shamazmazum:
@aartaka Deleting stylesheets would be incorrect at least for `search-buffer`. For example, look at the screenshot. A search is over, but there are still hints left (`Goog` is highlighted). It should be OK to delete stylesheet after `remove-search-hints` though. Also `follow-hint` does not clean up `nyxt-hint` attributes after it's done. You can check it with `view-source` command after using `follow-hint`.

I cannot promise, but maybe I'll write a patch for cleaning stylesheets later.

20230505_14h37m22s_grim

aartaka:
@shamazmazum, you're right, no need to remove stylesheets!

PR #2954, opened 2023-05-10
renderer/gtk: Unexport some implementation-detail slots.

### LINKED ISSUE #1431: Overhaul user-classes

Ambrevar:
# Description

Fixes #1431.

# Checklist:

- [ ] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:

### DIFF
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -64,7 +64,8 @@ See also the `web-contexts' slot."))
                              (make-context name :ephemeral-p ephemeral-p)))
 
 (define-class gtk-window ()
-  ((gtk-object)
+  ((gtk-object
+    :export t)                          ; TODO: Unexport?
    (handler-ids
     :documentation "See `gtk-buffer' slot of the same name.")
    (root-box-layout)
@@ -82,8 +83,7 @@ See also the `web-contexts' slot."))
    (message-view)
    (key-string-buffer))
   (:export-class-name-p t)
-  ;; TODO: Unexport?
-  (:export-accessor-names-p t))
+  (:export-accessor-names-p nil))
 
 (define-class gtk-buffer ()
   ((gtk-object)
@@ -95,6 +95,7 @@ When given the same context, multiple buffers share their internal browsing data
 `+default+' is the default context.
 `+internal+' is a context that's not persisted to disk.")
    (handler-ids
+    :export nil
     :documentation "Store all GObject signal handler IDs so that we can disconnect the signal handler when the object is finalised.
 See https://developer.gnome.org/gobject/stable/gobject-Signals.html#signal-memory-management.")
    (gtk-proxy-url (quri:uri ""))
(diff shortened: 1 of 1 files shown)


aadcg:
All good, but the tests seem to be failing.

Ambrevar:
Very strange:

```log
Component ASDF/USER::CL-GOBJECT-INTROSPECTION not found, required by #<NASDF-SYSTEM "nyxt/gi-gtk">
Error: Component ASDF/USER::CL-GOBJECT-INTROSPECTION not found, required by
       #<NASDF-SYSTEM "nyxt/gi-gtk">
```

Note that cl-colors2 is not fetched properly either.  Any idea what's going on?

aadcg:
@Ambrevar I can't reproduce any of the 2 problems that CI reports locally. Please ignore and push to master.

Ambrevar:
I've rerun the jobs.

aartaka (approved):

aadcg:
@Ambrevar re-running resulted in the same issue. 

We don't know what's going on (and we must find the root cause), but I don't think it's important right now, since the goal is to merge this PR.

PR #2964, opened 2023-05-14
Add REPL tab-completion functionality.

aartaka:
# Description

Y'ALL ON ATLAS TEAM, WE'RE ALL ON VACATION, DON'T REVIEW THIS PR, GET SOME TIME FOR YOURSELF ヽ(ｏ`皿′ｏ)ﾉ

This allows us to tab-complete things in REPL (at least for `lisp-cell`s), which I myself found a painfully lacking feature.

# Discussion

- I'm not yet sure which shape the API should have. Should `complete` return 
  - A list of string? 
  - A list of symbols?
  - A single most intuitive symbol?
  - Most intuitive string (as it does right now)?
  - Should it maybe return nothing and do all the completion work itself, so that all this `ffi-buffer-paste` thing is more customizable per cell?

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:

### DIFF
--- a/source/mode/repl.lisp
+++ b/source/mode/repl.lisp
@@ -107,6 +107,11 @@ Generic function to specialize for all the REPL cell types.
 Use CELL (and, likely, its `input') and set its `results' and `output' to the
 meaningful values."))
 
+(defgeneric complete (cell)
+  (:documentation "Get a single most intuitive tab-completion string for CELL contents.
+Can rely on current contents of a cell or just use some context-free suggestions
+and `prompt'-ing."))
+
 (export-always 'render-input)
 (defgeneric render-input (cell)
   (:method ((cell cell))
@@ -185,6 +190,7 @@ Generic function to specialize against new REPL cell types."))
     :documentation "The package where Lisp cell reading and `evaluate'-ion happens.")
    (actions
     '(("Eval" evaluate-cell)
+      ("Tab-complete" complete-cell)
       ("Cancel" cancel-cell)
       ("Add cell below" add-cell)
       ("Clean" clean-cell)
@@ -252,6 +258,18 @@ The `input' should be a valid Lisp code `read'-able in the `eval-package'.
                      (setf results (multiple-value-list (eval s-exp))
                            output (get-output-stream-string *standard-output*)))
                    (safe-slurp-stream-forms in)))))))
+  (:method complete (self)
+    (alex:when-let ((completion (prompt1 :prompt "Completion"
+                                         :sources '(nyxt::variable-source
+                                                    nyxt::function-source
+                                                    nyxt::class-source
+                                                    nyxt::variable-non-nyxt-source
+                                                    nyxt::function-non-nyxt-source
+                                                    nyxt::class-non-nyxt-source
+                                                    nyxt::variable-internal-source
+                                                    nyxt::function-internal-source
+                                                    nyxt::class-internal-source))))
+      (prini-to-string completion :package (eval-package self))))
   (:method render-results (self)
     (spinneret:with-html-string
       (unless (uiop:emptyp output)
@@ -374,6 +392,7 @@ The `input' should be a valid Lisp code `read'-able in the `eval-package'.
     (define-keyscheme-map "repl-mode" ()
       keyscheme:default
       (list
+       "tab" 'complete-cell
        "C-return" 'evaluate-cell)
       keyscheme:emacs
       (list
@@ -393,7 +412,8 @@ The `input' should be a valid Lisp code `read'-able in the `eval-package'.
        "C-_" 'move-cell-down
        "C-k" 'clean-cell
        "C-M-k" 'delete-cell
-       "M-q" 'reformat-cell)
+       "M-q" 'reformat-cell
+       "C-i" 'complete-cell)
       keyscheme:vi-normal
       (list
        ;; TODO: deleting chars/words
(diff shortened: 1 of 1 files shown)


source/mode/repl.lisp
```diff
@@ -107,6 +107,11 @@ Generic function to specialize for all the REPL cell types.
 Use CELL (and, likely, its `input') and set its `results' and `output' to the
 meaningful values."))
 
+(defgeneric complete (cell)
+  (:documentation "Get a single most intuitive tab-completion string for CELL contents.
+Can rely on current contents of a cell or just use some context-free suggestions
+and `prompt'-ing."))
+
```
aadcg, line 114: Suggestion: "Perform completion on the text around the CELL's cursor.
The completion suggestions rely on the CELL's contents, when possible."

aadcg (commented):
> I'm not yet sure which shape the API should have. Should complete return 

I think it should return the top suggestion as a string. This suggestion may be set interactively when the user uses the prompt buffer.

-------

Issue #1: When completing, the symbol around the cursor doesn't get pasted onto the prompt buffer.



---- 

Issue #2: If I type a prefix of the symbol I'm trying to complete, it won't appear at the top of the suggestions. This may be a limitation of the prompt buffer. Perhaps it doesn't sort sources, I didn't check it.

Ambrevar:
Looks good to me other than André's remarks.

source/mode/repl.lisp
aartaka: It doesn't always rely on text around the cursor. It doesn't do that for `lisp-cell`, at least.

aartaka:
> I think it should return the top suggestion as a string. This suggestion may be set interactively when the user uses the prompt buffer.

Good, then no change to the current API :D

> Issue #1: When completing, the symbol around the cursor doesn't get pasted onto the prompt buffer.

Yes, doesn't. That deliberate: instead of parsing the contents of the cell (compiler-complete problem) and somehow replacing it (requiring JS), we just list all the suggestions and paste the selected one verbatim.

This is not a perfect API, but:
- It's reliable.
- And it's an improvement over the previous state.

So, unless we come up with something better, this one would be fine :)
 
But then, should we call it tab-completion if it's not actually completing? Maybe call it `suggest` or something?

> Issue #2: If I type a prefix of the symbol I'm trying to complete, it won't appear at the top of the suggestions. This may be a limitation of the prompt buffer. Perhaps it doesn't sort sources, I didn't check it.

Yeah, it doesn't. I've reordered the sources to be more intuitive in 82663789a5a74be4381839a11ebf16a4b238f98e, but there's not much to do beyond that :)

source/mode/repl.lisp
aadcg: If I'm understanding you correctly, it seems that you should delete "Can rely on current contents of a cell" from the docstring.

aartaka: I mean, some of the methods can potentially rely on it, and this docstring is a generic documentation, which means that this umbrella more or less applies :)

aartaka:
So, putting it on a backburner? Or delete a button and ship :D I mean, yeah, something like general tab-completion won't hurt, but this REPL-local hack is okay for the use-case e(((o(*ﾟ▽ﾟ*)o)))

aartaka:
All addressed:
- Button removed.
- `complete` and friends are renamed to `suggest` etc.

Rebasing, squashing, merging?

jmercouris:
Hello Artyom, thank you very much for this! It does in fact make the REPL infinitely more useful. I find myself struggling to remember symbols, and this *really helps*.

With regards to this PR, it is I believe an improvement. Ideally, we need to consider a completion framework at the level of input-edit as Andre suggested.

That said, I think this should be merged in because it begins the development of a framework. When the next time happens that we want to add tab-completion we'll have a basis of which to build off. A way in order to see how the code could be generalized for reuse. We'll discover the API.

I am thinking of the approach as delineated in Practical Common Lisp of writing code and then replacing things with Macros as the correct solution emerges (https://gigamonkeys.com/book/practical-a-simple-database.html).

source/mode/repl.lisp
aadcg: I can now understand your perspective! Clear, thanks.

aadcg:
@aartaka I'd like to emphasize that this is a fantastic improvement! 

The remarks I made only make sense when reviewing the PR based on the title. I think it would be premature to jump at solving a big and general problem when we can have this simple and elegant solution that enhances the REPL experience.

Thank you for working on this.

For my side, it's ready to be merged. I'm not sold on the name of the command, `suggest-into-cell, but I don't have an amazing alternative. The best I can come up with is `suggest-at-point`. Please decide as you find most adequate, it's not very important.

aartaka:
> For my side, it's ready to be merged. I'm not sold on the name of the command, `suggest-into-cell`, but I don't have an amazing alternative. The best I can come up with is `suggest-at-point`. Please decide as you find most adequate, it's not very important.

I mean, the `*-at-point` Emacs convention is exactly the content-based actions. Which the current state not necessarily is. So I'd leave `*-into-cell` name, because the only thing it implies is adding something to the content.

aadcg:
> I mean, the *-at-point Emacs convention is exactly the content-based actions. 

This is not correct. See `face-at-point`. At point simply means "where the cursor is".  Nonetheless, as I've said, I don't think it's very important :)

PR #2978, opened 2023-05-23
Re-enable global history

### LINKED ISSUE #2748: Drop one of `global-history-p` and `conservative-history-movement-p`?
`global-history-p` cuts off buffer interrelations, while `conservative-history-movement-p` only hides buffer interrelations in the interface and commands. Seems like those two are doing roughly the same thing, but `global-history-p` is more radical and simple, while `conservative-history-movement-p` is slightly more cautious, at the cost of more mental gymnastics.

So, which one do we retain of those duplicative concepts. I think we should... drop `conservative-history-movement-p` in favor of `global-history-p`?

_Originally posted by @aartaka in https://github.com/atlas-engineer/nyxt/pull/2657#discussion_r1081559876_

### LINKED ISSUE #2862: Re-enable global history
Setting `global-history-p` to NIL by default cuts off one of the (subjectively) most useful and novel Nyxt features—history globality. Given that this change was smuggled through in #2708 [without any announcement](https://github.com/atlas-engineer/nyxt/compare/15e3ad0c78c86710f6347d4fdd8faa4afefc2d65..b929361ba0b342da75ec56aa89e6cce919b14e10), there wasn't much discussion (except for a private call between @aadcg and @jmercouris) about whether it's right to set it to NIL. 

#1677 cannot be used as a justification for this change: the last state of it was
- GHT is unreliable.
- But so it every new approach and code.
- Let's fix the actual bugs there are, and see how it compares to the latest, bug-free, state of per-buffer history.

Not all bugs were fixed prior to #2708 (and #2708 itself doesn't fix any either):
- #2660
- #2252
- #1117
- #1848

thus global history may not be said to be inferior and one must have a solid reason to disable it. There's none of such reasons that I've seen. Prove me (or, rather, global history) wrong @jmercouris.

aartaka:
# Description

This toggles global history back on and removes the `conservative-history-movement-p` slot deprecated by `global-history-p`.

Fixes #2862
Fixes #2748

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:


source/mode/history.lisp
```diff
@@ -37,10 +37,6 @@ are useful internal APIs for history reusal:
 Example: DuckDuckGo redirections should be ignored or else going backward in
 history after consulting a result reloads the result, not the DuckDuckGo
 search.")
-   (conservative-history-movement-p
-    nil
-    :type boolean
-    :documentation "Whether history navigation is restricted by buffer-local history.")
```
Ambrevar, line 43: We've made the mistake of releasing 3.0.0 without marking this as deprecated.  We can't remove it on the 3.x branch without breaking backward compatibility.

So retain the slot and mark it as deprecated.

Exercise: Maybe add a `slot-value-using-class` method for deprecated slots that warns the user on each read/write.  Or maybe just on the first one.

source/buffer.lisp
```diff
@@ -470,7 +470,7 @@ Suggestions are computed by the default search engine.")
     :documentation "Select a download engine to use, such as `:lisp' or
 `:renderer'.")
    (global-history-p
-    nil
+    t
```
aadcg, line 473: In my understanding, making this change means that we're changing the functionality in a backwards-compatible way, which required bumping the MINOR version. I.e., our next release will be 3.1.0.

Does this make sense @jmercouris?

Ambrevar: Yes!

source/mode/history.lisp
Ambrevar: To expand on the exercise, it would be nice if we could somehow automate this.  For instance expand user-class to support a `:deprecated t` slot option, then in Nyxt we define our `slot-value-using-class` on user-classes which checks if the slot is deprecated and triggers a warning accordingly.

aartaka: > We've made the mistake of releasing 3.0.0 without marking this as deprecated. We can't remove it on the 3.x branch without breaking backward compatibility.

Yes ヽ(ｏ`皿′ｏ)ﾉ

> So retain the slot and mark it as deprecated.

OK!

> Exercise: Maybe add a `slot-value-using-class` method for deprecated slots that warns the user on each read/write. Or maybe just on the first one.

Too early for this one, I guess. I'll just update all the uses of `global-history-p` with the `(or (global-history-p ...) (conservative-...p ...))`.

aartaka:
Okay, it all seems addressed:
- `conservative-history-movement-p` is deprecated and all the uses of it are replaced with `global-history-p`, which itself hooks into `conservative-history-movement-p` in some cases.
- Changelog updated to state 3.1.0.
- And the change I couldn't resist temptation to add: parent buffer is provided in much more cases, which makes global history much more useful and makes the behavior much more predictable!

aartaka:
One more round of review y'all, and we can merge :)

source/manual.lisp
```diff
@@ -256,10 +256,17 @@ commands like " (:nxref :command 'nyxt/mode/history:history-backwards) " and "
...
+            " to be NIL and " (:nxref :slot 'nyxt/mode/history:conservative-history-movement-p
+                                      :class-name 'nyxt/mode/history:history-mode)
+            " to T:"
+            (:ncode
+             '(define-configuration :history-mode
+               (conservative-history-movement-p t))
+             '(define-configuration :context-buffer
+               (global-history-p nil)))
+            " This would make all buffers to have their own history, not connected to the
+other buffers at all. All the history commands (like "
```
aadcg, line 269: I'm confused. The changelog entry says that `conservative-history-movement-p` has been deprecated in favour of `global-history-p`. But the manual states and one can set these two to opposite boolean values and get a specialized behaviour...

source/mode/history.lisp
```diff
@@ -92,6 +93,23 @@ Experimental, may not always produce intuitive enough history trees.")
...
+      ;; Quite a messed up heuristic, but it has the desired properties:
+      ;; - If `global-history-p' is on, then it's on.
+      ;; - If `conservative-history-movement-p' is on, while `global-history-p'
+      ;;   is off, then it's all off (kind of a safety switch to make it
+      ;;   consistent).
+      ;; - Any other combination (only global off conservative off
+      ;;   really)---off.
+      (global-p t) ; Global history enabled and has a priority in general.
+      (conservative-p nil) ; Global history disabled.
+      (t nil))))
```
aadcg, line 111: Hm... Wouldn't it better to set `global-history-p` to `t` if either `global-history-p` or `conservative-p` is `t`?

source/manual.lisp
aadcg: Ok, it actually makes sense because the negation of `global-history-p` is equivalent to the truth value of `conservative-history-movement-p`.

aadcg: I'd recommend referring to `global-history-p` in the manual exclusively. Otherwise it's more confusing for the user.

source/mode/history.lisp
aadcg: I'm not sure if I agree here. I think `global-history-p` should be `t` when either `global-history-p` is `t` or `conservative-p` is `nil`.

In the way you wrote it, it means that when `global-history-p` is `f` and `conservative-p` is `t`, then `global-history-p` will be `f`. I think it should be set to `t` in that case.

source/mode/hint.lisp
```diff
@@ -420,7 +420,6 @@ FUNCTION is the action to perform on the selected elements."
 
 (defmethod %follow-hint-new-buffer-focus ((a nyxt/dom:a-element) &optional parent-buffer)
   (make-buffer-focus :url (url a)
-                     :parent-buffer parent-buffer
```
Ambrevar, line 423: Why remove this?

source/mode/history.lisp
Ambrevar: Why early?  I think the time is right indeed, considering we've just released a major version.

But if you believe this can wait for later, could you open an issue about it then?

source/mode/history.lisp
```diff
@@ -92,6 +93,23 @@ Experimental, may not always produce intuitive enough history trees.")
        "L" 'history-forwards-maybe-query
        "M-H" 'history-all-query)))))
 
+(defmethod global-history-p :around ((buffer context-buffer))
+  ;; TODO: Remove in 4.0 after `conservative-history-movement-p' is deprecated.
+  (let ((conservative-p (when (find-submode 'history-mode)
+                          (conservative-history-movement-p (find-submode 'history-mode))))
```
Ambrevar, line 99: Let-bind the mode resulting from the first `find-submode`.

Time for a macro? :)

source/mode/history.lisp
```diff
@@ -92,6 +93,23 @@ Experimental, may not always produce intuitive enough history trees.")
...
+    (cond
+      ;; Quite a messed up heuristic, but it has the desired properties:
+      ;; - If `global-history-p' is on, then it's on.
+      ;; - If `conservative-history-movement-p' is on, while `global-history-p'
+      ;;   is off, then it's all off (kind of a safety switch to make it
+      ;;   consistent).
+      ;; - Any other combination (only global off conservative off
+      ;;   really)---off.
+      (global-p t) ; Global history enabled and has a priority in general.
+      (conservative-p nil) ; Global history disabled.
```
Ambrevar, line 110: The above condition is not needed then: the function returns nil if it's not there, because of the `(t nil)` below.

To simplify your logical statement: Only `global-p` is obeyed :)

source/mode/history.lisp
Ambrevar: > `conservative-p` is deprecated, but it should have the advertised effect I believe.

It's a tough one, because we have not clear user-facing way to see if a slot is set or not.
So if `global-history-p` had priority over `conservative-p`, then there is no way to set `conservative-p` and have Nyxt obey it, thus bypassing `global-history-p` and breaking the priority.

source/manual.lisp
aartaka: So the reason manual mentions it is that disabling the global history is only possible with both slots redefined. Which is kind of borked up, but that's the best heuristic I was able to come up with :( 

I mean, if we make `conservative-history-movement-p` do nothing, that wouln't be cool either (҂⌣̀_⌣́)

source/mode/history.lisp
aartaka: > In the way it's written: when `global-history-p` is `f` and `conservative-p` is `t`, then `global-history-p` will be `f`. I think it should be set to `t` in that case.

Waaait. What this combination says is: disable global history and enable the conservative (=buffer-local) movement. Which is a tautologous "disable globality" statement. Am I reading it wrong?

In general, this heuristic, however weird it is, encapsulates the most logical behavior I was able to come up with:
```
|                     | global... t                        | global... nil                     |
| conservative... t   | global (global-history-p priority) | local (tautology)                 |
| conservative... nil | global                             | local (global-history-p priority) |
```

Does it make sense? Which or this truth table cells feels off to you?

source/mode/hint.lisp
aartaka: Fixed in d08d050af985ca6691b5de5e09f1719ef730df58!

source/mode/history.lisp
aartaka: Yeah, could be a time for a macro :D

source/mode/history.lisp
aadcg: > Waaait. What this combination says is: disable global history and enable the conservative (=buffer-local) movement. Which is a tautologous "disable globality" statement. Am I reading it wrong?

You're correct. I don't understand why I've said it, since it makes no sense at all! What I wanted to say was that in the way it's done, `conservative-p` is not honored.

@Ambrevar has summarized it already: 

> So if global-history-p had priority over conservative-p, then there is no way to set conservative-p and have Nyxt obey it, thus bypassing global-history-p and breaking the priority.

> To simplify your logical statement: Only global-p is obeyed :)

And he pointed out accurately why this is not an easy:

> It's a tough one, because we have not clear user-facing way to see if a slot is set or not.

aadcg:
With respect to the slot options `global-history-p` and `conservative-history-movement-p` (on master, before this PR), we need to acknowledge that we've already messed up big time. Notice that the defaults are `nil` and `nil`, respectively, which semantically makes no sense.

We can't change the API because would require releasing a new major version. The "best" thing that comes to my mind is to keep the slot `conservative-history-movement-p` as a no-op and write in its documentation that it is ignored and users must use `global-history-p`. The manual should not mention it too.

What do you think? @aartaka @Ambrevar

source/mode/history.lisp
Ambrevar: Thanks!

Ambrevar:
I agree with @addcg.

aartaka:
Good, will adjust and merge then!

PR #2980, opened 2023-05-24
bash to uiop syntax for external-editor-program

simendsjo:
`VISUAL` and `EDITOR` might contain parameters, which must be converted to a list of strings for uiop:launch-program to work.

E.g. `"emacsclient --create-frame"` to `'("emacsclient" "--create-frame")`


aadcg:
Thanks @simendsjo.

@Ambrevar @aartaka do you approve?  If we agree on merging it, then it must be part of 4.0.0 as it breaks the 
logic of this slot.

Before we merge, we need to add a commit that to re-write the docstring and another that adds an entry to the changelog. @simendsjo, don't worry, we'll take it from here by ourselves!

simendsjo:
> @simendsjo, don't worry, we'll take it from here by ourselves!

I don't mind getting to know more of the process and do the work required. Doesn't seem like there's any rush as it's delayed until the next major release. I'll look into the docstring and changelog, but you want that as a separate commit rather than together in the same commit?

aadcg:
Thanks @simendsjo. The docstring fix could be squashed onto 55dcdaa3bb61dd07a178ab254acba9a5598b38c6, but a separate commit is required for the changelog. You may need to create a new entry for version 4.0.0.

aartaka:
> @Ambrevar @aartaka do you approve? 

I approve, yes.

> If we agree on merging it, then it must be part of 4.0.0 as it breaks the logic of this slot.

I don't think so. Using a single string with spaces as a value for external editor was quite unreliable anyway, because `uiop:run-program` handles it differently on different implementations and for different inputs. So I believe splitting it is a safe option, and quite a backwards-compatible one (because it was slightly broken previously :P)

aartaka:
The only thing I'm worried about is that splitting by mere spaces is not enough for proper shell invocations, because string literals are not respected. This will get parsed wrong, for instance:
``` sh
commands --arg --string-arg "a string with spaces in it"
```

Something like rudimentary parsing for the value won't hurt, but we can use the same reasoning as above, saying that providing proper input is the user responsibility (no, don't do that) xD

simendsjo:
> The only thing I'm worried about is that splitting by mere spaces is not enough for proper shell invocations, because string literals are not respected. This will get parsed wrong, for instance:

Yes, it isn't a 100% solution. I thought about only splitting on the first space, but uiop fails on something like `'("a" "-b -c")` too unfortunately.

> Something like rudimentary parsing for the value won't hurt, but we can use the same reasoning as above, saying that providing proper input is the user responsibility (no, don't do that) xD

We could avoid splitting inside `'` and `"` and support escaping, but maybe that's a bit overkill? In practice, you probably have an external program defined without strings. But if you do, this solution will make it impossible for you to laurch the program. Maybe we should accept a string and do the stupid str:split, but also accept a list which is sent verbatim to uiop? Then a user have full flexibility, but we also support a standard VISUAL/EDITOR environment variable.

aartaka:
@simendsjo, let's merge your implementation and then I may get to updating it some day (my `emacsclient` invokation is a five-lines long set of eval expressions, so I'll likely do that soon enough xD)

@aadcg, so you take care of this PR cleanup and merge?

simendsjo:
> @simendsjo, let's merge your implementation and then I may get to updating it some day (my emacsclient invokation is a five-lines long set of eval expressions, so I'll likely do that soon enough xD)

Sure it's not better to move it down the stack instead? Ref https://github.com/atlas-engineer/nyxt/pull/2980#issuecomment-1562381454

aartaka:
Sure, thanks!

simendsjo:
> Sure, thanks!

Updated.

aadcg:
To be honest, after the amazing exchange of ideas you had, I'm inclined to say that this would actually be a regression. I'd like to listen to @Ambrevar on this one.

simendsjo:
> Makes sense? @simendsjo @aartaka ?

Sounds good to me :+1: I'll make a new change iff aartaka agrees.

aartaka:
Yes, that should work alright!

simendsjo:
Not sure what your branching and versioning strategy is. Rebased on master, but add a changelog entry like it's 3.0.1.. :shrug: Should I rather just add it as an entry to the 3.X.Y changelog? Guess it's not worth backporting and supporting multiple versions at this point?

source/external-editor.lisp
```diff
@@ -13,12 +13,15 @@ so invoke on a separate thread when possible."
...
     (with-protect ("Failed editing: ~a" :condition)
-      (uiop:run-program (append (uiop:ensure-list (external-editor-program *browser*))
-                                (list (uiop:native-namestring p)))
-                        :ignore-error-status t))
+      (let* ((program (external-editor-program *browser*))
+             (path (uiop:native-namestring p))
+             (command (if (stringp program)
+                          (str:concat program " " path)
+                          (append program (list path)))))
+        (log:debug "External editor `~s` opens `~s` by executing `~s`"
```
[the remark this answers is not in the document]
Ambrevar: Also I suppose that only `command` is relevant now, no need to mention `program` or `path`.

Ambrevar:
This is backward compatible, to it will end on the 3-series branch.  Keep the changelog as it is, it's perfect.

source/external-editor.lisp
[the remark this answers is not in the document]
simendsjo: Fixed the message. Also noticed that the exact same pattern was used when viewing files too, so I extracted a method and fixed that too. Probably not too controversion, so I squashed them together (making review more difficult).

Ambrevar:
Looks good to me!

source/external-editor.lisp
```diff
@@ -3,6 +3,14 @@
 
 (in-package :nyxt)
 
+(defun %run-external-editor-command (path &optional (program (external-editor-program *browser*)))
+  (let* ((native-path (uiop:native-namestring path))
+         (command (if (stringp program)
```
aartaka, line 8: Note that the `stringp` never returns because of:
``` lisp
;; browser.lisp, L291
(defmethod external-editor-program ((browser browser))
  (alex:ensure-list (slot-value browser 'external-editor-program)))
```

Possible fix would be:
- Make a check for spaces in `defmethod external-editor-program` and only return list if there's no spaces (unreliable, but mostly DWIM).
  - Return string if there are spaces (likely a spaced command, so no lists).
- Profit!

source/external-editor.lisp
```diff
@@ -3,6 +3,14 @@
...
+                      (str:concat program " " native-path)
+                      (append program (list native-path)))))
+    (log:debug "External editor opens ~s" command)
```
aartaka, line 11: I'd retain the old
``` lisp
(log:debug "External editor ~s opens ~s" command native-path)
```

aartaka (commented):
I've tested it and it breaks with spaced string values because they get converted into lists unconditionally (see the full recipe below.)

source/external-editor.lisp
simendsjo: Hmm, I've had a too narrow search for usage, and I see there are even more places where this is used. Let me do another pass, extending my search and doing this more thoroughly.

image

simendsjo:
Updated the code to fix the other instances of external-editor-program I found. I wanted to write some tests for my new helper function %append-uiop-command, but I had some problems getting it to work. Running `(asdf:test-system :nyxt)` just returns `t`. Have to admit I don't have the most elegant development environment here yet.

source/external-editor.lisp
aartaka: Ohhhh, right! Forget it then :)

source/external-editor.lisp
aartaka: > The external-editor-program slot is already defined as list or string, and we know that we need to take care to avoid mixing for uiop. So couldn't we just return the value as-is instead? ensure-list at this point will break when it contains spaces, and returning a list will stop us from adding additional parameters without trying to reverse it later.

Yes, removing this custom method soulds like the best solution.

makefile
```diff
@@ -42,8 +42,8 @@ lisp_eval:=$(LISP) $(LISP_FLAGS) \
 	--eval '(asdf:load-asd "$(makefile_dir)/libraries/nasdf/nasdf.asd")' \
 	--eval '(asdf:load-asd "$(makefile_dir)/nyxt.asd")' \
 	--eval '(format t "~&; NASDF ~a: ~a~&" (asdf:component-version (asdf:find-system :nasdf)) (nth-value 2 (asdf:locate-system :nasdf)))' \
-  --eval '(when (find-package :ql) (funcall (read-from-string "ql:quickload") :cffi))' \
```
aartaka, line 45: This doesn't belong to the PR, because it doesn't directly relate to the feature in question. But I guess it's okay to have it 😅

source/external-editor.lisp
```diff
@@ -3,6 +3,34 @@
 
 (in-package :nyxt)
 
+(defun %append-uiop-command (command &rest args)
+  "Appends `ARGS' to an existing UIOP `COMMAND' (for `UIOP:RUN-PROGRAM' or
```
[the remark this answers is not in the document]
aartaka: Adding to that, having a bit more "air" in the docstring would be nice too. Like putting every
- `If COMMAND is a string ...`
- `If COMMAND is a list ...`
on their own fresh line or even paragraph.

And note that we have a clear-cut distinction between markups for:
- Arguments: all caps, no backtick-quote reference: `COMMAND`,
- And external symbols: lowercase (unless the symbol is actually mixed/upper case), and with backtick-quotes: 
```
`uiop:run-program'
```

The rationale behind it is that:
- Symbols beyond the function benefit a lot from backtick-quote (`xref`) syntax, because Emacs/SLY/Nyxt automatically parses these and resolves them to the right function/class/etc., making the documentation interlinkable.
- Function arguments don't need backtick-quote references, because there's nothing to resolve them to. So we jusp put them uppercase, which is a tradition in the community 😉

aartaka (approved):
That's an iteration I really like! Some minor docstring fixes (look through our `docs/README#programming-conventions` and the referenced style guides so that we spend a bit less time on thise), and we're good to go!

source/external-editor.lisp
simendsjo: I didn't remove the method, only the `ensure-list`. Guess both cases is a breaking change. Should I remove it completely and just require people to use `slot-value` instead?

makefile
simendsjo: It was a separate commit, but I moved the commit to its own PR.

source/browser.lisp
```diff
@@ -289,7 +289,7 @@ prevents otherwise.")
   (make-instance 'theme:theme))
 
 (defmethod external-editor-program ((browser browser))
```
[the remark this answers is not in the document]
simendsjo: Nice, I'll push a new change.. Cannot say I like the extremely verbose defclass.

source/external-editor.lisp
simendsjo: Noticed your other comment, so I removed it completely along with :writer t

source/external-editor.lisp
```diff
@@ -3,6 +3,34 @@
...
+`UIOP:LAUNCH-PROGRAM'.) If `COMMAND' is a string, `ARGS' is concatenated to it
+with spaces between the arguments. If `COMMAND' is a list, `ARGS' is appended to
+it. Note: These functions expect either the entire command as a string, or a
```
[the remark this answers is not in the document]
simendsjo: I don't get along very well with the CL type system or its syntax. Tried various things, but I always get errors similar to the one below. How should it look like? It takes `((or string list) &rest list)` as input and returns `(or string list)` as output.

```
; (VALUES &OPTIONAL (OR LIST STRING) &REST T).
; (VALUES &OPTIONAL (OR LIST STRING)).
```

source/external-editor.lisp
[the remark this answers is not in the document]
simendsjo: Thanks, very useful! I couldn't find any CONTRIBUTING file or similar, but this might be common knowledge?

source/external-editor.lisp
```diff
@@ -3,6 +3,40 @@
 
 (in-package :nyxt)
 
+(-> %append-uiop-command ((or string list) &rest string) (values (or list string) &optional))
+(defun %append-uiop-command (command &rest args)
+  "Appends ARGS to an existing uiop COMMAND (for `uiop:run-program' or
```
Ambrevar, line 8: Remove "uiop",

source/external-editor.lisp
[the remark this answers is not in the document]
simendsjo: See the documents folder contains a lot. I'll dig into those: https://github.com/atlas-engineer/nyxt/tree/master/documents

simendsjo:
> That's an iteration I really like! Some minor docstring fixes (look through our `docs/README#programming-conventions` and the referenced style guides so that we spend a bit less time on these), and we're good to go!

I'll look over the documents and make necessary changes. But I need some help on the type declaration for `%append-uiop-command` before it's ready, ref https://github.com/atlas-engineer/nyxt/pull/2980#discussion_r1225497477.

source/external-editor.lisp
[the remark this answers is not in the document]
Ambrevar: This is correct.  You need `(values ... &optional)` because `apply #'str:concat` type cannot be properly inferred (most likely an issue with str:concat).  You could try using `uiop:strcat` instead, not sure that would help.

makefile
aartaka: Thanks!

source/external-editor.lisp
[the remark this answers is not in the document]
aartaka: Or you can just remove the return type or set it to `t` (that's what I myself often do, because SBCL is often _too_ strict about types, so it's simpler to just give up on some functions 😛)

source/external-editor.lisp
simendsjo: I interpreted it as change "uiop COMMAND" to "COMMAND" in the first sentence.

source/external-editor.lisp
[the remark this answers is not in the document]
simendsjo: > This is correct. You need (values ... &optional) because apply #'str:concat type cannot be properly inferred (most likely an issue with str:concat). You could try using uiop:strcat instead, not sure that would help.

The problem is the intersperse, so I'd need a macro to run it without `apply`, or is there another way of writing this?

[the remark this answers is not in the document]
simendsjo: > I even found a more specialized function: (uiop:reduce/strcat `(,command " " ,@(sera:intersperse " " args)))
> 
> Is `append (list ...)` better than `,@` here?

Can avoid this by using `str:unwords` instead: `(uiop:reduce/strcat (list command " " (str:unwords args)))`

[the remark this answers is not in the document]
simendsjo: >> I even found a more specialized function: (uiop:reduce/strcat `(,command " " ,@(sera:intersperse " " args)))
> Right!

Even better, and I'll stop here: `(uiop:reduce/strcat (list command " " (str:unwords args)))`

> (..) because the behavior of literals is unpredictable when modified.

Ouch. I don't understand why this is happening though. Is it because it reuses the cdr of the list as it's a literal? While `list` will always allocate a new list?

> While for apply the number of arguments is unknown, because the last argument is a list.
> (..) And every element of that list needs to be allocated on the stack. (..)

I see. I'll favor folding when possible.

> Looked it up in the spec (for function type) and has gotten extremely suprised:

For me, either made just as much sense as the other :)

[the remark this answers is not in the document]
aartaka: > Ouch. I don't understand why this is happening though. Is it because it reuses the cdr of the list as it's a literal? While `list` will always allocate a new list?

Yes `list` always allocates a new unique (per `eq`) list. While [most of the literals in CL have undefined behavior when modified, including the list ones](https://cl-community-spec.github.io/pages/Modification-of-Literal-Objects.html). Which is an unfortunate side of the standard 😅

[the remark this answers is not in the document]
aartaka: > > Looked it up in the spec (for function type) and has gotten extremely suprised:
> 
> For me, either made just as much sense as the other :)

Well, it kinda makes sense, because it's hard to have typed lists (see our `list-of` type—it's a hack). While typing the `&rest` as the type of its arguments allows to have type-checking for every arg, which is a huge win 😃

source/external-editor.lisp
aartaka: Ah, right, sorry 😅

simendsjo:
Thanks for the great input so far. I believe it's necessary for a new full review as there has been a lot of changes and some back and forth. I haven't read all the style guides yet, but I don't believe it can be too bad.

Note that I signal errors in the append function to avoid getting into a situation where the first argument is treated as the command.

And we have two changes here; the bugfix to support spaces in command strings, and the breaking change where the method which fetches the slot value no longer listifies the value.

Sorry for taking up so much of your time on this little change.

source/external-editor.lisp
```diff
@@ -3,6 +3,44 @@
...
+
+(export-always '%run-external-editor)
+(defun %run-external-editor (path &optional (program (external-editor-program *browser*)))
+  "Calls `uiop:run-program' with PATH as an extra parameter to PROGRAM.
+PROGRAM defaults to `external-editor-program'"
+  (let ((command (%append-uiop-command program (uiop:native-namestring path))))
+    (log:debug "External editor opens ~s" command)
+    (uiop:run-program command :ignore-error-status t)))
+
+(export-always '%launch-external-editor)
```
Ambrevar, line 36: Same about `%`.

source/external-editor.lisp
```diff
@@ -3,6 +3,44 @@
...
+  ;; subsequent elements. Mixing them signals an error. This is the reason for
+  ;; this custom append function.
+  (cond
+    ((null command) (error "Unable to append arguments to a null command."))
+    ((str:emptyp command) (error "Unable to append arguments to an empty command."))
+    ((null args) command)
+    ((stringp command) (uiop:reduce/strcat (list command " " (str:unwords args))))
+    ((consp command) (append command args))))
+
+(export-always '%run-external-editor)
```
Ambrevar, line 28: Prefixing with % (which is not idiomatic in Nyxt) contradicts the export.

Remove the prefix?

source/external-editor.lisp
```diff
@@ -3,6 +3,44 @@
...
+(defun %run-external-editor (path &optional (program (external-editor-program *browser*)))
+  "Calls `uiop:run-program' with PATH as an extra parameter to PROGRAM.
+PROGRAM defaults to `external-editor-program'"
+  (let ((command (%append-uiop-command program (uiop:native-namestring path))))
+    (log:debug "External editor opens ~s" command)
+    (uiop:run-program command :ignore-error-status t)))
```
Ambrevar, line 34: Shall we really ignore the status?  What happens on error?

simendsjo: I have no idea. The existing calls did this, so I just kept it. I hate swallowing errors though, so it looks scary.

source/external-editor.lisp
```diff
@@ -3,6 +3,44 @@
 
 (in-package :nyxt)
 
+(-> %append-uiop-command ((or string (list-of string)) &rest string) (values (or string (list-of string)) &optional))
+(defun %append-uiop-command (command &rest args)
+  "Appends ARGS to an existing COMMAND (for `uiop:run-program' or
+`uiop:launch-program').
```
aartaka, line 9: First line of docstring should be a complete sentence and vice versa—first sentence of the docstring should be the first line of it 😉

aartaka (approved):
Thanks for getting through all of this together with us @simendsjo! I envy your persistence ❤️

This is ready for merging, I believe. Just:
- Fix the small docstring thing in my comments below.
- Rebase over master.
- And we'll happily merge it 😃

source/external-editor.lisp
simendsjo: Ok, I'll remove this in it's own commit and use your comment as a basis for the commit message.

source/external-editor.lisp
simendsjo: Not sure what you mean. Should I avoid the line wrapping? Or rephrase it to avoid the parenthesis?

simendsjo:
> Thanks for getting through all of this together with us @simendsjo! I envy your persistence :heart: 

I like the thorough review and focus on quality! I'm not in a hurry, and I'm learning a lot, so I don't mind.

Ambrevar:
Almost there! :)
Please squash "PR FIX: Don't prefix exported functions with %" into the previous commit :)

simendsjo:
> Almost there! :)
> Please squash "PR FIX: Don't prefix exported functions with %" into the previous commit :)

Still missing the docstring fix as I'm not sure how @aartaka means it should be fixed.

source/external-editor.lisp
aartaka: I mean this:
``` lisp
(defun %append-uiop-command (command &rest args)
  ;; Notice the sentence up until the full stop being all on first line
  "Appends ARGS to an existing COMMAND (for `uiop:run-program' or `uiop:launch-program').

..."
  #|...|#)
```

simendsjo: >   ;; Notice the sentence up until the full stop being all on first line

I see. My editor wraps lines automatically, so it's difficult to maintain too long lines.
But I'm fixing manually.

simendsjo:
> This is ready for merging, I believe. Just:
> 
>     * Fix the small docstring thing in my comments below.
> 
>     * Rebase over master.
> 
>     * And we'll happily merge it smiley

Done

aartaka:
Merged as 7a7d1f292368f14dbca487895bb831b0e20453a3! Thanks a lot @simendsjo!

simendsjo:
> These changes seems to not break the API, so can we ship them in last release in the 3-series? Thanks.

Guess it depends a bit on the definition of bugfix :)

The slot accessor `external-editor-program` would always return a list. When the slot was just a single string without any spaces, e.g. "emacs", everything would work as expected, but the accessor returned `'("emacs")`. After this change, the same value would return `"emacs"`.

It also returned a list when the value contained spaces, but then it would break when trying to launch the program, so I see that difference in behavior as part of the bugfix.

> Small detail: changelog commits should be separate.

Ok, I'll do that next time.

aartaka:
> The slot accessor `external-editor-program` would always return a list. When the slot was just a single string without any spaces, e.g. "emacs", everything would work as expected, but the accessor returned `'("emacs")`. After this change, the same value would return `"emacs"`.

I guess that it's a minor breakage, because we mostly used `external-editor-program` (pun intended) internally 😜

simendsjo:
Sorry, but I fail to see the bug or how the change actually has any effect at all. You're just moving the consp check further up? But it shouldn't interfere with the string checks?

The following test shows that the old and new implementation gives the same result, so maybe I'm missing something?

```lisp
(defun %append-uiop-command-old (command &rest args)
  (cond
    ((null command) (error "Unable to append arguments to a null command."))
    ((str:emptyp command) (error "Unable to append arguments to an empty command."))
    ((null args) command)
    ((stringp command) (uiop:reduce/strcat (list command " " (str:unwords args))))
    ((consp command) (append command args))))

(defun %append-uiop-command-new (command &rest args)
  (cond
    ((null command) (error "Unable to append arguments to a null command."))
    ((consp command) (append command args))
    ((str:emptyp command) (error "Unable to append arguments to an empty command."))
    ((null args) command)
    ((stringp command) (uiop:reduce/strcat (list command " " (str:unwords args))))))

(%append-uiop-command-old "emacsclient" "-c") ;; => "emacsclient -c"
(%append-uiop-command-old "emacsclient -c") ;; => "emacsclient -c"

(%append-uiop-command-new "emacsclient" "-c") ;; => "emacsclient -c"
(%append-uiop-command-new "emacsclient -c") ;; => "emacsclient -c"
```

PR #2988, opened 2023-05-27
Allow webkit-specific schemes

shamazmazum:
Hi! After I had updated my WebKit to a newer version with PDF.js pdf files stopped downloading. This patch allows nyxt to show pdf files using PDF.js

### DIFF
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -2316,3 +2316,6 @@ See `make-buffer' for a description of the other arguments."
                                   nyxt/mode/user-script:renderer-user-script))))
 
 (setf nyxt::*renderer* (make-instance 'gtk-renderer))
+
+(defmethod browser-schemes append ((browser gtk-browser))
+  '("webkit" "webkit-pdfjs-viewer"))
--- a/source/urls.lisp
+++ b/source/urls.lisp
@@ -189,28 +189,36 @@ documentation."
 Public Suffix list, T otherwise."
   (sera:true (cl:ignore-errors (cl-tld:get-tld hostname))))
 
+(export-always 'browser-schemes)
+(defgeneric browser-schemes (browser)
+  (:method-combination append)
+  (:documentation "Return a list of schemes supported by a browser"))
+
+;; Set specifier to T because *BROWSER* can be bound to NIL
+(defmethod browser-schemes append ((browser t))
+  (let ((nyxt-schemes (append '("blob" "javascript") (alex:hash-table-keys *schemes*)))
+        ;; List of URI schemes: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
+        ;; Last updated 2020-08-26.
+        (iana-schemes
+         '("aaa" "aaas" "about" "acap" "acct" "cap" "cid" "coap" "coap+tcp" "coap+ws"
+           "coaps" "coaps+tcp" "coaps+ws" "crid" "data" "dav" "dict" "dns" "example" "file"
+           "ftp" "geo" "go" "gopher" "h323" "http" "https" "iax" "icap" "im" "imap" "info"
+           "ipp" "ipps" "iris" "iris.beep" "iris.lwz" "iris.xpc" "iris.xpcs" "jabber"
+           "ldap" "leaptofrogans" "mailto" "mid" "msrp" "msrps" "mtqp" "mupdate" "news"
+           "nfs" "ni" "nih" "nntp" "opaquelocktoken" "pkcs11" "pop" "pres" "reload" "rtsp"
+           "rtsps" "rtspu" "service" "session" "shttp" "sieve" "sip" "sips" "sms" "snmp"
+           "soap.beep" "soap.beeps" "stun" "stuns" "tag" "tel" "telnet" "tftp"
+           "thismessage" "tip" "tn3270" "turn" "turns" "tv" "urn" "vemmi" "vnc" "ws" "wss"
+           "xcon" "xcon-userid" "xmlrpc.beep" "xmlrpc.beeps" "xmpp" "z39.50r" "z39.50s"))
+        ;; https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xml
+        ;; TODO: Remove when https://github.com/lu4nx/cl-tld/issues/2 is fixed.
+        (special-use-schemes
+         '("example" "invalid" "local" "localhost" "onion" "test")))
+    (append nyxt-schemes iana-schemes special-use-schemes)))
+
 (export-always 'valid-scheme-p)
 (defun valid-scheme-p (scheme)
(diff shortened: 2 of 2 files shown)


shamazmazum:
This is why PDF.js is turned off by default:
20230528_09h38m18s_grim

WebKit always opens PDF.js' viewer with URL `webkit-pdfjs-viewer://pdfjs/web/viewer.html?file=`. I've googled and found that `?file=` part must contain a name of pdf document which is used when the document is saved to disk. With empty `file` parameter the name will always be `document.pdf`. To avoid possible confusion I switched PDF.js off by default, but if somebody knows a solution how this URL can be changed by means of Nyxt, I would like to hear it.

aartaka:
> This patch fixes downloading of pdf files and also allows a user to view pdf in the browser if `always-download-pdfs-p` slot in `context-buffer` is set to `nil` (it's `t` by default).

But then, there's a download option in PDF.js, so why have a fallback for plain downloads (besides being one keyshord/click less)?

aartaka:
Except for `always-download-pdfs-p`, this one looks good to me and I'm eager to merge it!

aadcg:
Nothing of interest to be shared from my side. @aartaka feel free to proceed as you find best.

shamazmazum:
@aartaka PDF.js always saves pdf files under the name `document.pdf` because WebKit does not pass a correct name to PDF.js. See my comment with a screenshot which illustrates the problem. Do you have any ideas how to fix this problem?

aartaka:
> @aartaka PDF.js always saves pdf files under the name `document.pdf` because WebKit does not pass a correct name to PDF.js. See my comment with a screenshot which illustrates the problem. Do you have any ideas how to fix this problem?

The only idea I'm able to come to is renaming the existing `document.pdf` (if any) in the save directory. Like we do with download names in `download-manager::ensure-unique-file`.

Would you be willing to hack on file uniqueness, or should I take care of it?

source/renderer/gtk.lisp
```diff
@@ -1904,8 +1904,25 @@ local anyways, and it's better to refresh it if a load was queried."
...
+;; https://stackoverflow.com/questions/47098206/pdf-js-downloading-as-document-pdf-instead-of-filename .
+;; Here we restore the original file name from an URL if a suggested
+;; file name looks suspicious.
+(sera:-> maybe-fix-pdfjs-filename (string quri:uri)
+         (values string &optional))
+(defun maybe-fix-pdfjs-filename (suggested-file-name uri)
+  (let ((pathname (pathname (quri:uri-path uri))))
+    (if (and (string= suggested-file-name "document.pdf")
+             (string= (pathname-type pathname) "pdf"))
+        (namestring (make-pathname :name (pathname-name pathname)
```
aartaka, line 1919: Could be worth using `uiop:native-namestring` instead, so that it's 100% portable across systems.

jmercouris:
I think this is ready now, no?

shamazmazum:
@aartaka Thanks for the tip with `uiop:native-namestring`! It's done.
@jmercouris Yes, it's ready.

jmercouris:
Thank you so much :-)

aadcg:
@jmercouris, I'm officially denying you the right to merge pull requests :p

Here you have done two mistakes. 

One, you merged it without rebasing against master (resulting in the git log mess you can in the screenshot below). Btw, I believe you have clicked the merge button from GitHub, and I don't understand why it doesn't rebase when merging (by default). I'm sure that there's a setting to do it. Nonetheless, the button should _never_ be used because it won't merge as [per our standard](https://github.com/atlas-engineer/nyxt/blob/master/documents/README.org#branch-management). 

Two, you didn't double check whether this PR meets all requirements. This has been merged without a changelog entry. 

As a punishment, you're required to push a changelog entry relative to this PR. Recall that you need to create a new changelog heading called "3.X.Y".

jmercouris:
Oof. I thought I was doing the right thing... I thought hard about whether to do a rebase and merge... I will no longer use the GitHub UI...

jmercouris:
Sorry about that :-\!

jmercouris:
So I'll move 3.x.y to 3.1.0 and then make a new 3.x.y? correct?

aadcg:
No worries! Nothing personal, as you very well know :)

> So I'll move 3.x.y to 3.1.0 and then make a new 3.x.y? correct?

I think you're correct! @Ambrevar might have a better understanding of how it used to be done, but it seems that we're reasoning correctly. 

So, concretely, I believe we need to do as per the diff below. I'd probably do it in 2 commits. One for renaming 3.X.Y, and another one to add an entry about this PR.

```diff
modified   source/changelog.lisp
@@ -694,7 +694,7 @@ define-version "3.0.0"
 color-picker support as an example application for this feature.")
      (:li "New " (:nxref :function 'match-port) " URL designator predicate for auto-rules."))))
 
-(define-version "3.X.Y"
+(define-version "3.1.0"
   (:ul
    (:li "REPL provides symbol suggestions with " (:code "tab") "/" (:code "C-i") ".")
    (:li "Re-enable global history.")
@@ -704,3 +704,7 @@ define-version "3.X.Y"
   (:ul
    (:li (:code "conservative-history-movement-p") " was deprecated in favor of "
         (:nxref :slot 'global-history-p :class-name 'buffer) ".")))
+
+(define-version "3.X.Y"
+  (:ul
+   (:li "New entry.")))

```

Ambrevar:
Yes, I think this "3.X.Y" naming convention works alright.

aadcg:
@jmercouris it's good that didn't complete the punishment task yet :p

See commit 37088a88d. When I released 3.1.0, I've revised the changelog. I should have sent that commit from the `3-series` branch to `master`. I did it now. Next time I'll do it on the spot to avoid confusion.

Please don't forget to add a changelog entry that sums up the changes on this PR. Thanks.

jmercouris:
Phew :-D

jmercouris:
My punishment is no pay for a year :-D

Ambrevar:
> I'd like to remind everyone that the smallest of changes require changelog entries. See commit [13119df](https://github.com/atlas-engineer/nyxt/commit/13119dfaf7af739e5a346a19e967d2d00632dea0). As you can see, I had to add most changelog entries before the actual release. CC @aartaka @Ambrevar.

Roger that!

aartaka:
> leverage PDF.js to open PDF files that are available locally?

The first idea that comes to my mind is checking whether the loaded resouce is a PDF and redirecting to webkit-pdf: scheme if it is. I'm doing something like that in my unpdf config: https://github.com/aartaka/nyxt-config/blob/master/unpdf.lisp

shamazmazum:
@aadcg Yes, it's possible. Firstly open a directory which contains the needed file with `C-x C-f` (with emacs scheme), then click on the file. Example:
20230909_06h59m10s_grim

It's not working when opening PDF directly.

PR #2821, opened 2023-02-28
IFrame hinting

aartaka:
# Description

This is a huge work-in-progress to finally allow iframes in hints. Only the foundations are implemented, like parsing iframes (done before this PR in cf2a7bf83f4b573835f6b2bbd11cdf151b77ed57 and adjacent commits) and allowing nested search for Nyxt DOM nodes.

Fixes #1526.

# Discussion

Don't review yet, it's neither complete nor working :D

# Checklist:
Everything in this checklist is required for each PR.  Please do not approve a PR that does not have all of these items.

- [ ] I have pulled from master before submitting this PR
- [ ] There are no merge conflicts.
- [ ] I've added the new dependencies as:
  - [ ] ASDF dependencies,
  - [ ] Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    git submodule add https://gitlab.common-lisp.net/nyxt/py-configparser _build/py-configparser
    ```
  - [ ] and Guix dependencies.
- [ ] My code follows the style guidelines for Common Lisp code. See:
  - [Norvig & Pitman's Tutorial on Good Lisp Programming Style (PDF)](https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf)
  - [Google Common Lisp Style Guide](https://google.github.io/styleguide/lispguide.xml)
- [ ] I have performed a self-review of my own code.
- [ ] My code has been reviewed by at least one peer.  (The peer review to approve a PR counts.  The reviewer must download and test the code.)
- [ ] Documentation:
  - [ ] All my code has docstrings and `:documentation`s written in the aforementioned style.  (It's OK to skip the docstring for really trivial parts.)
  - [ ] I have updated the existing documentation to match my changes.
  - [ ] I have commented my code in hard-to-understand areas.
  - [ ] I have updated the `changelog.lisp` with my changes if it's anything user-facing (new features, important bug fix, compatibility breakage).
  - [ ] I have added a `migration.lisp` entry for all compatibility-breaking changes.
  - [ ] (If this changes something about the features showcased on Nyxt website) I have these changes described in the new/existing article at Nyxt website or will notify one of maintainters to do so.
- [ ] Compilation and tests:
  - [ ] My changes generate no new warnings.
  - [ ] I have added tests that prove my fix is effective or that my feature works.  (If possible.)
  - [ ] New and existing unit tests pass locally with my changes.

Issue #1400, opened 2021-05-12
Improve hinting mechanism 


aartaka:
Our `hint-mode` can still be greatly improved.  We need to hint more elements and be better at filtering false positives.

Sources of inspiration:
- [ ] Aria roles ([spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles), [listing of the useful ones by @heiwiper](https://github.com/atlas-engineer/nyxt/issues/1400#issuecomment-1606288036))
- [ ] [list of selectors Tridactyl uses](https://github.com/tridactyl/tridactyl/blob/18a699ba3fb8a7fcee072a0feacc7f618fdcb296/src/lib/dom.ts#L583)
- [ ] [the method Vimium uses to determine hintable elements](https://github.com/philc/vimium/blob/bdf654aebe6f570f427c5f7bc9592cad86e642b5/content_scripts/link_hints.js#L1043) 
- [x] [the algorithm Saka Key uses to determine if an element is overlapped](https://github.com/lusakasa/saka-key/blob/master/src/modes/hints/client/findHints.js#L114)

Ambrevar:
Great investigation, thanks!

aartaka:
[ARIA roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) are something to concider too.

aartaka:
> I have checked the referenced function and It seems to include Shadow DOM

Shadow DOM unawareness is severely lacking in our hinting mechanism, and help would be much appreciated if one knows what they're doing :) See #1828 for more context and examples.

There's also #1526 which I'm trying to fix, and which would also benefit from some help in #2821.

> and also there is the part where it checks whether the corners of the elements are clickable. 

Yeah, this one's weird, not sure what to do with it :thinking:

heiwiper:
I have made progress on the issue related to Shadow DOMs, I did inspire from the pull request you have mentioned which tackles IFrame hinting. However, I'm currently stuck somewhere and I wanted to ask for advice.

I'm using almost the same function as the one you have added in [this commit](https://github.com/atlas-engineer/nyxt/pull/2821/commits/5b33d4c032141106f5f6c8f03b45339b60c65ee1) but for some reason when I inspect the values of `,node` I notice that the root node is going through the second `if` statement which generates the following external error:
```
WARNING:
   Error on separate prompter thread: There is no applicable method for the generic function
[...]

PR #3001, opened 2023-06-04
Add hinting support for Shadow DOM nested elements

### LINKED ISSUE #1828: Hints do not appear for elements within a Shadow DOM
**Describe the bug**
Hints will never be created for page elements which exist inside of a [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM), even if a hint would normally apply to that particular element otherwise.

1. Visit a web page that heavily utilizes Shadow DOM ([example](https://developer.servicenow.com/dev.do))
2. Activate hints mode
3. Observe missing hints (e.g.: No hint on the "Sign In" link for the previously linked example)

**Information**
- OS name+version: Linux 5.10.68-1-MANJARO
- Graphics card and driver: Radeon RX 570 Series / amdgpu 21.0.0-1
- Desktop environment / Window manager name+version: Xorg 1.20.13-3 / Xmonad 0.15
- How you installed Nyxt (Guix pack, package manager, build from source): Via AUR using yay package manager
```
Nyxt version: 2.2.1
Renderer version: GI-GTK
Operating system kernel: Linux 5.10.68-1-MANJARO
Lisp implementation: SBCL 2.1.7 (Dynamic space size: 1073741824)
Features: (:WEBKIT2 :WEBKIT2-2.32 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :GLIB-2-58 :GLIB :NYXT-2 :FSET-EXT-STRINGS :CUSTOM-HASH-TABLE-NATIVE :SWANK
 :PLUMP-UTF-32 :GLOBAL-VARS :DECLARE-TYPES :PARENSCRIPT
 :SBCL+SAFE-STANDARD-READTABLE :NAMED-READTABLES :LPARALLEL :21BIT-CHARS
 :CHUNGA :CLOSER-MOP :CL-PPCRE-UNICODE :FLEXI-STREAMS :CL-UNICODE :CL-PPCRE
 :CL-JSON-DOUBLE-FLOAT-IS-SUBSUMED :CL-JSON-SINGLE-FLOAT-IS-SUBSUMED
 :BORDEAUX-THREADS :LPARALLEL.WITH-CLTL2 :LPARALLEL.WITH-CAS
 :LPARALLEL.WITH-STEALING-SCHEDULER :SPLIT-SEQUENCE CHIPZ-SYSTEM:GRAY-STREAMS
 CFFI-FEATURES:FLAT-NAMESPACE CFFI-FEATURES:X86-64 CFFI-FEATURES:UNIX :CFFI
 CFFI-SYS::FLAT-NAMESPACE ALEXANDRIA::SEQUENCE-EMPTYP :FAST-IO-SV :FAST-IO
 :SBCL-USES-SB-ROTATE-BYTE :CL-JSON-CLOS :CL-JSON :THREAD-SUPPORT :ASDF3.3
 :ASDF3.2 :ASDF3.1 :ASDF3 :ASDF2 :ASDF :OS-UNIX :NON-BASE-CHARS-EXIST-P
 :ASDF-UNICODE :X86-64 :GENCGC :64-BIT :ANSI-CL :COMMON-LISP :ELF
 :IEEE-FLOATING-POINT :LINUX :LITTLE-ENDIAN :PACKAGE-LOCAL-NICKNAMES
 :SB-CORE-COMPRESSION :SB-LDB :SB-PACKAGE-LOCKS :SB-THREAD :SB-UNICODE :SBCL
 :UNIX)

ASDF version: 3.3.1
Critical dependencies: (/home/chao/.cache/yay/nyxt/src/nyxt/_build/cl-cffi-gtk/gtk/cl-cffi-gtk.asd
 /home/chao/.cache/yay/nyxt/src/nyxt/_build/cl-gobject-introspection/cl-gobject-introspection.asd
 /home/chao/.cache/yay/nyxt/src/nyxt/_build/cl-webkit/webkit2/cl-webkit2.asd)

```

heiwiper:
# Description

Shadow DOM nested elements are not currently handled by the hinting feature. This pull request aims to add support for these type of elements by expanding the search for hintable elements through Shadow DOM trees.

Fixes #1828

# Discussion

I tried as much as I can to introduce less code additions and changes but I had to use the new macros `rqs` and `rqsa` instead of `qs` and `qsa` respectively in `mode/hint.lisp` and `dom.lisp` (Please keep in mind that for the replacements I made in `dom.lisp` I mainly tested the function `click-element` and then added the same changes to the functions that are similar like `check-element`, I wasn't sure how to test the rest and I did not want to waste more time figuring that out so I thought I would add this working code and confirm the other functions are working as expected).

I would also like to precise that this pull request was mostly inspired by #2821 in the beginning due to the similarity between Shadow DOM and IFrame elements, but then I had to change most of the code due to some changes that introduced bugs like the one I have mentioned [here](https://github.com/atlas-engineer/nyxt/issues/1400#issuecomment-1546753803). One of the changes that I had to drop was adding multiple stylesheets such as in the [commit](https://github.com/atlas-engineer/nyxt/pull/2821/commits/eecb78ef50336d66bbf28aa19ddf6948e9b49900) from the pull request I linked earlier.

I also want to suggest moving all the hint elements into a single parent element outside the main DOM body, the current behavior is appending the hint overlay elements to the body which makes it very cluttered and we would have to remove them one by one which wouldn't be the case if we had all elements under the same custom nyxt node, this is the case for [saka-key](https://github.com/lusakasa/saka-key) and [LinkHints](https://github.com/lydell/LinkHints/) .

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:


heiwiper:
I have just tested the following interactive functions in `dom.lisp`, the checked elements are work as expected:
- [X] click-element
- [X] check-element
- [X] focus-select-element
- [X] toggle-details-element
- [X] select-option-element
- [ ] scroll-to-element (*I didn't know how to test this function*)
- [ ] set-caret-on-start (*Generates errors because `visual-mode` doesn't yet support Shadow DOMs, maybe an issue should be created for that. I will revert back the changes for this function*)

aartaka:
> I would also like to precise that this pull request was mostly inspired by #2821 in the beginning due to the similarity between Shadow DOM and IFrame elements, but then I had to change most of the code due to some changes that introduced bugs like the one I have mentioned [here](https://github.com/atlas-engineer/nyxt/issues/1400#issuecomment-1546753803). One of the changes that I had to drop was adding multiple stylesheets such as in the [commit](https://github.com/atlas-engineer/nyxt/pull/2821/commits/eecb78ef50336d66bbf28aa19ddf6948e9b49900) from the pull request I linked earlier.

Right, new stylesheeds are not needed for Shadow DOM, because it inherits from the main frame already!

> I also want to suggest moving all the hint elements into a single parent element outside the main DOM body, the current behavior is appending the hint overlay elements to the body which makes it very cluttered and we would have to remove them one by one which wouldn't be the case if we had all elements under the same custom nyxt node, this is the case for [saka-key](https://github.com/lusakasa/saka-key) and [LinkHints](https://github.com/lydell/LinkHints/) .

Yes, that'd be much cleaner. No hurry, though—this can be addressed separately.

aartaka:
> I have just tested the following interactive functions in `dom.lisp`, the checked elements are work as expected:
> ...
> * [ ]  scroll-to-element (_I didn't know how to test this function_)

It's only used in heading scrolling, so testing it on the headings inside Shadow DOM seems the only testcase :P 

It should also be used for `highlight-selected-hint` someday, I guess.

> * [ ]  set-caret-on-start (_Generates errors because `visual-mode` doesn't yet support Shadow DOMs, maybe an issue should be created for that. I will revert back the changes for this function_)

Right, it'd need quite some refactoring to work with Shadow DOM. Let's figure it out with hints first and leave lots of `TODO`s for visual mode :P

heiwiper:
> Right, new stylesheeds are not needed for Shadow DOM, because it inherits from the main frame already!

Shadow DOM elements don't inherit stylesheet by default, the reason why I dropped that method is because I didn't see a point in drawing hint overlays in every Shadow DOM root element (this could be also the case for IFrames) when I can just append them to the main DOM and they would all be visible and would use the same stylesheet.


> 
> Right, it'd need quite some refactoring to work with Shadow DOM. Let's figure it out with hints first and leave lots of `TODO`s for visual mode :P

It doesn't seem complicated now that I implemented the the feature for hints, I will be working on it after I fix the hint scope probably.

source/mode/hint.lisp
```diff
@@ -110,17 +110,12 @@ For instance, to include images:
       (setf (ps:@ element text-content) hint)
       element))
 
-  (let ((fragment (ps:chain document (create-document-fragment)))
-        (hints (ps:lisp (list 'quote hints)))
-        (i 0))
-    (dolist (element (nyxt/ps:qsa document "[nyxt-hintable]"))
-      (let ((hint (aref hints i)))
-        (ps:chain element (set-attribute "nyxt-hint" hint))
-        (ps:chain fragment (append-child (create-hint-overlay element hint)))
```
aartaka, line 119: Yeah, removing fragment was the sorry part of the IFrame PR, and it's also here. Maybe push them to some clobal variable and then push them all into a fragment?

source/parenscript-macro.lisp
```diff
@@ -31,6 +31,21 @@
...
   `(chain ,context (query-selector (lisp (format nil "[nyxt-identifier=\"~a\"]" ,id)))))
 
+(export-always 'rqs-nyxt-id)
+(defpsmacro rqs-nyxt-id (context id)
+  "context.querySelector() tailored for Nyxt IDs."
+  `(flet ((recursive-query-selector (context selector)
+            (let ((node (qs context selector)))
+              (if node
+                  node
+                  (let ((node-iterator (chain document (create-node-iterator context (@ *Node -e-l-e-m-e-n-t_-n-o-d-e))))
```
aartaka, line 41: Instead of this hyphenated thing, you can use `|ELEMENT_NODE|` so that it translates into the right JS spelling. See the uses of `|innerHTML|` in our code.

If I was doing this, I'd use uninterned `#:|ELEMENT_NODE|` so that we don't pollute the package with unneeded symbols. But that's not critical, unless we have colliding symbols, like `window` (that's why it's often referenced as `#:window`).

source/parenscript-macro.lisp
```diff
@@ -151,13 +166,15 @@
           (radius (parse-float (chain computed-style border-top-left-radius)))
           (rounded-border-offset (ceiling (* radius (- 1 (sin (/ pi 4))))))
           (offset (max coord-truncation-offset rounded-border-offset))
-          (el (chain document (element-from-point (+ (chain rect left) offset)
-                                                  (+ (chain rect top) offset)))))
+          (el (chain ,element (get-root-node ,element) (element-from-point (+ (chain rect left) offset)
```
aartaka, line 169: Oh wow, it was there all this time?! That'd simplify a lot of code for IFrame indexing...

source/parenscript-macro.lisp
```diff
@@ -167,3 +184,16 @@
           0)
        (= (chain window (get-computed-style ,element) "visibility")
           "hidden")))
+
+(export-always 'rqsa)
```
aartaka, line 188: Actually, why define new variations for the existing Parenscript macros? Maybe simply replace `qsa` with this `rqsa`? Or is there something I'm missing?

aartaka (approved):
I like it, it's extremely elegant! I've left some minor notes regarding style here.

Regarding the non-working commands/macros/functions: proceed as you think best and ping me when you consider it ready :)

What won't hurt is a set of links you've tested the implementation on, so that I can see how it works.

source/parenscript-macro.lisp
aadcg: I think that `rqsa` will perform worse for the cases that `qsa` currently covers.

aadcg:
Thanks @heiwiper!

Not much to add. I second the request for testing URLs.

heiwiper:
Thank's a lot for the review! I will finish testing the remaining  elements and review your comments as soon as possible.

In the meantime, you can test the feature on the link that was posted as an example on the issue linked to this PR (here https://developer.servicenow.com/dev.do).

heiwiper:
Thanks a lot for the review! I will finish testing the remaining  elements and review your comments as soon as possible.

In the meantime, you can test the feature on the link that was posted as an example on the issue linked to this PR (here https://developer.servicenow.com/dev.do).

heiwiper:
Thank's a lot for the review! I will finish testing the remaining  elements and review your comments as soon as possible.

In the meantime, you can test the feature on the link that was posted as an example on the issue linked to this PR (here https://developer.servicenow.com/dev.do).

source/parenscript-macro.lisp
aartaka: Right. Here's an idea for optimization: test whether there are any Shadow DOM nodes in the page, and call a regular `qsa` whene there's none.

Although I'm not sure how easy that'd be. Leaving it to your judgement, @heiwiper :)

heiwiper: > Right. Here's an idea for optimization: test whether there are any Shadow DOM nodes in the page, and call a regular `qsa` when there's none.

This would not be possible as far as I know, Shadow DOMs don't seem to have a straighforward API to select them from the main DOM, one would need to check every element and call the shadowRoot function to test whether it a Shadow root or not. I did some tests on pages containing Shadow DOMs and others that do not contain this type of elements using `performance.now()`. It is definitely lacking in terms of speed, but it doesn't seem to have a signigicant impact but that is also relative to every webpage and how much content it contains.

There's actually a workaround that we could implement, we could check if an element is a shadow root in the function `update-document-model` in `buffer.lisp` and we add an attribute `nyxt-shadow-root` to the shadow root elements. we already do some checks in that function and we modify elements by adding `nyxt-identifier`. This would allow us to query for Shadow DOM elements using that attribute and we would be able to implement the optimization that you have suggested.

source/parenscript-macro.lisp
heiwiper: I didn't come accross that in the reference docs, and now that you have mentioned it I thought there's finally some syntax that would allow me to use [spread dots](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) I get a deprecation warnings...

source/mode/hint.lisp
heiwiper: I didn't understand the way you wanted to push everything into a global variable since it's a mix of js and lisp code, but I tried to address this differently, check the following [commit](https://github.com/atlas-engineer/nyxt/pull/3001/commits/20bebe571f2e40882034dfdaf8102731c363bc3e). This might still have some performance issues since the elements are queried one after one instead of the previously used `qsa`. The reason behind switching to `qs-nyxt-id` is because the hintable elements are selected in lisp code and js code which is not only redundant but it also uses different approaches, one is based on `CLSS` and the other is based on the new macro `rqsa`, and even tho the results are the same, the elements are still not ordered in the same way. So there's also another solution which is sorting both lists of elements in lisp and js based on `nyxt-identifier`, though I'm not sure about the performance between both methods.

heiwiper:
> > * [ ]   scroll-to-element (_I didn't know how to test this function_)
> 
> It's only used in heading scrolling, so testing it on the headings inside Shadow DOM seems the only testcase :P

I'm getting a weird behaviour when trying to fix this, the following [expression](https://github.com/atlas-engineer/nyxt/blob/b7248d8c82ebd922d117dd11f715dbf14de51b4b/source/mode/document.lisp#L453) returns `nil` because the current heading element object doesn't seem to match the same object that is in the headings list, it is the same heading element but not the same object. Can you point me to the right direction to know why could this happen ?

source/parenscript-macro.lisp
heiwiper: Should be resolved in [077455881](https://github.com/atlas-engineer/nyxt/pull/3001/commits/07745588138911d90b37258dde794481cd4066ca) and [967facdef](https://github.com/atlas-engineer/nyxt/pull/3001/commits/967facdefe8a8bd28d8fd96e08c64ce4da2c17a1).

aartaka:
> I'm getting a weird behaviour when trying to fix this, the following [expression](https://github.com/atlas-engineer/nyxt/blob/b7248d8c82ebd922d117dd11f715dbf14de51b4b/source/mode/document.lisp#L453) returns `nil` because the current heading element object doesn't seem to match the same object that is in the headings list, it is the same heading element but not the same object. Can you point me to the right direction to know why could this happen ?

My guess would be that the underlying `document-model` has gotten refreshed and thus the new object for the heading is not identical (as per `eq`) to the old one. What you can do instead is to get their text of some other value related to them that's unlikely to change. Like:
``` lisp
(position (element (current-heading buffer))
          headings
          :key (compose #'nyxt/dom:body #'element)
          :test #'equal)
```

source/mode/hint.lisp
aartaka: Yes, unifying the data for the actual hinting sounds like a good idea!

source/parenscript-macro.lisp
aartaka: Yes, `update-document-model` (or the underlying `nyxt/dom::get-document-body-json`) is where I'd put it!

heiwiper: Should be resoved in the commits [efa13d037](https://github.com/atlas-engineer/nyxt/pull/3001/commits/efa13d037722e2bff6c97d03251421b0923edc35), [5fb2654d1](https://github.com/atlas-engineer/nyxt/pull/3001/commits/5fb2654d113647a4a028a973ca096a69a5877acf#diff-df4a73c41e08deb14ee31a54e9b8b99ce31b8e2482baa94de3983b199de9a2afR183) and [5b0f340bf](https://github.com/atlas-engineer/nyxt/pull/3001/commits/5b0f340bf6ee9fd7fbe266c4b81d31fb2818235d#diff-df4a73c41e08deb14ee31a54e9b8b99ce31b8e2482baa94de3983b199de9a2afR48).

heiwiper:
> My guess would be that the underlying `document-model` has gotten refreshed and thus the new object for the heading is not identical (as per `eq`) to the old one. What you can do instead is to get their text of some other value related to them that's unlikely to change. Like:
> 
> ```lisp
> (position (element (current-heading buffer))
>           headings
>           :key (compose #'nyxt/dom:body #'element)
>           :test #'equal)
> ```

This seems to solve the issue, please see the commit [4e751aa97](https://github.com/atlas-engineer/nyxt/pull/3001/commits/4e751aa97c0e3cf280c665199052beeeabd6f429).

The PR should be ready now for another review @aartaka. Please let me know if there's any other changes to add.

aartaka (approved):
It looks perfect to me, I've nothing to add.

So let's 
- Rebase over master.
- Squash commits (I believe squashing into one would be the most sensible thing).
- Fixup commit messages (no backtics—plaintext commit messages are more accessible and don't depend on the Git/GUI parsing Markdown.)
- Go through the PR checklist once again (documentation and changelogs, mostly).

source/mode/hint.lisp
```diff
@@ -112,14 +112,15 @@ For instance, to include images:
 
   (let ((fragment (ps:chain document (create-document-fragment)))
         (hints (ps:lisp (list 'quote hints)))
-        (i 0))
-    (dolist (element (nyxt/ps:qsa document "[nyxt-hintable]"))
-      (let ((hint (aref hints i)))
+        (nyxt-identifiers (ps:lisp (list 'quote nyxt-identifiers))))
```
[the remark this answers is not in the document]
heiwiper: - Should I update the doctstring for the function `nyxt:update-document-model` to mention the new attribute `nyxt-shadow-dom` ? 
- Should I add docstring for already existing functions that I have modified ?
-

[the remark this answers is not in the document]
heiwiper: > Note that you can also `(coerce nyxt-identifiers 'vector)`, although I'm not sure about the performance of such an approach stuck_out_tongue
> 
> Anyway, not critical.

I tried to use the same code as the one used for `hint` binding

heiwiper:
Hello! I would like to follow up on this pull request to ask whether there's anything else to be done on my side.

jmercouris:
Hey @heiwiper thanks for the ping, and I'm sorry for our negligence. I'll take a look at this on Monday. I believe it is ready. So I'll handle the conflicts and hopefully merge it as well!

jmercouris:
Yes, feel free André!

heiwiper:
Alright, I will elaborate on this tomorrow!

aartaka:
Yes, IFrame hinting is dropped and is totally unrelated to this PR.

PR #3015, opened 2023-06-14
Refactor zoom logic.

### LINKED ISSUE #2979: zoom-ratio-default not respected
I’ve set zoom-ratio-default for document-buffer, but the zoom ratio is more often than not set as 1.0 when entering a site.
I have to call reset-page-zoom for the zoom to be set to zoom-ratio-default

```lisp
  ((zoom-ratio-default 1.2)))
```

Ref https://discourse.atlas.engineer/t/zoom-ratio-default-doesnt-trigger-automatically/678

### LINKED ISSUE #3020: Simplify zooming API
We have five (!) slots related to zooming. I routinely confuse these, and wonder why all of them are here. I mean, yeah, it ensures the right zoom ratio and makes all the process real configurable, but is it worth the complexity? I think it's not.

I'm thinking that we should remove these five slots for just one of two:
- `zoom-ratio` (currently called `current-zoom-ratio`),
- and maybe `zoom-ratio-step`.

All the other ones are not really necessary:
- `zoom-ratio-min` is useless and restrictive. The real restriction here is the renderer and its zooming restrictions. No need to reproduce them on the Lisp side.
- Same for `zoom-ratio-max`—if one wants to scroll 10x in and renderer allows that, why should we stop them?
- `zoom-ratio-default`—this one is managed by the renderer, and synchronising renderer default and Nyxt default is not really worth it, I guess.
  - I myself often confuse `zoom-ratio-default` and `current-zoom-ratio`, because both are implied to Just Work ™️ on the page. But it seems that zoom-ratio-default is not even used where it should #2979. So it'll be easier to remove it. 

And then, we make a setter method for `zoom-ratio`, which would call the `ffi-buffer-zoom-level` directly:
```
(defmethod (setf zoom-ratio) (value (buffer document-buffer))
  ;; Because zooming to negative values is meaningless?
  (unless (not (plusp value))
    (setf (slot-value buffer 'zoom-ratio) value
          (ffi-buffer-zoom-level buffer) value)))
```

With that setter, `zoom-page` and `unzoom-page` etc. turn into one-liners:
```
(define-command zoom-page (&key (buffer (current-buffer)))
  "Zoom in the current page BUFFER."
  (incf (zoom-ratio buffer) (zoom-ratio-step buffer)))

(define-command unzoom-page (&key (buffer (current-buffer)))
  "Zoom out the current page in BUFFER."
  (decf (zoom-ratio buffer) (zoom-ratio-step buffer)))

(define-command reset-page-zoom (&key (buffer (current-buffer)) (ratio 1.0))
  "Reset the BUFFER zoom to the RATIO."
  (setf (current-zoom-ratio buffer) ratio))
```

Yes, we imply several thing here:
- Default zoom is 1.0.
- Zoom does not fall below 0.
- `ffi-buffer-zoom-level` is `setf`-able (it is) and is lenient to arbitrary floats (it is: I just set it to 30 and it zoomed in like a good machine.)

Leaving it as is? It kinda works, so why break it?

**Additional context**
#2979 and #3015.

hgluka:
# Description

Fixes a bug where `zoom-ratio-default` wasn't respected when loading a page and refactors the zoom logic to make it easier to understand and configure.

Fixes #2979.
Closes #3020.

# Discussion

I've tested it and it seems to be working properly

# Checklist:

- [x] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -380,19 +380,13 @@ scroll-right will scroll.")
    (current-zoom-ratio
     1.0
     :type float
-    :documentation "The current zoom relative to the default zoom.")
+    :reader t
+    :export t
+    :documentation "The current zoom ratio.")
    (zoom-ratio-step
     0.2
     :type float
     :documentation "The step size for zooming in and out.")
-   (zoom-ratio-min
-    0.2
-    :type float
-    :documentation "The minimum zoom ratio relative to the default.")
-   (zoom-ratio-max
-    5.0
-    :type float
-    :documentation "The maximum zoom ratio relative to the default.")
    (zoom-ratio-default
     1.0
     :type float
@@ -408,6 +402,11 @@ down."))
   (:export-predicate-name-p t)
   (:metaclass user-class)
   (:documentation "Buffers holding structured documents."))
+
+(defmethod (setf current-zoom-ratio) (value (buffer document-buffer))
+  (when (plusp value)
+    (setf (slot-value buffer 'current-zoom-ratio) value
+          (ffi-buffer-zoom-level buffer) value)))
 
 (define-class context-buffer (buffer)
   ((last-access
--- a/source/mode/document.lisp
+++ b/source/mode/document.lisp
@@ -266,8 +266,7 @@ Important pieces of functionality are:
   nil)
 
 (defmethod nyxt:on-signal-load-finished ((mode document-mode) url)
-  (reset-page-zoom :buffer (buffer mode)
-                   :ratio (current-zoom-ratio (buffer mode)))
+  (reset-page-zoom :buffer (buffer mode))
   url)
 
 (define-internal-page show-url-qrcode (&key url)
@@ -361,26 +360,18 @@ The amount scrolled is determined by the buffer's `horizontal-scroll-distance'."
   (ps-eval (ps:chain window (scroll-by 0 (- (* (ps:lisp (page-scroll-ratio (current-buffer)))
                                              (ps:@ window inner-height)))))))
 
-(defun ensure-zoom-ratio-range (zoom &optional (buffer (current-buffer)))
-  (let* ((ratio (funcall zoom (current-zoom-ratio buffer) (zoom-ratio-step buffer))))
-    (setf ratio (max ratio (zoom-ratio-min buffer)))
-    (setf ratio (min ratio (zoom-ratio-max buffer)))
-    (setf (current-zoom-ratio buffer) ratio)))
-
 (define-command zoom-page (&key (buffer (current-buffer)))
   "Zoom in the current page BUFFER."
-  (ensure-zoom-ratio-range #'+ buffer)
-  (setf (ffi-buffer-zoom-level buffer) (current-zoom-ratio buffer)))
+  (incf (current-zoom-ratio buffer) (zoom-ratio-step buffer)))
 
 (define-command unzoom-page (&key (buffer (current-buffer)))
   "Zoom out the current page in BUFFER."
-  (ensure-zoom-ratio-range #'- buffer)
-  (setf (ffi-buffer-zoom-level buffer) (current-zoom-ratio buffer)))
+  (decf (current-zoom-ratio buffer) (zoom-ratio-step buffer)))
 
 (define-command reset-page-zoom (&key (buffer (current-buffer))
                                       (ratio (zoom-ratio-default buffer)))
   "Reset the BUFFER zoom to the `zoom-ratio-default' or RATIO."
-  (setf (ffi-buffer-zoom-level buffer) (setf (current-zoom-ratio buffer) ratio)))
+  (setf (current-zoom-ratio buffer) ratio))
 
 (define-internal-page summarize-buffer (&key (summary-length 5) (id (id (current-buffer))))
     (:title "*Summary*")
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -719,3 +719,8 @@ color-picker support as an example application for this feature.")
    (:li "Output multiple values when evaluating Lisp expressions from "
         (:nxref :command 'nyxt:execute-command) ".")
    (:li "Bug fixes.")))
+
+(define-version "3.X.Y"
+  (:ul
+   (:li "Remove restrictions on zoom level in the form of "
+    (:code "zoom-ratio-min") " and " (:code "zoom-ratio-max") ".")))
--- a/source/help.lisp
+++ b/source/help.lisp
@@ -126,8 +126,8 @@ Note that some settings may require creating a new buffer to take effect.")
                    "Set default new buffer URL"))
      (:li (:button :class "button"
                    :onclick (ps:ps (nyxt/ps:lisp-eval
-                                    (:title "set-zoom-ration")
-                                    (nyxt::configure-slot 'current-zoom-ratio 'document-buffer)))
+                                    (:title "set-zoom-ratio")
+                                    (nyxt::configure-slot 'zoom-ratio-default 'document-buffer)))
                    "Set default zoom ratio"))
      (:li (:button :class "button"
                    :onclick (ps:ps (nyxt/ps:lisp-eval


aartaka (approved):
I'm still uneasy about how complex our zoom API is :( 

@hgluka, any chance you'd be interested in simplifying it? (I've specified it out in #3020)

That's a collateral thing, though. Fix this one thing I've commented, and we're good to merge!

EDIT: change -> chance

source/mode/document.lisp
```diff
@@ -267,7 +267,7 @@ Important pieces of functionality are:
 
 (defmethod nyxt:on-signal-load-finished ((mode document-mode) url)
   (reset-page-zoom :buffer (buffer mode)
-                   :ratio (current-zoom-ratio (buffer mode)))
+                   :ratio (zoom-ratio-default (buffer mode)))
```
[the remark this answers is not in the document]
hgluka: Ah, I did a search for `current-zoom-level`, but I missed it in help.lisp. I'll change that now as well.

source/buffer.lisp
```diff
@@ -377,22 +377,17 @@ buffer.")
     :type integer
     :documentation "Horizontal scroll distance. The distance scroll-left or
 scroll-right will scroll.")
-   (current-zoom-ratio
+   (zoom-ratio
```
aartaka, line 380: This breaks backwards-compatibility (even though it pleases my aesthetic feelings), so we have several options here:
- Retain this slot named `current-zoom-ratio`.
  - Maybe mark it as deprecated and rename in 4.0?
- Rename this slot, while adding a `current-zoom-ratio` no-op one, just to not break the contract.
- Rename this slot, add `current-zoom-ratio`, but move all the data from `current-zoom-ratio` into `zoom-ratio` in `customize-instance :after` or elsewhere it makes sense.

@aadcg, @hgluka, what would y'all prefer? I seem to lean to the first option, because this way we have a new API, an old name, and much less user config breakages (I commited a mistake by configuring zoom with `current-zoom-ratio` in my config, which seems to do rounds in our community 😄)

source/buffer.lisp
```diff
@@ -377,22 +377,17 @@ buffer.")
...
     1.0
     :type float
+    :writer nil                          ; We use a custom writer.
```
aartaka, line 383: `:reader t` alone should cut it. But yeah, I myself often get confused over it 😅

source/buffer.lisp
```diff
@@ -409,6 +404,11 @@ down."))
   (:metaclass user-class)
   (:documentation "Buffers holding structured documents."))
 
+(defmethod (setf zoom-ratio) (value (buffer document-buffer))
+  (unless (not (plusp value))
```
aartaka, line 408: Wait, did I suggest to write it this way? 😱 

Seems like `when (plusp value)` does the same with much less cognitive load 😆 

I mean:
- `when (not ...` = `unless`
- `unless (not ...` = `when`
- `unless (or a b ...` = `when (and (not a) (not b) ...`
- `unless (and a b ...` = `when (or (not a) (not b) ...`

Ahh, my Formal Logic classes and De Morgan's Law...

source/buffer.lisp
hgluka: Ah, good point. I also think the first option is best. There's no real need to change the name (other than aesthetics, as you said), especially if it breaks old configs. I didn't think far enough ahead :smile:

source/buffer.lisp
hgluka: Yeah, there's a slot defined in the same file that uses `:reader nil` and `:writer t`, so I thought both were necessary.

hgluka: In this case it's `:reader t` and `:writer nil`, but for example `document-model` in the same class uses a custom reader, so it's the other way around.

aartaka: > Also, it takes getting to the depths of defclass-star (in particular process-slot-definition) to understand how we're managing :reader, :writer and :accessor.

Not really, just a couple of `macroexpand-1` on the `define-class`...

source/buffer.lisp
aartaka: > Actually, why do we want to rename current-zoom-ratio to simply zoom-ratio? We have default-zoom-ratio, so having an additional current-zoom-ratio seems more sensible than simply zoom-ratio.

It's not `default-zoom-ratio`, it's `zoom-ratio-default`, which is quite weirdly named. So `current-zoom-ratio` and `zoom-ratio-default` are not consistent at all, while `zoom-ratio` and `zoom-ratio-default` kinda make sense 🤷

source/buffer.lisp
```diff
@@ -409,6 +403,11 @@ down."))
   (:metaclass user-class)
   (:documentation "Buffers holding structured documents."))
 
+(defmethod (setf current-zoom-ratio) (value (buffer document-buffer))
+  (when (plusp value)
+    (setf (slot-value buffer 'current-zoom-ratio) value
+          (ffi-buffer-zoom-level buffer) value)))
```
hgluka, line 409: One  thing I noticed is that the documentation for `current-zoom-ratio` says: "The current zoom **relative** to the default zoom."
This isn't correct, strictly speaking, and I don't think it was correct before I made the changes. I could fix the doc or change the code to match it by doing something like
```lisp
(setf (slot-value buffer 'current-zoom-ratio) value
      (ffi-buffer-zoom-level buffer) (* zoom-ratio-default value))
``` 
@aadcg @aartaka What do you guys think is the intended behavior?

hgluka: Yeah, it can be read that way, but I changed it anyway to make it more clear.

source/buffer.lisp
aartaka: That's a strategy too, yes.

source/mode/document.lisp
aartaka, line 270: Note that this line is not necessary, because `:ratio` argument is already set to `(zoom-ratio-default buffer)`.

source/buffer.lisp
```diff
@@ -380,19 +380,13 @@ scroll-right will scroll.")
    (current-zoom-ratio
     1.0
     :type float
-    :documentation "The current zoom relative to the default zoom.")
+    :reader t
+    :export t
```
aartaka, line 384: Does it not export the symbol if you omit the `:export t` part? It should export it due to `(:export-accessor-names-p t)` below, I believe.

aartaka (approved):
Looks good to me! Let's:
- Fix the small things I've marked.
- Add a changelog entry.
- Squash all the commits modifying the code (changelog commit can stay on its own or get squashed too).
- And merge it! 😃

source/buffer.lisp
hgluka: That's what I thought, as well, but no, without `:export t`, 
I get an error that `The function NYXT/MODE/DOCUMENT::CURRENT-ZOOM-RATIO is undefined.`

aartaka: Okay then, let it be 😮

hgluka:
> > Squash all the commits modifying the code
> 
> [Keeping in mind my comment above](https://github.com/atlas-engineer/nyxt/pull/3015#discussion_r1231351151).

Oh, I didn't end up changing the name. I left it as `current-zoom-ratio` for now. Should I do it?

aartaka:
I would add a comment and leave the names as is until 4.0 😛

hgluka:
Okay, I'll just add a `TODO` comment to rename it in 4.0, is that alright?

hgluka:
And one last question, just so I don't mess something up. In the changelog, should I add my entry to version 3.2.0 (that one is already released, right?) or define a new version?

hgluka:
In that case, I think it's done. Thanks for the help!

aartaka:
Merged with 82babe9a31d49fceaf985bb4924d4e5a0b2d1e3a!

@hgluka, I've only noticed it now that you've been using your master branch for this pull request. A better approach would be to create a separate branch and open the PR with it:
- It keeps your fork master branch clean.
- It allows opening multiple pull requests—a branch per PR.
- It makes Nyxt-side commit log cleaner, because instead of `Merge remote-tracking branch 'hgluka/master'` we'd have `Merge remote-tracking branch 'hgluka/refactor-zoom-ratio'` 😉

hgluka:
Ah, yeah. Sorry about that, this PR started pretty small, so I just did it on master. But I'll keep that in mind and use separate branches from now on. Thanks!

PR #3024, opened 2023-06-15
buffer(switch-buffer): Add a source to create a new buffer.

aartaka:
# Description

I've noticed it in Emacs, that, when you switch buffers via Helm, there's a small field under all the existing buffers, where you can input arbitrary name and create a new buffer with that name. In our case, name won't make sense, but a URL will. Thus, `switch-buffer` now has a `new-url-or-search-source` to create new buffers out of nothing!

@lansingthomas, neat, huh?

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [X] Documentation:
- [X] Compilation and tests:


Ambrevar:
I do not have a strong opinion about this change: I can see why some would find it useful, but I can also understand that this is a feature that not the majority would use.

> "Do one thing and do it well" is an amazing design motto. _It's just that things changed a bit after the C programming language._

I personally beg to differ that it's an amazing design motto, but opinions aside, Nyxt (as well as CL) are at a polar opposite of it! :p

jmercouris:
I could indeed see the addition of a new command for this. Fascinating idea for sure... I've not found myself needing it, but that doesn't mean others dont!

aartaka:
> I could indeed see the addition of a new command for this. Fascinating idea for sure... I've not found myself needing it, but that doesn't mean others dont!

Why new commands if a new source is enough? And how do you imagine this command to be described as? `switch-buffer-or-create-new`? 

We have an assymetry: `set-url` allows switching buffers, while `switch-buffer` lacks the ability to create new ones. If we create a new commands that's both switching and creating buffers, we're essentially creating a poor person's version or `set-url`.

What I'm suggesting is adding a source to `switch-buffer` exactly to have it powerful enough to be useful outside the boring and predictable buffer switching. Helm allows creating new buffers from inside the buffer switching, and I myself used that a dozen time. Because it's too easy not to use.

And then yes, it might be just me who might use it. But what are we losing? It's either brings no benefits or really benefits someone. No negative outcome (at least none that I can imagine). So, win-win?

jmercouris:
that's actually interesting that you say that, because long ago when we came up with the idea of sources, Pierre and I discussed the permutation problem of sources. Consider this:

`set-url-from-x`
and
`set-url-from-y`

instead, in our codebase we simply have `set-url`, where `x` and `y` are defined as sources.

This is all good if you like manually paginating through the sources to find what you want, but what if you *specifically* want to set the URL from a bookmark? or from a `x`? Well, we figured that there *should* be a way to automatically generate all of the permutations from `set-url` based on the set of all sources.

What do you think?

aartaka:
That I've thought about too, with recent manic idea: we can store sources inside command objects and allow derivative commands to inherit/modify/reconfigure the sources. And for `define-configuration` to modify commands' sources too. No need to provide lots of commands if we can provide a sensible set of default ones and a good way to configure them.

aartaka:
In the meantime, merging this?

Ambrevar:
OK with me.

aartaka:
😵

jmercouris:
it's fine, no worries :-)

Issue #3019, opened 2023-06-14
Save the permission choices per website


aartaka:
**Is your feature request related to a problem? Please describe.**
Most of the engines that we can bind to, including WebKit, have their APIs for permission requests. In case of WebKit, the API is quite simple and we already handle it quite well. One thing is missing, though: saving the user preference across restarts. We make every permission request to go through again and again, causing lots of prompts, which is annoying. Would be better if we saved chosen permissions somehow.

**Describe the solution you'd like**
Maybe create dummy modes in gtk.lisp and use auto-rules to save these per website? Seems like the most natural solution, but makes it quite renderer-dependent. And auto-rules don't handler non-existent/unparseable modes well enough. So some infra work needed here.

And then, we'd have to modify the `if-config` call too, to allow the "always"/"never" responses: #3018

**Describe alternatives you've considered**
Maybe some simple hash-table or another simple storage for permissions, as suggested in https://discourse.atlas.engineer/t/grant-deny-os-access-isnt-remembered-across-restarts/775/5 ? This would only work inside a single session, though.

Context: https://discourse.atlas.engineer/t/grant-deny-os-access-isnt-remembered-across-restarts/775

aadcg:
Due to this annoyance, 5f57d160c04b72231b01c4bd70b14bf0383fbd9a introduced `handle-permission-requests-p`.

At this stage, it makes sense to disable the feature by default. We may revisit this topic in the future.

PR #3036, opened 2023-06-20
Respect `zoom-ratio-default`.

### LINKED ISSUE #2979: zoom-ratio-default not respected

hgluka:
# Description

Make sure that `zoom-ratio-default` is respected on page load and when changing the default zoom ratio in common-settings.

# Discussion

This implements the part of #3015 that fixes #2979 and doesn't break backwards compatibility by removing `zoom-ratio-min` and `zoom-ratio-max`. 
That makes it eligible for the next patch or minor version.

cc @aadcg 

# Checklist:

- [ ] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:


aartaka (approved):

source/mode/document.lisp
```diff
@@ -266,8 +266,7 @@ Important pieces of functionality are:
   nil)
 
 (defmethod nyxt:on-signal-load-finished ((mode document-mode) url)
-  (reset-page-zoom :buffer (buffer mode)
-                   :ratio (current-zoom-ratio (buffer mode)))
+  (reset-page-zoom :buffer (buffer mode))
```
Ambrevar, line 269: Wait, does this mean that loading a page resets the zoom ratio to the default value?  This seems wrong (other browsers don't do this) and remove the main use for `current-zoom-ratio`.

hgluka: Ah, I think you're right. The explanation in [#2979 ](https://github.com/atlas-engineer/nyxt/issues/2979#issuecomment-1561761656) made sense to me, but I didn't bother to check. So we should only `reset-page-zoom` in a brand new buffer, right?

But also, other browsers (I checked Firefox and Chromium) have "site-specific" zoom as well. Once you zoom in on GitHub, for example, even if you open it separately in another tab, your zoom level will persist.

I don't zoom in often personally, so I'm not sure if we want the buffer-specific or site-specific option.

hgluka: I agree with your second point, but I just double-checked, and Firefox does have a site-specific zoom level. You can read more about it  [in their docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserSettings/zoomSiteSpecific). I tested it in Chromium as well and it works the same way.

I think that if I had to chose, I'd say the buffer-specific way is better. Site-specific zoom could be a feature, but I wouldn't make it the default. Also, implementing something like that may be related to #3019?

source/buffer.lisp
```diff
@@ -861,6 +861,9 @@ store them somewhere and `ffi-buffer-delete' them once done."))
   "Finalize buffer.
 When NO-HOOK-P is nil, run `*browser*'s `buffer-before-make-hook'.
 Return the created buffer."
+  (when (document-buffer-p buffer)
```
hgluka, line 864: @Ambrevar This makes it so `current-zoom-ratio` is set to `zoom-ratio-default` only for new buffers, but I'm not sure if this is the right place to put that code? Is there a better place?

source/mode/document.lisp
Ambrevar: Site-specific zoom sounds like something for auto-rules.
I think we used to have a zoom-mode, maybe it was a bit shortsighted to remove it :p

source/buffer.lisp
Ambrevar: Move it to `initialize-instance :after ((buffer document-buffer))`.

hgluka: I tried that before, and it doesn't work. My guess is that `initialize-instance :after ((buffer document-buffer))` is triggered after `nyxt:on-signal-load-finished` in mode/document.lisp, which is where/when the zoom level is set for the renderer. I suppose that part could be moved around as well.

Ambrevar: No, `initialize-instance :after ((buffer document-buffer))` is the very first thing that's triggered after a new document-buffer object is instantiated.

You can run `(make-instance 'document-buffer)` in a REPL and inspect the resulting object to observe the slot.

source/mode/document.lisp
jmercouris: I agree, another PR, another dream, but always Lisp.

source/buffer.lisp
Ambrevar: Oh, sorry you are right, it should be part of `customize-instance`.  Kind of got lost in what we were trying to achieve here :p

hgluka: Right. Thanks for the help!

hgluka:
I think this is ready to merge, unless I missed something. We'll need to fix the issue @Ambrevar found in `master` as well.

Ambrevar:
> We'll need to fix the issue @Ambrevar found in `master` as well.

Which issue?

Ambrevar:
Looks good to me otherwise.

Ambrevar:
Sure!

PR #3077, opened 2023-07-11
Import history from other browsers

### LINKED ISSUE #2435: Importing browsing history, bookmarks and downloads' history from Luakit
1. There is an option to import bookmarks from HTML. So that seems straightforward. However,
    1. when I tried creating a new bookmark in Nyxt (for testing), I couldn't create tags with hiphen(`-`) or underscore(`_`). Is it not supported?

2. In Luakit, history and session history (opened/closed tabs & windows) are kept in seperate files. But it looks like Nyxt keep them in a single file?

3. Downloads' history: where does Nyxt store this?

---

Hopping between browsers, syncing these things is the main problem! So I am thinking about using external software for these, so...

4. How difficult is it to use an external bookmarks manager like `buku`?
5. I would like to use an SQlite database for browsing history and downloads' history. How difficult is to do this?

hgluka:
# Description

This PR aims to add commands for importing history from other browsers that use sqlite for data storage.
Some browsers that are on the docket:
- [x] Firefox
- [x] Google Chrome
- [x] Chromium
- [x] Brave
- [x] Vivaldi
- [ ] ~~Safari~~ (Not yet)
- [ ] ~~Luakit~~ (Not yet)
- [ ] Others (feel free to add to this list if I'm missing something)

Resolves #2435

# Discussion

I'm not sure if reordering the components in `nyxt.asd` was the right way to go, but I needed to use `nyxt/mode/file-manager:file-source` in my function because the `places.sqlite` file is user-supplied. 
The idea is to make the command search all the usual places and find it automatically, so I might be able to revert the order in `nyxt.asd` when that's done.

As for the actual history side of things, this imports everything as separate entries with `htree:add-entry`, but maybe we want a different structure?

# Checklist:

- [x] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:


hgluka:
@aadcg I dismissed that idea at first because it seems way too small for its own mode, but if the end goal is 4 or 5 different commands (one for each browser) then maybe it's warranted.

aartaka:
>  I dismissed that idea at first because it seems way too small for its own mode, but if the end goal is 4 or 5 different commands (one for each browser) then maybe it's warranted.

We have modes with 0 commands, so you're safe here!

hgluka:
> For the sake of tidiness, I'd suggest creating a mode nonetheless.

Done! But this is my first time adding a mode, so maybe I missed something.
I could also add it as a default mode (maybe for `context-buffer`) so that the command doesn't have to be global, but I'm not sure if that's necessary/worth it/the right thing to do.

aartaka:
I hate to change the direction, but why not putting it into the `history-mode`?

We have `import-bookmarks-from-html` in `bookmark-mode`, so there's a precedent for import commands belonging to the mode they influence.

hgluka:
> I hate to change the direction, but why not putting it into the `history-mode`?
> 
> We have `import-bookmarks-from-html` in `bookmark-mode`, so there's a precedent for import commands belonging to the mode they influence.

That was my idea originally. The issue is (or could be) that the command uses `nyxt/mode/file-manager:file-source` and that isn't loaded until after `history-mode`. My solution was to just reorder the components in `nyxt.asd` and while it worked well enough, it might be dangerous later down the line. I assume that there's a reason why `history-mode` is one of the Core modes (and thus loaded earlier than the Prompter modes).

hgluka:
@aadcg @aartaka I realized that all the commands for different browsers would be very similar in structure, so I made a macro to define them. Is that safe to do? 
The only downside I see is that `describe-command` will show the non-expanded form in the Source section, which obscures things somewhat.

hgluka:
I see your point. I just ordered the parameters by when they appear in the macro.

My thinking right now is that the `prompt` parameter could be renamed to something like `file-path` and expanded into a bigger `cond` expression that tries to find the file in the usual places and only prompts the user if the file isn't found.

In that case, the SQL string should definitely come before it.

aartaka:
>> I hate to change the direction, but why not putting it into the `history-mode`?
>> 
>> We have `import-bookmarks-from-html` in `bookmark-mode`, so there's a precedent for import commands belonging to the mode they influence.
>
>That was my idea originally. The issue is (or could be) that the command uses `nyxt/mode/file-manager:file-source` and that isn't loaded until after `history-mode`. My solution was to just reorder the components in `nyxt.asd` and while it worked well enough, it might be dangerous later down the line. I assume that there's a reason why `history-mode` is one of the Core modes (and thus loaded earlier than the Prompter modes).

A bad, but nonetheless working solution could be offloading the symbol reading until file manager mode is available. Something like (read-from-string "nyxt/mode/file-manager:file-source") or (sym:resolve-symbol ...) should be enough.

This way, you don't have to reorder files, while having both convenient prompt sources and commands belonging to the right modes.

hgluka:
> A bad, but nonetheless working solution could be offloading the symbol reading until file manager mode is available. Something like (read-from-string "nyxt/mode/file-manager:file-source") or (sym:resolve-symbol ...) should be enough. This way, you don't have to reorder files, while having both convenient prompt sources and commands belonging to the right modes.

Perhaps someone else (maybe @Ambrevar or @jmercouris?) can chime in, because I don't really know the pros and cons of using `sym:resolve-symbol`. I see that it's used in a few places in the project, so maybe it's alright to use here as well?

On the other hand, I think that a separate mode could still be okay. Yes, all the commands would fit very well in 
`history-mode`, but they would also make up a distinct part of it, so at least it's not like we're splitting `history-mode` down the middle.

Anyway, I'm still working on the feature in the meantime, so feel free to discuss! I have a moderate amount of popcorn at the ready.

jmercouris:
Resolve sym is effectively a hack to defer loading something so the compiler is happy. I would put it in a separate mode for now.

hgluka:
So, it turns out that I don't have access to a Mac to test out and import from Safari. @jmercouris mentioned that it's not a priority. We can add that and Luakit in later.

As far as Chrome and Firefox go, there are a couple of questions regarding them.
1. In Firefox, it's not easy to determine the default profile is. There is a `profiles.ini` file, but I found some conflicting information about it. Some people say that the profile with `Default = 1` is the default one, but in my case it's the opposite! Although I have a different variable called `DefaultProfile` in there which is correct. Right now, I left it up to the user to choose a profile by dropping the file source into the "top-level" `.mozilla/firefox` directory that has all the default profile dirs.
2. Chrome is easier in that sense because the History file is at `.config/google-chrome/Default/History`. However, the question here is about other Chromium based browsers (and Chromium itself). They all have the same History file, but the location is not always the same. Should I make a separate command for a couple of them (like Google Chrome, Chromium, Brave, Vivaldi, etc.) or would that introduce too much clutter?

hgluka:
@aadcg Interesting. What does your `profiles.ini` look like? 
In any case, I did a bit (a lot) more reading on the issue and it seems overly complicated. Apparently, there should be one default profile per installation of Firefox (Release, Nightly, Beta, etc.) and finding out which is which from the `profiles.ini` or `installs.ini` files isn't easy.

Maybe we can work on that after this is merged, but my lazy solution (showing the folder with all the profiles in the prompt buffer) works for now. The only thing I'd add to it is some kind of `path-filter` for the `file-source`, so that the irrelevant folders aren't shown.

hgluka:
> I've tested by issuing `import-history-from-firefox`.
> 
> I think we must improve the UI, since these commands should be usable by the simplest of users. I don't think `path-filter` will work, because the predicate will filter given a certain input, i.e. within a folder. I would suggest creating a custom class that inherits from `prompter:source` or `nyxt/mode/file-manager:file-source` (probably the former). In the process, you'll get acquainted with one of the key Nyxt classes (those related to the prompt buffer).
> 
> The `constructor` would be set to the output of:
> 
> ```
> (uiop:run-program (list "find" (uiop:native-namestring "~/") "-name" "places.sqlite") :output t)
> ```
> 
> This is just a draft idea. What do you think?

Interesting! It didn't occur to me to just find the `places.sqlite` that way. This should work great for Firefox, although Chrome's file is called `History` and there are a bunch of other applications that use that.

> Ideally we should also catch errors such as the one below, and tell the user what to do via `echo`.
> 
> ```lisp
> Could not open sqlite3 database /home/aadcg/.mozilla/firefox/apez2u28.aadcg/
> Code CANTOPEN: no message.
>    [Condition of type SQLITE:SQLITE-ERROR]
> ```

Yeah, I still need to handle some errors and also echo messages at certain steps like "Import started", "Import finished", etc.

> After closing Firefox and attempting to import history again I was greeted with another error:
> 
> ```lisp
> The value
>   NIL
> is not of type
>   REAL
>    [Condition of type TYPE-ERROR]
> 
> Restarts:
>  0: [ABORT] abort thread (#<THREAD tid=22714 "Nyxt run-async command" RUNNING {1005CC0AA3}>)
> 
> Backtrace:
>  0: (LOCAL-TIME:UNIX-TO-TIMESTAMP NIL :NSEC 0)
>       Locals:
>         NSEC = 0
>         UNIX = NIL
>  1: ((:METHOD NYXT/MODE/HISTORY-MIGRATION:IMPORT-HISTORY-FROM-FIREFOX NIL)) [fast-method]
>       Locals:
>         NYXT/MODE/HISTORY-MIGRATION::CONN = #<DBD.SQLITE3:DBD-SQLITE3-CONNECTION {10095FDF93}>
>         NYXT/MODE/HISTORY-MIGRATION::HISTORY = #<HISTORY-TREE:HISTORY-TREE {1005248403}>
>  2: (NYXT::RUN-COMMAND #<NYXT:COMMAND NYXT/MODE/HISTORY-MIGRATION:IMPORT-HISTORY-FROM-FIREFOX (1)> NIL)
>       Locals:
>         ARGS = NIL
>         #:G7 = #<NYXT:COMMAND NYXT/MODE/HISTORY-MIGRATION:IMPORT-HISTORY-FROM-FIREFOX (1)>
>         OLD-%BUFFER = NIL
> ```
> 
> Have you seen it?

I've never run into this issue. I think it means that there's a row in your `moz_places` table with an empty `last_visit_date`. I'll do some testing and see if it's a bug on my end, though.

hgluka:
@aadcg I implemented the changes you mentioned. Let me know what you think.

If I didn't miss anything, I think this is pretty much ready to merge. I can squash it into a couple of commits + a changelog commit when someone approves it.

hgluka:
> Finally, users may be confused as to what needs to be done for the imported history to be available. Issuing `set-url` after importing didn't show the entries in the "Global history" source. If restarting Nyxt is required (what I did), it needs to be echoed to the user. If there's an alternative, the user needs to be notified about it as well.

That's weird, because for me the imported stuff shows up as entries in the "Global history" source immediately.

source/mode/history-migration.lisp
```diff
@@ -0,0 +1,87 @@
...
+                                                          :title (getf row :|title|)
+                                                          :implicit-visits (getf row :|visits|))
+                                           (local-time:unix-to-timestamp (getf row :|lastaccess|))))))))))))
+
+(define-history-import-command import-history-from-firefox
+  "Import history from Mozilla Firefox."
+  "SELECT url, title, last_visit_date/1000000 as lastaccess, visit_count as visits FROM moz_places WHERE last_visit_date not null"
+  (prompt1 :prompt "Choose Mozilla Firefox places.sqlite file."
+           :sources (make-instance 'external-browser-history-source :browser-pattern ".*/places\.sqlite$")))
+
```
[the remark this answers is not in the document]
hgluka: Yeah, I noticed that, but I wasn't sure how to go about it. I'll echo a message before running `find` as well.

As for the speed, the issue is probably the use of `-regex`, so maybe I'll try to figure out a way to get rid of it (it's easy for Firefox, but all the different Chromium-based browsers might be more complicated.

source/mode/history-migration.lisp
```diff
@@ -0,0 +1,87 @@
...
+               (loop for row = (dbi:fetch query)
+                     while row
+                     do (unless (url-empty-p (getf row :|url|))
+                          (htree:add-entry history
+                                           (make-instance 'history-entry
+                                                          :url (getf row :|url|)
+                                                          :title (getf row :|title|)
+                                                          :implicit-visits (getf row :|visits|))
+                                           (local-time:unix-to-timestamp (getf row :|lastaccess|))))))))))))
+
```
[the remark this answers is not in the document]
hgluka: Oh, right. I mentioned that earlier in the PR, but I forgot about it. I'll do it now.

source/mode/history-migration.lisp
```diff
@@ -0,0 +1,87 @@
...
+(define-history-import-command import-history-from-google-chrome
+  "Import history from Google Chrome."
+  "SELECT url, title, last_visit_time/1000000-11644473600 as lastaccess, visit_count as visits FROM urls"
+  (prompt1 :prompt "Choose Google Chrome History file."
+           :sources (make-instance 'external-browser-history-source :browser-pattern ".*/google-chrome/Default/History$")))
+
+(define-history-import-command import-history-from-chromium
+  "Import history from Chromium."
+  "SELECT url, title, last_visit_time/1000000-11644473600 as lastaccess, visit_count as visits FROM urls"
+  (prompt1 :prompt "Choose Chromium History file."
```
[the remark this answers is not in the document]
hgluka: Isn't `prompt1` correct here? 
If I use `prompt`, I need to do `(first (prompt ...))` for everything, because it returns a list with one element instead of the element itself.

source/mode/history-migration.lisp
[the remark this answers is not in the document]
hgluka: Alright, I'll mark this as resolved then.

source/mode/history-migration.lisp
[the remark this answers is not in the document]
jmercouris: Yes, I think prompt1 is correct here :-D

.gitmodules
```diff
@@ -601,3 +601,7 @@
 	path = _build/cl-colors2
 	url = https://notabug.org/cage/cl-colors2
 	shallow = false # Notabug does not advertize commits.
+[submodule "_build/cl-dbi"]
+	path = _build/cl-dbi
+	url = https://github.com/fukamachi/cl-dbi
+	shallow = true
```
Ambrevar, line 607: Why cl-dbi?  Are there databases other than SQLite you need to parse?
If not, just use cl-sqlite, since cl-dbi is too big for our purpose here.

source/mode/history-migration.lisp
```diff
@@ -0,0 +1,95 @@
...
+  ((prompter:name "History files")
+   (browser-pattern :accessor browser-pattern :initarg :browser-pattern)
+   (prompter:constructor
+    (lambda (source)
+      (echo "Searching for history file. This may take some time.")
+      (or (sera:filter #'str:non-empty-string-p
+                       (str:split #\newline
+                                  (nth-value 0
+                                             (uiop:run-program (list "find" (uiop:native-namestring "~/")
+                                                                     "-regex" (browser-pattern source))
```
Ambrevar, line 25: This is OK for a draft, but the final version should not rely on `find`.
Instead, use `uiop:collect-sub*directories` to search for files recursively.

Also don't use `(user-homedir-pathname)` to resolve the HOME dir.

source/mode/history-migration.lisp
```diff
@@ -0,0 +1,95 @@
...
+                                                               :output :string))))
+          (echo-warning "No history files found."))))))
+
+(defmacro define-history-import-command (name docstring sql-expr file-path)
```
Ambrevar, line 29: I would use keyword arguments to make the macro invocations more readable.

source/mode/history-migration.lisp
Ambrevar: And use a lambda instead of a regex, it's more powerful, readable, and efficient.

hgluka: > Also don't use `(user-homedir-pathname)` to resolve the HOME dir.

I'm assuming you mean "don't use `~/`" here, because that's what I'm using?
It seems like `(user-homedir-pathname)` is used in a couple of places in Nyxt as well as to resolve HOME in `nfiles`.

.gitmodules
hgluka: Good point! We don't need the whole thing, just the SQLite part. And cl-dbi even uses cl-sqlite for that.

source/mode/history-migration.lisp
Ambrevar: Oops, sorry, yes, I meant "don't use `~/`" :)

hgluka: I'm having some trouble with `collect-sub*directories`. I tried this, for example:
```lisp
(let ((results '()))
  (uiop:collect-sub*directories
           (uiop:ensure-directory-pathname (user-homedir-pathname))
           (constantly t)
           (constantly t)
           (lambda (subdir)
             (when (uiop:file-exists-p (nfiles:join subdir "places.sqlite"))
               (push (nfiles:join subdir "places.sqlite") results)))))
```
And it takes a very long time to run (more than 10 minutes). 
Now, my home directory is huge, so restricting the search to `~/.mozilla` should be a lot faster (although it might exclude some possible locations).

But I still feel like I must be missing something, since `find` takes a couple of seconds at most. Any idea why this would happen?

hgluka: Ah, I figured out that the issue is following symlinks (that happen to form a loop). `find` apparently detects those loops, so I either need to do the same or just avoid all symlinks outright.

hgluka:
@Ambrevar @aadcg I managed to get the `uiop:collect-sub*directories` based file search to run in about 20 seconds on my PC. GNU find is still a lot faster, though, and then there's locate which is even faster than that.

jmercouris:
That's a huge performance penalty. Maybe use UIOP as a fallback when find is not available or something like that. We'll talk about it during our call today :-)

.gitmodules
```diff
@@ -599,4 +599,10 @@
 [submodule "_build/cl-json"]
 	path = _build/cl-json
 	url = https://github.com/sharplispers/cl-json
+[submodule "_build/cl-dbi"]
```
jmercouris, line 602: remove plox

jmercouris:
Please let me know when you have squashed/rebased and are ready to merge this in!

hgluka:
@jmercouris Well, after our talk yesterday, I think it's ready to merge now.
I did a rebase, although I didn't squash it all into one commit. I wanted to leave the ones where I add a dependency separate.

source/mode/history-migration.lisp
```diff
@@ -0,0 +1,115 @@
...
+(define-history-import-command import-history-from-brave
+  "Import history from Brave."
+  :sql-query "SELECT url, title, last_visit_time/1000000-11644473600, visit_count FROM urls"
+  :file-path (prompt1 :prompt "Choose Brave History file"
+                      :sources (make-instance 'external-browser-history-source
+                                              :browser-lambda (lambda (subdir)
+                                                                (when (and (string= "Brave-Browser"
+                                                                                    (nfiles:basename (nfiles:parent subdir)))
+                                                                           (uiop:file-exists-p (uiop:merge-pathnames* "History" subdir)))
+                                                                  (uiop:merge-pathnames* "History" subdir))))))
```
Ambrevar, line 104: Same.

source/mode/history-migration.lisp
```diff
@@ -0,0 +1,115 @@
...
+(define-history-import-command import-history-from-vivaldi
+  "Import history from Vivaldi."
+  :sql-query "SELECT url, title, last_visit_time/1000000-11644473600, visit_count FROM urls"
+  :file-path (prompt1 :prompt "Choose Vivaldi History file"
+                      :sources (make-instance 'external-browser-history-source
+                                              :browser-lambda (lambda (subdir)
+                                                                (when (and (str:starts-with-p "vivaldi"
+                                                                                              (nfiles:basename (nfiles:parent subdir)))
+                                                                           (uiop:file-exists-p (uiop:merge-pathnames* "History" subdir)))
+                                                                  (uiop:merge-pathnames* "History" subdir))))))
```
Ambrevar, line 115: Same.

Ambrevar (commented):
You did a great job at switching to `uiop:collect-sub*directories`, thanks!

Can you open a new pull request to address my review?

jmercouris:
I don't think it's important to address those points now. I've assigned Luka higher priorities right now. Those changes are easy enough to make, but i don't think the context switch is good.

hgluka:
I don't mind context switching every once in a while! Plus, there aren't any big changes to make. I'll do this when I have the time (and need a break from looking at `common-settings`).

source/mode/history-migration.lisp
```diff
@@ -0,0 +1,115 @@
...
+   (browser-lambda :accessor browser-lambda :initarg :browser-lambda)
+   (prompter:constructor
+    (lambda (source)
+      (echo "Searching for history file. This may take some time.")
+      (or (let ((results '()))
+            (uiop:collect-sub*directories
+             (user-homedir-pathname)
+             (constantly t)
+             (lambda (subdir)
+               (equal (iolib/os:file-kind subdir) :directory))
```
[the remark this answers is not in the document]
hgluka: This is for avoiding symbolic links, since `uiop:directory-pathname-p` doesn't detect them. 
I could change it to `(not (equal (iolib/os:file-kind subdir) :symbolic-link)))` to make that clear.

source/mode/history-migration.lisp
Ambrevar: Wait, if it loops, then it should not terminate, how come it took "10 minutes"?

source/mode/history-migration.lisp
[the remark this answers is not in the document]
Ambrevar: Still, no need for `iolib` just for this.  You can use `uiop:truename*`, or `uiop:resolve-symlinks`.  To avoid cycling, you can push each directory to a hash-table and check if the next subdir was visited already.

Ambrevar:
Most of it are trivial changes.  The main significant change is the home-dir traversal, still, it's an easy fix.

Issue #3096, opened 2023-07-25
Typical Emacs binding in the minibuffer


mmahmoudian:
**Is your feature request related to a problem? Please describe.**
There are some keybindings when dealing with minibuffer that make users' lives much easier and have been generally use also in Emacs and also other software. Some of them are missing from Nyxt at the moment (version 4):

1. When the mini buffer is invoked (e.g switching buffer, killing buffers, allowing website for tracking, hint list, ...) user need to navigate between the options and items in the list. At the moment user have to use arrow keys to move up and down
2. To close the minibuffer, user have to only use the <kbd>Esc</kbd>

**Describe the solution you'd like**
1. Use C-n and C-p for navigating between options in the minibuffer as alternative to arrow up and down
2. use C-g as alternative to <kbd>Esc</kbd> to close the minibuffer

**Describe alternatives you've considered**
I cannot think of alternative ways. If I come up with any, I will share.

**Additional context**

[Peek 2023-07-25 11-38.webm](https://github.com/atlas-engineer/nyxt/assets/390889/b5e1b104-0008-4931-a119-23cbff9b2b8b)

mmahmoudian:
@aadcg Thanks for the response, but as you can clearly see from the video I attached, I have the emacs mode enabled (I can scroll through the page by C-n and C-p)

Also this is my current config:

```lisp
(define-configuration web-buffer
  ((default-modes (pushnew 'nyxt/mode/emacs:emacs-mode %slot-value%))))

```

If these should already work in Emacs mode, then this is a bug, if not, it is a feature request.

aartaka:
> That should be the default behavior, I agree with you.

I remember some VI users protesting against enabling the same kescheme mode in buffer and prompt buffer. Because prompt buffer is not really command-friendly, and it only works effectively in CUA/Emacs keyscheme.

mmahmoudian:
@aartaka thanks for the reply

> Because prompt buffer is not really command-friendly, and it only works effectively in CUA/Emacs keyscheme.

Do you happen to have any examples for this?

odanoburu:
I seem to have the same problem as the OP, but I do have emacs-mode turned on my prompt buffer and elsewhere:

```lisp
(define-configuration (web-buffer prompt-buffer panel-buffer
                       nyxt/mode/editor:editor-buffer)
  ((default-modes (pushnew 'nyxt/mode/emacs:emacs-mode %slot-value%))))
```

[...]

PR #3097, opened 2023-07-25
Fix keyscheme common settings

aadcg:
# Description

A hot fix solution until issues #2351 and #2296 are solved since the current state is hurting our users (see #3096).

# Discussion

I believe that `(define-configuration buffer ...)` is superior than `(define-configuration (web-buffer prompt-buffer) ...)` since the keyscheme should be propagated to all classes (including user defined ones) inheriting from `buffer`.

# Checklist:

- [x] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -814,7 +814,9 @@ buffers.")
    (:li "Add commands for importing history from other browsers.
 Currently, the supported browsers are Firefox, Google Chrome,
 Chromium, Brave and Vivaldi.")
-   (:li "Add Public Sans font and set it as the default.")))
+   (:li "Add Public Sans font and set it as the default.")
+   (:li "Fix keyscheme configuration via "
+          (:a :href (nyxt-url 'common-settings) "Common Settings Menu") ".")))
 
 (define-version "4-pre-release-1"
   (:li "When on pre-release, push " (:code "X-pre-release")


jmercouris (approved):
I even got tripped up on this, like the other user, I was wondering why my Emacs bindings didn't work!

Ambrevar:
This issue was possibly introduced with e5c8920d2a1284ed0538ecc3738abcebcae77e51.

@aartaka Why did you switch from `input-buffer` to `web-buffer`?

@aadcg I suggest to use `input-buffer` in this hot fix.

aartaka:
Leaving `input-buffer`/`modable-buffer`/`web-buffer` bug aside, I'm still cautious about imposing all the keyschemes on all the "input" buffers. Regular buffers and prompt buffers (I'm setting panel buffers aside for now) are different in modality, and forcing the same kescheme on both is not always suitable.

Instead of the direction this PR takes, I'd suggest splitting the web/prompt buffer configuration into two separate sections. So that it's:
``` md
# Regular buffer keyscheme (keyboard shortcuts)
- CUA
- Emacs
- VI

# Prompt buffer (input menu) keyscheme
- CUA
- Emacs
- VI command
- VI insert
```

Notice that VI insert is the `prompt-buffer`-specific option—it's been requested in the past, and, while mostly fixed with input focus mode switching, might still prove useful to force when in prompt buffer.

-----------------------------------------------------------------------

Having both web and prompt buffers to have the same keyscheme takes away flexibility and can be harmful in some corner-cases (I myself want to switch to VI bindings for regular buffers, but I'll never surrender Emacs mode in the prompt buffer, because Emacs input scheme is a perfect match for a mode-less input area with instant feedback).

On the opposite, having separate prompt buffer configuration makes the behavior more flexible and makes it clear that there are two really different modalities one should consider separately. We lose nothing (except the screen space another `<select>` might take), while gaining a more explicit interface and highlighting modality difference. Win-win, in other words.

Ambrevar:
I agree with @aartaka and @mmahmoudian .

jmercouris:
I don't care, just make it obvious for the user what is going to happen! :-D

aartaka:
> I'm not convinced by the argument that proposes separating setting the keyscheme for `web-buffer` and `prompt-buffer`. From the user's perspective, using technical jargon such as prompt buffer in the `common-settings` (a page where they may have just landed after installing Nyxt) is not wise.

This is a matter of language used in `common-settings`, not the matter of `web-buffer`/`prompt-buffer` separation. Language can be easily fixed, so this is not much of a hindrance.

> From a technical perspective, the separation isn't exhaustive as well. Think of `editor-buffer` and note that it doesn't inherit from either `web-buffer` of `prompt-buffer`. Users will be confused for no good reason when starting Nyxt with the config below. Notice that `emacs-mode` won't be enabled in `editor-buffer`s. It breaks the principle that @jmercouris raised above about managing the user's expectations.

Indeed, we've forgotten the `editor-buffer`. We might include it too, for the sake of completeness!

> ```lisp
#| ... |#
> (define-configuration buffer
>   ((default-modes (pushnew 'nyxt/mode/editor:plaintext-editor-mode %slot-value%)))
> ```

Did you mean `editor-buffer` here? I'm confused about what idea you're trying to convey by this piece of code. Can you expand?

> > (@aartaka) emacs-mode/vi-mode is a keyscheme-mode, and, by extension, a mode. Which means it only works in modable-buffer.
> 
> The observation above is sensible, and motivates commit [...] I.e. the first sexp is seperior than the second to achieve the goal of propagating the keyscheme.
> 
> ```lisp
> (define-configuration (input-buffer modable-buffer)
>   ((default-modes (pushnew 'nyxt/mode/emacs:emacs-mode %slot-value%))))
> ```

But wait, this enables Emacs mode for:
- `input-buffer`s (including `status-buffer`),
- `modable-buffer`s,
- and the intersection of these.

I guess that you meant is just the intersection. For that, you need to target a class that is an intersection of `input-buffer` and `modabe-buffer`. Which `web-buffer`, `prompt-buffer`, `editor-buffer`, and `panel-buffer` (now that we talk of completeness) are.

aartaka:
> Today that is correct. We just forgot about `editor-buffer`, which shows that this is not robust.

Not including the things we're not sure/aware/remembering about is more robust than including the things we're not sure/aware/remembering about. It's more dangerous to implicitly rely on all the entities in a set—including forgotten ones—than relying on a small yet safe subset of these. So not including forgotten `editor-buffer` is more robust than including the forgotten `editor-buffer` and half a dozen eldritch horrors lurking in the shadow of `input-buffer` umbrella.

> While I agree that it's odd to add a keyscheme mode to an entity like `status-buffer`, in the sense that it's a no-op, it does nothing harmful. 

Once we add one/N more `input-buffer` subclass for which keyscheme is not a no-op, we're at risk of breaking things the exact moment we add this subclass.

> I still find the first configuration to be more robust than the second, in the snippet below:
> 
> ```lisp
> (define-configuration input-buffer
>   ((default-modes (pushnew 'nyxt/mode/emacs:emacs-mode %slot-value%))))
> 
> (define-configuration (web-buffer prompt-buffer panel-buffer editor-buffer)
>   ((default-modes (pushnew 'nyxt/mode/emacs:emacs-mode %slot-value%))))
> ```

I find the second one more robust, because:
- It clearly lists the classes it influences.
  - Which means that if we add other `input-buffer` subclasses, then it won't break in unexpected ways. 
- It is easier to modify to toggle the keyscheme per class.
  - When one opens their config file, which one looks more intimidating—abstract `input-buffer` or obvious (given at least half an hour of Nyxt use) `web-buffer` etc.?

In the context of this PR—with most of the team agreeing that `web-buffer`/`prompt-buffer`/whatever difference should at least be highlighted, at most separated on the level of UI—the second form is superior too, because the mapping from UI to configuration is extremely clear. You configure keyscheme for `web-buffer`—you see the configuration form for it.

source/help.lisp
```diff
@@ -60,24 +60,27 @@ to the next."
...
+                                       panel-buffer nyxt/mode/editor:editor-buffer)
                  ((default-modes (remove-if (lambda (m)
                                               (find (symbol-name (name m))
                                                     '("EMACS-MODE" "VI-NORMAL-MODE" "VI-INSERT-MODE")))
                                   %slot-value%))))))
-      '((emacs "Use Emacs")
+      '((emacs "Emacs")
         (nyxt::auto-configure
-         :form '(define-configuration web-buffer
+         :form '(define-configuration (web-buffer prompt-buffer
```
aartaka, line 77: Nut wait, that doesn't solve the problem of `prompt-buffer`/`web-buffer` discrepancy yet 😜 Maybe:
- An additional `prompt buffer`-related `:nselect`?
- Or at least a comment explaining that the configuration has this totality?

-------------------------------------------------------------------

In any way, it would also be nice to split the configuration in two:
- `web-buffer`, `panel-buffer`, and `editor-buffer`.
- And `prompt-buffer`.


So that it's easier for one to edit the configuration.

----------------------------------------------------------------------

Yet another detail: `define-configuration` can have a docstring, and it wouldn't hurt adding one here 😄

aartaka: The `prompt-buffer` vs. anything is the minimum actionable thing, everthing else is optional.

aartaka: > Well, if there's a single dropdown isn't it rather obvious?

Given the uniqueness of prompt buffer as an interface, it might be unclear whether it also respects this configuration.

> The point of auto configuration is that users set options via the UI. 

So they can also choose prompt buffer setting via UI.

> Black belts can craft their own configs.

Making their (and first-time users transitioning though the Nyxt karate dans) work slightly easier is a net-positive change, especially given how little work it requires. And making Nyxt more approachable (to everyone) is a central goal, as I understood.

> But we digress again from the focus of this PR.

Based on the linked issue (#3096), the proper fix is also disambiguating betwen prompt and web buffers (which was the initial confusion in the issue). So making the configuration more global may solve some of the problems (propagation of Emacs/VI bindings to both web and prompt buffer), but it consts us another problem ("why is `web-buffer` configuration doesn't influence `prompt-buffer`" i.e. disambiguation of the two).

jmercouris: Two problems.

1. This doesn't do what the user expects
2. The user wants more fine grained control

this PR solves point 1. Point 2 is a separate one, and probably not suited for Common Settings. A single user does not constitute a pattern. We shouldn't make the experience of everyone worse for the benefit of a single user. As this PR stands right now, it is ready.

jmercouris (approved):

source/help.lisp
jmercouris: Let's agree to disagree. Reminder: this is also not related to your main tasks.

aartaka:
> Reminder: this is also not related to your main tasks.

😰

Ambrevar:
I agree (;p) that it is not a very big deal, and that using different bindings for different things can easily be done outside of Common Settings for now.

It's also not hard to add a second `<select>`, but this can be done later when the Common Settings are more polished, and feature and "advanced" section for instance.

jmercouris:
OK, will do.

jmercouris:
Sorry about that, thank you for merging.

PR #3098, opened 2023-07-26
Updated instructions for openSUSE

### LINKED ISSUE #3043: Should nyxt dlopen versioned libfixposix and libwebkit2gtk rather than .so ?
Currently, the `nyxt` binary dynamically load (`dlopen`) .so files for `libfixposix` and `libwebkit2gtk` on startup.

For example, with my openSUSE RPM package it will dlopen `libfixposix.so` and `libwebkit2gtk-4.1.so`.

The problem is that at least on openSUSE (and I suppose Debian), these .so files are in the devel packages of `libfixposix` and `libwebkit2gtk-4.1` while the actual libs they point to are in their respective non-devel packages.

So for the nyxt RPM package to have these 2 required .so files, there are 2 solutions:

- make the package depends on the  devel packages of libfixposix and libwebkit2gtk at runtime. This is not the best idea at it will pull a ton of unnecessary files (especially for webkit) and adding a runtime dependency on a devel package not looking right.
- implement a hack in the RPM .spec file to create the 2 .so links pointing to the correct libs

For the openSUSE package, I opted for the second solution (see [spec file](https://build.opensuse.org/package/view_file/network/nyxt/nyxt.spec?expand=1)) resulting in the script creating the 2 .so files in `/usr/libexec/nyxt`:

```
/usr/libexec/nyxt> ls -l
total 8
lrwxrwxrwx 1 root root 28 Jun 19 18:13 libfixposix.so -> ../../lib64/libfixposix.so.4
lrwxrwxrwx 1 root root 34 Jun 19 18:13 libwebkit2gtk-4.1.so -> ../../lib64/libwebkit2gtk-4.1.so.0
```

And a wrapper shell script to launch nyxt with LD_LIBRARY_PATH pointing to directory above:

```
#!/bin/sh
LD_LIBRARY_PATH=/usr/libexec/nyxt exec -a nyxt nyxt.bin "$@"
```

I tried to get this package included in the official openSUSE Tumbleweed repo but it was refused due to how nyxt handled these .so which was deemed "wrong". The argument of the openSUSE devs is that no program should dynamically load system installed .so as there could be versioning problems and that the only valid case for loading .so is for a program's private plugins.
Long story short, it should not load system installed .so directly but instead the versioned files. So in that case it should dlopen `libfixposix.so.4` and `libwebkit2gtk-4.1.so.0`. The appended number suffix for each lib is its major lib version that the binary was built upon.

It would be great if the nyxt build could be modified to accommodate that, which should also make the wrapper hack above not needed. Not sure how it is easy to do, but it would have at build time to figure out the major version of the libs and make sure that whatever is calling dlopen make use of it.

bubbleguuum:
# Description

Updated instructions for openSUSE

Fixes #3043

### DIFF
--- a/README.org
+++ b/README.org
@@ -82,22 +82,14 @@ the following platforms:
 - [[https://nixos.org/nix/][Nix]]: Install with =nix-env --install nyxt=.
 - [[https://guix.gnu.org][Guix]]: Install with =guix install nyxt=.
 - [[https://voidlinux.org/][Void]]: Install with =xbps-install nyxt=.
-- [[https://www.opensuse.org/][OpenSuse]] (user-maintained package of stable and latest `git master` build):
-  - Tumbleweed:
-
-    #+begin_src sh
-    zypper ar https://download.opensuse.org/repositories/home:/bobbie424242:/nyxt/openSUSE_Tumbleweed/ nyxt
-    #+end_src
-
-  - LEAP:
-
-    #+begin_src sh
-    zypper ar https://download.opensuse.org/repositories/home:/bobbie424242:/nyxt/15.4/ nyxt
-    #+end_src
-
-  Then install with =zypper in nyxt= for the latest stable release or =zypper in
-  nyxt-git= for the latest `git master` build. Both versions cannot be installed
-  at the same time.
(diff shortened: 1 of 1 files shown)


Ambrevar:
Looks good to me, assuming `nyxt-git` does not exist anymore in Suse.

Issue #3099, opened 2023-07-26
Fresh icons for the status buffer


lansingthomas:
**Purpose**
visual communication

**What is your proposed change?**
use these fresh svgs in the prompt buffer
1. - [ ] replace the existing icons with these SVGs
2. - [ ] ensure that their color is `text` (#0C0C0D in light theme, #EDEDED in dark theme)

**Rationale**
looks more Nyxty
these do not rely on user system fonts or whatnot

**Constraints**
please make good choices with the spacing and alignment

**Additional context/User story**

**Screenshots/Mock ups**
the vector graphics:
![lambda clear bg](https://github.com/atlas-engineer/nyxt/assets/6666079/90014ae3-b7bd-4450-b3af-197aed1109f1)
![reload clear bg](https://github.com/atlas-engineer/nyxt/assets/6666079/352ea902-ec7c-4d30-8c06-874c08b2ebf6)
![Forward-history clear bg](https://github.com/atlas-engineer/nyxt/assets/6666079/99b68d2f-3df9-4178-a719-368473f01388)
![backward-history clear bg](https://github.com/atlas-engineer/nyxt/assets/6666079/10220759-99fc-43c9-b4a3-d07901174698)

should look like:
![publishing](https://github.com/atlas-engineer/nyxt/assets/6666079/2575e1f5-e649-4ec7-b125-17a61d6816ea)

# Pre-Build Checklist:
_talk through the change with someone, then come back and mark off what you can_
- [x] I have consulted with the team. 
- [x] One or more developers have signed off on this change.
- [x] One or more user researchers have signed off on this change.

lansingthomas:
in terms of spacing... 
the left arrow, right arrow, and lambda should be the same total width.
the reload icon should be slightly wider / have a little more room
I have them like this:
![publishing july 26](https://github.com/atlas-engineer/nyxt/assets/6666079/8827be6f-2f2d-49ad-a607-058682cc610e)

PR #3107, opened 2023-07-29
Fix buttons

jmercouris:
Fixes: #3099

Screenshot_20230730_004558


jmercouris:
And here is on hover
Screenshot_20230730_004625

jmercouris:
Indeed, also got to make sure it works on dark theme and that the color of the SVG can be changed! I think it can, just have to test.

aartaka (approved):
Oh, so the design is finalized now! Even though I still feel strange about the rounded lines, it's a progress that I can't deny :upside_down_face: All good!

lansingthomas (approved):

jmercouris:
Screenshot_20230801_023846

lansingthomas:
@jmercouris 

Yay dark theme!

- [ ] can the tile be purple too?

I know you mentioned doing the color change programmatically, and here is a purple version in case it helps.

tile component purple action

PR #3108, opened 2023-07-30
Fix font

jmercouris:
I can't seem to load fonts from a local resource. It is not allowed.

Not sure how to circumvent this without running our own HTTP server to load the assets. This seems way overboard though. Any ideas @atlas-engineer/atlas ?


hgluka:
Maybe data URLs could work? We even have a convenient `make-data-url` function in `web-extensions.lisp`, although it's not exported.

I don't know if there are any drawbacks performance-wise, though.

aartaka:
> I can't seem to load fonts from a local resource. It is not allowed.
> 
> Not sure how to circumvent this without running our own HTTP server to load the assets. This seems way overboard though. Any ideas @atlas-engineer/atlas ?

What Epiphany does is defining their own `ephy-resource://` custom theme, which can gain access to the local files. And then they use `ephy-resource` links to load bundled resources.

Here's a definition:
``` C
webkit_web_context_register_uri_scheme (priv->web_context, "ephy-resource",
                                        (WebKitURISchemeRequestCallback)ephy_resource_request_cb,
                                        NULL, NULL);
webkit_security_manager_register_uri_scheme_as_secure (webkit_web_context_get_security_manager (priv->web_context),
                                                       "ephy-resource");
```

What we can do is:
- Include all the static assets as `:static-file`s with ASDF.
- Define a scheme that's resolve files with ASDF.
- And pack these assets together with Nyxt (should be automatic on Guix, but could be non-trivial elsewhere.)

We can mimic Epiphany with:
``` lisp
(define-internal-scheme "nyxt-resource"
  (lambda (url buffer)
    #|...|#
    (asdf:system-relative-pathname :nyxt (quri:uri-path (url url)))
    #|...|#)
  :secure-p t)
```

aartaka:
> Maybe data URLs could work? We even have a convenient `make-data-url` function in `web-extensions.lisp`, although it's not exported.
> 
> I don't know if there are any drawbacks performance-wise, though.

That's a solution we rely on in several places, yes. But it's better to avoid it. 

There are several drawbacks, starting with the [length restrictions for `data:` URLs](https://developer.mozilla.org/en-US/docs/web/http/basics_of_http/data_urls), and up to [security risks](https://blog.mozilla.org/security/2017/11/27/blocking-top-level-navigations-data-urls-firefox-59/). So if we have any option to not use `data:` URLs, then we probably should pick that non-`data:` option.

jmercouris:
Thank you for the idea Artyom, I have implemented it! Still not sure about its portability or what `asdf:static-file` does. Does that mean that the system relative files will be available on a compiled system even if the binary is moved?

jmercouris:
there's an NASDF file thing, perhaps that can be used. Beyond this, do you think it is ready? (I'll test the flatpak)

jmercouris:
flatpak failed, have to load the fonts into the image somehow and cache them.

jmercouris:
I've figured out a strategy for caching the fonts in the image. I will continue to think about it a little bit more before committing.

aartaka:
> Beyond this, do you think it is ready? (I'll test the flatpak)

It works interactively, but I can't make it work when built with
Guix. Problem is on my side, most probably. Can someone else test it
when built with Guix?

jmercouris:
The problem is not on your side, static file will not do what we wish. We must load things into the image.

jmercouris:
@aartaka take a look at my latest commits. For some reason, the `*static-data*` is populated interactively, but not when compiled... I don't get it :-(

jmercouris:
@Ambrevar any ideas?

jmercouris:
note that 5bbc97d9 does work, but that is no way to do things...

jmercouris:
I have made it work even portably. Static-files is *not* the way to go here.

jmercouris:
I used a merge commit because I want to show the process and show *why* static files doesn't work, and how it could work. Perhaps someone could pick it up in the future.

source/browser.lisp
```diff
@@ -184,6 +184,10 @@ prompt buffer.")
     (make-instance 'theme:theme)
     :type theme:theme
     :documentation "The theme to use for all the browser interface elements.")
+   (glyph-logo
+    #.(alexandria:read-file-into-string
+       (asdf:system-relative-pathname :nyxt "assets/nyxt.svg"))
+    :documentation "The logo of Nyxt as an SVG.")
```
Ambrevar, line 190: Why call it "glyph-logo" and not "nyxt-logo" or "application-logo"?
Does it have to be an SVG?

source/global.lisp
```diff
@@ -178,4 +178,11 @@ Return nil on error."
     (when (and commits (not (zerop commits)))
       (push-feature "UNSTABLE"))))
 
-(defvar +logo-svg+ (alexandria:read-file-into-string (asdf:system-relative-pathname :nyxt "assets/nyxt.svg")))
+(export-always '*static-data*)
```
Ambrevar, line 181: I suggest we unexport this and instead add a function to get the static data.  This would be more future-proof.

source/urls.lisp
```diff
@@ -457,6 +457,15 @@ guarantee of the same result."
                 (setf (url buffer) (quri:uri url))
                 (apply (form internal-page) args))))))))
 
+(define-internal-scheme "nyxt-resource"
+    (lambda (url buffer)
+      (declare (ignore buffer))
+      (multiple-value-bind (data exists)
+          (gethash (quri:uri-path (url url)) *static-data*)
+        (declare (ignore exists))
+        data))
```
Ambrevar, line 466: Why the `multiple-value-bind` if you are ignoring `exists`?
This is effectively equivalent to `(gethash (quri:uri-path (url url)) *static-data*)`.

source/browser.lisp
jmercouris: Yes, has to be an SVG. Name was set to be consistent with other glyphs

source/urls.lisp
jmercouris: I know what you are thinking, but if you remove it, Webkit loses its mind. I didn't add it for no reason :-D

source/global.lisp
jmercouris: A good idea!

source/urls.lisp
Ambrevar: Oh, I see, this is because the callback multiple values have specific expectations for an Internal scheme, see `define-internal-scheme` documentation.

Anyways, to effectively discard multiple values beside the primary value, use `nth-value 0` or `prog1`.

Done in b3ac06d97204183fd44237c73f211140847c9f30.

jmercouris: I almost used nth value, since I was on the fence, we go with it!

PR #3109, opened 2023-07-31
documents/EXTENSIONS: Update nx-router and nx-tailor synopsis.

migalmoreno:
# Description

I've re-worked both extensions over the past months, and I feel an update to their description was needed, as the current ones are too vague IMO.

# Discussion

I'd appreciate if you could test the extensions since I've recently updated them to work with the latest Nyxt APIs. Also any feedback would be more than welcome to know if I'm going in the right path.


migalmoreno:
Apologies, I missed it! Squashed the changes.

PR #3129, opened 2023-08-18
Fix `dark-mode`

### LINKED ISSUE #3128: Enabling `dark-mode` shows warning and doesn't work
**Describe the bug**

Enabling `dark-mode` shows the following warning/error:

```
<WARN> [18:03:03] Warning: JavaScript error: GError: Domain: "WebKitJavascriptError", Code: 699, Message: nyxt:new:1: SyntaxError: Unexpected identifier 'e'. Expected ')' to end an argument list.
<INFO> [18:03:03] Dark mode enabled.
```
But dark mode is not actually enabled.
The same warning is shown when running the `darken` bookmarklet command directly.

1. Open any webpage in Nyxt.
2. Enable `dark-mode`.
3. Read the message buffer.

or

1. Open any webpage in Nyxt.
2. Enable `bookmarklets-mode`.
4. Execute the `darken` command.
5. Read the message buffer.

hgluka:
# Description

This fixes `dark-mode` by removing the `javascript:` URL prefix from the `darken` bookmarklet.

Fixes #3128.

# Checklist:

- [ ] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [ ] Documentation:
- [ ] Compilation and tests:


hgluka:
Ah, I missed those. Should've checked for any earlier work on it.

With that in mind, here's what I found:

```lisp
(source (if (str:starts-with-p "javascript:" source)
                               (quri:url-decode (subseq source 11))
                               source))
```

If you look at the above code, `quri:url-decode` is only called if the url starts with `javascript:`. 
If we change the last line to `(quri:url-decode source)` as well, then one other bookmarklet breaks: 
`invert-colors`, which happens to be the only one other than `darken` that's not of the form:
```
(function(){ ...code here... })();
```

Maybe that's the issue here?

aartaka:
> hgluka it seems that we need to answer the question: do we need `quri:decode`? CC @aartaka.

We do need it in case of `javascript:` URLs, because these _should_ be percent-encoded. In case there's no `javascript:` URL, then don't decode.

We can have an additional check, for `javascript:` URLs: if there's any character that has to be percent-encoded, then this is not a valid `javascript:` URL and, by extension, bookmarklet. Throw warnings when that happens.

hgluka:
I agree with @aartaka. The only question is, do we want to throw an error when a bookmarklet with `javascript:` isn't properly encoded, or just echo a warning and let things continue?

aartaka:
Same as @jmercouris:
- Echo a warning.
- Ignore the bookmarklet.
  - Or define it to alert the user (mayve with JS' `window.alert()`.

hgluka:
@aadcg @aartaka I don't think there's a way to prove whether a string is percent-encoded or not.
What I did here is check if decoding the `source` for the bookmarklet has any effect.

Still, if a bookmarklet that starts with `javascript:` has a `+`, by this logic it could possibly be encoded (encoding changes spaces to plus signs, decoding does the opposite), so that would still slip through the cracks.
That's why I left `darken` without the `javascript:` prefix.

source/mode/bookmarklets.lisp
```diff
@@ -28,7 +28,12 @@ By default, this mode does nothing but expose the default bookmarklet commands."
                              (pathname (files:content (make-instance 'files:file :base-path source)))
                              (string source)))
                    (source (if (str:starts-with-p "javascript:" source)
-                               (quri:url-decode (subseq source 11))
+                               (let ((source-original (subseq source 11)))
+                                 (if (string= source-original (quri:url-decode source-original :lenient t))
```
aartaka, line 32: This decoding check I'm not sure about.

The more we discuss it, the more I'm leaning to the initial implementation:
- Decode when `javascript:`,
- And do nothing otherwise.

hgluka: Yeah, you're probably right.
Then the main issue was with `darken` not being properly encoded, and that's fixed. I'll just revert this commit.

hgluka:
@aadcg Should be ready now.

PR #3148, opened 2023-09-07
Add `plaintext-editor-mode` to default modes.

### LINKED ISSUE #3006: Edit User Files button still not working for me
**Describe the bug**
3.1.0
default config
Ubuntu

"Edit User Files" button on the common settings internal page is still not getting me where I want to go.

I try to open the wide world of editing configurations from the `settings` page, and get this message in the message buffer. Warning: Please set `external-editor-program` browser slot"

This is fine, I can start figuring all that out... Do we not have a default text editor already? Is there not an extension? I asked John to try this on his machine and the result was different. His machine brings up a prompt buffer with some options. 

from my attempt at editing configuration files from within nyxt:
config file button screenshot

**solution might look like**
My dream experience here teaches me a bit about configurations generally, and holds my hand (just a bit!) while I setup the app for more general use.
```
1. press the button
2. the nyxt default editor opens up for me
3. my configuration file is open before me
4. comments in that file explain what it is, and how I might use it.
```

I would like this to be easy for me and for those who come after : )
Thank you!
Please advise.

hgluka:
# Description

Add `plaintext-editor-mode` to default modes for `editor-buffer`. This is the default, built-in editor for Nyxt.

Fixes #3006.

# Discussion

I don't think we need both of the modes enabled for this to work, but I kept `editor-mode` in just in case.
Also, as mentioned in #3006, we could change the `Edit user files` button in `common-settings` to launch the built-in editor instead of an external editor.

# Checklist:

- [x] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [x] Documentation:
- [x] Compilation and tests:

### DIFF
--- a/source/mode/editor.lisp
+++ b/source/mode/editor.lisp
@@ -97,8 +97,8 @@ See the `scheme' documentation for the format and the number of values that
 contains an `nyxt/mode/editor:editor-mode' instance (or a subclass thereof)."))
 
 (defmethod nyxt:default-modes append ((buffer editor-buffer))
-  "Add `editor-mode' to `editor-buffer' by default."
-  (list 'editor-mode))
+  "Add `editor-mode' and `plaintext-editor-mode' to `editor-buffer' by default."
+  (list 'editor-mode 'plaintext-editor-mode))
 
 (defmethod nyxt:default-modes :around ((buffer editor-buffer))
   ;; REVIEW: Really remove document-mode from editor-buffer?


aartaka:
The default behavior of editor-mode is to explicitly say that it's configurable. And, ideally, mention plaintext-editor-mode. The latter it doesn't mention yet, but that's fixable.

In case we set plaintext-editor-mode as the default, one would be extremely dissappointed in Nyxt (built-in editor) as the idea if they open some file.

No one changes the defaults (* in case of Nyxt it is less true than for most other tech products, but at least people coming from qutebrowser certainly expect a good default behavior AND we're aiming at good and friendly first-timer experience). If we provide an incomplete and almost unusable editor by default, one would simply stop using the built-in editor, or, at worst, Nyxt. If we say that they can pick and choose, they will be aware of the behavior being changeable.

So here's a set of options that I'm ready to accept:

- Retain the old behavior. And, maybe mention plaintext editor in the message.

- Move the editor configuration to common settings so that plaintext-editor is a conscious choice.

- At least echo the configurability message in plaintext-editor (although that's semantically confusing—why is an editor mode mentioning installing some editor mode if there's one installed and running already?)

hgluka:
True, but then again, the people who don't change defaults would almost certainly have trouble configuring `editor-mode` on their own.

I think we can combine your first two suggestions into a good middle ground, but I'll wait for @jmercouris and @aadcg to chime in as well, since they suggested this change in #3006.

aartaka:
plaintext-editor-mode is still incomplete, so including it by default is no different than erroring. Let's approach the issue of not being able to edit user file in Nyxt holistically and make at least some useful enhancements to plaintext-editor-mode before mandating it on the user.

But I'm not qualified talk of user ***@***.*** is. Tom, can you test and see what's a better option? Here's an (untested) config snippet to test the plaintext editor:

(define-configuration :editor-buffer
  ((default-modes `(:plaintext-editor-mode ,@%slot-value%))))

jmercouris:
I've tested it. It isn't feature complete, but it is better than nothing. The user can read, and can save files. For most new users, this suffices. They usually wish to copy and paste things into their config. @lansingthomas Please weigh in with your verdict.

lansingthomas:
Sorry its long -- TLDR; yes merge, then keep improving.

### I see two problems here we want to solve for users;
**Problem 1** - editing user files
   - must include: locating the appropriate files
   - must include: discovering which user files one should notice, or care about
   - must NOT: be so junky that we scare users away from Nyxt entirely

**Problem 2** - Setting that external text editor / organizing the file editing process for fun/customizability/and perhaps keeping a favorite text editor in the mix.

### I see several existing routes for solving problem 1 - editing a text file from the Nyxt environment.
1. `edit-file`  -- the built in editor -- not 100% -- but workable with a few more elements. -- does NOT help users locate user files.
2. `edit-user-file-with-external-editor` -- leads to great things, with prerequisite configuration outside of many user's immediate capabilities (such as: setting the _editor_ slot) -- DOES help users locate user files
3. `open-file` -- can be used to select a text file from the prompt buffer, &rarr; the OS then prompts for available text editors... this is functioning as an "edit with external editor" button already. -- though it requires most users to specify an editor each time, which is burdensome. -- does NOT help users locate user files.
4. `editor-open-file` -- only discoverable when an editor is already open in Nyxt -- changes which file the user is editing.  -- related because the similar name adds to the chaos.

### And we have a few open questions
A. should we enable `plaintext-editor-mode` in the default configuration? (to enable the built in editor by default)
   - I think yes, we should. It removes a barrier to Problem 1, which brings people a step closer to understanding the structure of their user files, and how they might go about customizing Nyxt.
   - I like the simplicity of the built in editor for new users. (particularly those who arrived by clicking on the button in Common Settings) -- I do not think that it needs to do much beyond plaintext editing and saving for this specific thing.
   - I see how using an external editor could be preferable, but I would like for people in the nyxt environment to stay there. There is something cool about editing Nyxt, from within Nyxt. Let's enable that out of the box.
 
B. What should the button(s) in `Common Settings` accomplish?
   - At this stage, people are probably just trying to figure out which user files they might care to edit and find out what the options might be. 
   - _one button_ should get people **locating** and **editing** user files as directly as possible in the built-in editor.
      - I see that the current broken button runs `edit-user-file-with-external-editor`, which is the only option we have so far that helps with the locating user files part of the process. But clearly is not working for editing because `editor slot` might not yet be configured!
      - could we create a new command; `edit-user-file-with-internal-text-editor` and set the "Edit user files" button to it to solve problem 1?
   - _a separate button_ should help configure the default text editor.
 
C. Is the plaintext editor good enough to share?
   - I have a few obvious changes in mind (for another thread) -- Its no so bad now that we need to hide it xD -- I vote to leave it published and make it better soon.

**John**
> It isn't feature complete, but it is better than nothing. The user can read, and can save files. For most new users, this suffices. They usually wish to copy and paste things into their config.

**Tom**
I concur. This might be the first time ever for many users to copy lisp code into a configuration file. A simple plaintext editor, out of the box, with no config required is a wonderful solution.

---

**Artyom**
> Tom, can you test and see what's a better option? Here's an (untested) config snippet to test the plaintext editor:...

Yes happily, thank you for the snippet. I share your fear that a bad experience for users in the built in editor would be unfortunate. Let's open a separate issue for improving the built-in editor. (Or find whichever issues are already open and discuss there.  -- I get the sense that we all agree the built in editor needs to be improved)

as far as the better option... It seems like we have both `open-file` and `edit-file` so, I am happy with `open-file` relying on the OS to prompt for other programs, and setting `edit-file` to the nyxt built in editor as default -- I think it is clear enough that the slot is configurable, if we have a button for it in common settings. We could even use text-editor as an example in the Slot configuration section of the Manual.

**Andre**
> My opinion is that if the user invokes `edit-file`, something should happen. Raising an error that says that a parameter isn't configured is unacceptable. 

totally agree -- even an imperfect solution is better than a confusing error message.

hgluka:
Thanks for the input, everyone!

>  could we create a new command; `edit-user-file-with-internal-text-editor` and set the "Edit user files" button to it to solve problem 1?

@lansingthomas Yes, that's easy enough to do. The only question is, should we have two buttons in `common-settings`, one for the internal and one for the external editor? I assume that having the option to use either one is good.
Although that might be a question for the larger `common-settings` rewrite that will be coming soon!

PR #3155, opened 2023-09-09
mode/file-manager: Open PDF files locally with OPEN-FILE

shamazmazum:
Since PDF.js is supported by nyxt, it may be a good idea to open PDF files which are stored locally right in the browser (with `C-x C-f` or `C-Space open-file`), so add `"pdf"` to `supported-media-types`.


shamazmazum:
I have made this PR in response to this:
https://github.com/atlas-engineer/nyxt/pull/2988#issuecomment-1711921820

and tested it in Nyxt 3.5.0, should work in master too.

Issue #2877, opened 2023-03-25
Use `define-configuration` internally?


aartaka:
So we have these complex use-cases where we need to modify a certain source (`url-sources`) return actions with lots of mode-specific commands. To do that, we refactor all the source into fragile one-off functions concatenating mode-specific suggestions into a final list of return actions. This asks for some refactoring and a better composition.

But then... we have exactly this tool for composition of slot modifications: `define-configuration`! Yes, it's configuration-specific and otherwise restricted in its core usage, but why not introduce something like that in the core, given the abundance of use-cases?

Use-cases and examples where `define-configuration` or a copy thereof coul help:
- Composing actions from multiple places without a single source of authority: `url-sources`.
  - Simply appending over `%slot-value%` suffices here.
- Forward-resolving: we have a whole set of `default-modes` methods that have `(sym:resolve-symbol :annotate-mode :mode)` in them. While `nsymbols` is good at what it does (resolving the necessary symbols at runtime), it's still a bad forward-reference that we have here. 
  - Instead, we can compose over the `%slot-value%` again.
- Making defaults overridable. I recently had this problem in https://github.com/aartaka/nyxt-config/issues/4 trying to wrap my head around removing the pre-set `default-modes`. Given that `default-modes` are partially constructed by accessor methods and `progn` method combinations, it's hard to revert what the methods do.
  -  In this case, we can use this `define-configuration-internal` something instead of the `progn` method combination. It would also be much easier to revert in the config, when necessary.

Now, a signinficant problem that we may have is the `pre-request-hook` over again: the undecidability of the hook activation order may make our internal configuration to interfere with the user one, thus negating or confusing the latter. As a possible solution, we can make a dedicated hook for core configuration in `user-class`, so that `customize-method` first runs this internal hook and then the actual user one. Not a nice model, but it could replace lots of models that haunt us for a long time now :)

Ambrevar:
Can you give an example?  (For instance with `url-source`.)

aartaka:
Oh wait. `url-sources` is not configurable due to being a method. But here's an example with `default-modes`. Instead of the current
``` lisp
(defgeneric default-modes (buffer)
  #| ... |#
  (:method append ((buffer context-buffer))
    (list
     ;; TODO: No need for `sym:resolve-symbol' if we move `context-buffer'
     ;; declaration in a separate file, loaded after modes.
     (sym:resolve-symbol :annotate-mode :mode)
[...]

PR #3199, opened 2023-10-02
[WIP] Buffers panel highlight current

MaxGyver83:
# Description

While `switch-buffer` is very efficient for finding a specific buffer among many, it requires that you remember what to search for. Having a list of open buffers always visible (as p.e. in Firefox) helps to keep an overview. Nyxt's buffer panel is very similar to a tab bar in other browsers.

In this pull request, I try to make the buffer panel even a bit more useful.

## To Do:

* ✔ Make the reload button stand out by paining it green and moving it above the header.
* ✔ Highlight the current buffer by painting its button blue (currently only updated when the buffer is changed using the buffer panel).
* ✘ Reload the buffer panel via `window-set-buffer-hook`.

### Before:

buffers-panel-old

### After:

buffers-panel-new

# Discussion

Please tell me if such a changed is welcome at all. If yes, you might prefer a different design. I actually want to use the default `prompt-buffer` highlight color instead of a simple blue but I couldn't manage to select this theme color. I also don't yet understand how to set a hook.

One detail: In line 100, I create either a regular or a blue `:nbutton` depending on the `highlight` argument. Actually, I would prefer to write less duplicate code but this does not work:

```lisp
(:nbutton :text (title buffer)
                  :buffer panel-buffer
                  @(if highlight '(:style "background-color:blue;"))
                 ...
```

I don't know why. Probably easy to solve but I'm a Common Lisp beginner.

# Checklist:

- [x] Git hygiene:
- [ ] Documentation:
- [ ] Compilation and tests:


source/mode/buffer-listing.lisp
```diff
@@ -88,36 +88,43 @@ With LINEAR-VIEW-P, list buffers linearly instead."
...
   "Display a list of buffers with easy switching."
-  (flet ((buffer-markup (buffer)
+  (flet ((buffer-markup (buffer &optional highlight)
            "Create the presentation for a buffer."
            (spinneret:with-html-string
-             (:p (:nbutton :text (title buffer)
-                   :buffer panel-buffer
-                   `(nyxt::switch-buffer :buffer ,buffer))))))
+             (:p (if highlight
+                     (:nbutton :text (title buffer)
```
aartaka, line 101: You can rely on Spinneret expanding style attributes from Lisp expressions:
``` lisp
(:nbutton
  :text (title buffer)
  :buffer panel-buffer
  :style (when highlight "background-color:blue;")
  ...)
```

Spinneret as a library is quite smart, so you can pass regular Lisp expressions to it and be (99%) sure that they will be interpreted properly.

source/mode/buffer-listing.lisp
```diff
@@ -88,36 +88,43 @@ With LINEAR-VIEW-P, list buffers linearly instead."
...
     (spinneret:with-html-string
       (:nstyle '(.button
                  :white-space nowrap
                  :overflow-x hidden
                  :display block
                  :text-overflow ellipsis))
       (:body
-       (:h1 "Buffers")
        (:nbutton :text "Update ↺"
+         :style "background-color:green;"
```
aartaka, line 118: You can use style classes exported by Nyxt to color things in a more consistent and theme-dependent shades. Like:
``` lisp
;; For green.
(:nbutton
  :class "success"
  ...)
;; For blue.
(:nbutton
  :class "action"
  ...)
```
----------------------------------------------------

Also note that inline styles are hard to maintain, so you better put the styles into the `style` slot of the respective mode (`buffer-listing-mode` here) or use the built-in style classes, as I've suggested.

source/mode/buffer-listing.lisp
```diff
@@ -88,36 +88,43 @@ With LINEAR-VIEW-P, list buffers linearly instead."
...
-            :test #'string=
-            :key (compose
-                  #'render-url #'url))))
+         `(reload-panel-buffer ,panel-buffer))
+       (:h1 "Buffers")
        (loop for buffer in (buffer-list)
-             collect (:raw (buffer-markup buffer)))))))
+             collect (:raw (buffer-markup buffer (eq buffer (current-buffer)))))))))
+
+(defun reload-panel-buffer (panel-buffer)
```
aartaka, line 125: This might go into the body of the panel command as:
``` lisp
(reload-buffer panel-buffer)
```

So no need for a separate function that works out of context.

source/mode/buffer-listing.lisp
```diff
@@ -88,36 +88,43 @@ With LINEAR-VIEW-P, list buffers linearly instead."
...
-                   :buffer panel-buffer
-                   `(nyxt::switch-buffer :buffer ,buffer))))))
+             (:p (if highlight
+                     (:nbutton :text (title buffer)
+                      :buffer panel-buffer
+                      :style "background-color:blue;"
+                      `(nyxt::switch-buffer :buffer ,buffer))
+                     (:nbutton :text (title buffer)
+                      :buffer panel-buffer
+                      `(progn
```
aartaka, line 107: You can pass several expressions as `:nbutton` body:
``` lisp
(:nbutton
  ...
  `(nyxt::switch-buffer :buffer ,buffer)
  `(reload-buffer ,panel-buffer))
```

aartaka (commented):
I like the idea! I've left some comments, mostly cosmetic and helping with Spinneret and our custom tags, like `:nbutton`.

source/mode/buffer-listing.lisp
MaxGyver83: This works like a charm! Thank you. I'll change it.

source/mode/buffer-listing.lisp
MaxGyver83: I have tried this. It works for `:class "action"` but not for `:class "success"`. Where is the according CSS defined?

MaxGyver83: OK, I think I have found it:
<https://github.com/MaxGyver83/nyxt/blob/be7ca1c5cabde6eb857581a35b0734a1694c0acf/source/prompt-buffer.lisp#L136-L138>

aartaka:
>I have tried this. It works for `:class "action"` but not for `:class "success"`. Where is the according CSS defined?

It's buffer.lisp, line 167 on master.

source/mode/buffer-listing.lisp
aartaka: Oh, right. I haven't taken the window hook into account. For that, a separate function might be necessary. But here, in buttons, you can benefit from the panel buffer accessibility.

source/mode/buffer-listing.lisp
aartaka: No, prompt buffer is not related to panel buffer. What you need is buffer.lisp, and `style` slot inside `panel-buffer`. 

But there were no proper CSS classes there, so I added them in 8cb76d447c6ed5e6e26b8b561763c1f95532fb83. Rebase over master and try again. Should work this time.

source/mode/buffer-listing.lisp
MaxGyver83: Applied in 6ebccfc5.

source/mode/buffer-listing.lisp
MaxGyver83: Now, green and blue are use as foreground colors:
buffers-panel-new2
Was this intended? I think the contrast of blue on grey could be better.

source/mode/buffer-listing.lisp
MaxGyver83: Applied in commit 6ebccfc51.

MaxGyver83:
So far, my best try  to add `reload-panel-buffer` to the `window-set-buffer-hook`:

```lisp
(defun buffers-panel-handler (window buffer)
  (reload-panel-buffer (first (nyxt::panel-buffers (current-window)))))

(define-configuration window
  ((window-set-buffer-hook
    (hooks:add-hook %slot-default% 'buffers-panel-handler))))
```

AFAIU, this warning lets the compilation fail:

```
Compilation warning: The variable WINDOW is defined but never used.
```

When I remove the function arguments `window` and `buffer`, I get a different error.

aartaka:
>Compilation warning: The variable WINDOW is defined but never used

It's just a compiler being picky. Add (declare (ignore window)) to the handler right after argument list and it'll work.

source/mode/buffer-listing.lisp
MaxGyver83: I have added the window hook in commit bec09b2b. Now I call `reload-panel-buffer` in the hook. Is it correct to call this top-level function also in the buffers panel? Or should I create a `flet` expression nevertheless?

source/mode/buffer-listing.lisp
aartaka: Hmmmmm, that's not right. And I know why. Fixed in 08573888a2cda3826faaa57719733ebdbe275241. Thanks for highlighting (pun not intended) it!

source/mode/buffer-listing.lisp
```diff
@@ -88,36 +88,47 @@ With LINEAR-VIEW-P, list buffers linearly instead."
...
          :buffer panel-buffer
-         `(reload-buffer
-           (find
-            (render-url (url ,panel-buffer))
-            (nyxt::panel-buffers (current-window))
-            :test #'string=
-            :key (compose
-                  #'render-url #'url))))
+         :class "success"
+         `(reload-panel-buffer ,panel-buffer))
```
aartaka, line 114: This still feels like an unnecessary indirection. Can't you use a raw 
``` lisp
(reload-buffer ,panel-bufer)
```
here? I might be out of context, so correct me if I'm missing something.

source/mode/buffer-listing.lisp
```diff
@@ -88,36 +88,47 @@ With LINEAR-VIEW-P, list buffers linearly instead."
...
+                   :test #'string=
+                   :key (compose #'render-url #'url))))
+
+(defun buffers-panel-handler (window buffer)
+  (declare (ignore buffer))
+  (let ((panel-buffer (first (nyxt::panel-buffers window))))
+    (when panel-buffer
+      (reload-panel-buffer panel-buffer))))
+
+(define-configuration window
```
aartaka, line 132: We had an issue about actually using `define-configuration` in our codebase: #2877. Bu it needs a new architecture to work properly. Until then, we leave `define-configuration` as a strictly user config construct.

How about dropping this `define-configuration`, using `reload-buffer` in `buffers-panel`, and removing `buffers-panel-handler`&`reload-panel-buffer`? We lose the updates this way, but it's more reliable this way. You can leave a `TODO`/`FIXME` in the code with the mention of live updates.

Sorry if it's a bit messy and non-interactive—need a bit of refactor to make it all work properly :sweat_smile:

source/mode/buffer-listing.lisp
MaxGyver83: :-) Thank you for fixing it. It looks better now! I have updated the screenshot in the pull request description.

source/mode/buffer-listing.lisp
MaxGyver83: This seems to work. At least I don't get an error. Pressing the reload button has no visible effect. I'm wondering if it's needed at all with this pull request. The buffers panel is reloaded via `window-set-buffer-hook` whenever the user switches to another buffer. And when a buffer is closed, the previous(?) buffer is loaded automatically which triggers the hook, too. Is there still a use case which requires a manual reload?

Also, if `` `(reload-buffer ,panel-bufer)`` works, why was `` `(reload-buffer (find ...))`` used initially? Does it make a difference when there are multiple panel buffers!?

source/mode/buffer-listing.lisp
MaxGyver83: I don't yet understand why `define-configuration` isn't allowed. (I only had a quick look at #2877.)

Don't you think it's pointless highlighting the current buffer when it doesn't get updated automatically!? This would mean that in some cases (p.e. when using `switch-buffer`) a wrong buffer remains highlighted which is rather confusing. Maybe it's better to have no buffer highlighted than a wrong one.

aartaka:
@aadcg, take a look. Most of my programming-related questions are resolved, so what's left is the feature assessment.

source/mode/buffer-listing.lisp
```diff
@@ -88,36 +88,47 @@ With LINEAR-VIEW-P, list buffers linearly instead."
...
-;; - Update button and buffer names are styled the same and are not visually
-;;   separated.
 ;; - Buffer as a button is quite confusing and does not match the idea of a
 ;;   switchable-to entity better separate action (switch, delete) from
 ;;   presentation.
 (define-panel-command-global buffers-panel ()
     (panel-buffer "*Buffers panel*")
   "Display a list of buffers with easy switching."
-  (flet ((buffer-markup (buffer)
+  (flet ((buffer-markup (buffer &optional highlight)
```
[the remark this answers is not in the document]
MaxGyver83: I have changed it: 62559ce1e4d70c46767fb5fc81a2defff78c0d29.

source/mode/buffer-listing.lisp
MaxGyver83: Do you mean to keep `(reload-buffer (find ...))` or to keep `(reload-panel-buffer ,panel-buffer)` (which is a wrapper around the former to prevent duplicate code)?

Maybe the first question is if I can add a hook that calls `reload-panel-buffer` at all!? @aartaka [said](https://github.com/atlas-engineer/nyxt/pull/3199/files#r1344010508) that I can't use `define-configuration` in the codebase. Is there a different way to add a hook directly in the code?

MaxGyver83:
> Steps to reproduce:
> 
>     * `(nyxt:start :urls '("https://en.wikipedia.org/wiki/Tomato") :failsafe t)`
> 
>     * Invoke command `buffers-panel`
> 
>     * Invoke command `set-url-new-buffer`
> 
>     * Notice the behavior above.

True. It looks like my testing was too narrow. I have committed two changes:

* Print "Loading..." instead of a tiny empty button when the title of a new buffer is not yet known.
* I have added a `buffer-loaded-hook`. I think both hooks are necessary.

I'm still trying to add the `window-set-buffer-hook` directly in `window.lisp` in `initialize-instance`.

MaxGyver83:
> @MaxGyver83 would you like to make the changes yourself or shall I do it? Thanks.

I have committed your suggested change, updated the changelog and rebased.

PR #3226, opened 2023-10-17
mode/hint: Add hints-alignment-x slot

### LINKED ISSUE #3179: Position hints next to links (instead of covering them)
Whenever I use hints, they cover the associated links (partly). The (not yet released) pull request #3173 allows the user to define pixel offsets for the hints. This is a great improvement but not ideal because the user has to find the right offset to have the hints next to the links via trial-and-error.

It would be more user-friendly, if the user could choose if they want the hints to the left of a link, to the right of a link or on top of a link (current default). 

None.

**Additional context**
Example of how it looks like in Firefox with the Surfingkeys add-on: [Hints cover link text - Nyxt / Configuration Questions - Atlas](https://discourse.atlas.engineer/t/hints-cover-link-text/857)

MaxGyver83:
# Description

Add `hints-alignment-x` slot to allow placing the hint overlays to the left or the right of a link. Possible values: `"left"`, `"on-top"` (current behavior, default), `"right"`.

Fixes #3179.

## Screenshots

`"left"`:
left

`"on-top"`:
on-top

`"right"`:
right

# Discussion

The left alignment is achieved by setting `transform: translate(-100%, 0);`. This translation is applied after ensuring that `style.left` is ≥ 0. This is why some hints may be not completely visible (like `AA` in the first screenshot). Any ideas how to fix?

# Checklist:

- [x] Git hygiene:
- [ ] Documentation:
- [ ] Compilation and tests:


source/mode/hint.lisp
```diff
@@ -79,6 +79,10 @@ A positive value shifts to the right.")
     :type integer
     :documentation "The number of pixels that hint overlays are vertically shifted by.
 A positive value shifts to the bottom.")
+   (hints-alignment-x
+    "on-top"
+    :type string
```
hgluka, line 84: I feel like `symbol` or `keyword` is a better type for this slot. What do you think?

MaxGyver83: You are probably right. I'll try to find an example for how to create keywords.

hgluka: Well, don't quote me on this, but keywords behave basically the same as symbols. You just use `:` to refer to them instead of `'`, because they are (automatically) exported in the `keyword` package and `:my-keyword` is shorthand for `keyword:my-keyword`.
You can find an example of how they're used in the `height` slot of a `prompt-buffer`.

And if you want to be really fancy, you can do something similar to `default-cookie-policy` and the `cookie-policy` type, which is basically an enum with keyword symbols.

MaxGyver83: I have changed it to keywords.

MaxGyver83:
> @MaxGyver83 I think this is going in a good direction but the implementation is too cumbersome - the second loop seems inappropriate.
> 
> In the same way that `ps:max` is being used to ensure that the top-left corner is within viewport, I think that a similar strategy could be used as to ensure that the bottom-right corner is too via `window.innerWidth` and `window.innerHeight`.

@aadcg :
* The default (`hint-alignment-x` = `:on-top`) is simple: The top-left corner of the hint is set and must be >= 0.
* `:left`: Here, the top-left corner of the hint overlay is set and then it's shifted by 50% of its width to the left. But this width is unknown. This is why I don't know the new top-left position which would be one argument of `ps:max`.
* `:right`: Here, the top-left corner of the hint overlay is set and the right edge might be outside the viewport. Same problem as for `:left`: The width of the hint overlay is unknown and the position of the top-right (or bottom-right) corner can't be calculated.

This is why I let the renderer do the calculations and check (and adjust) the values afterwards (second loop).
Do you know how I can calculate the width of the hint overlays? It depends on the number of characters, which characters are used (if the font isn't monospace) and all the CSS (paddings, margins, border width).

MaxGyver83:
> @MaxGyver83 I've thought again about this and my conclusions seem to be aligned with yours.
> 
> Given issue #3250, which impacts hint mode before this PR, it makes little sense to request from your PR that all hints must be fully drawn in viewport. They should be, however, partially visible. I believe that the PR meets this requirement, even if the added loop is deleted. I'm sorry that I have requested from your PR to meet a goal that we don't currently meet, as I was unaware of #3250.
> 
> I suggest deleting the aforementioned loop and reviewing the code once again. `create-hint-overlay` became a rather complex function, and it would be great to abstract the key ideas into smaller pieces. Your [description](https://github.com/atlas-engineer/nyxt/pull/3226#issuecomment-1802675224) of how the location of the hints is computed should be clear in the implementation. For instance, perhaps a small function that implements the idea behind `(ps:@ element style transform) "translate(-100%, 0)")` paired with a proper docstring would made the implementation more readable.
> 
> Does that make sense? Thanks.

Sounds good. I'll update this pull request.

MaxGyver83:
> I suggest deleting the aforementioned loop ...

Done in commit 756bdd7.

I have also removed the transform/translate part. This was only a workaround to set the hint's right edge/border position. I think it clearer to set the style.right position using `window.innerWidth` (I think you mentioned this before).

Now, I check the position for a min and a max value: It should be >= 0 and not too close to the window border to make sure the hints are at least partly visible. I'm using now a magic number (25 px). This is probably better than nothing (until we have found a better solution).

> `create-hint-overlay` became a rather complex function, and it would be great to abstract the key ideas into smaller pieces. Your [description](https://github.com/atlas-engineer/nyxt/pull/3226#issuecomment-1802675224) of how the location of the hints is computed should be clear in the implementation. For instance, perhaps a small function that implements the idea behind `(ps:@ element style transform) "translate(-100%, 0)")` paired with a proper docstring would made the implementation more readable.

This is still open. I tried this at first for the example you mentioned but without success. I guess I have to copy the element as argument to a new function and then return the modified element. Can I pass an element pointer/reference?

MaxGyver83:
> You don't need to return the element if you're just interested in the side effects.
> 
> ```lisp
> (defun test (element) (setf (ps:@ element style top) "2px") nil)
> ```
> 
> To check how it compiles to JS, pass it as an argument to `ps:ps` and see that it returns the JS code below.
> 
> ```js
> function test(element) {
>     element.style.top = '2px';
>     return null;
> };
> ```

Thank you!

> Let me know if you're interested in giving it a try. Regardless, one of us can take over and finish this PR.

I'll give it a try! (I have started with a rebase + squash.)

MaxGyver83:
@aadcg:
Here I'm trying to move some code from `create-hint-overlay` to a new function `set-alignment`:
[MaxGyver83:nyxt:hints-next-to-links-dev](https://github.com/atlas-engineer/nyxt/compare/master...MaxGyver83:nyxt:hints-next-to-links-dev?expand=1)
The build fails with:

```
Unhandled SIMPLE-ERROR in thread #<SB-THREAD:THREAD tid=50675 "main thread" RUNNING
                                    {10013D0133}>:
  Compilation warning: The function PARENSCRIPT:CHAIN is called with two arguments, but wants exactly one.
```

`ps:chain` is called with multiple arguments in many places. Can you tell me what's wrong here?

MaxGyver83:
> Regardless, one of us can take over and finish this PR.

This pull request has been open for quite a while. If you want to get this merged quickly, it's no problem for me if you finish this PR.

PR #3256, opened 2023-11-16
hint.lisp: Tweak appearance.

jmercouris:
Makes the hints look better.


jmercouris:
Interesting, seems we need to set the font-size to a static number.

jmercouris:
Turns out it was the border-width. I had to set it to an even number to avoid some strange behavior. It is fixed now. Waiting for tests to pass, then I will merge.

I have of coursed rebased on master ;-)

jmercouris:
Aha, I didn't see that. I was looking in the source code for something like that. Thank you for the tip. I will probably push a follow-up commit with the change.

jmercouris:
Done!

PR #3267, opened 2023-12-01
Internal define configuration

### LINKED ISSUE #2877: Use `define-configuration` internally?

aartaka:
# Description

This allows for use of `define-configuration` in Nyxt code, making several things (`define-modes` in particular) simpler to do. See the discussion in https://github.com/atlas-engineer/nyxt/pull/3199/files#r1344010508 for the potential problems using `define-configuration` internally would cause.

Fixes #2877.

# Checklist:

- [X] Git hygiene:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    ```
  - and Guix dependencies.
- [X] My code has been reviewed by at least one peer.  (The peer review to approve a PR counts.  The reviewer must download and test the code.)
- [X] Documentation:
- [X] Compilation and tests:


jmercouris:
I see, so the idea is that we can add configuration for a given mode/class anywhere in the codebase?

As an example, in this case, you are using it to add several default modes to a context-buffer? This allows us to remove the resolve-sym hack to add the default modes for a context-buffer? This also allows us to load our files in a more flexible order then?

aartaka:
> I am not sure about the distinction between `pre-customize-hook` and `customize-hook`. What do we gain from it, and what would we loose without it?

So imagine we set up some modes in Nyxt.
I'm already doing this in this PR.
And imagine that the user sets up their modes too—in their config.
So we have several hook handlers customizing `default-modes`, for example.
Now, the order of hook handlers is not guaranteed to be... ordered in any way.
So, if the user removes some mode in their config, while we add this same mode, there might be different scenarios:

```
Nyxt enables A
Nyxt enables B
User disables A
User enables C
Enabled: B and c
```

or the actually problematic:
```
Nyxt enables B
User disables A
Nyxt enables A
User enables C
Enabled: A, B, and C
```

So we need to somehow prevent Nyxt from running config before the user one.
Mainly so that user config can easily override the defaults.
That's why there's `pre-customize-hook`—it runs before any user configuration and ensures that user code can override our config.

aartaka:
> I see, so the idea is that we can add configuration for a given mode/class anywhere in the codebase?
> 
> As an example, in this case, you are using it to add several default modes to a context-buffer? This allows us to remove the resolve-sym hack to add the default modes for a context-buffer? This also allows us to load our files in a more flexible order then?

Yes, yes, yes, and yes.

jmercouris:
I am OK with the use if the `define-configuration` internally, I think it is a smart strategy. I too would like to understand why we need another hook. If that can be clarified, I think this is a good improvement!

aartaka:
`pre-customize-hook` removed in 2aa591e5c9a4671182e1e400bb9c3c7d649f2dfd. Merging?

aartaka:
Updated, rebased onto master, removed the changelog entry. Safe to merge (or just donate—it's a single commit now), I believe.

PR #3262, opened 2023-11-27
Fix nasdf

aadcg:
# Description

Fixes #2981

# Checklist:
Everything in this checklist is required for each PR.  Please do not approve a PR that does not have all of these items.

- [x] Git hygiene:
  - I have pulled from master before submitting this PR
  - There are no merge conflicts.
- [x] I've added the new dependencies as:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    git submodule add https://gitlab.common-lisp.net/nyxt/py-configparser _build/py-configparser
    ```
  - and Guix dependencies.
- [x] My code follows the style guidelines for Common Lisp code. See:
  - [Norvig & Pitman's Tutorial on Good Lisp Programming Style (PDF)](https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf)
  - [Google Common Lisp Style Guide](https://google.github.io/styleguide/lispguide.xml)
- [x] I have performed a self-review of my own code.
- [ ] My code has been reviewed by at least one peer.  (The peer review to approve a PR counts.  The reviewer must download and test the code.)
- [x] Documentation:
  - All my code has docstrings and `:documentation`s written in the aforementioned style.  (It's OK to skip the docstring for really trivial parts.)
  - I have updated the existing documentation to match my changes.
  - I have commented my code in hard-to-understand areas.
  - I have updated the `changelog.lisp` with my changes if it's anything user-facing (new features, important bug fix, compatibility breakage).
    - Changelog update should be a separate commit.
  - I have added a `migration.lisp` entry for all compatibility-breaking changes.
  - (If this changes something about the features showcased on Nyxt website) I have these changes described in the new/existing article at Nyxt website or will notify one of maintainters to do so.
- [x] Compilation and tests:
  - My changes generate no new warnings.
  - I have added tests that prove my fix is effective or that my feature works.  (If possible.)
  - I ran the tests locally (`(asdf:test-system :nyxt)` and `(asdf:test-system :nyxt/gi-gtk)`) and they pass.

PR #3268, opened 2023-12-02
Review testing and ci

aadcg:
# Description

Follow-up to https://github.com/atlas-engineer/nyxt/pull/3262#issuecomment-1837154415.

# Discussion

The testing underlying these 2 reader macros entail loading the nyxt system, pushing them to `*features*` and force loading the nyxt system again. This is not how a robust test suite should work. 

The features themselves are either under developed or were added as a workaround to deeper issues. 

The commit messages provide extra details.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,8 +1,5 @@
-# Inspired by http://3bb.cc/blog/2020/09/11/github-ci/.
 name: Tests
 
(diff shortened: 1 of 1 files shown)


jmercouris:
I have often thought that we should source, cache, and host copies of all submodules to protect against failure and ensure complete reproducibility within our control. That is however beyond the discussion of this pull request.

aartaka (approved):

Issue #2989, opened 2023-05-28
Support Chrome-based renderer


aadcg:
WebKitGTK has enabled us to mature Nyxt, but it is undeniable that its performance as a web renderer is subpar when compared to Chrome-based ones. This is particularly noticeable in resource intensive tasks such as watching a video on YouTube. 

Supporting new renderers is in itself beneficial since it puts Nyxt's API under the test. As a remainder, Nyxt is renderer agnostic.

To achieve this goal there are two alternatives: CEF or Electron. The former has proved to be tough to achieve, while the latter seems [promising](https://github.com/atlas-engineer/cl-electron). 

WebKitGTK will continue to be supported.

This move will also helps us to distribute Nyxt on MacOS and Windows.

aartaka:
There's always QtWebEngine as yet another Chromium wrapper that's more or less ethical and modularized. Just that it's C++ and pain to bind to (⌒_⌒;)

LinArcX:
Really? Electron? 

Did you measure performance?

aadcg:
@LinArcX performance is certainly an important factor. But the ability to easily write Lisp bindings is actually the most important aspect to achieve this goal. 

Others can expand on the technical issues, if you'd like to :)

aartaka:
Yep, ease of Lisp bindings is the main feature for us, because we don't have:
- A C++ person to make https://github.com/atlas-engineer/cl-webengine complete.
- A CEF-aware person to conform to all the opaque requirements that CEF sets up, mostly locking us to C :D

fnicastri:
I really like the idea of Nyxt and I would like to be capable to use it on Mac but,
honestly, Electron would be a very bad choice.
Please don't.

aadcg:
Would you like to expand?

LinArcX:
I don't understand the word of "promising".

Guys you are not creating a toy tool. This is a browser.

We're living in a world that most people's work are migrating to web. So look at it as an operating system.

Performance matters and it matters a lot in your domian.

Please don't be hype with "promising" technologies.

ILoveGoulash:
Are there numbers about the performance gap relevant to real world use? What about complexity (incl. bloat) or features (e.g. the recently talked about JPEG XL thing? Is supporting more than one implementation going to be a massive pain in the "glue" part of the code?

[...]

PR #3290, opened 2023-12-15
Review foreign interface

### LINKED ISSUE #3265: Actually override FFI methods in renderer system
Our FFI design relies on renderer class inheritance for method specialization. So `renderer-buffer` is a superclass of `buffer`, allowing one to specialize methods on `renderer-buffer` to add some renderer-specific logic. The problem is: regular classes, being subclasses of renderer ones, are more specific and thus their methods override the renderer-specific methods. Even though it should've happened the other way around.

Here's one example:
- Invoking zooming commands modifies buffer zoom settings.
- This setting modification triggers `(setf ffi-buffer-zoom-level)` method.
- This method tries to dispatch over the current buffer.
  - If it's a regular `buffer`, use fallback JavaScript-based zooming.
  - If it's a `gtk-buffer`, use WebKit-provided scrolling.
- Here's when it breaks: `buffer` is inherited from `gtk-buffer` and thus is more specific.
- So JavaScript-invoking method is picked instead of FFI one.
- And thus the zooming is slightly broken, because JS-based zooming is somewhat broken by design.

One option for fixing it is making our code a bit dirtier with actual renderer-specific overrides. Even though we make our FFI architecture harder to grasp, it ensures that we actually override the necessary methods. In case of example above, we can do:
```
(defmethod (setf current-zoom-ratio) (value (buffer document-buffer))
  (when (plusp value)
    (setf (webkit:webkit-web-view-zoom-level (gtk-object buffer)) value)))
```
in renderer/gtk.lisp. This will override the method from buffer.lisp, which is bad. But it'll also ensure that only the renderer-specific methods are called. And it's also quite Lispy—monkey-patching the imperfect behavior to make it better 😛

_Originally posted by @aartaka in https://github.com/atlas-engineer/nyxt/issues/3250#issuecomment-1833558175_

aadcg:
# Description

A review of certain foreign interface quirks noticed while working on #2989. 

Fixes #3265.

Won't be part of the 3-series.

### DIFF
--- a/source/foreign-interface.lisp
+++ b/source/foreign-interface.lisp
@@ -33,45 +33,37 @@ If the `:setter-p' option is non-nil, then a dummy setf method is defined."
                  (declare (ignore value ,@arguments)))))))))
 
 (define-ffi-generic ffi-window-delete (window)
-  "Delete WINDOW, possibly freeing the associated widgets.
-After this call, the window should not be displayed.")
+  "Delete WINDOW.")
 
 (define-ffi-generic ffi-window-fullscreen (window)
-  "Make WINDOW fullscreen.
-Removes all the Nyxt interface element and leaves the open page as the only
-thing on the screen.")
+  "Set fullscreen WINDOW state on.")
 (define-ffi-generic ffi-window-unfullscreen (window)
-  "Un-fullscreen the WINDOW.")
+  "Set fullscreen WINDOW state off.")
 
 (define-ffi-generic ffi-window-maximize (window)
-  "Make the WINDOW to fill the screen.
-Usually possible via desktop environment interface too.")
+  "Set WINDOW to a maximized state.")
 (define-ffi-generic ffi-window-unmaximize (window)
-  "Switch the WINDOW to \"windowed\" display.
-May or may not fill the screen.")
+  "Set WINDOW to an unmaximized state.")
 
 (define-ffi-generic ffi-buffer-url (buffer)
-  "Return the `quri:uri' associated with the BUFFER.
+  "Return the URL associated with BUFFER as a `quri:uri'.
 This is used to set the BUFFER `url' slot.")
 (define-ffi-generic ffi-buffer-title (buffer)
-  "Return as a string the title of the document (or web page) showing in BUFFER.")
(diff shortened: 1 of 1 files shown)


source/foreign-interface.lisp
```diff
@@ -33,143 +33,134 @@ If the `:setter-p' option is non-nil, then a dummy setf method is defined."
...
 
 (define-ffi-generic ffi-window-active (browser)
-  "The primary method returns the focused window as per the
-renderer.
+  "Return the focused window.
 
-The `:around' method automatically ensures that the result is set to
-`last-active-window'.
+The specialized method must fallback on the primary method below, as to account
+for the case when the renderer reports that none of the windows is focused.
```
jmercouris, line 74: "are focused"

source/foreign-interface.lisp
```diff
@@ -228,42 +219,47 @@ spontaneous events from programmed ones.
...
 
 (define-ffi-generic ffi-generated-input-event-p (window event)
-  "Return non-nil if EVENT was generated by `ffi-generate-input-event'.")
+  "Return non-nil when EVENT was generated by `ffi-generate-input-event'.")
 
-(define-ffi-generic ffi-within-renderer-thread (browser thunk)
-  "Run THUNK (a lambda of no argument) from the renderer's thread.
-This is useful in particular for renderer-specific functions that cannot be run on random threads."
-  (funcall thunk))
+(define-ffi-generic ffi-within-renderer-thread (thunk)
```
jmercouris, line 224: I'm not certain this isn't just an implementation detail of GTK. Is this necessary for other ports?

source/foreign-interface.lisp
```diff
@@ -441,8 +441,7 @@ Example:
 
 ;;; Signals
 
-(export-always 'on-signal-notify-uri)
```
jmercouris, line 441: I presume we don't need to export the symbols for signal handlers?

source/foreign-interface.lisp
jmercouris: Even if it /is/ an implementation detail of GTK, it is quite necessary for GTK. It'll just be a no-op in other ports.

jmercouris (approved):

PR #3294, opened 2023-12-20
pull_request_template.md: Greatly minimize the size.

jmercouris:
-------

# Description

This is designed to make it not so intimidating. Yes, it is less verbose, it is far easier however.


jmercouris:
Why this change? We skip over lots of the content that is in gigantic walls of text!

Issue #3295, opened 2023-12-21
`hint-mode` Shadow DOM elements support


aadcg:
As mentioned in https://github.com/atlas-engineer/nyxt/pull/3269#issuecomment-1862897209, `hint-mode`'s support for Shadow DOM may need to be re-designed from scratch.

Related to #3254. 

Performance benchmarks are needed as to ensure that the addition of this feature won't impact performance. E.g., mock an HTML document with thousands of links and return the time it takes for the renderer to draw the hints.

heiwiper:
Hello @aadcg !

I've been working on this issue these past days and I think there's no need for a major rewrite. The main problem which was causing the performance issue was the fact the use of the macro `rqs-nyxt-id` on every hintable element of the DOM. The reason I was using that macro instead of `rqsa` (which would've made more sense as a replacement for `qsa`) is because the results of that macro should match the exact same order as the function `clss:select` which is used to generate hints. At that time the macro `rqs-nyxt-id` didn't cause any noticeable performance issue so I made the choice to use that macro in order to avoid having to keep `rqsa` and `clss:select` using the same ordering logic, otherwise changing one of them would break the other.

Anyway, I have resorted to updating the `rqsa` macro to produce a depth-first ordered elements, which worked fine for now.

However there's one issue that I would like to discuss with you, and which I think would affect other websites. So basically when the document model is updated through `update-document-model` command (which happens twice for hint-mode, once after adding the `nyxt-hintable` attribute to hintable elements, and the other one is after removing them) the new elements that are added to the dom and which weren't part of it wouldn't get added to the dom for timing reasons I would assume.

(I'm getting 2 breakpoints in the video because I was setting a sticker inside the add-hints function)
https://github.com/atlas-engineer/nyxt/assets/12894695/162fd0ff-4612-4eac-b904-a450cb9034ca

heiwiper:
I have followe your advice about creating an example of the issue and after some debugging I found out that the issue had nothing to do with `update-document-model`. I updated the `rqsa` algorithm to make use of `nyxt-shadow-root` attribute to improve its performance but then I noticed that the function `set-hintable-element` gets called before `update-document-model` which causes `rqsa` macro that is used within `set-hintable-element` function to ignore some shadow root elements that were added to the DOM.

Now I have reverted to walking through all DOM elements using Tree Walker which apparently has good performance compared to the other javascript methods.

I have created a draft PR to fix this issue, please let me know when you take a look into it.

PR #3303, opened 2024-01-03
Drop `hint-mode`'s Shadow DOM support 

### LINKED ISSUE #3254: `hint-mode` performance
Commit b43ec712d450e64605e1a4a167ea6c992a3a036e introduced a considerable performance penalty when invoking `follow-hint`. To confirm the claim, invoke the command over https://en.wikipedia.org/wiki/List_of_municipalities_in_Quebec and notice that the time to draw the hints is impacted by a factor of ~ 2.

Requires revisiting the #3001 thread.

Indicates that performance benchmarks are needed as to access how the addition of a feature impacts performance. In this case, the benchmark entails writing a mock HTML document with thousands of links and outputting the time it takes for the renderer to draw the hints.

@heiwiper if you have a suspicion of where the performance penalty might be coming from, please share. Thanks!

aadcg:
# Description

Fixes #3254

After a detailed examination, there is no fast way to re-write the feature in a timely fashion given the current priorities. On the other hand, the performance penalty introduced by the feature is rather severe and it is hurting the users. Issue #3295 hints at how it could be implemented.

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -945,7 +945,9 @@ invoked via " (:code "flatpak-spawn --host <command> <command-args>") "."))))
     (:ul
      (:li "Fix commands "
           (:nxref :command 'reopen-last-buffer) " and "
-          (:nxref :command 'reopen-buffer) ".")))
+          (:nxref :command 'reopen-buffer) ".")
+     (:li "Fix performance issues related to "
+          (:nxref :class-name 'nyxt/mode/hint:hint-mode) ".")))
   (:nsection :title "UI/UX"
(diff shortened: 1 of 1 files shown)


jmercouris:
Thanks Andre. I shall have a look!

jmercouris:
I would have thought it would be enough to swap `rqsa` for `qsa`! In any case, it looks good. Thank you.

jmercouris:
P.S. this is WAY faster.

jmercouris (approved):

PR #3308, opened 2024-01-13
Review handling of proportional and monospace fonts

### LINKED ISSUE #3307: Monospace fonts aren't honored on `code` tags
The way we set the font we ship (Public Sans) overrides the cases when monospace fonts should be used.

2024_01_13_15:40:22-nyxt

### LINKED ISSUE #3314: Public Sans isn't set for all interfaces
For instance, it is not honored in the status buffer.

The issue is observed from the moment the feature was added (https://github.com/atlas-engineer/nyxt/pull/3108).

### LINKED ISSUE #3315: Non-monospaced fonts for some `prompt-buffer` elements?
As a follow-up to #3314, we could enable Public Sans for the `prompt-buffer`. Currently, it uses monospaced fonts everywhere.

aadcg:
Closes #3307.
Closes #3314.
Closes #3315.

### DIFF
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -968,6 +968,10 @@ allows access to common functions that are defined within the mode."))))
      (:li "Review the appearance of all help system pages such as"
           (:nxref :command 'describe-class) ".")
      (:li "Review font sizes.")
+     (:li "Add possibility to specify a monospace font via "
+          (:nxref :class-name 'theme:theme :slot 'theme:monospace-font-family) "."
+          "By default, it is set to DejaVu.")
+     (:li "Review usage of proportional and monospace fonts.")
      (:li "Use monospace fonts in "
           (:nxref :class-name 'nyxt/mode/editor:plaintext-editor-mode) ".")
      (:li "Replace source heading buttons in the "


jmercouris:
I didn't make this change because it changed A LOT of the UI, including the describe interface. I'll look at it closely.

jmercouris:
I tested this, and interestingly enough, now noto sans was being set in the status buffer. I have no idea why this is. This is for sure not the solution!

jmercouris:
> Is there a font (technically it couldn't be called a single font) that delivers both proportional and monospaced versions?

I do believe that there exists a monospaced version of public sans.

jmercouris:
On second search I've been unable to find a reference to monospace version of Public Sans.

jmercouris:
> Is there a font (technically it couldn't be called a single font) that delivers both proportional and monospaced versions?

Yes there is. However, I haven't really liked how they look. I have found that Bitstream Vera Sans works really well with Public Sans. I'll submit a pull request.

jmercouris:
I've added a commit that addresses the size of the prompt buffer fonts. I didn't want to force push on you, so feel free to include it however you like in the commit history. This is important for making things look reasonably sized.

jmercouris (approved):

Issue #3297, opened 2023-12-22
Delete Functions


jmercouris:
The following functions should be deleted:

[ ] clean-configuration
[ ] copy-placeholder
[ ] delete-no-procrastinate-host
[ ] disable-hook-handler
[ ] duplicate-buffer
[ ] duplicate-buffer-with-current-modes
[ ] enable-hook-handler
[ ] follow-hint-with-current-modes-new-buffer
[ ] go-up
[ ] history-forwards-query
[ ] load-file
[ ] no-procrastinate-buffer-host
[ ] no-procrastinate-current-host
[ ] no-procrastinate-host
[ ] nyxt-init-time
[ ] print-bindings
[ ] reload-buffer
[ ] reload-with-modes
[ ] resume-prompt
[ ] save-input-data
[ ] set-action-on-current-suggestion
[ ] set-input-data-from-saved
[ ] set-input-data-from-saved-domain

jmercouris:
Are there any reasons not to delete them?

jmercouris:
Firstly. I am OK with all of your suggestions about what *not* to delete that you've listed above.

With regards to `record-input-field-mode`, yes, we can fully delete it.

With regards to `go-up`, `go-next` et al., the problem is that it works on so /few/ pages, and thus presents to the user a degraded experience. If we could generalize it for websites that do not support this pattern, or provide the user with a helpful message, we could leave them in.

PR #3310, opened 2024-01-15
Refactor `{enable,disable}-hook-handler` commands.

aadcg:
# Description

These commands were suggested for deletion on #3297 but, on a closer look, I think they're in line with our goals.

I've revised the facilities.

I don't think it deserves a changelog entry.

# Checklist:

  - Tests are sufficient.


jmercouris:
I'm not sure this is an improvement. Yes you are /technically/ removing code, but is there any real benefit? Also, do these /need/ to be commands? When would we ever use these?

jmercouris:
OK!

PR #3311, opened 2024-01-15
Delete modes and commands

### LINKED ISSUE #1633: Cleanly separate optional components and report them with `system-information`
Currently OSPM only works on Guix (os-package-manager-mode.lisp commands do nothing on systems without Guix) and `lisp-system.lisp` is not available when Nyxt is compiled without Quicklisp.

Furthermore, users may wish to compile Nyxt without certain modes (for whatever reason).

So what about moving `lisp-system` to a mode, then as part of #1327 report which modes are compiled-in in `system-information`?

### LINKED ISSUE #3297: Delete Functions

aadcg:
# Description

Fixes #1633
Fixes #3297

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -997,4 +997,26 @@ enabling its customization.")
    (:li "The height of the " (:nxref :class-name 'message-buffer) " and "
         (:nxref :class-name 'status-buffer) " can be dynamically set and its UI
 elements are scaled accordingly.")
-   (:li "Remove experimental support for WebExtensions via " (:code "libnyxt") ".")))
+   (:li "Remove experimental support for WebExtensions via " (:code "libnyxt") ".")
+   (:li "Delete commands:"
+        (:ul
+         (:li (:code "clean-configuration"))
+         (:li (:code "copy-placeholder"))
+         (:li (:code "dashboard"))
+         (:li (:code "duplicate-buffer"))
+         (:li (:code "duplicate-buffer-with-current-modes"))
+         (:li (:code "follow-hint-with-current-modes-new-buffer"))
+         (:li (:code "go-next"))
+         (:li (:code "go-previous"))
+         (:li (:code "go-to-homepage"))
+         (:li (:code "go-up"))
+         (:li (:code "print-bindings"))
+         (:li (:code "reload-buffer"))
+         (:li (:code "reload-with-modes"))
+         (:li (:code "resume-prompt"))))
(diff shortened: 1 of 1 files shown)


source/buffer.lisp
```diff
@@ -1945,15 +1945,10 @@ specified for their contents."
...
-(define-command reload-buffer
-    (&optional (buffer
-                (prompt
-                 :prompt "Reload buffer(s)"
-                 :sources (make-instance 'buffer-source))))
-  "Reload BUFFER.
-Return it."
-  (when buffer
-    (buffer-load (url buffer) :buffer buffer)))
+(export-always 'reload-buffer)
```
jmercouris, line 1948: I'm trying to understand why this was ever a function. We have reload-buffers, and reload-current-buffer, why do we need this? I really can't imagine.

source/mode/document.lisp
```diff
@@ -169,35 +169,6 @@ Important pieces of functionality are:
   "Scroll to top if no input element is active, forward event otherwise."
   (call-non-input-command-or-forward #'scroll-to-top :buffer buffer))
```
jmercouris, line 170: I see you deleted all of the go-* functions. Yes, why not. I don't think anyone will miss them. Plus they need some reworking to consider /all/ cases, to be more reliable.

jmercouris (approved):
I've gone through commit by commit. I've  tested it on my local system. All appears well. Thank you Andre!

source/buffer.lisp
jmercouris: Yes, I meant a command, indeed.

PR #3312, opened 2024-01-15
Delete html-string-p and related logic.

aadcg:
# Description

I don't see the need for `html-string-p` in our API.

No need to for a changelog entry.

# Checklist:

  - Tests are sufficient.


jmercouris:
I think the idea was that modes could generate things that are either HTML, or just strings, and somehow we would do different actions depending. I don't know. In any case, you are right. Not needed.

Issue #1263, opened 2021-03-30
Keep page scroll position after reload


tg-x:
It would be handy if the page would retain the scroll position after reload.
Currently it jumps to the top after reload.

tg-x:
After some further testing I found out that even if I navigated within the page to an anchor, the anchor position is not restored either after reload.

Ambrevar:
This is #1164 I believe, right?

No idea why there would be an issue with local files only, need to investigate.

For a related feature, see #392.

aadcg:
Confirm if the bug report is reproducible after closing #2999.

jmercouris:
This is however a tricky question, can we /really/ retain position on reload? The contents may be completely different, what is the original "position" even?

PR #3324, opened 2024-01-29
Restore Scroll Position on Reload

### LINKED ISSUE #3325: Retain scroll position on `reload-current-buffer`
Self-explanatory.

jmercouris:
# Description

- When the user uses the command "reload-buffer" or "reload-current-buffer" they can have it restore the scroll position that they were at.

Fixes #3325

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1646,14 +1646,21 @@ specified for their contents."
     (current-buffer)))
 
 (export-always 'reload-buffer)
-(defun reload-buffer (buffer)
-  "Reload BUFFER and return it."
+(defun reload-buffer (buffer &key retain-scroll-position)
+  "Reload BUFFER and return it.
+
+If retain-scroll-position is t, calculate the position of the scroll, and
+attempt to restore it upon reload."
+  (when retain-scroll-position
+    (let ((scroll-position (ps-eval :buffer buffer (ps:chain window page-y-offset))))
+      (hooks:once-on (buffer-loaded-hook buffer) (buffer)
+        (ps-eval :buffer buffer (ps:chain window (scroll-to 0 (ps:lisp scroll-position)))))))
   (buffer-load (url buffer) :buffer buffer))
 
-(define-command reload-current-buffer ()
+(define-command reload-current-buffer (&key (retain-scroll-position t))
   "Reload current buffer.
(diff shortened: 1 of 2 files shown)


jmercouris:
> I'm afraid that there are some misunderstandings with regards to the report from #1263. It is both a non reproducible bug report (wrt local files not honoring URIs with fragments) and a feature request. This PR addressed the latter, which is superseded by #3325.

I don't believe it has anything to do with URI fragments. Here is the use case I think it is:

I am writing a LaTeX file. I compile it to HTML. I want to update the HTML view in Nyxt that is rendering my LaTeX file. For this use case, it works perfectly. It is not a function of the renderer. The render can't actually *generally* solve this problem. We are just providing a naieve solution that works for 99% of cases.

> Concretely, this solution seems to suffer from a fundamental issue. Navigate to `https://en.wikipedia.org/wiki/Tomato#Nutrition`, issue `reload-current-buffer` and the conclusion follows.

Yes, that *is* a problem. To solve this we need to compare the scroll position that we are upon reload rather than blindly moving by some offset. In other words, scrolling to an absolute position would solve this problem.

> In conclusion, I'd suggest closing this PR and think about a correct implementation in the future.

I've outlined a straightforward path to make this pull request correct for 100% of cases! Therefore, I disagree.

jmercouris:
This is not a trivial thing. In my opinion it *is* a fix. The fact that on reload things scroll willy-nilly is infuriating. This fixes a LONG standing issue I've had.

jmercouris:
Thank you for finding the first bug, please give it another try, I don't think there remain any problems.

jmercouris:
https://github.com/atlas-engineer/nyxt/assets/1691662/e04b6556-7da1-43f1-b99c-e457567b5884

Are you sure that you are loading the branch correctly?

jmercouris:
With regards to the information about WebKit supporting a buffer reload function, that will require changing our FFI. Which is of course possible, but now we will have `buffer-load` AND `buffer-reload`.

jmercouris:
OK, I will implement it that way then. I know this is indeed not an important task... but I am already this deep. Plus, I find it interesting.

jmercouris:
Ah yes!

PR #3341, opened 2024-02-13
Fix lisp urls

aadcg:
# Description

This is a pre-requisite to accomplish #2989. It simplifies the logic and paves the way towards renderer agnosticism.

Fixes #2606

@jmercouris please test as many UI buttons from our internal buffers (and pages) as possible.

# Checklist:

- [x] Git branch state is mergable.
- [x] Changelog is up to date (via a separate commit).
- [x] New dependencies are accounted for.
- [x] Documentation is up to date.
- [x] Compilation and tests (`(asdf:test-system :nyxt/gi-gtk)`)
  - No new compilation warnings.
  - Tests are sufficient.

PR #3348, opened 2024-02-26
Refactor scheme architecture

aadcg:
# Description

A pre-requisite to https://github.com/atlas-engineer/nyxt/issues/2989, in the sense that it refactors the architecture of custom schemes in a renderer-agnostic fashion. It is a follow-up to #3341.

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -27,7 +27,7 @@
 
 (define-version "4.0.0"
   (:ul
-   (:li "Refactor lisp schemes URLs API.")
+   (:li "Refactor custom schemes URLs API.")
    (:li "Deprecate slot " (:code "status-buffer-position") "in favour of"
         (:nxref :slot 'placement :class-name 'status-buffer) ".")
    (:li "Deprecate slot " (:code "prompt-buffer-open-height") " since "


jmercouris:
I have been looking at this, and haven't found any problem yet. I will continue to review it.

source/renderer/gtk.lisp
```diff
@@ -884,6 +982,17 @@ See `gtk-browser's `modifier-translator' slot."
      pointer)
     (cffi:foreign-free pointer)))
 
+(defmethod ffi-register-custom-scheme ((scheme gtk-scheme))
+  ;; FIXME If a define-internal-scheme is updated at runtime, it is not honored.
```
jmercouris, line 986: Can check if Nyxt started or not.

source/renderer/gtk.lisp
```diff
@@ -810,11 +810,109 @@ See `gtk-browser's `modifier-translator' slot."
...
+  ;;   page of the same scheme.
+  ;; - :secure-p and :cors-enabled-p are too permissive for a scheme that allows
+  ;;   evaluating Lisp code.
+  ;; Therefore, no settings provide the best configuration so that:
+  ;; - <iframe> embedding and exploitation are impossible.
+  ;; - Redirection—both as window.location.(href|assign|replace) and
+  ;;   HTTP status code 301—works.
+  ;; - nyxt scheme pages are linkable from pages of other schemes.
+  (match (name scheme)
+    ("nyxt-resource" (setf (secure-p scheme) t))
```
jmercouris, line 911: So you are hardcoding the rules that we were dynamically setting before to make them cross-renderer compatible. That makes sense.

source/urls.lisp
```diff
@@ -343,37 +308,34 @@ Authority is compared case-insensitively (RFC 3986)."
...
+                :scheme "nyxt"
+                :path (symbol->param-name fn)
+                :query (unless (uiop:emptyp query) query)))))
+    (if (internal-page-symbol-p fn)
+        url
+        (error "URL ~a is undefined." url))))
+
+(export-always 'internal-page-name)
+(-> internal-page-name ((or string quri:uri)) t)
+(defun internal-page-name (url)
```
jmercouris, line 332: Since this is often used as a predicate, perhaps consider a `-p` suffix.

jmercouris: Or at least a variant of this command that is `-p` suffixed that returns t/nil.

source/mode/editor.lisp
```diff
@@ -92,8 +92,8 @@ See `describe-class editor-mode' for details."))
   (uiop:parse-native-namestring (quri:uri-path (url buffer))))
 
 (define-internal-scheme "editor"
-    (lambda (url buffer)
-      (markup (find-submode 'editor-mode buffer)
+    (lambda (url)
+      (markup (find-submode 'editor-mode)
```
jmercouris, line 96: I believe there /is/ a scenario in which we could find a different editor mode than the one we intend.

What if we set the URL of a buffer that is *NOT* the current buffer, WHILE we are viewing a buffer that contains editor mode. Would that break in this case?

jmercouris: This is a generalized problem with our concept of finding submodes rather than using methods. I don't want to go into depth right here, but it is orthogonal to the changes you've made to schemes.

source/urls.lisp
```diff
@@ -505,8 +423,7 @@ TITLE is purely informative."
                (j:encode callback-output)))
            (log:warn "Request ~a isn't bound to a callback in buffer ~a" %url buffer))
          "application/json")))
-  :cors-enabled-p t
-  :error-callback (lambda (c) (log:debug "Error when evaluating lisp URL: ~a" c)))
+  (lambda (c) (log:debug "Error when evaluating lisp URL: ~a" c)))
```
jmercouris, line 426: why `c`? condition?

jmercouris:
The `manual` command works, but not when clicked from the button on the home page.

jmercouris:
If I click on a symbol within a code block, it also does not work. I am brought again to the debugger.

jmercouris:
Will do!

jmercouris:
No problems found :-)

PR #3358, opened 2024-03-11
Delete command deprecation facilities.

### LINKED ISSUE #2212: Report replacement/version in `define-deprecated-command` ?
Questions:

- If we deprecate a command, shall we remove it for the next major release?
  If so, then no need to report the version in which a command was deprecated, because it can only be the last one.
- We should probably echo / log:warn some text explaining with what code the command should be replaced.
- Do we still need the `define-deprecated-command` macro?  Maybe it's enough to just do

```lisp
(define-command foo (...)
 ...)
(setf (deprecated-p #'foo) t)
```

Thoughts?

aadcg:
# Description

Fixes #2212

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/source/command.lisp
+++ b/source/command.lisp
@@ -21,13 +21,6 @@ packages.
 - `:mode' lists it when the corresponding mode is active.
 
 - `:anonymous' never lists it.")
-   (deprecated-p
-    nil
-    :type boolean
-    :reader t
-    :writer nil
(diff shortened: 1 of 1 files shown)


jmercouris (approved):

jmercouris:
I guess we will introduce these concepts again later, as discussed in the thread.

PR #3365, opened 2024-03-23
Fix Shadow DOM performance issue and restore its support

### LINKED ISSUE #3295: `hint-mode` Shadow DOM elements support

heiwiper:
# Description

The algorithm for the macro `rqsa` has been improved by using Tree Walker instead of Node Iterator and the matching elements are now sorted in depth first order. The sorting order was necessary to fix the performance issue, You can read more about the details of that issue in the references issues below.

The algorithm for the macro `rqs-nyxt-id` has been improved by directly retrieving Shadow root elements if the matching element cannot be found by executing `querySelect` function. Previously, we would walk through all elements to check whether each element is a Shadow root. Please note that we can't use this same method in the `rqsa` macro because the that macro might be executed on elements that do not yet have nyxt attributes assigned which isn't the case for `rqs-nyxt-id`.    

Fixes #3295

# Checklist:

  - Tests are sufficient.


jmercouris:
Thank you @heiwiper! I want you to know that this PR is not unnoticed!

heiwiper:
Thanks for the followup! I totally understand, especially considering that the related issue is marked as low priority.

heiwiper:
I have just added the changelog entry and rebased into master.

Thanks for taking the time to test these changes!

aadcg:
Thanks again @heiwiper.

heiwiper:
My pleasure!

Feel free to tag me in related issues.

Issue #3349, opened 2024-02-26
set-url: entering url without http(s):// prefix no longer works v3.11.3


fictitiousexistence:
**Describe the bug**
Typing fqdn in set-url (github.com / google.com/etc) without http(s):// prefix no longer takes you to the direct url, instead shows search results. 

Issue occurs in Nyxt 3.11.3. Works properly in ≤3.11.2

**Precise recipe to reproduce the issue**

1. Run set url command (o in vi mode)
2. type github.com > enter
3. Takes you to search results instead of https://github.com

**Information**
- OS name+version: 
gentoo 
- Graphics card and driver:
- Desktop environment / Window manager name+version: 
dwl
- How you installed Nyxt (Guix pack, package manager, build from source):
source
- Information from `show-system-information`:
Nyxt version: 3.11.3
Renderer: GI-GTK
Operating system kernel: Linux 6.7.6-gentoo
Lisp implementation: SBCL 2.4.1 (Dynamic space size: 3221225472)
Features: (:NYXT-GI-GTK :NYXT-GTK :NYXT-GI-GTK :NYXT-GTK :NYXT-3.11 :NYXT-3 :NYXT-3.11.3
 :CLIPBOARD-CONTENT-METHOD :SWANK :SLYNK :PLUMP-UTF-32 :PARENSCRIPT :NSYMBOLS
 :FSET-EXT-STRINGS :SBCL+SAFE-STANDARD-READTABLE :NAMED-READTABLES :GLOBAL-VARS
 :CL-FAD :LPARALLEL :21BIT-CHARS :CUSTOM-HASH-TABLE-NATIVE :CL-PPCRE-UNICODE
 :CL-UNICODE :CL-JSON-DOUBLE-FLOAT-IS-SUBSUMED
 :CL-JSON-SINGLE-FLOAT-IS-SUBSUMED :CHUNGA :FLEXI-STREAMS :CL-PPCRE :WEBKIT2
 :WEBKIT-2.42.5 :WEBKIT-2.42 :WEBKIT-2 :WEBKIT2-CORS-ALLOWLIST
 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING :WEBKIT2-MUTE :WEBKIT2-EMOJI
 :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22 :GTK-3-20 :GTK-3-18 :GTK-3-16
[...]

Issue #2134, opened 2022-03-06
Nyxt freezes when typing very fast in prompt buffer


Nselm:
**Describe the bug**
Nyxts's GUI freezes after typing fast in any prompt buffer with selection filters.

**Precise recipe to reproduce the issue**
When typing fast in the prompt buffer, nyxt just freezes.
A video I had running in the background continued to play, but 
1. I could not do anything 
2. When I switched first to another window, then back, it did not display nyxt anymore, but the previous window

Possibly relevant:
- I have a history of ~100 tabs
- I have ~2000 bookmarks
- I have greatly increased the key repeat rate and reduced the delay (not sure how much, but its probably around twice as fast a normal)

**Information**
- OS name+version:
Arch linux

- Graphics card and driver:
Stoney [Radeon R2/R3/R4/R5 Graphics], AMD

- Desktop environment / Window manager name+version:
i3-gaps 4.20.1-2
- How you installed Nyxt (Guix pack, package manager, build from source):
Nyxt package in aur

- Information from `nyxt --system-information`:
```
Nyxt version: 2.2.4
Renderer version: GI-GTK
Operating system kernel: Linux 5.16.12-arch1-1
Lisp implementation: SBCL 2.2.1 (Dynamic space size: 3221225472)
Features: (:WEBKIT2 :WEBKIT2-2.34 :WEBKIT2-PASTE-PLAINTEXT :WEBKIT2-TRACKING
 :WEBKIT2-MUTE :WEBKIT2-EMOJI :WEBKIT2-MEDIA :WEBKIT2-SANDBOXING :GTK-3-22
 :GTK-3-20 :GTK-3-18 :GTK-3-16 :GTK-3-14 :GTK-3-12 :GTK-3-10 :GTK-3-8 :GTK-3-6
 :GTK-3-4 :GTK :GDK-3-22 :GDK-3-20 :GDK-3-18 :GDK-3-16 :GDK-3-14 :GDK-3-12
 :GDK-3-10 :GDK-3-8 :GDK-3-6 :GDK-3-4 :CAIRO-1-10 :CAIRO-1-12 :GDK-PIXBUF
 :GLIB-2-30 :GLIB-2-32 :GLIB-2-34 :GLIB-2-36 :GLIB-2-38 :GLIB-2-40 :GLIB-2-42
[...]

PR #3385, opened 2024-04-09
Fix SET-URL

shamazmazum:
As discussed in #3383, I open another PR which fixes navigation to URLs like "github.com/atlas-engineer".

It brings two changes:
1) Uses `valid-url-p` to check if `"https://"` + query is a valid URL.
2) In `valid-url-p` move the check of TLD validity out of `(or (not check-dns-p) ...)` block. Maybe it is my fault (I'm too lazy to do git blame), but this has nothing to do with DNS lookup.

Now you can set URL to `github.com`, `github.com/foobar`, `wiki foobar` or `foobar` and get what you want in all those cases.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1398,22 +1398,18 @@ Loads the entry with default `prompter:actions-on-return'."))
   (:documentation "Structure that processes a new URL query from user input.
 Checks whether a valid https or local file URL is requested, in a DWIM fashion."))
 
-;; TODO: Remove CHECK-DNS-P from everywhere in this file and url.lisp
-(defmethod initialize-instance :after ((query new-url-query)
-                                       &key check-dns-p &allow-other-keys)
-  (declare (ignore check-dns-p))
-  (with-slots (query engine)
-      query
+(defmethod initialize-instance :after ((query new-url-query) &key &allow-other-keys)
+  (with-slots (query engine) query
     ;; Trim whitespace, in particular to detect URL properly.
     (setf query (str:trim query))
     (cond
       (engine
        ;; First check engine: if set, no need to change anything.
        nil)
-      ((valid-url-p query :check-dns-p nil :check-tld-p nil)
+      ((valid-url-p query :check-tld-p nil)
        ;; Valid URLs should be passed forward.
        nil)
-      ((valid-url-p (str:concat "https://" query) :check-dns-p nil :check-tld-p t)
+      ((valid-url-p (str:concat "https://" query) :check-tld-p t)
        (setf query (str:concat "https://" query)))
       ;; Rest is for invalid URLs:
       ((uiop:file-exists-p query)
@@ -1449,19 +1445,17 @@ Checks whether a valid https or local file URL is requested, in a DWIM fashion."
       (fallback-url (engine query)))
      (t (query query)))))
 
-(defun make-completion-query (completion &key engine (check-dns-p t))
+(defun make-completion-query (completion &key engine)
   (typecase completion
     (string (make-instance 'new-url-query
-                           :engine      engine
-                           :check-dns-p check-dns-p
-                           :query completion))
+                           :engine engine
+                           :query  completion))
     (list (make-instance 'new-url-query
                          :engine engine
-                         :check-dns-p check-dns-p
                          :query (second completion)
                          :label (first completion)))))
 
-(defun input->queries (input &key (check-dns-p t) (engine-completion-p))
+(defun input->queries (input &key (engine-completion-p))
   (let* ((terms (sera:tokens input))
          (engines (let ((all-prefixed-engines
                           (remove-if
@@ -1477,14 +1471,12 @@ Checks whether a valid https or local file URL is requested, in a DWIM fashion."
                                          (mapcar #'shortcut engines)
                                          :test #'string=))
               (list (make-instance 'new-url-query
-                                   :query       input
-                                   :check-dns-p check-dns-p)))
+                                   :query input)))
             (or (mappend (lambda (engine)
                            (append
                             (list (make-instance 'new-url-query
-                                                 :query       (str:join " " (rest terms))
-                                                 :engine      engine
-                                                 :check-dns-p check-dns-p))
+                                                 :query  (str:join " " (rest terms))
+                                                 :engine engine))
                             ;; Some engines (I'm looking at you, Wikipedia!)
                             ;; return garbage in response to an empty request.
                             (when (and engine-completion-p
@@ -1492,8 +1484,7 @@ Checks whether a valid https or local file URL is requested, in a DWIM fashion."
                                        (completion-function engine)
                                        (rest terms))
                               (mapcar (rcurry #'make-completion-query
-                                              :engine      engine
-                                              :check-dns-p check-dns-p)
+                                              :engine engine)
                                       (with-protect ("Error while completing search: ~a" :condition)
                                         (funcall (completion-function engine)
                                                  (str:join " " (rest terms))))))))
@@ -1506,8 +1497,7 @@ Checks whether a valid https or local file URL is requested, in a DWIM fashion."
                                 (completion (completion-function engine))
                                 (all-terms (str:join " " terms)))
                   (mapcar (rcurry #'make-completion-query
-                                  :engine      engine
-                                  :check-dns-p check-dns-p)
+                                  :engine engine)
                           (with-protect ("Error while completing default search: ~a" :condition)
                             (funcall (completion-function engine) all-terms))))))))
 
@@ -1516,14 +1506,14 @@ Checks whether a valid https or local file URL is requested, in a DWIM fashion."
    (prompter:filter-preprocessor
     (lambda (suggestions source input)
       (declare (ignore suggestions source))
-      (input->queries input :check-dns-p t :engine-completion-p nil)))
+      (input->queries input :engine-completion-p nil)))
    (prompter:filter-postprocessor
     (lambda (suggestions source input)
       (declare (ignore source))
       ;; Avoid long computations until the user has finished the query.
       (sleep 0.15)
       (append suggestions
-              (input->queries input :check-dns-p nil :engine-completion-p t))))
+              (input->queries input :engine-completion-p t))))
    (prompter:filter nil)
    (prompter:actions-on-return #'buffer-load*))
   (:export-class-name-p t)
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -1398,33 +1398,33 @@ Loads the entry with default `prompter:actions-on-return'."))
   (:documentation "Structure that processes a new URL query from user input.
 Checks whether a valid https or local file URL is requested, in a DWIM fashion."))
 
+;; TODO: Remove CHECK-DNS-P from everywhere in this file and url.lisp
 (defmethod initialize-instance :after ((query new-url-query)
                                        &key check-dns-p &allow-other-keys)
-  ;; Trim whitespace, in particular to detect URL properly.
-  (setf (query query) (str:trim (query query)))
-  (cond
-    ((engine query)
-     ;; First check engine: if set, no need to change anything.
-     nil)
-    ((valid-url-p (query query)
-                  :check-dns-p nil)
-     ;; Valid URLs should be passed forward.
-     nil)
-    ((and check-dns-p
-          (valid-tld-p (query query)))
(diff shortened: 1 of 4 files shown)


shamazmazum:
Oops, one test fails, investigating

shamazmazum:
Done! As I said, I would like to remove DNS lookups completely in later commits, so URLs like `http://foobar` would be valid URLs. DNS is not used to check if an URL is valid since 377e9ab, so as for now I think this is a good idea to check if a query is an URL without a scheme or a search: if a query contains a dot and `(valid-url-p (str:concat "https://" (query query)) :check-dns-p nil)` is non-NIL, then it is not a search.

aadcg:
@shamazmazum I'll review soon, thanks.

aadcg:
@shamazmazum I think there are two issues in this PR.

1. As we have agreed, we don't want to change type signatures in this PR. See the change in `initialize-instance :after`.
2. The issue we're solving is related to the call `(and check-dns-p (valid-tld-p (query query)))`. Note that `(cl-tld:get-tld "github.com") => "com"`, while `(cl-tld:get-tld "github.com/") => NIL`. For this reason, the later case is interpreted as a search engine query.

shamazmazum:
1. `initialize-instance` still can accept `check-dns-p`, obviously, but OK, I bring it back.
2. The solution I propose it simple: if it looks like an URL, check if it is a valid URL with quri. cl-tld just does not work here. Moreover, quri was also used before 377e9ab (the commit which introduced this bug). If you think that this is not OK, I would like to see any examples where this approach does not work.

shamazmazum:
@aadcg Thanks! I agree. I think, `valid-tld-p` should be incorporated in `valid-url-p` (in a way that `https://about.this` will also be a search query, just like in firefox). Do you agree with it? I prefer `valid-url-p` because in parses a string first and checks if it is an URI. Than, it can pass the host part to `get-tld` if we wish so).

shamazmazum:
@aadcg I agree with the first example. I still suggest using `valid-url-p` because it parses an URL and then you can pass the host part to `get-tld`. We can do `valid-url-p` which checks TLD and a weaker predicate which omits this test (e.g. when the scheme was explicitly supplied). Obviously, you will have to change the type of `valid-url-p` to something like `(-> valid-url-p (string &key (:check-tld-p boolean)) (values boolean &optional))`.

Otherwise, I don't know. Cut everything after `/` in the query? But `github.com/foo bar baz` is obviously not an URL. Reinvent URL parsing in `valid-tld-p`? What for if we have quri.

I am out of ideas :) I'll just let it be if you don't like the idea with TLD check in `valid-url-p`.

shamazmazum:
BTW, there is a TLD check in `valid-url-p` which never runs.

aadcg:
> @aadcg I agree with the first example. I still suggest using `valid-url-p` because it parses an URL and then you can pass the host part to `valid-tld-p`. We can do `valid-url-p` which checks TLD and a weaker predicate which omits this test (e.g. when the scheme was explicitly supplied). Obviously, you will have to change the type of `valid-url-p` to something like `(-> valid-url-p (string &key (:check-tld-p boolean)) (values boolean &optional))`.

Overall, I agree. If we follow this approach, then we don't need to check for dots in the query or am I missing something?

-------

To put things into perspective, I've noticed the mess while fixing #2134. I took the approach of changing the minimum. I did mistakes nonetheless (8b00a2f8c and the bug you mention when querying `"github.com/user/repo"`), but the goal was to keep all of the knobs in place to easily revert to a previous state. It is now clear that it makes little sense to try to fix what is fundamentally broken. Therefore, [I take my word back](https://github.com/atlas-engineer/nyxt/pull/3383#issuecomment-2045538196) and suggest that we iterate on what the ideal solution would look like without any constraints. Sorry, I had the hope that this wouldn't spiral so deep. 


------

Some things  to take into account:

1. The only "DNS check" we care for is the one done by `cl-tld`. I'm not sure whether, linguistically and technically, we should call it a DNS check or TLD check.
2. Note that the DNS/TLD-check (in the above sense) is a needed parameter, see `prompter:filter-preprocessor` and `prompter:filter-postprocessor` from `new-url-or-search-source`.
3. Ensure that issues #2134 and #3349 remain fixed.
4. Bonus points for fixing [this issue](https://github.com/atlas-engineer/nyxt/pull/3351#issuecomment-2034941731).

-----

Let me see the full solution we suggest and then we'll go from there. Thanks!

shamazmazum:
@aadcg  I hope, I did it :) I split it into two commits: the first fixes the issue with minimal changes and the second does some clean-up which removes `:check-dns-p` parameter and also `lookup-hostname` function, because it is not used anymore. If you now allow me to change signatures (as I think, you do), I suggest accepting all those commits.

#2134 remains fixed (I hope, I see not delays on my computer), #3349 fixed, IP addresses work (`127.0.0.1` goes directly to `https://127.0.0.1`). Additional test cases added.

Please, re-check if I missed something (I hope not).

source/urls.lisp
```diff
@@ -137,17 +137,6 @@ signatures."
...
-(defmemo lookup-hostname (name)
-  "Resolve hostname NAME and memoize the result."
-  ;; `sb-bsd-sockets:get-host-by-name' may signal a `ns-try-again-condition' which is
-  ;; not an error, so we can't use `ignore-errors' here.
-  (handler-case
-      #+sbcl
-    (sb-bsd-sockets:get-host-by-name name)
-    #-sbcl
-    (iolib/sockets:lookup-hostname name)
-    (t () nil)))
```
aadcg, line 149: If we're getting rid of it, then see commit 363cc9e5e36aef039d4bb49a9ef0b83e23992f42. We should also get rid of the changes made to `nyxt.asd`.

source/buffer.lisp
```diff
@@ -1398,33 +1398,32 @@ Loads the entry with default `prompter:actions-on-return'."))
...
-           (str:concat
-            "file://"
-            (uiop:native-namestring
-             (uiop:ensure-absolute-pathname
-              (query query) *default-pathname-defaults*)))))
-    (t
-     (setf (engine query)
-           (or (engine query)
-               (default-search-engine))))))
+  (with-accessors ((query  query)
```
aadcg, line 1403: Please replace with `(with-slots (query engine) query ...)`.

source/buffer.lisp
```diff
@@ -1398,33 +1398,32 @@ Loads the entry with default `prompter:actions-on-return'."))
   (:documentation "Structure that processes a new URL query from user input.
 Checks whether a valid https or local file URL is requested, in a DWIM fashion."))
 
-(defmethod initialize-instance :after ((query new-url-query)
-                                       &key check-dns-p &allow-other-keys)
+(defmethod initialize-instance :after ((query new-url-query) &key &allow-other-keys)
   ;; Trim whitespace, in particular to detect URL properly.
```
aadcg, line 1402: This comment needs to go to the right place.

source/buffer.lisp
```diff
@@ -1398,33 +1398,32 @@ Loads the entry with default `prompter:actions-on-return'."))
...
-              (query query) *default-pathname-defaults*)))))
-    (t
-     (setf (engine query)
-           (or (engine query)
-               (default-search-engine))))))
+  (with-accessors ((query  query)
+                   (engine engine))
+      query
+    (setf query (str:trim query))
+    (let ((scheme+query (str:concat "https://" query)))
```
aadcg, line 1407: Let's get rid of this `let` and accept the redundancy in the single cond clause where it's needed.

source/urls.lisp
```diff
@@ -204,14 +192,9 @@ validation may take significant time since it looks up the DNS."
...
-               (not check-dns-p)
-               (valid-tld-p (quri:uri-host %url))
-               ;; "http://algo" has the "algo" hostname but it's probably invalid
-               ;; unless it's found on the local network.  We also need to
-               ;; support "localhost" and the current system hostname.
-               (or (quri:ip-addr-p (quri:uri-host %url))
-                   (lookup-hostname (quri:uri-host %url)))))))))
+              (or (not check-tld-p)
+                  (quri:ip-addr-p (quri:uri-host %url))
+                  (valid-tld-p (quri:uri-host %url))))))))
```
aadcg, line 198: Let's not miss this opportunity to fully refactor `valid-url-p`.

The version below probably works.

```lisp
(defun valid-url-p (url &key (check-tld-p t))
  "Return non-nil when URL is a valid URL.
When CHECK-TLD-P is non-nil, check if the host is a known TLD."
  (let ((%url (ignore-errors (quri:uri url))))
    (and %url
         (valid-scheme-p (quri:uri-scheme %url))
         (if (and check-tld-p (quri:uri-http-p %url))
             (or (quri:ip-addr-p (quri:uri-host %url))
                 (valid-tld-p (quri:uri-domain %url)))
             t))))
```

source/buffer.lisp
```diff
@@ -1516,14 +1509,14 @@ Checks whether a valid https or local file URL is requested, in a DWIM fashion."
...
+      (input->queries input :engine-completion-p nil)))
    (prompter:filter-postprocessor
     (lambda (suggestions source input)
       (declare (ignore source))
       ;; Avoid long computations until the user has finished the query.
       (sleep 0.15)
       (append suggestions
-              (input->queries input :check-dns-p nil :engine-completion-p t))))
+              (input->queries input :engine-completion-p t))))
    (prompter:filter nil)
```
aadcg, line 1520: On master, `valid-tld-p` is called once on each prompt input change. With the drop of `:check-dns-p`, it is called twice. 

I don't think it matters. The solution would be a refactor of the interplay between `input->queries` and `filter-{pre,post}processor`, which is outside the scope.

source/urls.lisp
```diff
@@ -137,17 +137,6 @@ signatures."
   (setf (gethash scheme-name *schemes*)
         (list callback error-callback)))
 
-(defmemo lookup-hostname (name)
```
aadcg, line 140: Out of curiosity, why was it introduced in  the first place?

aadcg (commented):
@shamazmazum overall, I think it's almost ready.

I couldn't find regressions when comparing with master and many issues have been solved.

When comparing to how Firefox behaves, I could only see that it handles `localhost` in a more DWIM fashion. E.g. issue `python3 -m http.server` and request `localhost:8000`. On Firefox it works without adding `http://`, unlike Nyxt.

source/urls.lisp
jmercouris: Defmemo was introduced to speed up looking hostnames up on slow connections.

aadcg: That much is clear. I meant why we added the logic of `lookup-hostname`.

source/urls.lisp
shamazmazum: Oops, thanks!

source/buffer.lisp
shamazmazum: Are you sure? The original code uses accessors. What if you decide to add `(defmethod (setf engine) :after ...)` later?

shamazmazum:
OK. Thanks. I'll do it all later, but firstly I would like to know why use `with-slots` instead of `with-accessors`. These two mean different things. If you are sure that you want the first, then OK, sure :) But do you really want it?

source/urls.lisp
shamazmazum: @aadcg AFAIK `https://foobar` was interpreted as a search request if there was no host `foobar` in the network. IMHO this is too complicated for a browser.

source/buffer.lisp
aadcg: What do you mean by original? It was added in this PR.

I'm aware of the differences between `with-slots` and `with-accessors`. I'd prefer the former in this specific case, but feel free to proceed as you wish.

source/urls.lisp
aadcg: @shamazmazum I agree.

shamazmazum:
@aadcg I did what you said. You still have to type `http://localhost` if you want to visit localhost, I hope it's OK more or less. BTW check something like `Images.jl` (a Julia package) in Nyxt and Firefox :)

source/urls.lisp
aadcg: @shamazmazum what about this?

shamazmazum: I have tried it, all tests pass. It's now in PR.

aadcg:
@shamazmazum I've pushed your work in commits 14dac1e88, 424eff657, 1f583161c and ec082a380. Thank you for your patience!

PR #3429, opened 2024-07-07
mode/password: filename-based usernames for pass

### LINKED ISSUE #3293: Password Store: Support harvesting username via filename without decrypting
Currently, the Nyxt password manager integration with `pass` will query for a username by decrypting the secret file and searching for a key by one of the following names: `login`/`user`/`username`. This convention, however, is not the only one in common use.

Many users (myself included) choose to instead use the convention of including the username as part of the secret's filename (e.g.: rms@example.org) so that it may remain easily and automatically disambiguated with other credentials in the same domain (e.g.: Microsoft SSO).

For illustrative purposes, here's what a password store directory using this format might look like:
```mermaid
graph TD
    A["~/.password-store/"]-->B["example.com/"]
    A-->C["sso.example.net/"]
    B-->D["me@example.org"]
    C-->E["me@example.org"]
    C-->F["work-me@example.org"]
    A-->G["me@example.org"]
   class D,E,F,G secret
   classDef secret stroke:#333
```

Because I follow the above convention rather than storing usernames as keys in the encrypted file body, my credential usernames cannot currently be retrieved by Nyxt when invoking `copy-username`. I'd very much like it if Nyxt was made to support this convention on an optional basis so that I could take full advantage of its password management features without having to considerably rework my existing setup.

**Relevant prior art**:
* [Emacs auth-source](https://www.gnu.org/software/emacs/manual/html_node/auth/The-Unix-password-store.html)
* [Qutebrowser qute-pass](https://github.com/qutebrowser/qutebrowser/blob/fb383a68816a0b45661cee8deefc07089d2dacad/misc/userscripts/qute-pass#L83-L84)
* [rofi-pass](https://github.com/carnager/rofi-pass/blob/e77cbdbe0e885f0b1daba3a0b6bae793cc2b1ba3/rofi-pass#L528-L529)
* [Android Password Store](https://github.com/android-password-store/Android-Password-Store/blob/36f93cd60903dfbb86536ee7a16575afc782567a/crypto/pgpainless/src/test/resources/alice_owner%40example_com#L4) (*via PGPainless*)

When `copy-password` is invoked with the password-store backend active, I would like the backend to first attempt extracting the username from the filename. Ideally, it should be able to automatically detect and extract the expected username from each of the following common well-known filename patterns ([source](https://www.gnu.org/software/emacs/manual/html_node/auth/The-Unix-password-store.html))

| Directory                   | Filename                | Extracted Username  |
|-----------------------------|-------------------------|---------------------|
| .password-store             | me@example.org.gpg      | me                  |
| .password-store             | me@example.org:22.gpg   | me                  |
| .password-store             | me@example.org@example.org.gpg      | me@example.org                  |
| .password-store             | me@example.net@example.org.gpg      | me@example.net                  |
| .password-store/example.org | me.gpg                  | me                  |
| .password-store/example.org | me@example.org.gpg      | me@example.org      |
| .password-store/example.org | me@example.net.gpg | me@example.net |

I've considered abandoning this scheme and moving the username into the content of the secret file. This is probably viable, but I'd very much rather not rewrite and reencrypt hundreds of secrets when my current & preferred format is quite standard and commonly supported elsewhere -- not to mention that reencrypting hundreds of keys would take me hours since I am using a strict mandatory-touch hardware-based encryption.

chaorace:
# Description
Introduces a new filename-based code pathway for determining the username of a GNU password-store credential. This new pathway simply uses the credential file name (_not including extension_) as the username.

By default, the new pathway is only used as a fallback strategy for when no matching username keys could be found using the current strategy, though users who *only* want the new filename-based behavior can entirely skip using the original strategy via the new `scan-for-username-entries` slot. Disabling scanning credential files for a key-based username is particularly useful in use-cases where the store credentials are encrypted and would require touch verification to open.

In the special case that a credential is on the root directory of the password store, we also trim off the domain signifier to match the behavior of other password-store based utilities (*e.g.: "me@example.net" becomes "me", "me@example.org@example.net" becomes "me@example.org"*)

Fixes #3293

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/libraries/password-manager/password-pass.lisp
+++ b/libraries/password-manager/password-pass.lisp
@@ -6,10 +6,20 @@
 (define-class password-store-interface (password-interface)
   ((executable (pathname->string (sera:resolve-executable "pass")))
    (sleep-timer (or (uiop:getenv "PASSWORD_STORE_CLIP_TIME") 45))
-   (password-directory (or (uiop:getenv "PASSWORD_STORE_DIR")
-                           (format nil "~a/.password-store" (uiop:getenv "HOME")))
-                       :type string
-                       :reader password-directory))
+   (password-directory
+    (or (uiop:getenv "PASSWORD_STORE_DIR")
+        (format nil "~a/.password-store" (uiop:getenv "HOME")))
+    :type string
+    :reader password-directory)
+   (scan-for-username-entries
+    t
+    :type boolean
+    :documentation "When non-nil, a two-step process determines the username:
+1. Scan credential content for an entry with a key matching `*username-keys*'.
+2. If no match could be found, use the credential filename as a fallback.
+
+When nil, Nyxt always immediately uses the fallback strategy.
+If your store doesn't utilize username keys, this skips credential decryption."))
   (:export-class-name-p t)
   (:export-accessor-names-p t))
 
@@ -69,22 +79,46 @@ The first line (the password) is skipped."
 (defvar *username-keys* '("login" "user" "username")
   "A list of string keys used to find the `pass' username in `clip-username'.")
 
-(defmethod clip-username ((password-interface password-store-interface) &key password-name service)
-  "Save the multiline entry that's prefixed with on of the `*username-keys*' to clipboard.
+(defun username-from-name (password-name)
+  "Select a username using the path of the credential file.
+The strategy for deriving the username is context-dependent:
+- If credential exists in a subdirectory (e.g.: x.example/me@y.example),
+  username is taken as-is from the filename (me@y.example)
+- Otherwise, if no subdirectory is used (e.g.: me@x.example@y.example),
+  username is taken from the first half of the filename (me@x.example)"
+  (multiple-value-bind (_ parent-dirs credential-name)
+      (uiop/pathname:split-unix-namestring-directory-components password-name)
+    (declare (ignore _))
+    (if parent-dirs
+        credential-name
+        (subseq credential-name 0 (position #\@ credential-name :from-end t)))))
+
+(defun username-from-content (password-interface password-name)
+  "Select username from first entry in credential matching `*username-keys*'"
+  (let* ((content (execute password-interface (list "show" password-name)
+                           :output '(:string :stripped t)))
+         (entries (parse-multiline content))
+         (username-entry (when entries
+                           (some (lambda (key)
+                                   (find key entries :test #'string-equal :key #'first))
+                                 *username-keys*))))
+    (when username-entry (second username-entry))))
+
+(defmethod clip-username ((password-interface password-store-interface)
+                          &key password-name service)
+  "Save username of the `password-name' credential to clipboard.
+See the `scan-for-usename-entries' slot for details.
 Case is ignored.
-The prefix is discarded from the result and returned."
+The resulting username is also returned."
   (declare (ignore service))
   (when password-name
-    (let* ((content (execute password-interface (list "show" password-name)
-                      :output '(:string :stripped t)))
-           (entries (parse-multiline content))
-           (username-entry (when entries
-                             (some (lambda (key)
-                                     (find key entries :test #'string-equal :key #'first))
-                                   *username-keys*))))
-      (when username-entry
-        (trivial-clipboard:text (second username-entry))
-        (second username-entry)))))
+    (let ((username
+           (or (when (scan-for-username-entries password-interface)
+                 (username-from-content password-interface password-name))
+               (username-from-name password-name))))
+      (when username
+        (trivial-clipboard:text username)
+        username))))
 
 (defmethod save-password ((password-interface password-store-interface)
                           &key password-name username password service)
--- a/source/changelog.lisp
+++ b/source/changelog.lisp
@@ -96,6 +96,20 @@ elements are scaled accordingly.")
   (:li "When on pre-release, push " (:code "X-pre-release")
        " feature in addition to " (:code "X-pre-release-N") "one."))
 
+(define-version "3.12.0"
+  (:nsection :title "Features"
+    (:ul
+     (:li "Added new filename-based fallback strategy to "
+          (:nxref :command 'nyxt/mode/password:copy-username)
+          " when using the default password-store password interface."
+     (:li "Allow skipping the primary key-based username strategy utilized by "
+          (:nxref :command 'nyxt/mode/password:copy-username)
+          " via the new slot "
+          (:nxref :class-name 'password:password-store-interface
+            :slot 'scan-for-username-entries)
+          ". By default, it is set to "
+          (:code "t") " (do not skip).")))))
+
 (define-version "3.11.7"
   (:nsection :title "Bug fixes"
     (:ul


chaorace:
Some checklist-related notes:
1. I wasn't sure what to do about the changelog. There's no documented developer instructions and `changelog.lisp`, which seems organized by release number, leaves me scratching my head because I'm not sure whether features like these belong in semantically major or minor releases.
2. Because the current password-store interface's behavior is not documented in manual.lisp, I decided to assume that my method docstrings would generally be sufficient. What I'm _less_ sure about, however, is whether or not I need to document the newly added slot in `manual.lisp`.

chaorace:
Oh, and fair warning: I've never worked on a Common Lisp project before. Please assume that any fishy code you see is the product of inexperience and nothing particularly avant garde. All critique is welcome and highly useful.

libraries/password-manager/password-pass.lisp
```diff
@@ -9,7 +9,14 @@
    (password-directory (or (uiop:getenv "PASSWORD_STORE_DIR")
                            (format nil "~a/.password-store" (uiop:getenv "HOME")))
                        :type string
-                       :reader password-directory))
+                       :reader password-directory)
+   (scan-for-username-entries t
```
aadcg, line 13: Please indent as described [here](https://github.com/atlas-engineer/nyxt/blob/e08b99830253b120d12312a4b51b3f03d49c5709/documents/README.org?plain=1#L361).

Indeed, it is wrong in other instances in this file, but we'll fix it later.

libraries/password-manager/password-pass.lisp
```diff
@@ -9,7 +9,14 @@
...
+                              :type boolean
+                              :documentation "If t, Nyxt uses a two-step process to determine a credential's username:
+1. Open the credential file and scan for an entry with a key matching `*username-keys*'.
+2. If no username key could be found, fall back to determining username based on the credential's filename.
+
+If nil, Nyxt skips the first step and instead always uses the fallback strategy. This is useful if you don't use key-based usernames in your password store as it skips unnecessarily decrypting the credential file (if originally encrypted)."))
```
aadcg, line 19: Please ensure that the docstring doesn't go beyond 80 characters per line. See file `.dir-locals.el` and `fill-column`.

Also, please replace `"If t,"` with "`When non-nil,"`.

Applicable to all docstrings in the PR.

libraries/password-manager/password-pass.lisp
```diff
@@ -69,22 +76,48 @@ The first line (the password) is skipped."
...
+  "Select a username using the path of the credential file.
+The strategy for deriving the username is context-dependent:
+- If the credential exists in a subdirectory of the password store (e.g.: example.com/me@example.net),
+username is taken as-is from the filename (me@example.net)
+- If the credential exists in the root of the password store (e.g.: me@example.net@example.com),
+username will be the first half of the filename (me@example.net)
+
+This is meant to handle the naming patterns described in Emacs documentation
+https://www.gnu.org/software/emacs/manual/html_node/auth/The-Unix-password-store.html
+"
```
aadcg, line 89: It doesn't make sense to link the Emacs manual here. It would be more sensible to reference the official `pass` documentation or to simply refer the relevant pattern.

aadcg (commented):
@chaorace, I've left some minor comments.

Regarding the changelog, this change belongs to 3.12.0.

Regarding the manual, I don't think this piece of information belongs there. It's too specific.

libraries/password-manager/password-pass.lisp
chaorace: Removed reference link. The preceding docstring was already sufficient to describe the filename-based pattern

libraries/password-manager/password-pass.lisp
chaorace: All lines in the file directly associated with this PR have been reformatted or otherwise reworded to fit the 80 character limit

libraries/password-manager/password-pass.lisp
chaorace: Class slots are now reformatted in a manner compliant with the style doc

chaorace:
@aadcg All pending feedback should now be addressed. Please take a look when you can, thanks!

libraries/password-manager/password-pass.lisp
```diff
@@ -69,22 +79,48 @@ The first line (the password) is skipped."
 (defvar *username-keys* '("login" "user" "username")
   "A list of string keys used to find the `pass' username in `clip-username'.")
 
-(defmethod clip-username ((password-interface password-store-interface) &key password-name service)
-  "Save the multiline entry that's prefixed with on of the `*username-keys*' to clipboard.
+(defun username-from-name (password-name)
```
aadcg, line 82: Can we come up with a better name for the function and argument? It seems a bit misleading.

libraries/password-manager/password-pass.lisp
```diff
@@ -6,10 +6,20 @@
...
+    :reader password-directory)
+   (scan-for-username-entries
+    t
+    :type boolean
+    :documentation "When non-nil, a two-step process determines the username:
+1. Scan credential content for an entry with a key matching `*username-keys*'.
+2. If no match could be found, use the credential filename as a fallback.
+
+When nil, Nyxt always immediately uses the fallback strategy.
+If your store doesn't utilize username keys, this skips credential decryption."))
```
aadcg, line 22: I gave it a second thought.

The slot name needs to be named as a predicate, i.e. it must have the suffic `-p`. 

Since I am not a `pass` user please help me figure out an adequate default behavior. Does it make sense to scan for _both_ the content and the directory tree structure? Why is that better than scanning the content only, as before this PR? Am I right that you'd use this mode with this slot set to `nil`, such that the username is computed exclusively via `username-from-name`?

Upon answering, we can decide the slot name.

aadcg (commented):
@chaorace, thank you for taking a look.

Please help me to clarify some details. Thanks.

libraries/password-manager/password-pass.lisp
chaorace: Here's some candidates, do any stand out to you?
* username-from-filename
* username-from-path
* username-fallback

libraries/password-manager/password-pass.lisp
chaorace: `pass` is -- in what might be described as typical GNU fashion -- highly unprescriptive. As far as the software is concerned, there are exactly two distinct types of data:
1. The filename & path relative to the root of the store
2. The password text

Higher-level concepts, such as URLs and usernames, exist at a level of abstraction that `pass` is unconcerned with  -- note how [the manual entry](https://git.zx2c4.com/password-store/about/) contains zero mentions of the word "username". The original author's [website](https://www.passwordstore.org/) _does_ touch on the author's _personally preferred_ schema for storing extra metadata...
> One approach is to use the multi-line functionality of pass (--multiline or -m in insert), and store the password itself on the first line of the file, and the additional information on subsequent lines

... though it should be noted that this statement is immediately preceded by the disclaimer that the software itself prescribes no notion of schema whatsoever:
> The password store does not impose any particular schema or type of organization of your data, as it is simply a flat text file, which can contain arbitrary data.

I lay this all out so that I can assert a very important point: it is **impossible** to use `pass` as an internet password manager without making opinionated assumptions about how the user chooses to utilize `pass`. At best, all one can do is study the conventions invented by the surrounding `pass`-based tooling ecosystem and accomodate the most common approaches.

The most common convention is to place the website URL & username metadata directly in the path of the credential. The case for storing these like so are, respectively:
* **URL**: If the URL metadata were instead stored within the password text, then password managers would be required to decrypt every single password in the store when attempting to produce an exhausting list of credentials for any given website. This scales _very_ badly with larger password stores.
* **Username**: There exist several common use-cases where users may have multiple credentials associated with the same website (e.g.: "Single Sign On" portals). For this reason, some additional disambiguating value needs to exist in order for the pattern to accomodate the concept of multiple credential files being applicable to the same URL. True, the disambiguating value doesn't _strictly_ need to be the username... but _something_ needs to be used and usernames make for good unique identifiers.

The main point of contention here among implementations has historically been *how* these two metadata elements are organized in the credential path. Some implementations used flat directory structures while others used nested structures. Luckily, these two approaches are not mutually exclusive, so the tooling ecosystem eventually settled on supporting use of either of these patterns (*even both simultaneously!*). This dual-support convention is most canonically codified in [the Emacs documentation](https://www.gnu.org/software/emacs/manual/html_node/auth/The-Unix-password-store.html), though it should be noted that _many_ password management tools in the `pass` ecosystem expect the same patterns ([example](https://github.com/carnager/rofi-pass/blob/e77cbdbe0e885f0b1daba3a0b6bae793cc2b1ba3/rofi-pass#L528-L529), [example](https://github.com/qutebrowser/qutebrowser/blob/fb383a68816a0b45661cee8deefc07089d2dacad/misc/userscripts/qute-pass#L83-L84)).

As you may have noticed, both of the previously linked examples also support getting the username via a "field value". These "field values" are yet another invented convention with [various competing implementations](https://github.com/gopasspw/gopass/blob/74a5e62d064068a22c94f314bca081d66a6e04b6/docs/secrets.md). The general consensus for how such username keys should be named is relatively weak, so implementations which support extracting usernames via field value generally try to pack in a customizable list of commonly used key variants.

Right, so with all of that baggage out of the way, let's circle back and address your questions...

> Does it make sense to scan for both the content and the directory tree structure?

Yes, the reason for this being that it's the easiest way to support both styles without making the user do additional configuration. Indeed, there are many common GPG configurations out there (*some of them even secure!*) which do not require additional input from the user in order to decrypt credentials once the keyring (and/or hardware key) has been unlocked. For these users, simply attempting both approaches is painless and not something they would be overly concerned about.

> Why is that better than scanning the content only, as before this PR?

Because storing usernames as "field values" is no more official or correct than storing usernames as part of the file path. More importantly, _I_ use the pathname-based approach and _I_ want Nyxt to support my (_quite common!_) password store schema. The only way to logically square excluding support for this on the grounds of canonicity would be removing support for *both* schemas and always returning `""` for the username.

> Am I right that you'd use this mode with this slot set to nil, such that the username is computed exclusively via username-from-name?

Correct. This is because I use hardware-based encryption which actively withholds decryption from GPG until I manually reach across my desk and press a physical button. In other words: without this feature, I'd have to engage in this awkward ritual twice as often as would be otherwise necessary -- once (_unnecessarily!_) for the username and once again for the password. Indeed, this is one of the better justifications for _not_ storing the username as a credential field value.

libraries/password-manager/password-pass.lisp
```diff
@@ -69,22 +79,48 @@ The first line (the password) is skipped."
...
+(defun username-from-content (password-interface password-name)
+  "Select username from first entry in credential matching `*username-keys*'"
+  (let* ((content (execute password-interface (list "show" password-name)
+                           :output '(:string :stripped t)))
+         (entries (parse-multiline content))
+         (username-entry (when entries
+                           (some (lambda (key)
+                                   (find key entries
+                                         :test #'string-equal
+                                         :key #'first))
```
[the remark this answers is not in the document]
chaorace: If I put it all on the same line, that would be 87 characters long. I'll make the change... but I just wanted to point that out.

libraries/password-manager/password-pass.lisp
chaorace: As for the argument name... I don't know what to say. It's identical to the name given to the exact same value via the pre-existing `clip-username`, `clip-password` functions. I don't really have a horse in this race but I think choosing a new name for it is beyond the scope of my responsibilities and the work I'm doing for this PR.

aadcg: Since it takes a path as a string I'd suggest calling it `username-from-path` and renaming the argument to path.

Indeed, as you correctly note, it is called in `clip-username` with an argument called `password-name` (which I find odd), but that is beyond the scope of this PR. My goal is to guarantee that the added code is maintainable and easily understood in the future. Note that I am not the author of the `password-manager` library.

libraries/password-manager/password-pass.lisp
aadcg: Thanks for the detailed explanation.

I suggest the following.

Name the slot `username-scan-method`. It takes the following keywords as values: `:credential-content-and-path`, `:credential-content` and `:credential-path`. By default it is set to `:credential-content-and-path` and it follows the logic you have implemented (checks the content and fallbacks on the path approach). For your use-case, you'd set it to `credential-path`.


Does that make sense? I wonder if it makes sense to allow setting it to `:credential-content`.

aadcg (commented):
@chaorace Sorry for the late reply.

Thanks for the information shared and check my relies.

Note that your first commit isn't using the correct pathname in the commit message. Suggestion:

`libraries/password-manager/password-pass: Refactor username scan.`

libraries/password-manager/password-pass.lisp
chaorace: @aadcg Hmm... I can't imagine any hypothetical reasons anyone would want such a `:credential-content` option other than offering strict reverse-compatibility.

If you ask me, I don't personally think the added level of technical completeness really justifies the (_admitedly small_) amount of additional code complexity... but I'll happily ignore that instinct and implement the option, should you deem it important -- it's your codebase after all :wink:

aadcg: Feel free to let the implementation handle `:credential-content-and-path` and `:credential-path` exclusively. Thanks!

aadcg:
@chaorace would you like to finish the PR or should I take over it? Thanks.

aadcg:
Stale.

PR #3430, opened 2024-07-08
first attempt at bitwarden integration

ehrt74:
# Description

- Please include a summary of the change.

Fixes # (issue)

# Checklist:

- [ ] Git branch state is mergable.
  - Tests are sufficient.


aadcg:
@ehrt74  I'm really sorry that I've pushed you to open the PR since, after a closer look, it is now clear that it's only fit as an extension. That is because it calls symbols from the `nyxt` package, while classes inheriting from `password-interface` are not expected to rely on them. Also, there's no correspondent to a `password:executable`.

Note that your extension has been listed (see e08b99830).

PR #3440, opened 2024-07-25
.github/workflows/tests: Bump nix version.

aadcg:


aadcg:
The CI fails since Nix is not successfully providing built binaries for `sbcl` on `darwin`. A reliable tool it is...

aadcg:
Interesting. There are no binaries for darwin on a tagged release, but there they're available on the unstable channel. YOLO Nix, living on the edge. Good software practices, to infinity and beyond... 🚀

PR #3505, opened 2024-09-13
delete-panel-buffer calls buffer-delete-hook

kchanqvq:
# Description

delete-panel-buffer now calls buffer-delete-hook of the panel buffer

# Checklist:

- [X] Changelog is up to date (via a separate commit).
- [X] New dependencies are accounted for.
- [X] Documentation is up to date.
  - Tests are sufficient.


aadcg:
I'm not sure about this one.

I don't like the panel buffer implementation for version 4 and I'll drop it in favor of being able to have multiple buffers in a single window.

Regardless, this is a method, so it's easy to redefine it in config files or extensions.

PR #3508, opened 2024-09-16
Dispatch command hook

kchanqvq:
# Description

I apologize for the noises made last time as I didn't realize there are other places using `dispatch-command` in the code base. This version also fixes the uses in describe-key and repeat-mode.

Change `command-dispatcher` to the buffer local `dispatch-command-hook`. This makes the slot easier to be used/customized by modes, use cases include:

- per-buffer synchronous command loop
- buffer-local pre/post command hooks
- other custom processing like custom error handling

This PR also does the following:
- bug fix: `repeat-mode` was broken due to calling `last-key` on window instead of buffer.
- using hook-based API, users of `dispatch-command-hook` no longer need to assume the value of `command-dispatcher` before they modify it (they were assuming the old dispatcher was always `#'dispatch-command`). Now they just push/pop handlers onto `dispatch-command-hook`. This makes it easier to correctly recursively call commands that changes command-dispatcher, similar to Emacs' recursive edits.

# Checklist:

- [X] Changelog is up to date (via a separate commit).
- [X] New dependencies are accounted for.
- [X] Documentation is up to date.
  - Tests are sufficient.

### DIFF
--- a/source/browser.lisp
+++ b/source/browser.lisp
@@ -131,11 +131,6 @@ Which commands will they invoke next?")
     nil
     :type (maybe function)
     :documentation "The last command invoked by the user.")
-   (command-dispatcher
-    #'dispatch-command
-    :type (or sym:function-symbol function)
-    :documentation "Function to process the command processed in `dispatch-input-event'.
-Takes the function/command as the only argument.")
    (prompt-buffer-generic-history
     (make-ring)
     :documentation "The default history of all prompt buffer entries.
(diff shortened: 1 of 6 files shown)


kchanqvq:
RFC: maybe the default `dispatch-command` should be a hardcoded fallback instead of being a handler itself in `dispatch-command-hook`, so that it is
- always there and not possible to remove
- does not get `nhooks` related restarts established around it

aadcg:
Stale.

PR #3518, opened 2024-10-03
Simplify Codebase

jmercouris:
# Description

These are all fixes to simplify the codebase in preparation for improving the electron port.


jmercouris:
I am thinking the next thing I would like to focus on is the simplification of the start procedure for the browser. It is unnecessarily complicated right now. Any better suggestions?

aadcg:
@jmercouris I've pushed the relevant commits in the draft to master, thanks.

> I am thinking the next thing I would like to focus on is the simplification of the start procedure for the browser. It is unnecessarily complicated right now. Any better suggestions?

Please prioritize the issue you're assigned to, and open an issue before starting to work on a task. 

"Simplying the start procedure" is rather abstract so I'd suggest opening an issue and coming up with a more concrete action plan.

PR #3538, opened 2024-12-03
electron.lisp: Add support for undo/redo.

jmercouris:
Unfortunately we *cannot* currently check if undo/redo is possible and display a message to the user.

In the future, should we figure out a way to programmatically trigger a context menu, we'll then be able to access the `editFlags` of the renderer and inspect them to see if undo/redo is possible. I am not certain why this property is hidden behind this interface. I'll file a request on GitHub for this functionality.


jmercouris:
I've filed the request: https://github.com/electron/electron/issues/44936

Huludiabo:
I wanna join the electron build

aadcg:
Thanks @jmercouris!

PR #3540, opened 2024-12-04
Package electron port as a flatpak

### LINKED ISSUE #3534: Package experimental Electron port as a flatpak
The first alpha release of the Electron port should be a Flatpak since it is easy for users to install.  

Related to #2989.

aadcg:
# Description

Fixes #3534

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/engineer.atlas.Nyxt-Electron.yaml
+++ b/engineer.atlas.Nyxt-Electron.yaml
@@ -1,0 +1,110 @@
+app-id: engineer.atlas.Nyxt-Electron
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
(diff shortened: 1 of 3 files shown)


aadcg:
The remaining issue is that data is written to the sockets and the Flatpak container is read-only.

aadcg:
> The remaining issue is that data is written to the sockets and the Flatpak container is read-only.

That isn't the issue. The issue is that `cl-electron` is buggy and raises errors related to socket communication. We need to handle them all otherwise the compiled program will terminate.

aadcg:
The good news is that it is packaged as a flatpak. The bad news is that it doesn't work for an orthogonal reason.

PR #3270, opened 2023-12-04
Remove web extensions

aartaka:
# Description

This removes all the WebExtensions-related code from Nyxt, in favor of external `libwebextensions` (to be open-sourced separately).

Fixes #2917

# Checklist:
Everything in this checklist is required for each PR.  Please do not approve a PR that does not have all of these items.

- [X] Git hygiene:
  - I have pulled from master before submitting this PR
  - There are no merge conflicts.
- [X] I've added the new dependencies as:
  - ASDF dependencies,
  - Git submodules,
    ```sh
	cd /path/to/nyxt/checkout
    git submodule add https://gitlab.common-lisp.net/nyxt/py-configparser _build/py-configparser
    ```
  - and Guix dependencies.
- [X] My code follows the style guidelines for Common Lisp code. See:
  - [Norvig & Pitman's Tutorial on Good Lisp Programming Style (PDF)](https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf)
  - [Google Common Lisp Style Guide](https://google.github.io/styleguide/lispguide.xml)
- [X] I have performed a self-review of my own code.
- [ ] My code has been reviewed by at least one peer.  (The peer review to approve a PR counts.  The reviewer must download and test the code.)
- [X] Documentation:
  - All my code has docstrings and `:documentation`s written in the aforementioned style.  (It's OK to skip the docstring for really trivial parts.)
  - I have updated the existing documentation to match my changes.
  - I have commented my code in hard-to-understand areas.
  - I have updated the `changelog.lisp` with my changes if it's anything user-facing (new features, important bug fix, compatibility breakage).
    - Changelog update should be a separate commit.
  - I have added a `migration.lisp` entry for all compatibility-breaking changes.
  - (If this changes something about the features showcased on Nyxt website) I have these changes described in the new/existing article at Nyxt website or will notify one of maintainters to do so.
- [X] Compilation and tests:
  - My changes generate no new warnings.
  - I have added tests that prove my fix is effective or that my feature works.  (If possible.)
  - I ran the tests locally (`(asdf:test-system :nyxt)` and `(asdf:test-system :nyxt/gi-gtk)`) and they pass.

PR #3562, opened 2025-01-11
Initialize WebExtension directory.

### LINKED ISSUE #3554: Why to remove WebkitWebExtensions?
Why to remove the code which sets WebkitWebExtensions (.so ones) directory?

I currently use a default Nyxt in conjunction with [blockit](https://github.com/dudik/blockit) to block ads. Obviously, this is not possible with the original browser. Was the removal deliberate? Since Nyxt does not have its own ads blocker, this decision does not seem good for me.

Do you plan to bring it back anytime soon? Doing so it trivial, anyway.

UPD: I am complaining about master branch. 3.x.x are OK AFAIK

shamazmazum:
# Description

Initialization of WebExtensions (Webkit-specific .so plugins) was removed in aa2351. However, minimal support for them are required for useful add-ons like blockit, an ads blocker.

Fixes #3554

There was an idea (by @aadcg) to include information about blockit to the manual. I can do that too, but only if you tell me how to add custom WebKit extensions to the official nyxt build. On my system `*libdir*` is set to `/usr/local/lib` and I install blockit to `/usr/local/lib/nyxt` via `pkg`, FreeBSD's package manager. This, however, may not work for other systems.

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/source/manual.lisp
+++ b/source/manual.lisp
@@ -18,7 +18,8 @@ of Nyxt.")
   (spinneret:with-html-string
     (let ((auto-config-file (namestring (files:expand *auto-config-file*)))
           (config-file (namestring (files:expand *config-file*)))
-          (rules-file (namestring (files:expand (make-instance 'auto-rules-file)))))
+          (rules-file (namestring (files:expand (make-instance 'auto-rules-file))))
+          (gtk-extensions-directory (namestring (uiop:merge-pathnames* "nyxt/" nasdf:*libdir*))))
       (:nsection :title "Configuration"
         (:p "Nyxt is written in the Common Lisp programming language which offers a
 great perk: everything in the browser can be customized by the user, even while
@@ -923,6 +924,17 @@ core to finalize the instance."))
         (:p "Extensions are regular Common Lisp systems.")
         (:p "Please find a catalog of Nyxt extensions "
             (:a :href (nyxt-url 'list-extensions) "here") "."))
+
+      (:nsection :title "Blocking ads using AdBlock rules"
+        (:p "With WebkitGTK backend you can use "
+            (:a :href "https://github.com/dudik/blockit" "BlocKit")
+            " extension to block ads.")
+        (:p "In short, you have to install "
+            (:a :href "https://crates.io/crates/adblock-rust-server" "adblock-rust-server")
+            " to a directory visible in " (:code "PATH")
+            " environment variable and the shared library ("
+            (:code "blockit.so") ") to " (:code gtk-extensions-directory)
+            ". After that, follow instructions on BlocKit github page."))
 
       (:nsection :title "Troubleshooting"
 
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -266,7 +266,17 @@ the renderer thread, use `defmethod' instead."
                    (handler-case
                        (nth-value 1 (ensure-directories-exist gtk-extensions-path))
                      (file-error ()))))
-      (log:info "GTK extensions directory: ~s" gtk-extensions-path))
+      (log:info "GTK extensions directory: ~s" gtk-extensions-path)
+      (gobject:g-signal-connect
+       context "initialize-web-extensions"
+       (lambda (context)
+         (with-protect ("Error in \"initialize-web-extensions\" signal thread: ~a" :condition)
+           ;; The following calls
+           ;; `webkit:webkit-web-context-add-path-to-sandbox' for us, so no need
+           ;; to add `gtk-extensions-path' to the sandbox manually.
+           (webkit:webkit-web-context-set-web-extensions-directory
+            context
+            (uiop:native-namestring gtk-extensions-path))))))
     (gobject:g-signal-connect
      context "download-started"
      (lambda (context download)
--- a/source/renderer/gtk.lisp
+++ b/source/renderer/gtk.lisp
@@ -230,6 +230,12 @@ the renderer thread, use `defmethod' instead."
   ;; starting again.
   (unless nyxt::*run-from-repl-p* (gtk:leave-gtk-main)))
 
+(define-class gtk-extensions-directory (nyxt-file)
+  ((files:name "gtk-extensions")
+   (files:base-path (uiop:merge-pathnames* "nyxt/" nasdf:*libdir*)))
+  (:export-class-name-p t)
+  (:documentation "Directory to load WebKitWebExtensions from."))
+
 (define-class gtk-download ()
   ((gtk-object)
    (handler-ids
@@ -241,7 +247,8 @@ the renderer thread, use `defmethod' instead."
   (let* ((context (make-instance 'webkit:webkit-web-context
                                  :website-data-manager
                                  (make-instance 'webkit-website-data-manager)))
-         (cookie-manager (webkit:webkit-web-context-get-cookie-manager context)))
+         (cookie-manager (webkit:webkit-web-context-get-cookie-manager context))
+         (gtk-extensions-path (files:expand (make-instance 'gtk-extensions-directory))))
     (webkit:webkit-web-context-set-spell-checking-enabled context t)
     ;; Need to set the initial language list.
     (let ((pointer (cffi:foreign-alloc :string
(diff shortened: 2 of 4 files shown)


aadcg:
Thanks @shamazmazum!

> There was an idea (by @aadcg) to include information about blockit to the manual. I can do that too, but only if you tell me how to add custom WebKit extensions to the official nyxt build.

If I understood the question, note that you can include `(namestring (files:expand ...))` constructs in the manual.

shamazmazum:
> note that you can include (namestring (files:expand ...)) constructs in the manual.

@aadcg Thanks, but what if the path is GTK specific like this?

```lisp
(gtk-extensions-path (files:expand (make-instance 'gtk-extensions-directory)))
```

Should `gtk-extensions-directory` be `nyxt:nyxt-extensions-directory` (I think, this is a bad solution)?
Maybe `manual-sections` should be a generic with `:append` combinator?

I'll try the second option, but I'll be glad to hear what you think.

aadcg:
> Surely, ARGLIST was removed a while ago. I don't know if English has an idiom "time bomb". If it does, that's surely it.

Thanks for noticing it. The culprit is 425e728350e8aa63eaab50d6879c43f62fa21e25. Please don't re-introduce `arglist`, instead call `swank-backend:arglist`. I'd suggest opening another PR since it's an orthogonal topic.

aadcg:
> I'll try the second option, but I'll be glad to hear what you think.

I don't follow. My suggestion was as below. Am I missing something?

```diff
modified   source/manual.lisp
@@ -18,7 +18,8 @@ defun manual-sections
   (spinneret:with-html-string
     (let ((auto-config-file (namestring (files:expand *auto-config-file*)))
           (config-file (namestring (files:expand *config-file*)))
-          (rules-file (namestring (files:expand (make-instance 'auto-rules-file)))))
+          (rules-file (namestring (files:expand (make-instance 'auto-rules-file))))
+          (gtk-extension-directory (namestring (files:expand (make-instance 'gtk-extensions-directory)))))
       (:nsection :title "Configuration"
         (:p "Nyxt is written in the Common Lisp programming language which offers a
 great perk: everything in the browser can be customized by the user, even while
```

shamazmazum:
@aadcg 

The problem is that `gtk-extensions-directory` is in `nyxt/renderer/gtk` package and is not available in other backends. I don't want to introduce a dependency to a specific backend just for a manual section. Any ideas?

aadcg:
@shamazmazum I see now. Can we use `nasdf:*libdir*` instead?

shamazmazum:
Good idea! It's too late, my head is not working. Then, I'll remove `gtk-extensions-directory` entirely and add documentation to `*libdir*`.

aadcg:
> Good idea! It's too late, my head is not working. Then, I'll remove `gtk-extensions-directory` entirely and add documentation to `*libdir*`.

Hmm, I had the following in mind:

```diff
modified   source/manual.lisp
@@ -18,7 +18,8 @@ defun manual-sections
   (spinneret:with-html-string
     (let ((auto-config-file (namestring (files:expand *auto-config-file*)))
           (config-file (namestring (files:expand *config-file*)))
-          (rules-file (namestring (files:expand (make-instance 'auto-rules-file)))))
+          (rules-file (namestring (files:expand (make-instance 'auto-rules-file))))
+          (gtk-extension-directory (namestring (uiop:merge-pathnames* "nyxt/" nasdf:*libdir*))))
       (:nsection :title "Configuration"
         (:p "Nyxt is written in the Common Lisp programming language which offers a
 great perk: everything in the browser can be customized by the user, even while
```

shamazmazum:
@aadcg Done what you suggested. I am still not 100% content, since other paths in the manual are controlled by `nfiles`, but at least it works.

source/renderer/gtk.lisp
```diff
@@ -252,6 +259,24 @@ the renderer thread, use `defmethod' instead."
...
       (cffi:foreign-free pointer))
+    (when (and (not (nfiles:nil-pathname-p gtk-extensions-path))
+               ;; Either the directory exists.
+               (or (uiop:directory-exists-p gtk-extensions-path)
+                   ;; Or try to create it.
+                   (handler-case
+                       (nth-value 1 (ensure-directories-exist gtk-extensions-path))
+                     (file-error ()))))
+      (log:info "GTK extensions directory: ~s" gtk-extensions-path)
+      (gobject:g-signal-connect
```
aadcg, line 270: Are you sure that this needed? It has been deleted in https://github.com/atlas-engineer/nyxt/pull/3270 (a year ago).

I thought you noticed that blockit stopped working after aafb5ada67830153f49c5112f8c3f3866f9a346c.

aadcg (commented):
I've left a review @shamazmazum. Thanks.

>  I am still not 100% content, since other paths in the manual are controlled by nfiles, but at least it works.

It's not ideal but it's OK for now. It's part of a bigger issue that we'll need to address someday.

source/renderer/gtk.lisp
shamazmazum: Yes, Exactly connecting a handler to that signal is needed to tell WebKit where to find extensions:

https://webkitgtk.org/reference/webkit2gtk/2.42.2/signal.WebContext.initialize-web-extensions.html

aadcg: I see.

Then please organize the changes in 3 commits:

1. Reverts https://github.com/atlas-engineer/nyxt/commit/aafb5ada67830153f49c5112f8c3f3866f9a346c;
2. Add the handler and mention the commit that deleted it;
3. Manual changes.

I'll probably give blockit a try as well before merging.  Thanks.

source/manual.lisp
```diff
@@ -924,6 +925,18 @@ core to finalize the instance."))
         (:p "Please find a catalog of Nyxt extensions "
             (:a :href (nyxt-url 'list-extensions) "here") "."))
 
+      (:nsection :title "Blocking ads using AdBlock rules"
+        (:p "With Webkit backend you can use "
```
aadcg, line 929: /
```suggestion
        (:p "With WebKitGTK backend you can use "
```

source/manual.lisp
```diff
@@ -924,6 +925,18 @@ core to finalize the instance."))
...
+            (:a :href "https://github.com/dudik/blockit" "BlocKit")
+            " extension to block ads.")
+        (:p "In short, you have to install "
+            (:a :href "https://crates.io/crates/adblock-rust-server" "adblock-rust-server")
+            " to your " (:code "PATH")
+            " directory and the plugin itself (the needed file is called "
+            (:code "blockit.so") " on Linux/BSD systems) to "
+            (:code gtk-extensions-directory)
```
aadcg, line 937:
```suggestion
            " directory and the shared library ( "
            (:code "blockit.so") " ) to "
            (:code gtk-extensions-directory)
```

shamazmazum:
Done

aadcg:
Thanks @shamazmazum.

shamazmazum:
@aadcg BTW, did you try blockit? How do you find it?

aadcg:
@shamazmazum I didn't try it yet unfortunately. What's your feedback?

shamazmazum:
@aadcg Here are my thoughts:

Pros: Blocks all ads, popups etc. (I use ruadlist) with exception of ads embedded in youtube videos.
Cons: Spawns 3 adblock-rust-server processes (I guess Nyxt uses 3 WebKitWebViews somewhere, maybe additional Web View for a promter?) with no means to kill them (only with killall). This can be partially solved by running adblock-rust-server manually with dtach (this way you get only one process).

For me personally, its advantages definitely outweigh disadvantages.

aadcg:
Those views should correspond to the message, status and main buffer. Indeed, I'd expect them to be 4 instead of 3. The view associated with the prompt buffer is created once for every window, i.e. it is not created/destroyed on display/close.

Also, does that mean that you only use a single main buffer?

shamazmazum:
> Also, does that mean that you only use a single main buffer?

How can I check that. Also, currently I use blockit with this patch, which prevents automatic launching of adblock-rust-server:
https://github.com/shamazmazum/freebsd-ports/blob/master/www/nyxt-blockit/files/patch-blockit.c

Do you want me to reinstall it without that patch and count how many additional adblock-rust-servers are launched? :) I cannot remember, maybe it's 4.

aadcg:
> How can I check that?

Via command `list-buffers` or `switch-buffer` for example.

> Do you want me to reinstall it without that patch and count how many additional adblock-rust-servers are launched?

I think that the only critical issue here is that web extensions should not be initialized for message, status and prompt buffers. It would be easy to fix by conditionally triggering what you added in a20be046f135339dcd6935f2257e4c7e141f2303.

shamazmazum:
> Via command `list-buffers` or `switch-buffer` for example.

When I guess I have many main buffers ;)

> I think that the only critical issue here is that web extensions should not be initialized for message, status and prompt buffers

Agreed. I will rebuild blockit with that patch removed and test how that works. If it's fine, i'll sent a PR, OK?

aadcg:
Sure!

shamazmazum:
@aadcg Good news, no patch is needed ;)

I have build nyxt from master, now blockit spawns only one adblock-rust-server. Maybe it's because of this commit: 205d8f5

aadcg:
That does make sense. Thanks for taking a look!

PR #3567, opened 2025-01-12
Fix the manual

shamazmazum:
# Description

The manual is broken. Fix it as suggested by @aadcg in discussion to #3562


aadcg:
Merged via 1d3f1fa5b and 8ac08f1a3. Thanks again @shamazmazum!

PR #3571, opened 2025-01-16
electron.lisp: Add listeners for load starting and redirection.

jmercouris:
# Description

Add listeners for load starting and redirection.


aadcg:
Thanks!

PR #3572, opened 2025-01-17
Enhance/electron

jmercouris:
# Description

Continue cleaning up renderer code.


aadcg:
Merged with some changes. I use `nix` to setup the dev env for macOS.  Left `uninstall` since that part of the API is actually well-designed (even though useless, as of today).

Thanks.

PR #3574, opened 2025-01-19
Enhance/electron

jmercouris:
# Description

Refactor the API for restoring buffers, create a new endpoint to avoid misusing `ffi-buffer-make`. Depends upon: https://github.com/atlas-engineer/cl-electron/pull/54

### DIFF
--- a/source/foreign-interface.lisp
+++ b/source/foreign-interface.lisp
@@ -136,8 +136,16 @@ Usually setf-able."))
 Dispatches over `window' and classes inheriting from `buffer'.
 Usually setf-able."))
 
-(define-ffi-generic ffi-buffer-make (buffer)
+(define-ffi-generic ffi-buffer-make (browser)
+  (:method ((browser t))
+    (make-instance 'buffer))
   (:documentation "Return BUFFER and display it."))
+
+(define-ffi-generic ffi-buffer-initialize-foreign-object (buffer)
+  (:documentation "Create and configure the foreign object for a given buffer.
+This differs from `ffi-buffer-make' because it takes an existing buffer object
+and creates the foreign objects necessary for rendering the buffer."))
+
 (define-ffi-generic ffi-buffer-delete (buffer)
   (:documentation "Delete BUFFER."))
 
--- a/source/renderer/electron.lisp
+++ b/source/renderer/electron.lisp
@@ -129,6 +129,10 @@ Note that by changing the default value, modifier keys can be remapped."))
   ;; Needed for buffers whose HTML document is set via JS
   ;; (e.g. `status-buffer').
   (electron:load-url buffer "about:blank")
+  (initialize-listeners buffer)
+  (finalize-buffer buffer :extra-modes extra-modes :no-hook-p no-hook-p))
+
+(defmethod initialize-listeners ((buffer electron-buffer))
   (electron:add-listener buffer :before-input-event
                          (lambda (buffer event)
                            (on-signal-key-press-event buffer event)))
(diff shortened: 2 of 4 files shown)


jmercouris:
Arguably we don't need `ffi-buffer-make` anymore or `ffi-window-make`, however, let's not strip those from the API just yet.

source/foreign-interface.lisp
```diff
@@ -136,8 +136,13 @@ Usually setf-able."))
 Dispatches over `window' and classes inheriting from `buffer'.
 Usually setf-able."))
 
-(define-ffi-generic ffi-buffer-make (buffer)
+(define-ffi-generic ffi-buffer-make (browser)
+  (:method ((browser t))
+    (make-instance 'buffer))
   (:documentation "Return BUFFER and display it."))
+(define-ffi-generic ffi-buffer-initialize-foreign-object (buffer)
```
aadcg, line 143: Let's call it `ffi-buffer-initialize`. The rest of the naming is redundant.

source/foreign-interface.lisp
```diff
@@ -136,8 +136,13 @@ Usually setf-able."))
...
+  (:documentation "Recreate and reconfigure the foreign object for a given buffer. In other words,
```
aadcg, line 144: `Re*` isn't the verb to use. It does all of those things without "re*", even though it can be used (and its existence is justified) for those "re*" operations.

This docstring should explain why this is needed and how it differs from `ffi-buffer-make`. Also, it shouldn't mention concrete specialization objects since it defines an abstraction.

source/renderer/electron.lisp
```diff
@@ -126,9 +126,12 @@ Note that by changing the default value, modifier keys can be remapped."))
   (:documentation "Electron buffer class."))
 
 (defmethod initialize-instance :after ((buffer electron-buffer) &key extra-modes no-hook-p)
-  ;; Needed for buffers whose HTML document is set via JS
-  ;; (e.g. `status-buffer').
+  ;; Needed for buffers whose HTML document is set via JS.
```
aadcg, line 129: I think this was OK and is unrelated to the PR.

aadcg (commented):
I like the idea. As mentioned in atlas-engineer/cl-electron#54, I think that the new level of indirection belongs here.

aadcg:
> Arguably we don't need `ffi-buffer-make` anymore or `ffi-window-make`, however, let's not strip those from the API just yet.

Yes, I have made the same the same obs regarding `ffi-window-make`. And now the same holds for `ffi-buffer-make`. We should get rid of those in another PR.

source/foreign-interface.lisp
jmercouris: I disagree, otherwise it is no more specific than `ffi-buffer-make`!

source/foreign-interface.lisp
jmercouris: OK!

source/renderer/electron.lisp
jmercouris: It just bothered me that it was two lines, you are right, it is unrelated. Just being obsessive about aesthetics...

jmercouris:
I believe I've addressed all comments!

source/foreign-interface.lisp
aadcg: All `define-ffi-generic` constructs deal with foreign objects. In this case, the buffer. We just need a verb to express the idea behind the procedure.

If this PR got rid of the old `ffi-buffer-make`, then the so called `ffi-buffer-initialize-foreign-object` could be renamed to `ffi-make-buffer`. 

Proceed as you find suitable.

aadcg:
Let me know when it's ready. I'd like to take a final look before merging.

Thanks.

jmercouris:
It is ready!

source/renderer/electron.lisp
```diff
@@ -150,12 +154,15 @@ Note that by changing the default value, modifier keys can be remapped."))
...
                            (lambda (_) (declare (ignore _))
-                             (on-signal-notify-title buffer (ffi-buffer-title buffer)))))
-  (finalize-buffer buffer :extra-modes extra-modes :no-hook-p no-hook-p))
+                             (on-signal-notify-title buffer (ffi-buffer-title buffer))))))
 
-;; TODO Needed for reopen-dead-buffer.
-;; (defmethod ffi-buffer-make ((buffer electron-buffer))
-;;   (make-instance 'electron-buffer))
+(defmethod ffi-buffer-initialize-foreign-object ((buffer electron-buffer))
+  (electron::message
```
jmercouris, line 160: This is the only part where we are `breaking through` the abstraction of `cl-electron`, but I don't see another way to do it!

jmercouris: I also think it is good enough, for now.

PR #3577, opened 2025-01-20
Refactor input handling

### LINKED ISSUE #2922: Handling of `shift` in keybindings
While working on #2921 I noticed the following with respect to how we define keybindings.

``` lisp
"C-s-Y" 'nyxt/mode/download:list-downloads ;OK
"C-s-y" 'nyxt/mode/download:list-downloads ;NOP
"C-Y" 'nyxt/mode/download:list-downloads ;OK
```

### LINKED ISSUE #3560: Electron Input
If I press `Shift+alt+"."` I get the following:

```
<DEBUG> [23:58:22] Forward key "\"M-s-Period\"".
```

while this is *technically* correct, we want to see `M-s->` for our bindings to work as users expect.

We could also just add `M-s-.` to our list of bindings for common commands, and it would work, however it would be surprising when trying to bind something.

### LINKED ISSUE #3561: Rethinking keyspecs
In the context of #3560, I came to the conclusion that it would be beneficial that the values of keys don't reflect the influence of modifiers (think of Shift) or OS input method transformations.

1. `meta-shift-a` is valid, while `meta-shift-A` and `meta-A` aren't.
2. `meta-shift-1` is valid, while `meta-shift-!` and `meta-!` aren't.
3. `meta-shift-.` is valid, while `meta-shift->` and `meta->` aren't.
4. `meta-a` is valid, while `meta-ф` isn't (`a` becomes `ф` on the russian input method).

--------------

Regarding 4., it doesn't seem sensible to bind keys to different commands depending on the input method. There are smarter ways to ensure that a key is bound to a command regardless of the IM. For instance, by mapping values to codes. E.g. `C-a` becomes `C-<code-bound-to-key-a>`.

------------

Regarding 1, 2 and 3, the idea seems radical at first, but it actually isn't. 

As of today, when `meta-shift-.` is hit, Nyxt sees `meta-shift->` (which is a bogus keyspec, from the user's perspective). That why `nkeymaps` sets `keymaps:*translator*` to `translate-shift-control-combinations`. The later checks if `meta->` is bound, in case `meta-shift->` isn't. From the user's perspective, `meta->` is indeed intuitive and the key has been resolved properly.

My suggestion is simple. Nyxt should see `meta-shift-.` instead of `meta-shift->`. Since the former may not be intuitive, `keymaps:*translator*` comes to the rescue and checks if `meta->` is bound. In other words, `keymaps:*translator*` defines a map to fallback on shifted values. E.g. `shift-4` -> `$`.

aadcg:
# Description

Depends on atlas-engineer/nkeymaps#28.

Fixes #3560

# Checklist:

  - Tests are sufficient.

### DIFF
--- a/source/mode/base.lisp
+++ b/source/mode/base.lisp
@@ -33,8 +33,7 @@ This mode is a good candidate to be passed to `make-buffer'."
        "f1 s" 'describe-slot
        "f1 v" 'describe-variable
        "f11" 'toggle-fullscreen
-       "C-shift-y" 'nyxt/mode/download:list-downloads
-       "C-shift-Y" 'nyxt/mode/download:list-downloads
+       "C-Y" 'nyxt/mode/download:list-downloads
        "C-space" 'execute-command
        "C-M-space" 'execute-extended-command
        "C-shift-space" 'execute-predicted-command)
@@ -50,7 +49,6 @@ This mode is a good candidate to be passed to `make-buffer'."
        "C-pageup" 'switch-buffer-previous
        "C-w" 'delete-current-buffer
        "C-n" 'make-window
-       "C-shift-W" 'delete-current-window
        "C-W" 'delete-current-window
        "M-w" 'delete-window
        "M-c l" 'copy-url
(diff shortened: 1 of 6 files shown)


aadcg:
Still need to review more thoroughly and test against all platforms and ports.

PR #3582, opened 2025-01-23
Refactor zoom logic

jmercouris:
# Description
Fixes zoom for both GTK/Electron.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -499,7 +499,7 @@ buffer.")
     :type integer
     :documentation "The distance in pixels for `scroll-left' or `scroll-right'.")
    (zoom-ratio-step
-    0.2
+    0.1
     :type float
     :documentation "The step size for zooming in and out.")
    (zoom-ratio-default
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -498,12 +498,6 @@ buffer.")
     50
     :type integer
     :documentation "The distance in pixels for `scroll-left' or `scroll-right'.")
-   (current-zoom-ratio
-    1.0
-    :type float
-    :reader t
-    :export t
-    :documentation "The current zoom ratio.")
    (zoom-ratio-step
     0.2
     :type float
@@ -526,8 +520,11 @@ down."))
 
 (defmethod (setf current-zoom-ratio) (value (buffer buffer))
   (when (plusp value)
-    (setf (slot-value buffer 'current-zoom-ratio) value
-          (ffi-buffer-zoom-level buffer) value)))
+    (setf (ffi-buffer-zoom-level buffer) value)))
+
+(export-always 'current-zoom-ratio)
+(defmethod current-zoom-ratio ((buffer buffer))
+  (ffi-buffer-zoom-level buffer))
 
 (define-class context-buffer (buffer)
   ((last-access
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -524,7 +524,7 @@ down."))
   (:metaclass user-class)
   (:documentation "Buffers holding structured documents."))
 
-(defmethod (setf current-zoom-ratio) (value (buffer document-buffer))
+(defmethod (setf current-zoom-ratio) (value (buffer buffer))
   (when (plusp value)
     (setf (slot-value buffer 'current-zoom-ratio) value
           (ffi-buffer-zoom-level buffer) value)))
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -503,13 +503,9 @@ buffer.")
     :type float
     :documentation "The step size for zooming in and out.")
    (zoom-ratio-default
-    1.0
-    :type float
-    :documentation "The default zoom ratio.")
-   (force-zoom-ratio-default
     nil
-    :type boolean
-    :documentation "Set the zoom ratio to the default value on document load.")
+    :type (or float null)
+    :documentation "The default zoom ratio.")
    (page-scroll-ratio
     0.90
     :type float
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -506,6 +506,10 @@ buffer.")
     1.0
     :type float
     :documentation "The default zoom ratio.")
+   (force-zoom-ratio-default
+    nil
+    :type boolean
+    :documentation "Set the zoom ratio to the default value on document load.")
    (page-scroll-ratio
     0.90
     :type float
--- a/source/foreign-interface.lisp
+++ b/source/foreign-interface.lisp
@@ -229,7 +229,7 @@ PROXY-URL is a `quri:uri' and IGNORE-HOSTS a list of strings."))
   (:setter-p t)
   (:documentation "Return the zoom level of the document.
 Setf-able."))
-(defmethod (setf ffi-buffer-zoom-level) (value (buffer buffer))
+(defmethod (setf ffi-buffer-zoom-level) (value (buffer t))
   "Use JavaScript, if the renderer does not allow zooming natively."
   (ps-eval :buffer buffer
     (ps:let ((style (ps:chain document body style)))
--- a/source/renderer/electron.lisp
+++ b/source/renderer/electron.lisp
@@ -186,6 +186,13 @@ Note that by changing the default value, modifier keys can be remapped."))
   (electron:reload (electron:web-contents buffer))
   buffer)
 
+(defmethod ffi-buffer-zoom-level ((buffer electron-buffer))
+  (electron:get-zoom-factor (electron:web-contents buffer)))
+
+(defmethod (setf ffi-buffer-zoom-level) (value (buffer electron-buffer))
+  (when (and (floatp value) (plusp value))
+    (electron:set-zoom-factor (electron:web-contents buffer) value)))
+
 (defmethod ffi-buffer-evaluate-javascript ((buffer electron-buffer) javascript
                                            &optional world-name)
   ;; TODO world-name is used in user-script mode.
--- a/source/renderer/electron.lisp
+++ b/source/renderer/electron.lisp
@@ -190,8 +190,9 @@ Note that by changing the default value, modifier keys can be remapped."))
   (electron:get-zoom-factor (electron:web-contents buffer)))
 
 (defmethod (setf ffi-buffer-zoom-level) (value (buffer electron-buffer))
-  (when (and (floatp value) (plusp value))
-    (electron:set-zoom-factor (electron:web-contents buffer) value)))
+  (if (and (floatp value) (plusp value))
+      (electron:set-zoom-factor (electron:web-contents buffer) value)
+      (echo-warning "Zoom level must be a positive floating point number.")))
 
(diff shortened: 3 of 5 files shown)


source/foreign-interface.lisp
```diff
@@ -229,7 +229,7 @@ PROXY-URL is a `quri:uri' and IGNORE-HOSTS a list of strings."))
   (:setter-p t)
   (:documentation "Return the zoom level of the document.
 Setf-able."))
-(defmethod (setf ffi-buffer-zoom-level) (value (buffer buffer))
+(defmethod (setf ffi-buffer-zoom-level) (value (buffer t))
```
aadcg, line 232: Outch, this is such a crude mistake. On how earth did this ever landed? Oh right, good old f4b9788f2a3817c049fa2da323660ec02f9ccdc5.

It even gets funnier: the setter was generic (JS fallback), while the getter was the renderer's specialization. 🤡

aadcg:
I merged the commits unrelated to the zoom logic. I'd like to take a deeper look at this PR later. Thanks.

aadcg:
One PR per subject please, otherwise it's a mess and makes reviewing impossible.

I've merged commits unrelated to the zoom logic pushed to this branch via b276ca285.

source/buffer.lisp
```diff
@@ -524,10 +522,13 @@ down."))
   (:metaclass user-class)
   (:documentation "Buffers holding structured documents."))
 
-(defmethod (setf current-zoom-ratio) (value (buffer document-buffer))
-  (when (plusp value)
-    (setf (slot-value buffer 'current-zoom-ratio) value
-          (ffi-buffer-zoom-level buffer) value)))
+(defmethod (setf current-zoom-ratio) (value (buffer buffer))
+  (when (>= value 0)
```
aadcg, line 526: This is incorrect. The zoom ratio can't be 0, but rather strictly positive.

source/renderer/electron.lisp
```diff
@@ -186,6 +186,13 @@ Note that by changing the default value, modifier keys can be remapped."))
 (defmethod ffi-buffer-reload ((buffer electron-buffer))
   (electron:reload (electron:web-contents buffer)))
 
+(defmethod ffi-buffer-zoom-level ((buffer electron-buffer))
+  (serapeum:parse-float (electron:get-zoom-level (electron:web-contents buffer))))
```
aadcg, line 190: This actually belongs to `cl-electron`. It must return a float instead of a string.

source/renderer/electron.lisp
```diff
@@ -186,6 +186,13 @@ Note that by changing the default value, modifier keys can be remapped."))
...
+
+(defmethod (setf ffi-buffer-zoom-level) (value (buffer electron-buffer))
+  (when (and (floatp value) (>= value 0))
```
aadcg, line 193: Same here, only strictly positive values.

source/buffer.lisp
```diff
@@ -498,20 +498,18 @@ buffer.")
...
     :type float
     :documentation "The step size for zooming in and out.")
    (zoom-ratio-default
     1.0
     :type float
     :documentation "The default zoom ratio.")
+   (force-zoom-ratio-default
+    nil
+    :type boolean
+    :documentation "Set the zoom ratio to the default value on document load.")
```
aadcg, line 512: Under what circumstances would be user set this to `t`?

aadcg (changes requested):

source/buffer.lisp
jmercouris: Yes, yes it can be 0. Look at the implantation of renderers, they allow 0!

source/renderer/electron.lisp
jmercouris: Not true!

source/buffer.lisp
jmercouris: If the user for example wants every website they visit in every tab to be zoomed in because they run a very large display.

source/renderer/electron.lisp
jmercouris: OK!

source/buffer.lisp
aadcg: They don't and they couldn't - it's a ratio. What would be the meaning of 0 - stop displaying the view?

This is obvious but see https://github.com/electron/electron/blob/ae56a03e339f76cfdab9c77932026b47686af6f6/spec/api-web-contents-spec.ts#L1386.

source/buffer.lisp
aadcg: I see now, because we need to enact `zoom-ratio-default`. Then actually, why can't we always run `reset-page-zoom` on `nyxt:on-signal-load-finished`, as it was before?

source/buffer.lisp
jmercouris: I don't know, but I was able to render 0, so I assume it means don't change the ratio. I'll try again!

source/buffer.lisp
jmercouris: Because when the user manually changes the value of some setting, it'll erase that setting. For example, visit tomato.com, press zoom in a bunch of times. As you navigate around tomato.com, it'll stay zoomed in. However, if we add back that change, tomato.com zoom will reset on every page navigation!

jmercouris: Open firefox, visit some website, increase the zoom, navigate around, notice that it doesn't change back to the default on every page load!

aadcg: Then the solution is simple. `zoom-page` and `unzoom-page` must set `zoom-ratio-default`. This way the user can override `zoom-ratio-default` to say 1.6 in their config file, and still maintain another zoom value until the buffer is destroyed. Newly created buffers will resume the 1.6 value. Am I missing something?

jmercouris: You are missing something. That would change the zoom for ALL sites. We only want to change it on `tomato.com` we don't want to change it for everything! This allows you that flexibility. You can either zoom in ALL sites, or some sites with the current set-up.

jmercouris: The current behavior mimics that of Firefox, which is in my opinion, correct!

source/buffer.lisp
jmercouris: Wait, you are correct! I don't know how I got this wrong! I have fixed it now.

source/renderer/electron.lisp
jmercouris: You were correct, fixed!

jmercouris:
Depends upon: https://github.com/atlas-engineer/cl-electron/pull/55

jmercouris:
Actually it is not so simple, look at the documentation for `setZoomLevel`:

```
Changes the zoom level to the specified level. The original size is 0 and each increment above or below represents zooming 20% larger or smaller to default limits of 300% and 50% of original size, respectively. The formula for this is scale := 1.2 ^ level.
```

As you can see, it starts at zero! The zoom factor is what we are interested in instead it seems!

Zoom factor documentation:

```
Changes the zoom factor to the specified factor. Zoom factor is zoom percent divided by 100, so 300% = 3.0.

The factor must be greater than 0.0.
```

jmercouris:
So, it turns out that I had used zoom level when I should have used zoom factor! I have now fixed it, and everything appears to be logical and working as expected!

source/renderer/gtk.lisp
```diff
@@ -1718,7 +1718,7 @@ custom (the specified proxy) and none."
 (define-ffi-method ffi-buffer-zoom-level ((buffer gtk-buffer))
   (webkit:webkit-web-view-zoom-level (gtk-object buffer)))
 (define-ffi-method (setf ffi-buffer-zoom-level) (value (buffer gtk-buffer))
-  (when (and (floatp value) (>= value 0))
+  (when (and (floatp value) (plusp value))
     (setf (webkit:webkit-web-view-zoom-level (gtk-object buffer)) value)))
```
aadcg, line 1722: 30c8b3959decee20104f89d07bb42d30adf057d5 got it wrong big time. Bigheaded engineers who don't know what a ratio is... fun times. I'm also ignorant, but at least I don't boast myself as being enlighten. 

Not only it can't be 0 but it also doesn't need to be a float, obviously. 

Also, it shouldn't silently be a no-op if the value isn't fit. A warning should be raised instead.

Let's fix it since we're here.

aadcg, line 1722: Should it silently be a no-op if the value isn't fit? A warning should be raised instead.

Let's fix it since we're here.

aadcg:
I'd suggest renaming to `ffi-buffer-zoom-level` to `ffi-buffer-zoom-percentage`. This way it's crystal clear that it is a ratio. As we've seen "level" is defined differently in Electron and WebKitGTK.

This implies that `zoom-ratio-default` and related slots become strictly positive integers. From the user's perspective, this is also easier. For instance, currently, if they set it to `2` (instead of `2.0`), then it silently enacts as a no-op. 

If the renderer specializations require floats, we can simply `(coerce (/ 210 100) 'float)`.

source/buffer.lisp
jmercouris: > > You are missing something. That would change the zoom for ALL sites.
> 
> I don't think it would because `zoom-ratio-default` is set for each buffer instance. What you state would be true if the slot would be allocated per class instead of per instance.

It would change the zoom for all sites on a given buffer. The whole point is that the zoom is PER URL ORIGIN, not per buffer.

> > This allows you that flexibility.
> 
> That also doesn't seem to be true. If the user wants to set `zoom-ratio-default` to a non-default value, then he needs to set `force-zoom-ratio-default` to `t`. But then, as soon as he uses `(un)zoom-page`, the desired behavior you mention crumbles.

Not true!

aadcg: Does `force-zoom-ratio-default` achieve conserving zoom ratios per URL domain? I don't think so.

aadcg:
> How about just changing the names to `ffi-buffer-zoom-ratio` and avoiding doing all of the percentage manipulations?

I agree.

source/buffer.lisp
```diff
@@ -498,19 +498,13 @@ buffer.")
...
-    :documentation "The current zoom ratio.")
    (zoom-ratio-step
-    0.2
+    0.1
     :type float
     :documentation "The step size for zooming in and out.")
    (zoom-ratio-default
-    1.0
-    :type float
+    nil
```
aadcg, line 506: I don't like this idea of setting `zoom-ratio-default` to nil.

The goal is to avoid resetting the zoom on page reload, while also noting that we can't replicate the behavior of conserving zoom ratios per URL domain. Even if we could, let's be good at doing the basics.

In the current implementation, if the user sets `zoom-ratio-default` to non-nil and then (un)zooms the page, then reloading the page will would still frustrate the goal mentioned above.

Again, as I've mentioned, the only sane way forward is to conserve zoom ratios on a per-buffer fashion. Concretely:

> `zoom-page` and `unzoom-page` must set `zoom-ratio-default`. 

------------

Let me know if you'll finish this PR.

aadcg (changes requested):

jmercouris:
Opening a new PR.

aadcg:
> Opening a new PR.

Well, I've left a review here.

PR #3585, opened 2025-01-29
Fix zoom

jmercouris:
# Description

This includes fixes to the zoom.

### DIFF
--- a/source/buffer.lisp
+++ b/source/buffer.lisp
@@ -527,7 +527,7 @@ down."))
 (defmethod (setf current-zoom-ratio) (value (buffer document-buffer))
   (when (plusp value)
     (setf (slot-value buffer 'current-zoom-ratio) value
-          (ffi-buffer-zoom-level buffer) value)))
+          (ffi-buffer-zoom-ratio buffer) value)))
 
 (define-class context-buffer (buffer)
   ((last-access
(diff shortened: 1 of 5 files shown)


aadcg (changes requested):
https://github.com/atlas-engineer/nyxt/pull/3582#discussion_r1933584112

jmercouris:
I did try doing:

```
(find 'zoom-ratio-default 
            (closer-mop:class-slots (class-of (current-buffer))) 
            :key (lambda (i) (slot-value i 'sb-pcl::name)))
```

I'm just worried that it is *too* brittle, I see too many weird edge cases that can be avoided with another slot.

jmercouris:
If the user does:

```
  ((zoom-ratio 1.2)))
```

Then, the following:
```
(closer-mop:slot-definition-initform
           (find 'zoom-ratio
                 (closer-mop:class-slots (class-of (current-buffer)))
                 :key #'closer-mop:slot-definition-name))
```
Will return `1.0` instead of `1.2`. Therefore, I suggest that we make `reset-page-zoom` just set the value to `1.0` OR introduce another slot. For this reason, I've introduced the slot `zoom-ratio-reset`.

aadcg (approved):

PR #3598, opened 2025-02-09
Fix history.

jmercouris:
# Description

Strip history tree for greatly simplified and more reliable history. This solves many issues within the codebase (particularly startup).

The global history tree has shown us that it does not work, and is not worth it. This simplification is the first step in building something reliable. 

As soon as Nyxt is stable, we will resume with the per-buffer history tree. So we will bring back the old behavior before the global history tree.


aadcg:
I think that this PR is still very raw to be reviewed. But I like the direction that it is taking.

jmercouris:
@aadcg  I think it is ready for review. Please let me know if I missed anything!

nyxt.asd
aadcg, line 1: Fix deleted tests files in `nyxt.asd`.

nyxt.asd
```diff
@@ -134,14 +133,13 @@
                  (:file "buffer-listing")
                  (:file "message")
                  (:file "passthrough")
-                 (:file "history-tree")
                  (:file "list-history")
```
aadcg, line 136: This should be deleted as well.

aadcg (changes requested):
Commit c48d6d72980708d8aca2c9a60c3e207a41729279 changes git submodule by mistake it seems.

-----

I've noticed that the history file is populated by entries whose URL and title conflict with each other.

```lisp
(
...
(:url "https://en.wikipedia.org/wiki/Tomato" :title "Tomato - Wikipedia")
(:url "https://en.wikipedia.org/wiki/Tobacco" :title "Tomato - Wikipedia")
...
)
```

Which trips the user:

Screenshot From 2025-02-13 14-09-57


-------

I'm not convinced about the idea of keeping track of every URL, which results in duplicates. Instead, the data structure should be shuffled accordingly. This is important from the user perspective, as to guarantee that they see the visited URLs in a meaningful order.


------

Complete the changelog with deleted commands and other relevant API changes.

------

Ensure that custom spinneret tags don't reference deleted symbols.

------

Objects of class `history-entry` are being instantiated with deleted slots.


-------

Tests should be added as well.

------

This may not be an exhaustive review, but it a start. Thanks.

jmercouris:
I have figured out how to recreate the mismatched titles/urls. If you navigate very quickly it'll load two URLs in quick succession, the signal will be fired, but the buffer page title will be updated at a different time.

jmercouris:
> Commit [c48d6d7](https://github.com/atlas-engineer/nyxt/commit/c48d6d72980708d8aca2c9a60c3e207a41729279) changes git submodule by mistake it seems.

Fixed.

> I've noticed that the history file is populated by entries whose URL and title conflict with each other.
> 
> ```commonlisp
> (
> ...
> (:url "https://en.wikipedia.org/wiki/Tomato" :title "Tomato - Wikipedia")
> (:url "https://en.wikipedia.org/wiki/Tobacco" :title "Tomato - Wikipedia")
> ...
> )
> ```

Fixed. The only problem that can occur now is due to the architecture of our system using threads to invoke `on-signal-finished`. Because we don't necessarily handle and fire off all threads at once we can have the following situation; the user has a lot of history, they press "navigate-backwards" very very quickly. The `on-signal-finished` event will not necessarily be fired in time by `electron` for us to find out for WHICH urls the load was finished. That is because we are handling the response in Lisp. Unless we were to update `cl-electron` to add a custom handler to `did-finish-load`, I don't think we could get this data synchronously.
  
> I'm not convinced about the idea of keeping track of every URL, which results in duplicates. This duplicated information doesn't seem to be used anywhere.

It is used, in `list-history`.

> Complete the changelog with deleted commands and other relevant API changes.

Done.
 
> Ensure that custom spinneret tags don't reference deleted symbols.

Done.

> Objects of class `history-entry` are being instantiated with deleted slots.

Done.

> Tests should be added as well.

Done.

@aadcg  Please take a look again, thanks!

jmercouris:
> Please note: due to bugs on master introduced by [9c73940](https://github.com/atlas-engineer/nyxt/commit/9c739404c9bc674fcbe0214c5ef346de618964a8), you'll have to manually enable `history-mode` to test.

Note; these bugs only appear on the Electron port.

jmercouris:
The suggested fix would be:

```
From 256a7b2ae59b8ad5094d77d4581424ac86d1209f Mon Sep 17 00:00:00 2001
From: John Mercouris <john@mercouris.email>
Date: Mon, 17 Feb 2025 18:09:34 -0600
Subject: [PATCH] tmp

---
 source/renderer/electron.lisp | 8 +++++++-
 1 file changed, 7 insertions(+), 1 deletion(-)

diff --git a/source/renderer/electron.lisp b/source/renderer/electron.lisp
index ddfff6c09..c3513e21f 100644
--- a/source/renderer/electron.lisp
+++ b/source/renderer/electron.lisp
@@ -128,7 +128,13 @@ Note that by changing the default value, modifier keys can be remapped."))
   ;; Otherwise the HTML document won't be set via JS.
    (when (member (type-of buffer) '(status-buffer message-buffer prompt-buffer))
     (electron:load-url buffer "about:blank"))
-  (initialize-listeners buffer)
+  (initialize-listeners buffer))
+
+(defmethod customize-instance :after ((buffer electron-buffer) &key extra-modes
+                                                                 no-hook-p
+                                      &allow-other-keys)
+  "Make BUFFER with EXTRA-MODES.
+See `finalize-buffer'."
   (finalize-buffer buffer :extra-modes extra-modes :no-hook-p no-hook-p))
 
 (defmethod initialize-listeners ((buffer electron-buffer))
-- 
2.43.0

```

aadcg:
> > Commit [c48d6d7](https://github.com/atlas-engineer/nyxt/commit/c48d6d72980708d8aca2c9a60c3e207a41729279) changes git submodule by mistake it seems.
> 
> Fixed.

Not fixed, see 8f7eda6ee.

aadcg (changes requested):
The goal of a reviewing process is to improve the quality of a piece of work through dialogue. Obviously, it's natural to expose conflicting perspectives and reach consensus. But it is not acceptable when suggestions or remarks are ignored without justification. I wouldn't be bringing up the topic if I hadn't observed a pattern. Maybe, it all comes down  to lack of attention. Still, I'd like to kindly ask for extra care and respect for the work of the reviewer.

Going through a new round of reviews, I note that:

(1)

> > Commit [c48d6d7](https://github.com/atlas-engineer/nyxt/commit/c48d6d72980708d8aca2c9a60c3e207a41729279) changes git submodule by mistake it seems.
> 
> Fixed.

Not fixed, see 8f7eda6ee.

(2)

> (@aadcg) From the user perspective, it makes sense to see the last visited URL first. Either the data structure must be free of duplicates and shuffled accordingly, or :from-end t must be passed to remove-duplicates.

The remark above was ignored without justification. Note that I didn't press on any architectural decision. I suggested solutions that fit the bill on either setting. 

At this point, I can't avoid but feel that the first review round wasn't fully addressed. 

A note about github and its review facilities. A review was requested and I delivered it (by requesting changes). Github enables notifying that changes have been made and re-requesting a review. See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review#requesting-reviews-from-collaborators-and-organization-members.







--------

I've deleted the comments that were copied from https://github.com/atlas-engineer/nyxt/commit/9c739404c9bc674fcbe0214c5ef346de618964a8, since they're orthogonal to this PR and the issue is now fixed.

aadcg:
> > (@aadcg) I'm not convinced about the idea of keeping track of every URL, which results in duplicates. This duplicated information doesn't seem to be used anywhere.
> 
> (@jmercouris) It is used, in `list-history`.

That wasn't true at the time of my first message.

jmercouris:
> > > (@aadcg) I'm not convinced about the idea of keeping track of every URL, which results in duplicates. This duplicated information doesn't seem to be used anywhere.
> > 
> > 
> > (@jmercouris) It is used, in `list-history`.
> 
> That wasn't true at the time of my first message.

Yes, but that isn't really relevant. I wanted simply to record that it is now used!

jmercouris:
> The goal of a reviewing process is to improve the quality of a piece of work through dialogue. Obviously, it's natural to expose conflicting perspectives and reach consensus. But it is not acceptable when suggestions or remarks are ignored without justification. I wouldn't be bringing up the topic if I hadn't observed a pattern. Maybe, it all comes down to lack of attention. Still, I'd like to kindly ask for extra care and respect for the work of the reviewer.
> 
> Going through a new round of reviews, I note that:
> 
> (1)
> 
> > > Commit [c48d6d7](https://github.com/atlas-engineer/nyxt/commit/c48d6d72980708d8aca2c9a60c3e207a41729279) changes git submodule by mistake it seems.
> > 
> > 
> > Fixed.
> 
> Not fixed, see [8f7eda6](https://github.com/atlas-engineer/nyxt/commit/8f7eda6ee869ef95659db514c6b32203e2858076).

I don't know how to fix this. I keep trying to change it to latest master, and it keeps breaking. At one point it *was* fixed.


> (2)
> 
> > (@aadcg) From the user perspective, it makes sense to see the last visited URL first. Either the data structure must be free of duplicates and shuffled accordingly, or :from-end t must be passed to remove-duplicates.
> 
> The remark above was ignored without justification. Note that I didn't press on any architectural decision. I suggested solutions that fit the bill on either setting.
> 
> At this point, I can't avoid but feel that the first review round wasn't fully addressed.

I am sorry about that. I actually had a commit where I did it, but then it was lost. I'm not sure what happened to it. It is not for lack of attention, I've been working really hard on this PR, I just made a mistake.

source/foreign-interface.lisp
```diff
@@ -516,14 +528,14 @@ Dispatches on buffers and modes."))
   (:documentation "Invoked when URL loading is approved in OBJECT.
 Dispatches on buffers and modes."))
 
-(define-ffi-generic on-signal-load-finished (object url)
-  (:method ((buffer buffer) url)
+(define-ffi-generic on-signal-load-finished (object url title)
```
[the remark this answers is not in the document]
jmercouris: This is accurate. `cl-electron` MUST return the url AND title within the handler for `on-signal-load-finished`, otherwise we'll never be able to run code that can grab us that data while it is fresh. In any case, we still want the URL AND TITLE to be passed to the signal handlers so they are also not getting stale data.

aadcg:
Still some work ahead to finish this PR.

Issue #3370, opened 2024-03-27
Clicking links on internal pages spawns new window 


aadcg:
The behaviour should be the same as in the GTK port.

aadcg:
The root of the issue is that when setting the `target` attribute of anchor elements to `_blank` results in inconsistent behaviour between the WebKitGTK and Electron ports. In the former, the URL is loaded in a new buffer while, in the latter, in a new window.

Recently (see PR https://github.com/electron/electron/pull/41432), a way to tweak the behaviour has been introduced for Electron 31 but our port only works for version 29.

On the other hand, we should question whether the use of this attribute value is justified. At `bookmarks-panel` it is, otherwise the bookmarks would be opened in the panel buffer. In the context of `macro-edit-mode` too, since it seems reasonable to open a help page in a new context (be it a buffer or window). However, in the context of `:nxref`, it is questionable. Currently, a new buffer is created each time the user clicks on a "`nxref`" link. That means that it's hard to go backwards/forwards between these help pages.

I conclude that the right way to proceed is to drop the `_blank` in the context of `nxref`.

PR #3605, opened 2025-02-16
Change electron create window behavior to focus a new buffer

DominikM:
# Description

Previously, in the electron renderer, clicking a link with `target="_blank"` or calling `window.create()` from JS would result in electron opening a new window outside the scope of nyxt. We want to prevent the default electron behavior by setting our own create window handler and denying the default behavior. See https://github.com/atlas-engineer/cl-electron/pull/59

# Checklist:

  - Tests are sufficient.


aadcg:
@DominikM great work, thanks! I have alluded to the issue in https://github.com/atlas-engineer/nyxt/issues/3370#issuecomment-2176765908. Now that we're running Electron >= 31, we can address it properly.

PR #3610, opened 2025-02-23
Fix bookmark commands.

### LINKED ISSUE #3227: Review bookmark commands
As discussed in #3217, we already have `bookmark-url` which includes a `buffer-source` and completely covers the behavior of `bookmark-buffer-url`.
This means that `bookmark-buffer-url` can be deprecated/removed in 4.0.

jmercouris:
# Description

Creates a single unified interface for adding bookmarks.

Fixes #3227

### DIFF
--- a/source/mode/bookmark.lisp
+++ b/source/mode/bookmark.lisp
@@ -4,7 +4,7 @@
 (nyxt:define-package :nyxt/mode/bookmark
   (:documentation "Package for `bookmark-mode', mode to manage bookmarks.
 The main object is `bookmark-entry'. The main function to add a bookmark is
-`bookmark-add'.
+`persist-bookmark'.
 
 See the `bookmark-mode' for the external user-facing APIs."))
 (in-package :nyxt/mode/bookmark)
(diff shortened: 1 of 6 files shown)


source/mode/bookmark.lisp
```diff
@@ -409,12 +445,12 @@ Splits bookmarks into groups by tags."
 
 (define-command import-bookmarks-from-html
     (&key (html-file (prompt1
-                       ;; TODO: Is there a more intuitive directory for bookmarks?
```
jmercouris, line 412: Fix indentation.

jmercouris:
There is minor code duplication between the `bookmark` methods for the various objects, but it is absolutely minimal, and I don't think extracting some common functionality would add any clarity.

jmercouris:
Should one wish, they can now very easily add commands to shortcut bookmarking of any different type, for example, if you wanted to create a simple `bookmark-from-buffer` command, without any duplication, you could do the following:

```
(define-command bookmark-from-buffer ()
  "Prompt for objects to bookmark."
  (prompt
   :prompt "Add Bookmark(s)"
   :sources (list
             (make-instance 'buffer-source
                            :actions-on-return
                            (lambda-mapped-command bookmark)))))             
```

This could be done for any arbitrary source, plain URLs, buffers, history entries, etc.

PR #3626, opened 2025-04-11
mode/file-manager: Run URL-DECODE on a URL in OPEN-FILE

shamazmazum:
The problem this PR fixes:

If you open a file (e.g. a PDF) residing in a directory which contains spaces or unicode characters and then invoke `open-file` again, you will see the directory percent-encoded and no hints will be available.

This is after the patch is applied:
20250411_18h30m14s_grim
Currently you can see `/home/vasily/Downloads/Telegram%20Desktop`


jmercouris:
Thank you kindly for the Pull Request/Fix, I will investigate as soon as I have some time!

jmercouris:
Thank you.

PR #3686, opened 2025-08-07
fix: non-QWERTY keyboard layouts need the UI event's key property ins…

### LINKED ISSUE #3677: pre-release 4.0.0.-pre12 does not respect keyboard map in X on Linux for its keybindings in emacs-mode
Keyboard map on Linux / X-Server is not respected for keybindings in emacs-mode, US layout has to be used.

BUT only for the keybindings; in prompt buffer, the keyboard map of X-Server IS respected.

**Steps to reproduce the issue**

Using Linux on X-Server, have not tried it on Wayland.
I'm using NEO layout (neo-layout.org)
For this, switch keyboard map: in a terminal emulator, type `setxkbmap de neo`.

Start Nyxt4 ( I'm in emacs mode now, but probably CUA ist affected, too)
Try Ctrl-l for opening a page – nothing works (note: the ›l‹ is on the key labeled ›e‹ in Neo layout!)
Try Ctrl-t – that is Control plus the key labelled ›l‹ on an Englisch or European keyboard, et voilà, the prompt for opening an URL appears.

Now try to type an URL – you're in Neo layout; especially the ›l‹ key on your keyboard will produce a ›t‹ in your buffer, because Neo.

It's probably the same with Dvořak and others, too.

Hint:
If you type `setxkbmap en` (or whatever your current layout is) before switching to NEO layout, you can just go back in history to re-enable your current layout, enabling you to work under X. ;-)

**Information**
- OS Name+Version: Linux 6.12
- Installation method (Flatpak, Guix, package manager, build from source): App.Image of Pre-Release-12
- Output of Nyxt command `show-system-information`:
Nyxt version: 4
Web Renderer: Electron
OS: Linux 6.12.37-1-MANJARO
Lisp implementation: SBCL 2.4.5 (Dynamic space: 3072MB)
ASDF version: 3.3.7
Features: (:NYXT-ELECTRON :NYXT-4 :CLIPBOARD-CONTENT-METHOD :PARENSCRIPT :NSYMBOLS
 :FSET-EXT-STRINGS :NAMED-READTABLES :GLOBAL-VARS :LPARALLEL :21BIT-CHARS
 :PLUMP-UTF-32 :CLOSER-MOP :CL-PPCRE-UNICODE :CL-UNICODE :CHUNGA :FLEXI-STREAMS
 :CL-PPCRE :BORDEAUX-THREADS :LPARALLEL.WITH-CLTL2 :LPARALLEL.WITH-CAS
 :LPARALLEL.WITH-STEALING-SCHEDULER :SPLIT-SEQUENCE
 CFFI-FEATURES:FLAT-NAMESPACE CFFI-FEATURES:X86-64 CFFI-FEATURES:UNIX :CFFI
 CFFI-SYS::FLAT-NAMESPACE ALEXANDRIA::SEQUENCE-EMPTYP :FAST-IO-SV :FAST-IO
 :CL-JSON-CLOS :CL-JSON :SBCL-USES-SB-ROTATE-BYTE CHIPZ-SYSTEM:GRAY-STREAMS
 :THREAD-SUPPORT :ASDF3.3 :ASDF3.2 :ASDF3.1 :ASDF3 :ASDF2 :ASDF :OS-UNIX
 :NON-BASE-CHARS-EXIST-P :ASDF-UNICODE :ARENA-ALLOCATOR :X86-64 :GENCGC :64-BIT
 :ANSI-CL :COMMON-LISP :ELF :IEEE-FLOATING-POINT :LINUX :LITTLE-ENDIAN
 :PACKAGE-LOCAL-NICKNAMES :SB-LDB :SB-PACKAGE-LOCKS :SB-THREAD :SB-UNICODE
 :SBCL :UNIX)

cvogl:
…tead of the code property

# Description
At least for Xorg@Linux, the keydown event provides two relevant properties:
- code
- key [ Cf. https://w3c.github.io/uievents/tools/key-event-viewer.html ]

_code_ is based on „standard” QWERTY keyboard layout, whereas _key_ reflects/respects the user's keyboard layout. Prominent example for German QWERTZ layout: pressing z (located on the key for y on a QWERTY layout) leads to this situation:

 pressed  |  key  | code
-------------------------
 Z        |   z  |  KeyY
 Y        |   y  |  KeyZ

This behavior is obviously worse for Dvořak or Neo layouts and the like.

Unfortunately, just changing to the _key_ property is not feasible because _key_ does not provide the useful keycodes, e. g. `ShiftRight`. On the other hand, just adding a matching line for the previous mapping of event _codes_ lead to conflicts with special character keys remapped to alphanumeric letters, e. g. in Neo layout the letter ›y‹ is produced by the key producing a single quote on QWERTY keyboards, which would have led to lack of C-y and M-y for emacs mode keybindings.

The patch / pull request goes a middle ground by special handling of alphanumeric characters without modifying the handling of special characters. It's highly probable that it needs further special treatment for characters like / (slash) or # (hash).
It's only tested on QWERTY, QWERTZ, and Neo layouts. Further testing is necessary, esp. for various country-specific layouts (French, Italian, Greek, …), but the patch will allow international / non-QWERTY users to start using Nyxt4.

Fixes #3677 

# Checklist:

  - Tests are sufficient for Linux with Xorg. Not Wayland on Linux, not MacOS.

### DIFF
--- a/source/renderer/electron.lisp
+++ b/source/renderer/electron.lisp
@@ -526,57 +526,71 @@ height of the status/prompt/message buffer."
   (when-let ((state input-event-modifier-state))
     (mapcar (lambda (modifier) (getf (modifier-plist buffer) modifier)) state)))
 
-(defun translate-code-string (code-string)
+(defun check-alpha (str)
+  "Check if a given event key is of length 1 and an alphanumeric character"
+  (if (and (eq (length str) 1)
+           (alphanumericp (aref str 0)))
+      str
+      nil))
+
+(defun translate-code-string (code-string key-string)
   "Return string representation of a keyval.
 Return nil when key must be discarded, e.g. for modifiers."
-  (match code-string
-    ;; Compatibility layer between GDK keycode names and those of Browsers.
-    ;; https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkkeysyms.h
-    ;; https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values
-    ((or "ControlLeft" "ControlRight"
-         "ShiftLeft" "ShiftRight"
-         "MetaLeft" "MetaRight"
-         "AltLeft" "AltRight")
-     nil)
-    ("Minus" "hyphen")
-    ("Equal" "=")
-    ("Space" "space")
-    ("Enter" "return")
-    ("Escape" "escape")
-    ("Tab" "tab")
-    ("Comma" ",")
-    ("Period" ".")
-    ("Slash" "/")
-    ("Semicolon" ";")
-    ("Quote" "'")
-    ("BracketLeft" "[")
-    ("BracketRight" "]")
-    ("Backslash" "\\")
-    ("Backquote" "`")
-    ("Backspace" "backspace")
-    ("ArrowUp" "up")
-    ("ArrowDown" "down")
-    ("ArrowRight" "right")
-    ("ArrowLeft" "left")
-    ("PageUp" "pageup")
-    ("PageDown" "pagedown")
-    ("Home" "home")
-    ("End" "end")
-    ("F1" "f1")
-    ("F2" "f2")
-    ("F3" "f3")
-    ("F4" "f4")
-    ("F5" "f5")
-    ("F6" "f6")
-    ("F7" "f7")
-    ("F8" "f8")
-    ("F9" "f9")
-    ("F10" "f10")
-    ("F11" "f11")
-    ("F12" "f12")
-    ((simple-string #\K #\e #\y key-value) (string-downcase (string key-value)))
-    ((simple-string #\D #\i #\g #\i #\t digit-value) (string digit-value))
-    (_ code-string)))
+  (cond
+    ;; First case: it's an alphanumeric character
+    ;; For Compatibility with non-QWERTY keyboard layouts,
+    ;;     the event's *key* is used for comparison
+    ((check-alpha key-string)
+     (string-downcase key-string))
+
+    ;; otherwise match the event's *code* for special keys
+    ((match code-string
+       ;; Compatibility layer between GDK keycode names and those of Browsers.
+       ;; https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkkeysyms.h
+       ;; https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values
+       ((or "ControlLeft" "ControlRight"
+            "ShiftLeft" "ShiftRight"
+            "MetaLeft" "MetaRight"
+            "AltLeft" "AltRight")
+        nil)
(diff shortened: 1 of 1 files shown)


cvogl:
One thing that definitely needs testing before accepting this pull request:
behavior on keyboard layouts without latin characters (Chinese, Japanese, cyrillic, …)

I have already reverted the removal of handling of alphanumeric keys/characters from the previous commit.

simendsjo:
I don't see why we should look at `code` at all. It's more or less the scancode, and only makes sense in low-level programming. `key` should be the translated result. There are some differences, e.g. `x` rather than `KeyX`, `0` rather than `Digit0`, ` ` rather than `Space`, and maybe others too. I've handled the cases mentioned here.

Ref https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

```patch
diff --git a/source/renderer/electron.lisp b/source/renderer/electron.lisp
index ece624616..b25e74e17 100644
--- a/source/renderer/electron.lisp
+++ b/source/renderer/electron.lisp
@@ -526,10 +526,12 @@ height of the status/prompt/message buffer."
   (when-let ((state input-event-modifier-state))
     (mapcar (lambda (modifier) (getf (modifier-plist buffer) modifier)) state)))
 
-(defun translate-code-string (code-string)
+(defun translate-key-string (key-string)
   "Return string representation of a keyval.
 Return nil when key must be discarded, e.g. for modifiers."
-  (match code-string
+  (match key-string
+    ((simple-string #\Space) "space")
+    ((simple-string key) (string-downcase key))
     ;; Compatibility layer between GDK keycode names and those of Browsers.
     ;; https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkkeysyms.h
     ;; https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values
@@ -574,9 +576,7 @@ Return nil when key must be discarded, e.g. for modifiers."
     ("F10" "f10")
     ("F11" "f11")
     ("F12" "f12")
-    ((simple-string #\K #\e #\y key-value) (string-downcase (string key-value)))
-    ((simple-string #\D #\i #\g #\i #\t digit-value) (string digit-value))
-    (_ code-string)))
+    (_ key-string)))
 
 (defmethod on-signal-key-press-event ((sender electron-buffer) event)
   (when (string= "keyDown" (assoc-value event :type))
@@ -584,7 +584,7 @@ Return nil when key must be discarded, e.g. for modifiers."
                                        (when (assoc-value event :control) :control)
                                        (when (assoc-value event :alt) :alt)
                                        (when (assoc-value event :meta) :meta))))
-          (key-string (translate-code-string (assoc-value event :code))))
+          (key-string (translate-key-string (assoc-value event :key))))
       (flet ((key () (keymaps:make-key :value key-string
                                        :modifiers (input-modifier-translator sender modifiers)
                                        :status :pressed)))
```

cvogl:
Closing this for simendsjo's PR is more informed.

